PR #2, opened 2016-01-20
several scalability fixes to krun

dwightguth:
@chathhorn @edgar-pek please review

### DIFF
--- a/kore/src/main/scala/org/kframework/unparser/Unparse.scala
+++ b/kore/src/main/scala/org/kframework/unparser/Unparse.scala
@@ -75,42 +75,43 @@ object ToKast {
    * label quote from combining with the bracket,
    * as in the incorrect {@code ```_+_`(...}
    *
-   * @param b The printed representation of the term is appended to this builder.
+   * @param accumulator The function that accumulates the string to date. ie, either a StringBuilder or
+   *                    an output stream of some kind.
    * @param inParen True if this term is the leftmost within a set of brackets
    * @param prec The current precedence level
    * @param k The term to print
    */
-  def unparse(b:String=>Unit, inParen: Boolean, prec: Int, k: K): Unit = k match {
-    case KToken(s, sort) => b("#token(\"" + escape(s) + "\",\"" + escape(sort.name) + "\")")
-    case InjectedKLabel(l) => b("#klabel("+apply(l)+")")
-    case KVariable(v) => b(v.toString)
-    case KApply(l, List()) => b(unparse(inParen,l)+"(.KList)")
+  def unparse(accumulator:String=>Unit, inParen: Boolean, prec: Int, k: K): Unit = k match {
+    case KToken(s, sort) => accumulator("#token(\"" + escape(s) + "\",\"" + escape(sort.name) + "\")")
+    case InjectedKLabel(l) => accumulator("#klabel("+apply(l)+")")
+    case KVariable(v) => accumulator(v.toString)
+    case KApply(l, List()) => accumulator(unparse(inParen,l)+"(.KList)")
     case KApply(l, args) =>
-      b(unparse(inParen,l))
-      b("(")
+      accumulator(unparse(inParen,l))
+      accumulator("(")
       var first = true
       for (a <- args) {
         if (!first) {
-          b(",")
+          accumulator(",")
         } else {
           first = false
         }
-        unparse(b, false, 0, a)
+        unparse(accumulator, false, 0, a)
       }
-      b(")")
-    case KSequence(Seq()) => b(".K")
+      accumulator(")")
+    case KSequence(Seq()) => accumulator(".K")
     case KSequence(a +: items) =>
-      unparse(b, inParen, 2, a)
+      unparse(accumulator, inParen, 2, a)
       for (i <- items) {
-        b("~>")
-        unparse(b, false, 2, i)
+        accumulator("~>")
+        unparse(accumulator, false, 2, i)
       }
     case KRewrite(l,r) =>
       val needParen = prec > 1
-      if (needParen) b("``")
-      unparse(b,needParen || inParen,1,l)
-      b("=>")
-      unparse(b,false,1,r)
-      if (needParen) b("``")
+      if (needParen) accumulator("``")
+      unparse(accumulator,needParen || inParen,1,l)
+      accumulator("=>")
+      unparse(accumulator,false,1,r)
+      if (needParen) accumulator("``")
   }
 }


dwightguth:
Jenkins: test this please

dwightguth:
@chathhorn @edgar-pek should be ready for review now

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -77,13 +85,15 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
             program = parseConfigVars(options, compiledDef);
         }
 
-        program = new KTokenVariablesToTrueVariables()
-                .apply(compiledDef.kompiledDefinition.getModule(compiledDef.mainSyntaxModuleName()).get(), program);
+        // store initial configuration in single mutable reference so that we can make sure it can be garbage collected
```
dwightguth: see here for what is going on with respect to the InitialConfiguration class. Basically I am trying to reduce the length of time that the initial configuration has to live in memory. it shows up a couple other places in this diff also, all aimed at the same thing.

k-distribution/include/ocaml/lexer.mll
```diff
@@ -23,6 +23,27 @@ rule token = parse
   | eof { EOF }
 
 { 
+
```
dwightguth: this code is designed to speed up parsing of binary terms in ocaml by making lookups in the intern list happen in constant time.

k-distribution/include/ocaml/prelude.ml
```diff
@@ -9,6 +9,8 @@ type k = K.t
 exception Stuck of k
 exception Not_implemented
 
+let interned_bottom = [Bottom]
```
dwightguth: this contains some code to speed up serialization and deserialization in ocaml also

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -53,72 +57,100 @@
  */
 public class BinaryParser {
 
-    private static final byte[] MAGIC = {0x7f, 'K', 'A', 'S', 'T'};
+    public static final byte[] MAGIC = {0x7f, 'K', 'A', 'S', 'T'};
 
     public static final int BEGIN = 0, KTOKEN = 1, KAPPLY = 2, KSEQUENCE = 3, KVARIABLE = 4, KREWRITE = 5,
-            INJECTEDKLABEL = 6, END = 7;
+            INJECTEDKLABEL = 6, END = 7, BACK_REFERENCE = 8;
```
dwightguth: here I am decreasing the size of binary configurations on disk by adding some sharing.

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -53,72 +57,100 @@
...
             type = data.get();
-            scala.collection.immutable.List<K> items;
+            K[] items;
             int arity;
             switch (type) {
             case KTOKEN:
-                stack.push(KToken(readString(), Sort(readString())));
+                String s = readString();
+                String sort = readString();
+                Map<String, KToken> sortCache = ktokenCache.computeIfAbsent(sort, sort2 -> new HashMap<>());
```
dwightguth: basically the rest of this stuff including the code in org.kframework.kore.mini is designed to decrease the memory footprint of terms parsed from binary in java.

kore/src/main/scala/org/kframework/unparser/Unparse.scala
```diff
@@ -34,9 +36,13 @@ object Unparse extends {
 object ToKast {
   def apply(k: K): String = {
     val b = StringBuilder.newBuilder
-    unparse(b, false, 0, k)
+    unparse(s => b ++= s, false, 0, k)
     b.toString()
   }
+  def apply(k: K, out: PrintStream): Unit = {
```
dwightguth: this was needed so that the entire string doesn't have to live in memory at once

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -53,72 +57,100 @@
  */
 public class BinaryParser {
 
-    private static final byte[] MAGIC = {0x7f, 'K', 'A', 'S', 'T'};
+    public static final byte[] MAGIC = {0x7f, 'K', 'A', 'S', 'T'};
 
     public static final int BEGIN = 0, KTOKEN = 1, KAPPLY = 2, KSEQUENCE = 3, KVARIABLE = 4, KREWRITE = 5,
```
edgar-pek: What is BACK_REFERENCE used for?

dwightguth: a back reference is a term we've seen before.

kore/src/main/scala/org/kframework/unparser/Unparse.scala
```diff
@@ -74,37 +80,37 @@ object ToKast {
    * @param prec The current precedence level
    * @param k The term to print
    */
-  def unparse(b:StringBuilder, inParen: Boolean, prec: Int, k: K): Unit = k match {
-    case KToken(s, sort) => b ++= "#token(\"" + escape(s) + "\",\"" + escape(sort.name) + "\")"
-    case InjectedKLabel(l) => b ++= "#klabel("+apply(l)+")"
-    case KVariable(v) => b ++= v.toString
-    case KApply(l, List()) => b ++= unparse(inParen,l)+"(.KList)"
+  def unparse(b:String=>Unit, inParen: Boolean, prec: Int, k: K): Unit = k match {
```
edgar-pek: `b` is not particularly revealing name (it made sense when in the previous version when it was StringBuilder)
if I understand correctly, it accumulates string either in a StringBuilder or in output stream, so having a more telling name would help readability

kore/src/main/scala/org/kframework/unparser/Unparse.scala
```diff
@@ -34,9 +36,13 @@ object Unparse extends {
 object ToKast {
   def apply(k: K): String = {
     val b = StringBuilder.newBuilder
-    unparse(b, false, 0, k)
+    unparse(s => b ++= s, false, 0, k)
```
edgar-pek: the anonymous function in this context requires looking at what unparse is doing, it could be more readable if it was a named function. 
don't have strong feelings though -- it might be just me getting used to the code base though

dwightguth: it's not super readable I guess but that's a fault of overuse of operator overloading in Scala's standard library rather than anything in particular we've done. I'd prefer to leave this and then use java.lang.StringBuilder if this code gets reimplemented in Java which it probably will at some point.

edgar-pek:
pretty clear now -- looks good to me.

PR #5, opened 2016-01-20
add compiler patch (cf ocaml/ocaml#406)

dwightguth:
@chathhorn please review

Should shave about 200 ms off the time to kcc.

### DIFF
--- a/k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
+++ b/k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
@@ -1,20 +1,101 @@
---- a/bytecomp/matching.ml  2015-10-19 10:50:12.000000000 -0500
-+++ b/bytecomp/matching.ml  2015-10-19 14:50:56.560932567 -0500
-@@ -79,7 +79,7 @@
+diff --git a/bytecomp/matching.ml b/bytecomp/matching.ml
+index 40169c4..5b2caa5 100644
+--- a/bytecomp/matching.ml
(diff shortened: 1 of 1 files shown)


k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
```diff
@@ -1,20 +1,101 @@
...
++++ b/typing/types.ml
+@@ -300,6 +300,12 @@ and constructor_tag =
+   | Cstr_extension of Path.t * bool     (* Extension constructor
+                                            true if a constant false if a block*)
+ 
++let equal_tag t1 t2 = match (t1, t2) with
++  | Cstr_constant i1, Cstr_constant i2 -> i2 = i1
++  | Cstr_block i1, Cstr_block i2 -> i2 = i1
++  | Cstr_extension (path1, b1), Cstr_extension (path2, b2) -> path1 = path2 && b1 = b2
++  | (Cstr_constant _|Cstr_block _|Cstr_extension _), _ -> false
```
chathhorn: Why is this faster, out of curiosity? Why is the last case not "_ -> false"?

dwightguth: this is faster because polymorphic comparison is super slow. The idea behind the last case is that by writing it this way, it's not any slower once compiled, but if a new case is added to the constructor_tag type, a warning will be outputted that this function is no longer correct. I implemented it this way at the request of the OCAML developers after I submitted the original PR to them.

chathhorn:
Ok, looks good then.

PR #6, opened 2016-01-20
roll back to good commit of domains

dwightguth:
**please review**


dwightguth:
@chathhorn please review

I recognize that this isn't very nice, but I'm just trying to get a working K as quickly as possible that we can use to execute RV-Match. When the time comes to deal with unification again,  we are probably going to reimplement much of what is there now.

chathhorn:
So you're adding the hook and removing it in the Java backend? Clearly I have a pretty superficial understanding about what's going on here, but I guess it looks good to me if you're comfortable with it.

PR #8, opened 2016-01-21
fix equals check on klabel

dwightguth:
@edgar-pek please review

We  are normalizing the klabel so that the equals check will work correctly.


dwightguth:
It's not really ideal, but some klabels are concrete klabels and some are variables. We basically want to only construct a concrete klabel if it was already concrete.

chathhorn:
Looks good with the usual caveat of my very superficial understanding about what's going on.

PR #9, opened 2016-01-25
make better error message for ambiguity

dwightguth:
This makes the warning when a rule parse is ambiguous much easier to read.


dwightguth:
@edgar-pek please review

I commented out the tests that broke because relying on the output of a toString method is pretty highly fragile, and the tests in question test a part of the code that has worked essentially unchanged for quite some time now, so I don't expect us to have to do further development on it in the near future.

edgar-pek:
It looks like the only thing missing are the attributes, e.g. 
`...[Location(#token("4",Int),#token("18",Int),#token("4",Int),#token("20",Int)) Source(#token("src/test/resources/convertor-tests/syntaxWithPriorities.k",KString)) klabel(#token("x",KString))]`
if that is ok, then it looks good to me.

PR #10, opened 2016-01-26
add a couple cases to make dealing with maps easier + add a few check…

dwightguth:
…s for rewrites

@edgar-pek please review

Basically we have two small changes here. The first, in prelude.ml, adds a couple cases for chaining map lookups together to make it easier to deal with bindings that aren't in a map when dealing with nested maps. The second adds some error checks for K that were missing at the point we split.


edgar-pek:
looks good to me.

PR #11, opened 2016-02-02
Uuid

dwightguth:
@edgar-pek please review

Basically adding a function to K that returns UUIDs


k-distribution/include/ocaml/prelude.ml
```diff
@@ -217,6 +216,8 @@ let get_exit_code(subst: k Subst.t) : int = match (Subst.fold (fun k v res -> ma
   | [Int i], Some _ -> failwith "Bad exit code pattern"
   | _ -> res) subst None) with Some i -> i | _ -> failwith "Bad exit code pattern"
 
+let () = Random.self_init ()
```
edgar-pek: How is this related to UIUD?

dwightguth: it was needed in an earlier version of this diff and I forgot to take it out.

edgar-pek:
got it. looks good.

PR #15, opened 2016-02-25
add Map type of cell collection

dwightguth:
@edgar-pek please review

Basically we are adding a new type of cell collection, Map, and ensuring that we can distinguish it from a Bag, and that we have the correct type for map items.


edgar-pek:
looks good.

PR #26, opened 2017-12-08
Fix the $JAVA environment variable.

virgil-serbanuta:
When running a script in the bin directory that sources checkJava (e.g. k-bin-to-text, which sources setenv, which sources checkJava), $0 is in the bin directory, making checkJava set the classpath to bin/java instead of lib/java.


dwightguth:
question: does this fix the issue where k-bin-to-text doesn't work without a kserver active? it would probably be good to fix that issue before we release iele.

traiansf:
Yes, I think that is the one of the issues it fixes; at least that's what I tried.

PR #30, opened 2018-02-12
add support for cell maps (mostly copy-pasta from rv-match)

andreistefanescu:


andreistefanescu:
this is just copy-pasta from rvk, with the exception keeping sorted variables. Since the java backend squashes all map klabels/sorts together, the output is technically incorrect, but readable.

dwightguth:
I tested this with the EVM semantics proofs and it seems to get stuck on the rule for SLOAD. The resulting configuration seems to not contain a map in the account cell, which I suspect is the problem. Can you please investigate and try to fix?

andreistefanescu:
sure, it's probably because the specifications themselves are not transformed into lookups...

andreistefanescu:
I'm thinking something like this, how do I test on the evm semantics?

dwightguth:
You can check out the evm semantics repo, run `git submodule init; git submodule update; cd .build/k; git fetch; git checkout cell_map; mvn package; cd ../..; make ocaml-deps; make; make proof-test`

andreistefanescu:
I checked, and the proofs run now, after changing `type="Bag"` into `type="Map"`

grosu:
Cool, thanks @andreistefanescu !  So are we done with this now?  @dwightguth does this require any changes to the K tutorial / exercises?  Let's talk on slack about what else is needed in order to have only one version of the EVM semantics, both for execution and for verification.

daejunpark:
I'll test this on the verification examples by updating specs.

daejunpark:
@dwightguth @andreistefanescu not sure about the usage of this repo. should i make another PR in rv-match via cloning this?

daejunpark:
@andreistefanescu Could you try to run the proofs on the `verification` branch of the evm-semantics repository? I found that this version failed to prove the specs.

daejunpark:
@grosu no, it's not done yet. I guess it is only tested against the master branch of the evm-semantics, where the `verification` branch should have been used. We will merge the verification branch soon btw.

andreistefanescu:
@daejunpark did you change `type="Bag"` into `type="Map"`? It fails with `type="Bag"` because the existing code tries to wrap the cells into some `BagItem`, I didn't look to much into it...

More generally, this might break because it does not translate yet the sort of `Bag` variables, to `Map`. However, the variable representing `...` don't have any sorts, so there is no way for me to translate. This would need fixing somewhere in kore before this step.

daejunpark:
@andreistefanescu ah sorry, my mistake. let me try again with it.

daejunpark:
@andreistefanescu the existing proofs work with the change of `type="Map"` as you suggested. Thanks! But, once we uncomment the following to capture the other irrelevant `<account>` cells, it still fails. Could you please take a look at it?
```
          </account>
       // ... // TODO: fix
        </accounts>
```

Indeed, the reason why we asked to support the bag-as-map-cell feature is to fix the above, where we thought that the kprover can benefit from the fact that all of the keys of the map are different. SLOAD leads to an infeasible path without knowing that.

PR #32, opened 2018-02-26
fix top-constraint issue

daejunpark:
@andreistefanescu This is a fix for another problem that a function is not evaluated in the initial term where it should be using the current constraint.

Can you think a better fix?  I mean, it could be fixed by adding `context.setTopConstraint(data.constraint);` in a proper place that you might miss last time.
For example, can we make `initialTerm.expandPatterns(true);` resolve such a function?


andreistefanescu (approved):
looks good! I don't think there is a nicer way to handle the top constraint without a refactoring.

daejunpark:
should be subsumed by #53

PR #34, opened 2018-02-28
pass the sort of the cell collection variables to the java backend

andreistefanescu:


andreistefanescu:
@daejunpark please check if this fixes the issue with `...` in `<accounts>` in the proof specs. I tried `tests/proofs/hkg/approve-spec.k` locally, and it works.

dwightguth:
Jenkins: test this please

dwightguth (approved):

dwightguth:
Jenkins: test this please

daejunpark (approved):

daejunpark:
@andreistefanescu @dwightguth I have tested this fix, cherry-picking this on the top of the proofmodule branch, and adding the `...` in the `<accounts>` cell of the spec. It works well! Thanks!!

dwightguth:
looks like some kind of bug in the OCAML backend was introduced by this change to SortCells. I will investigate tomorrow.

dwightguth:
So I found a fix for the issue in the OCAML backend and I can apply that when the rest of the diff is ready, but there is some kind of much deeper bug involved here. I don't understand exactly what is going on but this change makes the java backend's associative matching algorithm break very badly, and I was able to track it down to the fact that it changes the behavior of line 73 of AssocCommToAssoc.scala, but based on that information, I have no idea how it worked previously, and my attempts at a simple fix produced very bizarre behavior, so I have to give this up. @andreistefanescu will have to investigate this further. I am going to rerun the tests now so you can see that ktest fails.

dwightguth:
Jenkins: test this please

PR #37, opened 2018-03-08
Escaping back ticks and backslashes in Kast

traiansf:
Do not review yet


dwightguth (changes requested):
I would rather we use String.replace rather than relying on scala magic as it is likely to compile down to much better bytecode.

dwightguth:
looks like you may need to update some unit tests.

traiansf:
@dwightguth is it ok now?

dwightguth (approved):

PR #39, opened 2018-03-09
[WIP] Support for Opam 2.0

h0nzZik:
Do not merge this yet, does not work with opam 1.2.

Some linux distributions (including Fedora 27) ship with new version of opam, i.e. opam-2.0, instead of opam-1.2. This PR aims to support opam-2.0.

One problem is that the archive https://github.com/ocaml/ocaml/archive/4.03.tar.gz" does not contain ocaml-4.03.0, but ocaml-4.03.1 (with a bug fix), and when building the patched ocaml version the build complains that the versions do not match. This is fixed in 0d3e59d5d14eb607a5de596492f5b7d66e584931. Imho it should work even with opam-1.2, but I have not tried that yet, since I do not currently have any system with opam-1.2.

The patch 9b9f1387dbad987685702fe4f084727460455545 simply applies the changes described in the [migration guide](https://opam.ocaml.org/doc/2.0/Upgrade_guide.html). The changes break the compatibility with opam-1.2. One possible solution is to check the version of opam dynamically, another is is to create four scripts:
* k-configure-opam
* k-configure-opam-2.0
* k-configure-opam-dev
* k-configure-opam-2.0-dev

What would you suggest?


dwightguth:
Jenkins: add to whitelist

h0nzZik:
@dwightguth What do you think about having two different versions of `k-configure-opam*` scripts, one for opam 1.x and the other for opam 2.0?

h0nzZik:
Closed in favour of https://github.com/kframework/k/pull/183.

PR #43, opened 2018-03-26
add ac unification for maps with at most one map variable

andreistefanescu:
summary: add AC unification for maps when both the subject and the pattern consist of multiple entries and at most one map variable. The algorithm combines the entries in all possible ways, and the some of the combinations may become unsatisfiable later in the simplification process. To enable AC unification, the conversion to map lookups it turned off. The `--ac-unification` flag must be passed to both `kompile` and `kun`.

test plan: breakpoint to check that the 13 solutions are correct for the test file


ehildenb (changes requested):
Would a more appropriate name be `mapUnification` instead of `acUnification`? I don't want future users to be confused about what the engine does/doesn't support.

Also, when you say "test-plan", do you mean that you intend to implement tests? I see one example file of a specific use-case for map unification, but nothing turning that into a test on the repository. I think if we're adding this feature we will need more examples to test on, and an automated way to run the tests with the K test-set.

If you like, I can write more tests for you (also exercising other uses of map unification).

ehildenb:
@andreistefanescu I have added a few things.
First, a hook in `MetaK` which exposes a `#unify` function, which returns a BuiltinSet of unifiers (with each unifier represented as a BuiltinMap).

I have also added a directory `k-distribution/tests/unification`, with a simple testing harness (written in bash). Not sure how to integrate it into the existing testing framework, but I've included the K file, the "program" input (a bunch of unification problems), as well as the expected output. You can run `./test.sh` to run the single `basic.test` test.

I would like you to look over my code for the function `unify` in `MetaK`, which exposes the unification to the user. One problem I think it has is that it does not use the `FastRuleMatcher` codepath, which means that your Map unification code may not be present there. Perhaps it should be adapted to use the `FastRuleMatcher` code.

The reason for exposing a function to the user and specifying the tests at the K level is so that we can re-use these tests on the K-Haskell backend. It's also easier to write them this way than as a bunch of K proofs (which could have other things going wrong with them instead of just having problems with the unification).

ehildenb:
Jenkins: test this please

ehildenb:
@andreistefanescu I've also exposed the `matchable` function to the user. Could you please review? I need this functionality for SBC.

ehildenb:

dwightguth:

PR #47, opened 2018-04-02
Fix a bug introduced in constraint simplification

yzhang90:
I found a bug when prover was simplifying the ConstrainedTerm. 

Here is the initialConfig:
```
<k>... </k>
<balance> BAL </balance>
requires 0<=Int BAL andBool BAL <Int pow256
```

After several steps, it becomes the intermediateConfig :
```
<k> ... </k>
<balance> R_BAL </balance>
requires BAL ==K R_BAL andBool 0<=Int R_BAL andBool R_BAL <Int pow256
```

When the intermediateConfig matches and applies the following rule:
```
rule <k> #transferFunds ACCT ACCT VALUE => . ... </k>
         <account>
           <acctID> ACCT </acctID>
           <balance> ORIGFROM </balance>
           ...
         </account>
      requires VALUE <=Int ORIGFROM
```

the finalConfig becomes:
```
<k> ... </k>
<balance> R_BAL </balance>
requires BAL ==K R_ORIGINFROM andBool R_BAL ==K R_ORIGINFROM andBool 0<=Int R_ORIGINFROM andBool R_ORIGINFROM <Int pow256
```
which is OK. 
However, in SymbolicRewriter.java (L656 - L662), the kprover tries to simplify the constraint further by removing the bindings of `Sets.difference(cterm.constraint().substitution().keySet(), initialTerm.variableSet()))`. The substitution contains `R_BAL -> R_ORIGINFROM` and `BAL -> R_ORIGINFROM` and the initialTerm contains `BAL`. Thus, `R_BAL ==K R_ORIGINFROM` is removed from the constraint. As a result, `R_BAL` still appears in the term but there is no constraint regarding this variable.
After discussing with Daejun, we came up with a fix which will still keep the constraint. 
Another thing that came to my mind is  why in the finalConfig, `R_BAL` is not substituted by `R_ORIGINFROM`.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -658,7 +658,7 @@ public class SymbolicRewriter {
                             cterm.constraint().removeBindings(
                                     Sets.difference(
                                             cterm.constraint().substitution().keySet(),
-                                            initialTerm.variableSet())),
+                                            Sets.union(initialTerm.variableSet(), cterm.term().variableSet()))),
                             cterm.termContext());
                     if (visited.add(result)) {
                         nextQueue.add(result);


yzhang90:
jenkins: test this please.

yzhang90:
jenkins: test this please.

yzhang90:
jenkins: test this please.

yzhang90:
jenkins: test this please.

daejunpark (approved):

daejunpark:
I approve, but would like to have @andreistefanescu take a look as well.

andreistefanescu (changes requested):
The cause of the bug is that the variables in the substitution should have been eliminated already, and thus this modification should not be necessary. I'll have to dig into it, how can I reproduce?

yzhang90:
semantics:
```
module TEST-SYNTAX

   syntax Exp ::= "#call"
                | "#checkcall"
                | "#step1"
                | "#transfer"
                | "#step2"
                | "#end"

endmodule


module TEST
    imports TEST-SYNTAX
    imports DOMAINS

    configuration <k> $PGM:Exp </k>
                  <balance> 0 </balance>


    rule #call => #checkcall ~> #transfer ~> #step2


    rule <k> #checkcall => #step1 ... </k>
         <balance> BAL </balance>
      requires BAL >=Int 0

    rule <k> #step1 => . ... </k>
         <balance> BAL </balance>
      requires BAL >=Int 0

    rule <k> #transfer => . ... </k>
         <balance> ORIGINFROM </balance>
      requires ORIGINFROM >=Int 0

    rule <k> #step2 => #end ... </k>
         <balance> BAL </balance>
      requires BAL >=Int 0

endmodule
```

spec:
```
module TEST-SPEC
    imports TEST

    rule <k> #call => #end </k>
          <balance> BAL </balance>
      requires BAL >=Int 0

endmodule
```

The spec can not be proven because of the reason I mentioned above.

daejunpark:
close via #49

PR #52, opened 2018-04-22
Wasm support

ehildenb:
@dwightguth this is almost ready. What I have not figured out is how to add a compilation pass which inserts the `DEFAULT-LAYOUT` import in case there is nothing defined of sort `Layout`. Also, you mentioned we should use some special name for the `Layout` sort, so it wouldn't clash with users definitions. What was your suggestion again?

Anyway, I'd like some early feedback on this if you have time.


ehildenb:
@cos you may also want to take a look and see if it makes sense.

k-distribution/include/builtin/kast.k
```diff
@@ -196,6 +196,12 @@ module RULE-LISTS
 
 endmodule
 
+module DEFAULT-LAYOUT
+    syntax Layout ::= r"(\\/\\*([^\\*]|(\\*+([^\\*\\/])))*\\*+\\/)"
```
dwightguth: This should be #Layout.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -97,6 +97,18 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
       .groupBy(_.sort)
       .map { case (l, ps) => (l, ps) }
 
+  lazy val layouts: Set[String] =
+    productionsForSort
+      .get(ADT.Sort("Layout"))
```
dwightguth: Can we add a constant to Sorts for this?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -97,6 +97,18 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
...
+      .getOrElse(Set[Production]())
+      .collect({
+          case Production(_, Seq(Terminal(terminalString)), _)           => terminalString
```
dwightguth: You need to escape special characters here.

dwightguth (changes requested):
You can put the code for dealing with the special implicit import in DefinitionParsing.java.

k-distribution/include/builtin/kast.k
ehildenb: Looking at `kore/.../Sorts.scala`, it uses `K_` prefix instead of `#`. Should we pick `KLayout` instead for consistency?

kore/src/main/scala/org/kframework/definition/outer.scala
ehildenb: Well, I added print statements in `Scanner.java` to make sure the correct strings were making it through given the above `DEFAULT-LAYOUT` module. Basically, I left in the old `singleLine` and `multiLine` constants, and printed them side-by-side with the generated `layout`, and saw that they were the exact same. Perhaps string handling has already happened by the time it's dumped into `Terminal` and `RegexTerminal` by the parser?

k-distribution/include/builtin/kast.k
dwightguth: No

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth: I mean, what you have is fine for the regex terminals, which is what you tested, but if for example I have the terminal ".*", It needs to match the dot followed by the asterisk, not any character.

k-distribution/include/builtin/kast.k
dwightguth: We use hash prefixed sorts for anything that is not considered to be a subsort of sort K. The KString example is simply a very old sort name that predates that convention.

kore/src/main/scala/org/kframework/definition/outer.scala
ehildenb: Ah, I see what you mean. Hmmmm, I'll have to look and see if there is a way to interpret an entire string as literals instead of a regex.

ehildenb:
@dwightguth I have changed to `#Layout`, and added the constant `Layout` to `Sorts.scala`, and have made the program parser look for `layouts` and import `DEFAULT-LAYOUT` if it's not there.

Two remaining things: add tests and fix the escaping string sequences for `Terminal`. Now, I've been thinking about it, and I wonder if we should just not allow `Terminal` for it. It could be much simpler, and having `Terminal` Layout doesn't make much sense. Either way, they could always embed their constant string as a Regex.

ehildenb:
Also, @dwightguth can you look at dcd9581 and let me know if it's correct? I'm confused why it wasn't this way before (that is, with regards to the `this._` prefixes), but will not question the ways of Java.

ehildenb:
@dwightguth this should pass now. Note that I had to insert commits marked `sss`, which force the layout to always have whitespace. Perhaps there is a way to avoid this, but I do not know Flex well enough to make it work. It seems acceptable to me. If you approve those, then I think this is fully ready.

Note that the first two commits (topologically) also make some other cleanups/changes that I ran into while working on this, so make sure to double-check them.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -148,10 +157,18 @@ public Module getProgramsGrammar(Module mod) {
...
 
+            Set<Module> modules = new HashSet<Module>();
+            modules.add(newMod);
+
             // import PROGRAM-LISTS so user lists are modified to parse programs
-            scala.collection.Set<Module> modules = Set(newMod, baseK.getModule(PROGRAM_LISTS).get());
+            modules.add(baseK.getModule(PROGRAM_LISTS).get());
+
+            // check if `#Layout` has been declared, import `DEFAULT-LAYOUT` if not
+            if (mod.layouts().size() == 0) {
```
dwightguth: We want to use `mod.definedSorts(). contains(Sorts.Layout)` here

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Scanner.java
```diff
@@ -39,10 +39,18 @@
...
+
     public Scanner(ParseInModule module) {
-        tokens = KSyntax2GrammarStatesFilter.getTokens(module.getParsingModule());
-        scanner = getScanner();
-        this.module = module.seedModule();
+        this.tokens  = KSyntax2GrammarStatesFilter.getTokens(module.getParsingModule());
+        this.module  = module.seedModule();
+        String whites = "[\\ \\n\\r\\t]";
+        if (this.module.layout().length() == 0) {
+            this.layout = whites;
```
dwightguth, line 49: Can you help me understand what went wrong without this?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -97,6 +98,17 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
       .groupBy(_.sort)
       .map { case (l, ps) => (l, ps) }
 
+  lazy val layouts: Set[String] =
+    productionsForSort
+      .get(Sorts.Layout)
+      .getOrElse(Set[Production]())
+      .collect({
+          case Production(_, Seq(RegexTerminal(_, terminalRegex, _)), _) => terminalRegex
+          case p => throw KEMException.compilerError("Productions of sort `Layout` must be exactly one `RegexTerminal`.\nProduction: " + p.toString())
```
dwightguth, line 107: Should be #Layout

dwightguth (changes requested):

kore/src/main/scala/org/kframework/definition/outer.scala
ehildenb: It is `#Layout` defined as `val Layout = Sort("#Layout")` in `Sorts.scala`. Are you saying to also change the Scala/Java identifier name?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Scanner.java
ehildenb: Well, the symptom of the problem was that with no layout present, it would throw `Array access out of bounds` errors on line 285 for the tests in `ParserTest.java` (which does not go through `RuleGrammarGenerator.java`, and so does not get default layout added).

Originally it was throwing Flex errors, because the layout line being added to Flex was `() ;\n`, which it did not like. So I added the check that the layout was empty to `Scanner`, and did not put the line in unless it was non-empty. This fixed the Flex errors that were being thrown, but led into the array out of bounds access errors. As a test, to see if it was because it was expecting that line to be there in Flex, I changed the behaviour of layout to *always* include the whitespace at least, so that we would have something to put. It seems to have worked, fixing the tests in `ParserTest` while allowing custom comment layout (but throwing away whitespace).

One more thing to note: I peeked at the generated Flex files, and noticed that now that we have sort `#Layout` in the actual grammar, it's adding rules for those regexs to produce tokens in sort `#Layout`. This might give us a different approach to the problem, where we let the Scanner produce the tokens of sort `#Layout` for us, and we do a post-processing pass to throw them away. This might make it easier to eventually handle white-space sensitive languages like Python in the future.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth: Oh is this some kind of fancy string interpolation in scala I'm not aware of?

ehildenb: No, I don't think so. It's just a constant called `Layout` defined in `Sorts.scala`, and that constant has value `Sort("#Layout")`.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Scanner.java
dwightguth: Unfortunately it's simply not feasible for us to always treat whitespace as layout, even in a non-indent-sensitive language. A language like visual basic uses newline to delineate statements, but doesn't use indentation to delineate blocks, which means that this issue is the only blocking issue preventing us from defining parsers for such languages in K. Given that this is clearly some kind of underlying bug, I'd like to try to figure it out.

ehildenb: I should clarify, line 285 of `Scanner.java` was the one throwing the error, on the tests in `ParserTest`.

ehildenb:
Continued in #57

PR #53, opened 2018-04-24
evaluate spec rules with requires constraints when initializing

daejunpark:
Fix the initial evaluation of spec rules.

For example,

when we have the following spec rule:
```
module A-SPEC
  rule a(f(X)) => b
    requires X >Int 0
endmodule
```
now, the functional term `f(X)` is evaluated under the constraint `X >Int 0`.

This is needed when this spec rule is used to prove another spec rule.


daejunpark:
@andreistefanescu please review

andreistefanescu (approved):

PR #55, opened 2018-04-27
add [proj] attribute

daejunpark:
Add `[proj]` attribute, which will implicitly insert the projection lemma, that is, for all `f` with the `[proj]` attribute, we have:
```
f(X,0,N) == f(Y,0,N) /\ ... /\ f(X,N-1,N) == f(Y,N-1,N)
  iff
X == Y
```

Example.

Spec:
```
module D-SPEC
imports D

rule <k> call(X) => X ... </k>
     <calldata> _ => _ </calldata>
     <flag> 0 => 1 </flag>

rule <k> run => X ... </k>
     <calldata> bytes2(nth(X,0,2), nth(X,1,2)) </calldata>
     <flag> 0 => 1 </flag>
     [trusted]

endmodule
```

Semantics:
```
requires "domains.k"

module D
imports INT

configuration <T>
  <k> $PGM:Cmd </k>
  <calldata> bytes2(0,0) </calldata>
  <flag> 0 </flag>
</T>

// commands

syntax Cmd ::= call(Int)
             | "run"

rule <k> call(X) => run ... </k>
     <calldata> _ => split(X) </calldata>

rule <k> run => merge(CD) ... </k>
     <calldata> CD </calldata>

// utility functions

syntax Bytes ::= bytes2(Int, Int)

syntax Bytes ::= split(Int) [function]

syntax Int ::= merge(Bytes) [function]


// abstractions

syntax Int ::= nth(Int, Int, Int) [function, proj]

rule split(X) => bytes2(nth(X,0,2), nth(X,1,2))
     
rule merge(bytes2(nth(X,0,2), nth(X,1,2))) => X

endmodule
```

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
@@ -155,6 +155,24 @@ public class ConstrainedTerm extends JavaSymbolicObject {
             }
         }
 
+        // evaluate/simplify equalities
+        context.setTopConstraint(data.constraint);
+        for (Equality equality : constraint.equalities()) {
+            Term equalityTerm = equality.toK();
+            Term evaluatedTerm = equalityTerm.evaluate(context);
+            if (!evaluatedTerm.equals(equalityTerm)) {
+                constraint = constraint.addAll(Collections.singletonList(evaluatedTerm));
+                if (constraint == null || constraint.isFalse()) {
+                    return null;
+                }
+            }
+        }
+        context.setTopConstraint(null);
+        constraint = constraint.simplifyModuloPatternFolding(context);
+        if (constraint.isFalse()) {
+            return null;
+        }
+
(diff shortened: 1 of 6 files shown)


daejunpark:
@andreistefanescu please take a look. It may or may not be merged, depending on what you will have.

daejunpark:
Also, when unifying two functional terms, like `#unsigned(X)` and `#unsigned(Y)`, it generates an unification equality `#unsigned(X) =? #unsigned(Y)`. Now such equalities are simplified by applying lemmas, like `rule #unsigned(X) ==K #unsigned(Y) => true requires ...`.

daejunpark:
@andreistefanescu Please review. Note that there are still rooms for improving performance. But I'd like to go further once you confirm that this is the right direction to solve the issue.

daejunpark:
@dwightguth could you please merge this?

dwightguth (approved):

dwightguth:
Good enough for now since it only touches the java backend.

PR #57, opened 2018-05-01
Custom layout

ehildenb:
More appropriate name for #52 .


API/pom.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright (c) 2015 K Team. All Rights Reserved. -->
+<!-- Copyright (c) 2018 K Team. All Rights Reserved. -->
```
dwightguth: these should become 2015-2018, etc throughout

dwightguth (changes requested):
I have one request, but you should be able to implement it procedurally.

ehildenb:
@dwightguth I think I got all of the copyrights correct this time. I also ran `dos2unix` on all files in the repo, there were some with funky line-endings.

dwightguth (approved):
this looks really good and is an awesome new feature we've wanted for a while, thanks!

PR #61, opened 2018-05-07
[WIP] Support for debugging stuck configurations

h0nzZik:
When a function evaluation gets stucked, the user is given only the stucked term. With this patch the user is given a full stacktrace - a chain of terms ranging from the stucked term to the full configuration.

An example:
```bash
$ cd k-distribution/tests/ocaml-backtrack
$ kompile arith.k
$ krun inputs/input_side_stuck3.txt

< Stack trace >

Level: 3
isNice(#token("0", "Int"))

Level: 2
isPretty(#token("0", "Int"))

Level: 1
isBeautiful(#token("0", "Int"))

Level: 0
`<T>`(`<k>`(`_/__ARITH-SYNTAX`(#token("9", "Int"), #token("0", "Int"))))

</Stack trace >

<T>
  <k>
    9 / 0
  </k>
</T>
```
Notes:
* RV-Match needs a small patch to work with this.
* I haven't measured the performance impact. 
* TODO maybe add a switch to `kompile` to enable / disable this feature?
* TODO move the tests to appropriate directory and automate them

@dwightguth Could you please review this?
@chathhorn This is the feature I was talking about recently.


k-distribution/include/ocaml/run.ml
```diff
@@ -58,52 +78,80 @@ let schedule_next_thread (module Def: Plugin.Definition) (step_function: k -> (k
...
     match (try let res,func = (step_function config) in Step(res,func) with Stuck c -> NoStep c) with
     | Step (config,(StepFunc step_function)) -> (
         match schedule_next_thread (module Def) step_function false config thread_pool with
-          | None -> (config, n)
+          | None -> ([config], n)
           | Some (config, step_function, thread_pool) -> take_steps (module Def) step_function thread_pool config depth (n+1)
     )
-    | NoStep config -> (
+    | NoStep stack -> (
+      check_stack stack;
```
dwightguth, line 91: this doesn't seem right. you are potentially throwing a stuck exception at a point in the code where the rewriting is not necessarily stuck.

k-distribution/include/ocaml/run.ml
```diff
@@ -58,52 +78,80 @@ let schedule_next_thread (module Def: Plugin.Definition) (step_function: k -> (k
...
     | Step(config, StepFunc step_function) -> take_steps_no_thread (module Def) step_function config depth (n+1)
-    | NoStep config -> (config, n)
+    | NoStep stack -> (stack, n)
   )
 
-let rec strat_run (module Def: Plugin.Definition) (config: k) (depth: int) (n: int) : k * int =
-  let c0,threads = split_config config in
-  let c0 = context_switch c0 (List.hd threads) in
-  let c1,n1 = take_steps (module Def) Def.step (init_thread_pool (List.tl threads)) c0 depth n in
+type stepper_func = (k * int) -> (k * int)
```
dwightguth, line 107: I'm not comfortable with how complicated this segment of code is. You're introduced a whole bunch of new functions without it being clear why a pull request concerned primarily with error reporting should be altering the main rewriting code path so significantly. I'd prefer these changes to be localized to the portion of the code dealing with stuck configurations.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -453,29 +454,32 @@ public String interpreter() {
...
+        sb.append("with Stuck(stack2) -> output_string out (print_k_binary (hd (rev stack2))); exit 112\n");
         return sb.toString();
     }
 
     private void matchPatternAndPrint(StringBuilder sb) {
         sb.append("let _ = (try print_subst_binary subst (try_match res res (-1)) with Stuck c -> output_string subst \"\\x00\\x00\\x00\\x00\")\n");
     }
 
+    private void ocamlPrintKStacktrace(StringBuilder sb) {
+		sb.append("let () = if (length stack) <> 1 then print_string (print_stacktrace stack) \n");
```
dwightguth, line 479: I'm not sure I'm really comfortable with this being dumped on stdout without any kind of control happening here. It breaks the architecture of krun to be printing stuff to stdout like this. I'd feel a lot better if the terms were serialized to the java frontend and pretty printed there.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -453,29 +454,32 @@ public String interpreter() {
...
+    }
+
     private void runAndPrint(int depth, StringBuilder sb) {
-        sb.append("let res, steps = ");
+        sb.append("let stack, steps = ");
         constructRun(sb, String.valueOf(depth));
         sb.append("\n");
+        ocamlPrintKStacktrace(sb);
+        sb.append("let res = hd (rev stack)\n");
```
dwightguth, line 487: you seem to be doing this a lot, and it seems to me it could be very expensive in the case of a stack overflow. You should probably set a limit on how many stack frames get serialized, and also keep the bottom of the stack at the front of the list so you don't have to pay a linear cost.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -1260,7 +1267,7 @@ public void apply(KApply k) {
                         sb.append("| \"").append(sort.name()).append("\" -> (");
                         KLabel freshFunction = mainModule.freshFunctionFor().apply(sort);
                         encodeStringToFunction(sb, freshFunction);
-                        sb.append(" ([Int counter]) config (-1))\n");
+                        sb.append(" ([Int counter]) config [config] (-1))\n");
```
dwightguth, line 1288: again, I'm not comfortable with us changing the primary code path of execution so substantially just to introduce more error reporting. If you need to construct the stack trace when you reach a stuck configuration, use a series of try blocks.

dwightguth (changes requested):
this pull request has serious structural flaws that need to be addressed before I am comfortable with code that introduces functionality like this being merged. The biggest problems are that you seem to be breaking major abstraction boundaries in order to introduce this functionality, and that you seem to be substantially altering the primary code path of execution in order to introduce a change that should have no bearing on correctly executing semantics. The code needs to be seriously rewritten before it can be merged.

h0nzZik:
@dwightguth Ok, I would like to do it the right way now. What do you think about the following approach? When function evaluation gets stuck, the method [`org.kframework.rewriter.Rewriter.execute`](https://github.com/kframework/k/blob/master/kore/src/main/scala/org/kframework/rewriter/Rewriter.scala#L23)  throws an exception containing the stacktrace, and the exception is caught and pretty-printed in [`org.kframework.krun.KRun.run`](https://github.com/kframework/k/blob/master/kernel/src/main/java/org/kframework/krun/KRun.java#L87-L93).

Another question is how would the OCaml backend return the stacktrace back. I suggest adding a new line with an `ok`/`error` flag to the [output file](https://github.com/kframework/k/blob/master/ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java#L131-L138). If the function evaluation gets stuck, the file will contain the `error` flag and instead of single configuration there would be a `List` of them.

@chathhorn FYI

dwightguth:
One thing you could do if you wanted is just to construct a term of sort List instead of a list of terms of various sorts. So if `foo()` called `bar(1)` which called `1 /Int 0`, then you would return `ListItem(foo()) ListItem(bar(1)) ListItem(1 /Int 0)` from the ocaml backend.

dwightguth:
If you do that, you don't necessarily even have to print it in any special way, but of course you could, since any term returned by the backend which was of sort List would be known to be an error.

h0nzZik:
Yes, I meant I would print `ListItem(LastConfiguration) ListItem(bar(1)) ListItem(1 /Int 0)`. I still prefer to have some `ok`/`error` flags - I think that in principle the backends could do any kind of term rewriting, and a term of sort `List` is still a valid term, so it does not indicate an error.

h0nzZik:
Closed in favour of https://github.com/kframework/k/pull/167.

PR #64, opened 2018-05-08
don't compile all parsed modules as entry modules in prover

dwightguth:
Unfortunately the modules in the spec file could be part of a specification and therefore contain unbound variables on the right hand side and other features not supported in axioms, which means that it is not in general safe to compile them as part of the definition being proved. Fortunately, we don't actually need to do this, so this PR removes that functionality.


ehildenb:
#57 fixes the issue with checkstyle.

ehildenb:
@dwightguth does this work with the smart-contracts-verification repo? The only reason I could think that it doesn't is because extra modules are added after the fact for those proofs (the lemmas modules).

dwightguth:
I tested it with the master branch of the evm repo and all the proofs pass. I'm pretty sure that's because the modules that are actually used are part of the dependency graph of the main module and don't actually need to be added separately.

traiansf (approved):
seems reasonable; hopefully we will never need to compile (some of) those modules.

PR #66, opened 2018-05-14
Rule coverage tool for K

dwightguth:
For a sample coverage report, see https://office.runtimeverification.com/jenkins/job/iele-semantics-pull-request/cobertura

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -739,7 +739,7 @@ module K-IO
                                   String) // module name
                                   [function, hook(IO.parseInModule), impure]
 
-  rule #open(S:String) => #open(S:String, "rw")
+  rule #open(S:String) => #open(S:String, "r+")
 
   syntax Int ::= "#stdin"   [function]
                 | "#stdout" [function]
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -757,6 +757,17 @@ module K-IO
   syntax KItem ::= #system ( String ) [function, hook(IO.system), impure]
                  | "#systemResult" "(" Int /* exit code */ "," String /* stdout */ "," String /* stderr */ ")"
 
+  syntax K ::= #appendToFile(String, String) [function]
+             | #lockOpenFileThenAppendAndClose(Int, String) [function]
+             | #appendToOpenFileThenClose(K, Int, String) [function]
+             | #thenUnlockAndClose(K, Int, String) [function]
+             | #thenClose(K, Int) [function]
+  rule #appendToFile(Path:String, Text:String) => #lockOpenFileThenAppendAndClose(#open(Path, "a"), Text)
+  rule #lockOpenFileThenAppendAndClose(FD:Int, Text:String) => #appendToOpenFileThenClose(#lock(FD, 0), FD, Text)
+  rule #appendToOpenFileThenClose(.K, FD:Int, Text:String) => #thenUnlockAndClose(#write(FD, Text), FD, Text)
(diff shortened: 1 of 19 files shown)


k-distribution/include/builtin/domains.k
```diff
@@ -755,6 +757,17 @@ module K-IO
   syntax KItem ::= #system ( String ) [function, hook(IO.system), impure]
                  | "#systemResult" "(" Int /* exit code */ "," String /* stdout */ "," String /* stderr */ ")"
 
+  syntax K ::= #appendToFile(String, String) [function]
+             | #lockOpenFileThenAppendAndClose(Int, String) [function]
+             | #appendToOpenFileThenClose(K, Int, String) [function]
+             | #thenUnlockAndClose(K, Int, String) [function]
+             | #thenClose(K, Int) [function]
+  rule #appendToFile(Path:String, Text:String) => #lockOpenFileThenAppendAndClose(#open(Path, "a"), Text)
```
darko-marinov, line 765: Does this open and close the file for each write? Why not collect all coverage during execution and write once at the end? Maybe it's already done; I can't quite read K.

dwightguth: it's just for simplicity's sake. if we opened the file once at the beginning and wrote everything at the end, we would either have to implement backend-specific logic for the coverage, which we don't currently have to do, or else we would have to modify the configuration and add a cell containing the file descriptor, which would make the implementation much messier. I am aware this isn't particularly great for performance, but it seems to only be a factor of ~2 slower, which isn't so bad for something that doesn't happen in production.

darko-marinov: Fair enough; 2X is fine for a coverage tool. When K gets faster, this may need to be revised.

kernel/src/main/java/org/kframework/compile/ExpandMacros.java
```diff
@@ -46,11 +57,23 @@
...
+    private final PrintWriter coverage;
+    private final FileChannel channel;
 
     public ExpandMacros(Module mod, KExceptionManager kem, FileUtil files, GlobalOptions globalOptions, KompileOptions kompileOptions) {
         this.kem = kem;
         this.mod = mod;
+        this.cover = kompileOptions.coverage;
+        files.resolveKompiled(".").mkdirs();
         macros = stream(mod.rules()).filter(r -> r.att().contains("macro")).sorted(Comparator.comparing(r -> r.att().contains("owise"))).collect(Collectors.groupingBy(r -> ((KApply)RewriteToTop.toLeft(r.body())).klabel()));
+        try {
```
traiansf: maybe guard this by cover, too?

kernel/src/main/java/org/kframework/compile/GenerateCoverage.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+import static org.kframework.kore.KORE.*;
+
+public class GenerateCoverage implements AutoCloseable {
+    private final boolean cover;
+    private final FileUtil files;
+    private final PrintWriter allFiles;
```
traiansf: rename this to `allRulesFile` or something?  `allFiles` sounds strange

traiansf (approved):
Looks good to me.  I assume you do it through files rather than maps or similar stuff to make it independent from the backend.

traiansf:
BTW, thanks for asking for my review.  We need to do something similar for KORE (although there we already have the coverage from `stack`, but we still need to make it available for `Jenkins`)

PR #68, opened 2018-05-18
Improve performance of coverage tool

dwightguth:
As it turns out, the overhead on complex programs of the prior mechanism for measuring coverage was too substantial for me to be able to efficiently compute the coverage of the EVM semantics, which I am trying to do so I can get that data to IOHK.

In order to solve this, I rename #appendToFile to #logToFile and give it a hook IO.log which maintains a mapping from paths to string buffers and appends to the files only after rewriting ends.

Furthermore, it was discovered that the rule instrumentation used previously made it so that tail recursive K functions did not compile down to tailcalls in OCAML, which led to stack overflows occurring. Here we introduce a minor change to the OCAML backend in the form of a new attribute indicating a function that always returns .K, which is compiled down to a let statement instead of a list append statement, preserving the tailcall.

Finally, in the course of testing this PR I was required to test the behavior of a program on the latest version of OCAML. The bug was still present in that version, so I worked around it, which is why I am not updating the ocaml versions at this time, but in the interests of saving time if we do end up upgrading at some point in the future, I am checking in the changes to OPAM switches and patches required for OCAML 4.06.1 compatibility with K.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -757,16 +757,7 @@ module K-IO
   syntax KItem ::= #system ( String ) [function, hook(IO.system), impure]
                  | "#systemResult" "(" Int /* exit code */ "," String /* stdout */ "," String /* stderr */ ")"
 
-  syntax K ::= #appendToFile(String, String) [function]
-             | #lockOpenFileThenAppendAndClose(Int, String) [function]
-             | #appendToOpenFileThenClose(K, Int, String) [function]
-             | #thenUnlockAndClose(K, Int, String) [function]
-             | #thenClose(K, Int) [function]
-  rule #appendToFile(Path:String, Text:String) => #lockOpenFileThenAppendAndClose(#open(Path, "a"), Text)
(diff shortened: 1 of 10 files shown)


traiansf (approved):
Looks good to me.  I have two questions, though:
(1) Did you consider flushing the log more often, say every 1000 rewrites?
(2) it seems the changes you make to the Ocaml base / libraries are legit.  Are we submitting these patches to the Ocaml / Opam upstream?  If we do, did we get any reply from them?  If not, why not / should we?

BTW, some tests are failing :-)

ehildenb:
Will the logging work in the java backend? Should it be documented if it only works on the OCaml backend?

dwightguth:
I haven't tried flushing more often but also it doesn't seem to be an issue...

The patch is definitely submitted upstream, see https://github.com/ocaml/ocaml/pull/1786

As for the java backend, it will work if I add the hook, do you want me to do that? it doesn't really strike me as necessary...

bmmoore (approved):
Looks good.

PR #69, opened 2018-05-23
Add parameters to Sort and KLabel

dwightguth:
Do not review yet.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
The tests fail because of a race condition in the C semantics that requires either fixing or an updated reference file. Note that all the other rv-match tests pass and `mvn verify` succeeds from within the K module. So I think this is ready for review and to be merged despite the failing test. @chathhorn note that you will have to deal with this issue next time you update the K submodule after this has been merged, or you will get test failures. I tracked the issue down to a race in showQualifiers between outputting "const volatile" and outputting "volatile const", but I didn't want to spend time on a proper fix since that's Chris's responsibility.

traiansf (approved):
Looks good to me. Actually I think the code looks much cleaner in the new version.

I was wondering, though, whether it would make sense to have a stand-alone test for parameterized labels and sorts.

dwightguth:
@bmmoore does not appear to be working today, so I'm going to merge this, but he can provide feedback after the fact and I'll make a followup PR if necessary.

PR #85, opened 2018-05-29
Exponentiation

### LINKED ISSUE #84: No support for exponentials in rule requirements
I've problems with `^Int` expressions in an evm-semantics proof. They can't be translated into z3 and therefore the `impliesSMT` function crashes: this blocks from having expressions with `^Int` in require statements of lemmas:
e.g. I've a precondition a the circularity rule looking like this:
```...
X ^Int (2 +Int N) <Int pow256
1 <=Int X
X <Int pow256
...
```

at one point in the proof I multiply X*X which leaves `chop(X *Int X)` on the stack
in order to reason about it i have the following lemma:
```
rule chop(X *Int X) => X *Int X
  requires X ^Int (2 +Int N) <Int pow256
```
Since the actual proof is for an implementation of fixed-point exponentiation by squaring I don't see how omitting `^Int` expressions in requirements is possible.

dwightguth:
Fixes #84


traiansf (approved):
LGTM

bmmoore (approved):
Looks good to me too.

PR #86, opened 2018-05-29
update install instructions to fix bug on debian stable

dwightguth:
This fixes a bug in the installation on debian, because unlike ubuntu, in debian, flex does not depend on libfl-dev, even though K has a dependency on that library. Thus we manually add the dependency to the list.


traiansf (approved):
LGTM

PR #90, opened 2018-06-01
Sbc support

ehildenb:
Branch seems stable enough that merging should work out.


ehildenb:
Will re-open when if I need merged at some point.

PR #92, opened 2018-06-04
backport fix for latest gcc into 4.03.0+k switch

dwightguth:
Tested by uninstalling my opam switch, running k-configure-opam followed by mvn verify.

Part of a fix for kframework/evm-semantics#220

Ready for review.


traiansf:
@dwightguth: code looks OK, but I'm not sure I'm qualified to review this;  what does that signal business do?

traiansf (approved):

PR #93, opened 2018-06-05
backport another patch relating to new gcc versions

dwightguth:
Tested on Arch Linux with gcc 8.1.

Ready for review.

### DIFF
--- a/k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
+++ b/k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
@@ -20,6 +20,37 @@ index 7d54160..5a51708 100644
      let fail,jmps =
        scan_def
          []
+diff --git a/byterun/debugger.c b/byterun/debugger.c
+index 5e61584..55ccf62 100644
+--- a/byterun/debugger.c
++++ b/byterun/debugger.c
+@@ -161,6 +161,7 @@ static void winsock_cleanup(void)
+ void caml_debugger_init(void)
+ {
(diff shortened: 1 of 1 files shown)


k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
```diff
@@ -20,6 +20,37 @@ index 7d54160..5a51708 100644
...
+     sock_domain = PF_UNIX;
+     sock_addr.s_unix.sun_family = AF_UNIX;
++    a_len = strlen(address);
++    if (a_len >= sizeof(sock_addr.s_unix.sun_path)) {
++      caml_fatal_error("Debug socket path length exceeds maximum permitted length");
++    }
+     strncpy(sock_addr.s_unix.sun_path, address,
+-            sizeof(sock_addr.s_unix.sun_path));
++            sizeof(sock_addr.s_unix.sun_path) - 1);
++    sock_addr.s_unix.sun_path[sizeof(sock_addr.s_unix.sun_path) - 1] = '\0';
```
traiansf, line 46: If you fail with `fatal_error` above, why is this change still needed? Is  this just for safety, in case the above is just a warning?

traiansf (approved):

k-distribution/src/main/scripts/lib/opam/compilers/4.03.0/4.03.0+k/files/opam-compiler.patch
dwightguth: It's not really my code, I'm just backporting a change from the trunk of ocaml.

PR #94, opened 2018-06-07
Basic reflexion

andreistefanescu:
@ehildenb this should probably merged into #90 , but I want to get feedback first


ehildenb:
@andreistefanescu I think one problem we're going to run into (though it failed before hitting this), is that the OCaml backend won't have hooks for these functions. For other similar changes I have to my local K, the OCaml backend complains about compiling definitions that use KREFLECTION.

I propose moving the symbolic reasoning stuff to its own file. We can rename `unification.k` to `symbolic.k`, and then have the `UNIFICATION` module in there, and also a `NARROWING` module in there where this functionality is added. That way people using the OCaml backend can still use the `REFLECTION` module.

@dwightguth what do you think?

ehildenb:
@andreistefanescu had a brief discussion with @dwightguth and I think we will try moving forward with this. #96 makes this change of unification.k => symbolic.k.

dwightguth:

PR #96, opened 2018-06-08
k-distribution/: rename unification.k => symbolic.k

ehildenb:
Generalization of the file `unification.k` to allow other symbolic reasoning.


dwightguth:
I'm okay with renaming this, but you should probably keep the old filename for backwards compatibility.

ehildenb:
@dwightguth do you mean have a symlink, or just re-purpose the file without renaming?

dwightguth:
I mean just add a file with the old file name that requires the new file name.

ehildenb:
@dwightguth done

ehildenb:
Jenkins: test this please

PR #97, opened 2018-06-09
JSON Output

ehildenb:
Enables JSON serialization output option. Should have some test cases. Figure I should make a JSON reader as well, to enable other tools modifying K terms and wanting to invoke K on them.

Also need intermediate state dumping as well at some point.


ehildenb:
@dwightguth if you could provide a preliminary review of this, I would appreciate it.

I still need to add tests. My strategy for that is to have a bag of terms that I feed into both the Binary serializer and the Json serializer and read back in and check that it's the identity (so that as new tests are added, both serializers are tested).

dwightguth:
the code looks roughly okay but I haven't looked at the serializer/deserializer classes in detail because it's a little hard to visualize without an example json term for comparison. Can you provide me with an example of a term in the format you supported?

kernel/src/main/java/org/kframework/unparser/ToBinary.java
```diff
@@ -94,6 +94,7 @@ private void traverse(K k) throws IOException {
             add_intern(k);
             data.writeInt(seq.size());
 
+        // TODO is this case subsumed by line 84?
```
dwightguth: no, line 84 is a klabel variable, this is a variable of a sort subsorted to K.

ehildenb:
@dwightguth I've taken your comment into account, but I'm closing this until I've done more local testing. I want to make sure that the version 1 of the JSON serialization/deserialization at least meets @mhhf and @mrchico needs before merging.

PR #98, opened 2018-06-11
add log2Int function to prelude

dwightguth:
This is much faster than bitshifting manually, which is why I added it.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
@@ -99,6 +99,18 @@ public class BuiltinIntOperations {
         return IntToken.of(term.bigIntegerValue().abs());
     }
 
+    public static IntToken log2(IntToken term, TermContext context) {
+        BigInteger val = term.bigIntegerValue();
+        if (val.compareTo(BigInteger.ZERO) <= 0)
+            return null;
+        int log2 = 0;
+        while (val.compareTo(BigInteger.ONE) > 0) {
+            val = val.shiftRight(1);
+            log2++;
+        }
+        return IntToken.of(log2);
+    }
+
     public static BoolToken eq(IntToken term1, IntToken term2, TermContext context) {
         return BoolToken.of(term1.bigIntegerValue().compareTo(term2.bigIntegerValue()) == 0);
     }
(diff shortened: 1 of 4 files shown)


ehildenb:
@dwightguth can you add an implementation for the Java backend as well?

ehildenb:
I'm testing with KEVM now.

ehildenb:
Cannot approve until we pass tests on https://github.com/kframework/evm-semantics/pull/222.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
```diff
@@ -99,6 +99,18 @@ public static IntToken abs(IntToken term, TermContext context) {
         return IntToken.of(term.bigIntegerValue().abs());
     }
 
+    public static IntToken log2(IntToken term, TermContext context) {
```
traiansf, line 102: Code seems legitimate to me.  However, why wouldn't `bitLength()` work here?

k-distribution/include/ocaml/prelude.ml
```diff
@@ -835,6 +835,9 @@ struct
   let hook_min c _ _ _ _ = match c with
       [Int a], [Int b] -> [Int (Z.min a b)]
     | _ -> raise Not_implemented
+  let hook_log2 c _ _ _ _ = match c with
```
traiansf, line 838: similarly here, it seems you could be faster by using number of limbs, limb size, and computing the bit length of the most significant limb.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
dwightguth: bitLength ceilings the log2, which is not the behavior we want, we want it floored.

k-distribution/include/ocaml/prelude.ml
dwightguth: The performance is more than good enough with this as is.

dwightguth: I could maybe do what you suggest but it would be more fragile, so I'm inclined to leave it.

ehildenb (approved):
KEVM tests pass, which exercises this code path. Still want to hear from @traiansf on the issues above.

traiansf (approved):
If there is no issue of performance, I have no further comments.

PR #99, opened 2018-06-14
bitRange and sigExtendBitRange

dwightguth:
This implements an efficient way of getting out a range of bits in linear time in the number of bits to be extracted. We don't use Z.signed_extract because it is O(i) and not O(len).

IELE will use this to more efficiently implement chop, byte, twos, and sext. EVM can use it as well for similar purposes.


traiansf:
@dwightguth I'm pretty sure you got everything right, but could you add tests for it, esp for the conditions you have in case of overflow?

ehildenb:
@dwightguth I second @traiansf on writing some tests, also https://github.com/kframework/evm-semantics/pull/229 is failing, so looks like our initial idea of how to map the EVM operators to these new ones was not quite correct.

dwightguth:
@traiansf this is ready for re review

traiansf (approved):

ehildenb:
@dwightguth what happens if the second parameter is negative? The third parameter? Should we have tests?

dwightguth:
I'm fairly sure what happens is the ocaml backend will throw an exception and execution will halt. This is not the first function to possess this attribute, either. I don't think we want tests for this case.

ehildenb:
Looks like we run into it in KEVM: https://fsl-ci.cs.illinois.edu/#/builders/3/builds/269/steps/6/logs/stdio

dwightguth:
I mean, I'm still not persuaded that's a reason to change the behavior of the hook... a negative bit offset is pretty much meaningless. I could at best make the function not throw an exception and simply return the stuck function term, but that still would not get you the behavior you actually want in the EVM test... You are better off changing the semantics.

ehildenb:
I'm changing the semantics, just wanted to link you to what I'm looking at now/what prompted the question.

ehildenb (approved):
Approving with evidence of it working for KEVM at: https://office.runtimeverification.com/jenkins/job/evm-semantics-pull-request/1006/console

PR #102, opened 2018-06-20
Fix regression in java backend

dwightguth:
Ready for review.


traiansf (approved):

ehildenb (approved):
Fixes my issues with https://github.com/kframework/kat.

We'll just have to remember as we switch to the Haskell backend that the Java backend does not support parametric KLabels.

PR #103, opened 2018-06-20
add Bytes sort to K

dwightguth:
Adds a new sort, Bytes in module BYTES to domains.k and an efficient implementation of this type in the OCAML backend, as well as a fully axiomatized-in-K implementation of the type for use in the Java backend and other symbolic reasoning. The intent is to replace some of the instances of WordStack in the iele semantics with this new type in order to improve performance.

Ready for review.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

k-distribution/include/builtin/domains.k
```diff
@@ -590,6 +590,98 @@ module STRING-BUFFER
...
+  imports BYTES-SYNTAX
+  imports INT
+  imports K-EQUAL
+  imports STRING
+  imports STRING-BUFFER
+
+  syntax Bytes ::= "nilBytes"
+                 | Int ":" Bytes
+
+  rule .Bytes => nilBytes
```
ehildenb, line 626: Why do we have two different constants for the empty byte-string? (`nilBytes` and `.Bytes`)

k-distribution/include/builtin/domains.k
```diff
@@ -590,6 +590,98 @@ module STRING-BUFFER
...
+  syntax Int ::= lengthBytes(Bytes, Int) [function, klabel(lengthBytesAux), smtlib(lengthBytesAux)]
+  rule lengthBytes(BS) => lengthBytes(BS, 0)
+  rule lengthBytes(nilBytes, SIZE) => SIZE
+  rule lengthBytes(B : BS, SIZE) => lengthBytes(BS, SIZE +Int 1)
+endmodule
+
+module BYTES
+  imports BYTES-SYNTAX
+  imports BYTES-SYMBOLIC
+  imports INT
```
ehildenb, line 674: Is this `INT` import needed?

k-distribution/include/builtin/domains.k
```diff
@@ -590,6 +590,98 @@ module STRING-BUFFER
...
+  rule Int2Bytes(I, E, Unsigned) => Int2Bytes((log2Int(I) +Int 8) /Int 8, I, E)
+    requires I >Int 0
+  rule Int2Bytes(0, _::Endianness, _) => .Bytes
+  rule Int2Bytes(I, E, Signed) => Int2Bytes((log2Int(I) +Int 9) /Int 8, I, E)
+    requires I >Int 0
+  rule Int2Bytes(I, E, Signed) => Int2Bytes((log2Int(~Int I) +Int 9) /Int 8, I, E)
+    requires I <Int -1
+  rule Int2Bytes(-1, E, Signed) => Int2Bytes(1, -1, E)
```
ehildenb, line 682: Why are these here and not in `BYTE-SYMBOLIC`?

ehildenb (commented):
Overall looks ok, have a couple of questions (note: I focused on the K code, did not do an extensive look at the OCaml code).

Also, general design point, is it actually necessary to make it specific to 8-bit width? At least for the `BYTES-SYMBOLIC` side, I could see adding something like:

```
syntax Int ::= "#stackItemWidth" [function]
```

Then, when the user imports and uses `BYTES`, they add a rule:

```
rule #stackItemWidth => 8
```

if they want bytes.

Or perhaps even renaming the functions `Bytes2Int` to `IntStack2Int`, and adding an extra parameter which is the width to interpret them in. This would make it possible (on the symbolic side) to have both byte-stacks and EVM-word stacks.

How possible is this on the OCaml side? (in terms of still having a packed and efficient representation in memory)

k-distribution/include/builtin/domains.k
dwightguth: .Bytes is a function, nilBytes is the constructor that the function evaluates to. This is done so that we don't have to change the attributes of the production based on the backend, which we can't do currently.

k-distribution/include/builtin/domains.k
dwightguth: yes, because we use log2Int in this module and the BYTES-SYMBOLIC import will not be present in the ocaml backend since it is a java-backend-only module.

k-distribution/include/builtin/domains.k
dwightguth: because there is not an efficient hook written for this function, for simplicity's sake.

dwightguth:
I'm honestly not particularly concerned about the EVM word stack. It has a max length of 1024 and as such is not really subject to the asymptotic problems that I ran into with large memory buffers that led me to make this change. I could probably make something with a parametric value width, but it doesn't seem necessary to me and would add significant complexity to the module in the prelude. If you really want, it should be possible to implement a 32-byte-width byte string on top of the 1-byte-width byte string with the primitives that I provided.

traiansf (approved):

k-distribution/include/builtin/domains.k
ehildenb: @dwightguth I think I would prefer if these went in the `BYTES-SYNTAX` module instead of here. If someone imports `BYTES-SYNTAX`, the other hooked functions there will work on the ocaml backend whereas these ones will fail.

k-distribution/include/builtin/domains.k
```diff
@@ -590,6 +590,98 @@ module STRING-BUFFER
...
+  syntax Bytes [hook(BYTES.Bytes)]
+
+  syntax Bytes ::= ".Bytes" [function, hook(BYTES.empty), impure]
+  syntax Endianness ::= "LE" [klabel(littleEndianBytes)]
+                      | "BE" [klabel(bigEndianBytes)]
+  syntax Signedness ::= "Signed" [klabel(signedBytes)]
+                      | "Unsigned" [klabel(unsignedBytes)]
+  syntax Int ::= Bytes2Int(Bytes, Endianness, Signedness) [function, hook(BYTES.bytes2int)]
+  syntax Bytes ::= Int2Bytes(Int, Int, Endianness) [function, hook(BYTES.int2bytes)]
+                 | Int2Bytes(Int, Endianness, Signedness) [function, klabel(Int2BytesNoLen)]
```
ehildenb, line 603: This function (KLabel `Int2BytesNoLen`) will fail if someone imports the `BYTES-SYNTAX` module, even though the rest will succeed.

ehildenb:
@dwightguth changes look good, except that I'm concerned about the fact that people will get an inconsistent view if they import `BYTES-SYNTAX` vs `BYTES` on the ocaml backend. I think I would even go for renaming `BYTES-SYNTAX` to `BYTES`, and removing `BYTES` altogether (move it's definitions into `BYTES`). That way users importing `BYTES` will get all of the functionality for the OCaml backend, and will have to explicitely import `BYTES-SYMBOLIC` for the symbolic reasoning.

Also, I thought that the hooks were setup so that if the hook does not work, then it defaults to the K rules. So wouldn't it also be safe to just put everything in a single module `BYTES`, and we'll still get the efficient OCaml implementation for concrete code, but will fall back to the K rules otherwise?

My other question is, are all these tests run with both backends?

dwightguth:
@ehildenb I can't do what you suggest because both the rules in BYTES-SYMBOLIC and the rules in BYTES rely on the syntax productions in BYTES-SYNTAX. If I tried to merge BYTES with BYTES-SYNTAX I would end up with a circular import BYTES > BYTES-SYMBOLIC > BYTES, which is invalid.

I'm going to go ahead and merge this, but in answer to your second question, yes, these tests are run with both backends.

PR #104, opened 2018-06-22
Contradiction ellim

denis-bogdanas:
Kprove improvement: eliminating terms that contain some constraint "T" and its opposite "T ==K false" at the same time. Z3 is unable to handle such cases if T contains functional terms from the semantics that cannot be fully evaluated, such as "-Word" from KEVM.

Did not apply the improvement in all possible contexts, potentially this change can be expanded to more situations.


daejunpark:
jenkins: test this please

daejunpark:
@denis-bogdanas Please fix the compilation error (see the Jenkins log), and rebase it on the master, so that we can merge this.

denis-bogdanas:
jenkins: test this please

ehildenb:
@denis-bogdanas can I suggest a different way of integrating this with the source of K?

In `FastRuleMatcher.java`, we have checks for `continousSimplification`. Perhaps in those clauses is the time to apply this function? So that if we turn continuousSimplification on, this check will happen, if not then it won't. The reason I suggest this is so that we can keep all these optimizations in a single place. If you wanted to extract a separate function called `fullSimplify` which is what's called when `continuousSimplification` is checked, then have `fullSimplify` call both the regular `simplify` and your procedure, perhaps that would even help to keep all these optimizations in one places as well (so we can test them until we're confident in them).

The ideal way to implement this is as the K rewrite rule added to the prelude:

```
rule B:Bool ==K false #And B:Bool ==K true => false
```

Unfortunately, I don't think this will work because our assoc-comm matching on `#And` probably won't work well enough, so I think integrating it as a direct check

daejunpark:
@denis-bogdanas, @ehildenb's comment made me think twice with this PR. What will happen if you simply drop-in replace the `isFalse()` with the `isFalseExtended()`? Will it only affect the performance, or will it lead to any other unexpected behavior?

ehildenb:
@daejunpark I think it should be fine (shouldn't lead to bad behaviour) because it will take entire constraints to false, which will only be used to prune bad paths. I think if it replaced sub-components of constraints with false, it could be bad because we could lose (even redundant) information that the SMT solvers can be more efficient with.

denis-bogdanas:
@daejunpark I'm not sure what will happen. "Normally", e.g. according to reasonable expectations about the code that I only partially understand, it should only affect the performance. That is why I replaced `isFalse()` with `isFalseExtended()` conservatively. Only in places where it is followed by `checkUnsat()` - e.g. a more thorough check, that we expect should have detected this contradiction if the term would be fully translated to z3.

denis-bogdanas:
@ehildenb What you suggest might be valid but I don't understand the code well enough to make this change. The way I did it it's a safe optimization because it's always followed by the more thorough `checkUnsat()` as I explained above.

denis-bogdanas:
And besides it's not clear why last Jenkins build failed. Looks unrelated to this PR but to ocaml.

daejunpark:
Jenkins: test this please

daejunpark:
@andreistefanescu Could you review this change? Do you have any suggestion?

denis-bogdanas:
It failed at a different point this time:
`[exec] include/ktest-group.mak:11: recipe for target 'tutorial' failed`

denis-bogdanas:
Jenkins: test this please

denis-bogdanas:
@andreistefanescu Hi, do you have time to review this or we should manage ourselves?

denis-bogdanas:
@ehildenb Can you approve please? Looks like @andreistefanescu cannot be contacted.

daejunpark:
@ehildenb Considering the recent progress of the haskell backend, I guess we will be able to replace the java backend in the near future, which means that we shouldn't spend too much energy to maintain the java backend. So, I'd like to suggest to simply merge this PR, so that we can reduce our overhead of maintaining the long-standing branch in the smart contract verification effort.

daejunpark (approved):

daejunpark:
@dwightguth could you please merge this?

PR #109, opened 2018-07-02
Debugg, KProve, KProveoptions, KPrintOptions; path and omission debug…

MrChico:
…ging options


ehildenb:
@MrChico and @mhhf, I've included your patch in the `state-dumping` branch, which also now has some other patches from @denis-bogdanas for adding an option to use the faster `toString` method for printing terms instead of `unparseTerm` (using a similar `--output-tostring`), which is *much* faster.

The new commit id is: 1d798c863
My suggested changes are: ec3e86cab

If you test out the new branch, and approve of the suggested changes, I'll squash the suggested changes into your commit and close this PR.

mhhf:
Tested this changes and they seem to work with the current version of klab.
I created another PR #112 with changes we also need to support for.

ehildenb:
Alright, closing this and moving conversation over to there. I squashed my suggestions into the last commit.

PR #110, opened 2018-07-02
Improved unparsing

dwightguth:
Basically this fixes several soundness issues related to unparsing and also provides an unparsing mode suitable for outputting code that is intended to be parsed by the program parser. I have tested it on a iele example and the resulting contents of the k cell will parse to an identical AST as the original program.

Do not review yet.


ehildenb:
@dwightguth I'm very excited for this, as it seems to fix some issues with unparsing lists that we are having in the KAT repo.

dwightguth:
@ehildenb I was planning on making a follow-up pull request to the kat repository that uses the format attribute and this pull request to make it directly emit K syntax. But it looks like the tests are still failing, so I'll have to investigate tomorrow.

dwightguth:
Jenkins: test this please

ehildenb:
@dwightguth sorry for not waiting for you on this, but I need the fixed unparsing in KAT to make progress. Initial tests seem to indicate that it works correctly, but running more tests now. Branch `git-ci-setup` on the KAT repo (a better name for the branch would be `git-ci-removal`).

ehildenb:
@dwightguth further testing indicates that it's better, but still perhaps not perfect? I'm not sure honestly.

Basically, in the FUN language, there are three sorts:

```
syntax Names ::= List{Name,","}
syntax Vals ::= List{Val,","}
syntax Exps ::= List{Exp,","}
```

with `Vals` and `Exps` in a subsort relationship.

Before, it would produce `.Names` where it should definitely be producing `.Vals` or `.Exps`, now it does not do that anymore. But now it seems that there are places where it produces `.Exps` where `.Vals` is also correct, so it seems that `.Vals` would be the more optimal thing to produce (since it's lower in the subsort heirarchy).

Either way, this should fix the non-determinism issues I was having with how it picked which one (even if the subsort issue isn't figured out), I'll let you know if it passes CI.

dwightguth:
@ehildenb I believe I have fixed the issues in the example that I showed you, can you take another look?

ehildenb:
@dwightguth this doesn't quite fix it, now it seems to just pick the lowest sort in the hierarchy of list sorts (which is less sound than picking the highest sort, I believe).

An example program which demonstrates it is `pattern.fun`.

Things that show up in the resulting configuration:

```
Pair ( h1 , h2 , .Names )
```

This correctly infers `h1` and `h2` as the sort `Name`.

```
Pair ( 3 , b , .Names )
```

The lowest possible sort here should be `Val`, I believe.

```
Pair ( h2 + length t + b , a , .Names )
```

This should show up as sort `Exp`.

If finding the exact sort to use proves difficult (last time I tried to get the Java code to infer sorts correctly it didn't seem to have the necessary infrastructure), I think going with the highest sort in the hierarchy of Lists which share sorts should be safe. It will be parsed correctly and rules over the list will still only work if the elements match the correct sort. If you can make it do proper sort inference though, that would be cool.

The sort hierarchy here is:

```
Name Val < Exp
Name < Type
```

So at least it is now correctly inferring that the occurrences are not in the `Name < Type` side of the hierarchy.

dwightguth:
I mean, right now it picks the least production such that the term parses correctly. That's why you are getting `.Names`, because `.Names` is a Names and every Names is an Exps, so I can tell you that this behavior is definitely sound, it's just unintuitive. I can probably change it to select the greatest production such that the term parses, which might give you a result that you find more intuitive, but it would no longer be deterministic. The Java backend requires that terms have a least sort; it doesn't require that they have a greatest sort, which means that in cases where multiple greatest sorts exist, one will be picked at random. Moreover, choosing the least production is the logical choice because algorithmically, this is the production that is chosen when parsing, and the unparsing algorithm works by using the same disambiguation algorithm as the parser on an ambiguous term generated from the abstract syntax.

So while I recognize that it's unintuitive and it confused me at first too, it is nonetheless the most sound possible choice the algorithm can make in each of these cases.

ehildenb:
Hmmmm, I guess it makes sense then in terms of identifying all the empty lists as the same empty list when there is subsort overloading. Now that I think about it, Maude does a similar thing where it assumes that if you have subsort over-loaded collections, the unit element of the collections must be the same element. Here, that unique element becomes `.Names`.

ehildenb:
@dwightguth unfortunately I still think it's not working quite correctly.

In program `constructor-tree-mirror.fun`, I still see some `.Names` show up where it should be `.Vals`.

For example:

```
3 |-> Tree ( Leaf ( 1 , .Names ) , Leaf ( 2 , .Vals ) , .Vals )
```

Throughout the rest of the configuration, it correctly puts `.Vals` when all the elements in the list are `Val` and *not* `Name`, but then a few little spots show up like this. This was one example, but that same configuration demonstrated two other examples, and programs `pattern.fun` and `constructor-bst-sorting.fun` also demonstrates some issues.

My biggest concern is that this is due to nondeterminism, but I'm not sure.

Commit https://github.com/kframework/kat/commit/f6a7e398fa20f508a4394ac70274209e576b40bd are the changes I had to make to the expected output which I do not think should be made. Commit https://github.com/kframework/kat/commit/11179373c26f8406d1323da961f3f20ff76db9cc are the changes that I made to the expected output with the new unparsing which I believe are good changes.

dwightguth:
@ehildenb I made the change to always select the least production in a top-down manner rather than trying to determine the right production bottom-up, and it seems to have removed the remaining cases where a term was a Vals but it ended in .Names instead of .Vals. If you could take another look, I'd appreciate it.

ehildenb (approved):
Looks like it fixes the KAT repo unparsing issues! Thanks @dwightguth!

denis-bogdanas:
I noticed that unparsing is much slower now. Pretty printing K cell and the constraint in a kevm proof took <0.5 s before, and ~50s now. Can we have both old and new versions of unparsing?

dwightguth:
Do you have an example I can try out? I can look into performance and see if it can be improved, but if I don't see any obvious way to make it faster I can do what you suggest.

denis-bogdanas:
I opened a separate issue: https://github.com/kframework/k/issues/121

PR #111, opened 2018-07-04
Json output and kprint

ehildenb:
This PR achieves 2 things:

(1) Adds class `KPrint` which acts as a wrapper around the various serializers (`pretty|json|binary|none|kast`), holding the serialization options in hand. This way any other part of K wishing to output information can make sure the output is affected by the options passed to `KPrint` by using the class. Various abstraction passes are added for doing lossy serializations.

(2) Adds the `ToJson` and `JsonParser` classes, which are the serializer/deserializer for Json.

@dwightguth how do you think we should test this? Ideally, we would have one "testset" which we run through all the serializer/deserializers/lossy serializations I think.

### DIFF
--- a/kernel/pom.xml
+++ b/kernel/pom.xml
@@ -55,6 +55,11 @@
       <version>3.0.0</version>
     </dependency>
     <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
+      <version>1.0.4</version>
+    </dependency>
+    <dependency>
       <groupId>net.sf.jung</groupId>
       <artifactId>jung-api</artifactId>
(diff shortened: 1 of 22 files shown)


ehildenb:
By the way, branch `state-dumping` is using this class currently for some people using the prover, so there is some "real-world" testing of the code already, but we should add some static tests to the repo.

denis-bogdanas:
I'm waiting for this merge btw. I will still use my custom logs, but I want to use the KPrint options framework and better integrate them with master.

dwightguth:
I believe this can be closed now as a duplicate?

ehildenb:
Closed as subsumed in #117 and #136.

PR #117, opened 2018-07-16
[REFACTORING] Kprint

ehildenb:
Isolate printing and unparsing infrastructure to new module KPrint.

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/ColorSetting.java
+++ b/kernel/src/main/java/org/kframework/unparser/ColorSetting.java
@@ -1,5 +1,5 @@
 // Copyright (c) 2014-2018 K Team. All Rights Reserved.
-package org.kframework.krun;
+package org.kframework.unparser;
 
 /**
  * @author Denis Bogdanas
--- a/kernel/src/main/java/org/kframework/unparser/Indenter.java
+++ b/kernel/src/main/java/org/kframework/unparser/Indenter.java
@@ -1,3 +1,5 @@
+// Copyright (c) 2016-2018 K Team. All Rights Reserved.
+
 package org.kframework.unparser;
 
 public class Indenter implements Appendable {
--- a/kernel/src/main/java/org/kframework/unparser/KPrint.java
+++ b/kernel/src/main/java/org/kframework/unparser/KPrint.java
@@ -53,25 +53,24 @@ public class KPrint {
     }
 
     public static void prettyPrint(Definition def, Module module, FileUtil files, KompileOptions kompile, OutputModes output, Consumer<byte[]> print, K result, ColorSetting colorize) {
+        print.accept(prettyPrint(def, module, files, kompile, output, result, colorize));
+    }
+
+    public static byte[] prettyPrint(Definition def, Module module, FileUtil files, KompileOptions kompile, OutputModes output, K result, ColorSetting colorize) {
         switch (output) {
         case KAST:
-            print.accept((ToKast.apply(result) + "\n").getBytes());
-            break;
+            return (ToKast.apply(result) + "\n").getBytes();
         case NONE:
-            print.accept("".getBytes());
-            break;
+            return "".getBytes();
         case PRETTY: {
             Module unparsingModule = RuleGrammarGenerator.getCombinedGrammar(module, false).getExtensionModule();
-            print.accept((unparseTerm(result, unparsingModule, colorize, files, kompile) + "\n").getBytes());
-            break;
+            return (unparseTerm(result, unparsingModule, colorize, files, kompile) + "\n").getBytes();
         } case PROGRAM: {
             RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
             Module unparsingModule = RuleGrammarGenerator.getCombinedGrammar(gen.getProgramsGrammar(module), false).getParsingModule();
-            print.accept((unparseTerm(result, unparsingModule, colorize, files, kompile) + "\n").getBytes());
-            break;
+            return (unparseTerm(result, unparsingModule, colorize, files, kompile) + "\n").getBytes();
         } case BINARY:
-            print.accept(ToBinary.apply(result));
-            break;
+            return ToBinary.apply(result);
         default:
             throw KEMException.criticalError("Unsupported output mode: " + output);
         }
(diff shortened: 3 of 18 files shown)


ehildenb:
Jenkins: test this please

kernel/src/main/java/org/kframework/unparser/ColorSetting.java
```diff
@@ -0,0 +1,38 @@
...
+    private ColorSetting color;
+    static {
+        OFF.color      = OFF;
+        ON.color       = ON;
+        EXTENDED.color = EXTENDED;
+    }
+
+    public ColorSetting color(boolean ttyStdout, String outputFile, Map<String, String> env) {
+        boolean colorize = outputFile == null && ttyStdout;
+        ColorSetting colors = color;
```
dwightguth: can't you just say this instead of color? and then delete the color field?

kernel/src/main/java/org/kframework/unparser/Indenter.java
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2018 K Team. All Rights Reserved.
```
dwightguth: this file actually dates back to 2012.

kernel/src/main/java/org/kframework/unparser/OutputModes.java
```diff
@@ -7,5 +7,18 @@
  *
  */
 public enum OutputModes {
-    PRETTY, PROGRAM, KAST, BINARY, NONE
+    PRETTY, PROGRAM, KAST, BINARY, NONE;
+
+    private String extension;
```
dwightguth: this seems to be unused?

dwightguth (changes requested):

kernel/src/main/java/org/kframework/unparser/Indenter.java
ehildenb: Will fix.

kernel/src/main/java/org/kframework/unparser/OutputModes.java
ehildenb: This change facilitates a future PR, I'll move the commit to that PR instead.

kernel/src/main/java/org/kframework/unparser/ColorSetting.java
ehildenb: I'm not sure if I know how to address this, but I will remove the intermediate variable `colors`, and just update the class-level `color` directly.

ehildenb:
@dwightguth I've made the changes from above. I'll work on the failing tests tomorrow.

kernel/src/main/java/org/kframework/unparser/ColorSetting.java
dwightguth: I mean it looks to me like this code is setting an instance variable on each instance of the enumeration to itself. So `color == this` should always be true, which is why I don't think this field is necessary.

ehildenb: Alright, I've re-done using `this`. Is this the correct way?

kernel/src/main/java/org/kframework/unparser/ColorSetting.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @author Denis Bogdanas
+ *         Date: 10/9/13
+ */
+public enum ColorSetting {
+    OFF, ON, EXTENDED;
+
+    public ColorSetting color(boolean ttyStdout, String outputFile, Map<String, String> env) {
+        boolean colorize = outputFile == null && ttyStdout;
+        if (this == null) {
```
dwightguth: This is even more confusing. This can never be null, because you can't call methods on a null pointer. It makes more sense now that you don't have that confusing instance field that is always self referential, but you can't handle the null case like this. Why was this turned into an instance method anyway?

dwightguth (dismissed):

kernel/src/main/java/org/kframework/unparser/ColorSetting.java
ehildenb: Hmmm, this came initially from `PrintOptions` (which used to be `PrettyPrintOptions`). I guess I'll put it back there, and it will make more sense. The idea was to set the "default setting" `ColorSetting` logic should live with the `ColorSetting` definition. But I guess this is more about defaults for the `PrintOption` than about the `ColorSetting` itself.

ehildenb:
@dwightguth I've reverted the change which made `color` an instance method of `ColorSetting`, putting it back into `PrintOptions`.

ehildenb:
@dwightguth I've changed over to using `final` variables instead of `static` variables for the KPrint class. The idea is that we want one class which controls serialization (similar to how Pandoc has "writers"), so that eventually we can just use the Java backend as a converter between serialization formats (just invoke parser + printer).

For now, this enables also printing to JSON (or KAST or BINARY, etc...) in the next PR in a smooth way. I think we can also write a more generic testing setup for this class (and test all the serializers/deserializers with it).

kernel/src/main/java/org/kframework/unparser/KPrint.java
```diff
@@ -0,0 +1,149 @@
...
+    private final FileUtil          files;
+    private final TTYInfo           tty;
+
+    public final PrintOptions options;
+
+    public KPrint() {
+        this(new KExceptionManager(new GlobalOptions()), FileUtil.testFileUtil(), new TTYInfo(false, false, false), new PrintOptions());
+    }
+
+    public KPrint(KExceptionManager kem, FileUtil files, TTYInfo tty, PrintOptions options) {
```
dwightguth: This constructor needs the `@Inject` attribute

dwightguth (changes requested):

ehildenb:
@dwightguth I've added the `@Inject` annotation.

dwightguth:
Looks like you forgot an import.

ehildenb:
@dwightguth fixed.

dwightguth (approved):

dwightguth:
This is now ready to be merged.

ehildenb:
@dwightguth I guess you're the only one authorized to merge this.

PR #119, opened 2018-07-18
Windows support for kompile and kprove.

denis-bogdanas:


dwightguth (dismissed):
this looks good but you should also remove the reference to libfl-dev in README.md and k-distribution/INSTALL.md.

denis-bogdanas:
I suppose that's because I removed `-lfl` option? Ok will do that.

dwightguth (approved):

PR #122, opened 2018-07-20
Logs: Short summary box after a proof is finished,

denis-bogdanas:
2 more timing mess ages under option --verbose.


daejunpark:
Jenkins: test this please

denis-bogdanas:
Jenkins: test this please

denis-bogdanas:
@dwightguth Any idea why this fails? kprove now prints this box for each spec, followed by the same output as before. E.g. followed by `#True` if all specs were proved.

```
==================================
SPEC PROVED: /mnt/d/verified-smart-contracts-fork/specs/casper/proc_reward-spec.k Location(11,5,183,39)
Execution paths: 8
Longest path: 10183 steps
Time so far: 485.657 s
==================================
```

Could this be the reason?

daejunpark:
@denis-bogdanas I guess the additional output caused the failure. Is the information printed only under the `--verbose` option?

denis-bogdanas:
No. I added some bits under --verbose and some other by default. Will change that.

java-backend/src/main/java/org/kframework/backend/java/util/Profiler2.java
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2015-2018 K Team. All Rights Reserved.
+package org.kframework.backend.java.util;
+
+import org.kframework.main.Main;
+
+/**
+ * @author Denis Bogdanas
+ * Created on 23-Jul-18.
+ */
+public class Profiler2 {
```
dwightguth: Shouldn't this be set up so that each invocation of a K program only has one Profiler2 instance? If so, you are going to need the @RequestScoped attribute.

kernel/src/main/java/org/kframework/main/Main.java
```diff
@@ -38,6 +38,8 @@
 
 public class Main {
 
+    public static final long startTime = System.currentTimeMillis();
```
dwightguth: this is not going to play nice with the k server. Can you please put this in a request-scoped value that is requested by the Main class that is instantiated (see the constructor on line 90) and then inject it into the Profiler2?

dwightguth (changes requested):
looks pretty good but I have a couple requests relating to lifetime and dependency injection.

denis-bogdanas:
@dwightguth Thanks for tips! Will do. What is the scope of Profiler2 the way it is now?

dwightguth:
By default Guice creates a new instance of each object for every injection point.

dwightguth (approved):

dwightguth:
I assume you've tested the latest version and it still properly outputs what you want it to output? If so and the tests pass you can merge.

denis-bogdanas:
@dwightguth Yes I tested. I'm not authorized to merge.

denis-bogdanas:
I added this PR for you actually. For verification team we have the same changes on `gnosis` branch.

denis-bogdanas:
@dwightguth Can you merge please?

PR #125, opened 2018-07-24
Update K to generate kore from IELE

dwightguth:
Updates K to support generating KORE from the IELE semantics.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
The tests are not going to pass because of bugs in the C semantics that @chathhorn will fix after this is merged, but this is ready for review.

traiansf (commented):
Would it be possible to add some tests?  They would also serve to show how the output looks like.

dwightguth:
Quite difficult actually. I have a test that tests imp that was part of a previous pr, but it doesn't actually compare the result because as tests go that would be incredibly fragile. Doing a fuzzier comparison would require a diff engine that understood kore, which would be extremely complicated and time consuming to write. We are probably better off trying to do functional testing by feeding the result of the generated kore into the haskell backend when the time comes.

dwightguth:
Jenkins: test this please

traiansf (approved):
Looks ok to me.
Two generic observations, which should not prevent merging this, though:
- It might be useful to have a way to parse programs and output them in the Kore format; maybe using the krun command?
- Not sure though whether things like functionality should be stated as an axiom, or rather to generate a proof obligation... maybe that's ok for constructors, but it's not so clear we want it for function symbols as it can quickly lead to proving false for incompletely defined functions, e.g.

traiansf:
BTW, sorry it took so long to get it reviewed.

PR #134, opened 2018-08-13
remove functional for all functions and manually add it back for tota…

### LINKED ISSUE #132: [K2Kore] Remove the `functional` attribute from functions
Currently functions are tagged with the `functional` attribute when converting to KORE.  I don't thing this is the right thing to do for functions, as they might be partial, while `functional` guarantees totality.  While it might be desirable for us to prove that certain functions are total and use that, it seems `functional`ity for unctions if more of a proof obligation than a guarantee.

dwightguth:
…l functions.

Essentially I made it so that functions are not functional by default, but if you tag them with the functional attribute, a functional axiom is created. Then I added the functional attribute to all the functions in the prelude I thought were total.

@traiansf if you could check my work on domains.k and make sure I didn't add the functional attribute to any partial functions that would be great.

Fixes #132


traiansf (approved):
Looks good to me

PR #140, opened 2018-08-23
replace injection into k with list of injection into kitem

dwightguth:
Please let me know if this fixes your issue.


traiansf (approved):
seems to work fine.

daejunpark:
Um, @dwightguth or @traiansf, could you explain what was the issue?

traiansf:
The issue is the following:  According to the syntax of K, a syntactic element can either be injected directly into K, or into a KItem;  injecting it into K seems to be problematic because of the interactions with the K computation sequencing operation, so Dwight encoded syntactic elements in Kore similarly to how they are encoded in Ocaml:  each syntactic element is injected into KItem, which is then pushed in front of a .K to get something of sort K.

daejunpark:
I see. Thanks!

PR #142, opened 2018-08-28
add attributes for data structures to sort declaration

dwightguth:


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -106,6 +106,11 @@ public String convert() {
             Att att = module.sortAttributesFor().get(sort).getOrElse(() -> KORE.Att());
             if (att.contains(Attribute.HOOK_KEY) && !dummyHookedSorts.contains(att.get(Attribute.HOOK_KEY))) {
                 sb.append("hooked-");
+            } else if (att.contains(Attribute.HOOK_KEY)) {
```
traiansf, line 109: I find the logic here a bit hard to follow.  Do you mean that it contains the `HOOK_KEY`, but it is in the `dummyHookedSorts`?  If so, wouldn't it be better to merge the two `if`s and have one only for `!dummyHookedSorts`?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -922,7 +927,12 @@ private void convert(Map<String, Boolean> attributes, Att att) {
             Object val = attribute._2;
             String strVal = val.toString();
             sb.append(conn);
-            if (attributes.get(name) != null && attributes.get(name)) {
+            if (cls.equals(K.class)) {
```
traiansf, line 930: Do you mean here to say that an attribute of K class will be printed even if it's not in the `attributes` map?  
I would like a little more context on what this PR actually does.

traiansf (approved):

PR #144, opened 2018-09-04
Don't require interactivity in the build

maya-rv:
Answer "no" by default to the question:
Do you want OPAM to modify ~/.profile and ~/.ocamlinit?

### DIFF
--- a/k-distribution/src/main/scripts/bin/k-configure-opam-dev
+++ b/k-distribution/src/main/scripts/bin/k-configure-opam-dev
@@ -1,5 +1,5 @@
 #!/bin/sh
-opam init
+yes n | opam init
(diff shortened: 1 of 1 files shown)


dwightguth:
I'm not entirely convinced this is the way to go... If you don't update .profile then you will have to run `` eval `opam config env` `` in every new shell that you want to invoke k in. On the other hand, if we answer yes to this prompt by default then people may not like having their .profile changed automatically without being consulted. Can you elaborate on the reasoning behind why you don't want this to have to be chosen manually?

maya-rv:
> you will have to run eval `opam config env` in every new shell that you want to invoke k in.

that's probably fixable, but since I'm not going to fix it, OK.
defaulting to yes is definitely the worst option.

PR #146, opened 2018-09-11
add top cell initializer to definition attributes

dwightguth:


traiansf (approved):
LGTM;  however, note that if Grigore's plan to check attributes for well-formedness is effected, we will have issues here.  ONe way to address this would be to add a variable as an argument;  however, personally I like it better the way it is now.

PR #147, opened 2018-09-13
Extracted kore functionality to a proper backend.

traiansf:
Also added `krun` functionality.

### DIFF
--- a/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
+++ b/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
@@ -1,0 +1,190 @@
+// Copyright (c) 2014-2018 K Team. All Rights Reserved.
+package org.kframework.backend.haskell;
+
+import com.google.inject.Inject;
+import org.kframework.RewriterResult;
+import org.kframework.backend.kore.ModuleToKORE;
+import org.kframework.compile.AddSortInjections;
+import org.kframework.definition.Module;
+import org.kframework.definition.Rule;
+import org.kframework.kompile.CompiledDefinition;
+import org.kframework.kore.K;
+import org.kframework.krun.KRunOptions;
+import org.kframework.krun.RunProcess;
+import org.kframework.main.Main;
+import org.kframework.rewriter.Rewriter;
+import org.kframework.rewriter.SearchType;
+import org.kframework.unparser.OutputModes;
+import org.kframework.utils.errorsystem.KEMException;
+import org.kframework.utils.file.FileUtil;
+import org.kframework.utils.inject.RequestScoped;
+import scala.Tuple2;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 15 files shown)


dwightguth:
I actually don't want the kore code to be in a separate backend... It should stay part of the kernel since it's now becoming part of the core of the new K.

That said, you should introduce a haskell backend which assumes that kore is installed on the path and invokes it from krun.

traiansf:
@dwightguth please take another look.  I attempted to refactor it as you suggested.

traiansf:
@dwightguth I went on and added more commits to this PR. it now also converts heating/cooling rules to equations and changes the encoding of owise rules as per the design decision.  The commits are pretty self-contained.

traiansf:
@dwightguth I had a chat with Grigore on Thursday about how to encode subsorts and we got to the conclusion that the best way to do that is to add definedness (totality) axioms for `inj`, which is what I attempted to do in the commit above.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -0,0 +1,190 @@
...
+        this.files = files;
+        this.def = def;
+        this.haskellKRunOptions = haskellKRunOptions;
+        this.options = options;
+    }
+
+    @Override
+    public Rewriter apply(Module module) {
+        if (!module.equals(def.executionModule())) {
+            throw KEMException.criticalError("Invalid module specified for rewriting. Ocaml backend only supports rewriting over" +
```
dwightguth: this should probably say Haskell, not Ocaml.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -114,7 +114,7 @@ public int run() {
                 parsed = new ExpandMacros(compiledMod, files, def.kompileOptions, false).expand(parsed);
             }
             if (options.kore) {
-              ModuleToKORE converter = new ModuleToKORE(compiledMod, files, def.topCellInitializer);
+              ModuleToKORE converter = new ModuleToKORE(compiledMod, files, null);
```
dwightguth: why was this done?

dwightguth (approved):
looks good but I have one minor suggestion and one comment you should probably address before you merge.

dwightguth (changes requested):
actually I take it back, something about your changes broke the llvm backend.

dwightguth (approved):

dwightguth:
haha never mind, it was trivial to fix.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
traiansf: Sorry about that!  I think it happened while doing conflict resolution during the merge; for some reason I didn't noticed I can use `def.topCellInitializer` there.  I'm reverting the change.

traiansf:
@dwightguth I fixed the issues you pointed out, but I think it's you who need to do the merge.

PR #149, opened 2018-09-14
Changes to K to KORE to support the llvm backend better

dwightguth:
Some of these changes might be a bit controversial, but they are necessary in order to execute imp programs efficiently. @traiansf if you could let me know whether you have problems with them, that would be great.

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
@@ -42,6 +42,7 @@ public class KoreBackend implements Backend {
     private final KompileOptions kompileOptions;
     private final FileUtil files;
     private final KExceptionManager kem;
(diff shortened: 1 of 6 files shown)


traiansf:
@dwightguth : Actually, I do have problems with them :-) .  

First, for the haskell backend we will want heating/cooling equations, as having them as rules is neither correct, nor complete.  In the Haskell backend the plan is to encode checking of resultness as part of the strategy, as it should be.  If you prefer having them as heating/cooling rules, maybe we should branch and have a common core for what transformations kore requires, and then specializations for the various backends using kore.

Also, the general plan about isSort rules that is that they should go away completely now that we have injections, and whenever special cases are needed (e.g., lambda values) they should be manually defined.

traiansf (approved):

traiansf (approved):
LGTM

PR #154, opened 2018-09-21
Addded Scala kore parser and Kore2K translator and integrated them into K

traiansf:


traiansf:
@dwightguth only the last commit is relevant; the rest are just history imported from the kore repository.

dwightguth (changes requested):
Please see my comments on the last commit. I have some minor suggestions and also a few recommendations on how to translate kore back into k.

traiansf:
@dwightguth I think I've addresses all your recommendations.  Sorry for not understanding what you meant earlier;  I had just woken up.

dwightguth (approved):

dwightguth:
this is good enough, so I approved. Note however that what I actually suggested was to make TextToKore's interface be static. (ie, in scala, to say `object TextToKore { def apply(...) = ... }`)

PR #157, opened 2018-09-28
Allow making modifications/omissions to the output via command-line flags

ehildenb:
This adds (command-line) driven abstraction passes to `KPrint`. This is to be used with the serializers when communicating terms to other tools, not really by the pretty printer.

-   `--output-omit` will completely omit occurances of that KLabel from the output.
-   `--output-tokenize` will "tokenize" everything below specified KLabels (so it becomes a single AST node).
-   `--output-flatten` will flatten specified associated KLabels from cons-lists to assoc-ish lists.
-   `--output-tostring` calls `toString` on the given nodes in the AST and stores them as a `KToken` (faster than `--output-tokenize`).

Also `KPrint` has been refactored to take `kompileOptions` at construction time, so that there is one less argument to `prettyPrint`.

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/KPrint.java
+++ b/kernel/src/main/java/org/kframework/unparser/KPrint.java
@@ -186,37 +186,28 @@ public class KPrint {
         return new TransformK() {
             @Override
             public K apply(KApply orig) {
-                K omitted   = omitTerm(mod, orig);
-                K tokenized = tokenizeTerm(mod, omitted);
-                return tokenized;
+                String name = orig.klabel().name();
+                return options.omittedKLabels.contains(name)   ? omitTerm(mod, orig)
+                     : options.tokenizedKLabels.contains(name) ? tokenizeTerm(mod, orig)
+                     : orig ;
             }
         }.apply(term);
     }
 
-    private K omitTerm(Module module, K k) {
-        if (! (k instanceof KApply)) return k;
-
-        KApply kapp = (KApply) k;
-        if (! options.omittedKLabels.contains(kapp.klabel().name())) return k;
-
+    private K omitTerm(Module mod, KApply kapp) {
         Sort finalSort = Sorts.K();
-        Option<Sort> termSort = module.sortFor().get(kapp.klabel());
+        Option<Sort> termSort = mod.sortFor().get(kapp.klabel());
         if (! termSort.isEmpty()) {
             finalSort = termSort.get();
         }
         return KToken(kapp.klabel().name(), finalSort);
     }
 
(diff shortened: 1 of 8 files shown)


dwightguth:
I'm not sure I like the idea of --output-tostring, the tostring methods on k terms are pretty janky and should not be exposed to users. If it's just a matter of performance, have an --output-kast which just doesn't unparse the term

ehildenb:
@dwightguth see the newest commit, it makes the switch you suggest. I had to fallback to to `toString` for `UnsupportedEncodingException`, which basically should never happen I think? So more just to appease the exceptions type-system.

kernel/src/main/java/org/kframework/unparser/KPrint.java
```diff
@@ -42,19 +49,25 @@
...
 
     public final PrintOptions options;
 
     public KPrint() {
-        this(new KExceptionManager(new GlobalOptions()), FileUtil.testFileUtil(), new TTYInfo(false, false, false), new PrintOptions());
+        this(new KExceptionManager(new GlobalOptions()), FileUtil.testFileUtil(), new TTYInfo(false, false, false), new PrintOptions(), new KompileOptions());
+    }
+
+    public KPrint(KompileOptions kompileOptions) {
+        this(new KExceptionManager(new GlobalOptions()), FileUtil.testFileUtil(), new TTYInfo(false, false, false), new PrintOptions(), kompileOptions);
```
dwightguth: I should think this should be kompileOptions.global

kernel/src/main/java/org/kframework/unparser/KPrint.java
```diff
@@ -158,12 +176,72 @@ public K apply(KVariable k) {
...
+        } else {
+            items = kapp.klist().items();
+        }
+        return KApply(kapp.klabel(), KList(items), kapp.att());
+    }
+
+    public static K toKASTTerm(Module mod, KApply kapp) {
+        String kastTerm;
+        try {
+            kastTerm = new String(KPrint.serialize(kapp, OutputModes.KAST), "UTF-8");
```
dwightguth: You should be able to skip the exception handling here by just calling ToKast directly.

dwightguth: Not to mention the fact that this code currently is going to create mojibake as it serializes the string to the machine's default charset and then tries to parse it with utf-8.

dwightguth (changes requested):
looks good but there are some minor changes that need to be made.

dwightguth (approved):

PR #159, opened 2018-10-02
LLVM backend

dwightguth:


traiansf (approved):
I think you kind of cheated here, not using executeAndMatch to retrieve the exit code, but I'm going to approve it anyway; it might make the job easier for the Haskell backend, too, although there it makes sense to implement search functionality

dwightguth:
@traiansf the use of executeAndMatch for the exit code dates all the way back to the old way we were dealing with exit codes, where the exit code pattern was passed at execution time. Now the exit code pattern is computed at compile time from an annotation on a cell, so I saw no good reason to require the fully generic mechanism of matching on arbitrary patterns just in order to get the exit code.

PR #162, opened 2018-10-04
More llvm backend fixes

dwightguth:
* `\dv{SortString{}}` is unescaped
* terms shared unchanged by lhs and rhs are not reconstructed on rhs.


dwightguth:
@traiansf this is ready for review.

traiansf (approved):

PR #167, opened 2018-10-10
Print a stacktrace when function evaluation fails

h0nzZik:
A better reimplementation of https://github.com/kframework/k/pull/61. I still do not know where to put the tests, though (see the `mytest` directory).

@dwightguth can you review this, please?
@chathhorn (if interested)


kernel/src/main/java/org/kframework/EvaluationException.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2018 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+package org.kframework;
+
+import java.util.ArrayList;
+import org.kframework.kore.K;
+
+
+public class EvaluationException extends RuntimeException {
+    private int steps_;
+    private ArrayList<K> stacktrace_;
```
dwightguth: this should be List, not ArrayList. In general, in Java you should never declare variables to be of type ArrayList or HashSet or HashMap, but instead List or Set or Map.

kernel/src/main/java/org/kframework/EvaluationException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public EvaluationException(int steps, ArrayList<K> stacktrace) {
```
dwightguth: same here.

kernel/src/main/java/org/kframework/EvaluationException.java
```diff
@@ -0,0 +1,20 @@
...
+        super("EvaluationException");
```
dwightguth: Should have a better message than this. Something that actually describes the error.

kernel/src/main/java/org/kframework/EvaluationException.java
```diff
@@ -0,0 +1,20 @@
...
+        steps_ = steps;
+        stacktrace_ = stacktrace;
+    }
+
+    public int steps() { return steps_; }
+    public ArrayList<K> stacktrace() { return stacktrace_; }
```
dwightguth: List here as well.

kernel/src/main/java/org/kframework/EvaluationException.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2018 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+package org.kframework;
+
+import java.util.ArrayList;
+import org.kframework.kore.K;
+
+
+public class EvaluationException extends RuntimeException {
+    private int steps_;
```
dwightguth: also, you don't need the underscore. You can just say this.steps = steps; in the constructor.

mytest/arith-threaded.k
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2018 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
```
dwightguth: this is entirely the wrong place for these tests. they should be in a new directory under k-distribution/tests/regression-new

mytest/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
dwightguth: this is also the wrong test structure. You should be testing using the ktest makefile scripts that are used by the other tests in the directory I mentioned above.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -398,16 +398,19 @@ public static String enquoteString(byte[] value) {
 
     public String execute(K k, int depth, String file) {
         StringBuilder sb = new StringBuilder();
-        ocamlProgramHeader(sb, true);
+        ocamlProgramHeader(sb, true, "open List\n");
```
dwightguth: I don't like this; List has a bunch of functions in it that might not behave well when `open`ed. You should just prefix the functions with List.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -477,11 +480,34 @@ private void matchPatternAndPrint(StringBuilder sb) {
         sb.append("let _ = (try print_subst_binary subst (try_match res res (-1)) with Stuck c -> output_string subst \"\\x00\\x00\\x00\\x00\")\n");
     }
 
+    private void generateMain(StringBuilder sb) {
+        indentDepth = 0;
+        indented(sb).append("let main =\n");
```
dwightguth, line 496: this seems very weird and unnecessary to me. Why was this done? note that you haven't actually declared a function here.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -477,11 +480,34 @@ private void matchPatternAndPrint(StringBuilder sb) {
...
+        indented(sb).append("let res, steps = ");
         constructRun(sb, String.valueOf(depth));
+        sb.append("in\n");
+        indented(sb).append("output_string out (");
+        sb.append("(string_of_int steps) ^ \"\\n\"");
+        sb.append(" ^ (string_of_int 1) ^ \"\\n\"");
+        sb.append(" ^ print_k_binary(res))\n");
+        indentDepth = 1;
+        indented(sb).append(") with (FunctionEvalFailed' (n,stacktrace)) -> (output_string out (");
+        sb.append("(string_of_int n) ^ \"\\n\"");
```
dwightguth: This code is very difficult to make out, but it seems suspicious to me. I thought the whole idea was that we were just going to be returning a K term? why does this need to be modified at all?

dwightguth (changes requested):
this is still a much more intrusive change touching much more of the ocaml backend than I'd like. Recall our prior discussion; the idea was to just make the backend return a List of terms representing the stack when an exception occurs; As a result, you should need only minor modifications to the ocaml backend.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
h0nzZik: I did that because I could not find any way to use exceptions at toplevel. So I just declare a function and then call it.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
h0nzZik: This way we write 3 things to the output file:
1. number of steps
2. whether everything was ok or evaluation failed
3. the resulting term(s)

Concerning (2), I prefer to have such a such flag to the other options (either detecting a `List` term, or counting the number of terms), because the othre options are about having some special values. This way we may treat the term(s) opaquely. E.g. we do not have to rely on the assumption that a valid result is never a `List`.

Concerning (3), if we returned someting like `List(ListItem(k1), ..., ListItem(kn))`, the Java code would have to either print the term as-is, resulting in something like
```
ListItem ( isNice ( 0 ) )
ListItem ( isPretty ( 0 ) )
ListItem ( isBeautiful ( 0 ) )
ListItem ( <T>
  <k>
    9 / 0
  </k>
</T> )
```
, or somehow look into the `List` - which I am not sure how exactly to do it.

If your concern is that it is not easy to se what to generated ocaml code will do - I will clean it up.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
h0nzZik: Ok.

kernel/src/main/java/org/kframework/EvaluationException.java
h0nzZik: What about `Execution got stuck while evaluating a K function.` ?

h0nzZik:
@dwightguth May you have a look at it after the simplification?

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -453,35 +456,65 @@ private void constructRun(StringBuilder sb, String depth) {
...
-        } else {
-            sb.append("run_no_thread_opt");
-        }
-        sb.append("(input) (depth)\n");
-        sb.append("let () = output_string out (print_k_binary(res))\n");
-        sb.append("let () = try let subst = try_match res res (-1) in\n");
-        sb.append("let code = get_exit_code subst in\n");
-        sb.append("exit code\n");
-        sb.append("with Stuck(res) -> output_string out (print_k_binary res); exit 112\n");
+        sb.append("let main =\n");
```
dwightguth, line 472: This is still weird. You are declaring a top level value and then discarding its result. My advice would be to remove the `let main` and `let _ = main` entirely. If you get a syntax error, try adding a double semicolon to the end of the top-level expression.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -453,35 +456,65 @@ private void constructRun(StringBuilder sb, String depth) {
...
+        sb.append("  try (\n");
+        sb.append("    let res, _ = " + runfn() + "(input) (depth) in\n");
+        sb.append("    let () = output_string out (print_k_binary(res)) in\n");
+        sb.append("    let subst = try_match res res (-1) in\n");
+        sb.append("    let code = get_exit_code subst in\n");
+        sb.append("    exit code\n");
+        sb.append("  ) with\n");
+        sb.append("  | Stuck(res) -> (output_string out (print_k_binary res))\n");
+        sb.append("  | FunctionEvalFailed' (steps,stacktrace) -> (\n");
+        sb.append("    prerr_endline (\"Execution failed after \" ^ (string_of_int steps)  ^ \" steps (configuration dumped)\");\n");
```
dwightguth, line 482: I don't like that we've added output to stderr here, this will not play nicely with krun.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -1295,20 +1352,29 @@ public void apply(KApply k) {
...
+                    indented(sb).append("let lbl = ");
                     encodeStringToIdentifier(sb, functionLabel);
-                    sb.append(" and sort = \n");
+                    sb.append('\n');
+                    indented(sb).append("and sort = ");
                     encodeStringToIdentifier(sb, mainModule.sortFor().apply(functionLabel));
-                    sb.append(" in ");
-                    sb.append("match c with \n");
+                    sb.append(" in\n");
+                    indented(sb).append("try (\n");
```
dwightguth, line 1362: you can't have a try statement here, it will make these functions be unable to become tail recursive, which will break definitions that rely on tail recursion to bound the size of the stack.

dwightguth (changes requested):
I'm not convinced this diff is going to work out. You need to remove the try statement in K function bodies, but as far as I can tell that's pretty fundamental to your ability to generate a stack trace at all. If you can work around it, I am willing to review further, but I have no idea how one might go about it.

dwightguth:
I found this online: https://blog.janestreet.com/pattern-matching-and-exception-handling-unite/

Dunno if it will help you or not, but it seems pertinent to the problem you're having, so you can give it a try.

dwightguth:
I actually think it won't help though, because it's the tail recursive call that you're trying to put in a try block, which you just can't do. You might be able to work around it by putting everything /but/ the tail recursive call into a try block, at the cost of being missing some stack frames, but that's going to be a beast to implement. You're welcome to give it a shot though.

Honestly you might just be better off resigning yourself to using ocamldebug for debugging instead of trying to get this working though.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
h0nzZik: What if I just added a command-line parameter that would enable this feature only when the user asks for that? This way it would not be a performance problem.

dwightguth: That could definitely work, yes. Just bear in mind that this isn't a performance issue, it's a correctness issue; if you try to use this feature on a program that requires tail recursion, your program is going to segfault.

h0nzZik: Ok. In that case, I will add a test relying on tail recursion.

dwightguth: I think you might misunderstand: the stack trace being enabled at kompile time is going to make programs that depend on tail recursion segfault; it doesn't make much sense to add a test that will just fail.

h0nzZik: I will add a test testing that it will not segfault if the stacktracing is disabled.

dwightguth: Ah I see what you're saying. Okay yeah just make a function that recursively calls itself 100000 times.

h0nzZik: Ok. Btw can't we just eliminate the tail recursion before the backend starts? Just by transforming rules like
```
rule foo(P1::T1, ..., Pn::Tn) => foo(E1, ..., En)
```
to
```
rule foo(P1::T1 => E1, ..., Pn::Tn => En)
```
? Also, I understand the pipeline as
```
input definition => [frontend] => kore definition
                                          v
                                   [transformations]
                                          v
output   <=   [backend]   <=    kore definition
```
; is that correct?

dwightguth: That's roughly correct. And in response to your question, we could theoretically with the proper amount of work transform tail recursive functions into loops, but regardless this isn't a productive direction to take as it would be a very intrusive change to how the ocaml code is generated and that's not a good use of our time given that the ocaml backend is in the process of being phased out.

ehildenb:

dwightguth:

PR #173, opened 2018-10-18
general bugfix: Rules with inner rewrites are now possible in specifications.

denis-bogdanas:
Previously, due to a bug, such rules were causing NoSuchElementException.

Required to make this PR pass: https://github.com/kframework/evm-semantics/pull/261


dwightguth:
Can you help me understand why there would be nested rewrites in a specification? I understand that nested rewrites are in theory valid KORE semantics and have a specific logical meaning, but I seriously doubt that that meaning is what is being assigned by the Java backend, which makes me think that maybe this isn't the right fix.

dwightguth:
Oh wait, I think I misunderstood this.

dwightguth (approved):

PR #174, opened 2018-10-19
Haskell search

traiansf:
@dwightguth This should be adding search and match functionality for the Haskell backend.  The Haskell backend counterpart is implemented and tested in https://github.com/kframework/kore/pull/284

### DIFF
--- a/k-distribution/include/builtin/kast.k
+++ b/k-distribution/include/builtin/kast.k
@@ -76,7 +76,9 @@ module KSEQ-SYMBOLIC
   syntax #Pattern ::= "#True" [klabel(#True), symbol]
   syntax #Pattern ::= "#False" [klabel(#False), symbol]
   syntax #Pattern ::= K
-                    | #Pattern "#And" #Pattern [klabel(#And), symbol, assoc, comm, unit(#True), mlOp, format(%1 %2%n%3)]
+                    | "(" #Pattern ")" [bracket]
+                    > #Pattern "#Equals" #Pattern [klabel(#Equals), symbol, mlOp, format(%i%1%d%n%2%n%i%3%d)]
+                    > #Pattern "#And" #Pattern [klabel(#And), symbol, assoc, comm, unit(#True), mlOp, format(%i%1%d%n%2%n%i%3%d)]
                     > #Pattern "#Or" #Pattern [klabel(#Or), symbol, assoc, comm, unit(#False), mlOp, format(%i%1%d%n%2%n%i%3%d)]
 endmodule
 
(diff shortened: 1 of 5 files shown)


kore/src/main/scala/org/kframework/parser/kore/parser/KoreToK.scala
```diff
@@ -82,7 +82,7 @@ class KoreToK (headToLabel_ : java.util.Properties, sortAtt : Map[k.Sort, Att],
     case kore.Floor(s, rs, p) =>
       throw new TranslationError("Floor patterns currently unsupported")
     case kore.Equals(s1, s2, p, q) =>
-      throw new TranslationError("Equals patterns currently unsupported")
+      KORE.KApply(KLabels.EQUALS_K, apply(p), apply(q))
```
dwightguth: I'm not sure this is the way we want to go... I know we've done this in the past, but it might be better to add an `=` operator to kast.k since ==K returns a Bool which isn't exactly the same...

dwightguth (commented):
I'd like to hear your thoughts on one thing before I approve.

traiansf:
@dwightguth please re-review.  Sample output:
```
  VarResult ?=K
  <generatedTop>
    <k>
      final1
    </k>
    <generatedCounter>
      0
    </generatedCounter>
  </generatedTop>
#Or
  VarResult ?=K
  <generatedTop>
    <k>
      final2
    </k>
    <generatedCounter>
      0
    </generatedCounter>
  </generatedTop>
```

k-distribution/include/builtin/kast.k
```diff
@@ -76,6 +76,7 @@ module KSEQ-SYMBOLIC
   syntax #Pattern ::= "#True" [klabel(#True), symbol]
   syntax #Pattern ::= "#False" [klabel(#False), symbol]
   syntax #Pattern ::= K
+                    | #Pattern "?=K" #Pattern [klabel(?=K), symbol, mlOp, format(%1 %2%n%3)]
```
dwightguth: I would make the KLabel be `#Equals` for consistency.

dwightguth (approved):
Looks good but I suggest one minor alteration before you merge.

dwightguth (changes requested):
Actually I take it back, there's something else you need to fix. You need to remove the Var prefix from the variable name when converting from KORE back to K.

traiansf:
@dwightguth I think it's done

traiansf:
Jenkins:test this please.

traiansf:
Jenkins: test this please

dwightguth (approved):

traiansf:
Jenkins: test this please

traiansf:
Jenkins: test this please

PR #175, opened 2018-10-20
Debugg: State dumping for Java Backend to enable tools like KLab

ehildenb:
State dumping is enabled for `search`, `prove`, and `run` modes uniformly.

### DIFF
--- a/kore/src/main/scala/org/kframework/builtin/Sorts.scala
+++ b/kore/src/main/scala/org/kframework/builtin/Sorts.scala
@@ -39,4 +39,7 @@ object Sorts {
   val GeneratedCounterCell = Sort("GeneratedCounterCell")
 
   val Id = Sort("Id")
+
+  val Z3Query  = Sort("Z3Query")
+  val Z3Result = Sort("Z3Result")
 }
--- a/java-backend/src/main/java/org/kframework/backend/java/JavaBackendKModule.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/JavaBackendKModule.java
@@ -7,8 +7,10 @@ import com.google.inject.Module;
 import com.google.inject.TypeLiteral;
 import com.google.inject.multibindings.MapBinder;
 import com.google.inject.name.Names;
+import org.apache.commons.lang3.tuple.Pair;
 import org.kframework.backend.java.symbolic.InitializeRewriter;
 import org.kframework.backend.java.symbolic.JavaBackend;
+import org.kframework.backend.java.symbolic.JavaExecutionOptions;
 import org.kframework.compile.Backend;
 import org.kframework.kprove.KProve;
 import org.kframework.krun.ToolActivation;
@@ -24,6 +26,11 @@ import java.util.function.Function;
  * Created by dwightguth on 5/27/15.
  */
 public class JavaBackendKModule extends AbstractKModule {
+
+    @Override
+    public List<Pair<Class<?>, Boolean>> krunOptions() {
+        return Collections.singletonList(Pair.of(JavaExecutionOptions.class, true));
+    }
 
     @Override
     public List<Module> getKompileModules() {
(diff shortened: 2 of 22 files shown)


ehildenb:
@dwightguth if @denis-bogdanas confirms there isn't anything that needs fixed immediately with this, can you merge? We'd like to get this in because it's taking a while to get KLab working perfectly with it and there are other more immediate uses.

denis-bogdanas (approved):
Looks good.

kore/src/main/scala/org/kframework/builtin/Sorts.scala
```diff
@@ -39,4 +39,7 @@ object Sorts {
   val GeneratedCounterCell = Sort("GeneratedCounterCell")
 
   val Id = Sort("Id")
+
+  val Z3Query  = Sort("Z3Query")
```
dwightguth, line 43: these are not real sorts. Is there a reason why we need to encode these as fake K terms instead of just logging them directly?

dwightguth (changes requested):
Some obvious changes that need to be made:

* need to move Debugg to the java backend module, package org.kframework.backend.java, and also rename it to something better like ProofLogging or some such.
* need to move the options you added to somewhere better than GlobalOptionls. If you only need them for proofs, move them to KProveOptions, but if you want them for kprove, krun, and keq then you probably want to add them to JavaExecutionOptions.java and fix the bug that is causing those options to not be exposed on the command line.

denis-bogdanas:
Any news on this? We have 3 new hires in Bucharest. It would be helpful if they could use KLab.

dwightguth:
I'd still like to see the code quality improved, but the people who you're talking about could probably work off this branch potentially

ehildenb:
@dwightguth it doesn't only log in proof mode, also in normal execution and searching execution.

This is needed because we want to be able to use KLab for any execution type, so it doesn't make sense to name it `ProofLogging` because it's just generic logging of the Java backend (which can run in multiple modes).

I can move it to the Java backend package though.

kore/src/main/scala/org/kframework/builtin/Sorts.scala
ehildenb: The thought would be that eventually we'll have an actual sort of Z3Query and Z3Result (when it's integrated more properly). Not sure it will happen in the Java backend, but maybe in the Haskell backend. It also means that anything reading the logs only needs to understand KAST format and will be able to use standard getters to pick out the terms for z3 queries/results.

I can build the sort on the fly instead, but would prefer to call it out specially if we're going to use it specially.

ehildenb:
Hmmm, actually does moving it to the Java backend package make sense? It needs things from the front-end, for example the unparsing code.

ehildenb:
Jenkins: test this please

dwightguth:
The java backend depends on the kernel module so that's fine. If it's not just proving, you can give it another name, but Debugg is still a bad name. Moreover, GlobalOptions is used for options that affect the k framework as a whole and would be included in kompile and kast as well, which still isn't really suitable here. You can put it in JavaExecutionOptions then like I said.

ehildenb:
@dwightguth I have moved the Debugg to the `java-backend` module, and named it `StateLog` instead.

Also, I looked into the issue with JavaExecutionOptions. I'm not entirely sure, but I believe the problem is that options are exposed to the command line through code that lives in `kernel/`. `JavaExecutionOptions` is in `java-backend/`, but depends on `kernel/`. So when I try to add the import:

```
import org.kframework.backend.java.symbolic.JavaExecutionOptions
```

to `KRunOptions.java` (for example), so that I can do:

```
@ParametersDelegate
public JavaExecutionOptions JavaExecutionOptions = new JavaExecutionOptions();
```

it doesn't even know about the `org.kframework.backend` package (because `kernel` has no knowledge of `java-backend`).

So, I'm not sure what do, I don't think this issue with `JavaExecutionOptions` is directly fixable. So I'm inclined to move the options to `KRunOptions`, which would be a minimal change compared to the existing code but would move them out of `GlobalOptions`.

denis-bogdanas:
@ehildenb Can you move JavaExecutionOptions to kernel? This class actually contains interesting features.

dwightguth:
The solution with JavaExecutionOptions is to add a bind statement to the MapBinder of options to the guice modules for the Java backend. Please compare to OcamlKRunOptions for examples.

ehildenb:
Depends on #204 , closing for now.

PR #176, opened 2018-10-25
use KItem for MAP and SET keys

dwightguth:
Do not review yet.


dwightguth:
Merging with admin powers to break a dependency cycle. Ktest passes but rv match will need to be updated

PR #168, opened 2018-10-11
add [matching] attribute

daejunpark:
https://github.com/runtimeverification/verified-smart-contracts/pull/52#issuecomment-427107247

PR #177, opened 2018-10-26
kprove: Full variable substitution and function evaluation of specification rules

denis-bogdanas:
at initialization phase. Required for logging, performance and predictability.


java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -325,24 +325,36 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
...
             }
 
+            public org.kframework.backend.java.kil.Rule evaluateRule(org.kframework.backend.java.kil.Rule rule) {
+                //We need this ConsTerm only to evaluate the constraint. That's why we use an empty first argument.
+                ConstrainedTerm constraintHolder = new ConstrainedTerm(
+                        ConjunctiveFormula.of(termContext.global()),
+                        rule.getRequires(),
+                        termContext).expandPatterns(true);
+
+                ConjunctiveFormula constraint = constraintHolder.constraint();
```
daejunpark, line 340: could you explain what's the difference between `constraint` and `rule.getRequires()`?

java-backend/src/main/java/org/kframework/backend/java/kil/Term.java
```diff
@@ -82,6 +81,26 @@ public Term substituteAndEvaluate(Map<Variable, ? extends Term> substitution, Te
...
+     * <p>
+     * Denis, K branch `gnosis` only: Now that function evaluation results are cached,
+     * I suspect this function can be safely used instead of evaluate() and substituteAndEvaluate() in all contexts.
+     * These two no longer have performance advantages.
+     */
+    public Term fullSubstituteAndEvaluate(TermContext context) {
+        Map<Variable, ? extends Term> substitution = context.getTopConstraint().substitution();
+        Term newTerm = !Collections.disjoint(variableSet(), substitution.keySet())
+                ? (Term) this.accept(new SubstituteAndEvaluateTransformer(substitution, context))
+                : this;
```
daejunpark, line 98: hmm, so, the `evaluate` function does NOT consider the `context.getTopConstraint().substitution()`?

denis-bogdanas: Yes, `evaluate()` does not perform substitution. I believe we can safely replace all instances of `evaluate()` with function above, but there's no need to do this.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: The constraint is "fully evaluated and simplified" while `rule.getRequires()` is not. 2 ddifferences: 1. expandPatterns() can add new constraints, but this is not used in specifications that I know. 2. It is simplified. This is most important part, because it evaluates substitutions.

denis-bogdanas:
jenkins: test this please

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
daejunpark: Oh, now I see that `expandPatterns()` is applied to `rule.getRequires()`. Sounds good!

java-backend/src/main/java/org/kframework/backend/java/kil/Term.java
daejunpark: I didn't know that, and cannot see why.  There might be some specific reason why @andreistefanescu did that way.  Yes, let's do the drop-in replacement when needed later, not now.

daejunpark (approved):

daejunpark:
@denis-bogdanas looks good to me. But could you please add some tests, as I did in #168? Others can understand better what is the specific problem, and also prevent from the haskell backend team to make the same mistake.

denis-bogdanas:
@daejunpark This PR doesn't change the functionality. It only facilitates logging/debugging. So there's no need for tests.

PR #180, opened 2018-10-29
resolve overloaded 0-ary productions to normal form

dwightguth:
This does not handle overloads in the most general case, but it is assumed that overloads other than list terminators are handled in disambiguation during parsing. If I discover further cases where patterns have not been normalized in rules, I will revisit this.


dwightguth:
Don't merge this yet; it turns out to be insufficient to handle the case of List{A} vs. NeList{A}

dwightguth:
Never mind, this is actually ready for review.

dwightguth:
@traiansf can you review this tomorrow?

traiansf (approved):

PR #181, opened 2018-10-31
generate axioms for overloaded productions

dwightguth:
This generates axioms of the form `axiom {R} \equals{Exps{}, R}(exps{}(inj{Val{},Exp{}}(K1:Val{}),inj{Vals{},Exps{}}(K2:Vals{})),inj{Vals{},Exps{}}(vals{}(K1:Val{},K2:Vals{})))` for all overloaded productions. It also tags them with an attribute that can be used to infer semantically what the meaning of the axiom is.


traiansf (approved):
makes sense.  We'll have to figure it out how to handle it in the Haskell backend.

dwightguth:
Jenkins: test this please

PR #183, opened 2018-11-02
Support both Opam-2 and Opam-1

h0nzZik:
Some linux distributions (including Fedora 27) ship with new version of opam, i.e. opam-2.0, instead of opam-1.2. This PR aims to support opam-2.0.

One problem is that the archive https://github.com/ocaml/ocaml/archive/4.03.tar.gz" does not contain ocaml-4.03.0, but ocaml-4.03.1 (with a bug fix), and when building the patched ocaml version the build complains that the versions do not match.

This is a better version of https://github.com/kframework/k/pull/39.


dwightguth:
Looks good, but CI won't test this so before I approve I'd like to confirm that you've tested this on a fresh OS installation for both opam 1 and opam 2.

h0nzZik:
I have tested it on a fresh install (updated, though) of Ubuntu 16.04 (opam-1.2.2), Ubuntu 18.04 (opam-1.2.2) and Fedora 28 (opam-2.0.0~beta6). I used Vagrant for that purpose (see [my scripts](https://github.com/h0nzZik/c-semantics-deploy), or in a [particular revision](https://github.com/h0nzZik/c-semantics-deploy/tree/4a6df64f7d3cd873a6cc17d8ac5891c28277dfd6)). 

The `c-semantics-deploy` repo is only temporary, though - I will eventually merge it to the c-semantics repo.  I would like to run that weekly, so that we make sure everything works with current versions of packages.

dwightguth:
Jenkins: test this please

dwightguth (approved):

PR #186, opened 2018-11-06
Fix outdated help for kompile

hjorthjort:
Remove unused `--doc-style` flag and update valid backends list.

The help specifies that the available backends include Coq, but doesn't mention LLVM.

Also, the `--doc-style` indicates that there are PDF, LaTeX and HTML backends available for kompile, but as I understand it, that functionality has migrated to kdoc.

There is one flag remaining (which is experimental) which mentions `--backend=doc`. Maybe that should be removed as well.

I came across these inconsistencies while doing the tutorial, and spent some time being confused about this.


dwightguth (approved):

dwightguth:
Jenkins: test this please

dwightguth:
This build failure is unrelated to your change and I am working to address it.

dwightguth:
Jenkins: test this please

PR #187, opened 2018-11-06
Fix Array sort for Haskell and LLVM backends

dwightguth:
Basically I refactor the implementation of ARRAY in K to use List instead of Map, and import it as the implementation of the module for every backend except the ocaml backend.


traiansf:
Looks reasonable to me, but it seems tests are still failing.

traiansf (approved):

PR #191, opened 2018-11-11
Improved simplification of ConjunctiveFormula

denis-bogdanas:
When evaluating an equality, if it is also present in the topContext, it is removed from the evaluation context. 

Issue history
==============================
Evaluation in the context of itself can lead to information loss. As a result, a fix was implemented in the past that that, when simplifying term constraint, does not use any context:
Details: https://github.com/kframework/k-legacy/pull/2399#issuecomment-360680618

However, context is still needed when a term inside the constraint can be further evaluated with rules with side conditions, and those side conditions can be matched by other entries in the constraint.

To prevent information loss, now when an equality is evaluated, if it is also present in the constraint, it is temporarily removed from the constraint. Thus no equality can evaluate in the context of itself.

@dwightguth Please do not squash if possible. I prefer to keep the 2 commits separately in case we have to refer to this change in future.


daejunpark:
Could you please check if this doesn’t break the existing ERC20 proofs?

denis-bogdanas:
@daejunpark Yes all VSC tests pass: https://github.com/runtimeverification/verified-smart-contracts/pull/69
I added the same changes to K-gnosis.

denis-bogdanas:
@daejunpark @dwightguth Can anyone review please?

dwightguth:
I don't have any comments on functionality, but since it only touches the Java backend it seems okay from my perspective. I'd like @daejunpark to weigh in though.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -354,9 +358,7 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
 
             private Term evaluate(Term term, ConjunctiveFormula constraint, TermContext context) {
                 context.setTopConstraint(constraint);
-                Term newTerm = term.fullSubstituteAndEvaluate(context);
-                context.setTopConstraint(null);
-                return newTerm;
+                return term.fullSubstituteAndEvaluate(context);
```
daejunpark, line 362: could you please explain why `context.setTopConstraint(null);` is not needed here?

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -331,13 +331,17 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
             }
 
             public org.kframework.backend.java.kil.Rule evaluateRule(org.kframework.backend.java.kil.Rule rule) {
+                termContext.setTopConstraint(null);
```
daejunpark, line 335: why is this needed?

daejunpark: what was the original topConstraint here?

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -331,13 +331,17 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
...
                 //We need this ConsTerm only to evaluate the constraint. That's why we use an empty first argument.
                 ConstrainedTerm constraintHolder = new ConstrainedTerm(
                         ConjunctiveFormula.of(termContext.global()),
                         rule.getRequires(),
                         termContext).expandPatterns(true);
 
                 ConjunctiveFormula constraint = constraintHolder.constraint();
+                termContext.setTopConstraint(constraint);
+                //simplify the constraint in its own context, to force full evaluation of terms.
+                constraint = constraint.simplify(termContext);
```
daejunpark, line 345: what will happen if you simply the constraint before `setTopConstraint`?

daejunpark:
@denis-bogdanas `ConjunctiveFormula.java` looks good to me, but i cannot understand the changes made in `InitializeRewriter.java`. Could you please elaborate it?  (Sorry for the delay in reviewing this.)

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: Because it will be reset by setTopConstraint(null) above. Here it's just redundant. removing has no effect.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: Then it won't get fully simplified. In the current implementation an equality in constraint is simplified in the context of all other equalities. If constraint is null, then it will be simplified with no constraint.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: As Andrei once proposed, Constraint should be separated from context and passed as argument to every method that needs it. Then things would be much clearer. But that would be a very big change that is not worth at this point when java-backend is in maintenance mode.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: Essentially in this class constraint is simplified twice - once as part of expandPatterns(), in the presence of null context, the 2nd one - in the presence of itself. 2 steps are required because in the 1st step substitutions are calculated, and then you can use the constraint with proper substitutions to simplify itself once again.

All other changes are just refactoring/removing redundant code.

denis-bogdanas:
@daejunpark Added my replies.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
daejunpark: i mean, what's the problem with the 1st simplification under the non-null `constraint`?

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
daejunpark: if you are fully sure that the deleted line has no effect, then ok.  i'm still suspicious though. but we will find the problem if any later.

denis-bogdanas: Constraint is reset at the beginning of each rule evaluation:
https://github.com/kframework/k/blob/017c591d4b20802238afe431154659f232c8cca8/java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java#L350

So there's no need to reset it at the end. Code with no effect complicates understanding and should be removed. 

And the code before this PR is actually wrong. Not buggy, because it has no effect on final result, but sub-optimal performance and debug-wise.
Before this PR, constraint was set before LHS, but reset after both LHS and RHS.
As a result only LHS was evaluated in the presence of constraint.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: Initial constraint doesn't contain substitutions. First simplification generates those substitutions. Then we need another simplification to simplify in the presence of substitutions. Code is really cryptic and if you really want to understand all implications you have to look into many other classes not touched by this change.

But one thing you can probably agree - 2 simplifications instead of one cannot break things:)

daejunpark (approved):

PR #196, opened 2018-11-14
Splitted STRING-BUFFER between concrete, symbolic and kore.

traiansf:


k-distribution/include/builtin/domains.k
```diff
@@ -586,21 +598,22 @@ module STRING-BUFFER-SYNTAX
...
+  imports STRING-BUFFER-HOOKED
+endmodule
 
-  rule .StringBuffer => ""
+module STRING-BUFFER-SYMBOLIC [symbolic]
+  imports STRING-BUFFER-IN-K
+endmodule
 
-  rule StringBuffer2String(S:String) => S
+module STRING-BUFFER-KORE [kore]
```
dwightguth: this isn't actually what we want. We want to use the hooks for the ocaml and llvm backends, not just the ocaml backend. So you should delete this module and the kast attribute above.

dwightguth (changes requested):

k-distribution/include/builtin/domains.k
traiansf: @dwightguth Does this seem right?

k-distribution/include/builtin/domains.k
```diff
@@ -586,20 +598,16 @@ module STRING-BUFFER-SYNTAX
...
 
-module STRING-BUFFER-SYMBOLIC [symbolic]
-  imports STRING-BUFFER-SYNTAX
-  imports STRING-SYNTAX
-  imports BOOL-SYNTAX
-
-  syntax StringBuffer ::= String
-
-  rule .StringBuffer => ""
+module STRING-BUFFER-CONCRETE [concrete]
```
dwightguth: this is functionally correct but these modules now seem redundant to me. We should be able to do it with 2 modules and not 4.

traiansf: Is this what you meant?

dwightguth (approved):

dwightguth (changes requested):
actually it looks like you're missing +String in the IN-K version.

traiansf:
I know I am, I was just not sure it's necessary because I thought we already have `+String` on strings. But I agree it is.  However, I'm wondering whether it will use the same label as for the concatenation of strings and, if so, whether there is some automatically generated axiom to use string concatenation in the case that the left operand is a string.

dwightguth:
it's definitely a different label.

traiansf:
@dwightguth please re-review

dwightguth (approved):

PR #208, opened 2018-11-29
add IMAP builtin module

daejunpark:
This is an algebraic map theory that allows the symbolic map reasoning in the K level.

Currently, it is a monomorphic, and here IMAP is a map from Int to Int. But it is quite sufficient for many low-level languages such as EVM, LLVM, and x86.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -89,6 +89,24 @@ module ARRAY
   imports ARRAY-KORE
 endmodule
 
+module IMAP-SYNTAX
+    imports INT
+    syntax IMap [smt-prelude] // (define-sort IMap () (Array Int Int))
+    syntax IMap ::= store  ( IMap , Int , Int ) [function, smtlib(store),  smt-prelude]
+    syntax Int  ::= select ( IMap , Int )       [function, smtlib(select), smt-prelude]
+endmodule
+
+module IMAP-SYMBOLIC [symbolic]
+    imports IMAP-SYNTAX
+    rule select(store(M, K0, V), K) => V            requires K0  ==Int K
+    rule select(store(M, K0, V), K) => select(M, K) requires K0 =/=Int K
+endmodule
+
+module IMAP
+    imports IMAP-SYNTAX
+    imports IMAP-SYMBOLIC
+endmodule
+
 module MAP
   imports LIST
   imports SET


dwightguth (commented):
This is definitely not what an integer map module in the k prelude should look like. Please compare to the TypeMap module in the C++ semantics for a better idea, although the parallels are not complete.

daejunpark:
@dwightguth could you please provide a link to TypeMap?

daejunpark:
@dwightguth do you mean this https://github.com/kframework/c-semantics/blob/master/semantics/cpp/language/common/map.k by TypeMap?

could you explain why you think this is not enough?

btw, this is not for concrete execution, only for symbolic execution; and we don't need high performance.

dwightguth:
I mean, this code may suit your purposes for verifying a particular smart contract (although if you are modifying the semantics just for verification that's kinda sketchy), but I'm saying it's simply not structured syntactically or semantically in a manner consistent with how a K prelude module of a map from Int to Int should be written.

MrChico:
FYI @daejunpark this is what we are conforming to wrt this store: https://github.com/dapphub/evm-semantics/pull/3/files#diff-494b4c66d4128fe395a8c9607fab3629. I made the map even more conservative by not allowing `select` or `store(store...` to reduce unless the key in question can be shown to be distinct from all other entries of the store. This further improves the soundness of proofs, as the store will now take an unwieldy form if there is any uncertainty whether a collision can occur or not. For an example, see this discussion: https://github.com/dapphub/klab/pull/96#issuecomment-445487768

ehildenb:
@dwightguth this encoding of Maps is about storing the sequence of updates/reads from the map instead of storing it as key/value pairs (think transactional memory or journaling filesystems). The point is that we can do symbolic execution with `IM:IMap` variables directly instead of relying on any backend support with this encoding. We won't get away from the fact that we need two operations, an `update` which takes in the current `IMap` and the new key/value, and a `read` which takes in the current `IMap`.

This differs from the traditional algebraic associative-commutative map in that `union` will make less sense (indeed implementing union over this map would lead to a large state-space), but seeing the sequence up updates for this map is much simpler.

It's also not just for one smart contract that this is useful, it's useful for executing *any* smart contract (or C program with maps, etc...) where we want to start the map out fully symbolic. Currently we must specify the memory/storage layout fully upfront for smart contract specs, but this would let us start with a fully symbolic map and see the sequence of updates that results (enabling lightweight state-space exploration tools). It may not sound difficult to specify the memory layout up front, but currently if you get it wrong then K is not very helpful at all.

Another benefit is the improved coverage of proofs using this Map over the traditional K map. Currently, in a K spec if you specify a memory layout, it comes with implicit side-conditions which state that the map keys are non-unifiable. With this way of storing maps, you must make such side-conditions explicit if the prover can't prove it on its own.

Perhaps though, a better solution would be to allow using the existing map syntax, but kompile it to these operations which sequence the map effects instead.

ehildenb:
@MrChico I'm not sure that comparing against every key in the map actually improves soundness as much as it degrades performance. Since the structure of this map means that more current values are in outer layers of the map `update` operator, it should be fully sufficient to iterate through them in reverse order and only go over a key if you can prove non-unifiability.

MrChico:
> @MrChico I'm not sure that comparing against every key in the map actually improves soundness as much as it degrades performance. Since the structure of this map means that more current values are in outer layers of the map `update` operator, it should be fully sufficient to iterate through them in reverse order and only go over a key if you can prove non-unifiability.

Do you not agree that this "christmas spec" is an example of how a store that doesn't require key entries to be distinct is unsound? https://github.com/dapphub/klab/pull/96#issuecomment-445487768

ehildenb:
I'm not sure about the christmas spec, because AFAICT it uses the old maps.

I'm looking at this rule:

```
    //Reduces IMaps where multiple entries share the same key
    rule store(store(M, K0, V0), K1, V1) => store(M, K0, V1)
        requires K0 ==Int K1
        andBool notBool #inKeys(M, K1)
```

and wondering why you need the side condition `notBool #inKeys(M, K1)`? If we have that `M` is nested inside a `store(M, K0, V0)` then the semantics is that the update `K0 <- V0` happened most recently to map `M`, so if there is another key in `M` which unifies with `K0` then it's value is replaced by `V0`.

My understanding is that the two maps `store(store(M, K0, V0), K0, V1)` and `store(M, K0, V1)` are semantically equivalent. Is it correct @daejunpark?

MrChico:
> I'm not sure about the christmas spec, because AFAICT it uses the old maps.
> 
> I'm looking at this rule:
> 
> ```
>     //Reduces IMaps where multiple entries share the same key
>     rule store(store(M, K0, V0), K1, V1) => store(M, K0, V1)
>         requires K0 ==Int K1
>         andBool notBool #inKeys(M, K1)
> ```
> 
> and wondering why you need the side condition `notBool #inKeys(M, K1)`? If we have that `M` is nested inside a `store(M, K0, V0)` then the semantics is that the update `K0 <- V0` happened most recently to map `M`, so if there is another key in `M` which unifies with `K0` then it's value is replaced by `V0`.
> 
> My understanding is that the two maps `store(store(M, K0, V0), K0, V1)` and `store(M, K0, V1)` are semantically equivalent. Is it correct @daejunpark?

No, the christmas spec uses the new store. We just still use the mapping notation in our spec format.

`store(store(M, K0, V0), K0, V1)` and `store(M, K0, V1)` might be semantically equivalent, in the sense that they yield the same result when `select` is applied to them, but that does not mean that K can prove them to be equal. If the RHS of a spec contains the second expression then a term proof where the store equals the first will not satify it. The rewrite above can be seen as a "normal form" of the store. Without a normal form the store would look different based on "operational details" of code not the functional behavior:
```
i += 1;
i += 1;
i += 1;
```
and 
```
i += 3;
```


The side contition `notBool #inKeys(M, K1)` is added to avoid the edge case where we have a store
`store(store(M, K0, V0), K1, V1)` where we don't know whether `K0 == K1`. If we update with `K1 += V2` we would get `store(store(M, K0, V0), K1, V1 + V2)` and we might prove 
```
K0 |-> V0
K1 |-> V1 => V1 + V2
```
even though this spec does not make sense in the case `K0 == K1`. By adding the extra side condition, this spec would not be provable.

ehildenb:
@daejunpark how about we just add this to KEVM for now, and then later when we have proper parametric sorts `Map{Int,Int}` we can switch to using that instead?

MrChico:
After some discussions with @ehildenb , we came to the following conclusions:
We have been pursuing two different approaches. Mine, (which we may call a "normal form store") assumes storage is given explicitly in the `<store>` cell, as in:
```
<store> store(ACCT_ID_STORE, KEY VALUE_PRE) => store(ACCT_ID_STORE, KEY VALUE_PRE) </store>
```
while Everetts (which we might call "abstract store") keeps the `<store>` cell abstract and instead imposes restrictions in the `requires` (and `ensures`) clauses:
```
<store> ACCT_ID_STORE_PRE => ACCT_ID_STORE_POST </store>
...
requires
select(ACCT_ID_STORE_PRE, KEY) ==K VALUE_PRE
ensures
select(ACCT_ID_STORE_POST, KEY) ==K VALUE_POST
```
however, in order to make sure that no other storage locations than `KEY` have been meddled with, we also need to add the condition that `select(ACCT_ID_STORE_POST, A) ==K select(ACCT_ID_STORE_PRE, A)` for all `A =/=K KEY`. This can be done by adding extending the conditions to:
```
requires
select(ACCT_ID_STORE_PRE, KEY) ==K VALUE_PRE
A =/=K KEY
ensures
select(ACCT_ID_STORE_POST, KEY) ==K VALUE_POST
select(ACCT_ID_STORE_POST, A) ==K select(ACCT_ID_STORE_PRE, A)
```

It seems like the abstract store would work better for imports, as it will be hard for K to unify concrete nested `stores` if the order differs. OTOH, with the store abstract, K will not reduce an expression like:
`select(ACCT_ID_STORE_PRE, 0)`, even in the presence of a `require select(ACCT_ID_STORE_PRE, 0) ==K VALUE`, which ends up making gas expressions a bit gnarly:
```
VGas >=Int #if ( ( ( select( ACCT_ID_STORE_PRE , 0 ) ) ==K 0 ) andBool ( notBool( ( `_+Int__INT`( ( select( ACCT_ID_STORE_PRE , 0 ) ) , ABI_x ) ) ==K 0 ) ) ) #then 20517 #else 5517 #fi
```
instead of 
```
VGas >=Int #if (ABI_x ==K 0 ) #then 20517 #else 5517 #fi
```
P.S. the lemma  
```
 rule select(M, K) => V
      requires select(M, K) ==K V
```
does not seem to alleviate the situation

ehildenb:

dwightguth:
Closing as we are not going in this direction with the prelude.

PR #210, opened 2018-12-01
[BUGFIX] Conversion from Set2List omits ListItem klabels

ehildenb:


ehildenb:
@dwightguth I've added tests for converters between `Set` and `List` for the Java backend. It seems that prior to this the Java backend's variants of these were never tested. This was uncovered when I went to execute the VMTests on KEVM so that I could compare performance to the new Haskell backend and realized that the Java backend actually doesn't pass any tests because of this bug!

The last commit I am not very happy with. I found no way to make a compact set of changes which make `list2set` behave correctly, in fact I even had to add the singleton list case to `domains.k` directly! I think it's troubling that I cannot have a similar change to `list2set` that I had to `set2list` to get it working, so there must be something going on with the underlying data-structures that I am not aware of.

k-distribution/include/builtin/domains.k
```diff
@@ -273,7 +273,7 @@ module COLLECTIONS
 
   syntax List ::= Set2List(Set) [function, hook(SET.set2list)]
   syntax Set ::= List2Set(List) [function, functional, hook(SET.list2set)]
-
+  rule List2Set(ListItem(E) LS) => SetItem(E)
```
daejunpark: I don't understand this rule. Could you explain?

ehildenb: @daejunpark it's as confusing to me as it is to you. Without this line, the function `List2Set(ListItem(1))` remains unevaluated. Remove the line then build K then run `make -j6` in k-distribution/tests/builtins/collections to reproduce.

daejunpark: what will happen if you replace `LS` with `.List`?

ehildenb:
@daejunpark and @dwightguth, my understanding is this: The `set2list` functions worked OK by adding the `wrapListItem` for the resulting elements because `BuiltinSet` extends `AssociativeCommutativeCollection`, which provides a well-behaved `baseTerms()` accessor which `getKComponents` takes advantage of.

`BuiltList` does not extend said class, and so has no `baseTerms()` accessor. Moreover, the `getKComponents` accessor seems to have poor behavior. I think the only "correct" fix to this is to make an `AssociativeCollection` class, which `AssociativeCommutativeCollection` extends, and then have `BuiltinList` extend that.

As far as I can tell, the `BuiltinList` is also treated specially because it's used for KList? Not sure.

I'm not happy with this sequence of changes to make everything work, especially (1) the addition to `domains.k`, and (2) the additional `if` statement in `singleton` before the assertion. But every change is necessary for the added tests to pass.

k-distribution/include/builtin/domains.k
ehildenb: No, with `.List` it does not work.

ehildenb:
@dwightguth this should be ready for review. The singleton case has to be dealt with on its own, which is why `list2set` had to be lifted to work over `Term` instead of `BuiltinList` as we discussed.

dwightguth (approved):

PR #212, opened 2018-12-04
KILtoSMTLib: Improved translation of constraint and implication queries

denis-bogdanas:
KILtoSMTLib: 1 Improved translation of constraint and implication queries.
A term that has no smtlib attribute will be abstracted to a fresh anonymous variable. Previously the whole LHS/RHS of the equality that contained the term was abstracted to a variable. This lead to significant information loss making most Z3 queries useless.
This translation was also enabled for implciation RHS now. It is proved sound.

2. Making Z3 queries a bit prettier for easy debug.

Please only review 2nd commit. 1st is just refactoring.

A summary of the improvement for Z3 query generation
=============================================
Intro: A z3 query can have one of 3 results: unsat, sat or unknown. kprove treats unknown results as sat. So for the purpose of deductive capability, only unsat results matter. There are 2 types of queries: constraint and implication.

For constraint formulas
------------------
Previously, if we had a formula like

foo(A) == X /\ (baz(B) orBool goo(B) ) == true

where foo and bar and baz have smtlib attribute, but `goo` doesn't, it would be translated into
foo(A) == X /\ F1 == true

Where
F1 == (baz(B) orBool goo(B) )

The whole LHS of the 2nd equality was substituted with a fresh var.
As you see, this led to information loss that could lead to query result "unknown" where in fact it was unsat.

After K update:
Now only the term without smtlib is translated into fresh var, so we have this Z3 query instead:

foo(A) == X /\ (baz(B) orBool F1 ) == true

where
F1 == goo(B)

In some cases this is enough to prove the query unsat.

For implication formulas
------------------
In the old version of K:
  - For implication LHS substitution with fresh vars worked same way as for constraint formulas.
  - For implication RHS, if a term without smtlib attribute was encountered, an exception would be thrown which appears in logs as "query build failure" message.

In the new version of K:
  - Improved term substitution works for both LHS and RHS according to algorithm above. Variables map is shared between LHS and RHS.

Proof of soundness
--------------------------
For any logical formula foo(bar(X)):
```
foo(X, bar(Y)) <=> foo(X, F1) /\ F1 = bar(X)
```

where `F1` is a fresh var that substitutes all instances of bar(X) in initial formula.

But because kprove doesn't use the substitution itself when generating the query, the formula that Z3 actually tries to prove is reduced:
```
foo(X, F1)
```

Now proving that a formula is unsat is equivalent to:

`forall(vars, formula == false)`

But the reduced formula is more general than the full formula, thus if reduced formula is unsat, then full formula is too:

```
forall(X,F1, foo(X, F1) == false)
  =>
forall(X, foo(X, F1) /\ F1 = bar(X) == false)
```
because the 2nd is just a particuar case of the 1st.

Alternatively, it follows from this general truth:
`forall(X, not f(X)) => forall(X, not(f(X) /\ g(X)) )`

This is true for any form of f, whether constraint or implication, regardless of whether it uses existentially quantified vars inside or not.

I did a paper proof for the exact query that kprove generates for implications, with existentially quantified vars, and as expected it is also true. Below is only the formula:
```
L(X) -> exists(Y, R(X,Y)) 

is translated into:

forall(X, not (L(X) /\ not exists(Y, R(X, Y)) ) )
```

Now to question why Stefan didn't implement this sort of substitution before. At the time there were no methods to easily debug whether a Z3 query is fine-grained enough. So he just preferred the query generation to fail with a warning and he would add smtlib attribute for function terms mentioned in the warnings. But now we have easy to use log messages from which it is simple to figure out whether a Z3 query conveis all the required information of the original ConjunctiveFormula or not. Try the option `--debug-z3-queries` on some spec.

A bit about performance
----------------
On specs I worked on recently, with old K 90% of implication queries were generated anyway and only 10% failed with a warning. With new K all 100% were generated, as expected, total z3 time was 1-2 s more total execution time was about the same. So I guess this has no impact on performance on big specs.

Update: Middle ground solution
=============================================
Suppose we have 3 implications:

1. `f(X) => g(X)`
2. `f(X) => g(X, ugly(X))`
3. `f(X, ugly(X)) => g(X, ugly(X))`

where `f` and `g` are functions annotated with `[smtlib]`, thus translatable. And `ugly` is not annotated.

Old Z3 translation:
```
1. f(X) => g(X)
2. fail
3. fail
```

Current Z3 translation:
```
1. f(X) => g(X)
2. f(X) => g(X,Y)
3. f(X,Y) => g(X,Y)
```
(this is not the only change in "current Z3 translation", only the part related to this example)

"Middle ground" Z3 translation:
```
1. f(X) => g(X)
2. fail
3. f(X,Y) => g(X,Y)
```

Reason: if Y is a fresh variable that appears only on implication RHS, then implication can only be true if RHS is simplifiable to something that doesn't contain `Y`. This is unlikely to happen in kprove. Most likely such implications appear because kprove doesn't have all the information to further evaluate the term that caused the implication. Thus they are unpro and can be dropped. Just an observation based on an isolated case, so I really don't have a better argument. Here is the implication that gave me this idea:

```
...
  implies
ConjunctiveFormula(
  equalities:
    _==K_(
      chop(#bufElm(...)),,
      #bufElm(...)
    )
```

This implication happened because after I removed range conversion lemmas, there was no lemma for `#bufElm(_) < pow256 => true`. And rule for `chop` wants this in its side condition. So most likely this sort of queries happen due to rules with unsatisfiable side conditions.

How much this will increase performance? No idea. On old z3 translation, on a spec with ~4k implications, Only 10% impliations were failing at query build stage. On the current spec `encodeTransactionData` - we'll see in Jenkins.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
@@ -122,14 +122,14 @@ public class ConstrainedTerm extends JavaSymbolicObject {
     }
 
     /**
-     * Checks if this constrained term implies the given constrained term, assuming the variables
-     * occurring only in the given constrained term (but not in this constrained term) are
+     * Checks if {@code this} implies {@code matchRHSTerm}, assuming the variables
+     * occurring only in {@code matchRHSTerm} (but not in {@code this}) are
      * existentially quantified.
      */
-    public ConjunctiveFormula matchImplies(ConstrainedTerm constrainedTerm, boolean expand, Set<String> matchingSymbols) {
-        ConjunctiveFormula constraint = ConjunctiveFormula.of(constrainedTerm.termContext().global())
+    public ConjunctiveFormula matchImplies(ConstrainedTerm matchRHSTerm, boolean expand, Set<String> matchingSymbols) {
+        ConjunctiveFormula constraint = ConjunctiveFormula.of(matchRHSTerm.termContext().global())
                 .add(data.constraint.substitution())
-                .add(data.term, constrainedTerm.data.term)
+                .add(data.term, matchRHSTerm.data.term)
                 .simplifyBeforePatternFolding(context);
         if (constraint.isFalseExtended()) {
             return null;
@@ -149,7 +149,7 @@ public class ConstrainedTerm extends JavaSymbolicObject {
 
         /* apply pattern folding */
         constraint = constraint.simplifyModuloPatternFolding(context)
-                .add(constrainedTerm.data.constraint);
+                .add(matchRHSTerm.data.constraint);
         if (constraint.isFalse()) {
             return null;
         }
@@ -187,13 +187,15 @@ public class ConstrainedTerm extends JavaSymbolicObject {
         context.setTopConstraint(data.constraint);
         constraint = (ConjunctiveFormula) constraint.evaluate(context);
 
-        Set<Variable> rightOnlyVariables = Sets.difference(constraint.variableSet(), Sets.union(variableSet(), termContext().getInitialVariables()));
-        constraint = constraint.orientSubstitution(rightOnlyVariables);
+        Set<Variable> matchRHSOnlyVars = Sets.difference(constraint.variableSet(), Sets.union(variableSet(), termContext().getInitialVariables()));
+        constraint = constraint.orientSubstitution(matchRHSOnlyVars);
 
-        ConjunctiveFormula leftHandSide = data.constraint;
-        ConjunctiveFormula rightHandSide = constraint.removeBindings(rightOnlyVariables);
-        rightHandSide = (ConjunctiveFormula) rightHandSide.substituteAndEvaluate(leftHandSide.substitution(), context);
-        if (!leftHandSide.implies(rightHandSide, rightOnlyVariables)) {
+        ConjunctiveFormula implicationLHS = data.constraint;
+        ConjunctiveFormula implicationRHS = constraint.removeBindings(matchRHSOnlyVars);
+        implicationRHS = (ConjunctiveFormula) implicationRHS.substituteAndEvaluate(implicationLHS.substitution(), context);
+
+        boolean implies = implicationLHS.implies(implicationRHS, matchRHSOnlyVars);
+        if (!implies) {
             return null;
         }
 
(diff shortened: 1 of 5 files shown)


denis-bogdanas:
@traiansf Can you review please?

traiansf:
I'm pretty convinced that the first transformation is sound, but I don't fully understand what the second amounts to.  Could you please provide a concrete example, as you did for the first?

andreistefanescu (approved):
Overall looks good.

One thing to consider is that the sound smt encoding of implication is `\exists X . \phi /\ \forall Y . (not \psi)`, where `X` are the variables that abstract terms in the lhs, and `Y` are the variables that abstract terms in the rhs containing some rhs-only variables (because those are themselves universally quantified). So this would be unsound if any of the variables have empty sorts. In practice, I think we already assume in other places that the sorts are not empty, so it should not be an issue.

denis-bogdanas:
@traiansf It's similar to constraint transformation. From javadoc of function `translateImplication`:
An implciation `left -> right` will be translated into Z3 as `left /\ !right` (e.g. its negation), and it has to be unsat fo the implicaton to be proved. So in the end we have a conjunction, just like for `constraint transformation`.

Same logic applies here to why transformation is sound. It remains valid in the presence of existentially quantified terms as well (as Andrei explained), and is only invalid when you have empty sorts, but we assume they are not empty.

The difference in previous implementation is that in case of terms without `smtlib` attribute, for constraint formulas they would be overly-approximated, while for implication formulas the whole translation would fail. Everything else stays the same. The cause of this difference was not soundness but easy debugging. Now with better debug log messages we no longer need this.

I'm going to merge this once tests pass.

ehildenb:
@denis-bogdanas I think you should have gotten approval from a current maintainer of K for these changes as well before merging. It's dangerous to be merging whenever tests pass without review on code that has so few tests anyway (especially missing tests of the negative cases, eg. tests where we make sure our translation to Z3 isn't saying things are provable when they shouldn't be).

Also, if new commits are pushed to a PR, those new commits need to be reviewed as well.

denis-bogdanas:
@ehildenb I see. And who is the current maintainer of K?

ehildenb:
Mostly @dwightguth is who I've been deferring to (obviously you and I are also current maintainers, but you can't review your own PR for obvious reasons). But at the very least @daejunpark should have been a reviewer on this PR.

PR #213, opened 2018-12-04
Pass --smt-prelude to Haskell backend

phillipharr1s:


traiansf (approved):

traiansf (changes requested):
On a second glance, I think it might have been better to use a `@ParametersDelegate` to `org.kframework.utils.options.SMTOptions` and use the `--smt-prelude` from there (as we might be using more of the SMT options available there in the future, anyway). @u-- What do you think?

dwightguth:
You definitely should use SMTOptions.

phillipharr1s:
Hmm ok

traiansf:
And it turns out you can access it already through `kProveOptions.smt`

phillipharr1s:
Oops yeah didn't notice that existed.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -118,6 +118,10 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
                         args.add("--depth");
                         args.add(options.depth.toString());
                     }
+                    if (options.experimental.smt.smtPrelude != null) {
```
dwightguth, line 121: I'm fairly sure this will always evaluate to null if you are running kprove and not krun. You want to inject SMTOptions in the constructor above and then access it directly from the injected instance.

dwightguth: then again I guess it's fine since this is the execute method...

phillipharr1s: I don't know much about this repo, but this at least works in the case I'm testing.

traiansf: We will definitely need it for kprove, too. @u-- did you try it with `kprove`?

traiansf: Anyway, I'm going to approve it so you can test it now, but I might modify it later as we need support for kprove.

traiansf (approved):

PR #214, opened 2018-12-05
split DEFAULT-STRATEGY into concrete vs symbolic versions

dfilaretti:
We need this since the Haskell backend does not support `[anywhere]` rules.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -1308,10 +1308,28 @@ module STRATEGY
     configuration <s> $STRATEGY:K  </s>
 endmodule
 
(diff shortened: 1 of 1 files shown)


k-distribution/include/builtin/domains.k
```diff
@@ -1308,10 +1308,28 @@ module STRATEGY
     configuration <s> $STRATEGY:K  </s>
 endmodule
 
-module DEFAULT-STRATEGY
-    imports syntax STRATEGY
+// module DEFAULT-STRATEGY
```
traiansf: You should remove comment lines.

dwightguth (approved):

PR #215, opened 2018-12-05
`kserver --socket=./my/socket`

h0nzZik:
Specify the socket for KServer

Example
```
./k-distribution/bin/kserver --socket ./my-socket &
wait 10
env KSERVER_SOCKET=(pwd)/my-socket ./k-distribution/bin/kompile --help
```


kernel/src/main/java/org/kframework/kserver/KServerOptions.java
```diff
@@ -15,4 +15,8 @@
 
     @Parameter(names={"--port", "-p"}, description="The port to start the server on.")
     public int port = 2113;
+
+    @Parameter(names={"--socket"}, description="The socket to start the server on.")
```
dwightguth: this should say something like "The directory to put the unix domain socket in"

dwightguth (changes requested):
looks good and potentially very valuable but can you please modify the pom so that mvn verify makes use of this functionality?

h0nzZik:
@dwightguth Can you spot any obvious reason why is my integration (`pom.xml` and `spawn-kserver`) broken?

dwightguth:
@h0nzZik this looks like the same bug that @ehildenb is trying to fix. If you just wait a couple days it should hopefully be resolved.

dwightguth (approved):

kernel/src/main/java/org/kframework/kserver/KServerOptions.java
h0nzZik: Done.

dwightguth (approved):

PR #204, opened 2018-11-27
Expose JavaExecutionOptions on command line

ehildenb:
@dwightguth this exposes the JavaExecutionOptions on the command-line. I attempted to mark them as experimental, but they still just show up in `--help` and not `-X`.

@denis-bogdanas you may be interested in this as well.

PR #217, opened 2018-12-07
Ensure that with `--io on/off` we always get an extra newline on stdin

ehildenb:
Not sure if this is the correct fix, but it would make sense given the fact that we are consistently seeing one extra/too-few newlines in the stdin stream of SIMPLE programs in other PRs.


ehildenb:
Do not merge until I get a chance to check that the problems on #204 go away.

dwightguth (approved):

ehildenb:
Jenkins: test this please

dwightguth (approved):

PR #218, opened 2018-12-10
Error messages: improved error message for `--deterministic-functions`

denis-bogdanas:
when non-deterministic function is found. Print the 2 candidate rules and respective candidate results.

Unused code is used in K-gnosis and will be used in future PRs.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
@@ -1,0 +1,59 @@
+// Copyright (c) 2015-2018 K Team. All Rights Reserved.
+package org.kframework.backend.java.util;
+
+import org.kframework.attributes.Location;
+import org.kframework.backend.java.kil.Rule;
+import org.kframework.utils.file.FileUtil;
+
+import java.io.File;
+import java.util.Collections;
+import java.util.Formatter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Denis Bogdanas
+ * Created on 27-Aug-18.
+ */
+public class RuleSourceUtil {
+    private static final Map<Rule, String> cache = Collections.synchronizedMap(new HashMap<>());
+
+    public static boolean sourceShortEnough(Rule rule) {
+        Location location = rule.getLocation();
+        return location != null && location.endLine() - location.startLine() <= 20;
+    }
+
+    public static String loadSource(Rule rule) {
+        if (!cache.containsKey(rule)) {
+            if (rule.getSource() != null && rule.getLocation() != null) {
+                cache.putIfAbsent(rule, FileUtil.loadFragment(new File(rule.getSource().source()), rule.getLocation()));
+            }
+        }
+        return cache.get(rule);
+    }
+
+    public static void printSourceAndRule(Rule rule) {
+        File source = rule.source().isPresent() ? new File(rule.getSource().source()) : null;
+        System.err.format("\nSource: %s %s\n", source, rule.getLocation());
+        if (sourceShortEnough(rule)) {
+            System.err.println(loadSource(rule));
+        } else {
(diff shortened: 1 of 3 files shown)


java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
```diff
@@ -0,0 +1,59 @@
...
+        return cache.get(rule);
+    }
+
+    public static void printSourceAndRule(Rule rule) {
+        File source = rule.source().isPresent() ? new File(rule.getSource().source()) : null;
+        System.err.format("\nSource: %s %s\n", source, rule.getLocation());
+        if (sourceShortEnough(rule)) {
+            System.err.println(loadSource(rule));
+        } else {
+            System.err.println("rule too long or location unknown...");
```
dwightguth: this is not really nice; you might as well fall back on printing /something/ useful.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    public static void printRuleAndSource(Rule rule) {
```
dwightguth, line 35: having both a printSourceAndRule and printRuleAndSource is really confusing, especially when one calls another helper function that has a lot of duplicated code. Did you really intend to keep both around?

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
```diff
@@ -0,0 +1,59 @@
...
+public class RuleSourceUtil {
+    private static final Map<Rule, String> cache = Collections.synchronizedMap(new HashMap<>());
+
+    public static boolean sourceShortEnough(Rule rule) {
+        Location location = rule.getLocation();
+        return location != null && location.endLine() - location.startLine() <= 20;
+    }
+
+    public static String loadSource(Rule rule) {
+        if (!cache.containsKey(rule)) {
```
dwightguth, line 27: I'm fairly sure this function can return null, which I don't think is what you want.

dwightguth (changes requested):
it would be good to get some unit tests on your printing functions, too.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
denis-bogdanas: removed one of them.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
denis-bogdanas: Ok I slightly improved the message to the extent possible. When rule is too long developer can always look at the source, using the printed location.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
denis-bogdanas: @dwightguth replied and fixed what's possible.

Since this class is only for error reporting, is it really worth testing? It doesn't affect rewrite engine functionality. We are late with deadlines in VSC team.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
denis-bogdanas: Nope. Null source is valid option and is properly handled in `appendRuleAndSource`.

java-backend/src/main/java/org/kframework/backend/java/util/RuleSourceUtil.java
dwightguth: no, it's fine, I was more worried about the corner cases but it seems you've handled them now.

dwightguth (approved):

PR #228, opened 2018-12-21
Build Haskell backend as a submodule, hook into Maven

ehildenb:
This is in preparation of running the K tests on all the backends, not just the Java backend.

It will help root out deficiencies in the Haskell backend.


ehildenb:
Note that this should wait for: https://github.com/kframework/kore/pull/393

Once that is merged, I'll update the submodule and this will be ready.

traiansf (commented):
Looks good, but I think some of the default values here: `org/kframework/backend/haskell/HaskellKRunOptions.java`  need to change as well.

traiansf:
Also, it would be nice to add a test that the haskell backend works, or maybe even calling `make k-test` in the `kore` repo.

ehildenb:
@traiansf I'm holding off the changes for making sure the Haskell backend works for another PR. My plan is to also make it possible to run the test-set in k-distribution with any of the backends, then to have a list of tests to run with each backend. Currently we just run some of them with OCaml and others with Java.

The Haskell backend integration tests will be tested on it's PR job for now, so there I think I'm going to focus on running as many of the tests in k-distribution that the Haskell backend can in a reasonable amount of time.

I also plan on making a weekly build job that run all integration tests on nightly builds to find a nightly release which is working for all the semantics we care about. Once that build is up, the Haskell backend PR will use that version for testing instead of the nightly.

ehildenb:
Also, what defaults in HaskellKRunOptions need to change? We discussed on Slack, and I think the conclusion was that the current defaults actually work well. Right now it will default to using the binaries bundled in `k-distribution/target/release/k/bin`, and the binaries built by the haskell backend are being copied there.

You can override this by explicitly calling the options, which the testing for the Haskell backend needs. But the default is to use the version pinned by the submodule.

ehildenb:
Jenkins: test this please

haskell-backend/pom.xml
```diff
@@ -28,12 +28,42 @@
     </dependency>
   </dependencies>
 
+  <properties>
```
dwightguth: I'm reasonably sure this is not necessary as I added this flag specifically to deal with cmake.

dwightguth (changes requested):
looks good except one minor issue but I agree this should be tested.

haskell-backend/pom.xml
ehildenb: Removed.

ehildenb:

ehildenb:
@dwightguth this is ready for re-review.

I've parameterized the k-distribution integration tests in the backend to run them in (Makefile variable `KOMPILE_BACKEND`). I've also switched `1_k/2_imp` over to using the Haskell backend by default, to prevent regressions.

k-distribution/tutorial/1_k/2_imp/Makefile
```diff
@@ -1,3 +1,5 @@
 SUBDIRS=lesson_1 lesson_2 lesson_3 lesson_4 lesson_5
 
+KOMPILE_BACKEND=haskell
```
dwightguth: This is a bad choice as it's going to mess with the students building the tutorial in the next couple weeks and expose them to any lingering bugs in the Haskell backend. You should put the regression test for the Haskell backend in tests/regression-new

dwightguth (changes requested):
Minor change needed but the code still looks good

k-distribution/tutorial/1_k/2_imp/Makefile
ehildenb: Which regression tests are you referring to?

dwightguth: that's the name of the folder: k-distribution/tests/regression-new

ehildenb: Alright, I've made a new regression testsuite imp-haskell which exercises the haskell backend, leaving the tutorial alone.

dwightguth (approved):

dwightguth:
I approved but I'd like Traian to approve too before this gets merged.

ehildenb:
@dwightguth I've added an option `-Dhaskell.backend.skip`, tested it locally and it seems to work.

k-distribution/include/ktest-keq.mak
```diff
@@ -2,6 +2,7 @@
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 # path to the kompile binary of this distribuition
 KOMPILE=$(abspath $(MAKEFILE_PATH)/../bin/kompile)
+KOMPILE_BACKEND=java
```
traiansf, line 5: Don't you need a ? here to allow changing the backend?

traiansf (approved):

traiansf:
Nevermind my comment; I saw some tests have fixed backend.

k-distribution/include/ktest-keq.mak
ehildenb: The keq testset always uses the Java backend for now. It's a different execution mode.

PR #230, opened 2018-12-21
java-backend bugfix: ConjunctiveFormula.simplify()

denis-bogdanas:
properly restore the constraint on all execution paths.

Change description: added one big try-finally block, and restored initial constraint in finally{}.

After I implemented improved simplification in some previous PR, constraint changes during simplification process. Previously, it was not restored on all exit paths. Now it is properly restored in a finally block. Finally block is needed to minimize changes, because there are multiple return statements in the function.

DO NOT MERGE


java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
```diff
@@ -386,127 +386,130 @@ private ConjunctiveFormula simplify(boolean patternFolding, boolean partialSimpl
...
-        } while (change);
+                equalities = pendingEqualities;
+            } while (change);
 
-        context.setTopConstraint(originalTopConstraint);
-        return ConjunctiveFormula.of(substitution, equalities, disjunctions, global);
+            return ConjunctiveFormula.of(substitution, equalities, disjunctions, global);
+        } finally {
+            context.setTopConstraint(originalTopConstraint);
+        }
```
daejunpark, line 512: Github diff is not smart enough to show the meaningful difference here. :(

So, the main difference is the addition of the `finally` block here? Or, do you have other changes inside the `try` block?

daejunpark: If the `finally` block is the only change made, then it looks good to me, and you can merge.

denis-bogdanas: Yes, finally block is the only change.

daejunpark (approved):

ehildenb:

denis-bogdanas:
@ehildenb I want to run these 4 PRs on VSC first, but I'm busy with other tasks for the moment. Will get to them in 1-2 days. The 4 PRs were already reviewed by @daejunpark .

PR #231, opened 2018-12-21
java-backend bugfix: #getKLabelString

denis-bogdanas:
1. #getKLabelString properly handles variable arguments.
2. Added check that requires clause of a spec rule is not false.

Tested in verified-smart-contracts/k-test.

DO NOT MERGE

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/builtins/MetaK.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/builtins/MetaK.java
@@ -14,6 +14,7 @@ import org.kframework.backend.java.kil.Variable;
 import org.kframework.backend.java.symbolic.ConjunctiveFormula;
 import org.kframework.backend.java.symbolic.CopyOnWriteTransformer;
 import org.kframework.backend.java.symbolic.PatternMatcher;
+import org.kframework.kore.KVariable;
 
 import java.util.HashSet;
 import java.util.Set;
@@ -201,6 +202,8 @@ public class MetaK {
     public static StringToken getKLabelString(Term term, TermContext context) {
         if (term instanceof KItem) {
             return StringToken.of(((KItem) term).kLabel().toString());
+        } else if (term instanceof KVariable){
+            return null;
         } else {
             return StringToken.of("");
         }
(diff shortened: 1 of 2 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -351,6 +351,12 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
                 termContext.setTopConstraint(constraint);
                 //simplify the constraint in its own context, to force full evaluation of terms.
                 constraint = constraint.simplify(termContext);
+                if (constraint.isFalseExtended()) {
+                    StringBuilder sb = new StringBuilder();
+                    sb.append("Rule requires clause evaluates to false:\n");
+                    RuleSourceUtil.appendRuleAndSource(rule, sb);
+                    throw KEMException.criticalError(sb.toString());
+                }
```
daejunpark, line 363: This is orthogonal to the `#getKLabelString` fix, right?

java-backend/src/main/java/org/kframework/backend/java/builtins/MetaK.java
```diff
@@ -201,6 +202,8 @@ public static KItem getKLabel(KItem kItem, TermContext context) {
     public static StringToken getKLabelString(Term term, TermContext context) {
         if (term instanceof KItem) {
             return StringToken.of(((KItem) term).kLabel().toString());
+        } else if (term instanceof KVariable){
+            return null;
```
daejunpark, line 206: Could you explain in which case we need this?  An example?

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: It is made possible by `#getKLabelString` and is also a safety measure to avoid this sort of issue in future with builtin functions.

java-backend/src/main/java/org/kframework/backend/java/builtins/MetaK.java
denis-bogdanas: https://github.com/runtimeverification/verified-smart-contracts/blob/b9aa9543670a4b4bcf3d1450d44f66d028404cec/k-test/k-test-spec.ini#L59-L66

ehildenb:

denis-bogdanas:
@daejunpark  Can you approve pls? We are already using new `#getKLabelString` semantics in K-gnosis. All tests on Jenkins pass.

daejunpark (approved):

PR #232, opened 2018-12-21
Pr term init fix

denis-bogdanas:
java-backend: bugfix: properly evaluate the term, constraint and ensures clause before calling proveRule().

Before this change incorrect top constraint was sometimes present before calling lhs.expandPatterns(), from ProcessProofRules phase of another spec rule.

Summary of the changes before proveRule:
- Fully evaluate and simplify the constraint.
- LHS (initial term) is built from fully evaluated constraint and rule LHS which is already fully evalauted.
- initialTerm.expandPatterns() was moved from proveRule() to InitializeRewriter where it should have been in the first place.
- Evaluate the ensures clause.
- Build RHS (target term) from evalauted ensures and rule RHS, which is already fully evaluated.
- reset constraint (not needed, just for safety)

Problems with previous implementation:
- constraint and ensures clause were not fully evaluated. The previous pre-processing in ProcessProofRules only evaluates the LHS and RHS. This whole sequence is extraordinarily poorly written and error-prone, that's why it's not obvious.
- Cosntraint in termContext was not reset after initialization phase. Consequently, inside SymbolicRewriter.proveRule(), line `initialTerm.expandPattern()` could evaluate with wrong context and lead to funny issues.

@daejunpark Please review only 2nd commit. 1st commit is same as #231 .

DO NOT MERGE (I want to merge all PRs in a certain sequence).

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
@@ -228,9 +228,20 @@ public class InitializeRewriter implements Function<Module, Rewriter> {
             List<ConstrainedTerm> proofResults = javaRules.stream()
                     .filter(r -> !r.att().contains(Attribute.TRUSTED_KEY))
                     .map(r -> {
-                        ConstrainedTerm lhs = r.createLhsPattern(termContext);
-                        ConstrainedTerm rhs = r.createRhsPattern();
+                        //Build LHS with fully evaluated constraint. Then expand patterns.
+                        ConjunctiveFormula constraint = processProofRules.getEvaluatedConstraint(r);
+                        ConstrainedTerm lhs = new ConstrainedTerm(r.leftHandSide(), constraint, termContext);
+                        termContext.setTopConstraint(constraint);
+                        lhs = lhs.expandPatterns(true);
+
+                        //Build RHS with fully evaluated ensures. RHS term is already evaluated.
+                        ConjunctiveFormula ensures = (ConjunctiveFormula) processProofRules.evaluate(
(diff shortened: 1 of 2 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -371,6 +373,21 @@ public ProcessProofRules invoke(GlobalContext rewritingContext, long initCounter
...
+                //We need this ConsTerm only to evaluate the constraint. That's why we use an empty first argument.
+                ConstrainedTerm constraintHolder = new ConstrainedTerm(
+                        ConjunctiveFormula.of(termContext.global()),
+                        rule.getRequires(),
+                        termContext).expandPatterns(true);
+
+                ConjunctiveFormula constraint = constraintHolder.constraint();
+                termContext.setTopConstraint(constraint);
+                //simplify the constraint in its own context, to force full evaluation of terms.
+                constraint = constraint.simplify(termContext);
```
daejunpark, line 390: just to make sure, `constraint.simplify` ensures that each conjunct is simplified assuming all other conjuncts but itself, right?

daejunpark: ok. i just double-checked that the simplify function handles that.

daejunpark (approved):
what is described seems to be implemented in the second commit.
not sure whether this change will introduce (or trigger) another (hidden) bugs, though.
but no better option other than merging this for now.

ehildenb:
Jenkins: test this please

ehildenb:
Jenkins: test this please

ehildenb:

PR #237, opened 2018-12-28
Make Jenkins testing script runnable locally without Jenkins environment setup

ehildenb:
The script is organized slightly (so that parameters come at the top), and more notifications about build process are produced.

Also, the rv-match testing is done differently. Instead of cloning to subdirectory `$WORKSPACE/k-pr`, we're cloning directly to `$WORKSPACE`. K-Exercises tests are unaffected, but rv-match is now cloned to a subdirectory `$WORKSPACE/rv-match`, and we use manually update the K submodule inside RV-match to point directly at the HEAD of the checkout of K at `$WORKSPACE`. (so no more `rm ... ; cp ...` going on, just `git ...` operation).

### DIFF
--- a/src/main/scripts/jenkins
+++ b/src/main/scripts/jenkins
@@ -41,12 +41,14 @@ rm -rf k-exercises
 git clone "$K_EXERCISES_REPO" k-exercises
 
 notif "RUNNING K EXERCISES"
+exit_status='0'
 (   cd k-exercises/tutorial
     make -j"$NPROCS"
-)
+) || exit_status="$?"
 
 notif "KILLING KSERVER"
 ./k-distribution/target/release/k/bin/stop-kserver || true
+[[ "$exit_status" == '0' ]] || exit "$exit_status"
 
 notif "CLONING/PREPARING RV-MATCH"
 rm -rf rv-match
--- a/src/main/scripts/jenkins
+++ b/src/main/scripts/jenkins
@@ -14,6 +14,10 @@ KSERVER_SOCKET="$WORKSPACE/kserver"
 KSERVER_LOG="$WORKSPACE/kserver.log"
 export KSERVER_SOCKET
 
+notif "CLEANING REPO"
+mvn clean
+git clean -dfx
+
 notif "SETTING UP OPAM/PERL ENVIRONMENTS"
 eval `opam config env`
 eval `perl -I$HOME/perl5/lib/perl5 -Mlocal::lib`
--- a/src/main/scripts/jenkins
+++ b/src/main/scripts/jenkins
@@ -8,7 +8,7 @@ notif() {
             "\n" >&2
 }
 
-NPROCS="${NPROCS:-$(grep -c '^processor' /proc/cpuinfo)}"
+NPROCS="${NPROCS:-$(nproc)}"
 WORKSPACE="${WORKSPACE:-$(pwd)}"
 
 K_EXERCISES_REPO="${K_EXERCISES_REPO:-git@github.com:kframework/k-exercises.git}"
--- a/src/main/scripts/jenkins
+++ b/src/main/scripts/jenkins
@@ -10,6 +10,10 @@ notif() {
 
 NPROCS="${NPROCS:-$(grep -c '^processor' /proc/cpuinfo)}"
 WORKSPACE="${WORKSPACE:-$(pwd)}"
+
+K_EXERCISES_REPO="${K_EXERCISES_REPO:-git@github.com:kframework/k-exercises.git}"
+RVMATCH_REPO="${RVMATCH_REPO:-git@github.com:runtimeverification/rv-match.git}"
+
 KSERVER_SOCKET="$WORKSPACE/kserver"
 KSERVER_LOG="$WORKSPACE/kserver.log"
 export KSERVER_SOCKET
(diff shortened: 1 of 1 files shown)


src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
 set -e
 
-nprocs="$(grep -c '^processor' /proc/cpuinfo)"
+notif() {
+    echo -e "\n$@" \
+            "\n================================================================================" \
+            "\n" >&2
+}
 
+NPROCS="${NPROCS:-$(grep -c '^processor' /proc/cpuinfo)}"
```
dwightguth: You can just call nprocs here.

src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
+
+K_EXERCISES_REPO="${K_EXERCISES_REPO:-git@github.com:kframework/k-exercises.git}"
+RVMATCH_REPO="${RVMATCH_REPO:-git@github.com:runtimeverification/rv-match.git}"
+
+KSERVER_SOCKET="$WORKSPACE/kserver"
+KSERVER_LOG="$WORKSPACE/kserver.log"
+export KSERVER_SOCKET
+
+notif "CLEANING REPO"
+mvn clean
```
dwightguth: You don't need this if you run `git clean -fdx`

src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
+git clean -dffx
```
dwightguth, line 22: Also what does `-ff` do? Unless this option cleans submodules then this is still not going to work.

src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
+notif "STARTING KSERVER"
+./k-distribution/target/release/k/bin/spawn-kserver "$KSERVER_LOG"
 sleep 5
-make -j"$nprocs"
-../../k-distribution/target/release/k/bin/stop-kserver || true
 
-echo "=== RUNNING RV-MATCH TESTS ==="
-cd ${WORKSPACE}
+notif "CLONING K EXERCISES"
+rm -rf k-exercises
```
dwightguth: The workspace should be clean at this point so this is not needed

src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
+git clone "$K_EXERCISES_REPO" k-exercises
+
+notif "RUNNING K EXERCISES"
+exit_status='0'
+(   cd k-exercises/tutorial
+    make -j"$NPROCS"
+) || exit_status="$?"
```
dwightguth, line 45: Has this been tested in the case it fails? I ask because the last time you wrote something in approximately this syntax I had to change it because it didn't actually work.

src/main/scripts/jenkins
```diff
@@ -2,33 +2,66 @@
...
-git submodule update --init -- c-semantics
-rm -rf k
-cp -r ${WORKSPACE}/k-pr k
-mvn verify -U -DskipKTest -DbuildProfile=x86_64-gcc-glibc
+git clone "$RVMATCH_REPO" rv-match
+(   cd rv-match
+    git submodule update --init -- c-semantics k
+    cd k
+    git fetch ../../
+    git checkout FETCH_HEAD
```
dwightguth, line 58: What is this? I liked it better when we were just copying the directory into place. This feels error prone to me.

dwightguth (changes requested):

src/main/scripts/jenkins
ehildenb: From `man git clean`:

```
       -f, --force
           If the Git configuration variable clean.requireForce is not set to false, git clean will
           refuse to delete files or directories unless given -f, -n or -i. Git will refuse to delete
           directories with .git sub directory or file unless a second -f is given.
```

ehildenb: So it makes sure that the `rv-match` and `k-exercises` repositories are actually deleted along with any other junk.

src/main/scripts/jenkins
ehildenb: The reason to do this is because we want the rv-match tests to be runnable from directly in the K directory (without having to move to a temporary directory or work from one directory up like we were before).

The `git fetch ../../` will fetch the current `HEAD` of the repository at `../../`, which is the current checkout of K. `git fetch` with an anonymous remote (here `../../`) updates `FETCH_HEAD` to whatever the remote `HEAD` is. Nothing too fancy here.

Benefits of this approach:

-   Jenkins script is directly runnable from the K repo directory (trying to use `cp` to do this would result in recursively copying the rv-match repo into itself).
-   We are sure that the rv-match submodule is updated to exactly the version of K that is being tested, because we fetch it directly.

src/main/scripts/jenkins
ehildenb:
```
[dev@arch-ehildenb k-pr]$ exit_status='0'
[dev@arch-ehildenb k-pr]$ echo "$exit_status"
0
[dev@arch-ehildenb k-pr]$ ( cd k-distribution/
> false
> ) || exit_status="$?"
[dev@arch-ehildenb k-pr]$ echo $exit_status
1
[dev@arch-ehildenb k-pr]$ 
```


The parenthesis spawn a sub-shell (so that any `cd` that happen don't get carried over past that part of the script). The semantics of `$?` is that the last command run (in sub-shell or otherwise) by the bash interpreter will be stored in `$?`. So if `cd ...` fails, then we'll get the `exit_status` set to the failure code of `cd`, and if `cd` succeeds and `make` fails, then `exit_status` will be whatever `make` returns.

ehildenb:
Jenkins: test this please.

dwightguth (approved):

PR #240, opened 2019-01-04
fix smtlib attribute for %Int  and /Int

dwightguth:
This should be functionally correct; I have not tested whether it stands up to symbolic reasoning though. Would appreciate it if you tested it on the example that made you report this bug in the first place.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -356,8 +356,8 @@ module INT
                  Int "*Int" Int                 [function, functional, left, smtlib(*), latex({#1}\mathrel{\ast_{\scriptstyle\it Int}}{#2}), hook(INT.mul)]
                /* FIXME: translate /Int and %Int into smtlib */
                /* /Int and %Int implement t-division */
-               | Int "/Int" Int                 [function, left, smtlib(div), latex({#1}\mathrel{\div_{\scriptstyle\it Int}}{#2}), hook(INT.tdiv)]
-               | Int "%Int" Int                 [function, left, smtlib(mod), latex({#1}\mathrel{\%_{\scriptstyle\it Int}}{#2}), hook(INT.tmod)]
+               | Int "/Int" Int                 [function, left, smtlib((ite (>= #1 0) (div #1 #2) (div (- #1) (- #2)))), latex({#1}\mathrel{\div_{\scriptstyle\it Int}}{#2}), hook(INT.tdiv)]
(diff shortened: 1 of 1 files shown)


daejunpark:
the current encoding is:
```
x /Int y = if x >= 0 then x div y else   (-x) div (-y)
x %Int y = if x >= 0 then x mod y else -((-x) mod (-y))
```

while another equivalent encoding is:
```
x /Int y = if x >= 0 then x div y else -((-x) div y)
x %Int y = if x >= 0 then x mod y else -((-x) mod y)
```
in other words (conceptually):
```
x /Int y = sgn(x) * (abs(x) div y)
x %Int y = sgn(x) * (abs(x) mod y)
```

i haven't analyzed which one is better for z3 to solve, but at least the latter seems more uniform.

let me check.

daejunpark:
Let me formulate a bit for the record.

Let:
```
q = x div y
r = x mod y
```
where `x`, `y`, `q`, and `r` are integers (but `y =/= 0`).

Then we have (for any definition of `div` and `mod`):
```
x = y*q + r
```



----

Now let me formulate the different definitions of `div`.
(The definition of `mod` is determined by `div`, since `r = x - y*q`.)

Let `x / y` be the real arithmetic division result.


In SMT, `div` finds `q` such that `0 <= r < |y|`.

```
q := floor(x / y)  if y > 0
  :=  ceil(x / y)  if y < 0
```

In Java, `div` simply truncates numbers below the point, e.g., `1.3 => 1`, and `-1.3 => -1`, that is, flooring for positive, and ceiling for negative.

```
q := floor(x / y)  if (x / y) >= 0
  :=  ceil(x / y)  if (x / y) <  0
```

In Python, `div` takes flooring of `x / y`.

```
q := floor(x / y)
```




----

For example,

SMT:
```
 x =  y *  q +  r
-----------------
 4 =  3 *  1 +  1
 4 = -3 * -1 +  1
-4 =  3 * -2 +  2
-4 = -3 *  2 +  2
```

Java:
```
 x =  y *  q +  r
-----------------
 4 =  3 *  1 +  1
 4 = -3 * -1 +  1
-4 =  3 * -1 + -1
-4 = -3 *  1 + -1
```

Python:
```
 x =  y *  q +  r
-----------------
 4 =  3 *  1 +  1
 4 = -3 * -2 + -2
-4 =  3 * -2 +  2
-4 = -3 *  1 + -1
```

NOTE:
- SMT and Java agree when `x` is positive.
- SMT and Python agree when `y` is positive.
- Java and Python agree when `x / y` is positive.

ehildenb:

dwightguth:
@daejunpark do you still need this?

daejunpark:
i'm closing this for now, as we are currently not relying on z3 for reasoning about these operations, but will reopen when we need to.  btw, thanks @dwightguth for initiating this.

PR #241, opened 2019-01-04
fix form of rewrites

dwightguth:
This should fix the issue with the representation of rewrites in kore.


phillipharr1s (approved):

ttuegel:
@grosu [realized](https://runtimeverification.slack.com/archives/CC360GUTG/p1546705547159200), that this translation is not correct; [instead](https://runtimeverification.slack.com/archives/CC360GUTG/p1546708827210100) it should be:
for rewrite rules: `\rewrites(\and(\and(requires, ensures), L), R)`
for function rules: `\implies(\and(requires, ensures), \equals(L, R))`

dwightguth:
Since we decided to disallow ensures in axioms, I will merge this. I will do a followup PR that existentially quantifies RHS variables.

PR #242, opened 2019-01-09
Removed autoimport of definition main module.

traiansf:
This was generating ugly duplication of rules because of bad hashing of rules.  Instead, users will be required to manually include the main module of the definition in the main verification module.  Daejun said that is acceptable.


traiansf:
Jenkins: test this please

daejunpark:
what's the difference between auto-import and manual-import for the ugly duplication issue?

traiansf:
@daejunpark nothing interesting, really. Just the order in which attributes are serialized as strings, which leads to a hash value being different, too.

daejunpark:
ok, the usage should be fine.  but the code would need @dwightguth's approval.

dwightguth:
Jenkins: test this please

dwightguth (approved):
Looks fine if this is what the verification team wants.

ehildenb:

PR #243, opened 2019-01-09
Added ML implication, exists, and forall.

traiansf:


k-distribution/include/builtin/kast.k
```diff
@@ -84,6 +84,7 @@ module KSEQ-SYMBOLIC
                     > #Pattern "#Or" #Pattern [klabel(#Or), symbol, assoc, comm, unit(#False), mlOp, format(%i%1%d%n%2%n%i%3%d)]
                     > #Pattern "#Implies" #Pattern [klabel(#Implies), symbol, mlOp, format(%i%1%d%n%2%n%i%3%d)]
                     > "#Exists" #KVariable "." #Pattern [klabel(#Exists), symbol, mlOp]
+                    > "#Forall" #KVariable "." #Pattern [klabel(#Forall), symbol, mlOp]
```
traiansf: @vladciobanu using `>` there would indicate that exists has precedence over forall.  I would use `|` instead to signal they have the same precedence.

dwightguth (approved):
Looks good, but please make sure you've tested it before you merge.

PR #250, opened 2019-01-15
Adding simplification attribute to non-constructor-based equalities

traiansf:


kore/src/main/java/org/kframework/builtin/KLabels.java
```diff
@@ -42,6 +42,7 @@
     public static final KLabel SetItem = KLabel("SetItem");
     public static final KLabel DotMap = KLabel(".Map");
     public static final KLabel List = KLabel("_List_");
+    public static final KLabel DotSet = KLabel(".Map");
```
daejunpark: `.Set` or `.Map`?

daejunpark: BTW, the order of these constants makes little sense.

traiansf: Thanks for catching this!  And abput the order, I agree, but it's not like it was very ordered before :-)

kore/src/main/scala/org/kframework/compile/ConstructorsChecks.scala
```diff
@@ -0,0 +1,56 @@
+package org.kframework.compile
```
virgil-serbanuta: Some of the other files have copyright statements.

kore/src/main/scala/org/kframework/compile/ConstructorsChecks.scala
```diff
@@ -0,0 +1,56 @@
...
+    * Checks that a lhs pattern is constructor-based, i.e., an application of a
+    * function symbol to a list of constructor terms.
+    * @param m
+    * @param term
+    * @return
+    */
+  def constructorBased(m: Module, term: K): Boolean = term match {
+    case kapp: KApply => immutable(kapp.klist.items).forall(constructorOnly(m))
+    case t: KAs => constructorBased(m, t.pattern)
+    case other => true
```
virgil-serbanuta: Could you add a comment explaining why returning true here works?

virgil-serbanuta (commented):
This looks reasonable to me, but I'd like someone with more kframework experience to approve.

dwightguth (changes requested):
I'd really rather we didn't add to the amount of scala we are using in the project unnecessarily as it's already a source of problems on occasion

traiansf:
@dwightguth what is your exact suggestion here?

dwightguth:
I mean my suggestion is to write ConstructorChecks in Java.

traiansf:
@dwightguth I've rewritten it in Java.  Please re-review.

dwightguth (approved):

PR #252, opened 2019-01-16
Fix the format attribute for <generatedTop>

ehildenb:
This PR manually sets the format attribute for the `<generatedTop>` cell, after the `generatedCounter` production is added.

It could be moved to after the compilation pass if there are more `generated*` cells to not be printed.

Specifically, the logic is:

-   Do not print `generated*` sub-cells of `generatedTop`.
-   If there is only one non-`generated*` subcell of `generatedTop`, print that cell directly without printing the `generatedTop` cell.
-   If there are multiple non-`generated*` subcells of `generatedTop`, print the `generatedTop` cell as well as all non-`generated*` subcells.

I've tested this manually on Imp with/without the wrapping `<T>` cell, and it seems to behave.


ehildenb:
I've tested this on the Java and Haskell backends manually.

dwightguth (changes requested):
I thought we were going to make it not print the generatedTop cell period...

ehildenb:
The current behavior of `master` is to add the `<generatedTop>` cell if the top-level `configuration` declaration has multiple top-level cells.

I thought the goal was to maintain current behavior, but I can change it if you like. Note this will make for a change in output of KEVM (at the very least).

dwightguth:
I mean the goal is to have a consistent behavior across all backends. We can do that while also making the behavior more uniform overall, even though it is a change.

ehildenb:
@dwightguth do you think this should be pulled out into its own compiler pass which manually sets the formatting correctly for the `generatedTop`, and ignores anything from a set of given cells? It would make it easier to ignore other `generated` cells in the future.

ehildenb:
@dwightguth this has separated the code out into its own Kompile pass. I think we should wait to change the behavior of printing `<generatedTop>`. It will require the output to change in lots of other repos which I think we should wait until things quiet down a bit to do.

It should be easier now that it's pulled into its own compiler pass. I've enabled it for both the normal pipeline and the Kore pipeline.

kernel/src/main/java/org/kframework/compile/GeneratedTopFormat.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private static Production resolve(Production prod) {
+        if (prod.klabel().isDefined() && prod.klabel().get().equals(KLabels.GENERATED_TOP_CELL)) {
+            List<Integer> cellPositions = new ArrayList<Integer>();
+            int i = 1;
+            for (ProductionItem p: JavaConverters.seqAsJavaList(prod.items())) {
+                if (p instanceof NonTerminal) {
+                    NonTerminal nt = (NonTerminal) p;
+                    // if (! nt.sort().equals(Sorts.GeneratedCounterSort)) {
+                    if (! nt.sort().name().equals("GeneratedCounterCell")) {
```
dwightguth: why was this commented out?

ehildenb: It never existed, I put it there as a reminder that it's a better way to do it. Couldn't get it to work though.

dwightguth: Try reversing the order of the operands.

ehildenb:

kernel/src/main/java/org/kframework/compile/GeneratedTopFormat.java
ehildenb: It was an issue with how I was access the Scala object.

dwightguth (approved):

PR #253, opened 2019-01-16
Machine Int (bit-vector) semantics of left shift operation.

sdasgup3:
### Symptom
extractMInt(shlMInt(32'1, 1), 0, 32) yields 32'0 instead of 2.

### Reason
In case the bitwidth of the bitvector equals INTEGER.SIZE, the bitvector
is represented as a fixed width bit-vector `Int32Token` of size == 32.
`extractMInt` calls the Int32Token::extract which is:

    ```
    public BitVector extract(int beginIndex, int endIndex) {
      int resultBitwidth = endIndex - beginIndex;
      return BitVector.of( (value >> (bitwidth - endIndex)) & ((1 <<
                            resultBitwidth) - 1), resultBitwidth);
    }
    ```
The problem is with this leftshift operation. With beginIndex = 0 and
endIndex = 32, the operation is supposed to compute an "all 32 set bits
mask" using the opearation `(1 << 32) - 1 == 0 - 1 == 32'-1`. Instead, it
computes `(1 << 32%32)-1 == 0`.

### Proposed Fix
The proposed fix is about treating all the  machine int (bitvectors)
identically, without any special treatment for bit-vectors of specific sizes.


ehildenb:
Jenkins: test this please

dwightguth (approved):

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
@sdasgup3 if you still want this, please sign the cla.

sdasgup3:
@dwightguth I signed the cla. Thanks for the information.

Issue #254, opened 2019-01-16
Eliminate "Dot" variables from the Haskell Backend Match/Search Output  


traiansf:
The result of Search/Match operations is a substitution.

Th Haskell backend represents this substitution as an ML conjunction (`#And`) of ML equalities (`#Equals`) between a variable and a term.  However, some of the variables are generated when compiling the matching pattern and thus not interesting to the users.  The assignments corresponding to these variables need to be removed from the conjunction.

Related Slack context: around https://runtimeverification.slack.com/archives/CC360GUTG/p1547571211721400

ehildenb:
@traiansf is this still an issue?

traiansf:
I think not. And it seems to old for us to care anyway; moreover, lots of things have changes in the way we pretty print things, so I would say we can close this.

PR #255, opened 2019-01-16
Removing anonymous variables assignments from output substitution

traiansf:
Addressing Issue #254


dwightguth (changes requested):
So, K already has code to prune substitutions of variables that are tagged with the anonymous attribute in the k frontend. You should be adjusting that code if you need to rather than reinventing the wheel.

traiansf:
@dwightguth could you point to a location?  I wouldn't want to convert this into a JavaBackend substitution just to be able to call that pruning code.  The code I wrote would work for any backend outputing substitutions in Kore format.

dwightguth (changes requested):
it looks like you're correct that that was being handled inside the java backend previously. still, I have a couple requests.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -245,7 +248,8 @@ public K search(K initialConfiguration, Optional<Integer> depth, Optional<Intege
                         Pattern kore = textToKore.parsePattern(koreOutputFile);
                         KoreToK koreToK = new KoreToK(idsToLabels, mod.sortAttributesFor(), StringUtil::enquoteKString);
                         K outputK = koreToK.apply(kore);
-                        return outputK;
+                        K trimmedOutput = SubstitutionPruning.pruneSubstitution(generatedVars, outputK);
+                        return trimmedOutput;
```
dwightguth: I would like you to return the original output here but with the "anonymous" attribute added to the correct anonymous variables. And then move the substitution trimming code to KPrint.java and run it only when we are pretty printing.

kore/src/main/scala/org/kframework/compile/SubstitutionPruning.scala
```diff
@@ -0,0 +1,55 @@
+package org.kframework.compile
```
dwightguth: I would like you to rewrite this code in Java. We are trying to reduce the amount of scala code we have in the code base, not increase it.

dwightguth (changes requested):
it looks like you are correct that this was happening inside the java backend previously. Still, I would like you to fix a few things.

kore/src/main/scala/org/kframework/compile/SubstitutionPruning.scala
traiansf: Got it!  It wasn't clear to me so far (since my work was mostly on scala code lately) ...

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
traiansf: makes sense

traiansf:
I've made the following changes:

- translated the removal code to java
- added the call to the removal code in `KRunExecutionMode`
- removed the code from the Java backend that achieved the same thing for the Java backend
- modified java backend serialization of search results to use ML `#Equals` instead of `==K`
- added a krun print option to control whether assignments of anonymous variables are printed

Still to do:
- Update the tests which are failing because of the change from `==K` to `#Equals`

@dwightguth could you please take a look to check whether you are OK with the changes so far?  If so, I'll make another commit with the tests changes.

dwightguth (changes requested):
This looks fine, but term normalization happens during pretty printing in KPrint.java, not in KRunExecutionMode. You should revert those changes and move it to the right place, otherwise people who request the raw term from k will get something different than we want. Look for the code relating to sorting collections and alpha renaming, that's where this goes.

traiansf:
Well, I know that you proposed to put it in `KPrint`, but I don't think that it should be there because  by the time a pattern reaches KPrint it is simply a term, information on where it comes from is lost;  the problem I'm trying to address is specific to the output of a pattern match (which is supposed to have a specific form) and I'm not convinced that it should be applied on all potential K terms.

If you instead insist it should be there, I'm not sure how to ensure that for any K term, given in any form, once can (soundly) eliminate all equalities in which the LHS is an anonymous variable.  Do we even want to eliminate all of them?

dwightguth:
I mean it seems to me that any output term that has the form of a substitution should have this change applied to it.

traiansf:
Jenkins: test this please

traiansf:
Jenkins: test this please

ehildenb:

traiansf:
@dwightguth I've tested it on a separate machine and confirmed that the changes in this PR make krun seemingly loop forever for certain tests, so I'm going to have to debug it further.

traiansf:
@dwightguth can you please re-review?  Apparently the looping behavior was caused by a bug in the hashcode implementation of mini.KApply which was triggering an exponential (in number of KApply nodes) evaluation of hashcodes.

dwightguth:
Looks okay but I still need to review the tests which I can't do on GitHub. I'll try to get to this later today.

dwightguth:
I'm closing this as it has bitrotted. If you still need this functionality, feel free to fix the conflicts and issues and make a new PR.

PR #262, opened 2019-01-22
Build debian package

dwightguth:
Builds by running dpkg-buiildpackage in the root of the repo. Also requires rustup 1.28 and stack to be installed on the system in addition to the explicitly listed build dependencies. The built debian package is about 37 MB and all told when installed with dependencies on a base ubuntu system (via `sudo apt install kframework_5.0.0_amd64.deb`) it adds an additional gigabyte of unpacked data. But the installed tool can use the llvm, ocaml, and java backends with no additional manual setup.

@bmmoore since this is my first debian package if you have any feedback it would be appreciated. I figure we can do versioning in a future PR because we're not going to be using launchpad yet anyway.


ehildenb:

bmmoore (approved):
This looks reasonable as long as it works.
You might consider making the version include a `~` component giving a prerelease version, but maybe not if it's going to be used by the class.

Have you testing installation and removal? It's recommended to do that in a chroot for easy diffing: https://wiki.debian.org/HowToPackageForDebian#All_tests_in_chroot

dwightguth:
I tested installation and removal in a clean vm. I can set up a chroot if it matters though?

PR #266, opened 2019-01-24
Fix illegal reflective access warnings in Java 11

dwightguth:


java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
```diff
@@ -27,10 +27,6 @@
 
     @Before
     public void setUp() {
-        when(globalContext.getDefinition()).thenReturn(definition);
-        when(definition.subsorts()).thenReturn(subsorts);
-        when(subsorts.isSubsortedEq(Sort.LIST, Sort.LIST)).thenReturn(true);
-        when(subsorts.isSubsortedEq(Sort.LIST, Sort.INT)).thenReturn(true);
```
traiansf, line 33: Why is this not needed anymore?

traiansf (approved):

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
dwightguth: It was already not needed, but the update to the mockito version caused it to raise an error about unused stubs.

PR #268, opened 2019-01-25
Build and test debian package

dwightguth:
Ready for review. Builds the debian package for Ubuntu Bionic on every PR and tests the K tutorial on the installed system. Then the master job will upload the package to the github release it creates.


ehildenb (approved):
Looks good to me.

PR #269, opened 2019-01-28
Optimizations to take advantage of Dockerfile cacheing

ehildenb:


ehildenb:
@dwightguth it looks like this Dockerfile based build will shave off 20 minutes or so.

PR #270, opened 2019-01-28
Quote domain value strings in all cases.

virgil-serbanuta:


phillipharr1s (dismissed):

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1341,7 +1341,7 @@ public void apply(KToken k) {
                 convert(k.sort(), false);
                 sb.append("}(");
                 if (module.sortAttributesFor().get(k.sort()).getOrElse(() -> Att.empty()).getOptional("hook").orElse("").equals("STRING.String")) {
-                    sb.append(k.s());
+                    sb.append(StringUtil.enquoteKString(k.s()));
```
dwightguth, line 1344: uh, this is seriously going to break the llvm backend. The kore representation for the String
```
foo
bar
```

should be "foo\nbar", not "\"foo\\nbar\"".

Or am I misunderstanding what this is doing?

virgil-serbanuta: So, right now the k file contains domain values like
```
\dv{??}("foo
bar").
```

I am attempting to convince it to represent them as
```\dv{??}("foo\nbar")
```

dwightguth:
Closing as I believe this was fixed elsewhere.

PR #271, opened 2019-01-28
java-backend: function caching

denis-bogdanas:
Now evaluation of results of `resolveFunction()`
and `resolveFunctionAndAnywhere()` are cached in a map.
The cache is a map from <Term,Constraint> pair to evaluation result.
Yet, if term is ground term, constraint is unnecessary for evaluation,
and is always cached as null. Performance increase is 2x-4x for KEVM.
Enabled by default. User can disable it with `--cache-func false`.

Additional option: `--cache-func-optimized`. Clears cache at beginning of each spec evaluation, to free some memory.

I also fully generalized it. Cache now works properly with impure functions (does not cache them) and in Nailgun multithreaded environment. There's no need to synchronize it, because different Nailgun jobs will have different cache instances.

I was thinking to add a cache cleanup mechanism on low memory but turns out it is not worth the trouble. Memory consumption is low, on the order of ~100M for 2 M entris on larger tests I've run. And if someone gets OutOfMemoryError on very large runs, there is an option to disable the cache.


denis-bogdanas:
@dwightguth Questions:
1. Some time ago K test suite was running the tests from verified-smart-contracts. Does it happen now? If no, how can I test this new option? Are there any existing tests for kprove? In fact option works for `krun` too, so it could be activated for the whole test suite or a part of it.

2. Should I synchronize the cache? I guess it was needed with nailgun. Is it still needed now?
Here is the main data structure of the cache. It's a Table:
https://github.com/kframework/k/blob/40118a2e213e3f849779b1e38e17e1b9cce86ab1/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java#L38

dwightguth (changes requested):
You should use a concurrent map implementation with read write locking to allow concurrent reads, but yes, synchronize. You can test by modifying the Makefile for some of the definitions in the tutorial. Also, please change it to simply call the function every time if it's impure instead of crashing.

denis-bogdanas:
@dwightguth Impure functions are now fully supported, they are not cached.

Apparently all works. I did not make cache maps synchronized, only moved them under the right context, under GnobalContext. As far as I understand the code, they don't have to be synchronized. As long as I don't use any static mutable data structures, different running instances under same Nailgun server will not share any objects, and caches will be separate.

I also made --cache-func enabled by default, for kprove and krun Looks like tests finish ~4x faster compared to other PRs. I'll open a slack discussion about what the default should be.

AFAIK good for review and merge.

denis-bogdanas:
@dwightguth One more thing to do: I'll implement cache emptying policy. Then there will be no danger of OutOfMemoryError and feature could be left enabled by default. Will let you know.

denis-bogdanas:
@dwightguth Ready for review. Please see the updated PR description.

denis-bogdanas:
@dwightguth Can you review pls?

dwightguth (approved):

PR #278, opened 2019-01-30
Fix Stopwatch formatting.

radumereuta:
This was partially changed in #178. Now fixing the other print location.
Before:
```
Cleanup                                                      = 0.001 s
Total                                                        = 6056
```
After:
```
Cleanup                                                      = 0.001 s
Total                                                        = 6.056 s
```
Also, if the total time is more than one minute it should show:
```
Cleanup                                                      = 0.001 s
Total                                                        = 75.056 s (1m 15s)
```


kernel/src/main/java/org/kframework/utils/Stopwatch.java
```diff
@@ -37,8 +37,16 @@ public void printIntermediate(String message) {
 
     public void printTotal(String message) {
         printIntermediate("Cleanup");
-        if (options.verbose)
-            f.format("%-60s = %5d%n", message, lastIntermediate - start);
+        if (options.verbose) {
+            double ms = (lastIntermediate - start) / 1000.;
```
dwightguth: this is named ms but it seems to be a number of seconds, which is wildly misleading.

dwightguth (changes requested):
if you're going to change how times longer than a minute are printed, change it consistently and appropriately in both places with a helper method and using a mechanism that scales up to durations longer than an hour.

radumereuta:
Good points. There are a few automatic convertors, but I couldn't find any that I liked so I've implemented something custom. It should now print:
```
Cleanup                                                      =  7.135s
Total                                                        = 17.135s
```
Or for longer times:
```
 0.000s
 0.100s
10.000s
16m 40s
27h 46m 40s
277h 46m 40s
```

dwightguth (approved):

radumereuta:
@dwightguth do you have any idea why the deb package step failed?

PR #279, opened 2019-01-31
java-backend: performance optimization

denis-bogdanas:
1. Reduced the usage of StringBuilder.toString().
2. Replaced System.out.println with System.out.format() when makes sense.

Questionable effect, but other commits in K-gnosis depend on it. Prerequisite to PR the rest.


denis-bogdanas:
@dwightguth This build failed with an error about Haskell backend. Any idea why? It also fails on my local computer. I only touched Java backend here.

daejunpark (approved):

PR #280, opened 2019-01-31
Adding support for generation definability conditions for partial functions

traiansf:
We need this in the haskell backend to address partiality.

@dwightguth I think this should not impact the llvm backend; if needed the generated axioms are tagged with `ceil{}()` so they can be filtered out.


dwightguth:
Can you add an integration test for the llvm backend to ensure it in fact doesn't break?

ehildenb (changes requested):
The `ceil` attribute needs to be added to the pending documentation document, with an explanation of how it works and some simple examples of what using it looks like/what effect that has on the resulting definition.

PR #283, opened 2019-02-01
allow k tutorial to be built from any path

dwightguth:
Ready for review.


ehildenb (approved):
Looks good if tests pass.

PR #284, opened 2019-02-01
Java Backend logging batch 1

denis-bogdanas:
This is the first batch of logging features from K branch `gnosis`. Options implemented here are not in their final state, they were further generalized in later commits and made semantics-agnostic. Yet it is only feasible to put in the first PR this much. Later improvements will follow.

# Text from important commits

## Commit 1
Java Backend Logging PR1: options --log, --log-basic, --log-stmts-only, --branching-allowed.

Details:
  - logging options: --log, --log-basic, --log-stmts-only. They log a KEVM-specific subset of configuration at certain steps.
  - option --branching-allowed to limit branchings. If allowed number of branchings is exceeded, execution ends.

Additional:
  - added parameter `step` to some methods, for debugging.
  - optimized printing of the term when exception is thrown.
  - other minor logs. All enabled by at least --log-basic.

## Commit 2
Java Backend KEVM customization: Detecting final state, when first item in <k> is #halt and there is nothing left to execute.
If this state is reached and target term is not matched, execution is halted.

Useful in kprove, to prevent further symbolic execution and needless branching that would happen if spec would continue to evaluate past intended final state.

If specification does not use #halt in target term or is non-KEVM, this customization has no effect.
Temporary. Was replaced by more general and semantics-agnostic `--halt-cells`, that is about to be PRed.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -608,6 +608,17 @@ public class SymbolicRewriter {
 
             for (ConstrainedTerm term : queue) {
                 v++;
+                Term generatedTop = term.term();
+                KItem k = getCell((KItem) generatedTop, "<k>");
+                Term kContent = k != null ? ((KList) k.klist()).get(0) : null;
+                BuiltinList kSequence = kContent instanceof BuiltinList ? (BuiltinList) kContent : null;
+                boolean isHalt = kSequence != null
+                                 //case <k> has multiple items: first must be #halt, second - not #execute
+                                 ? kSequence.size() == 2 && kSequence.get(0) instanceof KItem
+                                         && kSequence.get(0).toString().equals("#halt_EVM(.KList)")
+                                         && !kSequence.get(1).toString().equals("#execute_EVM(.KList)")
+                                 //case <k> has only one item: it must be #halt
+                                 : kContent != null && kContent.toString().equals("#halt_EVM(.KList)");
                 boolean oldLog = globalOptions.log;
 
                 if (term.implies(targetTerm)) {
@@ -617,6 +628,18 @@ public class SymbolicRewriter {
                         System.out.println("\n============\nStep " + step + ": eliminated!\n============\n");
                         logStep(step, v, targetCallData, term, true);
                     }
+                    continue;
+                } else {
+                    logStep(step, v, targetCallData, term, step == 1 && globalOptions.logBasic);
+                }
+
+                //stopping at halt
+                if (isHalt) {
+                    if (!globalOptions.log) {
+                        logStep(step, v, targetCallData, term, globalOptions.logBasic);
+                    }
+                    System.out.println("Halt! Terminating branch.");
+                    proofResults.add(term);
                     continue;
                 } else {
                     logStep(step, v, targetCallData, term, step == 1 && globalOptions.logBasic);
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -12,6 +12,7 @@ import org.kframework.backend.java.builtins.BoolToken;
 import org.kframework.backend.java.builtins.FreshOperations;
 import org.kframework.backend.java.compile.KOREtoBackendKIL;
 import org.kframework.backend.java.kil.BuiltinList;
+import org.kframework.backend.java.kil.BuiltinMap;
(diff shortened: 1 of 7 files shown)


ehildenb:
@denis-bogdanas this puts a ton of KEVM specific stuff into the Java backend, I don't think that's appropriate at all.

Also, is there a chance we can re-use the existing `--state-log` functionality. It looks like you're mostly instrumenting the `SymbolicRewriter`, and the `StateLog` has already instrumented that code. You can add logging events if you like.

Anything KEVM specific should be done as a post-processing step with external scripts over the `StateLog` I think, instead of included in the Java backend.

denis-bogdanas:
I later generalized these logs, in branch `gnosis` they are no longer KEVM-specific. The list of cells to be logged is given as command-line option. But I can only put so much in a PR. Generating same logs from StateLog is infeasible as amount of effort. I tried this long ago and gave up. We don't have bandwidth to make big changes to a backend that will soon be obsolete.

dwightguth:
Yeah this definitely needs to be squashed into the later changes. Grigore is starting to teach for the semester and it's simply a non-starter to be distributing kevm-specific flags to krun to users.

denis-bogdanas:
@dwightguth I see. Looks like you didn't line flag descriptions:) In the latest version I removed the piece `KEVM specific`. I can try to squash with later changes. Problem is that later commits are intertwined with features unrelated to logging, it will be a pain to align them. All the flags in this PR are experimental; students should have no business to see them.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -584,14 +588,61 @@ private K disjunctResults(List<K> results) {
...
+                Term kContent = k != null ? ((KList) k.klist()).get(0) : null;
+                BuiltinList kSequence = kContent instanceof BuiltinList ? (BuiltinList) kContent : null;
+                boolean isHalt = kSequence != null
+                                 //case <k> has multiple items: first must be #halt, second - not #execute
+                                 ? kSequence.size() == 2 && kSequence.get(0) instanceof KItem
+                                         && kSequence.get(0).toString().equals("#halt_EVM(.KList)")
+                                         && !kSequence.get(1).toString().equals("#execute_EVM(.KList)")
+                                 //case <k> has only one item: it must be #halt
+                                 : kContent != null && kContent.toString().equals("#halt_EVM(.KList)");
+                boolean oldLog = global.javaExecutionOptions.log;
```
ehildenb, line 622: This is st ill KEVM specific.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -676,6 +784,114 @@ private void printSummaryBox(Rule rule, List<ConstrainedTerm> proofResults, int
...
+        BuiltinList kSequence = kContent instanceof BuiltinList ? (BuiltinList) kContent : null;
+
+        KItem output = getCell(top, "<output>");
+        KItem statusCode = getCell(top, "<statusCode>");
+        KItem localMem = getCell(top, "<localMem>");
+        KItem pc = getCell(top, "<pc>");
+        KItem gas = getCell(top, "<gas>");
+        KItem wordStack = getCell(top, "<wordStack>");
+        KItem callData = getCell(top, "<callData>");
+        KItem accounts = getCell(top, "<accounts>");
```
ehildenb, line 803: KEVM specific.

ehildenb (changes requested):
There are still KEVM specific stuff, it's not just about the command-line options there.

If we make the Java backend able to do proofs over KEVM by introducing KEVM specific stuffs, that will make switching over to the Haskell backend that much harder. Please implement this in a generic way (passing in the configuration specific parameters either on the command line or via a config file) instead of hard-coding it into the backend.

ehildenb:
I'm fairly certain all of the logging functionality here can be done as post-processing on the existing `StateLog` functionality (which also logs things like Z3 queries and results in addition). That would give a much less language-specific logging structure for the Haskell backend to re-implement. Since that outputs JSON files with the terms, the post-processing can be done in whatever language.

The actual logging is covered by `global.stateLog`. The extra logic checks (1) is it branching, or (2) is it terminating, which can be done as post-processing over the `StateLog`. Beyond that, everything else is post-processing the KEVM terms to only log the relevant parts, which I don't think belongs in the backend at all.

I'm willing to help develop a Python or Haskell library for post-processing the terms output by StateLog. I already helped develop the one that KLab uses (in javascript). The JSON format it outputs is fairly simple to walk and extract the needed parts from.

denis-bogdanas:
@ehildenb The semantics-independent logging and halt detection is already implemented in K-gnosis, I just couldn't separate it from the other commits and wanted to postpone to another PR. Will put all in one PR as you guys requested.

PR #288, opened 2019-02-08
KProve architecture extension: made pretty printing available to...

denis-bogdanas:
... implementations of Rewriter. Used in later PRs.

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/PrettyPrinter.java
+++ b/kernel/src/main/java/org/kframework/unparser/PrettyPrinter.java
@@ -1,0 +1,25 @@
+package org.kframework.unparser;
+
+import org.kframework.definition.Definition;
+import org.kframework.definition.Module;
+import org.kframework.kore.K;
+
+/**
+ * @author Denis Bogdanas
+ * Created on 07-Feb-19.
+ */
+public class PrettyPrinter {
+    public final KPrint kprint;
+    public final Definition def;
+    public final Module module;
+
+    public PrettyPrinter(KPrint kprint, Definition def, Module module) {
+        this.kprint = kprint;
+        this.def = def;
+        this.module = module;
+    }
+
+    public void prettyPrint(K target) {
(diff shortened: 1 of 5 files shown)


ehildenb:
@denis-bogdanas it's hard to know if an architectural change makes sense without seeing how it's used.

Would it be possible to just extend `KPrint.java` instead of making a new class? That was intended to be where all pretty and non-pretty printing functionality goes.

denis-bogdanas:
@ehildenb If you look at PrettyPrinter, you'll see that it incapsulates 3 objects. One of them is KPrint, the other 2 judging by your design choces don't belong to KPrint. Hence the need to create a new class.

kernel/src/main/java/org/kframework/unparser/PrettyPrinter.java
```diff
@@ -0,0 +1,25 @@
+package org.kframework.unparser;
```
ehildenb, line 2: copyright on first line

kernel/src/main/java/org/kframework/unparser/WantsPrettyPrinter.java
```diff
@@ -0,0 +1,12 @@
+package org.kframework.unparser;
```
ehildenb, line 2: copyright on first line

ehildenb (changes requested):
some nits

kernel/src/main/java/org/kframework/unparser/WantsPrettyPrinter.java
denis-bogdanas: I remember there was a check for copyright at build time. Why it was not triggered?

ehildenb (approved):

dwightguth:
I have reverted the merge of this PR because I needed to request changes to it but it was merged without waiting for my review. In the future please do not merge changes into this repo without my approval. If you remake the PR, I will add the comments that I wanted to make to the code.

PR #291, opened 2019-02-08
Java Backend Logging: First batch of logging features

denis-bogdanas:
## Main additions
- option --log to print a subset of configuration specified in `--log-cells` at each step.
- option --log-basic - same as --log but only at certain important steps such as branchings, application of spec rule.
  Plus enables multiple other minor logs.
- option --log-cells. Generalized logging. Added ability to specify the list of cells to log, and which of these cells to be pretty-printed.
- option --branching-allowed to limit branchings. If allowed number of branchings is exceeded, execution ends.
- option --log-rule, prints rule source for all invoked rules of in all contexts. Both kprove and krun.
- final implication debug: Logging the non-matching sub-terms when final implication does not match, if at least --log-basic is provided.

## Minor additions
- added parameter `step` to some methods, for debugging.
- Custom formatting when entire term is printed, without using the slow pretty printing.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaExecutionOptions.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaExecutionOptions.java
@@ -58,12 +58,12 @@ public final class JavaExecutionOptions {
     @Parameter(names="--branching-allowed", arity=1, description="Number of branching events allowed before a forcible stop.")
     public int branchingAllowed = Integer.MAX_VALUE;
 
-    @Parameter(names="--log", description="Log every step. KEVM only.")
+    @Parameter(names="--log", description="Log every step.")
     public boolean log = false;
 
     @Parameter(names="--log-basic",
-            description="Log most basic information: summary of initial step, final steps and final implications." +
-                    " All custom logging only works for KEVM-based specs.")
+            description="Log most basic information: summary of initial step, final steps and final implications, " +
+                    "branching points.")
     public boolean logBasic = false;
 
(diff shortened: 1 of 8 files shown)


denis-bogdanas:
@ehildenb ready for review.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaExecutionOptions.java
```diff
@@ -54,6 +55,37 @@
             description="Clear function cache after initialization phase. Frees some memory. Use IN ADDITION to --cache-func")
     public boolean cacheFunctionsOptimized = false;
 
+    @Parameter(names="--branching-allowed", arity=1, description="Number of branching events allowed before a forcible stop.")
+    public int branchingAllowed = Integer.MAX_VALUE;
+
+    @Parameter(names="--log", description="Log every step. KEVM only.")
```
ehildenb: This isn't accurate anymore, AFAICT.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaExecutionOptions.java
```diff
@@ -54,6 +55,37 @@
...
+    public boolean log = false;
+
+    @Parameter(names="--log-basic",
+            description="Log most basic information: summary of initial step, final steps and final implications." +
+                    " All custom logging only works for KEVM-based specs.")
```
ehildenb: Same here.

ehildenb (changes requested):

denis-bogdanas:
@ehildenb done.

ehildenb (approved):

PR #292, opened 2019-02-08
KProve architecture extension: made pretty printing available to Rewriter

denis-bogdanas:
Kernel: refactoring: rewriterGenerator now takes as argument kore Definition instead of Module.

Java Backend: building the class PrettyPrinter that incapsulates all components necessary to pretty print a K term. Used in logging PR.

### DIFF
--- a/kernel/src/main/java/org/kframework/kprove/KProve.java
+++ b/kernel/src/main/java/org/kframework/kprove/KProve.java
@@ -42,9 +42,9 @@ public class KProve {
         this.kprint = kprint;
     }
 
-    public int run(KProveOptions options, CompiledDefinition compiledDefinition, Backend backend, Function<Module, Rewriter> rewriterGenerator) {
+    public int run(KProveOptions options, CompiledDefinition compiledDefinition, Backend backend, Function<Definition, Rewriter> rewriterGenerator) {
         Tuple2<Definition, Module> compiled = getProofDefinition(options.specFile(files), options.defModule, options.specModule, compiledDefinition, backend, files, kem, sw);
-        Rewriter rewriter = rewriterGenerator.apply(compiled._1().mainModule());
+        Rewriter rewriter = rewriterGenerator.apply(compiled._1());
         Module specModule = compiled._2();
 
         K results = rewriter.prove(specModule);
--- a/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackendKModule.java
+++ b/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackendKModule.java
@@ -7,7 +7,7 @@ import com.google.inject.Module;
 import com.google.inject.TypeLiteral;
 import com.google.inject.multibindings.MapBinder;
 import org.apache.commons.lang3.tuple.Pair;
-import org.kframework.compile.Backend;
+import org.kframework.definition.Definition;
 import org.kframework.main.AbstractKModule;
 import org.kframework.rewriter.Rewriter;
 
@@ -56,8 +56,8 @@ public class HaskellBackendKModule extends AbstractKModule {
     private void installHaskellRewriter(Binder binder) {
         bindOptions(HaskellBackendKModule.this::krunOptions, binder);
 
-        MapBinder<String, Function<org.kframework.definition.Module, Rewriter>> rewriterBinder = MapBinder.newMapBinder(
-                binder, TypeLiteral.get(String.class), new TypeLiteral<Function<org.kframework.definition.Module, Rewriter>>() {
+        MapBinder<String, Function<Definition, Rewriter>> rewriterBinder = MapBinder.newMapBinder(
+                binder, TypeLiteral.get(String.class), new TypeLiteral<Function<Definition, Rewriter>>() {
                 });
         rewriterBinder.addBinding("haskell").to(HaskellRewriter.class);
     }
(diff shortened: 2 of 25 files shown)


kernel/src/main/java/org/kframework/unparser/WantsPrettyPrinter.java
```diff
@@ -0,0 +1,13 @@
...
+package org.kframework.unparser;
+
+/**
+ * To be implemented by instances of {@code Rewriter} that need an object of type PrettyPrinter.
+ *
+ * @author Denis Bogdanas
+ * Created on 08-Feb-19.
+ * @see org.kframework.rewriter.Rewriter
+ */
+public interface WantsPrettyPrinter {
```
dwightguth: I'm reasonably certain you don't need this interface You can just construct an instance of PrettyPrinter inside SymbolicRewriterGlue.

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -44,8 +46,14 @@ public KProve(KExceptionManager kem, Stopwatch sw, FileUtil files, KPrint kprint
 
     public int run(KProveOptions options, CompiledDefinition compiledDefinition, Backend backend, Function<Module, Rewriter> rewriterGenerator) {
         Tuple2<Definition, Module> compiled = getProofDefinition(options.specFile(files), options.defModule, options.specModule, compiledDefinition, backend, files, kem, sw);
-        Rewriter rewriter = rewriterGenerator.apply(compiled._1().mainModule());
+        Definition def = compiled._1();
+        Rewriter rewriter = rewriterGenerator.apply(def.mainModule());
         Module specModule = compiled._2();
+        Module languageModule = def.getModule("LANGUAGE-PARSING").get();
+        PrettyPrinter prettyPrinter = new PrettyPrinter(kprint, def, languageModule);
+        if (rewriter instanceof WantsPrettyPrinter) {
```
dwightguth: this type of mutable state is not very nice and is error-prone. You should be able to rewrite this so that you don't need to call any setter function like this.

dwightguth (changes requested):

kernel/src/main/java/org/kframework/unparser/WantsPrettyPrinter.java
denis-bogdanas: I scratched my head enough to come with least invasive solution. I cannot construct PrettyPrinter in SymbolicRewriterGlue, because I don't have access to Definition and Module there. Alternative would be to have a `WantsDefinitionAndModule` instead. Doesn't look better.

kernel/src/main/java/org/kframework/kprove/KProve.java
denis-bogdanas: I'm very well aware of this rule. The clean alternative is to pass `PrettyPrinter` to rewriteGenerator here: https://github.com/kframework/k/blob/8c813df953f27e1b13d57383d1920b4df52946c8/kernel/src/main/java/org/kframework/kprove/KProve.java#L50

or to Rewriter here: https://github.com/kframework/k/blob/8c813df953f27e1b13d57383d1920b4df52946c8/kernel/src/main/java/org/kframework/kprove/KProve.java#L58

Both would affect the interface of all backends and would be considerably more invasive.
This PR is really a compromise with design flaws that already exist in kernel module.

kernel/src/main/java/org/kframework/unparser/WantsPrettyPrinter.java
dwightguth: I'm super confused. A SymbolicRewriterGlue is already specific to a specific definition and module, it would be utterly trivial to ensure you had access to the appropriate information in that class, assuming it's not already there.

denis-bogdanas: Please look at my reply below. Looking at the whole code in Intellij also would help. To pass the right information to SymbolicRewriterGlue, one way or another, I have to alter the interfaces that create it. This will affect all backends.

kernel/src/main/java/org/kframework/kprove/KProve.java
dwightguth: Okay here's what we're going to do. We're going to change the signature of the K frontend so that a Rewriter is generated from a Definition instead of a Module. I've checked, and as far as I can tell in all cases we just use the main module of a definition anyway, so it should be an easy change to make, if extensive.

denis-bogdanas: makes sense. thx!

denis-bogdanas: @dwightguth After a more detailed look at the code, your proposal doesn't seem possible.

RewriterGenerator is shared between kprove and krun, but takes different module as argument:
KProve: https://github.com/kframework/k/blob/cc72c39b37aa4d07e8d7da4c34544d488ed24a3e/kernel/src/main/java/org/kframework/kprove/KProve.java#L47

KRun: https://github.com/kframework/k/blob/91a69ed56a92897677d16bbdb2065841a4446125/kernel/src/main/java/org/kframework/krun/modes/KRunExecutionMode.java#L43

So it's not possible to replace module with definition because you won't know how to get the right module in the right context.

The way I'd do it is to modify the interface of RewriterGenerator to take 2 arguments - module and PrettyPrinter. Then I'd keep `PrettyPrinter` because it nicely encapsulates all that's needed for pretty printing, but remove `WantsPrettyPrinter`.

Does this look good to you?

dwightguth: If you look at the implementation of the function, executionModule() is just kompiledDefinition.mainModule(). So it's still the main module of a definition in both cases, which means your concern doesn't really apply. It should still be doable to pass in a Definition unless you find another call site somewhere else that does something other than one of these two things.

denis-bogdanas:
@dwightguth Ready for review.

dwightguth (approved):
Looks much better, thank you!

PR #293, opened 2019-02-09
Java Backend: Z3 query build optimization

denis-bogdanas:
Java Backend: Z3 query build optimization:
removing substitution from implication LHS, because it's not used in query build process anyway. Massively improves cache hit rate on some specs and improves performance.

## Proof that HS substitution is not used

1. argument `allowNewVars` is true here:
https://github.com/kframework/k/blob/1768d7f317dd00bbd149995fe62203df68ba869d/java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java#L199

2. When `allowNewVars` is true, substitution is not used when translating ConjuncitveFormula:

https://github.com/kframework/k/blob/1768d7f317dd00bbd149995fe62203df68ba869d/java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java#L429-L433

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
@@ -871,8 +871,12 @@ public class ConjunctiveFormula extends Term implements CollectionInternalRepres
                 continue;
             }
 
-            global.stateLog.log(StateLog.LogEvent.IMPLICATION, left, right);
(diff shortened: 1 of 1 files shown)


denis-bogdanas:
@daejunpark If you remember we discussed on slack about this optimization. It reduced the number of Z3 queries and fixed last performance issue we had with Gnosis. Anyway, the proof of soundness (from programmer perspective) is in description.

daejunpark (approved):

PR #296, opened 2019-02-12
demonstrate bug in parsing definition on haskell backend

ehildenb:
Note that the test passes, because it arbitrarily picks the super-sort (which is undesired), and then executes fine.


dwightguth:
I'm going to close this, but I will reopen a PR that includes this test once I have the fix.

PR #300, opened 2019-02-17
java-backend: Cache ConjunctiveFormula.simplify()

denis-bogdanas:
Caching results of `ConjunctiveFormula.simplify()`. Disabled by default, enabled with `--cache-formulas`. 

Performance increase in current VSC Jenkins suite is only 5%, but maybe it's more noticeable on certain rules or it was bigger in the past. Might still be useful in future.


ehildenb:
@denis-bogdanas roughly how much time does this save? Out of curiosity?

denis-bogdanas:
@ehildenb See the description.

ehildenb (approved):
Looks good.

PR #302, opened 2019-02-18
Java Backend: toString() cache

denis-bogdanas:
Java Backend: Caching toString() for KItem. Speeds up all logging. Enabled with `--cache-tostring`.

Performance increase was ~25% some time ago before I optimized conjunctive formulas. At that time log size was several times bigger. Didn't use this option recently.


denis-bogdanas:
@ehildenb Can you review pls?

daejunpark (approved):

PR #303, opened 2019-02-19
Mainline klab

mhhf:
Changes are necessary for klab to work

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
@@ -69,13 +69,13 @@ public class GlobalContext implements Serializable {
         this.hookProvider = hookProvider;
         this.files = files;
         this.equalityOps = new EqualityOperations(() -> def);
-        this.stateLog = new StateLog(javaExecutionOptions, files);
-        this.constraintOps = new SMTOperations(() -> def, smtOptions,
-                new Z3Wrapper(smtOptions, kem, javaExecutionOptions, files, stateLog), kem, javaExecutionOptions);
+        prettyPrinter = new PrettyPrinter(kprint, coreDefinition);
+        this.stateLog = new StateLog(javaExecutionOptions, files, prettyPrinter);
+        this.constraintOps = new SMTOperations(() -> def, smtOptions, new Z3Wrapper(smtOptions, kem, javaExecutionOptions, files, stateLog), kem, javaExecutionOptions);
         this.kItemOps = new KItemOperations(stage, javaExecutionOptions.deterministicFunctions, kem, this::builtins, globalOptions);
         this.stage = stage;
         this.profiler = profiler;
-        prettyPrinter = new PrettyPrinter(kprint, coreDefinition);
+
     }
 
     private transient BuiltinFunction builtinFunction;
--- a/java-backend/src/main/java/org/kframework/backend/java/util/PrettyPrinter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/util/PrettyPrinter.java
@@ -30,4 +30,8 @@ public class PrettyPrinter {
     public void prettyPrint(K target, OutputStream out) {
         kprint.prettyPrint(def, module, output -> kprint.outputFile(output, out), target);
     }
+
+    public byte[] prettyPrintBytes(K target) {
+        return kprint.prettyPrint(def, module, target);
+    }
 }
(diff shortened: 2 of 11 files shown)


dwightguth:
I'll review as soon as you resolve the conflicts.

mhhf:
There are still some bugs which @ehildenb wanted to have a look at

ehildenb:
@mhhf can you try out this version? I reverted the variable printing change, since I don't think that one will work out without a bit more care (or at least updating test output).

MrChico:
@ehildenb this flatten seems to trigger infinite recursion leading to a stack overflow;
```
at scala.collection.AbstractMap.apply(Map.scala:59)
	at org.kframework.unparser.KPrint.flattenTerm(KPrint.java:224)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:195)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:189)
	at org.kframework.kore.KTransformer.apply(transformers.scala:11)
	at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
	at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:195)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:189)
	at org.kframework.kore.KTransformer.apply(transformers.scala:11)
	at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
	at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:195)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:189)
	at org.kframework.kore.KTransformer.apply(transformers.scala:11)
	at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
	at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:195)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:189)
	at org.kframework.kore.KTransformer.apply(transformers.scala:11)
	at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
	at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:195)
	at org.kframework.unparser.KPrint$3.apply(KPrint.java:189)
```
Also, I don't understand your skepsis towards the variable name printing. It's a lot easier to make sense of the output if the variable names are the same as was inputted in the proof claim instead of `Vn`. Many features of `K` depends on this. I don't push access to `mhhf/k` but I incorporated these changes here https://github.com/dapphub/k/tree/mainline_klab

mhhf:
@MrChico I gave you access to mhhf/k
@ehildenb the changes in KPrint and ToJson are as martin said essential for debugging. If they don't get included in the k#master we still will need to maintain a k fork which is dependant by klab.
The crucial information for us here is the originalName of the variable which is not provided in the original version, rather its a localized Variable name making things hard to debug. Logically speaking the renamed name should be equal to the original name so I don't see a strong reason against this change.  
Please help me understand where the problems with those changes are coming from.

ehildenb:
Changing the way variables is printed will change a lot of output. I agree though that this way of printing variables is much more informative, and a change for the better. Perhaps it is worth the change in output. Another problem I can see though, is that different backends will have different variable name generation schemes, so it's important to canonicalize the variable names before printing if we want to do automated comparisons of the outputs of the backends.

I wonder if we can add a flag `--original-variable-names` which turns on your behavior, and by default have the old behavior?

By the way, if you make this change, you'll need to make the same change to `ToBinary.java` as well.

mhhf:
A flag like `--original-variable-names` would solve our needs!

denis-bogdanas:
I had the same problem with variable names in branch `gnosis`. The variable names I got there are of the form `<originalName>_<generatedId>`. Either having original names by default or when an extra option is given would be fine.

ehildenb:
See https://github.com/kframework/k/pull/308.

MrChico:
@ehildenb 
It looks like the `output-flatten` flag isn't working quite as expected. [Here is a gist](https://gist.github.com/MrChico/7b00570f2c2bb4fb2464828ccff4ddef) of a cell that was supposed to be flattened (passed with `--output-flatten "_Map_ #And" \`)

Another thing that I noticed is that with the new `originalName` feature, `tokenized` cells are printed with their variable indices instead of the original, human readable name. Perhaps we can make it so that if the `no-alpha-renaming` flag is passed, `tokenize` prints the `originalName` instead?

dwightguth (approved):

PR #304, opened 2019-02-19
Java Backend: option --format-failures

denis-bogdanas:
Added a custom formatting of failure final states. Enabled with `--format-failures`. This required a small extension to `KPrint` and some refactorings in `SymbolicRewriter` to avoid duplicates. See commit messages for details.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -36,6 +36,7 @@ import org.kframework.kore.KORE;
 import org.kframework.rewriter.SearchType;
 import org.kframework.utils.errorsystem.KExceptionManager;
 
+import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -740,10 +737,28 @@ private K disjunctResults(List<K> results) {
...
+                printTermAndConstraint(term);
+            }
+            tweakedProofResults = ImmutableList.of(new ConstrainedTerm(BoolToken.FALSE, initialTerm.termContext()));
+        }
+
         if (global.globalOptions.verbose) {
             printSummaryBox(rule, proofResults, successPaths, step);
         }
-        return proofResults;
+        return tweakedProofResults;
```
ehildenb, line 751: I don't like that `tweakedProofResults` is returned instead of the actual `proofResults`. Can you do the tweaking as a post-processing step at the time you print them out instead of here?

ehildenb (changes requested):
Small change, would like to not change what `disjunctResults` returns, but instead just change how it's displayed.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: I don't think so. Custom printing is specific to java backend. All I want is to disable default result printing and use custom printing instead, when this option is given. Also I don't want the post-processing that is done on proof results - I want to display final states fully, one by one.

Other commits use this custom output format in different ways, like displaying success final states. So this customization is one way or another inevitable.

ehildenb (approved):
I thought this was changing the results of `disjunctResults`, no the overall `proveRule` method. But since it's the top-level `proveRule`, and it's predicated by an optional flag, I think changing the behavior of the method is OK.

PR #305, opened 2019-02-20
Send sentence labels to Kore

ttuegel:
Sentence labels are given to the `label` attribute in generated Kore definitions. Sentence labels must be unique in each module. The user-defined label is prefixed with the module name.

We discussed using the `name` Kore attribute instead of `label`, but I think consistency is important to avoid confusion, and the frontend code uses the term "label".


dwightguth (approved):

dwightguth:
looks good. We don't really have the infrastructure in place to write a test for this right now, but can you at least kompile some definition with a label and inspect the -kompiled/definition.kore file to make sure the attribute is appearing appropriately before it's merged?

ttuegel:
> can you at least kompile some definition with a label and inspect the -kompiled/definition.kore file to make sure the attribute is appearing appropriately before it's merged?

Yes, I have checked: the labels appear in the Kore definition, and they are escaped correctly.

PR #306, opened 2019-02-20
Various improvements to K frontend

dwightguth:
At a high level:

* refactor the parser generator so that modules other than ID can have separate syntax when parsing programs (needed for upcoming substitution PR)
* fixes a bug where macros were not expanded if they appear in the initial configuration
* adds support for `#Or` in the lhs of rules which is interpreted as an or-pattern a la OCAML
* adds new operators :=K and :/=K whose mathematical meaning is \in(...) and \not(\in(...))) and which is compiled efficiently by the frontend down to pattern matching on a fresh function symbol.
* fixes a bug in #266 which caused the incorrect guice jar to be pulled in as a dependency of some of our other guice dependencies.


k-distribution/include/builtin/kast.k
```diff
@@ -73,18 +73,16 @@ module KSEQ-SYMBOLIC
...
-                    > #Pattern "#Equals" #Pattern [klabel(#Equals), symbol, mlOp, format(%i%1%d%n%2%n%i%3%d)]
-                    > #Pattern "#And" #Pattern [klabel(#And), symbol, assoc, comm, unit(#True), mlOp, format(%i%1%d%n%2%n%i%3%d)]
-                    > #Pattern "#Or" #Pattern [klabel(#Or), symbol, assoc, comm, unit(#False), mlOp, format(%i%1%d%n%2%n%i%3%d)]
-                    > #Pattern "#Implies" #Pattern [klabel(#Implies), symbol, mlOp, format(%i%1%d%n%2%n%i%3%d)]
-                    > "#Exists" #KVariable "." #Pattern [klabel(#Exists), symbol, mlOp]
-                    | "#Forall" #KVariable "." #Pattern [klabel(#Forall), symbol, mlOp]
+  syntax KBott ::= "#True" [klabel(#True), symbol]
+  syntax KBott ::= "#False" [klabel(#False), symbol]
+  syntax KBott ::= "#Not" "(" K ")" [klabel(#Not), symbol, mlOp, poly(0, 1)]
+                 | "#Ceil" "(" K ")" [klabel(#Ceil), symbol, mlOp, poly(0, 1)]
```
traiansf: Does this mean that I'll be able to use (all) the ML connectives when writing K definitions?

traiansf (approved):
Looks good to me.

ehildenb:
I think we need to have a discussion about the surface syntax of ML that's exposed to users. @yzhang90 actually needs some other constructs for his bounded model checking, which probably we need an `#Always` and an `#Eventually`. But I would prefer if surface K exposed prettier versions of the operators, eg. `_/\_` for ML AND, and `_\/_` for ML OR, or `[]_` for Always (so that we're consistent with CTL/LTL).

traiansf:
Always and eventually are not even present in KORE :-)
But I fully agree with you;  I think that claims in specs, for example, should be written using the appropriate modal operators instead of the plain rewrite.

ehildenb (approved):
Should be OK, since as we discussed this is about just lifting ML operators to surface syntax, we can add versions of the operators which do some clever/userful things down the road.

kernel/src/main/java/org/kframework/compile/checks/ComputeUnboundVariables.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2019 K Team. All Rights Reserved.
+// Copyright (c 2017-2019 K Team. All Rights Reserved.
```
radumereuta: Is that right?

dwightguth:
This should hopefully finally be passing tests. I ended up having to take a dive into unparsing in order to get the tutorial working again with my changes, so @ehildenb and @traiansf should probably re-review.

traiansf (approved):

PR #308, opened 2019-02-20
Add option to turn off alpha renaming

ehildenb:
This moves alpha renaming into part of the abstraction pass of `prettyPrint`, which means that direct calls to `serialize` and `unparseTerm` will not have the alpha renaming happen anymore. I only found one direct call to `unparseTerm`, in the file `kernel/src/test/java/org/kframework/unparser/AddBracketsTest.java`, so I'm not sure if that one is affected.


ehildenb:
@mhhf and @mrchico, can you rebase on this PR, and try it out? I think it should affect both the ToJson serializer and the final output, since now the variable `_.name()` field won't be replaced with `V*` identifiers. Let me know if it doesn't do what you expect.

dwightguth:
If you're going to put the call to alphaRename inside abstractTerm you should do the same with the collection sorting code.

MrChico:
We still need something like this in order to actually print the original variable names https://github.com/dapphub/k/commit/e7a2dbd3b91a1eb873d2987a5a8849c946ee2f25

dwightguth (approved):

PR #310, opened 2019-02-21
Add smtlib attribute for Bool and Int sorts

phillipharr1s:


denis-bogdanas:
This could have unexpected implications for verification. Can you explain why is it necessary?

dwightguth:
I can't see how it would affect verification at all, since verification is not currently relying on the smtlib attribute of sort declarations. That said, it's unreasonable to expect an smtlib attribute on every single sort that is declared in the solver, so I think it's probably better if the prover, in the absence of an explicit attribute, defaults to the name of the sort. Which would make this change unnecessary. So I'm not sure I agree with it either, but for very different reasons.

denis-bogdanas:
I agree.

phillipharr1s:
I thought we had discussed this before, but maybe I'm misremembering. Anyway here's my reasoning for it.

Currently the way hooks work is you'll declare a sort to be hooked (for example Bool is hooked to BOOL.Bool) and then also declare the symbols associated with that sort are hooked (for example not is hooked to BOOL.not).

With respect to the smtlib attribute, we have to explicitly declare the symbols but not the sort. This is an odd asymmetry. I think it's more consistent and predictable to require sorts to be declared with the smtlib attribute if they are to be involved in SMT translation.

phillipharr1s:
I could certainly do it another way (hardcore int and bool, or blindly translate the sort name to the same name in the SMT language) but both strike me as rather hack-ish.

dwightguth:
I have no objection to an smtlib attribute on the sort declaration in kore, but we should automatically add it in the frontend if it's required and the user doesn't specify it explicitly.

We can't automatically add a hook attribute to a sort declaration because we can't know what its default value should be. If I have the productions
```
syntax Bool ::= Bool "andBool" Bool [hook (BOOL.and)]
syntax Bool ::= Int "<=Int" Int [hook(INT.le)]
```

What should the hook attribute be on `syntax Bool`? Any possible automatic choice we could make is not going to work universally. It's better to force the user to explicitly declare it.  However, there is a sensible value for the smtlib attribute of the Bool sort, namely, "Bool". If the user has a large specification full of many symbols and sorts which are not hooked but which need to be represented in smt, it vastly increases verbosity if we are forced to explicitly declare every single sort used in the spec. Indeed, I would go one step further and say that unless the user explicitly specifies an smtlib attribute on a symbol, we should simply use the name of the symbol. There may be downsides to this in the size of the queries sent to smt though so that's something we can discuss, but I don't see how the same concern applies to sorts. All we are doing is forcing the user to declare the name of their sort three times instead of once.

phillipharr1s:
The situation in `domains.k` currently looks like this:

>   syntax Bool [hook(BOOL.Bool)]
>   syntax Bool ::= "notBool" Bool          [smtlib(not), hook(BOOL.not)]

Where I have removed some other attributes for readability. 

We currently declare `hook`  for both the sort, and the associated symbols. But we only declare `smtlib` for the symbols. It's true that the string "Bool" happens to be the name of booleans in the smt2lib standard, and same for "Int," but that's just a coincidence, it could easily have been "bool" or "BOOL" etc. I think it's more principled to specify the SMT version of the sort explicitly.

Exactly how many sorts actually get translated to SMT in practice, besides Int and Bool? My understanding was that, even though SMT technically has facilities for reasoning about custom algebraic data types, it's not actually used for user-defined sorts in K.

dwightguth:
Your understanding is definitely incorrect. Reasoning in smt about symbols that are not builtins is a routine part of kevm proofs. If we force it to be explicitly declared, every single one of the sorts involved in such symbols will have to be named in the code an extra two times for no appreciable benefit. You say that it is a coincidence that Bool and Int just happen to be the same name in K and SMT and I agree, but that's why we should /allow/ the user to specify an attribute, not /force/ them to. In most cases in user code they will want to use the same name as the sort, and we should not spite the common case (user semantics) to make the k prelude cleaner.

phillipharr1s:
Wait, really? Can you show me an example of this feature?

For now I can just get around this by hardcoding Int and Bool, but I wasn't aware of this other use of SMT.

phillipharr1s:
So ultimately, the one extant example of a non-builtin sort that would need to have the `smtlib` attribute added by a user is `WordStack`. However there could be more examples in the future, so I'm going to just hardcode the case for Int and Bool for now.

PR #311, opened 2019-02-21
Java Backend debug: option `--boundary-cells`

denis-bogdanas:
Specifies the list of cells, for which, once they match the target, execution terminates. For KEVM, recommended value is "k,pc", e.g. cells commonly used as boundaries of evaluation.

Halt execution if boundary cells of current term match the same cells of target term, but entire current term does not match the target term.


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -883,6 +902,21 @@ private KItem getCell(KItem root, String label) {
...
 
+    public ConstrainedTerm buildHaltTerm(ConstrainedTerm term) {
+        KItem result;
+        try {
+            List<Term> cells = global.javaExecutionOptions.haltCells.stream()
+                    .map(name -> getCell((KItem) term.term(), String.format("<%s>", name)))
+                    .collect(Collectors.toList());
+            result = KItem
+                    .of(KLabelConstant.of(KLabels.HALT_CELLS_CELL, global.getDefinition()), KList.concatenate(cells),
+                            global);
```
ehildenb: Is this extra `HALT_CELLS_CELL` cell added as a "dummy" cell to append the other cells to? I don't understand the purpose.

dwightguth (changes requested):
I think it would be better if you passed a list of cell klabels on the command line and then the frontend converted that into a pattern that is compiled and made available to the backend. We will need this feature in the haskell backend, but the Haskell backend does not and should not understand what it means to be a cell.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: Yes, exactly.

denis-bogdanas:
@dwightguth I don't know how to do that. Can you give any hints?

dwightguth:
@denis-bogdanas you can create a rule that matches just one cell with code like the following:

```
Rule(IncompleteCellUtils.make(KLabel(cellLabel), false, cellBody, false), BooleanUtils.TRUE, BooleanUtils.TRUE)
```

To string multiple cells together, use `KApply(Klabels.CELLS, cell1, cell2)`

You then just need to get the rhs of the spec rule `RewriteToTop.toRight(rule.body())` and use VisitK to extract the contents of the cells you care about. Then you compile the rule like it's a search pattern (see krun for how to do this) and pass it somehow to the backend. I can answer questions if you need more details.

denis-bogdanas:
For the moment I don't understand much. Will have to look at the code tomorrow. Also this generalization needs a test. How should I write a test for a spec that is supposed to fail?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
ehildenb: Let's try to do this another way. I guess if a user has a cell named `<haltCells>` (however improbable), this will cause problems, and special-cased cell names now will only be a problem for us in the future.

ehildenb (changes requested):

denis-bogdanas:
@ehildenb This is the same principle as for `<generatedTop>`. Should I call it `<generatedBoundaryCells>`? Options is called `--boundary-cells` now.

ehildenb:
@dwightguth what do you think of putting this option in `KProveOptions` instead? So that we can also eventually thread it through to the Haskell backend. I guess we can migrate it later too.

dwightguth:
I agree with putting it in KProveOptions, but if you do it my way (and we have to if we want it to work in the Haskell backend, which we do), you will not need a specially named cell.

denis-bogdanas:
@dwightguth Can we leave this generalization for another PR?

ehildenb:
You don't need to actually thread it to the Haskell backend yet, but I think you should move it to KProveOptions to make sure you don't end up depending on it in places that can't be supplied by KProveOptions (or that you thread the appropriate state through now).

denis-bogdanas:
@ehildenb I agree with moving to KProveOptions. But can I just rename the generated cell and keep implementation as is?

ehildenb:
I think you should probably change that as well. It's something which will affect the frontend, so I'm going to hold it to a higher standard than Java backend code. If @dwightguth says it's OK then fine, but I think we shouldn't add shortcuts where we say "we promise to fix it later", because in reality we won't look at this code again for a long time.

dwightguth:
I mean, if we don't do it now, it will just have to be done later, and the whole point of this process is to get this functionality merged the "right" way. This code will simply not work in the Haskell backend, and I think we're all agreed it's a valuable feature, so I see no reason why we should take a shortcut that will just cause problems later. Plus, if we do this the right way now, it could be useful to haskell backend developers during debugging if they choose to implement the feature.

denis-bogdanas:
@dwightguth Can you answer my question about how to test the failure case? Can I write a test that expects kprove exit code != 0?

dwightguth:
@denis-bogdanas write such a test for imp++ (where we already have a test for  kprove) and when I have time later today I will push some commits that make it work in the testing infrastructure

denis-bogdanas:
@dwightguth I don't think your proposal will work. You propose that the backend receives a pattern instead of a list of cells, and all that all the cell to pattern conversion logic to be moved to Kernel (or Kore?) module.

And the way you propose to build the pattern is to simply concatenate the boundary cells into a KApply. But this will only work if all boundary cells are top-level. And even, if I want to match say cells `<a>` and `<b>` they have to be put into a pattern like `<generatedTop> <a>_</a><b>_</b>_<generatedTop>`.

But your proposal won't work if some cells are nested. The pattern will have to nest them at the right depth just as they sit in configuration. This is the case in VSC. Our boundary cells are `k,pc`, and `<pc>` is nested. This is not a problem with current implementation of --boundary-cells because I don't match the entire current term against a pattern. I just extract the required cells from current term, build an auxiliary term with these cells, and match aux term with a similarly built term for target.

Even if there is a way to do this, I expect it to be time consuming and error prone. And I'm totally unfamiliar with front-end so I'm not particularly enthusiastic to work on this. Would rather finish upstreaming. Can we postpone this extension until it's really needed in Haskell backend?

denis-bogdanas:
@ehildenb @dwightguth I believe this is ready for review if you agree on current implementation.

Why I cannot change the current implementation of boundary term:

Cell `<generatedBoundary>` cannot clash with actual cells in the configuration. Even if there is an actual cell with the same name, the 2 cells will be at different nesting levels. Boundary term will look like 

```
<generatedBoundary><generatedBoundary> ... </generatedBoundary> ... </generatedBoundary>
```
If you don't like this cell, I can use `<generatedTop>` instead.

I cannot use the trick `KApply(Klabels.CELLS, cell1, cell2)` because I have to generate a `ConjunctiveFormula`, and it cannot contain plain `KApply` objects.

dwightguth:
If you follow the path that I outlined, the pattern that you create will be compiled with the same compiler we use to compile search patterns and thus will automatically deal with configuration abstraction and nesting levels.

denis-bogdanas:
@dwightguth 1. "You then just need to get the rhs of the spec rule `RewriteToTop.toRight(rule.body())` and use VisitK to extract the contents of the cells you care about.". Can you elaborate please? What is the role of this `RewriteToTop.toRight` and of `VisitK`. I don't see them used to process search pattern.

dwightguth:
@denis-bogdanas I'm referring to the idea that you want to extract the rhs of the spec being proven and inject the contents of the cells named on the command line into the pattern being constructed. I would assume the output of this particular step is a map from cell klabel to the term on the rhs of that cell in the spec. RewriteToTop.toRight gets the rhs of a term containing rewrites, and VisitK is a visitor pattern for asts of type K. Presumably you can add a method apply(KApply) which checks the klabel and if it's one of the cells you care about, adds the contents to the map you're constructing.

denis-bogdanas:
@dwightguth @ehildenb Ready for review. Implementation is different that what @dwightguth suggested. I couldn't understand much from hints, so I just followed the search pattern as example. I can explain more details if needed.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,4 +54,12 @@ public synchronized File specFile(FileUtil files) {
 
     @Parameter(names="--depth", description="The maximum number of computational steps to prove")
     public Integer depth;
+
+    @Parameter(names = "--boundary-cells", description = "The comma-separated list of cells used to mark the boundary " +
+            "of evaluation. If option is specified, execution ends when these cells in the current term match same " +
+            "cells in the target term. (Except for step 1, for which boundary checking is disabled.)" +
+            "If the whole current term matches target, execution is successful. " +
+            "Otherwise it fails on the present path. If option is not specified, full target term implication is checked " +
+            "on every step. Recommended value for KEVM: \"k,pc\"")
```
dwightguth: you should not be talking about kevm in the command line flags of k.

dwightguth (changes requested):
I don't fully understand the code you wrote in the Java backend so I will have to stare at it longer, but I noticed one quick thing that needs to be changed and I wanted to let you know.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
denis-bogdanas: Explanation of what happens in Java backend:
When the pattern matches, it generates a substitution with 2 kinds of variables: 

1. Cell body for cells I mentioned in `--boundary-cells`. These vars will have the name starting with `BOUND_`. These have to be preserved.
2. The rest of the config - cells I don't care about. These will be anonymous vars with name starting with `_`. In function `getBoundarySubstitution` I only keep the vars in 1st category.

For each step, after I get the clean substitution for current term and target term, I consider them to match if the equality `currentMatch == targetMatch` is not false. That is, if the match contains symbolic terms, and equality is neither true nor false, `boundaryCellsMatchTarget` will be true. That is because the option is not designed to be used in conjuction with symbolic reasoning. Matching should not depend on symbolic values.

I suppose this fragment is most complex to understand, the point where result of boundary matching is actually used:
https://github.com/kframework/k/blob/0b7cbaa63d4ca14c8fd6530ec8c62f82c050dac1/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java#L648-L668

All 3 cases are covered by the tests, and explained in the description of the option.

dwightguth: This seems complicated enough to warrant comments inline in the boundaryCellsMatchTarget and getBoundarySubstitution methods to me.

denis-bogdanas: > 
> 
> you should not be talking about kevm in the command line flags of k.

Done.

denis-bogdanas: > 
> 
> This seems complicated enough to warrant comments inline in the boundaryCellsMatchTarget and getBoundarySubstitution methods to me.

Done.

dwightguth (changes requested):
Something still doesn't seem right here. The pattern you're passing to the backend seems to just have a variable for each cell, and then the backend is performing some logic I don't understand in order to decide whether the substitution of that pattern "matches" the right hand side of the specification. But I don't really understand how that can work in a backend independent fashion. Why aren't we just passing the appropriate pattern in that contains the right hand side of the target and then checking whether we match that pattern?

dwightguth (approved):
Never mind, I think I understand how this works and it looks like the backend piece of it should be implementable in the haskell backend using this as a solution. So I am approving.

denis-bogdanas:
@ehildenb Waiting your review. No auxiliary cells are used in new implementation.

ehildenb (approved):

PR #312, opened 2019-02-22
Java Backend: Bugfix - remove unnecessary cast in result building expression

denis-bogdanas:
Required for recent option --format-failures. That is because when option is enabled and evaluation fails, result is not of type `KApply`.


ehildenb (approved):
Looks fine, maybe @dwightguth  should double check in case that cast was needed.

dwightguth (approved):

PR #317, opened 2019-02-25
Java Backend: BuiltinList simplification

denis-bogdanas:
All seems to work, here and in VSC.


dwightguth (approved):
Looks good, I assume you tested it on all your proofs too? If so, go ahead and merge.

PR #319, opened 2019-02-27
add pending documentation file

dwightguth:
Adds to the repo all the pending documentation we have so far (including what I just wrote for the recent new features)

### DIFF
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -1,0 +1,121 @@
+# Partially completed documentation
+
+This document contains documentation that has been written up to some extent
+but still needs to be ultimately included in the K manual which has not been
+written yet. New features of K that affect the surface language should be
+added to this document.
+
+## As Patterns
+
+New syntax has been added to K for matching a pattern and binding the resulting match in its entirety to a variable. The syntax is:
+```
+Pattern #as V::Var
+```
+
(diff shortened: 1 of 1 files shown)


ehildenb:
@dwightguth in the cases where it constitutes a new Kompile pass, we should include (in the manual) examples of before/after for developers reading the manual.

Another nit is that we should move all the already existing features to `documentation.md` instead of pending. Pending should be for features that we are implementing or have PRs for, but don't exist yet (in my mind).

dwightguth:
I mean, the idea I had behind pending was that it was features that have not been integrated into the body of the real documentation yet. Which is at this point all of them.

ehildenb:
I guess I was thinking of PRs into "pending" as "Ready for feature design review", and the actual "documentation.md" being "whatever documentation we have available for it right now, complete or not". So my thoughts would be, if it's actually a feature you can use in K right now, it belongs in "documentation.md", however bad the documentation is, and if it's a proposed feature that doesn't have an implementation yet but has approval, it belongs in pending.

ehildenb:
@dwightguth is it possible to have a function declared as `withConfig` in the program that I write? Or can they only show up on the RHS of rules?

dwightguth:
@ehildenb as the configuration does not yet exist when the initial program is constructed, there is no value to pass as the configuration parameter at that time. I believe this should be disallowed; at the very least, users should not expect it to work.

ehildenb:
I also think it should be disallowed, I guess I was just wondering if the current implementation actually disallows it.

ehildenb (approved):
Looks good, thanks for adding the examples.

PR #321, opened 2019-02-28
kernel: Global option --shutdown-wait-time

denis-bogdanas:
# Description from GlobalOptions help
If the option is set, a shutdown hook will be registered that, once invoked, interrupts the main thread and waits its termination. The wait time is the argument of this option, in ms. Useful if K is interrupted by Ctrl+C, because it allows the backend to detect interruption and print diagnostics information. Currently interruption detection is implemented in Java Backend. If K is invoked from KServer (e.g. Nailgun), the option is ignored.

### DIFF
--- a/kernel/src/main/java/org/kframework/utils/InterrupterRunnable.java
+++ b/kernel/src/main/java/org/kframework/utils/InterrupterRunnable.java
@@ -1,0 +1,29 @@
+// Copyright (c) 2019 K Team. All Rights Reserved.
+package org.kframework.utils;
+
+/**
+ * A runnable that interrupts the given thread when invoked, then awaits termination for the given wait time, in ms.
+ *
+ * @author Denis Bogdanas
+ * Created on 24-Dec-18.
+ */
+public class InterrupterRunnable implements Runnable {
+
+    private final Thread thread;
+    private final int waitTime;
+
+    public InterrupterRunnable(Thread thread, int waitTime) {
+        this.thread = thread;
+        this.waitTime = waitTime;
+    }
+
+    @Override
+    public void run() {
+        thread.interrupt();
+        try {
+            thread.join(waitTime);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


kernel/src/main/java/org/kframework/utils/InterrupterRunnable.java
```diff
@@ -0,0 +1,29 @@
...
+        this.waitTime = waitTime;
+    }
+
+    @Override
+    public void run() {
+        thread.interrupt();
+        try {
+            thread.join(waitTime);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
dwightguth, line 26: This is not an acceptable catch block. You should either raise a KEMException or else call Thread.currentThread.interrupt() if you want to swallow the exception. In general we should never just call printStackTrace in a catch block, that's a sign you have written boilerplate code without properly considering what should happen if the exception occurs.

dwightguth (changes requested):

dwightguth:
Have you tested that this still terminates after the timeout period if the backend fails to finish?

kernel/src/main/java/org/kframework/utils/InterrupterRunnable.java
denis-bogdanas: The exception happens in a shutdown hook. If I interrupt the current thread, this will be ignored. I suspect throwing another exception in this thread also will be ignored. Notifying the user with a printStackTrace, particularly in this context, I believe is the best option.

denis-bogdanas:
Yes, I tested. If Ctrl+C happens when running some code that does not check for interruption, K will silently shutdown after a few seconds. I used wait time of 5s in my work setup.

kernel/src/main/java/org/kframework/utils/InterrupterRunnable.java
dwightguth: Alright, that makes sense.

dwightguth (approved):

PR #322, opened 2019-02-28
Pattern macros

dwightguth:
This is looking like it's going to be quite necessary in order to succinctly express the C semantics while removing undesirable subsorting and overloading over terms of sort Type. It may also have other uses elsewhere in the semantics, that remains to be seen.


pending-documentation.md
```diff
@@ -140,3 +140,7 @@ syntax Int ::= foo(Int, GeneratedTopCell) [function, withConfig]
...
 rule <generatedTop>... <k> something ...</k> ...</generatedTop> #as Configuration => <generatedTop>... <k> foo(0, Configuration> ...</k> ...</generatedTop>
 ```
+
+## Macros and Aliases
+
+A rule can be tagged with the `macro` or `alias` attribute. In both cases, what this signifies is that this is a macro rule. Macro rules are applied statically during compilation on all terms that they match, and statically before program execution on the initial configuration. Currently, macros are required to not have side conditions, although they can contain sort checks. When a rule is tagged with the `alias` attribute, it is also applied statically in reverse prior to unparsing on the final configuration. Note that a macro can have unbound variables in the right hand side. When such a macro exists, it should be used only on the left hand side of rules, unless the user is performing symbolic execution and expects to introduce symbolic terms into the subject being rewritten. However, when used on the left hand side of a rule, it functions similarly to a pattern alias, and allows the user to concisely express a reusable pattern that they wish to match on in multiple places.
```
ehildenb, line 146: Examples.

ehildenb (changes requested):

pending-documentation.md
```diff
@@ -140,3 +140,19 @@ syntax Int ::= foo(Int, GeneratedTopCell) [function, withConfig]
...
+
+```
+syntax KItem ::= "foo" | "foobar"
+syntax KItem ::= bar(KItem) | baz(Int, KItem)
+rule foo => foobar [alias]
+rule bar(I) => baz(?_, I) [macro]
+rule bar(I) => I
+```
+
+This will rewrite `baz(0, foo)` to `foo`. First `baz(0, foo)` will be rewritten statically to `baz(0, foobar)`. Then the non-macro rule will apply (because the rule will have been rewritten to `rule baz(_, I) => I`). Then foobar will be rewritten statically to foo via the alias.
```
ehildenb: Put `foo` and `foobar` in backticks so they render nicely, as well as non-`macro`. We're trying to build up usable documentation.

Also, the last sentence seems a little confusing, Should it say "Then `foobar` will be rewritten statically to `foo` via the alias before unparsing happens." instead?

ehildenb (changes requested):
Example is good, but should have clearer explanation.

pending-documentation.md
ehildenb: I mean, "rewritten statically (in reverse) ...", specifically, the in reverse seems to need to be mentioned.

ehildenb (approved):
Looks good.

kernel/src/main/java/org/kframework/compile/ResolveAnonVar.java
```diff
@@ -75,6 +76,8 @@ K transform(K term) {
             public K apply(KVariable k) {
                 if (ANON_VAR.equals(k)) {
                     return newDotVariable();
+                } else if (FRESH_ANON_VAR.equals(k)) {
```
ehildenb: Any reason this isn't just `||` in the previous case?

ehildenb (changes requested):

ehildenb (approved):

PR #323, opened 2019-02-28
fix hidden warnings in domains.k and add regression test

dwightguth:


ehildenb (approved):
Not sure what the "hidden warnings" were about, but looks good!

traiansf (approved):

PR #324, opened 2019-02-28
Java Backend: KItem.evaluateFunction() error logging

denis-bogdanas:
1. When exception happens in `KItem.evaluateFunction()`, add to error message the current term and the invoked rule. Especially useful when debugging StackOverflowError. Limited in message size and number of frames.

2. Added option `--no-exc-wrap` to disable wrapping of error messages.


java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -590,6 +592,15 @@ public Term evaluateFunction(KItem kItem, TermContext context) {
                             if (!deterministicFunctions && result != null) {
                                 return result;
                             }
+                        } catch (KEMException e) {
+                            addDetailedStackFrame(e, kItem, rule, context);
+                            throw e;
+                            // DISABLE EXCEPTION CHECKSTYLE
+                        } catch (RuntimeException | Error e) {
+                            // ENABLE EXCEPTION CHECKSTYLE
```
dwightguth, line 599: same here, if you catch an Error there's no way that further allocations and execution are going to work correctly.

dwightguth (changes requested):

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
denis-bogdanas: Likely true for OOME, but it doesn't matter for that error, anyway. I don't care where exactly it was thrown first time. I can add a separate clause for OOME if you want. But this code is very useful for other errors.

denis-bogdanas: Added a catch for OOME here and in 3rd PR. Should be enough to address your concern.

dwightguth: This really doesn't address my concerns at all. There's lots of subclasses of Error for which it's not reasonable to expect to be able to catch them; that is in fact the purpose of the Error class. So I would rather ask, what Errors do you need to be able to catch? And only catch those.

denis-bogdanas: Ok I changed as you asked. Catching RuntimeException, AssertionError, StackOverflowError.

dwightguth (approved):

PR #325, opened 2019-03-01
Java Backend error logging: SymbolicRewriter: improved exception handling

denis-bogdanas:
Main changes:
1. Print final configuration if Exception or Error is thrown during evaluation, including nr of paths in progress.
2. Separate handling of OutOfMemoryError to avoid an inner OutOfMemoryError.
3. Support for interruption. If current thread is interrupted summary box will be printed, just like for exception.
4. Added a warning if evaluation ends on step 1 with no successors.
5. Disabled toString caching when full term is printed, because it's more likely to trigger OutOfMemoryError at this point, and there are less cache reuse opportunities.


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -627,8 +628,11 @@ private K disjunctResults(List<K> results) {
             }
 
             for (ConstrainedTerm term : queue) {
+                boolean alreadyLogged = false;
+                try { //not formatting to minimize git merge conflicts.
```
dwightguth: uhhhhhhh... this is what a git merge strategy that ignores whitespace is for.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -726,6 +724,26 @@ private K disjunctResults(List<K> results) {
                         nextQueue.add(result);
                     }
                 }
+
+                if (Thread.currentThread().isInterrupted()) {
+                    throw KEMException.criticalError("Thread interrupted");
+                }
+                    // DISABLE EXCEPTION CHECKSTYLE
+                } catch (OutOfMemoryError e) {
```
dwightguth, line 761: can we just catch all Error and immediately rethrow them? there are probably a bunch of other Errors that we probably don't want to try to execute code after they're thrown. And there's no way that an allocation is going to work after you catch an OutOfMemoryError.

dwightguth (changes requested):

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: I did this to simplify code review mostly. Intellij knows to make diff modulo whitespaces. Can make another PR just for formatting. Are you ok with leaving as is?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: Then what's the point of catching them? There are many errors of interest: AssertionError and StackOverflowError in particular. They can't cause issues here.
The only error that could be hidden at this point by a new error is OutOfMemoryError. There is still a big chance that printSummaryBox here will work, and I've seen it working on Jenkins. A little allocation is possible because some pointers get released from the moment exception is thrown till the moment it is caught. And most OOME is GC limit exceeded - that means there is still a bit of free memory left but GC takes almost 100% of the time.

And I do more clever things with OOME in another PR that implements advanced memory profiling.

dwightguth: You name AssertionError and StackOverflowError in particular. I agree that these can probably be caught, but why not catch them explicitly rather than catching all errors?

denis-bogdanas: I named these 2 because I encountered them explicitly. What's wrong with catching Throwable? If we ever encounter problems because some other error is caught, this clause can be refined. But for now, maybe it's more useful this way.

denis-bogdanas: This looks like a typical case of law of triviality.

denis-bogdanas: Ok I changed as you asked. Catching RuntimeException, AssertionError, StackOverflowError.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
dwightguth: don't worry about code review, that's what ?w=1 is for.

dwightguth (changes requested):

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -691,72 +694,86 @@ private K disjunctResults(List<K> results) {
...
+                            logStep(step, v, term, true, alreadyLogged);
+                            // re-running constraint generation again for debug purposes
+                            if (global.javaExecutionOptions.logBasic) {
+                                System.err.println("\nApplying specification rule\n=========================\n");
+                            }
+                            if (visited.add(result)) {
+                                nextQueue.add(result);
+                            } else {
+                                if (term.equals(result)) {
+                                    throw KEMException.criticalError(
```
dwightguth, line 710: I'm not sure this should be an error, as the specification has technically been proven true at this point.

dwightguth (changes requested):

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: We already discussed this in some old PR. I don't see a valid use case for a spec rule that rewrites the term to itself.

Think of it like this: Any spec rule of the form `rule A => A` is technically valid, can be proved and represents no computation advancement. Without this check, once this rule matches during evaluation of another rule, that other rule gets proved as well. Why?

But it happened in the past that such rules were specification errors.

dwightguth: I'm fine if it's just a matter of forbidding a spec rule that rewrites to itself, but my understanding (correct me if I'm wrong) is that this will also occur if the configuration rewrites something to itself, in which case proving the spec correct is a valid proof of that spec relative to that semantics. I would rather not make it impossible to prove properties about certain types of semantics in order to solve this problem. If you just want to forbid a spec that rewrites something to itself, that can be done as a preprocessing step in the k frontend without interfering with the ability to prove a nonterminating semantics.

Am I misunderstanding the scope here?

denis-bogdanas: It could happen in a less obvious fashion, like `rule <context> X => 0 </context>` where X matches 0. Can this be done at pre-processing phase? Is it worth the effort? But honestly this case was already discussed when I introduced the warning in the first place. Here I just converted the warning into error. @ehildenb was reviewer that time.

I don't understand your last comment. What you mean by "configuration rewrites something to itself". Can you give an example?

dwightguth: So, what I mean is, suppose I have a semantics where a particular term doesn't terminate, and I want to prove that it doesn't terminate. Would not this change make such a proof impossible?

denis-bogdanas: Even so, you expect that return to a previously visited state will happen in more than 1 semantics step. By semantics steps I mean steps in which semantics rules are applied, not spec rules. If my change is in place, you would have to define spec rules in such a way that they don't capture one full loop, but a fragment of a loop. You'll have 2 rules, some `rule A=> B` and `rule B => A`, each of them encoding more than zero steps of the concrete semantics.

For the extremely unlikely case that we actually need to prove non-termination, this seems like a reasonable inconvenience. Much better then accidentally proving incorrect stuff due to easy to make errors in spec rules. (I have arguments why it's unlikely, but don't want to get into this debate.)

dwightguth (approved):

PR #326, opened 2019-03-01
Java Backend error logging: FastRuleMatcher: improved exception handling

denis-bogdanas:


dwightguth (approved):

dwightguth (changes requested):
I've decided that we probably shouldn't catch Throwable here like this.

dwightguth:
If you could fix the other case where you're catching Throwable and then executing code that performs allocations afterwards as well, I'd appreciate it. I didn't spot it during review or I would have asked it to be fixed.

denis-bogdanas:
Ok I changed as you asked. Catching RuntimeException, AssertionError, StackOverflowError.

dwightguth (approved):

PR #331, opened 2019-03-04
K ml to kore ml

traiansf:
An attempt to handle and translate K ML constructs into their corresponding KORE counterparts.


traiansf:
If successful, this should replace PR #280

traiansf:
@dwightguth I would like some feedback on the work so far (Commits c8f76bd and c81e7bd ). The next phase would be a transformation adding parameters to the ML KLabels with their corresponding sorts, and using these parameters when generating KORE from them.

dwightguth:
Before I can review further I need some indication why you have touched so many parts of the code in this PR. It doesn't seem like it should have to affect the compiler or the kore module.

The sort parameters should already be being added via the poly attribute on the production, if everything else works correctly. Maybe instead of keeping track of the K module, we should just move the ML operations into a separate module and import them from BASIC-K?

traiansf:
This work was restarted as part of https://github.com/kframework/k/pull/336

PR #335, opened 2019-03-04
install script

dwightguth:
This script is intended to be run from within semantics repos which want to download the binary release of a specific commit of K. I have a followup PR to evm-semantics which demonstrates how to use it to remove the requirement to build K from source.


ehildenb (approved):
LGTM

PR #336, opened 2019-03-05
use ml constructs in k and encode them into kore

traiansf:


k-distribution/include/builtin/kast.k
```diff
@@ -63,32 +63,39 @@ module KSEQ
...
-  syntax KBott      ::= #KVariable
-  syntax KBott      ::= KConfigVar
-  syntax KLabel     ::= #KVariable
 
-  syntax KBott ::= "#True" [klabel(#True), symbol, poly(0)]
-  syntax KBott ::= "#False" [klabel(#False), symbol, poly(0)]
-  syntax KBott ::= "#Not" "(" K ")" [klabel(#Not), symbol, mlOp, poly(0, 1), mlUnary]
-                 | "#Ceil" "(" K ")" [klabel(#Ceil), symbol, mlOp, poly(0; 1), mlUnary]
-                 | "#Floor" "(" K ")" [klabel(#Floor), symbol, mlOp, poly(0; 1), mlUnary]
+  syntax K ::= "#True" [klabel(#True), mlOp, symbol, poly(0)]
```
dwightguth: Now that these are all syntax k again, you can replace the > blocks and remove the syntax priority.

k-distribution/include/builtin/kast.k
```diff
@@ -63,32 +63,39 @@ module KSEQ
 endmodule
 
 
-// To be used when parsing/pretty-printing symbolic configurations
-module KSEQ-SYMBOLIC
-  imports KSEQ
+module ML-SYNTAX
+  imports BASIC-K
+
   syntax #KVariable ::= r"(?<![A-Za-z0-9_\\$!\\?])(\\!|\\?)?([A-Z][A-Za-z0-9'_]*|_)"   [token, prec(1), autoReject]
```
dwightguth: I don't want this production in this module because #KVariable is not a real sort.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -160,6 +162,9 @@ public String convert(boolean heatCoolEq) {
             }
         }
         for (Production prod : iterable(module.productions())) {
+            if (prod.att().contains("mlOp")) {
```
dwightguth: I would rather we call the function that enumerates the klabels than check an attribute like this.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -577,7 +585,11 @@ private void convertRule(Rule rule, boolean heatCoolEq, Sort topCellSort, Map<St
                     .map(i -> (NonTerminal) i)
                     .map(NonTerminal::sort).collect(Collectors.toList());
             productionLabel = production.klabel().get();
-            if (isFunction(prod) || rule.att().contains(Attribute.ANYWHERE_KEY)) {
+            if (ConstructorChecks.isBuiltinLabel(productionLabel)) {
```
dwightguth: Why is this necessary and is it possible we could instead improve the generic handling of the poly attribute?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -598,8 +610,11 @@ private void convertRule(Rule rule, boolean heatCoolEq, Sort topCellSort, Map<St
                     rule = rule.withAtt(rule.att().add(Attribute.SIMPLIFICATION_KEY));
                 }
             }
+            sb.append("  axiom{R");
+            StringBuilder oldSB = sb;
```
dwightguth: This is very weird and not nice. What are you trying to accomplish here?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -925,7 +945,7 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
 
     private boolean isFunction(Production prod) {
         Production realProd = prod.att().get("originalPrd", Production.class);
-        if (!realProd.att().contains(Attribute.FUNCTION_KEY)) {
+        if (!realProd.att().contains(Attribute.FUNCTION_KEY) && !realProd.att().contains(Attribute.ML_SYMBOL_KEY)) {
```
dwightguth: Again, I would rather call ConstructorChecks.isBuiltinLabel here.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1035,12 +1055,19 @@ private void collectAttributes(Map<String, Boolean> attributes, Att att) {
     }
 
     private static final Production INJ_PROD = Production(KLabel(KLabels.INJ, Sort("From"), Sort("To")), Sort("To"), Seq(NonTerminal(Sort("From"))));
+    private static final Production CEIL_PROD = Production(KLabel(KLabels.ML_CEIL.name(), Sort("From"), Sort("To")), Sort("To"), Seq(NonTerminal(Sort("From"))));
```
dwightguth: These should not be necessary as we now have productions in the grammar for them. That's what computePolyProd is for.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1035,12 +1055,19 @@ private void collectAttributes(Map<String, Boolean> attributes, Att att) {
...
+    private static final Production EQUALS_PROD =
+            Production(KLabel(KLabels.ML_EQUALS.name(), Sort("From"), Sort("To")), Sort("To"), Seq(NonTerminal(Sort("From")), NonTerminal(Sort("From"))));
 
 
     private Production production(KApply term) {
         if (term.klabel().name().equals(KLabels.INJ))
             return Production(INJ_PROD.klabel(), INJ_PROD.sort(), INJ_PROD.items(), Att.empty().add("originalPrd", Production.class, INJ_PROD));
-        scala.collection.Set<Production> prods = module.productionsFor().apply(((KApply) term).klabel());
+        if (term.klabel().name().equals(KLabels.ML_CEIL.name()))
```
dwightguth: I think we can revert this block of changes

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1113,6 +1140,12 @@ private void convert(Sort sort, Production prod) {
     }
 
     private void convert(Sort sort, boolean var) {
+        if (sort.name().equals(AddSortInjections.SORTPARAM_NAME)) {
```
dwightguth, line 1151: I have no idea why we have a "SortParam" sort but it seems very strange and hackish to me. What are you trying to accomplish?

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -118,14 +136,38 @@ public K addInjections(K term, Sort expectedSort) {
...
+
     private K visitChildren(K term, Sort parentSort, Sort actualSort) {
         Att att = term.att().add(Sort.class, actualSort);
         if (term instanceof KApply) {
             KApply kapp = (KApply)term;
             if (kapp.klabel().name().equals("inj")) {
                 return term;
-            } else if (kapp.klabel().name().equals("#Or")) {
-                return KApply(KLabel(kapp.klabel().name(), actualSort), KList(addInjections(kapp.items().get(0), actualSort), addInjections(kapp.items().get(1), actualSort)), att);
+            } else if (kapp.klabel().name().equals("#True")
```
dwightguth: I don't like this. This is what the generic poly machinery should be for.

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -182,11 +224,27 @@ public Sort sort(K term, Sort expectedSort) {
             KApply kapp = (KApply)term;
             if (kapp.klabel().name().equals("inj")) {
                 return kapp.klabel().params().apply(1);
-            } else if (kapp.klabel().name().equals("#Or")) {
+            } else if (kapp.klabel().name().equals("#True")
```
dwightguth: Same

dwightguth (changes requested):
A good start but it needs significant work. most importantly:

* We shouldn't treat the mlOp attribute as special, we should simply rely on a whitelisted list of klabels in ConstructorChecks.
* We should not hard code specific code for specific ml connectives. The sort signatures and parameters should all be handled directly and generically by the productions in the grammar and their poly attribute.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: I agree this is very weird and not nice, but it's because one might need to add extra sort variables which we don't know of until we go through the terms.

I guess we could be traversing the terms here and collecting the sort variables; that would be less weird, although not very nice as well.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: I hope to catch you on slack later and discuss this more interactively, but to give you a head-start, here is the issue I'm trying to address:


Well, the issue here is that ml opeations are parameterized.  Some of them (And, Or, Implies) have the sort of their operands, and perhaps for those it the poly mechanism might work, but others like Ceil and Equals, transport the truth from one sort to another, so in some sense they are injections.

the issue becomes even weirder when I have a combination of them, e.g. Ceil(Ceil(X:Int)) becomes \ceil{Q,R}(\ceil{Int, Q}(X)), where Q is a new sort parameter.  And of course, if any of Ceil, Equals, True or False are at the top, they would need a sort parameter as their output sort.

That is why I thought of annotating the ML constructs with their corresponding sort parameters during AddSortInhections.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: I would rather annotate the rule with an attribute containing the sort variables it needs when those variables are created.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: I should think Ceil(Ceil(X:Int)) would just become `\ceil{Int, R}(\ceil{Int,Int}(X))`. Then you wouldn't need new sort parameters. Is there a reason this is theoretically insufficient?

traiansf: tI think this would be unacceptable because then the pattern, being too specialized, might not match (and be able to be replaced by its RHS) in all contexts in which it would match.
Consider the rule motivating this PR:
`#Ceil(I1:Int /Int I2:Int) => #Equals(I2 =/=Int 0, true)`

Using the existing mechanism `#Ceil(I1:Int /Int I2:Int)` would become `\ceil{Int,Int}(I1:Int /Int I2:Int)` while `#Equals(I2 =/=Int 0, true)` would become `\equals{Bool,Bool}(I2 =/=Int 0, true)`.  However, I need to put an `\equals` on top of both of them, so if I leave the `\ceil` as Int and the `\equals` as bool, I would need to add inj to both of them.

In this particular case,the final result I currently generate (and which is the desirable one) is `\equals{Q, R}(\ceil{Int,Q}(I1:Int /Int I2:Int)), \equals{Bool, Q}(I2 =/=Int 0, true))`

But, as said before, you might get even more parameters, say if you want to express something like 
`#Equals(#Equals(#Not(#False), #True), #True)`, which should become something like `\equals{R,S}(\equals{Q,R}(\not{Q}(\bottom{Q}), \top{Q}), \top{R})`

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
traiansf: I manage to reduce it to the symbols which can have any output sort (True/False/Ceil/Floor/Equals).  I'm not sure how to reduce it further but we can talk about that.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: What if we go the other way? have it be `\equals{R, R}(\ceil{Int,R}(...), \equals(Bool,R}(...))` Does that have the same problem?

kernel/src/main/java/org/kframework/compile/ConvertDataStructureToLookup.java
```diff
@@ -106,23 +106,25 @@ public ConvertDataStructureToLookup(ConvertDataStructureToLookup copy) {
...
-            if (p.att().contains("filterElement")) {
-                set.add(Tuple2.apply(KLabel(p.att().get("filterElement")), p.klabel().get()));
-            }
-            if (p.att().contains("wrapElement")) {
-                set.add(Tuple2.apply(KLabel(p.att().get("wrapElement")), p.klabel().get()));
-            }
-            return set.stream();
-        }).distinct().collect(Collectors.toMap(Tuple2::_1, Tuple2::_2));
+        return stream(m.productions())
+                .filter(p -> p.att().contains(Attribute.ASSOCIATIVE_KEY) && p.att().contains("element"))
```
traiansf, line 110: except for `&& p.att().contains("element")` on this line, this block of changes is just indentation.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -598,8 +610,14 @@ private void convertRule(Rule rule, boolean heatCoolEq, Sort topCellSort, Map<St
                     rule = rule.withAtt(rule.att().add(Attribute.SIMPLIFICATION_KEY));
                 }
             }
+            sb.append("  axiom{R");
+            Option<Set> sortParams = rule.att().getOption("sortParams", Set.class);
+            if (sortParams.nonEmpty()) {
+                for (Object sort : sortParams.get())
+                    sb.append("," + sort);
```
dwightguth: I'd rather we cast this to a Sort and called name()

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -44,11 +45,37 @@
     private final Module mod;
     private final Map<KLabel, KLabel> collectionFor;
 
+    private int freshSortParamCounter = 0;
+    private Set<String> sortParams = new HashSet<>();
+    public static final String SORTPARAM_NAME = "SortParam";
```
dwightguth: can we add a hash in front of this sort name?

dwightguth (changes requested):
looks a lot better. If you fix a couple more minor issues, I will approve. It would be nice if you could test; kevm in the llvm backend against this version of k though just to be on the safe side. the few integration tests for the llvm backend in the k frontend don't provide the most coverage, and I'd like to test more thoroughly just to make sure the sort parameters thing doesn't break anything.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: Actually `sortParams` is a set of `String`s  holding the sort parameter names, so no casting is needed. I changed the name of the iterator to make that clear.

dwightguth (approved):

PR #338, opened 2019-03-06
[BUGFIX] Equals formatting correction

ehildenb:
I've added a regression test with fails before the commit which fixes it.

This is blocking the haskell backend because it relies on K integration tests which print out `#Equals`.


dwightguth (changes requested):
This doesn't look like the right value of the format attribute. I should think it should be `format(%i%1 %2%d%n%3%n%i%4 %5%d)`

ehildenb:
@dwightguth  it passes the tests that are there. So do you want different expected output for those tests?

dwightguth (approved):

PR #340, opened 2019-03-06
fix bug in or patterns when used outside of a rewrite

dwightguth:
There was a bug where or patterns were appearing on the rhs of rules when an or pattern was used outside of a rewrite, despite them not being syntactically valid there. This fixes that issue by introducing a compiler pass that guards or patterns with an #as pattern so that or patterns used outside of any rewrite have an unambiguous substitution on the rhs.

We also fix a bug where anonymous variables could not be used in or patterns in the ocaml backend because of a compiler error in ocaml.


traiansf (approved):
Looks good to me.  My only concerned is what if you actually want to write an #Or pattern in the RHS, say, e.g., for the RHS of a specification rule.

ehildenb (approved):

PR #344, opened 2019-03-07
Java Backend SMT warnings: do not print stack trace

denis-bogdanas:


dwightguth:
This feels iffy to me. If a stack trace exists for an error, it should be possible to obtain it. We already don't print stack traces by default, so if the user asks for a stack trace, they should get a stack trace if one exists.

denis-bogdanas:
I can rewrite the logic to not have a stack trace at all. I have to use --debug to show me stack traces of errors, not warnings. If I enable --debug on Jenkins now, I get logs full of thrash.

dwightguth:
Hmmm. Let's do this. Let's make it so that --debug enables stack traces on errors and --debug --debug enables stack traces on errors and warnings.

denis-bogdanas:
If I'm telling you that stack trace for these warnings contains no useful information, why don't you believe me? Every single run of kprove prints a couple of them.

dwightguth:
it's not that I don't believe you, it's that I believe that we should always have a way to print a stack trace if one exists, even if in the vast majority of cases it doesn't contain useful information. That's why I'm saying that we should just disable printing the stack trace of all warnings unless the user passes --debug twice. That's a much more measured and idiomatic solution to the problem than to pick and choose which warnings get stack traces and which ones don't.

denis-bogdanas:
I'm quite knowledgeable of the code that generates these warnings, I'm well aware of the context, and this allows me to state that stack trace is not useful. I guess you can believe if I say this code can be refactored to return some false, or null, instead of throwing exception. But this would be a time consuming and otherwise useless exercise. The message of these exceptions contains all information one needs to know, plus extra details can be inferred with --debug-z3 option. I never ever had to look at stack traces.

dwightguth:
This is why I am proposing to you an idiomatic solution such that you are able to see the stack trace of errors but not the stack trace of warnings, while still making it possible to see the stack trace of warnings if you so choose. Your solution however hardcodes one particular warning to make it impossible to see the stack trace without modifying the code, even if some situation arises for some developer wherein they wish to view that stack trace. I am aware of your statement that in the case of this particular error, it is unlikely that the user will want this stack trace regardless, and I even agree that's probably true, but I also don't know enough about the internals of z3 to know for sure that there might not be some uncommon case where you might want to have the stack trace. Furthermore, this is the first step down an extremely slippery slope that ultimately leads to a lack of a robust debugging infrastructure in K. Today we have one warning we don't want users to see the stack trace for. Tomorrow there may be others. Where can we draw the line in the sand? Sooner or later if we make every stack trace we don't think the user wants to see unavailable to developers, we will run into a case where the tool crashes on some users machine but the error is not reproducible locally and we have no way to obtain the stack trace because of our poor decisions, and we will thus have a very bad experience for our user, who has to download a whole new version of k and possibly even rewrite their semantics just to give us the information we need to debug their bug.

The entire reason KExceptions have a chained exception is because it is sometimes needed for debugging. You have not given me a suitable justification for violating this very important invariant in the code, not when I have already described how to get the behavior you want another way.

denis-bogdanas:
Fine, I can do the way you asked only to get over it. But how can I implement `--debug --debug` ? JCommander doesn't accept the same option multiple times.

denis-bogdanas:
@dwightguth I changed the warning message. Now it looks like:
```
[Warning] Internal: execution phase: missing SMTLib translation for #asByteStack . Please re-run with the --debug-z3 flag. Search the logs starting with 'Z3 warning' to see the Z3 implication that generated the warning.
```
And the log it's talking about is:
```
Z3 warning. Query not generated: missing SMTLib translation for #asByteStack

Z3 Implication (Function rule implication) RHS dropped (cannot be proved):
ConjunctiveFormula(
  equalities:
    _==K_(_<Int__INT(sizeWordStackAux(#asByteStack(P1_1641:Int),, Int(#"0")),, Int(#"32")),, Bool(#"true"))
)
```

Hope this is evidence that there is already more useful info logged than stack trace.

dwightguth:
http://jcommander.org/#_lists

denis-bogdanas:
@dwightguth I'm aware of JCommander tutorial. The section you mentioned does not allow specifying the same option twice without arguments, like you proposed `--debug --debug`. How about `--debug` for errors and `--debug-warnings` for errors and warnings?

dwightguth:
If you've tested it with List<Bool> and it doesn't work, then what you suggest sounds fine to me.

denis-bogdanas:
@dwightguth Implemented through `--debug-warnings`. I didn't find any clean way to chain options --debug and --debug-warnings, so 2nd does not include all functionality of the 1st.

dwightguth:
The solution to "chaining" options is to implement a method on GlobalOptions that returns the or of the individual options, and make the unneeded individual variables private.

denis-bogdanas:
Hoped to avoid that. Options chaining implemented.

kernel/src/main/java/org/kframework/main/GlobalOptions.java
```diff
@@ -99,4 +102,8 @@ public WarningsConverter(String optionName) {
 
     @Parameter(names={"--no-exc-wrap"}, description="Do not wrap exception messages to 80 chars. Keep long lines.")
     public boolean noExcWrap = false;
+
+    public boolean debug() {
+        return debug | debugWarnings;
```
dwightguth: ||

dwightguth (approved):

dwightguth:
I approved but it would be great if you fixed one minor issue before you merge.

PR #345, opened 2019-03-07
Update Haskell backend submodule version

ehildenb:
Hopefully this will catch breakages in the backend caused by the frontend before they get merged.


k-distribution/tests/regression-new/Makefile
```diff
@@ -32,7 +32,7 @@ SUBDIRS=issue-2273 \
 	kore-brackets \
 	quadratic-poly-unparsing \
 	io-java \
-	symbolic-int-function \
+	symbolic-int-function
```
dwightguth: this was done deliberately in order to simplify diffs that modify this file, which frequently conflict with one another. I'd appreciate it if it was kept the way it is now.

src/main/scripts/test-haskell-backend
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+rm -rf haskell-backend-kore
+git clone 'https://github.com/kframework/kore' haskell-backend-kore
```
dwightguth: this doesn't seem at all the right way to do this. you should simply use the source directory of the haskell backend that has already been checked out.

dwightguth (changes requested):

k-distribution/tests/regression-new/Makefile
ehildenb: Makes sense, fixed.

ehildenb:
@dwightguth this PR now just updates the Dockerfile and haskell backend submodule (which has new directory structure).

dwightguth (approved):

PR #349, opened 2019-03-09
Java Backend: eliminating non-determinism in logs

denis-bogdanas:
Sorting rules so that they always appear in collections in the same order, to eliminate non-determinism when running same task in different environments. Replaced some lists/maps with linked version.

Not all sources of non-determinism were eliminated. Some come from `kompile`. Surprisingly, this eliminated these spurious errors on Jenkins: `Critical: [non-deterministic function definition]: more than one rule can apply to the term`


java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
```diff
@@ -197,6 +199,10 @@ public void addKoreRules(Module module, GlobalContext global) {
         List<org.kframework.definition.Rule> koreRules = JavaConversions.setAsJavaSet(module.rules()).stream()
                 .filter(r -> !r.att().contains(AUTOMATON))
                 .collect(Collectors.toList());
+
+        //Ensures that rule order in all collections remains the same across across different executions.
+        koreRules.sort(Comparator.comparingInt(rule -> rule.body().hashCode()));
```
dwightguth, line 204: This is not going to guarantee a unique order as two rules can have the same body (and thus the same hash code) but different side conditions.

denis-bogdanas: That's right. But it still reduces non-determinism enough to eliminate the issues I mentioned above.

dwightguth (approved):
I'm still not entirely clear what this is used for, but as long as you're comfortable with the fact that the logs are not actually going to be determinstic as a result of this change, I don't have a particular opposition to it.

PR #350, opened 2019-03-10
smt.k: Interface for constructing smtlib2 scripts and calling into Z3 and CVC4

nishantjr:
This is currently done via calls to `#system`


dwightguth (commented):
I really need to know what you're trying to do before I can know whether this is an appropriate solution to the problem. This is a significant design decision and I have absolutely no context to decide whether it is appropriate as an addition to the K language.

nishantjr:
@dwightguth Sorry, wasn't ready for a review yet and didn't see your review. Should have put that in the title.

I had built this for use with the matching-logic-prover repository where we need to translate matching logic queries to SMTLIB and sent them to an SMT solver. I think this is something useful for multiple projects (I'm in the process of implementing a semantics for dafny that this will be useful for). It does not really affect the core K Language and is just a library on the side.

Initially, I had built a hook for Z3 but maintenance-wise that seems more difficult -- it only supported the Java backend, and only for Z3. It also required depending on the Z3 jar being around at mvn compile time.

dwightguth:
In that case, let's revert the change to the Dockerfile and remove the regression test for CVC4. I'm not ready to require K users to install multiple solvers and it is more important that we test that K works /without/ CVC4 installed than that we test one single rarely-used module with multiple solvers.

nishantjr:
Fair enough. I'm OK with testing the CVC4 functionality in the `matching-logic-prover` repository for now, where it is used. I'm still working on figuring out why older versions of Z3 (such as the one used in CI) don't appear to be reading the entire generated file.

ehildenb:
Please re-open if needed.

PR #353, opened 2019-03-11
Some regressions in how generation of fresh symbolic values was working.

ehildenb:
I guess we need to decide how much of this is desired behavior.

It makes sense to me that this function should be annotated with `impure`, since it will be returning a different value dependent on the backend environment each time.


dwightguth:
Neither the impure attribute nor the memo attribute should have any impact on the result value of a function call.

ehildenb:
Taken care of by #354

PR #354, opened 2019-03-11
Java Backend: function cache fix

denis-bogdanas:
Proper support for caching+fresh vars and impure functions. Terms using fresh vars and impure functions are not cached.

## Description
Added a field `JavaSymbolicObject.isCacheable`, computed recursively on the AST tree. It is only computed on fully evaluated terms. The value depends not only on the present term bot also on the source term from which present term was computed through function evaluation.
A result term is not cacheable if either:
1. Source term KLabel is `[impure]`.
2. A fresh variable is created during evaluation from source term to present term.
3. Any sub-term is not cacheable.

The entry `source -> result` is added to the cache only for cacheable results. But the entry `result -> result` is added for any term, to distinguish terms that are already evaluated from unevaluated terms.


denis-bogdanas:
@ehildenb Ready for review. See description. Also includes your tests. Only the latest commit contains the implementation.

ehildenb (approved):
LGTM, have you checked that you get good performance on KEVM?

denis-bogdanas:
@ehildenb Yes. It does not affect KEVM because it doesn't use impure/freshVar.

PR #358, opened 2019-03-13
Update README.md

daejunpark:
For building the haskell backend on Ubuntu 18.04


dwightguth:
Neither the haskell backend nor llvm backend dockerfilke installs this package, so I'm not quite clear what problem you think this is solving. Can you clarify?

daejunpark:
When I installed the haskell backend (and the repl) in a fresh ubuntu 18.04 machine, I had to install `libtinfo-dev`, otherwise the haskell build (especially the stack build) failed.

@ttuegel could you please clarify a bit more?

dwightguth:
I'm fine if there's a missed dependency, but if there is, I'd like to know why it is not listed in the haskell-backend dockerfile, since that would seem to suggest we may be missing other dependencies also.

ttuegel:
> I had to install `libtinfo-dev`, otherwise the haskell build (especially the stack build) failed.

`stack build` failed for @daejunpark with an error message about that library being missing. It appeared that the GHC build which Stack provides for Ubuntu requires that library. This is the extent of my knowledge.

ehildenb:
@dwightguth, the following are in the `Dockerfile` but not in the README. Perhaps one of these brings in `libtinfo-dev`, which is a fairly common package I think.

```
curl
debhelper
gcc
git
llvm-6.0-tools
```

dwightguth:
You're right, llvm-6.0-tools depends on llvm-6.0-dev which depends on libtinfo-dev. @daejunpark can you please add gcc and llvm-6.0-tools to the readme instead?

dwightguth (approved):

PR #359, opened 2019-03-13
Java Backend exception handling: FastRuleMatcher: adding stack frames when exception is caught

denis-bogdanas:
Essentially this replaces System.err messages with stack frames, like it is done in `KItem.evaluateFunction()`.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
@@ -33,10 +33,12 @@ import org.kframework.backend.java.util.StateLog;
 import org.kframework.backend.java.utils.BitSet;
 import org.kframework.builtin.KLabels;
 import org.kframework.kore.KApply;
+import org.kframework.utils.errorsystem.KEMException;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Formatter;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


denis-bogdanas:
@dwightguth ready for review

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -231,12 +233,15 @@ private BitSet matchAndLog(Term subject, Term pattern, BitSet ruleMask,
...
+        } catch (KEMException e) {
+            addDetailedStackFrame(e, subject, pattern);
+            throw e;
             // DISABLE EXCEPTION CHECKSTYLE
         } catch (RuntimeException | AssertionError | StackOverflowError e) {
             // ENABLE EXCEPTION CHECKSTYLE
-            System.err.format("\nException during matching rule pattern:\n\t%s\nSubject:\n%s\nPattern:\n%s\n",
-                    e.getMessage(), getLogString(subject), getLogString(pattern));
-            throw e;
+            KEMException newExc = KEMException.criticalError("", e);
```
dwightguth, line 242: There should be a message for this exception.

dwightguth (changes requested):

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
denis-bogdanas: it's the message of original exception. K-specific stack frames give more info. Tested, looks best this way.

dwightguth: If you don't pass --debug, you're going to get an error with no text. That's not really great.

denis-bogdanas: @dwightguth Nope. Nice exception message will be present with and without --debug.
With --debug:
```
org.kframework.utils.errorsystem.KEMException: [Error] Critical:  (unexpected subject class at matching: class org.kframework.backend.java.kil.KList)
  while matching rule pattern:
...
	at org.kframework.utils.errorsystem.KEMException.create(KEMException.java:108)
...
```

Without --debug:
```
[Error] Critical:  (unexpected subject class at matching: class org.kframework.backend.java.kil.KList)
  while matching rule pattern:
...
```

dwightguth: Ah, I didn't realize we were printing the exception message as part of the kexception now. It's possible that might generate some output that contains stuff the user doesn't care about, but we can address that if we find an example of that. Alright, I'm okay with this for now then.

dwightguth (approved):

PR #360, opened 2019-03-14
A few more allowed simplifications of the builtin Map.

ehildenb:


k-distribution/include/builtin/domains.k
```diff
@@ -155,6 +155,23 @@ module MAP
...
 
+module MAP-SYMBOLIC [symbolic]
+  imports MAP
+  imports K-EQUAL
+
+  rule ((K1 |-> V1) MAP) [ K2 ] => V1         requires K1  ==K K2
+  rule ((K1 |-> V1) MAP) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
+
+  rule ([ K1 <- K2 ] MAP) [ K2 ] => V1         requires K1  ==K K2
+  rule ([ K1 <- K2 ] MAP) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
```
daejunpark: syntax? `[ K1 <- K2 ] MAP` instead of `MAP [ K1 <- K2 ]`?

k-distribution/include/builtin/domains.k
```diff
@@ -155,6 +155,23 @@ module MAP
...
+
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V2) MAP                requires K1  ==K K2
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
```
daejunpark, line 170: so, you want to keep the new entry in the innermost?

k-distribution/include/builtin/domains.k
ehildenb: fixed

k-distribution/include/builtin/domains.k
ehildenb: I've commented out this rule because it could infinite loop.

ehildenb:
Re-opening as draft.

PR #361, opened 2019-03-14
BinaryLoader: lock parse cache file before saving the cache

denis-bogdanas:
Eliminates ObjectInputStream exceptions when several K instances run in parallel over same definition.

Tested on verified-smart-contracts.

Essentially only one line matters:
```
out.getChannel().lock();
```
Everything else is refactoring.

### DIFF
--- a/kernel/src/main/java/org/kframework/utils/BinaryLoader.java
+++ b/kernel/src/main/java/org/kframework/utils/BinaryLoader.java
@@ -16,7 +16,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.ObjectStreamException;
-import java.io.OutputStream;
+import java.nio.channels.FileLock;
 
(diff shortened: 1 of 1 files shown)


dwightguth (commented):
Have you tested this with multiple instances of kompile running concurrently that use --cache-file?

denis-bogdanas:
No, just multiple instances of kprove. Why would you run multiple instances of kompile over the same destination dir? If they run over different destinations, then each has its own cache file and there is no race condition.

dwightguth:
What I want to test is multiple different K definitions that compile to different directories but share the same cache file and have some files in common.

denis-bogdanas:
No I didn't test that but this fix should make them work properly. It doesn't matter who rewrites the cache after all - kprove or kompile.

dwightguth:
Okay, I will approve this and then try it once it's merged.

dwightguth (approved):

PR #363, opened 2019-03-14
Adds more reasoning for symbolic maps and some tests demonstrating desired behavior

ehildenb:
@yzhang90 and @daejunpark please add any tests you want to make sure the symbolic map passes to this PR. I figure this file can become our "specification of what symbolic maps in K mean".


java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinMapOperations.java
```diff
@@ -72,13 +72,16 @@ public static Term builtinMapLookup(Term map, Term key) {
...
-        if((map instanceof BuiltinMap) && ((BuiltinMap) map).isConcreteCollection() && key.isConcrete()) {
-            BuiltinMap.Builder builder = BuiltinMap.builder(context.global());
-            builder.put(key, value);
-            return updateAll(map, (BuiltinMap) builder.build(), context);
-        } else {
-            return  null;
+        if (map instanceof BuiltinMap) {
+            BuiltinMap bMap;
+            bMap = (BuiltinMap) map;
+            if((bMap.isConcreteCollection() && key.isConcrete()) || bMap.getEntries().containsKey(key)) {
```
daejunpark: @ehildenb `isConcreteCollection()` does not mean that all the map entries are concrete.  It just says that there is no other variables or abstractions, like `(1 |-> 2  M:Map)` or `(1 |-> 2  tree(X))`.

Thus, this allows `(X |-> 1) [2 <- 3]` to be evaluated to `(X |-> 1  2 |-> 3)`, which is wrong.

ehildenb: Ok, well I will come up with another fix, and make sure it passes/fails these tests correctly. Then will try out on KEVM master again and then ask you to try on KEVM Gnosis branch again.

daejunpark: I'll add more tests later. Need to sit down and spend time to come up with exhaustive test cases.

ehildenb: We don't need exhaustive tests. We just need any tests at all. Any test you want to write is a good test.

k-distribution/include/builtin/domains.k
```diff
@@ -155,6 +155,26 @@ module MAP
...
+  rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => V1         requires K1  ==K K2
+  rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
+
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V2) MAP                requires K1  ==K K2
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
+
+  rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP              [ K1 <- V2 ] requires K1  ==K K2
+
+  // potential infinite loop
+  // rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP [ K2 <- V2 ] [ K1 <- V1 ] requires K1 =/=K K2
```
daejunpark, line 175: https://github.com/runtimeverification/verified-smart-contracts/blob/master/gnosis/verification.k#L115-L125

daejunpark:
The following test scenario should be a good start.
(Note that `?=` means neither `==` nor `!=` can be proved.)

Here I consider the ideal symbolic map reasoning that tries to simplify a term as much as possible, i.e., reducing to a canonical term, at the cost of time overhead.

An efficient symbolic map reasoning may not fully simplify the term, but should be equivalent to the canonical term.

_EDIT: here I implicitly assume that all `x`, `y`, and `z` are not in `keys(M)`._

```
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 3                    when y == x /\ z == y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 2 z |-> 3            when y == x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 2 [z <- 3]           when y == x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 3 y |-> 2            when y != x /\ z != y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 y |-> 2 z |-> 3    when y != x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 y |-> 2 [z <- 3]   when y != x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 y |-> 3            when y != x /\ z == y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 y |-> 2 [z <- 3]   when y != x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 y |-> 2 [z <- 3]   when y != x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 3]           when y ?= x /\ z == y /\ z ?= x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 z |-> 3 [y <- 2]   when y ?= x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z ?= x

-------------------------------------------------------------------------------------------

M x |-> 1 [y <- 2] [z] => 2                                 when y == x /\ z == y /\ z == x
M x |-> 1 [y <- 2] [z] => M [z]                             when y == x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z] => M x |-> 2 [z]                     when y == x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z] => 1                                 when y != x /\ z != y /\ z == x
M x |-> 1 [y <- 2] [z] => M [z]                             when y != x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z] => M x |-> 1 [z]                     when y != x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z] => 2                                 when y != x /\ z == y /\ z != x
M x |-> 1 [y <- 2] [z] => M y |-> 2 [z]                     when y != x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z] => M x |-> 1 y |-> 2 [z]             when y != x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z] => 2                                 when y ?= x /\ z == y /\ z ?= x
M x |-> 1 [y <- 2] [z] => M [z]                             when y ?= x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z] => M x |-> 1 [z]                     when y ?= x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z == x
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z ?= x
```

daejunpark:
Another thing. I think we should use `krun` to test this behavior. `kprove` involves some implicit simplification over both LHS and RHS of the spec, and also does not structurally compare the output with RHS, so it's not adequate to test the function evaluation.

ehildenb:
@daejunpark I've tested as many of your rules as I can by translating them into `kprove` tests. I first tested them with no partial revert (so keeping @yzhang90 's #290), and the result was this: 3dbb676. But then, by importing `MAP-SYMBOLIC` from `domains.k`, it was able to prove more of them: 81961e3. Perhaps we can add enough lemmas to `MAP-SYMBOLIC` that they all pass.

I think we need to be able to turn off evaluation of the RHS before attempting matching, to make sure that the simplification happens to exactly the RHS and no further. You suggested doing this by running with `krun` instead of `kprove`, but I think it may be easier to just tell the prover "make sure it matches exactly what I wrote, and not some normal form of what I wrote".

We need to also need to find out why the negative proof cases are passing, because they probably shouldn't be.

k-distribution/tests/regression-new/map-symbolic-tests/assignment-11-spec.k.out
```diff
@@ -0,0 +1,4 @@
+<k>
+  assignmentResult ( Y |-> 2
+  ( MAP [ Z <- 3 ] ) )
```
daejunpark: the ideal output should be: `Y |-> 2 Z |-> 3 MAP` (or one with replacing Y with X).
the current output is not fully evaluated.
there would be many partially evaluated outputs.
so, your plan is to update this output if the next version changes the behavior and returns another form of the output?

k-distribution/tests/regression-new/map-symbolic-tests/assignment-16-spec.k.out
```diff
@@ -0,0 +1,4 @@
+<k>
+  assignmentResult ( ( X |-> 1
+  MAP ) [ Y <- 2 ] [ Z <- 3 ] )
+</k>
```
daejunpark: `M x |-> 1 z |-> 3 [y <- 2]` will be the fully evaluated term, for the record.

k-distribution/tests/regression-new/map-symbolic-tests/assignment-3-spec.k.out
```diff
@@ -0,0 +1,4 @@
+<k>
+  assignmentResult ( ( X |-> 1
+  ( MAP [ Y <- 2 ] ) ) [ Z <- 3 ] )
+</k>
```
daejunpark, line 4: not sure what ` X |-> 1 ( MAP [ Y <- 2 ] )` means.  will it be different from `( X |-> 1 MAP ) [ Y <- 2 ]`?  if so, not sure if it makes any problem.

btw, my expected idea output is `M x |-> 1 y |-> 2 [z <- 3]`.

daejunpark: hmm, now i can see that the term is generated by this rule:
```
rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
```

you may need to add more side condition like:
```
rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ])
     requires K1 =/=K K2 andBool K2 in keys(MAP)
```
which will require:
```
rule K1 in keys(K2 |-> _ MAP) => true when K1 ==K K2
rule K1 in keys(K2 |-> _ MAP) => K1 in keys(MAP) when K1 =/=K K2

rule K1 in keys(MAP [ K2 <- _ ]) => true when K1 ==K K2
rule K1 in keys(MAP [ K2 <- _ ]) => K1 in keys(MAP) when K1 =/=K K2
```

k-distribution/include/builtin/domains.k
daejunpark: you can avoid the infinite looping behavior by adding more side condition, like:
```
rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP [ K2 <- V2 ] [ K1 <- V1 ]
     requires K1 =/=K K2 andBool K2 in keys(MAP)
```
and have the following axioms:
```
rule K1 in keys(K2 |-> _ MAP) => true when K1 ==K K2
rule K1 in keys(K2 |-> _ MAP) => K1 in keys(MAP) when K1 =/=K K2

rule K1 in keys(MAP [ K2 <- _ ]) => true when K1 ==K K2
rule K1 in keys(MAP [ K2 <- _ ]) => K1 in keys(MAP) when K1 =/=K K2
```

k-distribution/include/builtin/domains.k
```diff
@@ -155,6 +155,26 @@ module MAP
...
+  rule .Map [ K1 <- V1 ] => K1 |-> V1
+
+  rule ((K1 |-> V1) MAP) [ K2 ] => V1         requires K1  ==K K2
+  rule ((K1 |-> V1) MAP) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
+
+  rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => V1         requires K1  ==K K2
+  rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
+
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V2) MAP                requires K1  ==K K2
+  rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
```
daejunpark, line 170: as explained below, i think it is better to add the additional side condition `K2 in keys(MAP)`.

k-distribution/tests/regression-new/map-symbolic-tests/failing/assignment-7-spec.k.out
```diff
@@ -0,0 +1,4 @@
+<k>
+  assignmentResult ( X |-> 1
+  ( MAP [ Z <- 3 ] ) )
+</k>
```
daejunpark: so, the current output is:
```
<k>
  assignmentResult ( X |-> 1
  ( MAP [ Z <- 2 ] [ Z <- 3 ] ) )
</k>
```

first, it is not unsound, right?

second, i cannot understand why `MAP [ Z <- 2 ] [ Z <- 3 ]` is not simplified to `MAP [ Z <- 3 ]`.  that is, why  the following rule is not applied?
```
rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP              [ K1 <- V2 ] requires K1  ==K K2
```

k-distribution/tests/regression-new/map-symbolic-tests/assignment-11-spec.k.out
ehildenb: This seems to imply that `MAP [ Z <- 3 ] => Z |-> 3 MAP` should be provable, but I don't think it should given the precondition?

k-distribution/tests/regression-new/map-symbolic-tests/assignment-16-spec.k.out
ehildenb: Why? We don't know that `notBool Z in_keys(MAP)` is `true`, so how can we get `MAP Z |-> 3`?

k-distribution/tests/regression-new/map-symbolic-tests/failing/assignment-7-spec.k.out
ehildenb: This is in the `failing` directory. These tests are not run on CI. They are there to document the fact that this test is broken with current Map reasoning.

k-distribution/tests/regression-new/map-symbolic-tests/assignment-16-spec.k.out
daejunpark: um, you're right.  i didn't think the case where `MAP` can be substituted by another map.

then, if there is no information about `MAP`, it is not possible to extend it, like `MAP  A |-> B`, which is the sound behavior.

k-distribution/tests/regression-new/map-symbolic-tests/failing/assignment-7-spec.k.out
daejunpark: i'm asking why this failed.  the current java backend with the MAP-SYMBOLIC module should pass this test, i think.

k-distribution/tests/regression-new/map-symbolic-tests/lookup-7-spec.k.out
```diff
@@ -0,0 +1,4 @@
+<k>
+  lookupResult ( X |-> 1
+  ( MAP [ Y <- 2 ] ) [ Z ] )
+</k>
```
daejunpark, line 4: `(X |-> 1 MAP) [Z]` is the fully evaluated term.

this should be obtained simply by adding the side condition to the rule, as suggested earlier:
```
rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
```

k-distribution/tests/regression-new/map-symbolic-tests/failing/lookup-8-spec.k.out
```diff
@@ -0,0 +1,3 @@
+<k>
+  lookupResult ( 2 )
+</k>
```
daejunpark, line 3: again, i don't understand why it is failed.

i mean, the current output is:
```
lookupResult ( MAP [ Z <- 2 ] [ Z ] )
```
which should have been simplified to `lookupResult ( 2 )`, by the following rule:
```
rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => V1         requires K1  ==K K2
```

do you think you can further look into it?

k-distribution/tests/regression-new/map-symbolic-tests/failing/lookup-9-spec.k.out
```diff
@@ -0,0 +1,3 @@
+<k>
+  lookupResult ( MAP [ Z ] )
+</k>
```
daejunpark: same here. i think this should have been passed.

the current term is
```
lookupResult ( MAP [ Y <- 2 ] [ Z ] )
```
which should have been simplified to the expected output, by the rule:
```
rule (MAP:Map [ K1 <- V1 ]) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2
```

daejunpark:
test scenario without the implicit assumption: (test spec id annotated)
```
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 3                    when y == x /\ z == y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 2 [z <- 3]           when y == x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 2 [z <- 3]           when y == x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 3 [y <- 2]           when y != x /\ z != y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y != x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y != x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 3]           when y != x /\ z == y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y != x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y != x /\ z ?= y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 3]           when y ?= x /\ z == y /\ z ?= x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z != y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z != y /\ z ?= x

M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z == x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z != x
M x |-> 1 [y <- 2] [z <- 3] => M x |-> 1 [y <- 2] [z <- 3]  when y ?= x /\ z ?= y /\ z ?= x

-------------------------------------------------------------------------------------------

M x |-> 1 [y <- 2] [z] => 2                                 when y == x /\ z == y /\ z == x  16
M x |-> 1 [y <- 2] [z] => M [z]                             when y == x /\ z != y /\ z != x   3
M x |-> 1 [y <- 2] [z] => M x |-> 2 [z]                     when y == x /\ z ?= y /\ z ?= x   1

M x |-> 1 [y <- 2] [z] => 1                                 when y != x /\ z != y /\ z == x  15
M x |-> 1 [y <- 2] [z] => M [z]                             when y != x /\ z != y /\ z != x   9
M x |-> 1 [y <- 2] [z] => M x |-> 1 [z]                     when y != x /\ z != y /\ z ?= x   7

M x |-> 1 [y <- 2] [z] => 2                                 when y != x /\ z == y /\ z != x   8
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y != x /\ z ?= y /\ z != x   6
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y != x /\ z ?= y /\ z ?= x   5

M x |-> 1 [y <- 2] [z] => 2                                 when y ?= x /\ z == y /\ z ?= x  12
M x |-> 1 [y <- 2] [z] => M [z]                             when y ?= x /\ z != y /\ z != x  14
M x |-> 1 [y <- 2] [z] => M x |-> 1 [z]                     when y ?= x /\ z != y /\ z ?= x  13

M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z == x  10
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z != x  11
M x |-> 1 [y <- 2] [z] => M x |-> 1 [y <- 2] [z]            when y ?= x /\ z ?= y /\ z ?= x   4
```

daejunpark:
@ehildenb do you think you can have another set of the 15 assignment tests where you provide the pre-condition saying that x, y, and z are not in the keys of the map?  such a test set will exercise more interesting behaviors.

daejunpark:
hmm, @ehildenb, now i realized that it is not trivial to add the `in_keys` side-condition to the other rules.  so i'll not ask you to complete the MAP-SYMBOLIC in this pr.

daejunpark:
looks good to me. but i'll let @denis-bogdanas to review the last commit.

daejunpark (approved):

denis-bogdanas:
@ehildenb @daejunpark PR is now rebased on top of 3 other PRs and still has issues. Do not merge until prerequisites are merged.

denis-bogdanas (changes requested):
Prerequisites must be merged first. They fix intermittent exceptions on Jenkins.

k-distribution/tests/regression-new/map-symbolic-tests/assignment-11-spec.k.out
ehildenb: Added tests with both `Z in_keys(MAP)` and `notBool Z in_keys(MAP)`, so we test both sides.

k-distribution/tests/regression-new/map-symbolic-tests/assignment-3-spec.k.out
ehildenb: Added tests for `in_keys`

k-distribution/include/builtin/domains.k
ehildenb: I've added tests with such side-condition to complement existing tests.

k-distribution/tests/regression-new/map-symbolic-tests/failing/lookup-8-spec.k.out
ehildenb: I'm not sure why it's failed either, but we have a record of it now, and can look into it in another PR.

k-distribution/tests/regression-new/map-symbolic-tests/failing/lookup-9-spec.k.out
ehildenb: For another PR.

k-distribution/tests/regression-new/map-symbolic-tests/lookup-7-spec.k.out
ehildenb: I've added tests which exercise both `in_keys` and `notBool in_keys` cases.

denis-bogdanas (approved):

PR #365, opened 2019-03-20
Added smt-hook attribute

traiansf:


traiansf:
I still need to check whether EVM proofs still work after this change, but I wanted to get some feedback on the PR in the meanwhile.

dwightguth (dismissed):
looks fine, but we should ask whether we want to use the smt-hook attribute for the functions in domains.k that are not actually z3 builtins.

traiansf:
I tried to only add it to functions which are z3 builtins.  But I agree, I wasn't sure about things which desugar to a formula;  if those don't generate declarations, maybe we should continue using smtlib.  Do you refer to such a thing?

dwightguth:
I'm referring to things like int_max and smt_seq_nil which are /not/ z3 builtins.

daejunpark:
@traiansf, @dwightguth is right.  Those are not z3 builtins, but they are already declared in the prelude file that is supposed to be given to the prover: https://github.com/kframework/k/blob/master/k-distribution/include/z3/basic.smt2

The thing is that in the smt query (at least for the text format), the order of declarations matters, and it is not trivial to implement the smt encoder to conform the order when generating a query.

You can still abuse the notation `smt-hook`, but not sure if that is what you want.

traiansf:
I don't mind removing them;  I was puzzled by the fact that they are in the `SMTLIB_BUILTIN_FUNCTIONS` list of the `KILtoSMTLib` class.  And I naturally assumed things in that list to be z3 builtins, because prelude defined things are excluded in a different way.  

Could you point to a place where I can extract a (complete) list of the current z3 builtins?

dwightguth:
https://stackoverflow.com/questions/38793301/list-of-all-built-in-symbols-in-z3

traiansf:
I think this is ready now, conforming  to our latest understanding of the matter.  However, I need an approval to merge.

daejunpark (approved):

PR #368, opened 2019-03-20
Fix bugs in K frontend compiler

dwightguth:
Fixes:

* issues with macro expansion when passing --non-strict or when variables are of sort K or when applying a macro to a ksequence.
* a bug with the compiler pass for function expressions.
* a too-vague error message in the ocaml backend.


k-distribution/tests/regression-new/equals-pattern/test.k
```diff
@@ -4,7 +4,9 @@ module TEST-SYNTAX
   imports INT
   imports K-EQUAL
 
-syntax KItem ::= foo(Int) | bar(Int) | baz(KItem) | stuff(K) | StringLiteral(K) | WStringLiteral(K)
+syntax KItem ::= foo(Int) | bar(Int) | baz(KItem) | stuff(K)
+
+syntax Contant ::= StringLiteral(K) | WStringLiteral(K)
```
traiansf: What does `Contant` stand for?  Did you mean `Constant`?

kernel/src/main/java/org/kframework/compile/ExpandMacros.java
```diff
@@ -220,10 +222,14 @@ public K apply(KVariable k) {
            for (int i = 0; i < kapp.items().size(); i++) {
               final int idx = i;
               Set<Sort> sorts = sort(kapp.items().get(idx), r);
-              prods.removeIf(p -> sorts.stream().noneMatch(s -> mod.subsorts().lessThanEq(s, p.nonterminal(idx).sort())));
+              if (!sorts.contains(null)) {
```
traiansf, line 225: When would `sorts` be `null` and what would this function do in that case?

traiansf (approved):

kernel/src/main/java/org/kframework/compile/ExpandMacros.java
dwightguth: Note that we return a sort of null if we examine the sort of a variable and it has no Sort attribute. This is done in order to make macro expansion work correctly when the user passes --non-strict to kompile. Basically, we're ignoring a sort check if the variable does not have a declared sort.

PR #371, opened 2019-03-24
Profiler2: made impliesSMTTimer non-static

denis-bogdanas:
Root cause of intermittent exceptions on Jenkins when using kprove+nailgun.


dwightguth (commented):
Why does this pr not introduce a new variable in GlobalContext?

denis-bogdanas:
Because `Profiler2` is a better place. It is `@RequestScoped` and all other stopwatches are there. And it makes more sense incapsulation-wise.

dwightguth:
Oh I see, I misread the pr.

dwightguth (approved):

PR #372, opened 2019-03-26
Java Backend logging: Option --log-success-pc-diff & better ConjunctiveFormula logging

denis-bogdanas:
1. Option --log-success-pc-diff to print new PC entries for success final states.
2. When ConjunctiveFormula is logged, group separately entries from initial state and new entries.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -90,9 +90,10 @@ public class SymbolicRewriter {
 
     public RewriterResult rewrite(ConstrainedTerm constrainedTerm, int bound) {
         stopwatch.start();
+        ConstrainedTerm initTerm = constrainedTerm;
         int step = 0;
         List<ConstrainedTerm> results;
-        while (step != bound && !(results = computeRewriteStep(constrainedTerm, step, true)).isEmpty()) {
+        while (step != bound && !(results = computeRewriteStep(constrainedTerm, step, true, initTerm)).isEmpty()) {
             /* get the first solution */
             constrainedTerm = results.get(0);
             step++;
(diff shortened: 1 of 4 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -154,7 +156,9 @@ public RewriterResult rewrite(ConstrainedTerm constrainedTerm, int bound) {
 
     }
 
-    public List<ConstrainedTerm> fastComputeRewriteStep(ConstrainedTerm subject, boolean computeOne, boolean narrowing, boolean proofFlag, int step) {
+    public List<ConstrainedTerm> fastComputeRewriteStep(ConstrainedTerm subject, boolean computeOne, boolean narrowing,
+                                                        boolean proofFlag, int step,
+                                                        ConstrainedTerm initTerm) {
```
yzhang90: Not sure if it is a good idea to add new argument (only for logging purpose) to the fastComputeRewriteStep function. Can we call the logging function after ``List<ConstrainedTerm> results = computeRewriteStep(term, step, false)``, can traverse the list to print the constraint diff?

denis-bogdanas: This would degrade the usability of logs. First, `fastComputeRewriteStep` is called from 3 places. Most important is here: https://github.com/kframework/k/blob/c98afd1439d4b50ad6133ebcc51cdf1546a43104/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java#L722-L723

Second, logs are most useful when they are printed just when the event that triggered them happened. In the current implementation, I can look at the logs generated before `New top-level constraint` to see what triggered new constraint. If `New constraint` logs would be printed later, it would be harder to make connection with rules and z3 queries preceding `New constraint` event. I'll add a comment to the argument to make it clear it's only used for logging.

yzhang90: ok

yzhang90 (approved):
Does PR require @dwightguth approval?

denis-bogdanas:
No I don't think we need another review. Dwight is mostly interested in changes outside Java Backend. This is mostly a re-formatting of existing log messages.

dwightguth:
That's more or less correct; I am not concerned with changes inside the Java backend except insofar as we should make sure that there is not a more generic way to do the same thing. That said, logging in particular is essentially impossible to do in a backend-independent way, so if your code only touches the backend and only deals with new logging features, you should not feel like you need my review.

PR #373, opened 2019-03-27
Added refresh rules implementation and owise test. CI should fail

traiansf:


dwightguth (commented):
I don't see how you are making sure different rules are normalized to different values. Am I missing something?

kernel/src/main/java/org/kframework/compile/RefreshRules.java
```diff
@@ -0,0 +1,81 @@
...
+    private Rule applyRefresh(Rule rule) {
+        return Rule(
+                applyRefresh(rule.body()),
+                applyRefresh(rule.requires()),
+                applyRefresh(rule.ensures()),
+                rule.att());
+    }
+
+    private Rule applyRefreshResetVars(Rule rule) {
+        vars.clear();
```
traiansf, line 59: @dwightguth Wrt. your question:  note that `refresh` is called for a collection of rules, and in between them only the `vars` map  is cleared, while the counter keeps growing.

dwightguth: Oh you're right, I missed that.

dwightguth (approved):

PR #374, opened 2019-03-27
Fixed dry-run issue with kprove on Haskell Backend

traiansf:


dwightguth (commented):
Can we just return early on dry run instead of making the indentation level bigger?

traiansf:
> Can we just return early on dry run instead of making the indentation level bigger?

Done that. I also did it to the execute and search methods. nothing else changed. please re-review.

dwightguth (approved):

PR #375, opened 2019-03-28
java-backend/FastRuleMatcher: more informative error message when crashing on map unification

ehildenb:
Instead of crashing when a map has unevaluated function symbols, just fail unification.

Need to double-check that this won't cause unsound behavior though.


daejunpark (approved):

daejunpark:
we already discussed about this code change.  please check if this doesn't break the vsc ci.

ehildenb:
Passes VSC: https://github.com/runtimeverification/verified-smart-contracts/pull/209

PR #376, opened 2019-03-29
Attempt to switch several `smtlib` attributes to `smt-hook`.

ehildenb:
The questionable ones are in the floating point module, where there are mixed declarations like `(not (== #1 #2))`. I have changed all uses of `==` (uninterpreted function) to `=` (builtin z3 equality), because Z3 seems happy enough to consume floating points under `=`.

There is an example of floating point verification in KernelC, at `k-distribution/samples/java_rewrite_engine/kernel-cil/proofs/cd2d/cd2d-spec.k`.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -470,27 +470,30 @@ module FLOAT
   imports BOOL
   imports INT-SYNTAX
 
+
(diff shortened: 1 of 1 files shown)


dwightguth (approved):

dwightguth:
Looks good. Not sure why it says you didn't sign the cla (maybe because the email address on the commits is wrong) but I'll merge anyway.

PR #377, opened 2019-03-31
kore: Att: removed lazy hashCode

denis-bogdanas:
For some reason it was computed incorrectly, as if Att would be a mutable object. This interfered with `Map.contains(Att)` algorithm and led to unnecessary parse cache invalidation.

### Effect
Performance impact is insignificant, but who knows might be important later/elsewhere.

Before, when running kprove sum-to-n for the first time after clean kompile:
```
Parse spec modules [1035/1037 rules]                         =  0.020s
```
after:
```
Parse spec modules [186/1037 rules]                          =  0.021s
```


dwightguth:
Let's fix this rather than just removing it. Att can be incredibly performance sensitive because we spawn tens of thousands of instances of it and call its methods a similar number of times. No reason not to be cautious.

PR #378, opened 2019-04-01
kompile flag to generate time to parse each rule

dwightguth:


kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -70,6 +70,9 @@ public String syntaxModule(FileUtil files) {
     @Parameter(names="--coverage", description="Generate coverage data when executing semantics.")
     public boolean coverage;
 
+    @Parameter(names="--profile-rule-parsing", description="Generate time in seconds to parse each rule in the semantics. Found in -kompiled directory under timing.log.")
+    public boolean profileRules;
```
ehildenb, line 74: Should we be setting this to `false` by default?

ehildenb (approved):
Looks good, one minor question.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
dwightguth: primitive fields in java are initialized to the default value for that type which is false in the case of booleans. so no, it's not required.

PR #380, opened 2019-04-02
Latex serializer/output mode

ehildenb:
Adds output mode `--output latex`, which serializes the AST to LaTeX.

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/ToLatex.java
+++ b/kernel/src/main/java/org/kframework/unparser/ToLatex.java
@@ -33,6 +33,10 @@ public class ToLatex {
         }
     }
 
+    public static String latexedKLabel(String orig) {
+        return orig.replaceAll("[^a-zA-Z]", "");
+    }
+
     public static void apply(DataOutputStream out, K k) throws IOException {
         if (k instanceof KToken) {
             KToken tok = (KToken) k;
@@ -41,7 +45,7 @@ public class ToLatex {
         } else if (k instanceof KApply) {
             KApply app = (KApply) k;
 
-            out.write(("\\" + app.klabel().name()).getBytes(StandardCharsets.UTF_8));
+            out.write(latexedKLabel("\\" + app.klabel().name()).getBytes(StandardCharsets.UTF_8));
 
             for (K item : app.klist().asIterable()) {
                 out.write("{".getBytes(StandardCharsets.UTF_8));
--- a/kernel/src/main/java/org/kframework/unparser/ToLatex.java
+++ b/kernel/src/main/java/org/kframework/unparser/ToLatex.java
@@ -14,6 +14,7 @@ import org.kframework.utils.errorsystem.KEMException;
 import java.io.IOException;
 import java.io.DataOutputStream;
 import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
 
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            apply(new DataOutputStream(out), k);
+            return out.toByteArray();
+        } catch (IOException e) {
+            throw KEMException.criticalError("Could not write K term to LaTeX", e, k);
+        }
+    }
+
+    public static String latexedKLabel(String orig) {
+        return "klabel" + orig.replaceAll("[^a-zA-Z]", "");
```
dwightguth: this is not great because different klabels could be rewritten to the same latex macro. Take a look at how we deal with encoding of klabels in the ocaml and kore backends, then look at what characters are allowed in latex macro names and go from there.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            apply(new DataOutputStream(out), sent);
+            return out.toByteArray();
+        } catch (IOException e) {
+            throw KEMException.criticalError("Could not write K term to LaTeX", e, sent);
+        }
+    }
+
+    public static void apply(DataOutputStream out, Sentence sent) throws IOException {
+        if (sent instanceof Rule) {
```
dwightguth: it would be nice if we used polymorphism for this instead of a big if/else block.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            writeString(out, "\\outerModuleComment{" + moduleComment.comment() + "}{");
+            apply(out, moduleComment.att());
+            writeString(out, "}");
+
+        } else {
+            KEMException.criticalError("Do not know how to serialize sentence: " + sent.toString());
+        }
+    }
+
+    public static void apply(DataOutputStream out, K k) throws IOException {
```
dwightguth, line 73: let's use a visitor for this.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+        } else if (k instanceof KSequence) {
+            KSequence kseq = (KSequence) k;
+
+            writeString(out, "\\klabelKSequence{");
+
+            for (K item : kseq.asIterable()) {
+                apply(out, item);
+                writeString(out, "}{\\klabelKSequence{");
```
dwightguth: let's call this kseq and dotk to align with kore generation.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            if (origName.isPresent()) {
+                writeString(out, origName.get());
+            } else {
+                writeString(out, var.name());
+            }
+
+        } else if (k instanceof KRewrite) {
+            KRewrite rew = (KRewrite) k;
+
+            writeString(out, "\\klabelKRewrite{");
```
dwightguth: can this just be \krewrites?

kernel/src/main/java/org/kframework/unparser/ToLatex.java
```diff
@@ -0,0 +1,208 @@
...
+            writeString(out, "}{");
+            apply(out, alias.alias());
+            writeString(out, "}{");
+            apply(out, alias.att());
+            writeString(out, "}");
+
+        } else if (k instanceof InjectedKLabel) {
+            InjectedKLabel inj = (InjectedKLabel) k;
+
+            writeString(out, "\\injectedKLabels{");
```
dwightguth: why s?

dwightguth (changes requested):
looks okay to start with. there's some comments I have about how to improve it. But the main comment is that we should make use of the latex attribute on KApply if it exists.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: Taken care of.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: I'm pulling out the `Sentence` and `Module` serializers from this PR, only including the `K` one for now.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: Done.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: Done.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: changed to `injectedklabel`.

kernel/src/main/java/org/kframework/unparser/ToLatex.java
ehildenb: I modeled this one off of the Binary/JSON serializers, and would prefer to keep the same structure as those ones here (then if desired, refactor them all in a later PR to abstract out the common details).

ehildenb:
@dwightguth for serializing k terms, I don't plan to use the `latex` attribute, but to just emit the klabel names. The idea is that as another pass you will tell K to emit the latex prelude for you, which will add in all the `newcommand` definitions needed to make the definitions valid. If you don't supply a `latex` attribute, it will generate some sensible `newcommand` for you, but if you do supply it it will directly use that.

So, given the following definition:

```
syntax Exp ::= Int | Bool
            | Exp "+" Exp
            | Exp "<=" Exp [latex(#1 \le #2)]
```

So if you run something like `krun --gen-latex-prelude`, it will produce the following latex prelude:

```
\newcommand{\klabelPlus}[2]{#1 + #2}            // auto-generated based on the production
\newcommand{\klabelLTEQ}[2]{#1 \le #2}        // use the user-supplied one
```

And you serialize the term `3 + 4 <= 8` with `--output latex`, it will produce (without the prelude):

```
\klabelLTEQ{\klabelPlus{3}{4}}{8}
```

This means that `--output latex` can be a strict serializer to abstract LaTeX ASTs, and leaves all the decisions about how to display things to the `newcommand` declarations. This will make it much easier to uniformly fiddle with the resulting display of the document when doing final touches by just changing the generated `\newcommand` declarations.

dwightguth:
Oh I see. That makes sense, actually. This is good enough for now, then. We can improve it if it becomes necessary.

dwightguth (approved):

PR #383, opened 2019-04-02
try a new way to colorize build log

dwightguth:


ehildenb (approved):
If it works and does what you want, looks good to me.

PR #385, opened 2019-04-02
build and deploy k on arch linux

dwightguth:


ehildenb (approved):
Tested that I could build and install it on a clean Arch system.

PR #388, opened 2019-04-02
Bug fixes for LLVM Backend

dwightguth:
I'm going to try to release a PR every day with the smaller fixes for the day rather than letting it clump up so much, in the future, but this is what has built up so far. Let me know if there's specific things you'd like to see given regression tests

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
+++ b/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
@@ -130,7 +130,7 @@ public class AddSortInjections {
             }
         } else {
             String hookAtt = mod.sortAttributesFor().get(expectedSort).getOrElse(() -> Att()).getOptional("hook").orElse("");
-            if (hookAtt.equals("MAP.Map") || hookAtt.equals("SET.Set") || hookAtt.equals("LIST.List")) {
+            if ((hookAtt.equals("MAP.Map") || hookAtt.equals("SET.Set") || hookAtt.equals("LIST.List")) && term instanceof KApply) {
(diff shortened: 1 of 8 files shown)


traiansf:
I was wondering whether we could have Jenkins run all Haskell backend tests whenever there is a change to ModuleToKore or to the supporting transformations (such as AddSortInjections).  This part is known to be fragile, so making it sure it passes tests would be better done now (since we know we've touch potentially sensitive code) rather than waiting for @ehildenb 's job to do it one in a while.

Other than that, the changes look reasonable to me.

traiansf:
```
 make[1]: *** [or-llvm] Error 2
```

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -147,7 +147,7 @@ private K internalAddSortInjections(K term, Sort expectedSort) {
                         }
                     }
                 }
-                throw new AssertionError();
+                return KApply(KLabel("inj", actualSort, expectedSort), KList(visitChildren(term, actualSort)), Att.empty().add(Sort.class, expectedSort));
             } else {
                 return KApply(KLabel("inj", actualSort, expectedSort), KList(visitChildren(term, actualSort)), Att.empty().add(Sort.class, expectedSort));
```
ehildenb: I think we can drop the `else` branch and hoist these identical `return KApply` to after the `if_else`.

ehildenb (approved):
Overall looks pretty good, one nit you could address, then you can merge. Approving now for speeeeed

dwightguth:
@traiansf I tested on the haskell integration tests by running scripts/ktest.sh and everything passed.

PR #391, opened 2019-04-03
FFI Module

mickyabir:
Contains the development of the FFI module


k-distribution/include/builtin/ffi.k
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015-2019 K Team. All Rights Reserved.
```
dwightguth: this should be Copyright (c) 2019

k-distribution/include/builtin/ffi.k
```diff
@@ -0,0 +1,18 @@
...
+require "domains.k"
+
+module FFI-SYNTAX
+  imports INT-SYNTAX
+  imports FLOAT-SYNTAX
+
+  syntax FFIType ::= Int | Float | "#void" [klabel(#void), symbol]
```
dwightguth: This isn't really what you want. You have made "0" and "0.0" instances of FFIType instead of making "Int" and "Float". Additionally, take a look at http://www.chiark.greenend.org.uk/doc/libffi-dev/html/Primitive-Types.html#Primitive-Types for the actual list of types. As I've said elsewhere, exclude pointers and structs for now.

k-distribution/include/builtin/ffi.k
```diff
@@ -0,0 +1,18 @@
...
+endmodule
+
+module FFI
+  imports FFI-SYNTAX
+  imports BYTES
+  imports LIST
+
+  /*@ The FFI call takes a list of bytes, a list of types, and a return type. */
+  syntax Bytes ::= "#ffiCall" "(" List "," List "," FFIType ")" [function, hook(FFI.call), impure]
```
dwightguth: I may not have given you the right info last time, but this also needs a String with the name of the symbol to call.

dwightguth (changes requested):
A good start, but there's a few changes that need to be made.

dwightguth (approved):

k-distribution/include/builtin/ffi.k
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2019 K Team. All Rights Reserved.
+require "domains.k"
+
+module FFI-SYNTAX
+  imports INT-SYNTAX
```
dwightguth: I don't think we need these imports anymore, do we?

dwightguth (changes requested):

dwightguth:
One other thing that would be good would be a test that requires this file just to make sure there are no parsing errors.

dwightguth:
I'm not sure you should test this on the llvm backend yet. I suspect that the test will fail to link. Just use the ocaml backend for now and we'll update it once we implement the hooks

dwightguth (approved):

Issue #389, opened 2019-04-02
maybe link to this Windows install walk-through


DustinWehr:
This is for installing K starting with a fresh Windows Subsystem for Linux install. Written April 2019 in/for Windows 10 Pro 64-bit. I'm a modern-Windows newb , and I wrote this for people who are as prone to installation frustration as I am. 

Install the Ubuntu package from the Windows Store, which as of now is an alias for the Ubuntu LTS 18.04 package. 
I downloaded kframework_5.0.0_amd64_bionic.deb to C:\linux.
From the windows command prompt:
`ubuntu`
It'll take a few minutes to install, then you'll make a user account, and then you're in a linux shell.
You need to download package source data before you can install anything. I first tried
`sudo apt update`
but that reported a couple errors, so I tried again with
`sudo apt-get update`
and that completed with no errors. Lastly:
`sudo apt-get install /mnt/c/linux/kframework_5.0.0_amd64_bionic.deb`

That installed 1.4GB of dependencies, some of which is basic dev stuff, since it's a fresh ubuntu install, and took a long time for me. It also, apparently, compiles OCaml. 

Some instructions beginning with `1. To configure OPAM in the current shell session` showed up during the installation, and I've ignored them for now. There was also a worrying part where OPAM asked if it could modify ~/.profile and ~/.ocamlinit, and apt-get apparently answered no for me. I haven't done anything about that.

I ended up with a k installation at /usr/lib/kframework with everything owned by root, so I did:
`sudo chown -R dustin:dustin /usr/lib/kframework`

I added `export PATH=$PATH:/usr/lib/kframework/bin` to my ~/.bashrc. That was enough for the test suggested in K's INSTALL.md to work:
`source ~/.bashrc`
`make /usr/lib/kframework/tutorial/2_languages/1_simple/1_untyped/`
`krun make /usr/lib/kframework/tutorial/2_languages/1_simple/1_untyped/tests/diverse/factorial.simple`

dwightguth:
You don't actually need to add anything to your path. It creates scripts that invoke everything in /usr/bin, like a normal ubuntu package. There are a few obscure scripts we don't create links for, but none of them are commonly used. If you end up needing them, you can update the path like you said.

K also works just fine with everything owned by root, and you can just copy the tutorial directory into your home directory and build everything with make or kompile.

[...]

PR #392, opened 2019-04-04
readme: Guide for installing on fresh Windows Subsystem for Linux

denis-bogdanas:


ehildenb:
@DustinWehr ping with regards to #389

ehildenb (approved):
I'm approving, would like Dustin's feedback to make sure it works for him too, but I don't think we should block merging on it.

denis-bogdanas:
I modified his instructions for what worked for me. There are some differences. I didn't add instructions creating a new user. Just for running with root.

DustinWehr:
Looks good! An issue for me with running as root is getting 
> [WARNING] Running as root is not recommended

with every use of kompile, which is worrisome. It led me to googling about WSL security, which took a lot longer than simply changing the permissions. Whether it actually helps or not, it was the shortest path to obtaining the minimal comfort I require.

denis-bogdanas:
@ehildenb Are you ok merging like this? I guess there are cleaner ways to install K on non-root user than what's in Dustin's guide. But this is a good enough starter.

I'd rather not spend time on verifying how to install on a different user.

ehildenb:
Let's just add a notice at the beginning which says "These instructions are for root user, make sure to create a non-root user to run as if you care not to do that" or something.

denis-bogdanas:
@ehildenb I added a new point with links to guides how to create a new user. Adding that sentence at the beginning might get confusing.

README.md
```diff
@@ -95,6 +95,27 @@ which will significantly speed up the incremental build process.
 
 After running `mvn package` for the first time, you should run ``k-distribution/target/release/k/bin/k-configure-opam; eval `opam config env` ``. This performs first-time setup of the OCAML backend.
 
+## Installing on fresh Windows Subsystem for Linux
+
+1. Install the Ubuntu package from the Windows Store, which as of now is an alias for the Ubuntu LTS 18.04 package. 
+2. Download the latest Ubuntu Bionic distribution from https://github.com/kframework/k/releases 
+    to a temporary directory, for example `d:\temp`
+    
+3. Optional. By default the WSL bash user is root. It is recommended to run K as a non-root user, otherwise K will print warnings for each executed command.
```
dwightguth: the first sentence here is not true.

README.md
```diff
@@ -95,6 +95,27 @@ which will significantly speed up the incremental build process.
...
+
+4. Open linux bash. 
+5. Run the following commands:
+    $ `sudo apt-get update`
+    $ `cd <download dir>`. In our example it is `cd /mnt/d/temp`
+    $ `sudo apt-get install ./kframework_5.0.0_amd64_bionic.deb`
+        This will install ~1.4GB of dependencies and will take some time.
+        F will be installed to `/usr/lib/kframework`
+6. Now you can try to run some programs:
+    $ `cd /usr/lib/kframework/tutorial/2_languages/1_simple/1_untyped`
```
dwightguth: this will not work as this directory will be owned by root. You need to tell people to `cp -R /usr/lib/kframework/tutorial ~` and then tell them to build the version of the tutorial they copied into their home directory.

dwightguth (changes requested):

denis-bogdanas:
@dwightguth fixed. Added the command to change owner, like Dustin din.

dwightguth:
As I said, the default WSL user is /not/ root and we should not be telling people that because it will just confuse them. Nor should we be telling them to chown the installation as that is not necessary.

denis-bogdanas:
@dwightguth The way I installed it today, from `WSLTools` utility, it is root. But if one uses the install wizard from windows store, it probably prompts to create a new user. So you suggest deleting the whole part about creating/changing user?

I'll change the other part.

dwightguth:
We should definitely give people the official Microsoft-supported instructions for installing WSL.

denis-bogdanas:
done

README.md
```diff
@@ -95,6 +95,42 @@ which will significantly speed up the incremental build process.
...
+examples from default installation dir if you are not `root`:
+
+```    
+$ cp -R /usr/lib/kframework/tutorial /mnt/d/k-tutorial
+```
+
+6. Now you can try to run some programs:
+
+```bash
+$ /mnt/d/k-tutorial/2_languages/1_simple/1_untyped
```
dwightguth: I think you meant for there to be a cd here.

dwightguth (approved):
looks good but one minor thing to fix before you merge.

README.md
denis-bogdanas: done

README.md
```diff
@@ -95,6 +95,42 @@ which will significantly speed up the incremental build process.
...
+
+    `$ sudo apt-get update`
+    
+    `$ cd <download dir>`. In our example download dir is `/mnt/d/temp`
+    
+    `$ sudo apt-get install ./kframework_5.0.0_amd64_bionic.deb`
+        This will install ~1.4GB of dependencies and will take some time.
+        K will be installed to `/usr/lib/kframework`
+
+5. Copy the tutorial to some work directory, for example `/mnt/d/k-tutorial`. Otherwise, you won't be able to run the
```
DustinWehr, line 119: Heh, ah, yeah that's better than chown.

PR #393, opened 2019-04-04
Strategy language updates

ehildenb:
Includes:

-   [BugFix] No longer forget the current term constraint when applying "secret" `#STUCK()` rule.
-   `#STUCK()` is defined as a KLabel instead of an Attribute.
-   Add file `builtin/Rules.java` which is analogous to `builtin/KLabels.java` for "secret"/"magic" rules (eg. the `#STUCK()` rule).
-   Emit a `STUCK_RULE` event on StateLog when that rule is applied.
-   Replace `BASIC-STRATEGY` with `STRATEGY-ABSTRACT`, separate out `RULE-TAG-SYNTAX` to make parsing definitions with `#RuleTag` variables easier.


dwightguth:
If STRATEGY doesn't import RULE-TAG-SYNTAX, DEFAULT-STRATEGY definitely needs to, or you /will/ break things.

dwightguth (approved):

PR #394, opened 2019-04-04
Factored identifier encoding

ehildenb:
Makes some common machinery for escaping identifiers to various encoding formats.


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1186,136 +1186,17 @@ private void convert(Map<String, Boolean> attributes, Att att) {
...
-            "y",// 79
-            "z",// 7a
-            "LBra",// 7b
-            "Pipe",// 7c
-            "RBra",// 7d
-            "Tild",// 7e
-            null// 7f
-    };
+    private static String[] asciiReadableEncodingKoreCalc() {
+        String[] koreEncoder = StringUtil.asciiReadableEncodingDefault;
```
dwightguth: are you sure this is actually doing a copy? it could just be assigning a pointer to the object, which would mean you're overwriting the default. You might want to use Arrays.copyOf.

ehildenb: Done.

dwightguth (approved):

PR #395, opened 2019-04-04
generate explicit projections for outer casts 

dwightguth:


traiansf (approved):
Looks reasonable to me, and the change is for the better, so I agree with it.

PR #396, opened 2019-04-04
don't emit no confusion for ml syntax

dwightguth:
This fixes a bug introduced by importing ML-SYNTAX into BASIC-K.


traiansf (approved):

ehildenb (approved):
I suppose you've tested that this plays nicely with the haskell backend.

dwightguth:
I mean, I have tested that it fixes a bug where the kore generated by kompile didn't validate in the version of the haskell backend used by the llvm backend, anyway.

PR #397, opened 2019-04-05
Java backend engine bugfix: Resetting constraint at the beginning of each step

denis-bogdanas:
Otherwise stale constraint from a previous operation might be present.

Previously this lead to function rules being evaluated in the context of another branch. This could cause false specs getting proved, or the other way around, in the presence of branching.


denis-bogdanas:
VSC passes: https://github.com/runtimeverification/verified-smart-contracts/pull/214
(look on Jenkins page for details)

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -653,6 +653,7 @@ private K disjunctResults(List<K> results) {
                 boolean alreadyLogged = false;
                 try {
                     v++;
+                    term.termContext().setTopConstraint(null); //To remove leftover constraint from previous step
```
ehildenb, line 656: Down on line 718/768, we're doing `nextQueue.add(result)`, where `result` is a `ConstrainedTerm`. So `result` could already have a constraint which we should be keeping, will this throw that constraint away? Or is that constraint different than the one in `_.termContext()`?

denis-bogdanas: It's different. ConstrainedTerm is immutable. While TermContaxt is just a bag of mutable fields. All of them should be evaluated, and (ideally) either replaced with parameters or made immutable. Or at lest documented.

ehildenb:
@daejunpark can you approve this one? It looks good to me, but I'm less confident about the proving code and would like one more set of eyes on it. Otherwise, looks good to me.

daejunpark (approved):

MrChico:
Thanks for the fix! :)

PR #398, opened 2019-04-05
Implemented injection compaction for the KORE rewrite rule generator.

theo25:
This PR fixes a bug where a nested injection could be generated in the output KORE when the input definition contained an injection that contains a rewrite that contains an injection. The fix compacts such nested injections using the injection transitivity axiom.


kore/src/main/scala/org/kframework/compile/RewriteToTop.scala
```diff
@@ -16,7 +17,7 @@ object RewriteToTop {
 
   def toRight(rewrite: K): K = rewrite match {
     case t: KRewrite => toRight(t.right) // recurse here because of KAs
-    case t: KApply => KApply(t.klabel, immutable(t.klist.items) map toRight, t.att)
+    case t: KApply => compactInjections(KApply(t.klabel, immutable(t.klist.items) map toRight, t.att))
```
dwightguth, line 20: can we do this on toLeft too?

dwightguth (changes requested):

kore/src/main/scala/org/kframework/compile/RewriteToTop.scala
theo25: Done

dwightguth (approved):

PR #399, opened 2019-04-08
pass --smt none to simple

dwightguth:
PR #390 intermittently fails on simple on mac os due to a z3 version issue. See https://office.runtimeverification.com/jenkins/blue/rest/organizations/jenkins/pipelines/k/branches/PR-393/runs/4/nodes/27/steps/65/log/?start=0

It's because of the int_min z3 declaration issue. Since this is concrete execution, we just pass --smt none.


denis-bogdanas:
@dwightguth Could you check the Z3 version on mac os? It should be `4.6.0`. This is the most likely cause.

daejunpark (approved):

PR #402, opened 2019-04-09
Java Backend log: --log-func-target, more KItem and global logs

denis-bogdanas:
Main new features:
- option --log-func-target that prints every KItem that reached function evaluation, along with result of evaluation costomzied by exit point.
- printing mode details of what's going on inside function evaluation.
- printing more details of whats going on as part of one global step. Distinguishing final implication phase, apply spec rule phase and multiple sub-phases as part of applying a regular rule.

Example log fragment:
```
KItem lvl 1,     starting evaluation: #if_#then_#else_#fi_K-EQUAL
 -------------
 KItem lvl 1,  matched, eval. constr.: #if_#then_#else_#fi_K-EQUAL, source: Source(/mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k) Location(829,8,829,67)

  KItem lvl 2,     starting evaluation: notBool_
  KItem lvl 2,      no rule applicable: notBool_(_<Int__INT-COMMON(X_81:Int,, Int(#"0")))

  Z3 Implication (Function rule implication) failed (cached result):
  ConjunctiveFormula(
    equalities:
      _==K_(_<Int__INT-COMMON(X_81:Int,, Int(#"0")),, Bool(#"true"))
  )
    implies
  ConjunctiveFormula(
    equalities:
      _==K_(_<Int__INT-COMMON(X_81:Int,, Int(#"0")),, Bool(#"false"))
  )
  
  Rule for formula above:
    rule #if C:Bool #then _ #else B2::K #fi => B2 requires notBool C
  	Source: /mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k Location(829,8,829,67)
  -------------
 -------------
 KItem lvl 1,  matched, eval. constr.: #if_#then_#else_#fi_K-EQUAL, source: Source(/mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k) Location(828,8,828,59)

 KItem lvl 1,  function rule applying: #if_#then_#else_#fi_K-EQUAL, source: Source(/mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k) Location(828,8,828,59)

 KItem lvl 1,            rule applied: #if_#then_#else_#fi_K-EQUAL
  rule #if C:Bool #then B1::K #else _ #fi => B1 requires C
	Source: /mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k Location(828,8,828,59)
KItem lvl 1,               evaluated: #if_#then_#else_#fi_K-EQUAL(_<Int__INT-COMMON(X_81:Int,, Int(#"0")),, Int(#"2"),, Int(#"1"))
                                  to: Int(#"2")
```

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
@@ -597,7 +597,9 @@ public class KItem extends Term implements KItemRepresentation {
 
                     if (result != null) {
                         if (kItem.global.javaExecutionOptions.logFunctionTargetPublic) {
-                            System.err.format("KItem evaluated: %s\n", kItem);
+                            System.err.format(""
+                                    + "KItem evaluated: %s\n"
+                                    + "             to: %s\n", kItem, result);
                         }
                         kItem.profiler.evalFuncRuleCounter.increment();
                         return result;
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
@@ -426,6 +426,7 @@ public class KItem extends Term implements KItemRepresentation {
 
             try {
                 KList kList = (KList) kItem.kList;
+                int nestingLevel = kItem.profiler.resFuncNanoTimer.getLevel();
 
                 if (builtins.get().isBuiltinKLabel(kLabelConstant)) {
                     try {
@@ -433,9 +434,9 @@ public class KItem extends Term implements KItemRepresentation {
                         Term result = builtins.get().invoke(context, kLabelConstant, arguments);
                         if (result != null && !result.equals(kItem)) {
                             Term evalResult = result.evaluate(context);
-                            if (evalResult != result && context.global().javaExecutionOptions.logRulesPublic) {
-                                System.err.format("\nEvaluated builtin KLabel, lvl %d\n" +
-                                        "-------------------------\n", kItem.profiler.resFuncNanoTimer.getLevel());
+                            if (context.global().javaExecutionOptions.logRulesPublic) {
+                                System.err.format("\nKItem lvl %d, %23s: %s\n",
+                                        nestingLevel, "builtin evaluation", kLabelConstant);
                             }
                             kItem.profiler.evalFuncBuiltinCounter.increment();
                             return evalResult;
@@ -480,6 +481,13 @@ public class KItem extends Term implements KItemRepresentation {
                     Term result = null;
                     Term owiseResult = null;
                     Rule appliedRule = null;
+                    if (kItem.global.javaExecutionOptions.logRulesPublic) {
+                        if (nestingLevel == 1) {
+                            System.err.print("-------------------------\n");
+                        }
+                        System.err.format("KItem lvl %d, %23s: %s\n",
+                                nestingLevel, "starting evaluation", kLabelConstant);
+                    }
 
                     // an argument is concrete if it doesn't contain variables or unresolved functions
                     boolean isConcrete = kList.getContents().stream().filter(elem -> !elem.isGround() || !elem.isNormal()).collect(Collectors.toList()).isEmpty();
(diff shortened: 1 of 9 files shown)


java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -582,13 +587,16 @@ private Term evaluateFunction(KItem kItem, TermContext context) {
...
                     if (result != null) {
+                        KItemLog.logEvaluated(kItem, result, nestingLevel);
                         kItem.profiler.evalFuncRuleCounter.increment();
                         return result;
                     } else if (owiseResult != null) {
+                        boolean anyRegRuleUnify = false;
                         if (!kItem.isGround()) {
                             if (context.global().stage != Stage.REWRITING) {
-                                return kItem;
+                                break processMatchResult;
```
daejunpark: not sure what's going on here. what is the semantics of this break here?

denis-bogdanas: It is a refactoring.
Approximate picture:
Before:
```
if (foo) {
  return kItem;
} else {
  stuff;
  return result;
}
return kItem;
```
After:
```
ifLabel:
if (foo) {
  break ifLabel;
} else {
  stuff;
  return result;
}
log(...);
return kItem;
```

I replaced ~2 such return locations with breaks. An equivalent return is the next non-logging statement executed after break. The reason is that I can log a message before the return, and not duplicate it in 3 places. (that is, 3 return places before became one now).

dwightguth: Java's ability to break to a label on an if is way too much like a goto for my comfort. If you want to make sure something happens prior to return on every code path, this is what a finally block is for. I will look into whether checkstyle has advanced enough to be able to forbid break-to-if (when we first set it up I believe it could not), but for now please replace this with a finally everywhere.

dwightguth (dismissed):

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
dwightguth: It seems there is still no checkstyle check for this. However, it is still something I would like to avoid in the K project as it makes code harder to understand and reason about.

denis-bogdanas:
@dwightguth Agree. This method is a goto fest. I've never seen so many return, throw, break and continue in one method. I extracted one piece into a new method and there's no need for break or finally now.

dwightguth:
@denis-bogdanas thanks. I will leave @daejunpark to review the rest.

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -107,8 +106,17 @@ public FastRuleMatcher(GlobalContext global, int ruleCount) {
         ruleMask.stream().forEach(i -> constraints[i] = ConjunctiveFormula.of(context.global()));
         empty = BitSet.apply(ruleCount);
 
-        BitSet theMatchingRules = matchAndLog(subject.term(), pattern, ruleMask, List(), false);
+        if (global.javaExecutionOptions.logRulesPublic) {
+            System.err.format("\nRegular rule automaton phase, step %d\n" +
+                    "==========================================\n", step);
+        }
+        Term automatonLHS = global.getDefinition().automaton.leftHandSide();
```
daejunpark, line 113: is there a specific reason for this change?  is it only for a better design or fixing an issue?

daejunpark: i mean, for this change, we need to ensure that `definition` inside `fastComputeRewriteStep()` be the same with `global.getDefinition()` here, which may be true for now, but could be violated later.

so, i want to know whether it is necessary or not.  (btw, if it is necessary, i guess it should be a separate pr..)

denis-bogdanas: This was a parameter inline refactoring. I needed it to distinguish 3 phases of regular rule matching: automaton phase, attempting to match, and processing match results. And to print a message at the beginning of each phase.

daejunpark: oh you mean that `global.getDefinition().automaton.leftHandSide()` involves some computation?

daejunpark: ok, if you are convinced that "definition inside fastComputeRewriteStep() is the same with global.getDefinition() here", then i'm ok.

denis-bogdanas: Yes, it is the same. This was automatic refactoring.

denis-bogdanas: `global.getDefinition().automaton.leftHandSide()` does not involve compuation. But making this move helped put log messages closer to the context they document. Without this it was not clear how logs relate to the logic of functions.

daejunpark (approved):

PR #403, opened 2019-04-09
Io extras

radumereuta:
Adding implementation for #system, #tempFilename, #remove, #parseAST for java-backend and ocaml-backend.


java-backend/pom.xml
```diff
@@ -55,6 +55,11 @@
       <version>2.23.4</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.apache.ant</groupId>
```
dwightguth: this is a really big dependency just for splitting a string into words. I'd rather we not pull this in.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
```diff
@@ -114,20 +116,24 @@ public static Term parseInModule(StringToken input, StringToken startSymbol, Str
...
+    }
+
+    public static Term tempFilename(StringToken prefix, StringToken suffix, TermContext termContext) throws IOException {
+        return StringToken.of(File.createTempFile("tmp" + prefix.stringValue(), suffix.stringValue()).getAbsolutePath());
+    }
+
+    public static Term remove(StringToken fname, TermContext termContext) {
+        File f = new File(fname.stringValue());
+        f.delete();
+        return BuiltinList.kSequenceBuilder(termContext.global()).build();
```
dwightguth: this needs to return an IOError if the delete operation fails.

k-distribution/include/builtin/domains.k
```diff
@@ -566,6 +566,7 @@ module STRING
 
   syntax String ::= Float2String ( Float )              [function, functional, hook(STRING.float2string)]
   syntax String ::= Float2String ( Float , String )     [function, klabel(FloatFormat), hook(STRING.floatFormat)]
+  syntax String ::= tokenToString( K )                  [function, hook(STRING.token2string)]
```
dwightguth: What is this used for? I would much rather we have a type-safe function instead unless you can demonstrate a use case where we need to be able to transform tokens of any sort via the same function.

k-distribution/include/builtin/domains.k
```diff
@@ -860,6 +861,12 @@ module K-REFLECTION-SYMBOLIC [symbolic]
   syntax Bool ::= #isVariable(K) [function, hook(KREFLECTION.isVariable)]
 endmodule
 
+module META
+  imports STRING-SYNTAX
+  // Takes as input a string and returns a K term
+  syntax KItem ::= #parseAST(String) [function, hook(META.parseAST), impure]
```
dwightguth: Three things. First, this is not an impure function. Second, we already have a parse function in K-IO. It's not currently implemented in any backend, so my suggestion would be to delete it, but it does need to be deleted if you want this to be the new signature of parsing in K. Third, this should be poly(0).

k-distribution/include/builtin/domains.k
```diff
@@ -980,6 +987,10 @@ module K-IO
                  | "#systemResult" "(" Int /* exit code */ "," String /* stdout */ "," String /* stderr */ ")"
 
   syntax K ::= #logToFile(String, String) [function, functional, hook(IO.log), impure, returnsUnit]
+  // #tempFilename ( prefix:String, suffix:String):IOString - creates a temporary file in the default system location
+  syntax IOString ::= #tempFilename(String, String)  [function, functional, hook(IO.tempFilename), impure]
```
dwightguth: this is not secure and is subject to the well known time-of-check-time-of-use fault. Please make the following function instead:
```
syntax IOFile ::= IOError | tempFile(path: String, fd: Int)
syntax IOFile ::= #mkstemp(String) [function, hook(IO.mkstemp)]
```

This function should create the temporary filename and open it atomically, then return the file descriptor and the path that was opened. It should return an IOError upon failure.

k-distribution/include/builtin/domains.k
```diff
@@ -980,6 +987,10 @@ module K-IO
...
+  syntax KItem ::= #remove(String) [function, functional, hook(IO.remove), impure, returnsUnit]
```
dwightguth: This should return a K, not a KItem, since its default return value is .K. Note that it needs to return an IOError upon failure and as such it is not sound to possess the returnsUnit attribute.

dwightguth (changes requested):
I have not reviewed the ocaml code in depth; I will take a look only after you have fixed the issues in domains.k and BuiltinIOOperations.k

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
radumereuta: What about the ocaml-backend? `val remove : string -> unit` returns nothing.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
```diff
@@ -114,20 +116,24 @@ public static Term parseInModule(StringToken input, StringToken startSymbol, Str
 
     public static Term system(StringToken term, TermContext termContext) {
         Map<String, String> environment = new HashMap<>();
-        String[] args = term.stringValue().split("\001", -1);
-        //for (String c : args) { System.out.println(c); }
+        String[] args = Commandline.translateCommandline(term.stringValue());
```
radumereuta: @dwightguth I use the apache library for this split. A simple split by whitespace can be wrong. Any workarounds so I can avoid including the whole library?

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
dwightguth: on ocaml you should be calling Unix.unlink which raises an exception if there is an error. We already have a function to handle those exceptions; see unix_error in prelude.ml.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
dwightguth: just invoke the command via `sh -c`

dwightguth: You probably don't want to support shell expressions though, in which case you should create a separate function, `#exec`.

nishantjr: In POSIX the `exec` family of functions replaces the current process, so I don't like the idea of reusing that name.

Python calls this functionality `subprocess.call()`
Java: `Runtime.exec()`
Rust: `std::process::Command`'s `spawn`

I think I like `#spawn`

dwightguth: That's fair, it's more like a fork-and-exec. Spawn works for me since there doesn't seem to be an immediate posix analogue.

k-distribution/include/builtin/domains.k
radumereuta: `#save(Fd, tokenToString(#token("contents", "Bubble"))`. After parsing you may end up with tokens of many types, depending on your grammar. I need a way to connect the contents of those tokens with different primitives like save-to-file, parse-strings...

k-distribution/include/builtin/domains.k
radumereuta: I've looked over Java and Ocaml libraries, and they have both options available. I added both options here too.
I did some minimal testing for `mkstemp`, not sure if this is the correct behavior you were thinking of.

radumereuta:
@dwightguth please have another look. I did some of the changes you requested and added comments on the rest. I'm not touching `#system` anymore (since I don't know who uses it) and I added `#spawn` instead and it behaves the same way on java and the ocaml backends.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
```diff
@@ -130,6 +124,57 @@ public static Term system(StringToken term, TermContext termContext) {
...
     }
 
+    /**
+     * Executes the given command line with `sh -c 'cmd'` on unix
+     * and `cmd /c 'cmd'` on windows.
+     * @param term
+     * @param termContext
+     * @return a #systemResult term containing the exit code, stdout:String, stderr:String
+     */
+    public static Term sharpSpawn(StringToken term, TermContext termContext) {
```
dwightguth: I suggested spawn as an alternative name for a function that takes a list of strings and executes that program. This does however look like an acceptable implementation of the IO.system hook finally.

k-distribution/include/builtin/domains.k
dwightguth: If you just need to convert a token of sort Bubble to a string, you can create a BubbleToString production in your semantics that hooks tto this production. We don't need a production in the prelude for this. I would prefer to keep operations on tokens type safe going forward.

k-distribution/include/builtin/domains.k
```diff
@@ -860,9 +861,16 @@ module K-REFLECTION-SYMBOLIC [symbolic]
   syntax Bool ::= #isVariable(K) [function, hook(KREFLECTION.isVariable)]
 endmodule
 
+module META
+  imports STRING-SYNTAX
+  // Takes as input a string and returns a K term
+  syntax KItem ::= #parseAST(String) [function, hook(META.parseAST), poly(0)]
```
dwightguth: let's call this #parseKAST since it operates on KAST.

k-distribution/include/builtin/domains.k
```diff
@@ -977,9 +979,22 @@ module K-IO
 // System
 
   syntax KItem ::= #system ( String ) [function, hook(IO.system), impure]
+                   // executes the given command line with `sh -c cmd:String` on unix or
+                   // `cmd /c cmd:String` on windows and returns #systemResult
+                 | #spawn  ( String ) [function, hook(IO.sharpSpawn), impure]
```
dwightguth: See above comment. If you're going to make an implementation that invokes sh -c, that implementation should be hooked to IO.system and you should not have this function at all.

k-distribution/include/builtin/domains.k
dwightguth: I'm aware that for various historical reasons both options are available in most programming languages. That is not a justification to make both options available in K, however.

k-distribution/include/builtin/domains.k
```diff
@@ -977,9 +979,22 @@ module K-IO
...
                  | "#systemResult" "(" Int /* exit code */ "," String /* stdout */ "," String /* stderr */ ")"
 
   syntax K ::= #logToFile(String, String) [function, functional, hook(IO.log), impure, returnsUnit]
+  // #tempFilename ( prefix:String, suffix:String):IOString - creates a temporary file in the default system location
+  syntax IOString ::= #tempFilename(String, String)  [function, functional, hook(IO.tempFilename), impure]
+  // #make safe temp ( prefix:String, suffix:String):IOFile - creates a temporary file in the default system location
+  syntax IOFile ::= #mkstemp(String, String) [function, hook(IO.mkstemp)]
```
dwightguth: this is impure.

dwightguth (changes requested):

k-distribution/include/builtin/domains.k
radumereuta: I was thinking the same hook could be implemented in the haskell-backend to take as input KORE. Then parseAST would be a better name. But I don't feel strongly about it. I don't know what are the plans for the haskell-backend.

dwightguth: I can say with 100% certainty that it makes no sense to have a hook that takes the same string to different outputs on different backends.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
```diff
@@ -104,30 +106,51 @@ public static Term write(IntToken term1, StringToken term2, TermContext termCont
...
         String stderr = output.stderr != null ? new String(output.stderr) : "";
         return KItem.of(klabel, KList.concatenate(IntToken.of(output.exitCode),
-            StringToken.of(stdout.trim()), StringToken.of(stderr.trim())), termContext.global());
+                StringToken.of(stdout.trim()), StringToken.of(stderr.trim())), termContext.global());
+    }
+
+    public static Term mkstemp(StringToken prefix, StringToken suffix, TermContext termContext) throws IOException {
+        FileSystem fs = termContext.fileSystem();
+        try {
+            File f = File.createTempFile("tmp" + prefix.stringValue(), suffix.stringValue());
```
dwightguth, line 139: this is not a valid way of doing this. You need to create the filename and open it atomically. From what I see online though, it /may/ work securely however if you pass f to the FileSystem object rather than f.getAbsolutePath(). You would then have to modify FileSystem.

k-distribution/include/ocaml/hooks.ml
```diff
@@ -0,0 +1,717 @@
...
+
+  let hook_stat _ _ _ _ _ = raise Not_implemented
+  let hook_lstat _ _ _ _ _ = raise Not_implemented
+  let hook_opendir _ _ _ _ _ = raise Not_implemented
+  let hook_system c _ _ _ _ = match c with
+    | [String path] ->
+      let ic, oc = Unix.open_process path in
+      let buf_out = Buffer.create 4096
+      and buf_err = Buffer.create 128 in
+      (try
```
dwightguth, line 285: not really sure what's going on here but this looks incredibly suspicious and is probably not the right way to read all the output of a process. Cf. https://stackoverflow.com/questions/29503960/in-ocaml-how-to-get-stdout-string-from-subprocess for a better solution and also bear in mind you need to use open_process_full so you get access to stderr as well.

k-distribution/include/ocaml/hooks.ml
```diff
@@ -0,0 +1,717 @@
...
+         while true do Buffer.add_channel buf_out ic 1 done
+       with End_of_file -> ());
+      (try
+         while true do Buffer.add_channel buf_err ic 2 done
+       with End_of_file -> ());
+      let exit_status = Unix.close_process (ic, oc) in
+      let exit_code = match exit_status with
+        | Unix.WEXITED n -> n
+        | Unix.WSIGNALED n -> n
```
dwightguth: this is almost  certainly not right. here your process has exited abnormally and you are treating a signal number like it's an exit code on the command line. If you want to return the true exit status, you are likely going to have to do a big if/else on all the signal numbers in the ocaml Sys module and then map them to their corresponding bash exit codes. I'm fine if you just want to raise Not_implemented in this case for now though.

dwightguth (changes requested):
looks much better. domains.k now looks good, but I have several comments on the implementation.

dwightguth:
Oh also: integration tests for the new hooks on both backends please.

k-distribution/tests/regression-new/io-java/test.k
```diff
@@ -25,6 +26,9 @@ rule close(I:Int) => #close(I)
 rule seek(I1:Int, I2:Int) => #seek(I1, I2)
 rule putc(I1:Int, I2:Int) => #putc(I1, I2)
 rule write(I:Int, S:String) => #write(I, S)
+rule system(S:String) => #system(S)
+rule mkstemp(Prefix:String, Suffix:String) => #mkstemp(Prefix, Suffix)
+rule #tempFile(Path:String, Fd:Int) => #close(Fd) ~> #remove(Path)
```
dwightguth: I'd like this test to actually test that you can read and write to this file.

k-distribution/tests/regression-new/io-ocaml/1.test.out
```diff
@@ -0,0 +1,3 @@
+<k>
```
dwightguth: should not have its own set of output files. You should just set TESTDIR in the makefile to point to the other directory.

k-distribution/tests/regression-new/io-ocaml/test.k
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2013-2019 K Team. All Rights Reserved.
```
dwightguth: this should be a symlink pointing to the other file.

dwightguth (changes requested):
implementation looks a lot better but the tests need a little bit of refactoring. Also, you still haven't addressed my comment about the suspicious way in which you are reading an input channel into a string.

k-distribution/tests/regression-new/io-ocaml/test.k
radumereuta: The two files are different. For some reason, the ocaml-backend fails with this error:
```
File "realdef.ml", line 2125, characters 180-195:
Error: This expression has type Constants.K.kitem
       but an expression was expected of type
         Constants.k = Constants.K.kitem list
```
realdef.ml area: `isTrue (evalisKResult((varHOLE_184)) config (-1))` (`varHOLE_184` is the problematic term)
test.k area: `context assertEqual(HOLE, _:KItem)`
You are right about the tests. They can be symlinks, but I don't know where this is comming from, so I just created another test.

dwightguth: Can you please post the diff between the two files then? That might help me figure out what is causing the error you're getting.

radumereuta:
Regarding the way I read the output from the process into a string.
I took the code from: https://www.rosettacode.org/wiki/Execute_a_system_command#OCaml
I'm not sure if it's the best way, but the tests work as I expect.
Ex: issue 1 from that link:
A call to `#system("echo sout ; >&2 echo serr ; sleep 60")` produces the following result
`#systemResult ( 143 , "sout\n" , "serr\nTerminated\n" )`. So both java and ocaml manage to get the stdout and stderr even if the process was killed.
Notice the error code. Java will always return that error code, but in ocaml you get extra information when the process was killed. In order to preserve uniformity, I had to reconstruct the error code. If you follow the link from stackoverflow, you will see it's not perfect, but works for the most common cases.
Issue 2 from the link:
I think `#system` should block the current execution until all the output is collected, then resume. If the process never terminates, it's the job of the user to stop it.
Am I missing something else?

k-distribution/tests/regression-new/io-ocaml/test.k
radumereuta: I removed the context declaration and the 3 rules at the end which seemed inactive.
The error points to the `HOLE` from the context.
Or are you talking about the diff from the resulting realdef.ml files?

radumereuta: Actually realdef.ml doesn't make sense because the other one is compiled with java.

k-distribution/include/ocaml/hooks.ml
dwightguth: I understand better what is going on here but can you increase the number of characters read at one time to 4096?

dwightguth (approved):

dwightguth:
Feel free to merge this, or wait until Nishant reviews if you'd rather.

PR #405, opened 2019-04-10
fix bug with stringbuffer on symbolic backends 

dwightguth:
@denis-bogdanas this should fix your issue.


traiansf (approved):
looks reasonable to me.

PR #406, opened 2019-04-12
Kast respects output, allows specifying any output format for `kast`

ehildenb:
This achieves two things:

-   `kast` program now gets a `PrintOptions`, and goes through `KPrint` to do its printing. This means any output format supported by `KPrint` is not supported by `kast` (instead of just KAST format).
-   ~~`KORE` output format is added to `KPrint`, so that we can now say `--output kore` for any of the given programs.~~

One thing that may need to be change: before this the behavior of `kast` was always `--output kast`. Now since we're getting the output format from `PrintOptions`, the default is `--output pretty`. Should we override this for the `kast` program so that the default is now `--output kast` when running `kast`? Currently, in the two places in the frontend where it needs to call `kast` manually, I'm manually adding `--output kast` to retain original behavior, but external users of the `kast` program will see a difference in behavior.


ehildenb:
@dwightguth I've removed the addition of `--output KORE` format from the PR, as there are some more issues with that I need to iron out locally first.

We still need to decide if `kast` should default to `--output pretty` (which it does in this PR), or if we should default to `--output kast` (to preserve original behavior).

ehildenb:
This also adds a testing harness for `kast` tests.

dwightguth (approved):

PR #407, opened 2019-04-12
[BUGFIX] Determining rule for getExitCode relies on top cell being named generatedTop

ehildenb:
When compiling to the Kore/Haskell/LLVM backends, the logic here https://github.com/kframework/k/blob/cebb9ab45fc99f2c2b30df9401a50311f5a496ca/kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java#L486 assumes that some cell `<generatedTop>` exists, which may not be the case (as this test demonstrates).


dwightguth:
@ehildenb you can't review this because you created it but please let me know if the fix looks good and if it does, I will approve.

ehildenb:
@dwightguth this seems to fix the immediate issue I was having, though I'm still having trouble building KEVM after changing the configuration slightly: https://github.com/kframework/evm-semantics/pull/335

The remaining issues have to do with the ocaml plugin though, not with this changeset, so this should be good to merge.

dwightguth (approved):

PR #410, opened 2019-04-15
java-backend: cleanup

denis-bogdanas:
Nothing particularly important but I did this while looking for other issues.


java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
```diff
@@ -17,38 +15,32 @@
 /**
  * An object containing context specific to a particular configuration.
  */
-public class TermContext extends JavaSymbolicObject {
+public class TermContext {
```
daejunpark, line 18: i cannot see why this is necessarily safe.

java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
```diff
@@ -17,38 +15,32 @@
...
     private final GlobalContext global;
 
-    private Set<Variable> initialVariables;
+    private Set<Variable> initialLhsVariables;
 
     private Term topTerm;
 
     private ConjunctiveFormula topConstraint;
 
-    private KOREtoBackendKIL converter;
```
daejunpark, line 32: i cannot see why this is necessarily safe.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -192,7 +189,6 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
             KOREtoBackendKIL converter = new KOREtoBackendKIL(module, definition, termContext.global(), false);
             ResolveSemanticCasts resolveCasts = new ResolveSemanticCasts(true);
             ExpandMacros macroExpander = new ExpandMacros(module, files, kompileOptions, false);
-            termContext.setKOREtoBackendKILConverter(converter);
```
daejunpark, line 195: i cannot see why this is necessarily safe.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -120,20 +120,17 @@ public void chainOptions() {
...
+        Definition definition = initializeDefinition.invoke(def.mainModule(), kem, initializingContext);
         GlobalContext rewritingContext = new GlobalContext(fs, globalOptions, krunOptions,
                 kproveOptions, javaExecutionOptions, kem, smtOptions, hookProvider, files, Stage.REWRITING, profiler,
                 kprint, def);
         rewritingContext.setDefinition(definition);
 
-        return new SymbolicRewriterGlue(mainModule, definition, definition, transitions,
-                initializingContext.getCounterValue(), rewritingContext, kem, files, kompileOptions, sw);
+        return new SymbolicRewriterGlue(def.mainModule(), definition, definition, transitions,
+                0, rewritingContext, kem, files, kompileOptions, sw);
```
daejunpark, line 133: I do remember there was an issue that was fixed by having `initializingContext.getCounterValue()` instead of `0`, but I cannot recall the original issue.

daejunpark (commented):
I'm not comfortable to review this change.  It doesn't seem to me that the changes are trivial or safe.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
daejunpark: I guess that if we have more than one spec rules, this initial counter value needs to be different for each spec proof.  This counter is used to handle `?X`, if I remember correctly.

java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
denis-bogdanas: Because this extension is wrong and not used anywhere. That's why this PR is a cleanup - removing unused stuff.

java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
denis-bogdanas: Same as above. If it wouldn't be safe, code wouldn't compile.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -120,20 +120,17 @@ public void chainOptions() {
 
     @Override
     public synchronized Rewriter apply(org.kframework.definition.Definition def) {
-        Module mainModule = def.mainModule();
-        TermContext initializingContext = TermContext.builder(new GlobalContext(fs, globalOptions, krunOptions,
```
denis-bogdanas, line 124: The TermContext created here is essentially not used. You see below, it is only used to access GlobalContext, and CounterValue that stayed zero. So this refactoring essentially removes this TermContext and leaves only the GlobalContext.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: Look at my comment above.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
denis-bogdanas: because `setKOREtoBackendKILConverter` was deleted.

daejunpark (approved):

PR #414, opened 2019-04-17
Fixes for the C semantics with the latest version of K

dwightguth:


ocaml-backend/src/main/java/org/kframework/backend/ocaml/SplitThreadsCell.java
```diff
@@ -106,7 +106,19 @@ public K apply(KApply k) {
...
+                            Tuple3<K, K, K> split = splitSide(threadsCell.get());
+                            K global = replaceThreadLocal(k);
+                            return KApply(KLabel("#Thread"), global,
+                                    split._1(),
+                                    split._2(),
+                                    split._3());
+                        }else {
+                            // these should be variable names not used anywhere else in the rule since only anonymous variables
+                            // can begin with _ in a user rule and only _0, _1, _2, etc are generated by ResolveAnonVar
+                            return KApply(KLabel("#Thread"), k, KVariable("_A"), KVariable("_B"), KVariable("_C"));
```
traiansf, line 120: Wouldn't it be possible to apply this transformation in two places, or to have another transformation adding variables?

I would at least name the variables more verbose, e.g.,  _SplitThreadsCell...

traiansf (approved):

PR #416, opened 2019-04-19
Java Backend cache: introduced KItem.evaluatedRecursively

denis-bogdanas:
that prevents recursive traversal of evaluated ground terms.

Tested on VSC: https://github.com/runtimeverification/verified-smart-contracts/pull/225
Total Jenkins suite time 40% less.


daejunpark:
Does this prevent only "substitution" of grounded terms, or any "function evaluation" as well?

I mean, a function term inside a ground term has no potential to be further evaluated later?

denis-bogdanas:
Only for ground terms, and yes, they don't have potential to be evaluated further on their own. For terms with variables, evaluation result depends on context, so this optimization is not possible for them or at least would be more complicated and probably less useful. Most terms in a configuration are ground, like <program> cell that contains ~10K items for gnosis. So this optimization captures the most benefit.

I added a comment for clarification.

daejunpark (approved):

PR #421, opened 2019-04-22
Update k frontend to use new scala pattern matching code

dwightguth:
@theo25 I can't mark you as a reviewer yet because you haven't yet accepted my invite into the kframework organization, but I'd still like you to review this as well since it's pretty much part of my last pr into llvm-backend.


traiansf (approved):
looks reasonable to me.

PR #424, opened 2019-04-24
Adding parseKAST tests

radumereuta:
@dwightguth I get an issue with test no 2 on the java-backend.
I get the following output:
```
[Error] Critical:  (unexpected subject class at matching: class
org.kframework.backend.java.kil.KSequence)
  while matching rule pattern:
    Subject: .K
    Pattern: parseKAST(R_S:String)
  while matching rule pattern:
    Subject: .K
    Pattern: #KSequence(parseKAST(R_S:String), R_DotVar0:K)
  while matching rule pattern:
    Subject: .K
    Pattern: #KRewrite(#KSequence(parseKAST(R_S:String), R_DotVar0:K),,
org.kframework.backend.java.kil.InnerRHSRewrite@aa14de04)
  while matching rule pattern:
    Subject: <k>(.K)
    Pattern: <k>(#KRewrite(#KSequence(parseKAST(R_S:String), R_DotVar0:K),,
org.kframework.backend.java.kil.InnerRHSRewrite@aa14de04))
```
Any ideas?
Also, how should I handle parse errors? Should I return `#noparse` or `#noparse("error message")` or should I just let krun throw an error?


dwightguth:
`#noparse` is fine. As for the error, I don't really know what to suggest, as I'm not that familiar with that part of the Java backend. Does @denis-bogdanas know anything?

denis-bogdanas:
I don't know the context. This error sometimes appears in kprove when there are problems with the spec. If it appears during parsing It's better to replace it with something more meaningful.

radumereuta:
@dwightguth please have another look.
You were right about the error. It needed that transformer.
I also added the error message to `#noParse` because it was easy and it could potentially save us time in the future. Made it camelcase so we are consistent across `domains.k`.

dwightguth (approved):

PR #431, opened 2019-04-26
Configuration formatting improvements

nishantjr:
This change allows use to use format attributes as cell properties, and improves pretty printing of Map, List and Set cells.

We can now do write `configuration <foo  format="foo(%2)"> ... </foo>` to have K pretty print the configuration as `foo(...)`.


dwightguth:
looks like you need to update the unit test for the transformer.

nishantjr:
@dwightguth Yes, done.

What kind of integration tests would you like? I had expected that I'd just have to change existing tests that use Map cells etc, but none seem to have failed (at least on my machine).

dwightguth:
you just need to write a test with a configuration with one of each type of cell collection and one cell with a custom format attribute, and then execute some program in that semantics, even if it takes zero steps.

dwightguth:
Also, don't forget to sign the CLA.

nishantjr:
@dwightguth For some reason the format attribute for MapItem etc is ignored by the Java backend (but not ocaml). Any suggestions?

dwightguth:
The cell collections are not handled properly on the java backend and are transformed into terms of sort List/Map/Set. There's not really anything you can do about this... it's too complicated and ingrained a change to make.

nishantjr:
@dwightguth  Ok. This is ready for your review now.

dwightguth:
I'm not going to review this yet because I want #432 to be merged first as it includes a much better change to the regressions-new/Makefile that is going to conflict with your changes. Will take a look once that's merged and you resolve conflicts though.

nishantjr:
Great. I like that change.

dwightguth (approved):

PR #441, opened 2019-05-02
Bug fixes for kore generation and also a new test harness for testing generated kore

dwightguth:
Includes:

* fix to bug that inserted append on lhs of rules.
* fix bug in ocaml backend involving sort of or patterns
* fix bug in kore generation involving term construction minimization happening under an or pattern
* test harness to test properties of generated .kore files.


k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala
```diff
@@ -0,0 +1,98 @@
...
+        case And(_, Top(_), And(_, _, rw @ Rewrites(_, _, _))) => Some(rw)
+        case Rewrites(s, And(_, Equals(_, _, _, _), l), And(_, _, r)) => Some(B.Rewrites(s, l, r))
+        case Rewrites(s, And(_, Top(_), l), And(_, _, r)) => Some(B.Rewrites(s, l, r))
+        case Implies(_, Bottom(_), p) => go(p)
+        case Implies(_, Equals(_, _, _, _), And(_, eq @ Equals(_, _, Application(_, _), _), _)) => Some(eq)
+        case Implies(_, Top(_), And(_, eq @ Equals(_, _, Application(_, _), _), _)) => Some(eq)
+        case Implies(_, And(_, _, Equals(_, _, _, _)), And(_, eq @ Equals(_, _, Application(_, _), _), _)) => Some(eq)
+        case Implies(_, And(_, _, Top(_)), And(_, eq @ Equals(_, _, Application(_, _), _), _)) => Some(eq)
+        case eq @ Equals(_, _, Application(_, _), _) => Some(eq)
+        case _ => None
```
ehildenb, line 66: How did you come up with this list of patterns? Could it be done as a visitor which accumulates a list of Rewrites and Equals instead?

dwightguth: @ehildenb it cannot be done as a visitor that accumulates all rewrites and equals, because some instances of those patterns do not correspond to actual rules in the original semantics, and therefore should not be included in this list. This list of patterns corresponds more or less to the structure of the axioms generated from rules in ModuleToKore.java and was taken originally from the llvm backend.

Last I heard, the haskell backend was also doing something very similar in matching on the structure of axioms, but I'm not sure if that's still true. @ttuegel may know more.

dwightguth: with that being said, I'm not opposed to moving this logic into the k frontend and sharing the same code path between this unit test and the llvm backend, if you'd like.

traiansf:
Changes seem reasonable to me; however I would like to see test K tests passing on the haskell backend. I might approve this even if they don't but we should at least know.

ehildenb:
@traiansf I'm not sure what you mean. Do you mean you want to run the Haskell backend `test-k` target on PRs into K? Or that you want to run all the K repo tests against the haskell backend? The former may be possible, the latter almost certainly won't be since the haskell backend hasn't implemented everything yet.

k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala
ehildenb: If it's common logic, then I think it should be moved. This list of patterns still seems rather ad-hoc to me (not saying that it is ad-hoc, just that I can't seem to see how it's derived). But if it's common for all backends then it seems ok.

ehildenb (approved):

traiansf:
@ehildenb I just meant `test-k` If those pass, than it should be fine.

dwightguth:
test-k passes on my machine.

PR #445, opened 2019-05-03
kernel: BinaryLoader - added shared file lock on reading

denis-bogdanas:
No concrete issues solved but now definitely this piece of code should not bring any issues.


dwightguth:
This is also giving an overlappingfilelockexception, so you need to lock a ReadWriteLock for same jvm access.

dwightguth (approved):

dwightguth:
I approved. I also kicked off the build again just to verify we don't have any obvious intermittent failures. If the tests pass, go ahead and merge.

PR #449, opened 2019-05-05
Update strategy language and add simple tests

ehildenb:
This PR does:

-   Renames `#RuleTag` sort to `RuleTag`, which makes it possible to now use ruletag variables like `RT:RuleTag` and write in object syntax instead of using klabels.
-   Removes the intermediate sort `AtomicStrategy` (in favor of using `RuleTag` directly).
-   Adds a testing harness for strategies, and some simple tests.
-   Fixes some bugs in alternation `_|_`.


dwightguth:
I haven't looked at the code yet, but how are you avoiding the syntax collision with RuleTag?

ehildenb:
I'm not sure which syntax issues you're referring to @dwightguth. I renamed the sort `#RuleTag => RuleTag`, do you foresee that causing problems? The issue I had before was that writing `RT:#RuleTag` wouldn't parse, so I couldn't have rules which had symbolic rule tags, but now I can do `RT:RuleTag` directly.

dwightguth:
@ehildenb what I'm worried about is the collision between the syntax of RuleTag and other productions. If I have something like:
```
syntax Foo ::= "foo"
```

Then `foo` would parse as either a Foo or a RuleTag, generating an ambiguity.

ehildenb:
Well that was an issue already, but in a previous PR I separated out the `RULE-TAG-SYNTAX` module, so that you can import it directly only if you need to refer to rule tags directly in your module.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -200,7 +200,7 @@ public static ParseInModule getCombinedGrammar(Module mod, boolean strict, boole
         if (mod.importedModuleNames().contains(AUTO_CASTS)) { // create the diamond
             Set<Sentence> temp;
             for (Sort srt : iterable(mod.definedSorts())) {
-                if (!isParserSort(srt)) {
+                if (!isParserSort(srt) || mod.subsorts().directlyLessThan(Sorts.KVariable(), Sorts.RuleTag())) {
```
dwightguth: I believe this should be directlyLessThan(Sorts.KVariable(), srt), right?

ehildenb: Haha, good catch. Fixed.

dwightguth (approved):

PR #451, opened 2019-05-08
FFI struct type

mickyabir:
Adds FFI struct type


dwightguth (changes requested):
This needs to be `| #struct(List)` because the struct type needs to contain a record of the children of the struct.

dwightguth (approved):
looks good, merge if tests pass.

PR #452, opened 2019-05-08
Allowing non-rewrite claims; proper encoding for ML binders

traiansf:


traiansf:
@yzhang90 In order to make your specification `kompile` and get the KORE output you want, you need to (update K to this PR and) add the ` [anywhere, kore]` attributes to the claim.

`anywhere` tells the compilation process that this claim can "apply" anywhere, hence no need to add `<k>` and other cells on top of it; I guess another way to achieve this would be to detect that the rule's sort is already `TCell`, but that might be harder to do at that  compilation stage.

The `kore` attribute added by this PR instructs the KORE generation algorithm that this claim is not of the rule/equation kind, so it has no LHS and RHS.

dwightguth:
can we make it so that the anywhere attribute is not required and `kore` implies that we don't add implicit cells?

traiansf:
>can we make it so that the anywhere attribute is not required and kore implies that we don't add implicit cells?

@dwightguth I can try. I looked into this, but it seemed we need to change more than a few lines to allow that.

traiansf:
@dwightguth turns out you were right; it was actually a simple change to make. please re-review.

dwightguth (approved):

PR #453, opened 2019-05-08
 support for fresh constants in function rules in llvm backend

dwightguth:
This is not handled in a way that would work for symbolic execution, but it's good enough for the llvm backend so we're going in this direction for now in order to unblock us.


dwightguth:
my next PR substantially reworks this one so I'm going to close this one.

PR #457, opened 2019-05-09
Refactor injected options in the haskell backend

ehildenb:
This PR:

- Makes it so haskell backend doesn't print out extra stuff unless you include `--verbose`.
- Injects a `KPrint` into `HaskellRewriter` and uses the options from there.
- Injects `GlobalOptions` and `SMTOptions` into haskell rewriter and uses options directly from there when applicable.
- Correctly passes `depth` parameter instead of looking up in options when invoking haskell backend.

### DIFF
--- a/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
+++ b/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
@@ -55,7 +55,7 @@ public class HaskellRewriter implements Function<Definition, Rewriter> {
 
     private final FileUtil files;
     private final CompiledDefinition def;
-    private final KRunOptions options;
+    private final KRunOptions krunOptions;
     private final KompileOptions kompileOptions;
     private final KExceptionManager kem;
     private final HaskellKRunOptions haskellKRunOptions;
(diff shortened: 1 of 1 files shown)


haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -112,18 +120,18 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
                         "--module", moduleName,
                         "--pattern", pgmPath,
                         "--output", koreOutputFile.getAbsolutePath()));
-                if (options.depth != null) {
+                if (depth.isPresent()) {
                     args.add("--depth");
-                    args.add(options.depth.toString());
+                    args.add(depth.toString());
```
dwightguth: the java library specification does not provide any guarantees about the value of Optional.toString so you really should be calling get() here.

dwightguth (changes requested):
looks good but as is it has a chance of breaking the --depth flag so we really should address my comment before we merge.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
ehildenb: Fixed.

dwightguth (approved):

PR #458, opened 2019-05-10
Allow #AG spec for bounded model checking

yzhang90:


k-distribution/include/kore/prelude.kore
```diff
@@ -47,4 +47,8 @@ module K
   alias weakAlwaysFinally{A}(A) : A
   where weakAlwaysFinally{A}(X:A) := X:A []
 
+  // Definitions for CTL aliases
+  alias allPathGlobally{A}(A) : A
+  where allPathGlobally{A}(X:A) := X:A []
```
dwightguth, line 53: I remember you said this is not sound, which obviously it's not since it's not actually doing anything here. Can we get a TODO then?

dwightguth (changes requested):

k-distribution/include/kore/prelude.kore
yzhang90: Added a same as weakExistsFinally.

dwightguth (approved):

PR #460, opened 2019-05-10
 don't add hook attribute or hooked- prefix to unknown hooks 

dwightguth:
I deleted the META module and moved parseKAST into K-REFLECTION. I also moved --hook-namespaces into KompileOptions and used it to implement the functionality of custom hooks on the llvm and haskell backends.


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -541,6 +545,17 @@ public String convert(boolean heatCoolEq) {
...
+    private boolean isRealHook(Att att) {
+      String hook = att.get(Attribute.HOOK_KEY);
+      if (hook.startsWith("ARRAY.")) {
+        return false;
+      }
+      if (options.hookNamespaces.stream().anyMatch(ns -> hook.startsWith(ns + "."))) {
+        return true;
+      }
+      return Hooks.namespaces.stream().anyMatch(ns -> hook.startsWith(ns + "."));
+    }
```
ehildenb, line 557: I guess these two could be combined with some sort of `Set.union(Hooks.namspaces, options.hookNamespaces).steam()....`

Up to you.

ehildenb (approved):

PR #464, opened 2019-05-13
[BUGFIX] Fix exit-code status of Java backend execution

ehildenb:
The java backend has incorrectly be reporting the exit-code for some time, giving the following error even for fully-concrete exit-codes:

```
[Warning] Critical: Found 0 integer variables in exit code pattern. Returning 111.
```

This complicates refactors to the KEVM testing harness which require that the exit-code work.

The method `getExitCode` in `KRun.java` is to blame, but the logic there is pretty weird. No idea what it's trying to accomplish, but perhaps it needs to be scrapped and made simpler.


ehildenb:
Note that `SHELL=bash -o pipefail` must be set for this test to work (it's set in the `Makefile`, so no action needed on users part).

PR #465, opened 2019-05-13
Java Backend builtin map: fixed map.removeAll when attempting to remove symbolic elements

denis-bogdanas:
Also tests. Only tests remove-01-03 & 16... were passing before this fix.

Tests are based on tests for lookup.


k-distribution/tests/regression-new/map-symbolic-tests/remove
```diff
@@ -0,0 +1,19 @@
+//copied from lookup test plan
+
```
ehildenb: Get rid of these two lines, I don't think the `gen-tests.sh` script will treat them correctly.

Also, you need to modify `gen-tests.sh` to also call `remove` in addition to `assignment lookup inkeys`.

ehildenb (changes requested):

ehildenb:
As far as checking that the tests give what you guys expect, I think @daejunpark should have a look as well, so I'll request him as a reviewer.

k-distribution/tests/regression-new/map-symbolic-tests/remove
ehildenb: The test files `remove-#-spec.k` should be populated by the `gen-tests.sh` script here.

denis-bogdanas: I didn't use the script. What is it supposed to do?

denis-bogdanas: Done.

ehildenb:
The tests look good, if @daejunpark approves it should be ready for merge.

denis-bogdanas:
@ehildenb I guess you should change your review status to "comment".

daejunpark (approved):

ehildenb (approved):

PR #468, opened 2019-05-14
Java Backend engine: Fixes required to make Bottom a valid RHS of Substitution

denis-bogdanas:
Prerequisite to improvements in ConjunctiveFormula.simplify().


ehildenb:
This looks OK to me, but I want @daejunpark and @dwightguth to weigh in as well.

daejunpark (approved):

PR #469, opened 2019-05-14
Java Backend: NPE fix

denis-bogdanas:


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -1100,7 +1102,13 @@ private ConstrainedTerm applySpecRules(ConstrainedTerm constrainedTerm, List<Rul
...
                 ConstrainedTerm result = buildResult(specRule, constraint, null, true, constrainedTerm.termContext(),
                         new FormulaContext(FormulaContext.Kind.SpecConstr, specRule, global));
-                global.stateLog.log(StateLog.LogEvent.SRULE, specRule.toKRewrite(), constrainedTerm.term(), constrainedTerm.constraint(), result.term(), result.constraint());
+                if (result != null) {
+                    global.stateLog.log(StateLog.LogEvent.SRULE, specRule.toKRewrite(), constrainedTerm.term(),
+                            constrainedTerm.constraint(), result.term(), result.constraint());
+                } else {
+                    global.stateLog.log(StateLog.LogEvent.SRULE, specRule.toKRewrite(), constrainedTerm.term(),
+                            constrainedTerm.constraint());
+                }
```
ehildenb, line 1117: We shouldn't be logging that a spec rule was applied `SRULE` if it doesn't apply correctly. So I think you can get rid of the `else` branch.

ehildenb (changes requested):

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
denis-bogdanas: Done.

ehildenb (approved):

PR #482, opened 2019-05-17
FFI variadic functions

mickyabir:


k-distribution/include/builtin/ffi.k
```diff
@@ -34,7 +34,14 @@ module FFI
 
   /*@ The FFI call takes an int for the symbol to call, a list of bytes, a list of types, and a return type. */
   syntax Bytes ::= "#ffiCall" "(" Int "," List "," List "," FFIType ")" [function, hook(FFI.call), impure]
+  /*@ The variadic FFI call takes an int for the symbol to call, a list of bytes, a list of fixed types, a list of variadic types and a return type. */
+                 | "#ffiCall" "(" Int "," List "," List "," List "," FFIType ")" [function, hook(FFI.call_variadic), impure]
+  /*@ The FFI call takes a boolean isVariadic, an int for the symbol to call, a list of bytes, a list of types and a return type. */
+                 | "#ffiCall" "(" Bool "," Int "," List "," List "," FFIType ")" [function, impure]
```
dwightguth: I think you forgot to put the Int for NFixed here.

dwightguth:
I'm not entirely sure why we're getting this error about impure functions when we have already previously gotten other impure functions to work, but it's not really worth digging into too deeply. Can you remove the impure attribute from all three functions for now? It's just a temporary fix, but it should be fine for the time being.

dwightguth:
This looks good but I don't think it's going to pass tests until we merge the llvm backend pr and update the submodule. So please focus on that for now.

dwightguth (approved):

PR #489, opened 2019-05-23
Fix bug involving fresh functions and top sort in ocaml backend

dwightguth:
@nishantjr FYI


kernel/src/main/java/org/kframework/compile/ExpandMacros.java
```diff
@@ -66,6 +66,9 @@
     private final boolean reverse;
     private final ResolveFunctionWithConfig transformer;
 
+    // note that because of the constraints required on ResolveFunctionWithConfig, this overload
+    // MUST NOT be called unless reverse=true OR mod is the main module of the definition. And if
+    // reverse=true, you MUST NOT call the resolve overload for sentences.
```
traiansf: could the comments here become assertions? or at least make this private and have a static factory method with a suggestive name?   I'm worried people might not look at the comment and ignorantly violate the invariants.

traiansf (approved):

kernel/src/main/java/org/kframework/compile/ExpandMacros.java
dwightguth: I will make a factory method. But it's not really feasible to make it an assertion because we can't assert that the user won't call a certain method or that something they pass in is the main module when we aren't passing in the entire definition.

PR #490, opened 2019-05-23
Upstream gen spec script

ehildenb:
This script comes from the VSC repo originally.

It differs in that you can specify any level in the proof tree (not just the leaves), and it will generate all sub-proofs in the same spec file. For example, if your ini file is:

```
[pgm]
specs: transfer transferFrom
...

[root]
...

[transfer]
...

[transfer-success]
...

[transfer-success-1]
...

[transfer-failure]
...

[transferFrom]
....
```

You could say `./gen-spec.py DEFN_TMPL RULE_TMPL INI_FILE transfer-success-1 transfer-failer transferFrom` and it will generate one rule per spec for each of the listed proof names (so 3 specs total). But if you do `./gen-spec.py DEFN_TMPL RULE_TMPL INI_FILE transfer transferFrom`, it will only generate 2 specs, one with both `transfer-success-1` and `transfer-failure`, and another with `transferFrom`. If you don't supply any spec names, it defaults to the list in `pgm.specs`.


ehildenb:
Here is the KEVM PR which uses this new script: https://github.com/kframework/evm-semantics/pull/356

dwightguth:
I don't have time to look at the script in detail yet, but can we give it a better name, remove the .py, and add a wrapper in the distro-bin folder for it?

ehildenb:
@dwightguth I've renamed the script `gen-specs.py => kprove-ini`, and added a wrappen in `distro-bin` for the script.

I've also added a testing harness in `ktest.mak` which tests that the generated `*-spec.k` files are what we expect.

Issue #491, opened 2019-05-23
Concretize cells on poly production


yzhang90:
In BMC, we need to recursively addTopCells to the children of a production whose sort are the same as the return sort.

yzhang90:
see https://github.com/kframework/k/pull/498

PR #498, opened 2019-05-24
Concretize cells on poly production

yzhang90:
see #491.

Quantify over implicit variables will come in the next PR.


kernel/src/main/java/org/kframework/compile/AddImplicitComputationCell.java
```diff
@@ -63,7 +63,7 @@ public Sentence apply(Sentence s) {
         }
     }
 
-    static boolean skipSentence(Sentence s) {
+    private boolean skipSentence(Sentence s) {
```
dwightguth, line 66: what is the rationale behind removing static? it doesn't look like anything was changed in the method body.

yzhang90: Previously this function is also used in `AddTopCellToRules.java`.(when adding `kore` attribute to the spec, it will skip both `AddTopCellToRules` and `AddImplicitComputationCell`). However, now we only want it to skip `AddImplicitComputationCell`. So I made another skipSentence function in the `AddTopCellToRules`. There is no need to make this function static.

dwightguth (approved):

PR #499, opened 2019-05-24
Faster sort inference of ml primitives in unparsing

dwightguth:


dwightguth:
Tested this and it seems to be passing the haskell backend integration tests.

dwightguth:
I have tested this and it sucessfully sorts the kore file such that the only differences are order of attributes within a rule and the values of the productionID and Source attributes.

ehildenb (approved):

PR #506, opened 2019-06-03
kerenl/KastFrontEnd: simplify logic for selecting parsing module

ehildenb:


ehildenb:
I'm also rolling together an LLVM backend submodule update so that this PR will work with the rule-summaries stuff on KEVM.

dwightguth (approved):

PR #511, opened 2019-06-03
fix strategy

dwightguth:
Fixes bug where strategy cell is not included in some rules.


ehildenb (approved):

dwightguth:
This appears to break the C semantics, so please don't merge until I fix it.

PR #513, opened 2019-06-04
Add `injective` to constructor-like Kore symbols

ttuegel:
The Haskell backend uses the `injective` symbol attribute to aid unification of some constructor-like symbols. This will be needed for unification with overloaded symbols, which are `injective` but not `constructor`. I tested this with the Haskell backend's integration test suite to check that the generated definitions have the attribute in the appropriate places.


dwightguth:
I'm confused by this. You are checking for the presence of the injective attribute, but it's not being added anywhere during compilation or in domains.k.

ttuegel:
> You are checking for the presence of the injective attribute, but it's not being added anywhere during compilation or in domains.k.

This pull request would add it during compilation.

dwightguth:
Sure, but as far as I can tell you are checking for its presence /before/ it is added. why?

dwightguth (approved):

PR #516, opened 2019-06-05
Support #Forall _/ #Exists _

yzhang90:


dwightguth (approved):
I assume you inspected the output of using this feature and it looks correct to you?

yzhang90:
> I assume you inspected the output of using this feature and it looks correct to you?

yes

PR #517, opened 2019-06-06
Tests for undefined term in K definition: should output meaningful kompile error

denis-bogdanas:


denis-bogdanas:
@dwightguth This PR contains 2 definitions that are expected to give kompile error. How to write tests for this?

dwightguth:
You can put the tests and their output files in the checks directory under regression-new.

PR #528, opened 2019-06-10
fix libffi link

dwightguth:
This should hopefully fix the broken testing on mac os.


traiansf (approved):
looks ok to me, if tests pass.

PR #530, opened 2019-06-10
fix bug in MinimizeTermConstruction

dwightguth:


traiansf (commented):
I'm not sure I understand what is this trying to fix. I kind of figured that you don't want to use As patterns in or or builtin collections, but I don't quite follow why is that, and the PR itself is not self-explanatory on this matter.

dwightguth:
So, we can't do it inside an or pattern because unless the pattern appears on every branch of the or, it can't actually be used on the rhs, and we don't want to bother to compute whether that is the case.

As far as collections, it considerably simplifies pattern matching to forbid as patterns which bind subcollections of a map, list, or set. so we are careful not to introduce such terms in this pass as the llvm backend does not yet support them.

traiansf (approved):

PR #531, opened 2019-06-11
context alias

dwightguth:
Basically this is exactly what we discussed but with two slight tweaks to the syntax: we use the same rule labelling syntax to label the added notation, and we use the "alias" keyword instead of "notation". Feel free to discuss either.


dwightguth:
Seems to have failed due to a transient issue connecting to opam online, so I'm rebuilding.

ehildenb (approved):
Can we add one test which uses a `context alias` with a `requires` clause?

ehildenb (approved):

PR #538, opened 2019-06-11
Kore to k 2

traiansf:


kernel/src/main/java/org/kframework/parser/KRead.java
```diff
@@ -45,6 +59,14 @@ public K prettyRead(Module mod, Sort sort, CompiledDefinition def, Source source
         }
     }
 
+    private K parseKore(Module mod) {
+        try {
+            return  KoreToK.parseKoreToK(options.fileToParse(), idsToLabelsProvider.getKoreToKLabels(), mod.sortAttributesFor());
+        } catch (ParseError parseError) {
+            throw KEMException.criticalError("Parse error" );
+        }
+    }
```
ehildenb: I don't like that `KRead` has a `KastOptions` put into it. Can we instead make this function just take a string to parse, and assume that reading the file has been handled elsewhere?

The main reason is that we intend to make KRun and KProve go through KRead as well (so that we can avoid parse-time by passing in JSON input or BINARY input, for example), and they do their own file-handling code.

ehildenb (changes requested):

kernel/src/main/java/org/kframework/parser/KRead.java
ehildenb: Note also that @gtrepta is working on making KRun go through KRead right now, which will make a new `KReadOptions.java` available. Currently we're only giving it the `--input` options, and we'll include `KReadOptions` into `KRunOptions` and `KastOptions`. But if there are more parameters which are inherent to KRead, we can move those over too once he's done.

But I'd like to do the refactoring incrementally to reduce chances of bugs, so we're just focusing on getting KRead to be used directly by other parts of the code (and the other parts of the code need to do their own file handling), then we'll refactor it to localize more of the file handling code into KRead directly.

dwightguth (commented):
I will let Everett review this one since I'm sick.

kernel/src/main/java/org/kframework/parser/KRead.java
traiansf: We could, but the Kore parsing code is in Scala and doesn't have a parse from string option and I was reluctant to add it. On the other hand, isn't it less efficient to pass a string, esp. Considering that the input can sometimes be quite large?

traiansf:
@ehildenb I've refactored it as you suggested, and more :-)
could you please take another look?

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -110,13 +105,24 @@ public int run() {
...
+                compiledMod = def.kompiledDefinition.getModule(options.module).get();
+            }
             Option<Module> maybeMod = def.programParsingModuleFor(options.module, kem);
             if (maybeMod.isEmpty()) {
                 throw KEMException.innerParserError("Module " + options.module + " not found. Specify a module with -m.");
             }
             Module mod = maybeMod.get();
-            Module compiledMod = def.kompiledDefinition.getModule(options.module).get();
 
             K parsed = kread.prettyRead(mod, sort, def, source, FileUtil.read(stringToParse), options.input);
```
ehildenb, line 128: We need to change the code which comes after this. The next `if` block should only check if `options.expandMacros`, and we should get rid of the `--kore` option (in favor of using `--input kore`).

So the second `if` block after this line should be removed, and we should go straight to the `else` branch.

ehildenb (changes requested):

ehildenb (approved):

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
ehildenb: @traiansf pointed out that the code following this block is dealing with _output_ and not _input_ options, so this comment is incorrect.

PR #542, opened 2019-06-12
java backend: Partial work on automaton for spec rules

denis-bogdanas:
Enabling refactorings in different PR: #541 
VSC build: https://github.com/runtimeverification/verified-smart-contracts/pull/247

Attempt to enable automaton for spec rules. Disabled after I realized this requires more extensive changes in pre-processing steps that we cannot afford right now. Still refactorings here clean up the code and are useful to keep.

Core idea: add spec rules and their automaton to the java backend definition. This won't interfere with regular rule evaluation because regular rules have their own automaton and cannot match spec rules, even if they are not otherwise distinguished in the definition.

Most of the work was various refactorings/code cleanup to make this possible.

I also removed the cache from `InitializeRewriter.InitializeDefinition`, because `InitializeDefinition` is a request-based object managed by Guice, so it will be instantiated anew each time it is requested. Cache is thus useless.

@dwightguth ping

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
@@ -27,6 +27,7 @@ import org.kframework.utils.errorsystem.KExceptionManager;
 import scala.collection.JavaConversions;
 import scala.collection.JavaConverters;
 
+import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


denis-bogdanas:
Not ready for review yet. Issues at VSC end.

denis-bogdanas:
@ehildenb I preserved all the refactorings but disabled and commented the WIP code on spec automaton. I believe it is useful to merge the work in this form until/in case it will be completed.

ehildenb:
Waiting on https://github.com/runtimeverification/verified-smart-contracts/pull/247

denis-bogdanas:
@ehildenb VSC tests passed. Looks like previous OOM error was flaky.

ehildenb (approved):

PR #543, opened 2019-06-12
prelude.kore: kseq and dotk are _functional_ constructors

ttuegel:
It #535, I marked the Kore symbols `kseq` and `dotk` as constructors, but due to a testing error, I missed the fact that they actually need to be `functional` constructors.


traiansf (approved):
LGTM

PR #544, opened 2019-06-12
K read additions

gtrepta:
Added --input flag to krun.


ehildenb:
@gtrepta this looks good to me, but I would like @dwightguth to also weigh in because he seemed to have second thoughts about adding `--input` to KRun.

ehildenb:
After discussing with @dwightguth, it doesn't make sense to add a KRead to KRun, because KRun has it's own parsing logic because it needs to operate as an interpreter for the surface language.

Instead, we should manually add another branch to `externalParse` which attempts to parse the input as JSON, and if it succeeds returns that. Note that all the functionality of having `--input JSON` can be achieve the same way with `--parser cat` and the check for JSON input in `externalParse`.

Additionally, we still will add a `KRead` to `KProve`, and a JSON deserializer for `Definition`, so that KProve can take programmatic JSON input.

ehildenb:
Closing this as outdated.

PR #545, opened 2019-06-13
KBMC

yzhang90:


kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
```diff
@@ -0,0 +1,58 @@
...
+@RequestScoped
+public class KBMCOptions {
+
+    @ParametersDelegate
+    public transient GlobalOptions global = new GlobalOptions();
+
+    @ParametersDelegate
+    public DefinitionLoadingOptions definitionLoading = new DefinitionLoadingOptions();
+
+    @Parameter(names={"--spec-file"}, description="Path to the specification file.")
```
dwightguth: let's call this --raw-spec and give it a description like "Path to a file containing the patterns to model-check. These patterns will be executed as-is without inserting any initial configuration."

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
```diff
@@ -0,0 +1,58 @@
...
+    public PrintOptions print = new PrintOptions();
+
+    @Parameter(names={"--spec-module", "-sm"}, description="Name of module containing specification to model check.")
+    public String specModule;
+
+    @Parameter(names={"--def-module", "-m"}, description="Name of module containing definition to model check under.")
+    public String defModule;
+
+    @Parameter(names="--depth", description="The maximum number of computational steps to model check.")
+    public Integer depth;
```
dwightguth, line 54: you seem to have forgotten to pass this anywhere.

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Parameter(names="--graph-search", description ="Search order of the execution graph. " +
```
dwightguth, line 56: let's just make this be --bfs or --dfs with some sensible default.

dwightguth (changes requested):
needs files added to k-distribution/src/main/scripts/bin and k-distribution/src/main/scripts/distro-bin, plus an integration test that actually BMC's something.

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
yzhang90: For the haskell backend, if user doesn't provide the `--graph-search` order at all, the default one is breadth first search. Apart from bfs and dfs, in the future, we may have other heuristic search order.

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
yzhang90: done

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
yzhang90: done

kernel/src/main/java/org/kframework/kbmc/KBMCOptions.java
yzhang90: For the haskell backend, if user doesn't provide the --graph-search order at all, the default one is breadth first search. Apart from bfs and dfs, in the future, we may have other heuristic search order.

dwightguth (approved):

PR #547, opened 2019-06-14
Allow kompile repeated optimization flags

charala1:


dwightguth:
What exactly are you trying to do here? if the goal is to be able to say something like -O3 -O2 and have -O2 override -O3, that's not going to happen here. If the goal is simply to be able to say -O2 -O2, why do we need that to work?

charala1:
I need to allow repeated appearances of -O2 (and other opt flags). Two
reasons: (a) crashing just because one passed two instances of the same
optimization flag is just not expected behavior. (b) the way flags are
handled in the semantics necessitates that we be able to append flags at
will, which leads to (a).

dwightguth:
I'm not sure I understand why two optimization flags need to be added to the same call to kompile. where exactly is this happening?

charala1:
It's not happening now, because we haven't updated the c-semantics
submodule (in rv-match). Once we do, the flag appending happening in the
c-semantics makefiles will reveal this problem. BTW I see how this PR is
not going to work, List<Boolean> expects -O2 <list of {true, false}>, right?

dwightguth:
very likely, yes. I don't understand though; I thought you removed the -O flag from the c semantics makefile? in which case, why will it already exist when we add it via rv-match?

PR #549, opened 2019-06-16
bug of `in` operator in java backend

QianyangPeng:
Adding a new test case to the k framework which will fail with the current java backend.

The `in` operator for `List` returns `false` where it should return `true`.


ehildenb (approved):

ehildenb:
Waiting on confirmation from @QianyangPeng that this fixes the bug we found in KWasm before we move forward with this.

QianyangPeng:
This patch fixes the bug posted on issue https://github.com/kframework/k/issues/550.

PR #554, opened 2019-06-18
add parsing for weakExistsFinally

ana-pantilie:


ehildenb (approved):
Maybe we should change the klabel for `#wEF` to be `#wEF` to match the others.

dwightguth (approved):

PR #555, opened 2019-06-18
Updated kernel/src/main/javacc/Outer.jj

charala1:
* Bugfix to avoid matching the empty string


dwightguth (changes requested):
I'm fairly sure you can just delete all the stuff in the RegEx section. It looks like dead code to me.

PR #556, opened 2019-06-18
Report error when trying to heat a KSequence.

dwightguth:


kernel/src/main/java/org/kframework/compile/checks/CheckHOLE.java
```diff
@@ -0,0 +1,95 @@
...
+                        ResolveStrict.setPositions(components[0].trim(), strictnessPositions, arity, prod);
+                    } else {
+                        for (int i = 1; i <= arity; i++) {
+                            strictnessPositions.add(i);
+                        }
+                    }
+                } else if (components.length == 2) {
+                    ResolveStrict.setPositions(components[1].trim(), strictnessPositions, arity, prod);
+                } else {
+                    errors.add(KEMException.compilerError("Invalid strict attribute containing multiple semicolons.", prod));
```
ehildenb, line 69: This is already checked in `ResolveStrict.java` above. Is it possible to only have it checked once?

ehildenb (approved):

PR #557, opened 2019-06-18
Pyk library: Library for handling frontend JSON KAST terms programatically

ehildenb:
This introduces the pyk library, which is being used by (1) the evm-coverage repo, and (2) the beacon-chain repo. Currently there are PRs into those repos waiting on this PR.

I have added a simple test-harness as well, and tested that the pyk library can generate well-formatted K pretty format terms that `kast` can parse. We can add more tests as features/bugs are uncovered I figure.


k-distribution/src/main/scripts/lib/py-kframework/kast.py
```diff
@@ -0,0 +1,228 @@
...
+                i += 1
+        return " ".join(result)
+    return _underbarUnparsing
+
+K_builtin_labels = { klabelRewrite : binOpStr("=>")
+                   , klabelCells   : (lambda *args: "\n".join(args))
+                   , klabelEmptyK  : (lambda : ".")
+                   }
+
+BOOL_expressions = { "_andBool__BOOL" : binOpStr("\nandBool")
```
dwightguth: this is not robust. relying on the klabels generated for productions that don't have the klabel and symbol attributes is going to be fragile and in particular it will break in the llvm and haskell backends. if you want to refer to all of these labels explicitly, let's give them explicit klabels and the symbol attribute.

k-distribution/src/main/scripts/lib/py-kframework/pyk.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python3
+
+import sys
+import subprocess
+
+from kast import *
+
+def match(pattern, kast):
```
dwightguth, line 8: what is this being used for? it doesn't seem to deal with sorts very intelligently.

ehildenb: No, it certainly doesn't. It's not meant to be a robust pattern matcher.

It's used for doing simple rewrites/manip over the AST before handing it over to K. Often it's easier to specify the desired behavior as a rewrite (which uses the matcher) than to manually traverse/update the term.

This is used in function `normalizeKLabels`, for example, which does things like replacing `true and TERM` with just `TERM`.

k-distribution/src/main/scripts/lib/py-kframework/kast.py
ehildenb: I've added explicit `klabel` and `symbol` attributes for many parts of `domains.k`, at least all the parts that `pyk` understands for now.

dwightguth (changes requested):
I still see several references to outdated _BOOL klabels.

ehildenb:
I've added a more useful testing harness, which catches this issue with `normalizeKLabels` having the incorrect klabels from the `domains.k`.

I've also added `symbol` to the attributes which `Production` uses when testing `equals`.

dwightguth (approved):

PR #558, opened 2019-06-18
autoDeserialize in KRead

gtrepta:
Can now do `krun --parser "cat" _.{json,kast,bin}`


ehildenb (dismissed):
I'm approving, but please don't merge until @dwightguth also approves.

kernel/src/main/java/org/kframework/parser/KRead.java
```diff
@@ -51,4 +58,21 @@ public static K deserialize(String stringToParse, InputModes inputMode) {
                 throw KEMException.criticalError("Unsupported input mode: " + inputMode);
         }
     }
+
+    public static K autoDeserialize(byte[] kast, Source source) {
+        if (BinaryParser.isBinaryKast(kast))
+            return BinaryParser.parse(kast);
+
+        try {
```
dwightguth, line 71: can we just check that the first non-whitespace character in the input is '{'? instead of trying to parse it, failing, and then trying to parse it again differently? it will be a lot faster that way.

dwightguth (changes requested):

ehildenb (approved):

kernel/src/main/java/org/kframework/parser/KRead.java
gtrepta: Sure, that makes more sense.

dwightguth (approved):

PR #561, opened 2019-06-19
Integrated KORE output into KPrint

traiansf:


kernel/src/main/java/org/kframework/unparser/KPrint.java
```diff
@@ -128,7 +137,17 @@ public void prettyPrint(Definition def, Module module, Consumer<byte[]> print, K
...
+            }
+            case KORE:
+                if (compiledDefinition.isPresent()) {
+                    CompiledDefinition cdef = compiledDefinition.get();
+                    ModuleToKORE converter = new ModuleToKORE(module, files, cdef.topCellInitializer, kompileOptions);
+                    K result = ExpandMacros.forNonSentences(module, files, kompileOptions, false).expand(orig);
+                    result = new AddSortInjections(module).addInjections(result);
+                    converter.convert(result);
+                    return converter.toString().getBytes();
+                }
```
ehildenb, line 144: We should have an `else` branch which throws a `KEMException.criticalError` so that devs can catch issues in invocations of this easier/earlier.

ehildenb (approved):

kernel/src/main/java/org/kframework/unparser/KPrint.java
traiansf: I thought that was unnecessary because if the condition doesn't hold it will fall through to the default case which throws an exception.

However, I agree this make the program logic harder, therefore I added another throw locally.

ehildenb: I wasn't aware that Java will cause a fall-through if the condition doesn't hold. If it's the case maybe we don't need the extra branch then, but I can make that refactor in a follow up PR.

PR #562, opened 2019-06-19
made Rewriter::prove return RewriterResult

traiansf:
To allow for kprove to report the actual exit code from the backend ( https://github.com/kframework/kore/issues/718 )I've refactored the prove API to return a `RewriterResult` instead of a `K`.


kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -59,16 +60,9 @@ public int run(KProveOptions options, CompiledDefinition compiledDefinition, Bac
...
-            KApply kapp = (KApply) results;
-            exit = kapp.klabel().name().equals("#True") ? 0 : 1;
-        } else {
-            exit = 1;
-        }
-        kprint.prettyPrint(compiled._1(), compiled._1().getModule("LANGUAGE-PARSING").get(), s -> kprint.outputFile(s), results);
-        return exit;
+        RewriterResult results = rewriter.prove(specModule, boundaryPattern);
+        kprint.prettyPrint(compiled._1(), compiled._1().getModule("LANGUAGE-PARSING").get(), s -> kprint.outputFile(s), results.k());
+        return results.exitCode().orElse(-1);
```
dwightguth: 255 is a bad choice of exit code for failure here. Can we make it 113?

dwightguth (approved):

PR #568, opened 2019-06-20
Fix regression involving strategies

### LINKED ISSUE #567: Exception: "Attempting to split non-cell term" when trying to kompile IELE
Hi!

I have been working on IELE VM for quite some time now. I am building a Go backend inspired by the OCaml one.

Today I decided to try to merge my work the latest version from K Framework master, and I noticed that it doesn't kompile anymore. This exception is not related to the backend, I could reproduce it on the OCaml backend, and occurs before any backend code gets called.

It doesn't seem to be a problem in IELE, since nothing changed on that project since my last merge 2-3 months ago. From what I was able to debug, it seems to be a bug when cells are being sorted, during kompile.

The error is as follows:

```
Compiler: Attempting to split non-cell term
isKResult(`_=shift_,__IELE-COMMON`(K0,K1,K2))
  while executing phase "concretizing configuration" on sentence at
	<none>
	<none>
```

it occurs in SortCells.java, line 444

dwightguth:
Fixes #567


k-distribution/tests/regression-new/issue-567/test.k
```diff
@@ -0,0 +1,16 @@
+module TEST-COMMON
+  imports INT-SYNTAX
+  imports BOOL
+
+  syntax ShiftInst ::= Int "=" "shift" Int "," Int [hybrid, seqstrict(2,3)]
```
h0nzZik, line 5: What does `hybrid` mean? Just curious.

dwightguth: hybrid introduces an axiom saying that if the term has been completely heated, then it is itself a KResult.

ehildenb (approved):

PR #571, opened 2019-06-25
Initial support for set variables

traiansf:


dwightguth (approved):
Looks good, but let's finish discussing the syntax on slack before merging this.

ehildenb (changes requested):
We need a check that axioms with set variables are not fed to the Java backend.

We also need to document in `pending-documentation.md` this feature.

ehildenb (approved):

nishantjr:
Could we have a small example of usage in the documentation please

traiansf:
Related to this: https://github.com/kframework/kore/issues/729

traiansf:
> Could we have a small example of usage in the documentation please

@nishantjr the latest commit adds an example. Hope that is helpful.

ehildenb:
@traiansf should the axioms for `/Int` in `domains.k` be updated to use set variables?

pending-documentation.md
```diff
@@ -197,6 +197,50 @@ This will rewrite I to . if and only if the state cell contains I + 1.
...
+However, it is sometimes useful to have simplification rules which work over any kind of pattern, be it undefined or
+non-deterministic.  This behavior can be achieved by using set variables to stand for any kind of pattern.
+
+### Syntax
+
+Any variable prefixed by `@` will be considered a set variable.
+
+### Example
+
+Below is a simplification rule which motivated this extension:
```
grosu, line 221: This is an interesting rule but it was not what motivated the extension.  There are many and complex reasons for the extension, which we do not need to get into here.  I would remove the "which motivated the extension" part.

pending-documentation.md
```diff
@@ -197,6 +197,50 @@ This will rewrite I to . if and only if the state cell contains I + 1.
...
+
+```
+  rule #Ceil(@I1:Int /Int @I2:Int) =>
+    {(@I2 =/=Int 0) #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) 
+    [anywhere]
+```
+
+This rule basically says that `@I1:Int /Int @I2:Int` is defined if `@I1` and `@I2` are defined and
+`@I2` is not 0. using sets variables here is important as it allows the simplification rule to apply
```
grosu, line 230: using => Using

pending-documentation.md
```diff
@@ -197,6 +197,50 @@ This will rewrite I to . if and only if the state cell contains I + 1.
...
+_any_ symbolic patterns, without caring whether they are defined or not. This allows simplifying
+the expression `#Ceil((A:Int /Int B:Int) / C:Int)` to
+`{(C =/=Int 0) #Equals true} #And #Ceil(C) #And ({(B =/=Int 0) #Equals true} #And #Ceil(B) #And #Ceil(A)` 
+
+See [kframework/kore#729](https://github.com/kframework/kore/issues/729) for more details.
+
+
```
grosu, line 237: @xc93 can you provide a small and simple example where set variables match actual non-functional patterns?  Maybe the (Lin) axiom that defines LTL?  Or something about maps or SL?

pending-documentation.md
ehildenb: These are the rules which motivated the extension to K (perhaps not the same extension to matching logic). These are rules we need to get the Haskell backend to prove several of the single-opcode KEVM proofs. I think as pending documentation for K, it's appropriate to use K rules motivations.

ehildenb:
Merging this so we don't get bogged down in perpetual review. Feel free to open new PRs which add documentation.

PR #572, opened 2019-06-26
Poly bracket parsing issue in definitions

ehildenb:
This demonstrates how rules which use `( ... )` productions where it could also be interpreted as `syntax K ::= "(" K ")" [bracket, poly(0,1)]` end up being parsed incorrectly even when the alternative production is marked as `prefer`.


ehildenb:
The program parser adds the empty list syntax `""` as allowed for empty lists, but the rule parser does not because it create horrible ambiguities in the grammar. This is a known issue, and workarounds are suggested. In this case, you can add a production:

```
syntax Instr ::= "(" PlainInstr ")" [prefer]
```

To be able to explicitly have the empty case for parsing in rules.

PR #573, opened 2019-06-26
Refactor KILToKORE to use new `FlatModule` data-structures in outer-ext.scala

ehildenb:


kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -311,6 +359,45 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
...
+
+object FlatModule {
+  def apply(name: String, unresolvedLocalSentences: Set[Sentence]): FlatModule = {
+    new FlatModule(name, Set(), unresolvedLocalSentences, Att.empty)
+  }
+}
+
+case class FlatModule(name: String, imports: Set[String], localSentences: Set[Sentence], @(Nonnull@param) val att: Att = Att.empty)
+  extends OuterKORE with Sorting with Serializable {
+}
```
ehildenb: Let's move these to `outer-ext.scala`.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -84,6 +84,54 @@ object Module {
...
+    var newModule = new Module(mainMod.name, importedModules, nonSyntaxItems, att)
+
+    newSyntaxModule.checkSorts()
+    newModule.checkSorts()
+
+    koreModules += ((newModule.name, newModule))
+    koreModules += ((newSyntaxModule.name, newSyntaxModule))
+
+    newModule
+  }
```
ehildenb: Move this code to `FlatModule` class in `outer-ext.scala`, call it `toModule`, and use `this` as the first argument (remove first argument).

ehildenb:
Also check if `Configuration` code in `outer-ext.scala` is dead-code, and remove it if so.

kore/src/main/scala/org/kframework/definition/outer.scala
gtrepta: `object Import` as well?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -311,6 +359,45 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
...
+}
+
+case class FlatModule(name: String, imports: Set[String], localSentences: Set[Sentence], @(Nonnull@param) val att: Att = Att.empty)
+  extends OuterKORE with Sorting with Serializable {
+}
+
+object Import {
+  val syntaxString = "$SYNTAX"
+
+  def isSyntax(name: String): Boolean = name.contains("$SYNTAX")
```
ehildenb: Let's use `name.endsWith(syntaxString)`.

kore/src/main/scala/org/kframework/definition/outer.scala
ehildenb: I think `object Import` is useful even for non-ext modules.

ehildenb: So lets keep it here.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -311,6 +311,35 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
...
+      })
+    (name, flatModules)
+  }
+
+}
+
+object Import {
+  val syntaxString = "$SYNTAX"
+
+  def isSyntax(name: String): Boolean = name.endsWith("$SYNTAX")
```
ehildenb: @gtrepta why are we using `"$SYNTAX"` here instead of `syntaxString` (the `val` declared two lines above)?

gtrepta: Oh, that's a leftover from debugging.

dwightguth (approved):

PR #589, opened 2019-07-02
ocaml-backend kompile: test for a missing kompile error. +Int in LHS

denis-bogdanas:
Need something similar for LLVM.


traiansf (approved):

denis-bogdanas:
This error is only for concrete backends. In Java & Haskell matching on functions should be allowed.

dwightguth (approved):

PR #592, opened 2019-07-03
Non-recursive and recursive macros, plus macros applied to tokens.

dwightguth:
This should change the default behavior of macro-expansion to not recursively apply a macro within its own expansion. It will also allow you to write macro rules that rewrite a variable of a particular sort nonrecursively, which is useful when parsing custom constant syntax.


kore/src/main/scala/org/kframework/Strategy.scala
```diff
@@ -63,7 +63,7 @@ class Strategy(heatCool: Boolean) {
 
           import rich._
           
-          if (!defn.mainModule.importedModuleNames.contains("STRATEGY$SYNTAX") || r.att.contains("anywhere") || r.att.contains("macro")) {
+          if (!defn.mainModule.importedModuleNames.contains("STRATEGY$SYNTAX") || r.att.contains("anywhere") || r.att.contains("macro") || r.att.contains("alias") || r.att.contains("macro-rec") || r.att.contains("alias-rec")) {
```
ehildenb, line 66: I suppose we can't use `ExpandMacros.isMacro` here because that lives in `kernel`.

ehildenb (approved):

kore/src/main/scala/org/kframework/Strategy.scala
dwightguth: Correct. Really this code should be rewritten in java and live in the kernel, but that's a bigger change than I want to make right now.

PR #597, opened 2019-07-08
Tests for nested #fun. Issues with LLVM backend

denis-bogdanas:
Tests for nested #fun, incremental increasing of complexity. Work for ocaml, the last test (commented) fails on LLVM backend.


kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -186,7 +186,7 @@ private K visitChildren(K term, Sort actualSort) {
                         List<Sort> polySorts = positions.stream()
                                 .map(p -> sort(kapp.items().get(p - 1), freshSortParam))
                                 .collect(Collectors.toList());
-                        expectedSort = lub(polySorts, null, kapp);
+                        expectedSort = lub(polySorts, null, kapp, mod);
```
traiansf, line 189: I don't get it: where is the extra argument `mod` being used?

traiansf (approved):
Looks good to me; I just want some clarification.

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
dwightguth: it's being used in upperBounds.

dwightguth: and in the Collection overload of lub.

PR #614, opened 2019-07-15
Changes for packaging KEVM on Arch/Ubuntu

ehildenb:


ehildenb:
This updates the PKGBUILD to be more self-contained. Currently there is an issue with `stack` on Arch where it can't install any GHC versions (or perhaps some specific versions) https://github.com/commercialhaskell/stack/issues/4935, but the PKGBUILD worked once for me. `makepkg -si` should work on a completely clean arch install now, and doesn't place anything in the users home directory anymore (from setting `CARGO_HOME` and `RUSTUP_HOME` before running commands).

ehildenb:
It seems that the issue goes away with `stack upgrade`, which our `Dockerfile` does anyway (gets the latest stack). So I think this should be good to merge, I did `makepkg -si`, it fails at the `mvn package` stage (because of the above posted issue on `stack`), then I do `stack upgrade`, then `makepkg -si` goes through.

PR #618, opened 2019-07-17
Updated Dockerfile.debian

charala1:
- apt source and ppa for llvm-8 and openjdk-11
- user creation akin to match and c-semantics
- code alignment


dwightguth (changes requested):
Most of these changes look fine but unfortunately it's going to break the deploy stage to be using jdk 11, so please revert those changes.

PR #623, opened 2019-07-17
Add the hasDomainValues attribute to kore

virgil-serbanuta:


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -114,6 +120,10 @@ public String convert(boolean heatCoolEq) {
         collectionSorts.add("MAP.Map");
         collectionSorts.add("LIST.List");
         collectionSorts.add("ARRAY.Array");
+        if (attributes.containsKey("token")) {
+            assert(!attributes.containsKey(HAS_DOMAIN_VALUES));
```
dwightguth: I'm reasonably sure this assertion could fail if the user added the attribute to their definition. In general it should be considered a bug for any k definition to generate an assertion failure when compiled

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -137,6 +147,9 @@ public String convert(boolean heatCoolEq) {
                     sb.append("hooked-");
                 }
             }
+            if (tokenSorts.contains(sort)) {
+                att = att.add(HAS_DOMAIN_VALUES, "");
```
dwightguth: Don't think you need the second parameter here

dwightguth (changes requested):
Looks good but a few things need to be improved

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
virgil-serbanuta: Fixed.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
virgil-serbanuta: Fixed.

dwightguth (approved):

PR #634, opened 2019-07-22
The test case about `String2Base` function under llvm backend.

QianyangPeng:
Currently with llvm backend `String2Base` function would fail if the string is in the format of `"+N"`, e.g. "+123", with error message:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Not a valid integer
```
But under other backend, such as java, "+123" would be parsed as Int: 123


mickyabir:
Fix is ~under review~ merged: https://github.com/kframework/llvm-backend/pull/191

ehildenb (approved):

PR #641, opened 2019-07-24
missing bytes hooks

dwightguth:


k-distribution/include/builtin/domains.k
```diff
@@ -731,6 +733,13 @@ module BYTES-IN-K [symbolic]
   rule String2Bytes(S) => ordChar(substrString(S, 0, 1)) : String2Bytes(substrString(S, 1, lengthString(S))) requires lengthString(S) >=Int 1
   rule String2Bytes("") => nilBytes
 
+  syntax Bytes ::= Bytes "[" Int "<-" Int "]" [function]
+  rule BS [ N <- M ] => substrBytes(BS, 0, N) +Bytes M : substrBytes(N +Int 1, lengthBytes(BS))
+
+  syntax Int ::= Bytes "[" Int "]" [function]
+  rule (B : BS) [ 0 ] => B
+  rule (_ : BS) [ I ] => BS [ I -Int 1] [owise]
```
ehildenb: Let's use `rule (_ : BS) [ I ] => BS [ I -Int 1 ] requires I >Int 0` instead, it will behave better for symbolic execution, and will keep us from decrementing indefinitely for negative numbers.

ehildenb (changes requested):

ehildenb (approved):

PR #655, opened 2019-07-29
Packaging updates

ehildenb:
-   Get packaging message from `INSTALL.md` instead of hard-coded string in `Jenkinsfile`.
-   Flesh out build and runtime dependencies for Arch, Ubuntu, and Debian Buster.
-   Switch `PKGBUILD` to a `git` VCS based build, and follow guidelines here about how to do that: https://wiki.archlinux.org/index.php/VCS_package_guidelines
-   Explicitly run `install-rust` in setup phase for packaging, as a fresh-vm has not done this.


INSTALL.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Ubuntu/Debian
+
+Install the package with (`X.Y.Z` is version number, `ID` is platform identifier):
+
+```sh
+sudo apt install ./kframework_X.Y.Z_amd64_ID.deb
+```
+
+On Debian Stretch, you also must first enable stretch-backports by running:
```
dwightguth: we support buster now, not stretch, and on buster this is not true.

INSTALL.md
```diff
@@ -0,0 +1,78 @@
+Installing Release Builds
+=========================
+
+These instructions explain how to download, install, and build the K packages.
+Current supported systems are:
+
+-   Arch Linux
+-   Ubuntu Bionic (18.04)
+-   Ubuntu Xenial (16.04)
```
dwightguth: we do not support xenial anymore.

dwightguth (changes requested):
I'm not sure I like this git stuff either. When we do a stable release, it's going to be based on the source tarball that doesn't even have a .git directory, so git submodule update will fail. let's leave source control to source control and the packaging system for building and packaging.

I also have a couple other comments.

INSTALL.md
ehildenb: Removed.

INSTALL.md
ehildenb: Updated.

dwightguth (approved):

PR #657, opened 2019-07-29
update submodule

dwightguth:


dwightguth:
@ehildenb I am merging this, but it's just a submodule update so that's pretty much within existing policy regarding code reviews.

PR #658, opened 2019-07-30
Installed pinned version of Stack from Haskell backend

ttuegel:
This should take care of the Docker cache problem. :crossed_fingers:

### DIFF
--- a/Dockerfile.arch
+++ b/Dockerfile.arch
@@ -27,7 +27,7 @@ ENV LC_ALL=C.UTF-8
 ADD --chown=user:user haskell-backend/src/main/native/haskell-backend/stack.yaml /home/user/.tmp-haskell/
 ADD --chown=user:user haskell-backend/src/main/native/haskell-backend/kore/package.yaml /home/user/.tmp-haskell/kore/
 RUN    cd /home/user/.tmp-haskell \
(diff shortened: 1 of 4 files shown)


Dockerfile.arch
```diff
@@ -21,10 +21,12 @@ RUN    cd /home/user \
     && ./.tmp-opam/bin/k-configure-opam-dev
 
 ENV LC_ALL=C.UTF-8
+ADD haskell-backend/src/main/native/haskell-backend/scripts/install-stack.sh /home/user/.install-stack/
+RUN /home/user/.install-stack/install-stack.sh
```
ehildenb: Run as root before the `useradd` and `USER` directives.

Dockerfile.debian
```diff
@@ -22,10 +21,12 @@ RUN    cd /home/user \
     && ./.tmp-opam/bin/k-configure-opam-dev
 
 ENV LC_ALL=C.UTF-8
+ADD haskell-backend/src/main/native/haskell-backend/scripts/install-stack.sh /home/user/.install-stack/
+RUN /home/user/.install-stack/install-stack.sh
```
ehildenb: same

ehildenb (changes requested):

ehildenb (approved):

PR #663, opened 2019-08-02
Use dockerhub images

charala1:
* Uses the `runtimeverification/kframework:*` base images (currently, debian-only)
* Pulls rust, ocaml and haskell from other images (currently, debian-only)
* Renames dockerfiles to the `*.Dockerfile` convention


charala1:
NOTE: unfortunately, `COPY --from` does not support variables, so we cannot say `COPY --from=runtimeverification:kframework:${OS}-${FLAVOR}`.

dwightguth (changes requested):
Please run a full build as it is obvious this would not pass the complete test suite.

charala1:
> Please run a full build as it is obvious this would not pass the complete test suite.

Why is it obvious? What did I miss

dwightguth:
You renamed the arch dockerfile but did not rename any references to it. Regardless, as a change to the CI system, it /should/ run a full build.

dwightguth:
Also, where do the dockerfiles live that build these dependencies?

charala1:
> You renamed the arch dockerfile but did not rename any references to it. Regardless, as a change to the CI system, it /should/ run a full build.

I didn't ask you to use admin access to merge it as-is. There has been a lot of work under the hood that would benefit from some input, that's all.

Anyway, I fixed the `arch.Dockerfile` issue, let me know what you think.

charala1:
> Also, where do the dockerfiles live that build these dependencies?

They are in this repository: https://github.com/runtimeverification/dockerfiles

charala1:
It's worth mentioning that these have been successfully used for a while now. Notably, the trufflecon demo on evm-semantics, and the open source c-semantics.

PR #676, opened 2019-08-08
Unparse set vars from Kore to K

traiansf:
This should be merged after PR #675


kore/src/main/java/org/kframework/parser/KoreParser.java
```diff
@@ -39,7 +39,7 @@ public K parseString(String koreString) {
             Pattern kore = textToKore.parsePattern(koreString);
             return koreToK.apply(kore);
         } catch (ParseError parseError) {
-            throw KEMException.criticalError("Parse error" );
+            throw KEMException.criticalError("Parse error.\n" + parseError.msg() );
```
dwightguth: This should probably chain the exception also.

dwightguth (changes requested):

kore/src/main/java/org/kframework/parser/KoreParser.java
traiansf: It's actually better this way; and it doesn't need the message anymore.

dwightguth (approved):

PR #680, opened 2019-08-09
Smtlibs int - add SMTLIB translation for bitwise Int operations

ehildenb:


dwightguth (approved):

daejunpark:
@ehildenb what specific problems are you trying to solve with this?

I know there will be some warning messages without the smtlib attribute. Is that what you're trying to solve here, or is there another issue?

Indeed, the java backend has a variable-abstraction mechanism that effectively does what you're trying to achieve with this. Or, is that for the haskell backend?

BTW, in the documentation, `smtlib` is used not only for the uninterpreted function, but also for builtin functions, e.g., `smtlib(+)` for `+Int`.

ehildenb:
@daejunpark yes several "Missing smtlib translation for errors" in the proofs I'm trying to fix. Also, `+Int` gets `smt-hook(+)`, not `smtlib(+)`.

Does it hurt to have the extra `smtlib` symbols?

daejunpark:
Now I see `smt-hook`.

Having smtlib wouldn't necessarily hurt, but I'm not sure that it wouldn't affects the existing proofs.  It would be great if we can run the verified-smart-contract CI with this update.

daejunpark:
@ehildenb Started the VSC CI against this.
https://github.com/runtimeverification/verified-smart-contracts/pull/250

If fails, I'll try to cherry-pick only this change and re-run to see the net effect.

Please let me know if you have a better idea of doing this test.

PR #682, opened 2019-08-12
fix bug involving strategies

dwightguth:


kore/src/main/scala/org/kframework/Strategy.scala
```diff
@@ -45,6 +47,12 @@ object Strategy {
   }
 }
 
+class ContainsSCell extends ExistsK {
+  override def apply(k: KApply): java.lang.Boolean = {
+    k.klabel.name == "<s>"
```
ehildenb: I think we may have a `KLabel.STRATEGY_CELL` somewhere you could use? Can you double-check?

ehildenb (approved):

PR #684, opened 2019-08-13
tests for #fun: One more failure case required for Beacon Chain.

denis-bogdanas:


dwightguth:
Unfortunately I cannot make this work as by definition a function that requires the configuration cannot be invoked before the configuration has been constructed. I can try to give a better error message in this case if you'd like, however?

denis-bogdanas:
I don't understand. Isn't the configuration constructed from the beginning and available to both functions, both #fun and the function of the rule?

dwightguth:
Your initial configuration contains a call to `funIntListAndConfig()`. This function needs to match on the `<cell>` cell, but this cell does not exist yet when the function is being evaluated, because the initial configuration is still being constructed. So it cannot evaluate this function because it does not have a configuration to pass as an implicit parameter to the function. Thus, you get some kind of failure from the backend. We can definitely make it a more friendly failure, but it is not possible to evaluate this function until after the initial configuration has been constructed and a `<cell`> cell exists.

denis-bogdanas:
The error appears during kompile, so it's not related to `$PGM`. I changed the test so that `funIntListAndConfig()` is called indirectly.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
```diff
@@ -372,11 +372,20 @@ public String toString() {
 
     public static boolean isFunctionRule(TermCons tc, boolean isAnywhere) {
         if ((tc.production().sort().name().equals("#RuleContent") || tc.production().sort().name().equals("#RuleBody")) && !(tc.get(0) instanceof TermCons && isFunctionRule((TermCons)tc.get(0), isAnywhere))) {
+            if (tc.get(0) instanceof Ambiguity) {
+              return false;
+            }
```
ehildenb, line 378: Why is it the case that if there is an `Ambiguity` we know this isn't a function rule? Seems to me we could have ambiguities in function rules too.

dwightguth: We don't; but when we don't know whether or not it's a function rule, we don't know that it's safe to apply the typing restrictions associated with function rules, so we don't apply them. it's a "degrade gracefully" kind of deal. it seems that the ambiguities in this example are resolved later in the ambiguity pipeline though.

I might be able to implement a case where if all of the ambiguous possibilities are function rules, we also return true. But at worst this is going to fail to catch a type error rather than just crashing, so it seems to me like an improvement.

dwightguth: We could also always return true in this case, in which case you would be guaranteed not to miss type errors, but in some cases might be unable to write certain (correct) rules. This seemed like a better solution to me...

ehildenb: Let's at least put a TODO here marking that it's a known issue, and (potentially?) add a card to the Parser column of our board mentioning it: https://github.com/kframework/k/projects/2#column-5804675

ehildenb (approved):

denis-bogdanas:
@dwightguth Seems to work now. Thanks!

PR #691, opened 2019-08-15
[draft] Implement mkstemp in llvm backend

nishantjr:


dwightguth:
This change looks reasonable but we need to fix tests on the llvm backend pr and get it merged first.

PR #695, opened 2019-08-16
[DRAFT] Use alias for LHS of semantics rules

yzhang90:
All commits marked as [REFACTOR] just break the huge `convert` function into smaller pieces. commit 419e498 passed all the tests.

commit 49727ba7 implements the alias stuff.


dwightguth:
Looks reasonable but it seems tests fail.

PR #700, opened 2019-08-20
refactor brew scripts

dwightguth:
This is done so we can reuse them in kevm.


src/main/scripts/brew-update-to-final
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh -ex
```
ehildenb: Maybe we should also add `-u` since we're relying on environment variables being set.

src/main/scripts/brew-build-bottle
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -ex
```
ehildenb: same

src/main/scripts/brew-update-to-local
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh -ex
```
ehildenb: same

src/main/scripts/brew-update-to-local-bottle
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh -ex
```
ehildenb: same

ehildenb (approved):

Jenkinsfile
```diff
@@ -5,6 +5,11 @@ pipeline {
   options {
     ansiColor('xterm')
   }
+  environment {
+    PACKAGE = 'kframework'
+    VERSION = '5.0.0'
+    ROOT_URL = 'https://github.com/kframework/k/releases/download/nightly'
```
ehildenb, line 11: More descriptive names might be good. Eg `PACKAGE_VERSION`, `PACKAGE_ROOT_URL`. Your choice.

PR #710, opened 2019-08-23
add basic bisimulation test

daejunpark:
An initial effort of implementing keq in kprove.

The idea is to generate verification conditions of checking bisimulation in kprove specs, by using external scripts (or manually for now). This is backend-independent, and can replace the existing keq implementation in java backend, avoiding re-implementing it in the haskell backend.

There are more work need to be done for improving usability (including https://github.com/kframework/k/issues/616), but the core functionality turned out to be supported sufficiently by kprove and strategy (thanks to @ehildenb).


daejunpark:
The current test checks bisimulation between two transition systems that model the essence of the abstract and concrete specs of EVM smart contracts.

`a.k` is an abstract model where it uses a logical map (i.e., direct key-to-value mapping), while `b.k` is a concrete model where it mimics the EVM storage using a symbolic hashed location. Essentially this test shows that we can establish a (bi)simulation relation between such different map structures.

Also, this test involves a loop, showing how the bisimulation proof obligations are encoded in pure kprove specs.

daejunpark:
Issues in parallel composition of definitions (i.e., having a tuple of two configurations, where each configuration evolves independently using the corresponding definition's rules):

1. cell name clash in configuration composition, requiring to rename cells to be unique. also, name clash for klabels and sorts.
2. (due to 1), configuration concretization does not work as expected for any non-function rule with no cell (e.g., `rule I1 + I2 => I1 +Int I2` is problematic, while `rule <k> I1 + I2 => I1 +Int I2 ... </k>` is ok.) and auto-generated heating/cooling rules.
3. no way to assign rule tags to auto-generated rules (such as heating/cooling rules).

k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/main.k
```diff
@@ -0,0 +1,55 @@
...
+configuration <c>
+  <k> $PGM:Pgm </k>
+  <s/>
+  <impT/>
+  <lambdaT/>
+</c>
+
+syntax Bool ::= isSyncPoint(#RuleTag, KItem, K) [function]
+
+rule isSyncPoint(a, K, Ks) => isKResult(K) andBool #isVariable(Ks)
```
denis-bogdanas: I understand most of this spec, except the logic around `a` and `b` where are they defined and why are they needed? Perhaps this is worth commenting.

denis-bogdanas (commented):
I see that `a` and `b` comes from rule tags. I didn't follow enough the discussion about strategies to review this PR.

ehildenb (dismissed):

daejunpark:
Now this contains all the previous keq tests.

Also, I added a new test, checking equivalence between sum-imp and sum-lambda, that is:
```
    int n, sum;
    n = N;
    sum = 0;
    while (!(n <= 0)) {
      sum = sum + n;
      n = n + -1;
    }
```
vs
```
    letrec f x = if x <= 0 then 0 else (x + (f (x + -1)))
    in (f N)
```
Note that this is quite challenging because the sum-lambda is *not* tail-recursive. See how the `sum` value in IMP is "related" to the continuation in LAMBDA: https://github.com/kframework/k/blob/5eb7b37300edd5a09da9107b09700d49e2bf61dd/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/sum-symbolic-spec.k#L5-L10

daejunpark:
@denis-bogdanas When the strategy feature is enabled, each rule, e.g.,
```
rule <k> l => r ... </k> [tag(a)]
```
is translated to:
```
rule <k> l => r ... </k> <s> ^ a => ~ a ... </s>
```
That is, each rule is applied only when the top of the strategy cell `<s>` is matched with its rule tag. (The default rule tag is `regular` if no `tag` attribute is presented.)

More precisely, the above rule is applied only when the top of the `<s>` cell is `^ a`.  Once a rule with tag `a` is applied, the top of `<s>` cell is rewritten to `~ a`. So, you can define some post-rule-application action using `~ a`.  That's all you need to understand this PR. (Indeed that's all about the strategy feature.)

ehildenb (approved):
Looks good, checked that `make --dry-run` actually runs these tests.

PR #711, opened 2019-08-23
KPrint: print available klables when an invalid klabel is encountered, in exception message

denis-bogdanas:


dwightguth (changes requested):
This would create an error message that is much too long. If you'd like to implement "Did you mean" functionality that suggests the most lexically similar klabel according to some kind of lexicographical distance function, that would be an acceptable change to make.

denis-bogdanas:
That's unfortunate.
Do you really think:
1. Here is the place to implement lexical distance algorithms? Besides, I might want to see all the labels. This really helped me today.
2. Shortness of the message is more important than its informativity.

Eventually long message may be enabled on `--debug`.

denis-bogdanas:
@dwightguth any opinion on my proposal?

denis-bogdanas:
@dwightguth Today is the 2nd time when I needed this functionality. An I need it precisely like here. I want to see all the available options, not just possible typo corrections.

My proposal is to allow long error messages on `--debug`. Does this sound reasonable to you? We already have plenty of long messages in java backend.

ehildenb:
I agree with @dwightguth that this is likely not a good way to solve whatever problem you're having. A better thing to do would be to just write a utility called `list-klabels` in bash which just uses shell commands (grep, cut, etc..) to print all the klabels that exist in `compiled.txt` if you need that.

denis-bogdanas:
That's an unmaintainable hack. I'm surprised you propose it.

A solution @dwightguth agreed on would be an option for `kompile` to print all the labels. Then the error message would point to that option for more info. Considering that we are going to eliminate compilation from kprove, all labels would be available at kompile time.

ehildenb:
Re-open if needed and updated.

PR #719, opened 2019-08-27
fix regression in rule labels

dwightguth:


kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -62,6 +62,7 @@ public KILtoKORE(org.kframework.kil.loader.Context context) {
     public FlatModule toFlatModule(Module m) {
         CheckListDecl.check(m);
         String name = m.getName();
+        moduleName = name;
```
ehildenb: Can we just get rid of the intermediate `name` instead? So say `moduleName = m.getName()` directly, and use `moduleName` later in this?

ehildenb (approved):

PR #725, opened 2019-08-29
[BUGFIX] fix non-config split bug

yzhang90:


ehildenb:
This appears to fix the issue I had, so LGTM, but I still would like @dwightguth to weigh in.

So it seems to me that comparing the modules instead of the names was indeed a bug, but in most cases it was equivalent to comparing the names. Since module names are globally unique, we should be OK to only compare names.

@yzhang90 's explanation:

There is one rule in ETHEREUM-SIMULATION module that introduces fresh variable. m and def.mainModule() are both VERIFICATION module.
Note that def is the definition before the transformation and m is the module under transformation. ETHEREUM-SIMULATION module gets transformed before VERIFICATION module. And then when compares m and def.mainModule(), they will flatten all the dependent modules and compares all the sentences. Thus, the rule in m will have generatedCounter cell but the same rule in def.mainModule() will not. As a consequence, all the code behind will never get executed. Those code generates declarations for generatedTopCell and so on. (edited)

dwightguth (approved):

PR #734, opened 2019-08-31
Infer definition module when proving specifications

ehildenb:
When proving we need two modules, the "spec module" and the "def module". The def module will be taken as axioms to be used in proving the claims in the spec module.

Currently, if the user does not supply the spec module, the filename passed on the command line is used to infer the spec module. This seems reasonable.

If the user does not supply the def module, we currently use the kompiled definitions `mainModule`. This may seem reasonable when no lemmas are being added to the definition, but all our real-world proofs have lemma files in addition.

This PR changes the logic so that if the user does not manually provide `--def-module`, the `mainModule` is not used as the default. Instead, it will inspect the spec module and see if there is a singular import, if so, it will use that imported module as the def module.

In KEVM/KWasm, for example, this would largely remove the need to ever supply `-m MODULE_NAME` on the command line, which has caused at least 2 RV devs to waste a day each trying to make the prover work, and is confusing for new users.

### DIFF
--- a/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-imp/if-const-propa-spec.k
+++ b/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-imp/if-const-propa-spec.k
@@ -16,7 +16,6 @@ endmodule
 
 module IF-CONST-PROPA-SPEC
 
-imports MAIN
 imports SYNC-POINT-CANDIDATES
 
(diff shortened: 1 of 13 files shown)


ehildenb:
Another option would be to:

-   Enforce that `--spec-module` module only imports exactly one module.
-   Use said imported module as the `--def-module`, but allow overriding by explicitely providing the `--def-module` as well.

PR #737, opened 2019-09-03
Pyk updates - fixes to JSON rep, main runner script

ehildenb:
-   Fixes the JSON representation of Module => FlatModule.
-   Respects the K_RELEASE environment variable when invoking K.
-   Adds ability to make an import graph from a given JSON definition.


ehildenb:
Should wait for https://github.com/runtimeverification/beacon-chain-spec/pull/68 to pass tests.

denis-bogdanas (approved):
I don't understand the underlying intention, mostly a formal review.

Question: does this require changes to BC scripts?

ehildenb:
@denis-bogdanas these changes are for the MKR project, but I wanted you to review since you're the only other main user of the pyk library.

It looks like it doesn't require any changes to beacon-chain: https://github.com/runtimeverification/beacon-chain-spec/pull/68

PR #740, opened 2019-09-03
update --debugger to use gdb

dwightguth:


kernel/src/main/java/org/kframework/krun/KRunOptions.java
```diff
@@ -229,14 +225,7 @@ public SearchType searchType() {
...
         @Parameter(names="--debugger", description="Run an execution in debug mode.")
-        private boolean debugger = false;
-
-        public boolean debugger() {
-            if (debugger && search()) {
-                throw new ParameterException("Cannot specify --search with --debug. In order to search inside the debugger, use the step-all command.");
-            }
-            return debugger;
-        }
+        public boolean debugger = false;
```
ehildenb, line 228: Are we intending to not support Java backend debugger anymore? Fine if so, but let's remove that code then.

If we are intending to continue supporting it, then maybe we should leave this guard in?

ehildenb (approved):

PR #741, opened 2019-09-04
Weaker configuration constraints for more flexible composition

ehildenb:
-   Only check that there is a unique top cell when the calling context explicitly requests for the top cell (by moving the check into `getRootCell` and getting rid of variable `topCell`).
-   Rule out the case that there are multiple productions for the same configuration item directly (perhaps this is redundant with a check elsewhere over the syntax?)
-    No need to tree `<k>` and `<s>` cell specially in ConfigurationInformation because multiple top cells allowed.


ehildenb:
Confirmed that this allows us to do the configuration composition we want in runtimeverification/mkr-mcd-spec

dwightguth (approved):

ehildenb:
Testing with KEVM here: https://github.com/kframework/evm-semantics/pull/477

ehildenb:
KEVM passed tests.

PR #746, opened 2019-09-04
Debugger fixes

dwightguth:


debian/rules
```diff
@@ -18,10 +18,10 @@ export DEB_BUILD_MAINT_OPTIONS=hardening=-stackprotector
 	dh $@ 
 
 override_dh_auto_build:
-	mvn package -DskipTests
+	mvn package -DskipTests -Dllvm.backend.prefix=/usr/lib/kframework -Dllvm.backend.destdir=$(shell pwd)/debian/kframework
 
 override_dh_auto_install:
-	DESTDIR=debian/kframework PREFIX=/usr src/main/scripts/package
+	DESTDIR=$(shell pwd)/debian/kframework PREFIX=/usr src/main/scripts/package
```
ehildenb, line 24: Hmmm, this one doesn't make sense to me. Why would we want to use the full path to the directory where we're doing the packaging when building the package? That seems like it could potentially break other people's systems.

pending-documentation.md
```diff
@@ -375,6 +375,28 @@ In the example above, we declare two productions `~Int_` and `_^%Int__`, and tel
 Set variables are currently only supported by the Haskell backend.
 The use of rules with set variables should be sound for all other backends which just execute by rewriting, however it might not be safe for backends which want to guarantee coverage.
 
+## Debugging
```
```suggestion
Debugging
---------
```

pending-documentation.md
```diff
@@ -375,6 +375,28 @@ In the example above, we declare two productions `~Int_` and `_^%Int__`, and tel
...
+
+The LLVM Backend has support for integration with GDB. You can run the debugger on a particular program by passing the `--debugger` flag to krun, or by invoking the llvm backend interpreter directly. Below we provide a simple tutorial to explain some of the basic commands supported by the LLVM backend.
```
ehildenb, line 381: Here you say "Below we provide a simple tutorial", but we seem to be missing the actual tutorial part.

I'm wondering if commits got mixed up between this PR updating the dependencies and another with the debugger tutorial.

debian/rules
dwightguth: it's necessary because the working directory when this variable gets used is actually something different. It is harmless, though, because DESTDIR does not get used after packaging finishes.

pending-documentation.md
dwightguth: it's a work in progress. I can factor out this part of the commit if you want? I just needed to get this merged because it contained necessary fixes and I hadn't finished writing the tutorial yet. But I was planning on finishing this in a followup PR. How would you like me to handle it?

ehildenb (approved):

PR #751, opened 2019-09-05
Remove some mvn warnings

ehildenb:
The remaining warnings are either about the assembly process (copying things into place), or in the llvm-backend repository code.

Also the warning in POSet about using JavaConversions vs JavaConverters, I couldn't figure out in a short amount of time, so I gave up.

### DIFF
--- a/kore/src/main/scala/org/kframework/kore/interface.scala
+++ b/kore/src/main/scala/org/kframework/kore/interface.scala
@@ -3,7 +3,6 @@ package org.kframework.kore
 import java.util.Optional
 
 import org.kframework.attributes._
-import org.kframework.kore.ADT.{KApply, KList}
 import org.kframework.unparser.ToKast
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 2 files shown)


kore/src/main/scala/org/kframework/kore/interface.scala
```diff
@@ -54,6 +53,7 @@ object K {
         case (_, _:KRewrite) => -1
         case (_:InjectedKLabel, _) => 1
         case (_, _:InjectedKLabel) => -1
+        case (_, _) => 0
```
dwightguth: most of this looks good but this is definitely /not/ correct. Two arbitrary terms that don't match any of the above classes are /not/ equal. You should probably raise an error here.

dwightguth (changes requested):

kore/src/main/scala/org/kframework/kore/interface.scala
ehildenb: Now throwing a `KEMException.criticalError`

kore/src/main/scala/org/kframework/kore/interface.scala
```diff
@@ -54,6 +54,7 @@ object K {
         case (_, _:KRewrite) => -1
         case (_:InjectedKLabel, _) => 1
         case (_, _:InjectedKLabel) => -1
+        case (_, _) => throw KEMException.criticalError("Cannot order these terms:\n" + a.toString() + "\n" + b.toString())
```
dwightguth: should probably be an internalError.

dwightguth (approved):

PR #757, opened 2019-09-09
More robust (un)parsing for PRETTY and PROGRAM

ehildenb:
Here we make sure to use the same module for parsing/expanding macros/unparsing, instead of switching between modules. By default, it's the languageParsingModule(), which is what KRun defaults to as well for unparsing.

This is a change in behavior for Kast, but if the tests pass and the beacon-chain repo works better with this commit, I see no reason to not merge it.

### DIFF
--- a/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
+++ b/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
@@ -106,19 +106,17 @@ public class KastFrontEnd extends FrontEnd {
                 }
             }
 
-            Module unparsingMod;
             if (options.module == null) {
-                options.module = def.mainSyntaxModuleName();
                 switch (options.input) {
                     case KORE:
-                        unparsingMod = def.languageParsingModule();
+                        options.module = def.languageParsingModule().name();
                         break;
                     default:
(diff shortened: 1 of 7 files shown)


dwightguth (changes requested):
This will not work as is. `kast $file` is intended to parse a file using the syntax module ONLY. We should not change this behavior even if tests pass, but I'm also not surprised this is failing.

ehildenb:
What needs to happen is (from @dwightguth ):

1. Import K-TERM implicitly if input or output is pretty, because the configuration or string could contain k syntax
2. Call programParsingModuleFor if the input or output is program, so we handle user lists correctly
3. Remove K-TERM from the imports of LANGUAGE-PARSING

This by itself should fix the issue with k syntax, but it still could crash on some inputs, so if we want to fix that, we also do:

4. Pass K-TERM (the module itself) to KOREToTreeNodes. If a production is not found, look up #KApply or #KToken production in the K-TERM module and create a TermCons or Constant with that production and the appropriate transformed children. This will guarantee we always print as best we can without crashing if there is a problem

ehildenb:
Waiting for #759.

PR #759, opened 2019-09-10
Refactor Kast/KRead/KPrint - less pretty printers

ehildenb:


dwightguth:
the changes to KastFrontEnd don't make sense. You are now completely bypassing the process of selecting a default module for the parser. I can't see how tests are going to pass for this.

ehildenb:
@dwightguth not sure what you mean, there are no semantic changes to KastFrontEnd.java, only renaming variables and formatting.

ehildenb:
And an improved error message.

dwightguth (approved):
This looks like it shouldn't have any changes to behavior, which I assume is what you intended. Thus, I approve.

PR #771, opened 2019-09-12
Emit json definitions when kompiling

ehildenb:
-   Improvements to `pyk` serialization to files when running `{krun,kast,kprove}JSON` functions.
-   Add function which splits configuration from substitution to pyk.
-   Add definition JSON serialization/deserialization methods.
-   Add `--emit-json` option to Kompile which emits the json serialized definition to the `-kompiled` directory (and test of it).

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -102,6 +102,9 @@ public class KompileOptions implements Serializable {
         }
     }
 
+    @Parameter(names="--emit-json", description="Emit JSON serialized version of parsed and kompiled definitions.")
+    public boolean emitJson = false;
+
     public static final class Experimental implements Serializable {
 
         // Experimental options
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -98,9 +98,6 @@ public class KompileOptions implements Serializable {
         }
     }
 
-    @Parameter(names="--emit-json", description="Emit JSON serialized version of parsed and kompiled definitions.")
-    public boolean emitJson = false;
-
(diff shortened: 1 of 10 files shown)


kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -83,6 +86,21 @@ public boolean isKore() {
...
+        }
+
+        @Override
+        public Class<OutputModes> enumClass() {
+            return OutputModes.class;
+        }
+    }
+
+    @Parameter(names="--emit-json", description="Emit JSON serialized version of parsed and kompiled definitions.")
+    public boolean emitJson = false;
```
dwightguth: this should probably be an experimental option.

dwightguth (commented):
What exactly is this being used for? I highly doubt it is going to be possible to serialize, deserialize, and then compile or krun using the result. Not without more work to ensure a faithful rendering.

ehildenb:
@dwightguth I'm going to use it to build Python unparsers for KLabels automatically using `pyk`s `underbarUnparsing` function, so that I can get rid of the tons of boilerplate in places like https://github.com/runtimeverification/polkadot-verification/blob/a50120c8493805d4146af4182bfb9dec44d298e0/pykWasm.py#L35 and https://github.com/runtimeverification/beacon-chain-spec/blob/a0975944c17537ecd070051cc08a387b2dda9403/buildConfig.py#L76

I do eventually want to be able to re-load definitions, but I removed the `--input json` option for `komplie` from this PR because it is not working yet. That's lower priority than having the definition available in structured form for external tools.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
ehildenb: Done.

dwightguth (approved):
Looks reasonable if you don't intend to deserialize it in the frontend yet.

PR #774, opened 2019-09-13
Generate projections for named nonterminals

dwightguth:


kernel/src/main/java/org/kframework/compile/GenerateSortProjections.java
```diff
@@ -63,5 +74,34 @@ public static KLabel getProjectLbl(Sort sort, Module m) {
         return stream(Set(Production(lbl, sort, Seq(Terminal(lbl.name()), Terminal("("), NonTerminal(Sorts.K()), Terminal(")")), Att().add("function").add("projection")), r));
     }
 
+    public Stream<? extends Sentence> gen(Production prod) {
+      if (prod.att().contains(Attribute.FUNCTION_KEY)) {
```
ehildenb, line 78: It seems that we'll ignore non-`function` constructors, which are really what I want record projectors for.

Is this correct then?

ehildenb (approved):

PR #789, opened 2019-09-18
Nice freezers for context aliases

h0nzZik:
Previously, the freezers had the shape of
```
#freezer#SemanticCastToK0_ ( 5 )
```
Now they look as usual:
```
#freezer_+__IMP-SYNTAX0_ ( 5 )
```


kernel/src/main/java/org/kframework/compile/ResolveContexts.java
```diff
@@ -176,7 +176,13 @@ public void apply(KApply k) {
         List<ProductionItem> items = new ArrayList<>();
         KLabel freezerLabel;
         if (cooled instanceof KApply) {
-            freezerLabel = getUniqueFreezerLabel(input, ((KApply)cooled).klabel().name() + finalHolePosition[0]);
+            KApply kApply = (KApply)cooled;
+            String name = kApply.klabel().name();
+            K firstArg = kApply.klist().items().get(0);
```
dwightguth: Can we move this declaration inside the name.equals condition? I'm a little worried about the possibility of an IndexOutOfBounds exception.

dwightguth (approved):
Looks good, thanks!

PR #793, opened 2019-09-23
haskell-backend/src/main/native/haskell-backend: update deps

traiansf:
This contains the latest updates for supporting WASM proofs.


ttuegel:
Duplicate of #786.

PR #809, opened 2019-09-30
Remaining json serialization for pyk

ehildenb:
Currently we only support a subset of the JSON nodes which the k frontend can put out for a given definition.

-   Add the remaining nodes (mostly having to do with sentences/productions).
-   Remove the KImport node, which doesn't get its own representation in KAST.
-   Changes `KApply(klabelRewrite, ...) => KRewrite`, since rewrites get their own node in the KAST json format output by the existing frontend.

### DIFF
--- a/k-distribution/src/main/scripts/lib/pyk/kast.py
+++ b/k-distribution/src/main/scripts/lib/pyk/kast.py
@@ -90,7 +90,7 @@ def isKBubble(k):
     return k['node'] == 'KBubble'
 
 def KModuleComment(comment, att = None):
-    return { "node": "KModuleComment", "commonte": comment, "att": att }
+    return { "node": "KModuleComment", "comment": comment, "att": att }
 
 def isKModuleComment(k):
     return k['node'] == 'KModuleComment'
--- a/k-distribution/src/main/scripts/lib/pyk/kast.py
+++ b/k-distribution/src/main/scripts/lib/pyk/kast.py
@@ -83,11 +83,65 @@ def KRule(rule, requires = None, ensures = None, att = None):
 def isKRule(k):
     return k["node"] == "KRule"
 
-def KImport(kimport):
-    return { "node": "KImport", "import": kimport }
+def KBubble(sentenceType, contents, att = None):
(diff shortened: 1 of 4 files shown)


k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -65,26 +65,86 @@ def KToken(token, sort):
...
-def KImport(kimport):
-    return { "node": "KImport", "import": kimport }
+def KBubble(sentenceType, contents, att = None):
+    return { "node": "KBubble", "sentenceType": sentenceType, "contents": contents, "att": att }
+
+def isKBubble(k):
+    return k['node'] == 'KBubble'
+
+def KModuleComment(comment, att = None):
+    return { "node": "KModuleComment", "commonte": comment, "att": att }
```
gtrepta: Is "commonte" the intended spelling here?

k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -95,29 +155,28 @@ def KRequire(krequire):
...
         newAtts = []
-        if kast["atts"] is None:
-            newAtts = atts
+        if kast["att"] is None:
+            newAtts = att
         else:
-            newAtts.extend(kast["atts"])
-        return KRule(kast["rule"], requires = kast["requires"], ensures = kast["ensures"], atts = newAtts)
+            newAtts.extend(kast["att"])
+        return KRule(kast["body"], requires = kast["requires"], ensures = kast["ensures"], att = newAtts)
```
gtrepta: Does the `att` parameter get added to `newAtts` when the else branch is taken?

ehildenb: Hmmm, looks like some bugs in attributes in general slipped through, I've updated.

k-distribution/src/main/scripts/lib/pyk/kast.py
ehildenb: fixed

gtrepta (approved):

PR #811, opened 2019-10-01
Better rule minimization for pyk unparsing

ehildenb:
-   Allows matching through `KRewrite` using `pyk.match`
-   Pushes rewrites down through `KSequence` in `pyk.minimizeRewrites`
-   Removes `#SemanticCast*` nodes in `pyk.minimizeRule`


ehildenb:
Subsumed by #809 #807 #811

PR #816, opened 2019-10-03
llvm: Custom #if+IntList crashes after a certain list range

denis-bogdanas:


dwightguth:
This appears to be a stack overflow. You should make your function tail recursive if it is going to be called in a deeply nested fashion.

denis-bogdanas:
@dwightguth You really don't see any problem here?
1. Replacing recursion with an iterative implementation requires appending to the end of the list - so we get N^2 complexity. And it looks less elegant.
2. Stack overflow at depth 32, with argument an IntList of length <=32, does look like a problem.

dwightguth:
You don't replace it with something iterative, you replace it with something tail recursive. You don't need to append to the end of the list either, you just append to the front and then reverse in the base case. And the stack limit is controlled by the OS so we can't really do anything about that. You can of course increase the stack limit on your computer, but that's a bad solution because it doesn't solve the underlying problem that your program will crash if the input is too big, and it would also fix the problem for your machine only. Stack overflow if you have too deep of recursion is pretty normal for functional languages and we are a functional language. So no, this seems pretty normal to me.

denis-bogdanas:
Fine, let's keep discussion on stack limit. My `ulimit -s` gives 8192 = 8 MB. This program has a recursion depth of 32 and functions have an IntList argument of length <=32. How this cannot fit into 8 MB of stack? This is nowhere near normal.

denis-bogdanas:
@dwightguth ping

dwightguth:
I will investigate. My guess is the step function is taking up a decent amount of stack space, but I'll look into it when I get a chance.

dwightguth:
My apologies, I think I misunderstood why this program was crashing. This program calls keepSmallerThan40 over 4 billion (2^32) times. It's probably not smashing the stack, it's smashing the heap. It's essentially behaving in the same way as a naive recursive implementation of the fibonacci sequence. We would have to implement memoization in the llvm backend to make something like this possible. How high priority is this for you?

denis-bogdanas:
Oh I see, that's horrible. I don't need this anymore. Managed to do all my work with workarounds. Just thought it's an important issue worth reporting. Should we keep this PR?

dwightguth:
If it's not really something you need currently, let's table this for now. We don't really need memoization yet for performance in any of our semantics, so I'd rather wait until it proves necessary. I will make an issue in the llvm backend repo to support the memo attribute, but we should close this PR.

PR #819, opened 2019-10-03
Remove 'poly' attribute

dwightguth:
This now is passing the regression test suite; testing with the tutorial and exercises now. Note that it's still not ready even if it passes those, though; I also want to test against KEVM and the Haskell backend integration tests.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
@@ -118,7 +118,8 @@ public class JavaBackend implements Backend {
         ModuleTransformer convertDataStructureToLookup = ModuleTransformer.fromSentenceTransformer((m, s) -> new ConvertDataStructureToLookup(m, false).convert(s), "convert data structures to lookups");
 
         //Commented steps are (possibly) required for spec automaton generation.
-        return m -> ModuleTransformer.fromSentenceTransformer(new ResolveAnonVar()::resolve, "resolve anonymous varaibles")
+        return m -> ModuleTransformer.fromSentenceTransformer(Kompile::removePolyKLabels, "remove poly klabels")
+                .andThen(ModuleTransformer.fromSentenceTransformer(new ResolveAnonVar()::resolve, "resolve anonymous varaibles"))
                 .andThen(ModuleTransformer.fromSentenceTransformer(s -> new ResolveSemanticCasts(kompileOptions.backend.equals(Backends.JAVA)).resolve(s), "resolve semantic casts"))
                 .andThen(AddImplicitComputationCell::transformModule)
                 .andThen(ConcretizeCells::transformModule)
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -845,7 +845,7 @@ module K-EQUAL
 
   rule K1:Bool ==Bool K2:Bool => K1 ==K K2
 
-  syntax K ::= "#if" Bool "#then" K "#else" K "#fi"   [function, functional, smt-hook(ite), hook(KEQUAL.ite), poly(0, 2, 3)]
+  syntax {Sort} Sort ::= "#if" Bool "#then" Sort "#else" Sort "#fi"   [function, functional, smt-hook(ite), hook(KEQUAL.ite)]
 
   rule #if C:Bool #then B1::K #else _ #fi => B1 requires C
   rule #if C:Bool #then _ #else B2::K #fi => B2 requires notBool C
@@ -869,8 +869,8 @@ module K-REFLECTION
   syntax List ::= #argv() [function, hook(KREFLECTION.argv)]
 
   // Takes as input a string and returns a K term
-  syntax KItem ::= #parseKAST(String) [function, hook(KREFLECTION.parseKAST), poly(0)]
-  syntax KItem ::= #parseKORE(String) [function, hook(KREFLECTION.parseKORE), poly(0)]
+  syntax {Sort} Sort ::= #parseKORE(String) [function, hook(KREFLECTION.parseKORE)]
+  syntax {Sort} Sort ::= #parseKAST(String) [function, hook(KREFLECTION.parseKAST)]
   syntax IOError ::= "#noParse" "(" String ")" [klabel(#noParse), symbol]
 
 endmodule
--- a/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
+++ b/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
@@ -317,7 +317,7 @@ public class AddSortInjections {
         if (term.klabel() instanceof KVariable) {
           throw KEMException.internalError("KORE does not yet support KLabel variables.", term);
         }
-        scala.collection.Set<Production> prods = mod.productionsFor().apply(((KApply) term).klabel());
+        scala.collection.Set<Production> prods = mod.productionsFor().apply(((KApply) term).klabel().head());
         if (prods.size() != 1) {
           throw KEMException.compilerError("Could not find production for KApply with label " + term.klabel(), term);
         }
--- a/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
+++ b/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
@@ -182,11 +182,7 @@ public class AddSortInjections {
             if (prod.params().nonEmpty()) {
                 for (Sort param : iterable(prod.params())) {
                     Sort expectedSort;
-                    Set<Integer> positions = IntStream.range(0, prod.nonterminals().size())
-                                .mapToObj(i -> Tuple2.apply(prod.nonterminals().apply(i), i))
-                                .filter(t -> t._1().sort().equals(param))
-                                .map(t -> t._2())
-                                .collect(Collectors.toSet());
+                    Set<Integer> positions = getPositions(param, prod);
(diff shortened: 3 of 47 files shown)


k-distribution/include/builtin/domains.k
```diff
@@ -859,7 +859,7 @@ module K-EQUAL
 
   rule K1:Bool ==Bool K2:Bool => K1 ==K K2
 
-  syntax K ::= "#if" Bool "#then" K "#else" K "#fi"   [function, functional, smt-hook(ite), hook(KEQUAL.ite), poly(0, 2, 3)]
+  syntax {Sort} Sort ::= "#if" Bool "#then" Sort "#else" Sort "#fi"   [function, functional, smt-hook(ite), hook(KEQUAL.ite)]
```
ttuegel, line 848: Is this going to change the symbol declaration generated in Kore?

traiansf: Not much, if at all.  `Sort` will become a sort variable; in some sense you may say it's bringing the syntax closer to KORE.

traiansf (commented):
These look reasonable to me. They are a bit more verbose, but I find them much easier to parse than the `poly` declarations.

k-distribution/include/builtin/domains.k
ttuegel: Ah, I see, I was mistaken about the generated Kore. I thought this generated a symbol in `SortK`, but in fact the Kore symbol is already parametric.

dwightguth: Yes, it's basically the same symbol in kore you were already getting, but it's more explicitly obvious in the syntax now. Also, users can now name their sort variables instead of them being automatically numbered.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/ApplyTypeCheckVisitor.java
```diff
@@ -35,28 +35,30 @@ public ApplyTypeCheckVisitor(POSet<Sort> subsorts, boolean isAnywhere) {
...
-                    j++;
-                } else {
-                    Term t = tc.get(j);
-                    Sort s = ((NonTerminal) tc.production().items().apply(i)).sort();
-                    Either<Set<ParseFailedException>, Term> rez = new ApplyTypeCheck2(s).apply(t);
-                    if (rez.isLeft())
-                        return rez;
-                    tc = tc.with(j, rez.right().get());
-                    j++;
+        if (tc.production().params().isEmpty()) {
```
ehildenb: What happens when this check fails? Should there be an assertion error? My guess is it should never happen? Or does type checking of parametric productions happen elsewhere?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/PushTopAmbiguityUp.java
```diff
@@ -31,7 +31,7 @@ public Term apply(TermCons tc) {
     private class PushTopAmbiguityUp2 extends SafeTransformer {
         @Override
         public Term apply(TermCons tc) {
-            if (tc.production().klabel().isDefined() && tc.production().klabel().get().equals(KLabels.KREWRITE)) {
+            if (tc.production().klabel().isDefined() && tc.production().klabel().get().head().equals(KLabels.KREWRITE)) {
```
ehildenb, line 34: Not sure why all these `.head()` are being inserted after `.klabel().get()` now. Why would there be multiple `klabel`? One per sort instantiation? How do we know the first is the correct one?

dwightguth: This pr introduces a new method KLabel.head which returns the head of a klabel (ie its name without parameters). It's necessary because for example the klabel associated with the rewrite production has a parameter and thus otherwise will not compare equal to KLabels.KREWRITE.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/ApplyTypeCheckVisitor.java
dwightguth: I had forgotten to rewrite this code when I made a change elsewhere, but the answer to your question is that type checking of parametric productions happens when we type check their parents.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/PushTopAmbiguityUp.java
ehildenb: Ah, I see. Was thinking it was the head of a sequence.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
```diff
@@ -118,7 +118,8 @@ public JavaBackend(KExceptionManager kem, FileUtil files, GlobalOptions globalOp
         ModuleTransformer convertDataStructureToLookup = ModuleTransformer.fromSentenceTransformer((m, s) -> new ConvertDataStructureToLookup(m, false).convert(s), "convert data structures to lookups");
 
         //Commented steps are (possibly) required for spec automaton generation.
-        return m -> ModuleTransformer.fromSentenceTransformer(new ResolveAnonVar()::resolve, "resolve anonymous varaibles")
+        return m -> ModuleTransformer.fromSentenceTransformer(Kompile::removePolyKLabels, "remove poly klabels")
```
ehildenb, line 121: I guess the plan is to just not support definitions with parametric productions on the Java backend?

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -251,15 +260,17 @@ public Sort sort(K term, Sort expectedSort) {
...
-                        Set<Integer> otherPositions = new HashSet<>(positions);
-                        otherPositions.remove(0);
+            if (prod.params().nonEmpty()) {
+                for (Sort param : iterable(prod.params())) {
+                    if (prod.sort().equals(param)) {
+                        Set<Integer> positions = IntStream.range(0, prod.nonterminals().size())
+                                    .mapToObj(i -> Tuple2.apply(prod.nonterminals().apply(i), i))
+                                    .filter(t -> t._1().sort().equals(param))
+                                    .map(t -> t._2())
+                                    .collect(Collectors.toSet());
```
ehildenb: duplicate code?

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -175,21 +178,27 @@ private K visitChildren(K term, Sort actualSort) {
...
+            List<Sort> args = new ArrayList<>();
+            if (prod.params().nonEmpty()) {
+                for (Sort param : iterable(prod.params())) {
                     Sort expectedSort;
-                    if (positions.contains(0)) {
+                    Set<Integer> positions = IntStream.range(0, prod.nonterminals().size())
+                                .mapToObj(i -> Tuple2.apply(prod.nonterminals().apply(i), i))
+                                .filter(t -> t._1().sort().equals(param))
+                                .map(t -> t._2())
+                                .collect(Collectors.toSet());
```
ehildenb: Duplicate code? Let's factor out if so, for maintainability.

ehildenb: (and readability)

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -223,7 +223,8 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
   @transient lazy val signatureFor: Map[KLabel, Set[(Seq[Sort], Sort)]] =
     productionsFor mapValues {
       ps: Set[Production] =>
-        ps.map {
+        ps.filter { p: Production => p.params.isEmpty }
+        .map {
```
ehildenb, line 227: Parametric productions aren't considered part of the signature? Or will the resolved variants of them already also be present in the definition whenever this is called?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -247,13 +248,13 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
     Att(union.filter { key => attMap(key._1._1).size == 1 }.toMap)
   }
 
-  lazy val definedSorts: Set[Sort] = (productions map {_.sort}) ++ (sortDeclarations map {_.sort})
+  lazy val definedSorts: Set[Sort] = (productions filter {p => !p.params.contains(p.sort)} map {_.sort}) ++ (sortDeclarations map {_.sort})
```
ehildenb: We seem to use this pattern `p.params().contains(p.sort())`, or `p.params().contains(...)`, or `!p.params().contains(...)`. I wonder if it makes sense to factor out a `isSortVariable()` method?

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
dwightguth: Well, they're supported inasmuch as they were before, we're simply not actually inserting parameters into the klabels that the backend sees.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth: the latter.

ehildenb (approved):

PR #820, opened 2019-10-03
[BUGFIX]: Haskell backend rejects definitions which import K-TERM

### LINKED ISSUE #43: BUG: `domains.k` overlapping symbols
`domains.k` has three `symbol` with the same `klabel` for `#EmptyK`, and two for `#EmptyKList`

Currently the Haskell backend will reject this definition. Should it be that the frontend generates `definition.kore` differently, or should the Haskell backend accept this definition?

For now, I am manually editing `definition.kore` to only have one of the productions for each of these before running the rule merger.

This issue only shows up because we are hacking around the unparsing issue by including K-TERM in the definition module.

ehildenb:
Fixes runtimeverification/polkadot-verification#43

Currently the Haskell backend will reject this definition _at runtime_ (not kompile time), with the following error:

```
kore-exec: Error in module 'EMPTY-K' -> (/home/dev/src/k-dev/k-distribution/tests/regression-new/empty-k-haskell/././empty-k-kompiled/definition.kore 91:10, /home/dev/src/k-dev/k-distribution/tests/regression-new/e
mpty-k-haskell/././empty-k-kompiled/definition.kore 90:10): Duplicated name: 'Lbl'Hash'EmptyK'.
```

This comes from:

```
  syntax K ::= ".K"      [klabel(#EmptyK), symbol, unparseAvoid]
             | "."       [klabel(#EmptyK), symbol]
             | ".::K"    [klabel(#EmptyK), symbol, unparseAvoid]
```

which gives three productions in `definition.kore` with the same symbol. A similar issue happens for `#EmptyKList`.

Should this:

1.  Be allowed by the Haskell backend?
2.  Be rejected by the frontend?
3.  Something else?


ehildenb:
One fix (which I've pushed up) is to remove the `symbol` attributes from the duplicated klabels. Then at least the Haskell backend doesn't get the same symbol for the haskell backend, and the backend no longer complains.

ttuegel:
The Haskell backend correctly rejects this definition because it uses the same name for different declarations; Kore requires _all_ declarations to have unique names. I'm not sure why the frontend does that in this case, though.

ehildenb:
Well, for now I'm hacking together some bash to remove the extra lines.

dwightguth (changes requested):
This is not going to work. Moreover, as I have said multiple times now, users should not be importing modules in kast.k except BASIC-K. So we really should fix whatever problem made you think doing so was a good idea, because you are importing a module that should not be imported. Indeed, importing this module directly should probably be an error.

ehildenb:
I mean come on, we should have valid definitions in `kast.k`. And I'm importing it to hack around the unparsing issues which we've been discussing (about not having `#EmptyK` symbol and `#EmptyKList` symbol available in pretty unparser).

ehildenb:
Closing because we've found other workarounds for the moment.

PR #822, opened 2019-10-04
Set LLVM backend as default

ehildenb:
It's getting old to have to tell people to use the llvm backend instead of the ocaml backend, because the ocaml backend is set as default. This:

-   Sets the llvm backend as default.
-   Re-orders the options in `kompile --help` to emphasize llvm and haskell backends.
-   Makes sure `--non-strict` only has an effect on the ocaml backend.


nishantjr:
for `--nonstrict` on non-ocaml backends, can we have an error instead of silently ignoring it.

dwightguth:
If this passes tests, I'm fine with it, but we should also introduce a warning when compiling with the ocaml backend that it is deprecated and will be deleted in the next major release.

ehildenb:
@nishantjr done, now it's a criticalError to use --non-strict with any backend other than OCaml.

@dwightguth also done, now `Kompile` registers a criticalWarning when built with the OCaml backend.

ehildenb:
Looks like we're hanging now (though able to `kompile` the `kool` definition).

ehildenb:
Subsumed by #883

PR #824, opened 2019-10-04
Document several attributes asked about by users

ehildenb:
This provides documentation for:

-   `function` and `functional`
-   `freshGenerator`
-   `anywhere`
-   `strict` and `seqstrict`

### DIFF
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -121,8 +121,8 @@ syntax AExp ::= Int
               | AExp "+" AExp [strict]
 ```
 
-This generates two heating rules (where the hole syntax `[]` is automatically
-added to create an evaluation context):
+This generates two heating rules (where the hole syntaxes `"[]" "+" AExp` and
+`AExp "+" "[]"` is automatically added to create an evaluation context):
 
 ```k
 rule <k> AE1:AExp + AE2:AExp => AE1 ~> [] + AE2 ... </k> requires notBool isKResult(AE1)
@@ -180,6 +180,18 @@ rule <k> REST && true  => REST  ... </k>
 rule <k> _    && false => false ... </k>
 ```
 
+This will generate rules like this in the case of `_||_` (note that `BE1` will
+not be heated unless `isKResult(BE2)` is true, meaning that `BE2` must be
+evaluated first):
+
+```k
+rule <k> BE1:BExp || BE2:KResult => BE1 ~> [] || BE2 ... </k> requires notBool isKResult(BE1)
+rule <k> BE1:BExp || BE2:BExp    => BE2 ~> BE1 || [] ... </k> requires notBool isKResult(BE2)
+
+rule <k> BE1:KResult ~> [] || BE2 => BE1 || BE2 ... </k>
+rule <k> BE2:KResult ~> BE1 || [] => BE1 || BE2 ... </k>
+```
+
 ### `function` and `functional` attributes
 
 Many times it becomes easier to write a semantics if you have "helper" functions
@@ -210,8 +222,8 @@ rule I1 /Word I2 => (I1 /Int I2) modInt (2 ^Int 256) requires I2 =/=Int 0
 
 In K, you can access "fresh" values in a given domain using the syntax
 `!VARNAME:VarSort` (with the `!`-prefixed variable name). This is supported for
-builtin sort `Int` already. For example, you can generate fresh memory locations
-for declared identifiers as such:
+builtin sorts `Int` and `Id` already. For example, you can generate fresh memory
+locations for declared identifiers as such:
 
 ```k
 rule <k> new var x ; => . ... </k>
@@ -400,7 +412,9 @@ syntax Stmt ::= Stmt ";" Stmt
 rule (S1 ; S2) ; S3 => S1 ; (S2 ; S3) [anywhere]
 ```
 
-Then after every step, all occurances of `_;_` will be re-associated.
+Then after every step, all occurances of `_;_` will be re-associated. Note that
+this allows the symbol `_;_` to still be a constructor, even though it is
+simplified similarly to a `function`.
 
 ### `smt-lemma`, `lemma`, and `trusted` attributes
 
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -435,7 +435,7 @@ A `strict` attribute with no rule label associated with it is equivalent to
 a `strict` attribute given with the following context alias:
 
 ```k
-context alias [default]: <k> HERE:K ...</k>
+context alias [default]: <k> HERE:K ... </k>
 ```
 
 Configuration Declaration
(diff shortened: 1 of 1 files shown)


pending-documentation.md
```diff
@@ -107,6 +107,143 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+a *result* is for K, by extending the  `KResult` sort.
+
+For example:
+
+```k
+syntax AExp ::= Int
+              | AExp "+" AExp [strict]
+```
+
+This generates two heating rules (where the hole syntax `[]` is automatically
```
dwightguth: this makes it sound like we are creating a new production `syntax AExp ::= "[]"` but actually it's more like
```
syntax AExp ::= AExp "+" "[]"
syntax AExp ::= "[]" "+" AExp
```

pending-documentation.md
```diff
@@ -107,6 +107,143 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+
+```k
+rule <k> AE1:AExp + AE2:AExp => AE1 ~> [] + AE2 ... </k> requires notBool isKResult(AE1)
+rule <k> AE1:AExp + AE2:AExp => AE2 ~> AE1 + [] ... </k> requires notBool isKResult(AE2)
+```
+
+And two corresponding cooling rules:
+
+```k
+rule <k> AE1:KResult ~> [] + AE2 => AE1 + AE2 ... </k>
```
dwightguth: @ttuegel needs to speak about the haskell backend and the isKResult predicate, but I know the llvm backend does not generate isKResult for cooling rules.

pending-documentation.md
```diff
@@ -107,6 +107,143 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+              | BExp "&&" BExp [strict(1)]
+
+rule <k> true  || _    => true ... </k>
+rule <k> false || REST => REST ... </k>
+
+rule <k> true  && REST => REST  ... </k>
+rule <k> false && _    => false ... </k>
+```
+
+If you want to force a specific evaluation order of the arguments, you can use
```
dwightguth, line 251: you really need to show what is generated by the seqstrict attribute.

pending-documentation.md
```diff
@@ -107,6 +107,143 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+
+syntax BExp ::= Bool
+              | BExp "||" BExp [seqstrict(2,1)]
+              | BExp "&&" BExp [seqstrict(2,1)]
+
+rule <k> _    || true  => true ... </k>
+rule <k> REST || false => REST ... </k>
+
+rule <k> REST && true  => REST  ... </k>
+rule <k> _    && false => false ... </k>
```
dwightguth, line 266: missing from this discussion are contexts, context aliases, the `heat` and `cool` attributes, and the `hybrid` attribute.

pending-documentation.md
```diff
@@ -107,6 +107,143 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+
+rule I1 +Word I2 => (I1 +Int I2) modInt (2 ^Int 256)
+rule I1 /Word I2 => (I1 /Int I2) modInt (2 ^Int 256) requires I2 =/=Int 0
+```
+
+### `freshGenerator` attribute
+
+In K, you can access "fresh" values in a given domain using the syntax
+`!VARNAME:VarSort` (with the `!`-prefixed variable name). This is supported for
+builtin sort `Int` already. For example, you can generate fresh memory locations
```
dwightguth: also `Id`

pending-documentation.md
```diff
@@ -247,6 +384,24 @@ attribute, it would instead expand (as the user likely intended) to
...
+For example, if you want to make sure that some associative operator is always
+right-associated anywhere in the configuration, you can do:
+
+```k
+syntax Stmt ::= Stmt ";" Stmt
+
+rule (S1 ; S2) ; S3 => S1 ; (S2 ; S3) [anywhere]
+```
+
+Then after every step, all occurances of `_;_` will be re-associated.
```
dwightguth: I would draw the distinction that _:_ is still a constructor modulo this axiom, whereas a function is not a constructor.

dwightguth (changes requested):
it's a good start but it needs fleshing out. I'm happy to assist with writing text if you want to assign specific portions to me.

Mainly I would suggest defining strictness in terms of context aliases and strict attributes being transformed into contexts, and contexts in terms of heating and cooling rules.

pending-documentation.md
ehildenb: I won't add them.

pending-documentation.md

pending-documentation.md
ehildenb: Done.

pending-documentation.md

pending-documentation.md
ehildenb: Sentence added.

pending-documentation.md
ttuegel: The `definition.kore` generated for the Haskell backend contains function definitions for `isKResult`.

dwightguth (approved):

pending-documentation.md
```diff
@@ -107,6 +107,338 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+In this case, the code generated will be as follows:
+
+```k
+rule <k> foo(HOLE:Bar) => bar(HOLE) ~> foo([]) ... </k> [heat]
+rule <k> bar(HOLE:Bar) ~> foo([]) => foo(HOLE) ... </k> [cool]
+```
+
+This can be useful if the user wishes to evaluate a term using a different
+set of rules than normal. For example, if the user wants to evaluate a term
+to an rvalue instead of an lvalue.
```
ehildenb: `rvalue` and `lvalue` are completely out of nowhere here, probably better to just remove this sentence rather than send the user searching for "what is `rvalue`.

Also, if `rvalue` and `lvalue` are supposed to be actual sorts (which I assume, but I don't know) then capitalize them correctly and put them in code brackets. But if you keep them, explain them.

pending-documentation.md
```diff
@@ -107,6 +107,338 @@ argument and parameter are K, and it has no effect on the resulting sort of
...
+attribute. For example, `strict(c)` will instantiate these contexts once for
+each argument of the production, whereas `strict(c; 1)` will instantiate it
+only for the first argument. The special variable `HERE` is used to tell the
+compiler where you want to place the production that is to be heated or cooled.
+
+A `strict` attribute with no rule label associated with it is equivalent to
+a `strict` attribute given with the following context alias:
+
+```k
+context alias [default]: <k> HERE:K ...</k>
```
```suggestion
context alias [default]: <k> HERE:K ... </k>
```

pending-documentation.md
```diff
@@ -125,20 +195,37 @@ This generates two heating rules (where the hole syntaxes `"[]" "+" AExp` and
...
-Note that these rules depend on the definition of `KResult`, so you need to
-define this for K. For instance, we tell K that a term is fully evaluated once it
-becomes an `Int` here:
+rule <k> HOLE:AExp ~>  [] + AE2 => HOLE +  AE2 ... </k> [cool]
+rule <k> HOLE:AExp ~> AE1 +  [] =>  AE1 + HOLE ... </k> [cool]
+```
+
+You will note that these rules can apply one after another infinitely. In
+practice, the `KResult` sort is used to break this cycle by ensuring that only
+terms that are not part of the `KResult` sort will be heated. In the theory of
```
ttuegel: > In the theory of K, these rules represent equalities between structurally distinct states rather
than transitions in the transition system.

This is not presently true of the generated Kore.

dwightguth: Did not know that, thanks. I will update.

PR #833, opened 2019-10-09
Fix a pretty serious bug in sort inference + better debugging info

dwightguth:
This is the first of many smaller PRs that I am making because the work I've done on parametric productions turned out to be quite invasive and I'm trying to break it into smaller diffs.

Basically this fixes a bug where we were not correctly computing the full range of possible sorts for variables in nested ambiguities and thus triggering inference incorrectly to the wrong sort.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
@@ -202,6 +202,7 @@ public class VariableTypeInferenceFilter extends SetsGeneralTransformer<ParseFai
             VarKey fails = null;
             Collection<Sort> constraint = null;
             for (Multimap<VarKey, Sort> variant : vars2.vars) {
+                boolean success = true;
                 // take each solution and do GLB on every variable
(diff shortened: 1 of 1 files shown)


radumereuta (approved):

radumereuta:
Just to make sure. This bug appeared since you added parametric productions?

dwightguth:
This is more like two bugs and an improved error report, but yes, both of the two bugs were discovered due to the code changes I made for dealing with polymorphic productions. They could theoretically have surfaced at any time though with the right definition, because they're orthogonal to the changes I made, those changes just happened to uncover the right set of circumstances to trigger them.

radumereuta:
Can you explain the circumstances then?

dwightguth:
well, the else cases I added where we call TermCons.with were fixing a bug where if you had an inferred variable as an immediate child of a polymorphic production, it would not correctly add the correctly sorted cast to the parse tree, thus leading to incorrectly inferred sorts of variables at runtime.

The changes made to the `Set<Multimap<>>` code fixed a bug where we would incorrectly fail to split the state of inference into two cases when we reached an ambiguity, thus causing the sorts of variables on both sides of the ambiguity to be lumped together into a single possible solution to the inference problem.

PR #835, opened 2019-10-09
Make KORE slightly more deterministic

dwightguth:
This is one of many smaller PRs that I am making because the work I've done on parametric productions turned out to be quite invasive and I'm trying to break it into smaller diffs.

Here we delete the productionID attribute that was not being used for anything, and then define a sorted order of sorts when referred to in ModuleTOKORE.


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -976,7 +976,7 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
         if (isMacro) {
             att = att.add("macro");
         }
-        return att;
+        return att.remove(Constants.ORIGINAL_PRD, Production.class);
```
yzhang90, line 981: is this change related to Production_ID?

yzhang90 (commented):
I would change `definedSorts` on line 194 to `sortedSorts` as well, though it doesn't matter much.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: No, it's related to the upcoming changes with polymorphic productions. We are removing this attribute because it doesn't belong in the generated kore and it was messing with my ability to compare the kore generated before and after my changes. I should be able to delete this line now that I'm done testing if you'd like though?

yzhang90: can you add a todo here to make sure that we don't forget it?

yzhang90 (approved):
LGTM

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
yzhang90: It's ok to remove this attribute since it is never used in kore generation. I would add a comment here pointing out this attribute is useless in the kore generation if possible.

PR #842, opened 2019-10-14
Added attributes for distinguishing claims

traiansf:
This PR comes to address an open issue for the Haskell backend the ability to distinguish between all path and one path claims: https://github.com/kframework/kore/issues/906


pending-documentation.md
```diff
@@ -845,6 +845,22 @@ Note that in the case of Set and Map, one guarantee is that K1, K2, K3, and K4
...
+be used to clarify what kind of claim a rule is.
+
+In addition of being able to annotate a rule with one of them
+(if annotating with more at the same time, only one of them would be chosen),
+one can also annotate whole modules, to give a default claim type for all rules
+in that module.
+
+Additionally, the Haskell backend introduces an extra command line option
+for the K frontend, `--default-claim-type`, with possible values
+`all-path-claim` and `one-path-claim` to allow choosing a default type for all
```
ttuegel: Could we use `--default-claim-type=all-path` and `--default-claim-type=one-path` (without `-claim` in the flag name) for consistency with the module attributes?

traiansf: Thanks!

dwightguth:
I'd like to see a test for the case when the attribute is added to an individual sentence because it doesn't look like that case works to me.

traiansf:
@dwightguth please take another look.

dwightguth (approved):

dwightguth:
Looks good but I'd like Tom to approve too before we merge.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKRunOptions.java
```diff
@@ -16,6 +18,19 @@
     @Parameter(names="--haskell-backend-home", description="Directory where the Haskel backend source installation resides.")
     public String haskellBackendHome = System.getenv("KORE_HOME");
 
-    @Parameter(names="--all-path-reachability", description="Whether to interpret claims as All Path Reachability Claims.")
-    public boolean allPathReachability = false;
+    @Parameter(names="--default-claim-type", converter = SentenceTypeConverter.class, description="Default type for claims. Values: [all-path|one-path")
```
ttuegel:
```suggestion
    @Parameter(names="--default-claim-type", converter = SentenceTypeConverter.class, description="Default type for claims. Values: [all-path|one-path=]")
```

ttuegel (approved):
I think I found one trivial typo, but it looks good to me!

PR #844, opened 2019-10-16
make StringBuilder an explict argument to convert functions

yzhang90:


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -398,67 +399,68 @@ private void genUnitAxiom(Production prod) {
...
+        convert(prod.sort(), prod, sb);
         sb.append(") [unit{}()] // left unit\n");
     }
 
-    private void genFunctionalAxiom(Production prod) {
+    private void genFunctionalAxiom(Production prod, StringBuilder sb) {
         // exists y . f(...) = y
         Production finalProd = prod;
-        functionalPattern(prod, () -> applyPattern(finalProd, "K"));
+        functionalPattern(prod, () -> applyPattern(finalProd, "K", sb), sb);
```
yzhang90, line 433: am i doing it right here?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -265,43 +266,43 @@ private void translateSymbols(Map<String, Boolean> attributes, SetMultimap<KLabe
...
+            convert(finalProd.getSubsortSort(), finalProd, sb);
             sb.append(", ");
-            convert(finalProd.sort(), finalProd);
+            convert(finalProd.sort(), finalProd, sb);
             sb.append("} (From:");
-            convert(finalProd.getSubsortSort(), finalProd);
+            convert(finalProd.getSubsortSort(), finalProd, sb);
             sb.append(")");
-        });
+        }, sb);
```
yzhang90, line 297: am i doing it right here(functionalPattern function)?

dwightguth:
It looks like this PR is /only/ deleting the sb field from ModuleToKore and passing it as a parameter everywhere, not actually taking advantage of the new API to construct multiple StringBuilders, is that correct?

yzhang90:
Currently yes. In my later PR, I will call `convert(K k, StringBuilder sb)` on a different String builder to build some terms that will be used later in the code generation.

dwightguth (approved):

PR #846, opened 2019-10-18
add haskell backend for itp proof

yzhang90:


daejunpark:
Yi, what changes are made to the haskell specs? The only difference between java and haskell is the attribute of k(Int)? If so, have you tried to add the same attribute for the java backend? What happened?

daejunpark (approved):

daejunpark:
ok, the java backend works for the additional attributes as well, so i reorged to remove duplicates.

PR #848, opened 2019-10-21
cache topCellSort String and make the code more readable

yzhang90:


dwightguth (commented):
Looks reasonable since sorts are pretty much always pretty short strings. Technically this is quadratic in the length of the sort string though, so you definitely can't do this with anything that might be very long. I'd also like it if you could confirm that a large semantics (eg kevm) doesn't take appreciably longer to kompile with `--backend llvm --no-llvm-kompile` than before during an incremental build. I could be wrong about the performance impact being minimal, so let's measure.

dwightguth (approved):

PR #850, opened 2019-10-22
Use BYTES-HOOKED with Haskell backend

ttuegel:
The regression tests pass, but I have no indication how the symbolic module performs with large byte arrays.


ttuegel:
This is ready. I see up to 45% reduction in runtime on KEVM.

ttuegel:
@dwightguth I made the changes you requested to the `bytes-haskell` tests, to avoid duplicating the test inputs and outputs.

dwightguth (approved):

Issue #828, opened 2019-10-07
Make ambiguities errors


dwightguth:
Long ago it  was decided that ambiguities should be warnings. However, in general there is pretty much always a bug in your semantics if it has an ambiguity, because you cannot guarantee that the correct ambiguous parse will be chosen. Indeed, different versions of K may choose differently because it is currently choosing on the basis of hash set ordering. 

Given that ambiguous parses are always indicative of a bug, I propose that we make ambiguities errors instead of warnings. I think there may be some ambiguities in the test suite right now, but if I recall correctly they should all be indicative of a bug in the handling of module attributes somewhere. I think we can probably fix the test suite rather than worrying about this case as if it's valid.

PR #855, opened 2019-10-29
Make ambiguities report as errors - b

radumereuta:
Fix for #828


dwightguth (changes requested):
I'm not comfortable with tests being deleted without justification. Please explain why you have deleted some tests in this PR.

radumereuta:
Also sent to the #k channel:
I only removed test 2 and 3, there is another one which tests for `[Error] Compiler: Could not find main syntax module with name TEST-SYNTAX in definition.  Use --syntax-module to specify one. Using TEST as default.`

radumereuta:
@dwightguth please review.
I've added the test as you asked.
I also found a bug related to cached error messages.

dwightguth (approved):

PR #862, opened 2019-10-31
Auto update dependents of K frontend

ehildenb:
This auto-updates the following dependents of K frontend on pushes to K master:

-   KEVM
-   KWasm
-   Haskell Backend
-   Beacon Chain Project

Note that KEVM and Haskell backend updates will _only_ go through if the release goes through.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -122,7 +122,7 @@ pipeline {
             stage('Build and Package on Debian Buster') {
               when {
                 anyOf {
-                  not { changeRequest() }
+                  branch 'master'
                   changelog '.*^\\[build-system\\] .+$'
                   changeset 'Jenkinsfile'
                   changeset 'Dockerfile'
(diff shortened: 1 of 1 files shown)


dwightguth (commented):
Code looks fine, but I'd like a bit more understanding of how it works before I approve. Can you explain what exactly the various attributes to the build step are doing?

ehildenb:
The build being triggered is the `rv-devops/master` build, which just opens the PR on the target repository and quits.

`propagate` means that if the `rv-devops/master` build fails, this build should fail too. The `rv-devops/master` build can fail for several reasons, one of them being for instance of someone else has manually bumped the submodule in the meantime. So we don't want to fail this build in that case.

The `wait` parameter says should it wait for that build to finish before continuing this build, we we don't want to do because it would lengthen the time of this build, especially in the case that the CI servers are under heavy load and those submodule updates are just queued instead of being run immediately.

The section in the `parameters` block simply tells the `rv-devops/master` build which submodules to attempt updating.

Note that there isn't a reliable way right now to actually get back the status of the submodule update PR itself (since the `rv-devops/master` build just opens the PR, which triggers a build on the target repo in the standard github hooks way on Jenkins). Also note that the `update-deps` script now correctly handles the case that an `_update-deps` branch already exists, by choosing to update that branch instead of deleting it and opening a new one.

dwightguth (approved):

PR #864, opened 2019-10-31
kernel/BinaryLoader: replacing default JDK serializer with FST, for performance

denis-bogdanas:


kernel/pom.xml
```diff
@@ -191,6 +191,19 @@
       <classifier>windows</classifier>
       <type>exe</type>
     </dependency>
+
+    <!--Custom serializer-->
+    <!--<dependency>
```
dwightguth, line 196: Please remove this comment if you're not using this library.

dwightguth (changes requested):
This needs more detailed performance information before I can approve. I need to know how much time in kprove is spent deserializing before/after this change both with and without an already-warmed-up kserver.

denis-bogdanas:
I benchmarked on my local computer, with a biggish proof from Gnosis.
Saving cache.bin,~35 MB: JDK: 8s. FST: 0.7s
Loading cache.bin:            JDK: 21s FST: 0.8s

This is after running same kprove command 2 times, on cleanly built KEVM. If cache contains runs from multiple specs, diff gets bigger.

I assume results in your scenarios will be of similar magnitude.

Regarding comment, kryo is more popular and actively developed, so it was the first choice. I thought it's good to document why I didn't use it. Let me know if you want the comment removed.

dwightguth:
This still does not answer the question of the relative performance when a kserver is running. JDK serialization is vastly faster the second time a particular jvm process deserializes a given class, so while it's clear that this is a vast improvement in performance when run for the first time, I'd like to still know if it's a regression in performance the second time. If it's not a major regresion, it's still probably a change worth making, but I really want to see the numbers.

denis-bogdanas:
@dwightguth What you ask is quite time consuming. I'd guess you are rather curious to see what is the performance improvement in your scenario of interest. Can you give me instructions? What commands to run.

dwightguth:
It's really very simple. Just start the kserver and then run your kprove command twice, profiling it the second time. I don't see how it's that time consuming. In particular I'm trying to determine whether FST is /slower/ on the second deserialization than the JDK serialization, which is extremely rapid the second time.

denis-bogdanas:
Results on kserver, JVM serializer:
- Run 1: 26s read, 8s write
- Run 2: 27s read, 8s write
- Run 2 total parse time: 104s

FST serializer:
- Run 1: 0.6s read, 0.8s write
- Run 2: 0.4s read, 0.4s write
- Run 2 total parse time: 71s

dwightguth:
Okay so your data set is wildly inaccurate with respect to the performance of Java serialization, but I am persuaded that this is in fact not a regresssion on warmed-up jvms.

dwightguth (approved):

PR #865, opened 2019-10-31
Document `simplification` and `concrete` attributes

ehildenb:
This PR adds documentation for:

-   `concrete`, `#isConcrete`, and `#isVariable`
-   `simplification`

Questions I have are:

-   @ttuegel and @virgil-serbanuta is my documentation of `simplification` correct?
-   @daejunpark is my documentation of `concrete`, `#isConcrete`, and `#isVariable` correct? In particular, is `#isVariable` _only_ true for a `KVariable`, or is it also true for any term which has symbolic subterms?


pending-documentation.md
```diff
@@ -648,6 +648,68 @@ ambiguity during parsing, so care should be taken to ensure that named
...
+   andBool #isConcrete(Z)
+   andBool #isVariable(Y)
+```
+
+### `simplification` attribute
+
+The simplification attribute identifies axioms that are useful for simplifying
+configurations, without being part of the main semantics. When a function rule
+is tagged as `simplification`, the Haskell backend will only apply that rule if
+the side condition has no remainder (that is, it's always true/valid).
```
yzhang90: that is, it's always true/valid under the top-level predicate (or the top-level predicate implies the side condition)

pending-documentation.md
```diff
@@ -648,6 +648,72 @@ ambiguity during parsing, so care should be taken to ensure that named
 nonterminals are sufficiently unique from one another to prevent such
 ambiguities. Of course, the compiler will generate a warning in this case.
 
+### `concrete` attribute, `#isConcrete` and `#isVariable` function
+
+**NOTE**: `#isConcrete` and `#isVariable` are _not_ supported by the Haskell
```
ttuegel: We should emphasize they _will never be_ supported by the Haskell backend.

pending-documentation.md
```diff
@@ -648,6 +648,72 @@ ambiguity during parsing, so care should be taken to ensure that named
...
+   andBool #isConcrete(Z)
+   andBool #isVariable(Y)
+```
+
+### `simplification` attribute
+
+The simplification attribute identifies axioms that are useful for simplifying
+configurations, without being part of the main semantics. When a function rule
+is tagged as `simplification`, the Haskell backend will only apply that rule if
+the side condition has no remainder given the current top-level predicate (that
```
ttuegel: And if the rule _matches_ the configuration.

ehildenb: I was under the impression that the haskell backend _always_ uses matching for functions? So I didn't want to put it here, because it might confuse people who take this to mean that usually it's not using matching.

pending-documentation.md

pending-documentation.md
ttuegel: Simplification rules are not function rules: they are not restricted to function-like equalities (in fact, they may not have function symbols at their head at all) and they are not included when closing over the definition rules for `owise`. Unlike definition rules, which are applied as a group, simplification rules are attempted one-by-one.

Furthermore, users should not assume that function rules are applied by matching. The operational semantics of function evaluation is undefined, the backend may do whatever it likes. It happens to use matching at this time, but we reserve the right to change that without notice. (For example, we may decide to automatically simplify `\ceil(function)` expressions, which would probably require using unification.) On the other hand, using matching is part of the contract of the `simplification` attribute.

dwightguth (approved):
looks good but obviously others should still review and approve.

virgil-serbanuta (approved):

pending-documentation.md
ehildenb: I believe I've captured this.

ttuegel (approved):

daejunpark (approved):

PR #868, opened 2019-11-02
kernel/BinaryLoader: Workaround for a FST bug causing OverlappingFileLockException

denis-bogdanas:
Commit 1 is just refactoring.
Commit 2 is the actual fix.


dwightguth (approved):
I'm approving, but you should probably check if the output stream has the same problem.

denis-bogdanas:
I did. Changing only what's needed.

PR #869, opened 2019-11-04
Adding kast option keepAmb

radumereuta:
Since we made ambiguities report as errors, I'm adding this option to kast to visualize ambiguous programs better.

### DIFF
--- a/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
+++ b/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
@@ -128,7 +128,7 @@ public class KastFrontEnd extends FrontEnd {
             }
             Module parsingMod = maybeMod.get();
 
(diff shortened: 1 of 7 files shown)


radumereuta:
This is the second step in solving #828.

dwightguth:
What is this actually used for? If it's just for describing ambiguities to users, why isn't the error message enough data?

radumereuta:
Sometimes, seeing the entire AST can be more descriptive.
Also, Grigore had plans at one point to do disambiguation with K rules. Since we're eliminating the ambiguities, we have no way of doing that. We'd have to rely on external parsers.

radumereuta:
After discussing with Dwight, we decided that: either the error message contains sufficient information to resolve the ambiguity, in which case we shouldn't need this feature, or it doesn't, in which case we should fix the error message.

We'll wait and see if the users complain about ambiguities being errors and proceed then.

PR #870, opened 2019-11-04
k-distribution,kernel: remove notInRules

ehildenb:


dwightguth (commented):
looks good but let's remove autoReject while we're at it since that is also unused.

dwightguth (approved):

PR #873, opened 2019-11-05
Maven: Use --batch-mode for CI

ttuegel:
The `--batch-mode` option makes logs significantly more readable by preventing
Maven from printing progress updates that don't display correctly.


dwightguth (approved):
nice! I had no idea this was possible.

PR #877, opened 2019-11-06
kernel: refactoring KPrint and dependencies to Guice injections

denis-bogdanas:
@dwightguth This is first part of refactorings I need, up to KPrint. Next I want to make KProve and Kompile similar Guice injections. Please let me know if it's ok for you.

### DIFF
--- a/kernel/src/main/java/org/kframework/kast/KastModule.java
+++ b/kernel/src/main/java/org/kframework/kast/KastModule.java
@@ -8,6 +8,7 @@ import org.kframework.main.Tool;
 import org.kframework.utils.inject.DefinitionLoadingModule;
 import org.kframework.utils.inject.Main;
 import org.kframework.utils.inject.Options;
+import org.kframework.utils.inject.RequestScoped;
 import org.kframework.utils.options.DefinitionLoadingOptions;
 
 import com.google.inject.AbstractModule;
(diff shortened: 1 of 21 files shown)


kernel/src/main/java/org/kframework/kast/KastModule.java
```diff
@@ -32,11 +35,16 @@ public void configure() {
         experimentalOptionsBinder.addBinding().toInstance(KastOptions.Experimental.class);
     }
 
-    @Provides
+    @Provides @RequestScoped
```
dwightguth, line 38: why was this made request scoped?

dwightguth (commented):
so, the change to make KPrint be constructed by Guice looks fine, but you've added a ton of RequestScoped attributes and it's not at all clear to me why.

kernel/src/main/java/org/kframework/kast/KastModule.java
denis-bogdanas: The @RequestScoped for `GlobalOptions` and other options objects which are just getters have no effect. I just thought it's good to mention it explicitly, as a for of documentation. Or mabye I was chasing a problem and thought this might be the solution.

For FileUtil, KPrint and TTYInfo it's good to avoid unnecessary instantiation. In case of FileUtil @RequestScoped does not work in KEq, because we want a different FileUtil for each definition, so it is overridden to `NO _SCOPE`.

dwightguth (approved):

PR #880, opened 2019-11-06
Fix another issue with sort inference

dwightguth:
We had a bug where we were not inferring casts in the direct children of polymorphic productions if the children are polymorphic in one another but not in the return sort. This should fix that case.


ehildenb (approved):
Checked that this fixes beacon-chain project.

PR #887, opened 2019-11-07
kernel/KProve, ProofDefinitionBuilder: refactored to Guice injections

denis-bogdanas:
ProofDefinitionBuilder is former KProve.getProofDefinition() and dependent members.


dwightguth:
I haven't had time to review this in depth yet, but let's not just indiscriminately add RequestScoped everywhere. The benefits when used correctly are tiny, whereas it can cause some extremely subtle bugs that are difficult to notice when used incorrectly.

denis-bogdanas:
I'll keep this in mind. This PR, and the next 2, are ready for some time. Not recent work.

Particularly for ProofDefinitionBuilder `@RequestScoped` is important because it will deal with definition caching later.

denis-bogdanas:
Ok. I removed all introduced `@RequestScoped`.

dwightguth (approved):

PR #891, opened 2019-11-08
kernel/Kompile, KBMC, KEq: refactored to Guice injections

denis-bogdanas:
There are 2 `@RequestScoped` added: for KBMC and KEq. These objects have 1-to-1 relationship with their respective front-end, so no reason why they shouldn't be RequestScoped. The only additional thing FrontEnd is doing is to open the DefinitionScope.


dwightguth:
This mostly looks fine but I don't understand your comment about RequestScoped. When there is only a single place where a dependency is constructed is exactly the place where you /don't/ want to use RequestScoped, because you are not changing behavior at all, but you are replacing a simple constructor call with a complex operation calling out into the scope.

denis-bogdanas:
I didn't know this is your position. Removed all @RequestScoped.

dwightguth (approved):

PR #892, opened 2019-11-08
kernel: Combining compiled.bin and cache.bin into one file - definitionAndCache.bin

denis-bogdanas:
This is necessary for next PR, to reduce object duplication. There's a significant object sharing between old compiled.bin and old parse.bin. So what could be one single object becomes 2 separate objects if it is loaded through compiled.bin and parse.bin.

If we combine them in one file, their sum is smaller than the parts, and future increases to cache size are even smaller. As a result after a single run of kprove, definitionAndCache.bin is now smaller than the old cache.bin alone.

Also parse cache now has `@DefinitionScope` and is shared across multiple definitions in kserver. The map and members are properly synchronized.

Next PR brings Proof definition caching, so reducing object duplication is even more important.

Code changes: cache saving/loading logic is moved from DefinitionParsing to DefinitionStorage. Most tricky changes are in `KompileFrontEnd`.

Warning: builds that rely on `compiled.bin` for compilation timestamp will no longer work properly. But they should use `timestamp` file anyway.

I delayed marking this PR ready because there's an intermittent failure of K builds. It is fixed now: https://github.com/kframework/k/pull/1037


kernel/src/main/java/org/kframework/kompile/DefinitionStorage.java
```diff
@@ -0,0 +1,56 @@
...
+        this.files = files;
+    }
+
+    public File getCacheFile() {
+        return files.resolveKompiled("definitionAndCache.bin");
+    }
+
+    public void save(DefinitionAndCache definitionAndCache) {
+        if (!cacheParses(definitionAndCache.compiledDefinition.kompileOptions)) {
+            definitionAndCache.parseCaches.clear();
```
dwightguth, line 33: this seems suspect to me. you are clearing the cache after having run the tool, when profile rules is enabled. This means that the following run, which does not have profile rules, will not use the cache. This is probably not what you want.

Instead, I think what you should do is make it so that the loading of the cache doesn't happen if profile rules is enabled, so that it sees an empty cache even though the cache isn't really empty, but then it should save the cache as normal.

kernel/src/main/java/org/kframework/kompile/DefinitionStorage.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        loader.saveOrDie(getCacheFile(), definitionAndCache);
+    }
+
+    public DefinitionAndCache load() {
```
dwightguth, line 38: it seems to me like this method should be private, and you should have public getters that return either the definition or the cache, which in turn call the private load method. Right now you have both a call to loadOrDie and loadCache, and that seems like badly factored code to me.

dwightguth (changes requested):
This code now only saves the caches during kompile if compilation succeeds, which is unacceptable. we need to save the caches immediately after parsing finishes so they can be reused even when there are parse errors.

kernel/src/main/java/org/kframework/kompile/DefinitionStorage.java
denis-bogdanas: This preserves exactly the old behavior of `KompileOptions.profileRules`. If it is enabled, parse cache is neither loaded nor saved. Besides, what you propose wouldn't change the behavior until the next `kompile`, in cases when you do `kompile --profile-rules` followed by a `kompile` without the option. I was careful here. I believe current behavior is more intuitive. You don't save what is not supposed to be used.

kernel/src/main/java/org/kframework/kompile/DefinitionStorage.java
denis-bogdanas: `DefinitionAndCache` guice object is required for `DefinitionStorage.save()`, called from 2 places. All problems you mention are just careful preservation of existing behavior. From diff alone this is not intuitive. You need `loadCache` in kompile, because you only need the parse cache component there. But you need `loadOrDie` in `DefinitionStorage.load()`, because you need both cache and definition and failure is not an option.

denis-bogdanas: Will work on point 3.

denis-bogdanas:
@dwightguth done

dwightguth (changes requested):
I don't see why we chose to modify the behavior of saving the cache immediately after parsing finishes.

denis-bogdanas:
Good question. There's another PR right on top of this one, which saves into cache runtime extended definitions: https://github.com/kframework/k/pull/910

So it puts the base definition, the parse cache and all the runtime extended definitions into one file. Then it makes sense to save the cache in the service that initiates both parsing and compuation of extended definition. That is `ProofDefinitionBuilder`. This behavior is equivalent to old one, including when exceptions are thrown. We just call `save` from another function.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -127,29 +120,8 @@ public DefinitionParsing(
         }
 
         def = resolveNonConfigBubbles(def, gen);
-        saveCachesAndReportParsingErrors();
```
denis-bogdanas, line 130: Place where caches were saved before - at the very end of `parseModules` invoked during runtime parsing. Similar for compile-time parsing.

dwightguth (commented):
I am suspicious of the fact that we have introduced a synchronized map to the parse cache where previously there was no synchronization. My experience is that if you're not careful, using synchronized caching instead of lockless caching significantly decreases performance. Can you explain the logic?

denis-bogdanas:
I don't know exactly what is multithreaded and what is not in parsing logic. It's hard to tell. The inner map inside `ParseCache` is synchronized from before this PR. So I thought the outer map must be synchronzied too.

https://github.com/kframework/k/blob/dep-inj-cacheFile/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java#L341-L348

kernel/src/main/java/org/kframework/kompile/DefinitionAndCache.java
```diff
@@ -0,0 +1,26 @@
...
+ * Created on 05-Nov-19.
+ */
+public class DefinitionAndCache implements Serializable {
+    public final @Nullable CompiledDefinition compiledDefinition;
+    public final Map<String, ParseCache> parseCaches;
+
+    public DefinitionAndCache(CompiledDefinition compiledDefinition,
+                              Map<String, ParseCache> parseCaches) {
+        this.compiledDefinition = compiledDefinition;
+        this.parseCaches = Collections.synchronizedMap(parseCaches);
```
denis-bogdanas: I added synchronization just for the outer map, here.

dwightguth: Let's make it a ConcurrentHashMap instead of a synchronizedMap. I think you're right that with the increased parallelism I added to the codebase recently, in theory this can be accessed from multiple threads, but we definitely need lockless reading from the map otherwise there will be a huge amount of contention between threads. It should be fine performance-wise if we make this change, because writing to the outer map happens at most once per module.

dwightguth (changes requested):
I realized another problem when I was rereading this PR. it seems that this code is no longer respecting the --cache-file kompile option... This option is being used by some of our customers who develop K semantics that are built with NIX.

denis-bogdanas:
This option cannot be implemented the same way in this PR. previously option was stored in compiled.bin and was used to load cache.bin. Now both are in the same file, so I cannot retrieve the option before loading definitionAndCache.bin.

@ehildenb said at some point that the option was not really helpful to our clients because they expected a krun option. Everett, can you confirm?

If option is still needed I would have to store it in a separate file.

ehildenb:
I'm actually not sure. I know that both @ttuegel and @asymmetric have attempted to package up K, and I think at least @asymmetric has tried to package up KEVM on NixOS. IIRC, the K packages work fine, but @asymmetric eventually they gave up having KEVM depend on the K Nix package, instead building K directly into the KEVM package independent of the K package, because of this cache-file option issue.

@asymmetric can you confirm whether you guys use `--cache-file` or not?

denis-bogdanas:
@ehildenb Looks like the better solution is an option that disables updates to parse cache. Cache produced by `kompile` would still be read, but krun/kprove-time updates would not be saved. This would make `kprove` slower when run repeatedly over same definition. But should not affect performance of other tools. Would that be acceptable?

dwightguth:
Any solution where krun and kprove do not write to the kompiled directory should be fine.

asymmetric:
@ehildenb this was over a year ago so can't remember the specifics, but yes, your description sounds correct: KEVM's build process requires writing to the location where K is (IIRC to write the chache file mentioned here), but in Nix dependencies are read-only, so I personally couldn't think of a way to nixify KVEM at that time.

So having a way to specify the location of the cache file would've been a solution.

> @asymmetric can you confirm whether you guys use --cache-file or not?

We don't use it because we don't package K or KEVM with nix at the moment, but we would like to in the future.

denis-bogdanas:
@dwightguth Ok I deleted `--parse-cache` since it's not needed. Will implement `--readonly-cache` in another PR. Please review.

ehildenb:
Re-open if needed and updated. But it seems that this centralizes things more than making them incremental, which is moving the wrong direction.

PR #907, opened 2019-11-15
Jenkinsfile, src/main/scripts/brew*: update ROOT_URL

ehildenb:


ehildenb:
These are all the uses of `ROOT_URL` that I could find with `git grep ROOT_URL`

dwightguth (approved):

PR #910, opened 2019-11-17
kernel/ProofDefinitionBuilder: storing compiled extended definitions in cache file

denis-bogdanas:


dwightguth:
Specifically what I'm trying to avoid is adding a bunch of @RequestScoped annotations when that will not necessarily make the creation of the injector faster. If you can demonstrate in each case that it does, fine, but otherwise I'd rather we leave the scoping of stuff alone unless it affects correctness.

ehildenb:
Re-open if needed.

PR #917, opened 2019-11-18
kernel/binaryLoader: corrected file locking mechanism issues caused by JDK limitations

denis-bogdanas:
Explanation in code comments.


kernel/src/main/java/org/kframework/utils/BinaryLoader.java
```diff
@@ -96,12 +96,12 @@ public void saveOrDie(File file, Object o) {
...
     public <T> T loadCache(Class<T> cls, File file) {
-        try (FileInputStream in = new FileInputStream(file)) {
-            return cls.cast(loadSynchronized(in));
+        try {
+            return cls.cast(loadSynchronized(file));
         } catch (FileNotFoundException e) {
             //ignored
         } catch (IOException | ClassNotFoundException e) {
-            kem.registerInternalHiddenWarning("Invalidating serialized cache due to corruption.", e);
+            kem.registerInternalWarning("Invalidating serialized cache due to corruption.", e);
```
dwightguth: this should not be a regular warning as it can happen during normal operation.

dwightguth:
This doesn't make sense to me. You said you did this thing with opening the file twice for the case when another thread reads the file between opening it and locking it, but the entire method is already guarded by a write lock, so that shouldn't be possible. Why not move the FileStream creation into the method, but otherwise leave the method alone? wouldn't that be a simpler but also correct fix?

kernel/src/main/java/org/kframework/utils/BinaryLoader.java
denis-bogdanas: It did happen during the normal operation so far, precisely because of the bug this PR fixes. It took me a long time to figure out the root of the problem because this warning was hidden. From now on this shouldn't happen as part of normal operation unless definition was not kompiled with latest K. If this warning happens in future, we want to see it.

Do you know other scenarios why it might happen, unrelated to this fix or stale definition?

denis-bogdanas:
> 
> 
> This doesn't make sense to me. You said you did this thing with opening the file twice for the case when another thread reads the file between opening it and locking it, but the entire method is already guarded by a write lock, so that shouldn't be possible. Why not move the FileStream creation into the method, but otherwise leave the method alone? wouldn't that be a simpler but also correct fix?

The write lock acts in kserver mode only. The file lock acts in standalone multi-process mode only. Hence stream workaround is needed for the 2nd case.

kernel/src/main/java/org/kframework/utils/BinaryLoader.java
dwightguth: definition was compiled with an old version of K /is/ normal operation.

kernel/src/main/java/org/kframework/utils/BinaryLoader.java
```diff
@@ -112,27 +112,30 @@ public void saveOrDie(File file, Object o) {
...
      */
-    public void saveSynchronized(FileOutputStream out, Object o) throws IOException, InterruptedException {
+    public void saveSynchronized(File file, Object o) throws IOException, InterruptedException {
         //To protect from concurrent access from another thread
         lock.writeLock().lockInterruptibly();
-        try {
+        //JDK API limitation: there's no API to atomically open a file for writing and lock it.
+        //Consequently, if another thread reads a file between the moments this thread opens a stream and acquires a
+        // lock, it will see an empty file. We use lockStream open for appending as workaround.
+        try (FileOutputStream lockStream = new FileOutputStream(file, true)) {
```
dwightguth, line 122: this would be less confusing if we just directly got a file channel via FileChannel.open(file.toPath()) rather than creating a file in append mode and then not appending anything.

kernel/src/main/java/org/kframework/utils/BinaryLoader.java
denis-bogdanas: Fine, I reverted to avoid arguing. But this practice of hidden warnings needs rethinking. It is precisely because this warning was hidden that we didn't discover this bug until now.

denis-bogdanas:
@dwightguth Turns out your solution although looks nicer, doesn't work for unknown reasons:
```
     [exec] [Error] Critical: Could not write to
     [exec] /home/jenkins-slave/workspace/k_PR-917@2/k-distribution/tests/regression-new/owise-haskell/./demo-kompiled/cache.bin
```
So I reverted to my code.

dwightguth:
Yeah, if it doesn't work, this is fine, I suppose.

dwightguth (approved):

PR #919, opened 2019-11-19
Add and improve #fun examples

xc93:
I improved two existing examples of using `#fun` and added a new one, borrowed from the existing language semantics. I also provided their formal translations to Kore patterns, which help the readers to understand the behaviors of `#fun` from a logical point of view.


xc93:
@traiansf Please review the above documentation of the discussion about `#fun` in https://runtimeverification.slack.com/archives/CBNGEMMUP/p1574184470310200

xc93:
This PR is closed because the added explanation doesn't belong to the Reference Manual, but Semantics of K document.

traiansf (commented):
I looks good to me, but I cannot but wonder why we don't encode it in Kore as we do in the frontend, i.e., adding a fresh symbol and extracting its definition as a rule for it instead of adding aliases and worrying about duplication.

dwightguth:
I tend to agree with Everett that this is not the file to describe proposals like this with regards to the K implementation. This is a document intended to describe the exact current behavior of the implementation oriented towards end-users.

ehildenb:
Either way, the documentation of `#fun` is currently lacking. So I don't want to close this, just want to get it into merge-able state.

ehildenb (approved):

xc93:
I'll come back to this PR over the weekend and make it in a way that is facing the end-users. Specifically, I'll restore the original description of using `lambda` to define `#fun`, but I'll include the new example which uses nested `#fun`. 

I may need further clarification about the exact behavior of `#fun` in the future.

ehildenb (approved):

dwightguth (approved):

PR #922, opened 2019-11-20
kernel: option --timeout: support for kserver mode

denis-bogdanas:


dwightguth (changes requested):
I'm not really sure why we're not just calling System.exit regardless of whether or not we are nailgun. Interrupting the main thread does not sound to me like it would be sufficient to get K to fully terminate if there are other threads associated with that nail.

denis-bogdanas:
Because timeout is for current task, not for entire kserver. There are usage scenarios in which some `kprove` commands are expected to timeout. We want the other commands to continue running.

dwightguth:
It will. We currently call System.exit after every frontend finishes, even when we are running a nail.

denis-bogdanas:
But in this case `System.exit()` is getting called from another thread than the nail - from the timeout thread. Will it know what nail to stop? Are the timeout thread and nail main thread linked by a common thread group?

dwightguth:
I think you're right that System.exit will not know what to do about this case, so thanks for pointing that out. I thought it would be handled on the basis of the parent thread, but it seems it's not. That being said, I still suspect that interrupting the main thread will not necessarily stop any threads that were spawned by that thread... This needs more investigation.

dwightguth:
Has this been tested? I took a look at the nailgun code and I wouldn't have expected this particular change to work, after having looked closer (contrary to what I thought earlier). I'd like to make sure that we've actually tested it before merging it.

denis-bogdanas:
There might be some problems with this PR so it's not ready yet. And I will test it in Nailgun to make sure it works.

ehildenb:
Re-open if needed.

PR #935, opened 2019-11-26
add --concrete-rules option

daejunpark:


java-backend/src/main/java/org/kframework/backend/java/kil/Rule.java
```diff
@@ -220,6 +220,11 @@ public boolean isConcrete() {
         return att().contains(Attribute.CONCRETE_FUNCTION_KEY);
     }
 
+    public boolean isConcrete(ImmutableSet<String> concreteRules) {
```
denis-bogdanas: This is worth renaming to `isConcreteRule`,and keep it with no arguments. And also include the old `isConcrete`. To distinguish it from `isConcrete` inherited from `JavaSymbolicObject` which is called in different context and should be kept separate.

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -343,15 +344,18 @@ public BitSet getChildrenDontCareRuleMaskForPosition(int i) {
         private final KExceptionManager kem;
         private final Provider<BuiltinFunction> builtins;
         private final GlobalOptions options;
+        private final ImmutableSet<String> concreteRules;
```
denis-bogdanas: rename to extraConcreteRuleLabels and `extraConcreteRuleLabels()` everywhere.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,6 +56,13 @@ public synchronized File specFile(FileUtil files) {
     @Parameter(names="--depth", description="The maximum number of computational steps to prove")
     public Integer depth;
 
+    @Parameter(names="--concrete-rules", description="List of concrete rule labels")
```
denis-bogdanas: Description should be expanded. I guess you mean:
```
List of rule labels to be considered concrete, in addition to rules marked with `[concrete]` attribute.
```

denis-bogdanas: Another concern: Option should be moved to `JavaExecutionOptions`. Then it will be available to other tools like KEq, and only in Java backend.

denis-bogdanas (changes requested):

denis-bogdanas (approved):

PR #948, opened 2019-12-04
Fix for sort annotations on parametric KApply

### LINKED ISSUE #945: #parseKORE doesn't take into consideration sort annotation
Version K-master, backend: LLVM
```
module TEST
  imports K-REFLECTION
  imports INT
  configuration <k> #parseKORE("\\dv{Int{}}(\"1\")"):Int </k>
endmodule
```
I would expect the output to be an Int, instead I get:
```
radu@ubuntu:~/work/test$ krun --output kast 
`<generatedTop>`(`<k>`(#token("1","K")),`<generatedCounter>`(#token("0","Int")))
```
parsed.txt has:
```
rule initKCell(.KList)=>`<k>`(#noDots(.KList),#SemanticCastToInt(`#parseKORE(_)_K-REFLECTION_Sort_String`(#token("\"\\\\dv{Int{}}(\\\"1\\\")\"","String"))),#noDots(.KList)) requires #token("true","Bool") ensures #token("true","Bool")
```
compiled.txt has this:
```
 rule `<generatedTop>`(`<k>`(`` `start_TEST_KItem`(.KList)=>`#parseKORE(_)_K-REFLECTION_Sort_String`(#token("\"\\\\dv{Int{}}(\\\"1\\\")\"","String"))``~>DotVar1),DotVar0) requires #token("true","Bool") ensures #token("true","Bool")
```
Somewhere along the way, the sort annotation is getting lost.
P.S. This is blocking me from progressing with the K-in-K definition.

dwightguth:
Fixes #945


radumereuta:
I know it's just a draft, but just wanted to see how much it works, and there are some issues with `#as`:
```
  configuration <k> .K </k>
  rule <k> .K #as Rest => 1 </k>
```

dwightguth:
Yeah I'm aware. There's a bug in the propagation of attributes I need to track down before this is going to work.

kernel/src/main/java/org/kframework/compile/ResolveSemanticCasts.java
```diff
@@ -116,24 +124,67 @@ public void apply(KApply v) {
...
+            return KApply(kapp.klabel(), kapp.klist(), f.apply(kapp.att()));
+        }
+        @Override
+        public K apply(KRewrite rew) {
+            return KRewrite(rew.left(), rew.right(), f.apply(rew.att()));
+        }
+        @Override
+        public K apply(KToken tok) {
+            return KToken(tok.s(), tok.sort(), f.apply(tok.att()));
+        }
```
ehildenb: This code is probably more broadly useful, consider moving it to where TransformK is defined?

ehildenb (approved):

PR #950, opened 2019-12-04
make RelWithDebInfo default build of llvm backend

dwightguth:


radumereuta:
StackOverflow:
> The default optimization level changes. With gcc and clang, at least, RelWithDebInfo uses -O2, but Release uses -O3

We might still need Debug if too many variables are `<optimized out>`. But that can be set manually.

dwightguth:
I'm going to do this a different way.

nishantjr:
Would just adding `-g` to the default flag for `Release` be reasonable? If someone really cares about the extra size they can strip the binary.

PR #951, opened 2019-12-04
kast.k: module KVARIABLE-SYNTAX

### LINKED ISSUE #934: Generated Kore contains illegal sort injections of `#KVariable`
The Kore generated by the frontend contains illegal sort injections such as (from `evm-semantics`):

```
// rule `is#KVariable`(inj{#KVariable,KItem}(#KVariable))=>#token("true","Bool") requires #token("true","Bool") ensures #token("true","Bool") []
  axiom{R} \implies{R} (
    \top{R}(),
    \and{R} (
      \equals{SortBool{},R} (
        Lblis'Hash'KVariable{}(kseq{}(inj{Sort'Hash'KVariable{}, SortKItem{}}(Var'Hash'KVariable:Sort'Hash'KVariable{}),dotk{}())),
        \dv{SortBool{}}("true")),
      \top{R}()))
  []
```

This sort injection is illegal because there is no subsort relation defined on `Sort'Hash'KVariable`.

ttuegel:
ML-SYNTAX does not need the `#KVariable` sort. That sort causes problems in Kore
generation because it does not properly become a subsort of `KItem`.

Fixes: #934

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -1342,6 +1342,7 @@ endmodule
 
 module STRATEGY
     imports ML-SYNTAX
+    imports KVARIABLE-SYNTAX
     imports K-EQUAL
 
     syntax #RuleTag ::= #KVariable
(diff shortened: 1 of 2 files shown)


dwightguth (approved):
This seems to not solve the issue if the definition uses strategies, but it's strictly better than what was there before. Really we only want #KVariable to be there when parsing rules, not when generating kore... We'll have to figure out a better solution eventually.

Issue #947, opened 2019-12-04
Frontend reports FileNotFoundException if error during backend


ttuegel:
If the backend returns an error, the frontend reports a `FileNotFoundException` because the output file was not created. In this case, the frontend should not expect to find a file, and should not report this error. The `FileNotFoundException` leads to confusing error reports from users.

nishantjr:
One case where this occurs is when the LLVM backend throws an exception. The correct behavior here should be for the LLVM backend to write the result.kore file  with the last good configuration, and exit failure. This will make debugging exceptions possible without resorting to the debugger. 

To support this, the frontend should print the result file if it exists even on failure.

ttuegel:
I like the idea of all Kore-based backends putting the last state in `result.kore` even if there is an error. The Haskell backend already prints the configuration, but not in the response file.

traiansf:
This was fixed by #954 and by https://github.com/kframework/kore/commit/c32cb34e5da799e8b3218d7037bee47b1783b73e

PR #954, opened 2019-12-06
Throwing a specific error if backend returned non-zero code without output

traiansf:


traiansf:
Fixes #947  (for the Haskell backend).

ttuegel (approved):

PR #967, opened 2019-12-11
Fix NoSuchElementException for klabel

virgil-serbanuta:


dwightguth:
Can we have a test?

virgil-serbanuta:
I added a commit with the test.. It fails on master, passes with this PR.

traiansf (approved):

dwightguth (approved):

PR #968, opened 2019-12-11
Remove "String" as builtin type in Z3

hjorthjort:
Emitting the function `(declare-fun string2wasmstring (String) WasmStringToken)`  causes an error saying there is no `String` sort. Declaring it explicitly fixes the problem.

Perhaps we should also delete the *string theory* set of builtin functions?

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
@@ -48,7 +48,6 @@ public class    KILtoSMTLib extends CopyOnWriteTransformer {
             Sort.INT,
             Sort.BIT_VECTOR,
             Sort.of(Sorts.Float()),
-            Sort.of(Sorts.String()),
             Sort.of(KORE.Sort("IntSet")),
             Sort.of(KORE.Sort("MIntSet")),
             Sort.of(KORE.Sort("FloatSet")),


ehildenb:
@hjorthjort let's try it with the *string theory* stuff removed, but you'll also have to maybe change `domains.k` for that (specifically, where those symbols are `smt-hook` they should become `smtlib`). Would be good to have @daejunpark 's feedback on this, I wonder why it is that these were added before but now it seems that our encoding to Z3 doesn't support it?

daejunpark:
@hjorthjort I cannot understand. String is a builtin sort. https://rise4fun.com/Z3/tutorial/sequences What's your z3 version? I'm using 4.6.0.

hjorthjort:
I need to declare String with version 4.4.1 (on my dev server), but can't (because it already exists) on version 4.6 on my own machine. Investigating further

ehildenb:
We officially are recommending Z3 4.6 or greater anyway @hjorthjort for many of our projects. So if switching to Z3 4.6 or greater solves it, let's update the KWasm documentation and close this.

hjorthjort:
Seems 4.4.1 is the newest one available for Ubuntu using `apt-get`. Building from source to see if that helps.

ehildenb:
Yeah, Ubuntu is a bit behind. But the Dockerfile's all have the instructions needed to build any of the versions from source.

daejunpark:
You can also just download the binary:
https://github.com/Z3Prover/z3/releases/tag/z3-4.6.0

and use without installing it only by setting some env variables, like:
```
Z3_DIR=/path/to/z3-4.6.0-x64-ubuntu-16.04
export PATH=$Z3_DIR/bin:$PATH
export LD_LIBRARY_PATH=$Z3_DIR/bin:$LD_LIBRARY_PATH
```

hjorthjort:
Latest version z3 worked, so this is indeed an issue with older versions. Thanks for the help!

PR #971, opened 2019-12-12
pyk code for working with K coverage logs

ehildenb:


k-distribution/src/main/scripts/lib/pyk/coverage.py
```diff
@@ -0,0 +1,129 @@
...
+                ruleAtt  = sentence['att']['att']
+                if    (isKApply(ruleBody)                                 and ruleBody['label']        == '<generatedTop>') \
+                   or (isKRewrite(ruleBody) and isKApply(ruleBody['lhs']) and ruleBody['lhs']['label'] == '<generatedTop>'):
+                    if 'UNIQUE_ID' in ruleAtt:
+                        dst_non_function_rules.append(ruleAtt['UNIQUE_ID'])
+
+    # Convert the src_coverage rules to dst_no_coverage rules via the maps generated above
+    dst_rule_list = []
+    for src_rule in src_rule_list:
+        if src_rule in src_rule_map:
```
nishantjr: inverting these `if`s would make this more readable

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -268,6 +298,17 @@ def minimizeRule(rule):
 
     return KRule(ruleBody, requires = ruleRequires, ensures = ruleEnsures, att = ruleAtts)
 
+def removeSourceMap(k):
+    def _removeSourceMap(att):
+        if isKAtt(att):
```
nishantjr, line 308: Same here

k-distribution/src/main/scripts/lib/pyk/coverage.py
ehildenb: Done.

k-distribution/src/main/scripts/lib/pyk/kastManip.py
ehildenb: I don't agree with this one, since there are not nested cases, and there is no failure case, it doesn't make sense to me.

nishantjr (approved):

PR #973, opened 2019-12-12
Check uniqueness of symbols in kore backends

dwightguth:


ehildenb:
@dwightguth will this handle overloads correctly?

eg. https://github.com/kframework/wasm-semantics/blob/master/data.md#basic-values

If so, add a test that it lets overloads through please.

dwightguth:
I'll add the test tomorrow, but yes, it allows overloads because overloads in kore are represented by distinct symbols modulo some equality axioms.

ehildenb (approved):

PR #990, opened 2019-12-20
apply function rules in order

daejunpark:
Sort function rules by their locations before applying. This helps avoid using #notKLabel in lemmas.

Performance overhead is not much.  Or, we can consider to cache the sorted list, if needed.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
@@ -34,6 +34,7 @@ import org.kframework.utils.errorsystem.KExceptionManager;
 
 import java.io.IOException;
 import java.io.ObjectOutputStream;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


denis-bogdanas (commented):
I don't think this is a correct change. You are introducing a hidden property of rule application, and abuse it in order to get rid of a visible but unwanted `#notKLabel`. The point to get rid of `#notKLabel` is to make specs compatible with Haskell. This change doesn't seem to help in this regard.

Can we use rule priority instead?

daejunpark:
My comment was misleading. It’s not mainly for notKLabel. I think we need
this feature, and found that it also helps for notklabel. This will make
our life a lot easier when writing lemmas.

Now a valid question is how hard it is for the haskell backend to implement
this. I guess that it is not that, from the fact that it supports the rule
priority.

denis-bogdanas:
Comments regarding the code itself, if you decide to keep this change. I can do them if you want.

1. Sorting must be done in Definition. I expect performance drop otherwise. I guess here, after kore rules were added: https://github.com/kframework/k/blob/dadcda08c422b015b544478369f8da0235e3f492/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java#L217
2. Use an inner class Rule.LocationComparator rather than implementing Comparable, because this is not a natural comparator for rules based on their meaning.
3. Comparator can be implemented purely functionally, like here: https://www.programcreek.com/java-api-examples/?class=java.util.Comparator&method=thenComparing

ehildenb:
Re-open if needed.

PR #992, opened 2019-12-27
Java Backend: improved simplification of equalities

denis-bogdanas:
Allows ==K term evaluation through regular rules before builtin simplification is applied.


daejunpark:
Have you tested this on VSC?  It seems nontrivial changes.

denis-bogdanas:
Yes of course. This PR is only for K test: https://github.com/runtimeverification/verified-smart-contracts/pull/300

Solar also uses it: https://github.com/runtimeverification/verified-smart-contracts/pull/301

daejunpark (approved):

PR #1001, opened 2020-01-03
domains.k: MAP imports BOOL and INT

ttuegel:
If the BOOL and INT modules are not imported into MAP, then the declaration of
Bool- and Int-valued functions in that module defines new Int and Bool sorts
that are not hooked to the builtin domains in Kore.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -90,6 +90,8 @@ module ARRAY
 endmodule
 
 module MAP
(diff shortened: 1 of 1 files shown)


dwightguth (changes requested):
let's make this import BOOL-SYNTAX and INT-SYNTAX instead.

ttuegel:
Done!

dwightguth (approved):

PR #1003, opened 2020-01-06
kast --gen-parser prototype

dwightguth:
This is definitely not ready for production use yet, but it seems to work for IMP at least. I intend to make a followup PR after trying out JSON and WASM.


kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -124,13 +135,59 @@ public int run() {
             }
             Module parsingMod = maybeMod.get();
 
-            K parsed = kread.prettyRead(parsingMod, sort, def, source, FileUtil.read(stringToParse));
+            if (options.genParser) {
+                try (ParseInModule parseInModule = RuleGrammarGenerator.getCombinedGrammar(parsingMod, true)) {
+                    try (Scanner scanner = parseInModule.getScanner()) {
+                        FileUtil files = this.files.get();
+                        File scannerFile = files.resolveTemp("scanner.l");
```
radumereuta: Why is this in the temp directory?
Shouldn't this be inside the `-kompiled` directory?
Could you put all the related files inside a single directory? `-kompiled` already has many things in it, and it would make it easier to look for debugging info when files are a bit more organized.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -392,9 +392,9 @@ public static ParseInModule getCombinedGrammar(Module mod, boolean strict, boole
                         Seq(NonTerminal(ul.childSort), Terminal(""), NonTerminal(Sort(ul.sort.name() + "#Terminator", ul.sort.params()))),
                         newAtts.add(Constants.ORIGINAL_PRD, Production.class, ul.pList));
                 // Es ::= Ne#Es
-                prod4 = Production(Seq(), ul.sort, Seq(NonTerminal(Sort("Ne#" + ul.sort.name(), ul.sort.params()))));
+                prod4 = Production(Seq(), ul.sort, Seq(NonTerminal(Sort("Ne#" + ul.sort.name(), ul.sort.params()))), Att().add("notInjection"));
```
radumereuta: Maybe make `notInjection` a static String.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/KSyntax2Bison.java
```diff
@@ -0,0 +1,297 @@
...
+          sentences.add(Production(prod.klabel(), prod.params(), newNT, Seq(NonTerminal(prod.getSubsortSort(), prod.nonterminals().apply(0).name())), prod.att()));
+        } else if (prod.klabel().isEmpty() || !item._2().contains(new Tag(prod.klabel().get().name()))) {
+          sentences.add(Production(prod.klabel(), prod.params(), newNT, prod.items(), prod.att()));
+        }
+      }
+    }
+    return Module(module.name(), module.imports(), immutable(sentences), module.att());
+  }
+
+  public static void getParser(Module module, Scanner scanner, Sort start, File path) {
```
radumereuta: `get` method returns void? Maybe find a better name?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Scanner.java
```diff
@@ -52,11 +53,44 @@ public Module getModule() {
...
+        return tokens.values().stream().map(v -> v._1()).collect(Collectors.toSet());
+    }
+
     // debugging method
-    private TerminalLike getTokenByKind(int kind) {
+    public TerminalLike getTokenByKind(int kind) {
         return tokens.entrySet().stream().filter(e -> e.getValue()._1() == kind).findAny().get().getKey();
     }
 
+    public void getStandaloneScanner(File path) {
```
radumereuta: Another get that returns void.

radumereuta (changes requested):
Looks interesting, although I don't really understand why this happens at kast time, and not kompile, since you usually only parse one type of programs per definition. This way you get more flexibility because you can specify which module to use when generating the program parser, but when do you need that?

Another issue that I would like to point out, is file organization in -kompiled. Since we have so many backends which all generate files, it becomes quite a mess. Directories would make it easier to navigate when debugging. I know, it's such a minor issue, but the fix is also very simple.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -124,13 +135,59 @@ public int run() {
...
+                    } catch(IOException | InterruptedException e) {
+                      throw KEMException.internalError("Failed to execute process.", e);
+                    }
+                }
+            } else {
+              K parsed = kread.prettyRead(parsingMod, sort, def, source, FileUtil.read(stringToParse));
+
+              if (options.expandMacros) {
+                  parsed = ExpandMacros.forNonSentences(unparsingMod, files.get(), def.kompileOptions, false).expand(parsed);
+              }
```
ehildenb, line 144: Wow, the logic here got a lot more complicated.

I think having `kast --gen-parser` doesn't make much sense. `kompile --backend bison` or `kompile --emit-bison-parser` makes more sense to me. I also agree that the generated parser should go in the `kompiled` directory, and should form a standalone binary/executable which the user can call directly (without having to invoke immediately via `kast`).

If you absolutely must keep it here, put all the bison generation stuff it its own method and call it under the `if (options.genParser)` instead of inlining it, so that it's easier to see the overall logic of this main method.

It would also make sense to me to add `kast --use-bison-parser` and `krun --use-bison-parser`, which would use the parser emitted during `kompile`.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Scanner.java
```diff
@@ -52,11 +53,44 @@ public Module getModule() {
...
+                flex.append(t.regex());
+            }
+            writeStandaloneAction(flex, key);
+        }
+        try {
+            FileUtils.write(path, flex);
+        } catch (IOException e) {
+            throw KEMException.internalError("Failed to write file for scanner", e);
+        }
+    }
```
ehildenb, line 97: Is there no way for `getStandaloneScanner` and `getScanner` to re-use chunks amongst themselves?

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
ehildenb: Or, since `kompile --emit-bison-parser` will make a standalone executable, should be able to say `krun --parser path/to/definition-kompiled/bison-parser`, and no more options are needed?

ehildenb:
Need to add at least one test of the new generated parser, if for no reason but to document how to use it in practice.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
dwightguth: it's in the temp directory because they're temporary files; if you want to keep them around, you can pass --debug, but otherwise there's no need for them to be kept around after the program terminates.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
dwightguth: You cannot say `krun --parser` because krun currently doesn't know how to deserialize and execute something that has already been translated into kore. But since the whole point of this is to /not/ invoke krun at all, I think that's fine.

radumereuta (approved):

ehildenb (approved):

PR #1010, opened 2020-01-09
kernel: KProveOptions: --concrete-rules is now implemented in front-end

denis-bogdanas:
Option `--concrete-rules` is now implemented in the front-end and is available for any backend. Right now only `KProveOptions` has it but it can easily be added to other proof-related tools like CBMC.

ktest.mak changes: extended to allow definition .k files in other directory than current running directory. This is required for this PR, because I want to run same compiled definition with different sets of kprove options, to test correct caching/non-caching.


haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
```diff
@@ -20,9 +22,10 @@
     @Inject
     public HaskellBackend(
             KompileOptions kompileOptions,
+            KProveOptions kproveOptions,
```
dwightguth: Architecturally, this is wrong. It should not require a KProveOptions object to construct something unless you are within the kprove tool's code path, but backends can also be constructed by kompile

dwightguth (changes requested):

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
denis-bogdanas: Yes I'm aware of this. Actually there are other changes needed that will make access to KProveOptions unnecessary.

dwightguth (approved):

PR #1025, opened 2020-01-14
Stopwatch print time of step

radumereuta:
instead of cumulative time


radumereuta:
A better fix for #956

denis-bogdanas (approved):

PR #1037, opened 2020-01-16
FST serializer: fixes to intermittent, concurrency failures

denis-bogdanas:
Without this PR K Jenkins fails in ~20% of runs. When definition cache will be merged, it is expected to fail more often.
Proof, 15 K builds passed on Jenkins: https://github.com/kframework/k/pull/1034

References:
https://github.com/RuedigerMoeller/fast-serialization/issues/274

### DIFF
--- a/k-distribution/src/main/scripts/lib/checkJava
+++ b/k-distribution/src/main/scripts/lib/checkJava
@@ -40,7 +40,9 @@ else
     if [ $ARCH -eq 64 ]; then
       TIERED=-XX:+TieredCompilation
     fi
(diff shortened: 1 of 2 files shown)


dwightguth (approved):
From what I can see, Java 8 already uses ParallelGC as the default, so that's not a change, but newer versions of Java use the G1 collector by default. Apparently ParallelGC has higher throughput than G1 but at the expense of higher pause times. Given that the Java code in K is not typically interactive, throughput is more important than latency, so this seems like a reasonable alteration to me.

PR #1039, opened 2020-01-16
Significant performance improvements to parsing and compilation time

dwightguth:
Includes several different changes:

* we now use lock-free synchronization in order to deal with initializing type inferencers in parser threads, which should remove a significant source of lock contention
* we don't call ParseInModule.initialize() unless at least one bubble in the module is a cache miss
* we call the implementation of ParseInModule.getCombinedGrammar lazily
* I fixed a serious performance problem in the constructor for ExpandMacros.


kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -92,19 +100,51 @@ public Module seedModule() {
...
      */
     public Module getExtensionModule() {
-        return extensionModule;
+        Module extM = extensionModule;
+        if (extM == null) {
+            Tuple3<Module, Module, Module> mods = RuleGrammarGenerator.getCombinedGrammarImpl(seedModule);
+            extM = mods._1();
+            disambModule = mods._2();
+            parsingModule = mods._3();
+            extensionModule = extM;
```
ehildenb, line 109: Can we factor out a `private void loadModules()` here which all of `getExtensionModule()`, `getParsingModule()`, and `getDisambiguationModule()` can use?

ehildenb: Or perhaps `initializeModules()` to mirror the `initialize()` below.

ehildenb:
I see several places where `parsingModule` is used directly in `ParseInModule.java`, without going through `getParsingModule()` loader function. Is it intentional?

ehildenb:
Sorry, not `parsingModule`, but `disambModule`.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
dwightguth: I mean, the implementation of the three methods is different... I'm not really sure how I would go about doing that. I'm not opposed, but I'm just unclear how that would be done.

dwightguth:
The code is easier to read without the call to the getter, and it is already guaranteed to be initialized by that point in the code, so I think it makes sense to leave it as it is.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
ehildenb: Well you could have a codeblock which says:

```
private void initializeModules() {
            Tuple3<Module, Module, Module> mods = RuleGrammarGenerator.getCombinedGrammarImpl(seedModule);
            extensionModule = mods._1()
            disambModule = mods._2();
            parsingModule = mods._3();
}
```

Then just call that function in each of these three functions.

Your call though.

ehildenb (approved):

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
dwightguth: Unfortunately that is not possible either because in order to implement the racy single-check pattern, we need slightly different assignment statements in each of the three functions.

PR #1040, opened 2020-01-16
Add option --skip-invoking-backend for Java backend

### LINKED ISSUE #1036: Port --dry-run to Java backend
As discussed with @ehildenb, it would be useful in order to have faster feedback about parsing errors.

ehildenb:
Fixes: #1036


ehildenb:
@asymmetric requested this feature for making a local IDE like thing integrated with KLab which can check if there are parsing/kompiling issues before pushing to a server which actually runs the proof.

asymmetric:
We're probably going to run this on the CI server, in a "sanity check" job that is a prerequisite of the actual prover job.

Just to clarify that we're not expecting sub-second parsing feedback.

dwightguth (approved):

PR #1042, opened 2020-01-16
Makefile system: tweaks to make it compatible with global --debug option

denis-bogdanas:
If some intermittent bug appears in future, this PR will allow to easy enable --debug in Jenkins.


denis-bogdanas:
Ready for review.

dwightguth (approved):

PR #1054, opened 2020-01-23
fix bug involving interaction of function context and macros

dwightguth:
This fixes a bug in the C semantics, but I couldn't figure out how to reproduce the issue in a standalone K definition, so that's why there isn't a regression test...

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/ExpandMacros.java
+++ b/kernel/src/main/java/org/kframework/compile/ExpandMacros.java
@@ -220,6 +220,9 @@ public class ExpandMacros {
                                 public K apply(KVariable k) {
                                     K result = subst.get(k);
                                     if (result == null) {
+                                      if (k.name().equals("_Configuration")) {
+                                        return k;
+                                      }
                                       result = newDotVariable(k.att());
                                       subst.put(k, result);
                                     }
--- a/kernel/src/main/java/org/kframework/compile/ResolveFunctionWithConfig.java
+++ b/kernel/src/main/java/org/kframework/compile/ResolveFunctionWithConfig.java
@@ -233,9 +233,6 @@ public class ResolveFunctionWithConfig {
     }
 
     public Sentence resolve(Module m, Sentence s) {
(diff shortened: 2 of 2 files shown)


kernel/src/main/java/org/kframework/compile/ExpandMacros.java
```diff
@@ -220,6 +220,9 @@ public K apply(KApply k) {
                                 public K apply(KVariable k) {
                                     K result = subst.get(k);
                                     if (result == null) {
+                                      if (k.name().equals("_Configuration")) {
```
ehildenb, line 223: I'm not entirely sure what's going on here, but I don't like that we're calling out variables by name specifically. Would this be affected if I named a variable as `BLAH_Configuration` in my definition?

kernel/src/main/java/org/kframework/compile/ResolveFunctionWithConfig.java
```diff
@@ -233,9 +233,6 @@ public Sentence resolveConfigVar(Sentence s) {
     }
 
     public Sentence resolve(Module m, Sentence s) {
-        if (ExpandMacros.isMacro(s)) {
-            return s;
-        }
```
ehildenb, line 238: So now we are expanding macros on macro rules? Just want to make sure that I understand what this is doing.

kernel/src/main/java/org/kframework/compile/ExpandMacros.java
dwightguth: No, it's an exact string match, and this particular string is not part of the syntax of variables in K, so it can't be written by the user. This isn't really a new design choice though; the function context pass was already written this way, we're just fixing a bug in how it interacts with macros.

kernel/src/main/java/org/kframework/compile/ResolveFunctionWithConfig.java
dwightguth: This is applying the function context pass to macro rules.

ehildenb (approved):

Issue #1046, opened 2020-01-20
kprove: should error given unbound RHS variable


MrChico:
If a variable appears only on the right hand side of a specification, this is silently interpreted as an existentially quantified variable. 
This is potentially dangerous behavior, as a typo in the spec writing can lead to a proof becoming trivial, resulting in a false certainty in the verification process.
For rewrite rules, variables appearing on the RHS of a rule throws an error, unless explicitly prefixed by a `?` https://github.com/kframework/kore/blob/master/docs/2019-12-16-Question-Mark-Variables-With-Ensures.md. 

It would be reasonable to expect `kprove` to have the same behavior, and throw an error if a variable appears only on the RHS (unless prefixed with a `?`). 

Cf. the conversation [here](https://riot.im/app/#/room/!MfVXcmpKuyudZHcqil:matrix.org/$1579293981718150SvIjN:matrix.org)

asymmetric:
Seems like this has been [addressed](https://github.com/kframework/k/pull/1058#issuecomment-579163219) in recent versions with a warning, so I think this can be closed?

ehildenb:
I have confirmed with the following example (`test.k`):

```
module TEST
    imports INT

    configuration <k> $PGM:Pgm </k>

    syntax Pgm ::= "a" | "b"
endmodule

module TEST-SPEC
    imports TEST

    claim <k> a => V:Pgm </k>
endmodule 
```

And the following commands:

```
% kprove --spec-module TEST-SPEC test.k 
[Error] Compiler: Found variable V on right hand side of rule, not bound on
left hand side. Did you mean "?V"?
        Source(/home/dev/src/k/test.k)
        Location(14,20,14,21)
        14 |        claim <k> a => V:Pgm </k>
           .                       ^
[Warning] Compiler: Variable 'V' defined but not used. Prefix variable name
with underscore if this is intentional.
        Source(/home/dev/src/k/test.k)
        Location(14,20,14,21)
        14 |        claim <k> a => V:Pgm </k>
           .                       ^
[Error] Compiler: Had 1 structural errors.
```

That we are indeed letting the user know that the variable is not bound correctly.

PR #1058, opened 2020-01-27
Checking verification modules similarly to regular modules

traiansf:
Up to now, `kprove` was not performing checks (similar to those performed at definition parsing time) on the spec files. 

Since we know there are specs where vars occurring only in the RHS are not existentially quantified, I decided not to stop execution when checks are violated, and just to report them at the end of the execution.

Once we merge this PR we should let interested parties know about it and advise them to update their specs asap.


dwightguth:
If you are going to put programs with error messages in the output files in the k regression test suite, you will need to modify ktest.mak to make sure that they get run through the sed snipped that transforms absolute paths into relative paths within error messages, so that the text seen by diff will be the same on all machines.

traiansf:
Fixes #1046 (for now, no hard error, but just warnings).

ttuegel (commented):
This seems to implement the behavior we agreed upon, but I would like someone with more frontend experience to review it.

k-distribution/include/ktest.mak
```diff
@@ -38,7 +38,9 @@ KAST_TESTS?=$(wildcard $(TESTDIR)/*.kast)
 # default KOMPILE_BACKEND
 KOMPILE_BACKEND?=llvm
 
-CHECK=| diff -
+CHECK?=| diff -
+REMOVE_PATHS=sed 's!'`pwd`'/\(\./\)\{0,1\}!!g'
+CHECK_ERRORS=2>&1 | $(REMOVE_PATHS) | diff -
```
dwightguth: unfortunately this will break the update-results target, because sed will not be run in that case and you will populate the output files with results that will not pass tests. The sed needs to be outside the CHECK variable.

kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
```diff
@@ -70,13 +66,15 @@ public ProofDefinitionBuilder(CompiledDefinition compiledDefinition, Backend bac
                 backend.excludedModuleTags());
         Map<String, Module> modulesMap = modules.stream().collect(Collectors.toMap(Module::name, m -> m));
         Definition parsedDefinition = compiledDefinition.getParsedDefinition();
+        Module specModule = getModule(specModuleNameUpdated, modulesMap, parsedDefinition);
+        kompile.structuralChecks(scala.collection.JavaConverters.asScalaSet(modules),
+                specModule, scala.Option.empty(), backend.excludedModuleTags(), false);
```
dwightguth, line 71: it doesn't make sense that you wouldn't throw an error here. You should never have errors in the output displayed by a tool unless the tool has exited abnormally.

dwightguth (changes requested):

kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
traiansf: The idea is to actually throw an error, eventually. However, there are still specifications (including in our Verification Team's work) where these checks would fail. I thought it would be good to implement this change in two steps: first issue the errors, but preserve the existing behavior, to give some time for people to update their specs, then change the behavior to throw a hard error here.

dwightguth: Well if you don't want to break existing specs, emit them first as warnings. But if they're errors, they should halt execution.

dwightguth (approved):
Looks good! Thanks!

PR #1059, opened 2020-01-27
Delete a bunch of outdated and dead code

dwightguth:
All of this stuff is either java/scala code that is completely unused by the current implementation, or k code that is highly out of date and not being used for anything, even teaching. No tests that were currently being run were deleted, nor any java/scala code that was being called by anything in the project.

I would suggest looking at it one commit at a time, but most of this is just git rm -rf of directories.


ehildenb (approved):

nishantjr:
`ktree` at the top-level?

ehildenb:
@nishantjr can you make an issue about it so @dwightguth can look at it tomorrow? Would rather get this merged and have a separate PR for that. As I understand, this list of deletions was mostly found by (1) looking for tests  that are not run, and (2) having IntelliJ perform a code use analysis starting from standard entry-points and removing anything unreachable with that. So would rather keep this PR to the results of that analysis.

dwightguth:
Yeah there's a ktree entry point so I didn't remove it. This was just intended to be a rough and ready PR that deletes a bunch of extremely obvious candidates for deletion, rather than being any kind of in depth effort to delete unused code

dwightguth:
I can make a follow-up that does more granular deletions if you'd like

PR #1061, opened 2020-01-28
Update lambda, imp, and lambda++ lesson 1 to use llvm backend in tutorial

dwightguth:


k-distribution/tutorial/1_k/1_lambda/lesson_5/tests/arithmetic-div-zero.lambda.out
```diff
@@ -1 +0,0 @@
-V ==K 1 /Int 0
```
ehildenb, line 1: Why was this test deleted? Because llvm backend crashes on this output?

k-distribution/tutorial/1_k/1_lambda/lesson_8/exercises/SK-combinators/tests/lambda-6.lambda.out
```diff
@@ -1 +1,3 @@
-V ==K ( SS ( SS ( SS ( KK cond ) ( KK a ) ) ( KK b ) ) ( KK c ) )
+<k>
+  SS ( SS ( SS ( KK cond ) ( KK a ) ) ( KK b ) ) ( KK c )
+</k>
```
ehildenb, line 3: So are all these tests now being `krun` with llvm backend instead of `krun --pattern '<k> V:K </k>'` with Java backend basically?

ehildenb (approved):

ehildenb:
Looks good to me, but if @grosu wants to make sure it won't affect his students that may be a good idea.

Note that it changes from using java backend to llvm backend for a lot of tests, so that may affect those that already have it setup. If they haven't set it up yet though, should be OK.

k-distribution/tutorial/1_k/1_lambda/lesson_5/tests/arithmetic-div-zero.lambda.out
dwightguth: Yes

k-distribution/tutorial/1_k/1_lambda/lesson_8/exercises/SK-combinators/tests/lambda-6.lambda.out
dwightguth: Yes

PR #1064, opened 2020-01-28
Pyk updates and bugfixes

ehildenb:


ehildenb:
The definition minimizer run in the test-suite minimizes the IMP language with respect to the sum.imp program.

The current output that it gives is here: [minimized-sum-imp.k.txt](https://github.com/kframework/k/files/4382025/minimized-sum-imp.k.txt)

This can be improved by using the pre-kompiled definition instead, as long as the pre-kompiled definition has the rule hashes. Otherwise we'll have to use location maps to get the correct rules out (which is also definitely possible).

gtrepta (approved):

PR #1066, opened 2020-01-29
existentially quantify fresh anonymnous variables

traiansf:
So far `?_` were handled similarly to `_` variables and their distinction was lost during the resolving phase. Now the distinction is maintained by adding a fresh attribute, and fresh anonymous variables are being existentially quantified upon transforming them to KORE.


k-distribution/tests/regression-new/haskell-backend-kprove/verif.k
```diff
@@ -7,12 +7,19 @@ module VERIF-SYNTAX
...
+    syntax Pgm ::= addCounter ( Int )
 endmodule
 
 module VERIF
     imports VERIF-SYNTAX
 
-    configuration <k> $PGM:Pgm </k>
+    configuration <k> $PGM:Pgm </k> <counter> 1 </counter> <sum> 0 </sum>
+
+    rule addCounter(0) => .
```
ehildenb:
```suggestion
    rule <k> addCounter(0) => . ... </k>
```

I want to get people out of the habit of omitting `<k>` tags. It makes it quite confusing in definitions which use a lot of `function`s.

ehildenb (approved):

ehildenb:
Looks good to me, but please accept my suggestion.

PR #1068, opened 2020-01-29
Improvements to search

dwightguth:
This PR adds two improvements to search in the haskell backend.

1. We do not display the substitution of anonymous variables in search patterns in search results unless that anonymous variable appears somewhere else inside the substitution. This behavior can be disabled via a flag in order to obtain the original substitution.
2. We add a new flag to kompile, --superstrict, which modifies the behavior of strict. This flag has two effects:
  * We remove side conditions from cooling rules
  * We add a new cell to the configuration, `<generatedProgress>` which is used to automatically transform the definition so that a cooling rule can never apply immediately following a heating rule. This gives us the full range of nondeterminism associated with strictness, including cases where you partially heat one argument, then cool it, then heat another argument, without introducing cycles into the transition system.


ttuegel (approved):
This looks good to me!

traiansf:
I'm kind of OK with this, but wouldn't it be better if the "progress" cell would stay close to the cell in which strictness is happening?  I'm asking this because a global "progress" cell doesn't seem like the best idea for a multithreaded definition.

dwightguth:
You raise a good point, but this issue isn't really high priority enough for me to want to spent time refactoring it at the current moment. I will bear that in mind as something to investigate if multithreaded definitions seem to be missing behaviors I would expect them to have, but otherwise I'm going to keep it as is for now.

ehildenb:
I agree with @traiansf that it seems necessary that we make one generatedProgres cell per k cell. I don't think this is something we can skip on, because it's pretty clear it will cause weirdness in a multi-threaded environment. A simple example:

thread 2 applies heating rule
thread 2 applies some semantic rule
thread 1 applies heating rule
-> Now thread 2 cannot apply a cooling rule

The changes to output look good to me, so if you want to pull those out into their own PR, we can get that merged.

traiansf (approved):
@dwightguth convinced me that while this is not perfect, it's still better than the current setting. But maybe we should find some time later for finding a better solution for this issue.

dwightguth:
Closing this as part of this PR will not get merged. I will reopen with just the commits that are relevant.

PR #1069, opened 2020-01-30
Added some documentation for RHS vars usage scenarios

traiansf:


dwightguth:
This attribute seems sketchy to me. Wouldn't it be better if we had some way of marking individual variables as universally quantified on the rhs?

traiansf:
> This attribute seems sketchy to me. Wouldn't it be better if we had some way of marking individual variables as universally quantified on the rhs?

I agree. I only put it there as a placeholder, in some sense.  There was no agreement on such a label.

How about something like `rhs-only(X, Y)` ?

traiansf:
@dwightguth I've updated the documentation to match the `unboundVariables` attribute introduced in the other PR. could you please take another look at it?

dwightguth (approved):

dwightguth:
This looks good to me but I'd like someone from the haskell backend to review it also.

ttuegel (approved):
This seems alright to me, but why not require explicit quantification with `#Forall`?

traiansf:
> This seems alright to me, but why not require explicit quantification with `#Forall`?

I guess it would make it harder to write; moreover, all variables are implicitly universally quantified; why quantify some and not others?

ttuegel:
> all variables are implicitly universally quantified

Well, except for the ones that are now banned! :smiley:

dwightguth:
Well, there are cases where you want a variable on the RHS that is unbound that will become bound later, like the rhs of a macro that is used on the lhs, so if you required explicit quantification, then you would have to somehow delete that quantifier later, which would be kinda messy.

PR #1070, opened 2020-01-30
Include LLVM backend header files in build assembly

ehildenb:


ehildenb:
I'm able to successfully run a KEVM program in the debugger using this release tarball, not sure what other functionality I should test.

This also enables me to use the release tarball to build Firefly using a binary package of K instead of building K from source, which means a 30mins/CI run speedup for the users of Firefly.

dwightguth:
As far as I can tell, this code will not be able to pretty print variables in the debugger on the following platforms:

* platform independent tarball
* mac os

The second of these two can be fixed with a fairly trivial change to how the brew bottle is created. You just need to pass the value of `#{prefix}` to `mvn package` using the `llvm.backend.prefix` maven variable. The first I'm not particularly worried if it doesn't work, although it's possible you are. Let me know your thoughts.

dwightguth (approved):
This seems at least as good as what was there before, and any problems it might have are likely related to debugging on Mac OS or with the tarball, which we can worry about later.

PR #1080, opened 2020-02-04
Collection cell documenation

ehildenb:
We're missing documentation of cell collection attributes.

I added some explanatory text, and some simple examples.

### DIFF
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -512,11 +512,17 @@ rule <k> introduce-set(I:Int) => . ... </k>
 rule <k> eliminate-set => I ... </k>
      <sets> <set> I </set> => .Bag </sets>
 
-rule <k> introduce-list(I:Int) => . ... </k>
+rule <k> introduce-list-start(I:Int) => . ... </k>
+     <lists> (.Bag => <list> I </list>) ... </lists>
+
+rule <k> introduce-list-end(I:Int) => . ... </k>
      <lists> ... (.Bag => <list> I </list>) </lists>
 
-rule <k> eliminate-list => I ... </k>
+rule <k> eliminate-list-start => I ... </k>
      <lists> (<list> I </list> => .Bag) ... </lists>
+
+rule <k> eliminate-list-end => I ... </k>
+     <lists> ... (<list> I </list> => .Bag) </lists>
 ```
 
 Notice that for `multiplicity="?"`, we only admit a single `<set>` instance at
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -492,7 +492,8 @@ configuration <k> $PGM:Pgm </k>
               <maps>
                 <map multiplicity="*" type="Map">
                   <map-key> 0:Int </map-key>
-                  <map-value> "":String </map-value>
+                  <map-value-1> "":String </map-value-1>
+                  <map-value-2> 0:Int     </map-value-2>
                 </map>
               </maps>
 ```
@@ -533,23 +534,41 @@ cell collection in all cases.
 Declaring `type="Map"` indicates that the first sub-cell will be used as a
 cell-key. This means that matching on those cells will be done as a map-lookup
 operation, so rules which match on the subcells _must_ mention the cell-key.
+Note that there is no special meaning to the name of the cells (in this case
+`<map>`, `<map-key>`, `<map-value-1>`, and `<map-value-2>`). Additionally, any
+number of sub-cells are allowed, and the _entire_ instance of the cell
+collection is considered part of the cell-value, including the cell-key
+(`<map-key>` in this case) and the surrounding collection cell (`<map>` in this
+case).
 
-For example, the following rules introduce, retrieve from, and eliminate
+For example, the following rules introduce, set, retrieve from, and eliminate
 `type="Map"` cells:
 
 ```k
 rule <k> introduce-map(I:Int) => . ... </k>
      <maps> ... (.Bag => <map> <map-key> I </map-key> ... </map>) ... </maps>
 
-rule <k> retrieve-map(I:Int) => S ... </k>
-     <map> <map-key> I </map-key> <map-value> S </map-value> </map>
+rule <k> set-map-value-1(I:Int, S:String) => . ... </k>
+     <map> <map-key> I </map-key> <map-value-1> _ => S </map-value-1> ... </map>
+
+rule <k> set-map-value-2(I:Int, V:Int) => . ... </k>
+     <map> <map-key> I </map-key> <map-value-2> _ => V </map-value-2> ... </map>
+
+rule <k> retrieve-map-value-1(I:Int) => S ... </k>
+     <map> <map-key> I </map-key> <map-value-1> S </map-value-1> ... </map>
+
+rule <k> retrieve-map-value-2(I:Int) => V ... </k>
+     <map> <map-key> I </map-key> <map-value-2> V </map-value-2> ... </map>
 
 rule <k> eliminate-map(I:Int) => . ... </k>
      <maps> ... (<map> <map-key> I </map-key> ... </map> => .Bag) ... </maps>
 ```
 
 Note how each rule makes sure that `<map-key>` cell is mentioned, and we
-continue to use `.Bag` to indicate the empty collection.
+continue to use `.Bag` to indicate the empty collection. Also note that
+when introducing new map elements, you may omit any of the sub-cells which are
+not the cell-key. In case you do omit sub-cells, they will recieve the default
+value given in the `configuration ...` declaration.
 
 Rule Declaration
 ----------------
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -470,6 +470,81 @@ invocations of `krun`. Additionally, we state that the default exit-code is `1`
 that the test fails until proven otherwise and only set the `<status-code>` cell
 to `0` if the test succeeds.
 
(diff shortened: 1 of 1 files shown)


pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+     <sets> <set> I </set> => .Bag </sets>
+
+rule <k> introduce-list(I:Int) => . ... </k>
+     <lists> ... (.Bag => <list> I </list>) </lists>
+
+rule <k> eliminate-list => I ... </k>
+     <lists> (<list> I </list> => .Bag) ... </lists>
+```
+
+Notice that for `multiplicity="?"`, we only admin a single `<set>` instance at
```
kmbarry1: Should "admin" be "admit"? I'm not sure what "admin" would mean as a verb here (maybe just short for "administer")?

ehildenb: Yeah it should, fixed.

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+indicates that the cell may appear any number of times in a runtime
+configuration. Setting `multiplicity="?"` indicates that the cell may only
+appear exactly 0 or 1 times in a runtime configuration.
+
+`type` can take on values `Set`, `List`, and `Map`. For example, here we declare
+several collecion cells:
+
+```k
+configuration <k> $PGM:Pgm </k>
+              <sets>  <set  multiplicity="?" type="Set">  0:Int </set>  </sets>
```
traiansf, line 494: What is the point of having type `Set` if we have multiplicity `?` ?

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+introduction and elimination rules for these collections:
+
+```k
+rule <k> introduce-set(I:Int) => . ... </k>
+     <sets> .Bag => <set> I </set> </sets>
+
+rule <k> eliminate-set => I ... </k>
+     <sets> <set> I </set> => .Bag </sets>
+
+rule <k> introduce-list(I:Int) => . ... </k>
```
traiansf: Maybe write both list introduction rules?

traiansf: (and elimination). I know you said it in words, but the rules are faster to comprehend, IMO

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+     <lists> ... (.Bag => <list> I </list>) </lists>
+
+rule <k> eliminate-list => I ... </k>
+     <lists> (<list> I </list> => .Bag) ... </lists>
+```
+
+Notice that for `multiplicity="?"`, we only admit a single `<set>` instance at
+a time. For the `type=List` cell, we can add/eliminate cells from the from or
+back of the `<lists>` cell. Also note that we use `.Bag` to indicate the empty
```
traiansf, line 535: The fact that we use `.Bag` seems a bit confusing...  I would prefer using `.` without any qualification then `.Bag` when I think about it as a `Set` or `List`, or `Map`

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+cell collection in all cases.
+
+Declaring `type="Map"` indicates that the first sub-cell will be used as a
+cell-key. This means that matching on those cells will be done as a map-lookup
+operation, so rules which match on the subcells _must_ mention the cell-key.
```
traiansf: So the name of the key cell does not have any influence?  Maybe that should be explicitly mentioned.  Would we want to advise as good practice to use a naming convention, though?

traiansf: Also, can we have more than two subcells?  As far as I remember, the map values are actually the contents of the entire cell, including the key.   if so, maybe we should say so.

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+
+For example, the following rules introduce, retrieve from, and eliminate
+`type="Map"` cells:
+
+```k
+rule <k> introduce-map(I:Int) => . ... </k>
+     <maps> ... (.Bag => <map> <map-key> I </map-key> ... </map>) ... </maps>
```
traiansf, line 554: This rule is interesting, because it says the values would be initialized with the defaults from the initial configuration.  Maybe we should mention that.

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+
+Sometimes a semantics needs to allow multiple copies of the same cell, for
+example if you are making a concurrent multi-threading programming language.
+For this purpose, K supports the `multiplicity` and `type` attributes on cells
+declared in the configuration.
+
+`multiplicity` can take on values `*` and `?`. Declaring `multiplicity="*"`
+indicates that the cell may appear any number of times in a runtime
+configuration. Setting `multiplicity="?"` indicates that the cell may only
+appear exactly 0 or 1 times in a runtime configuration.
```
traiansf: Maybe we should mention the effect that `*` and `?` have on building the initial configuration.

pending-documentation.md
```diff
@@ -457,6 +457,94 @@ A single configuration cell containing an integer may have the "exit"
...
+              <lists> <list multiplicity="*" type="List"> 0:Int </list> </lists>
+              <maps>
+                <map multiplicity="*" type="Map">
+                  <map-key> 0:Int </map-key>
+                  <map-value> "":String </map-value>
+                </map>
+              </maps>
+```
+
+Declaring `type="Set"` indicates that duplicate occurances of the cell should be
```
traiansf:
```suggestion
Declaring `type="Set"` indicates that duplicate occurrences of the cell should be
```

traiansf (approved):

pending-documentation.md
ehildenb: Either it exists  or not, I figured. Can we have `multiplicity="?"` without `type="Set"`? Not sure.

pending-documentation.md
ehildenb: Well, sure. But I would say open an issue for it rather than fix it before it's documented.

pending-documentation.md

pending-documentation.md
ehildenb: Added, please check it.

pending-documentation.md
ehildenb: Added a sentence towards the end explaining what happens when sub-cells are missing.

pending-documentation.md
ehildenb: I've added both left and right elim/intro rules.

pending-documentation.md
ehildenb: I've added a second sub-cell, and made it clear that the cell names mean nothing. I also mention that the cell map value includes the entire collection cell, though I think that may be confusing for users, but we'll see.

pending-documentation.md
dwightguth: it should really be .Cells but yeah, we can change that separately since it's not a backwards compatible change.

pending-documentation.md
dwightguth: You actually do not need to mention the cell key. It will just do a nondeterministic choice if you don't.

pending-documentation.md
```diff
@@ -457,6 +457,120 @@ A single configuration cell containing an integer may have the "exit"
...
+rule <k> eliminate-map(I:Int) => . ... </k>
+     <maps> ... (<map> <map-key> I </map-key> ... </map> => .Bag) ... </maps>
+```
+
+Note how each rule makes sure that `<map-key>` cell is mentioned, and we
+continue to use `.Bag` to indicate the empty collection. Also note that
+when introducing new map elements, you may omit any of the sub-cells which are
+not the cell-key. In case you do omit sub-cells, they will recieve the default
+value given in the `configuration ...` declaration.
+
```
dwightguth, line 578: you can only omit cells if you have a ... in the parent.

traiansf (approved):
Looks good! Thanks!

ehildenb:
@dwightguth this one still needs your review.

dwightguth (changes requested):

pending-documentation.md
```diff
@@ -457,6 +457,120 @@ A single configuration cell containing an integer may have the "exit"
...
+Sometimes a semantics needs to allow multiple copies of the same cell, for
+example if you are making a concurrent multi-threading programming language.
+For this purpose, K supports the `multiplicity` and `type` attributes on cells
+declared in the configuration.
+
+`multiplicity` can take on values `*` and `?`. Declaring `multiplicity="*"`
+indicates that the cell may appear any number of times in a runtime
+configuration. Setting `multiplicity="?"` indicates that the cell may only
+appear exactly 0 or 1 times in a runtime configuration. In both cases, the
```
dwightguth: this is not true. If a multiplicity cell contains configuration variables it will start with exactly one instance of that cell.

pending-documentation.md
ehildenb: Fixed.

pending-documentation.md

ehildenb:
@dwightguth I've addressed your comments, please re-review.

pending-documentation.md

dwightguth (approved):

PR #1082, opened 2020-02-04
default value of core.abbrev is different on Mac OS, leading to 404

### LINKED ISSUE #1078: Can't install on Mac OS X
The Mac installation instructions say 

    brew tap kframework/k "file:///$(pwd)"
    brew install kframework--X.Y.Z.ID.bottle.BN.tar.gz -v

Unfortunately, the first command gives an error:

    Cloning into '/usr/local/Homebrew/Library/Taps/kframework/homebrew-k'...
    fatal: '//Users/asloane/Downloads' does not appear to be a git repository
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights and the repository exists.
    Error: Failure while executing; `git clone file:////Users/asloane/Downloads /usr/local/Homebrew/Library/Taps/kframework/homebrew-k --depth=1` exited with 128.

since it seems to be assuming the URL provided is to a git repository.

Is there a fix to enable installation on Mac? I'm on Catalina if that matters.

dwightguth:
error

I believe this should fix, probably, #1078 once the release happens after being merged, [but I'm not 100% sure. At the very least it is harmless though.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -405,7 +405,7 @@ pipeline {
         }
         sshagent(['2b3d8d6b-0855-4b59-864a-6b3ddf9c9d1a']) {
           sh '''
-            release_tag="v${VERSION}-$(git rev-parse --short HEAD)"
+            release_tag="v${VERSION}-$(git rev-parse --short=7 HEAD)"
             mv bionic/kframework_${VERSION}_amd64.deb bionic/kframework_${VERSION}_amd64_bionic.deb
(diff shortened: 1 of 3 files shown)


ehildenb:
@dwightguth maybe you should also change the script `src/main/scripts/brew-update-to-local` as well?

dwightguth:
No, that script doesn't use --short

ehildenb (approved):

PR #1083, opened 2020-02-05
support priority attribute in moduleToKORE

yzhang90:


dwightguth:
Again, I would appreciate an example of what this looks like in practice.

dwightguth:
Also, have you tested when the user mixes `owise` and `priority`?

dwightguth:
I would prefer we skip the lowest priority block rather than specifically skipping priority 200, because if the user writes priority > 200, you will get some very weird behavior.

dwightguth (approved):

PR #1085, opened 2020-02-05
Kool on llvm backend

dwightguth:
This removes the last reference to the ocaml backend in the K test suite, aside from regression tests that are duplicated for the llvm backend. That leaves the C semantics the last remaining place in the company where the ocaml backend is still being used.


src/main/scripts/test-in-container
```diff
@@ -9,5 +9,5 @@ echo 'Testing tutorial in user environment...'
 make -j`nproc`
 cd ~
 echo "module TEST imports BOOL endmodule" > test.k
-kompile test.k --backend llvm
+kompile test.k --backend ocaml
```
ehildenb, line 12: What's going on here? Why are we only running this smoke test with ocaml/haskell now?

dwightguth: So, the smoke test exists in order to test that the backends that were not part of the K tutorial still work after being installed via packaging. We also run the tests in the K tutorial. In the past, the only backends not tested by the tutorial were the llvm and haskell backend. Now, they are the ocaml and haskell backends. Hence the change.

ehildenb (approved):

ehildenb:
Looks like tests passed, and changes look good.

Feel free to merge k-exercises, and back out change to branch.

dwightguth:
3 tests seem to have gotten stuck rewriting forever when built via the debian package of K. I will investigate tomorrow.

dwightguth:
As near as I can tell, with everything running in parallel on the kserver, and a 4G heap, the other processes use enough heap space that the increased memory requirement of generating the decision tree for the `profiling` exercises of kool-typed-dynamic causes the GC to severely thrash, which essentially deadlocks everything running at that time and eventually leads to a java OOM after several hours. The solution is to up the heap size given to the k server when executing the exercises.

PR #1100, opened 2020-02-13
Performance improvements to type inferencer

dwightguth:
This contains a few minor tweaks to the type inferencer which seem to substantially improve the performance of the inferencer on the C semantics. The only actual change to the type system as far as I am aware is that we are now treating A:K as an exact type equality rather than a <= constraint. Note that foo:K where foo is not a variable is still a <= constraint.


radumereuta (approved):

radumereuta:
- what happens if I have `rule A:Int => A:Exp`. Do I get an error message?
- if all variables are typed, do you still need to run Z3? I would guess yes because you still have parametric productions for which you want to infer sorts.
- is it worth applying a type check visitor to the AST between collecting the sorted variables and Z3 in order to trim down some ambiguities, and minimize the problem you send to the SMT?

dwightguth:
In order:

* Yes, you get a type error at one of the sites for the typed variable, saying that one sort was expected when the other was required
* Yes, you still need to run z3, for the reason you mention
* It's not necessary to do what you describe because z3 will intelligently trim the search space based on the equalities you mention. Moreover, it is best left to z3 because we cannot necessarily guarantee that the type equality constraint will not be appearing under a disjunction.

PR #1103, opened 2020-02-13
parse modules in parallel

dwightguth:
This is not enabled by default because it can sometimes make things slower. However, it seems to cause about a 3x speedup on the C semantics on my machine.

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -249,11 +249,8 @@ public class DefinitionParsing {
         Module ruleParserModule = gen.getRuleGrammar(defWithConfig.mainModule());
         ParseCache cache = loadCache(ruleParserModule);
         try (ParseInModule parser = RuleGrammarGenerator.getCombinedGrammar(cache.getModule(), isStrict, profileRules, files)) {
-            if (options.parallelModules) {
-                return DefinitionTransformer.parFrom(m -> this.resolveNonConfigBubbles(m, parser.getScanner(), gen), "parsing rules").apply(defWithConfig);
-            } else {
-                return DefinitionTransformer.from(m -> this.resolveNonConfigBubbles(m, parser.getScanner(), gen), "parsing rules").apply(defWithConfig);
-            }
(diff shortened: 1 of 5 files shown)


h0nzZik:
Nice, 3x is a lot. Is it for the parsing time, or overall kompilation time?

dwightguth:
Specifically parsing time. I got it down from 10 minutes before both of the changes (this one and the type inference one) to about 2 minutes. Note that was with a machine with a ton of cores though, so it's probably less if you only have 2 or 4 cores.

dwightguth:
Parse time isn't really the deciding factor in performance on compiling the C semantics after this change though; it's time spend in the backend. For example, I just compiled the execution semantics and parsing took 5m30s and the backend took almost 11m. At this point, if we want to get compile times faster, we will have to do more work to shrink the size of the decision tree that it generates.

radumereuta:
Wait a second. If you look at `DefinitionParsing.java:276, 285, 294` we are already calling the parallel streams from java. Is that not working? That was the intended way to parallelize rule parsing.

dwightguth:
That parallelizes parsing of rules within a single module. It does not parallelize parsing of the modules themselves, and thus if significantly more time is spent initializing modules than parsing rules (ie because you have many modules with lots of syntax but few rules), you end up gaining very little parallelism prior to this patch.

radumereuta:
Something feels off.
If you look at `transformers.scala:111`  you have a mapping that applies the module transformer to all modules. That can be parallelized: `d.entryModules.par.map(moduleTransformer).toList.toSet`. Why aren't you using that?
In any case, this needs more comments to explain why are you choosing this solution.

dwightguth:
I created two transformers: one that doesn't parallelize and one that does. The one on line 111 is the one that doesn't.

I'm not sure what exactly you have in mind in terms of comments. Can you point me to what you think should be commented?

dwightguth:
After talking with Radu I pushed a better version of  this fix that does not have the problem of parsing rules more than once. It seems to exhibit the same 3x speedup on C-semantics, but does not slow anything down for EVM or the other smaller semantics I tried. Indeed, it is marginally faster on EVM as well, but only by a couple seconds.

radumereuta (approved):
I was not expecting it to be that simple.

Issue #985, opened 2019-12-19
K should infer sorts when using abstract syntax


radumereuta:
```
module TEST
  syntax E ::= "a"         [klabel(elma), symbol]
  syntax Lst ::= E "," Lst [klabel(constr), symbol]
               | ".Lst"    [klabel(emptyLst), symbol]
  configuration <k> constr(a, emptyLst(.KList)) </k>
  rule constr(I, L) => L
endmodule
```
This rule should apply, but right now, the sorts for `I` and `L` are being inferred as sort `K`. It used to work in the ocaml backend because we didn't have such strict sort checks and injections.

This is blocking the K-in-K definition. Since we're defining K, we have to avoid using the concrete syntax so we don't get parsing errors. The alternatives would be to either type all variables in every rule or have duplicate productions with syntax only for rules. Both of these solutions are error-prone.

dwightguth:
I would recommend creating macros instead of using abstract syntax.

PR #1106, opened 2020-02-18
Kapp2termcons

radumereuta:
This is supposed to be a fix for #985
I've added a new filter at parsing time that transforms instances of KApp into the equivalent TermCons
This way the type inferencer can apply correctly

I also found a pretty-printing issue in TreeNodesToKORE.scala and simplified the code a bit.


kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+        super();
+        this.mod = mod;
+    }
+
+    @Override
+    public Either<java.util.Set<KEMException>, Term> apply(TermCons tc) {
+        assert tc.production() != null : this.getClass() + ":" + " production not found." + tc;
+        if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#KApply")) {
+            if (!(tc.items().get(1) instanceof Constant) || !((Constant) tc.items().get(1)).production().sort().equals(Sorts.KLabel()))
+                // TODO: maybe return a hidden warning?
```
radumereuta: Not sure about adding a warning here.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+    public KAppToTermConsVisitor(Module mod) {
+        super();
+        this.mod = mod;
+    }
+
+    @Override
+    public Either<java.util.Set<KEMException>, Term> apply(TermCons tc) {
+        assert tc.production() != null : this.getClass() + ":" + " production not found." + tc;
+        if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#KApply")) {
+            if (!(tc.items().get(1) instanceof Constant) || !((Constant) tc.items().get(1)).production().sort().equals(Sorts.KLabel()))
```
dwightguth: Let's use TermCons.get everywhere so we don't get confused by the reversed indices

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+                return super.apply(Ambiguity.apply(tcs, tc.location(), tc.source()));
+            }
+        }
+        return super.apply(tc);
+    }
+
+    /**  Recurse under #KList and flatten all the terms */
+    private static PStack<Term> flattenKList(Term t) {
+        if (t instanceof Ambiguity) {
+            Ambiguity amb = (Ambiguity) t;
```
dwightguth: The logic in this block doesn't look sound to me. You are basically dropping some of the parses in the forest

kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
```diff
@@ -29,21 +29,23 @@ class TreeNodesToKORE(parseSort: java.util.function.Function[String, Sort], stri
...
   }
 
-  def termConsToKApply(t: TermCons, items: PStack[Term], p: Production): KApply = {
-    if (p.att.contains("recordPrd", classOf[Production])) {
-      val realProd = p.att.get("recordPrd", classOf[Production])
-      val map = new util.ArrayList(items).asScala.reverse.zipWithIndex.map { case (item, idx) => (p.nonterminal(idx).name.get, apply(item))} toMap
+  def termConsToKApply(t: TermCons): K = {
+    if (t.production.att.contains("recordPrd", classOf[Production])) {
+      val realProd = t.production.att.get("recordPrd", classOf[Production])
+      val map = t.items.asScala.reverse.zipWithIndex.map { case (item, idx) => (t.production.nonterminal(idx).name.get, apply(item))} toMap
```
dwightguth: This needs to be converted to an arraylist still because otherwise reverse will take quadratic time.

kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
```diff
@@ -29,21 +29,23 @@ class TreeNodesToKORE(parseSort: java.util.function.Function[String, Sort], stri
...
-        throw KEMException.internalError("Missing klabel in production: " + p, t)
-      val klabel = if (p.klabel.get.name == "#OuterCast") KLabel("project:" ++ p.sort.toString) else p.klabel.get
-      KApply(klabel.head, KList(new util.ArrayList(items).asScala.reverse map apply asJava), locationToAtt(t.location, t.source).add(classOf[Production], realProd))
+      val realProd = if (t.production.att.contains("originalPrd", classOf[Production])) t.production.att.get("originalPrd", classOf[Production]) else t.production
+      if (t.production.att.contains("bracket"))
+        return apply(t.items.get(0))
+      if (t.production.klabel.isEmpty)
+        throw KEMException.internalError("Missing klabel in production: " + t.production, t)
+      val klabel = if (t.production.klabel.get.name == "#OuterCast") KLabel("project:" ++ t.production.sort.toString) else t.production.klabel.get
+      KApply(klabel.head, KList(t.items.asScala.reverse map apply asJava), locationToAtt(t.location, t.source).add(classOf[Production], realProd))
```
dwightguth: Same here

dwightguth (changes requested):

dwightguth:
It looks alright aside from the minor issues I mentioned, but it really should be the first step in the disambiguation pipeline, not left until after priority is done.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
radumereuta: I am applying an early `prefer` filter, only for `#KList`. If I get a KApp, I gather all the KList items and put them in a single list. If I get different ambiguities, like user lists, I discard them since I'm only interested in KList in this place.
I want to apply this filter in exactly this place because it simplifies this part. It is possible to have user lists mixed in, then the number of KList items is going to be smaller. That's why I'm only taking the list with the highest amount of elements. Since KList has the left attribute, I'm only going to find one element with the highest amount of items.

It's possible that I have a production with arity 2, but a KList with 3 elements. Then that's going to be an error. You can still write that if you add a set of braces (see test:495)

If I make it the first filter, I have to handle more cases when flattening KList. It could also hinder priorities. I can write `mul(plus(1,2),3)` but when I transform it into TermCons, it's going to be rejected by the priority filter.

radumereuta:
I've fixed most of the comments and answered to another.
I would also want you to look at the `TODO` I've added. I'm not sure what to do if I fail to transform the KApp into TermCons (cases like a KApp with a variable instead of KLabel constant). One suggestion was to add a hidden warning, but I'm not sure what is the support or future plans for that kind of pattern.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
```diff
@@ -43,57 +51,68 @@ public KAppToTermConsVisitor(Module mod) {
...
+        public Term apply(TermCons tc) {
+            if (tc.production().klabel().isDefined() && tc.production().klabel().get().head().equals(KLabels.KLIST)) {
+                Term v0 = super.apply(tc.get(0));
+                Term v1 = super.apply(tc.get(1));
+                Set<Term> t0 = v0 instanceof Ambiguity ? ((Ambiguity) v0).items() : Sets.newHashSet(v0);
+                Set<Term> t1 = v1 instanceof Ambiguity ? ((Ambiguity) v1).items() : Sets.newHashSet(v1);
+                Set<Term> rez = Sets.newHashSet();
+                for (Term t00 : t0)
+                    for (Term t11 : t1)
+                        rez.add(TermCons.apply(ConsPStack.singleton(t00).plus(t11), tc.production(), tc.location(), tc.source()));
```
dwightguth, line 112: Pretty sure this is reversed from the order you want.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
```diff
@@ -43,57 +51,68 @@ public KAppToTermConsVisitor(Module mod) {
...
+            assert false : KAppToTermConsVisitor.class + " expected all ambiguities to already be pushed to the top:\n" +
+                    "   Source: " + ((Ambiguity) t).items().iterator().next().source().orElse(null) + "\n" +
+                    "   Location: " + ((Ambiguity) t).items().iterator().next().location().orElse(null);
         } else if (t instanceof TermCons) {
             TermCons tc = (TermCons) t;
-            if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#KList"))
-                return flattenKList(tc.items().get(1)).plusAll(flattenKList(tc.items().get(0)));
-            else if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#EmptyKList"))
+            if (tc.production().klabel().isDefined() && tc.production().klabel().get().equals(KLabels.KLIST))
+                return flattenKList(tc.get(0)).plusAll(flattenKList(tc.get(1)));
```
dwightguth: Pretty sure this is reversed too

dwightguth (changes requested):
This needs some tests that will actually break if the children of a TermCons get reordered.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
```diff
@@ -472,4 +474,35 @@ public void testLayout() {
...
+    @Test
+    public void test25() {
+        String def = "" +
+                "module TEST\n" +
+                "  syntax E ::= \"a\"         [klabel(elma), symbol]\n" +
+                "  syntax Lst ::= E \",\" Lst [klabel(constr), symbol]\n" +
+                "  syntax Lst2 ::= E \",\" Lst2 [klabel(constr), symbol]\n" +
+                "                | Lst \n" +
+                "endmodule";
+        parseRule("constr(I, L) => L", def, 0,
```
radumereuta, line 489: Test 1

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
```diff
@@ -472,4 +474,35 @@ public void testLayout() {
...
+                KApply(KLabel("#ruleNoConditions"),KApply(KLabel("#KRewrite"),
+                        KApply(KLabel("constr"),
+                                KApply(KLabel("#SemanticCastToE"), KToken("I",Sort("#KVariable"))),
+                                KApply(KLabel("#SemanticCastToLst2"), KToken("L",Sort("#KVariable")))),
+                        KApply(KLabel("#SemanticCastToLst2"), KToken("L",Sort("#KVariable")))
+                )));
+        parseRule("`constr`(I, (a, L)) => L", def, 0,
```
radumereuta, line 496: Test 2

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
dwightguth: This is wrong, I misunderstood that plus does a prepend in constant time rather than an append.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/KAppToTermConsVisitor.java
dwightguth: I was wrong here too; plusAll prepends the argument in reverse order, so if I have ((1, 2),(3, 4)) as the children, I am going to end up with 1,2,4,3 as the final arguments.

dwightguth: the second flattenKList call needs to be reversed.

dwightguth (approved):

PR #1107, opened 2020-02-18
Document limitations on the memo attribute

### LINKED ISSUE #1540: Document restrictions on the memo attribute
We would like to automatically (without an attribute) cache functions called with constructor-like arguments, but there are some obstacles to this. For the moment, we should just document the limitations of the `memo` attribute, namely:

- a `memo` function must be independent of any side condition; particularly it must not depend on any uninterpreted symbol.
- a `memo` function generally should not be recursive.

See also: #1522

ttuegel:
Fixes: kframework/kore#1540


traiansf (approved):
I wonder whether it would be a good idea to give an example for the warning. Users might be confused about what an uninterpreted function might mean in K.

dwightguth:
It looks good but can we write the documentation in such a way that it's clear that it's not just the haskell backend that might support this attribute? the ocaml backend also has it. In particular, it should say something like "the attribute is a hint to the backend that the user wishes to memoize this function. Not all backends support memoization, but the Haskell and OCAML backends do and when this attribute is present and the definition is compiled with that backend, calls to the function will be cached."

dwightguth (approved):

PR #1110, opened 2020-02-19
print class of exception

### LINKED ISSUE #1108: Ensure exception class is shown in critical errors
When [`StackOverflowError`](https://github.com/kframework/k/blob/d23fb1bb15d31de94ba6dc9424fffcd3ddbd0b35/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java#L584) gets thrown due to e.g. a bug in the semantics, all the user sees is `[Error] Critical: (null)`.

It would greatly help with debugging to know that e.g. a `StackOverflowError` has occurred. In my case, that was due to an error in my semantics where more than one `function` rule matched, which in some cases ended up producing infinite recursion.

dwightguth:
Fixes #1108


traiansf (approved):
Looks good to me; hopefully it will also benefit the person requesting the feature.

PR #1112, opened 2020-02-20
KRunOptions: improved help message for `--config-var`

denis-bogdanas:


dwightguth (commented):
looks good but can we have an example not from a file too just for good measure?

dwightguth (approved):

PR #1113, opened 2020-02-21
Consider the priority attribute for claims

traiansf:
This is needed in order for the `priority` attribute (more precisely, its value) to be visible in the generated `.kore`.

Related question for @dwightguth : I don't remember the rationale for `consideredAttributes` when generating KORE.   Does this date back to the time when we were declaring attributes? if so, is it relevant anymore?


ttuegel (approved):

ttuegel:
@yzhang90

dwightguth:
The map from string to boolean for attributes tells ModuleToKore whether the attribute has a child or not, since k does not otherwise make such a distinction, but kore does.

traiansf:
@ttuegel please disregard the above request fro review; wrong click!

PR #1116, opened 2020-02-22
Adding support for .md files

radumereuta:
with fenced k code blocks

I'm going to need someone to check if the licensing is ok for us:
https://github.com/vsch/flexmark-java

I've looked over several java parsers for Markdown and this provided the location information
I needed to correctly preserve location information.

I've manually tested it on the K-in-K definition and it kompiled all rules w/o errors. Not sure if it's worth adding tests since we're mostly interested in preserving location information here.

I've also thought of printing the output somewhere in -kompiled when `--debug` is given, but I would have to add a new dependency to `ParserUtils`. Not sure if it's worth.


dwightguth:
I would like to see a pr into evm testing that everything works as expected if we remove our dependency on pandoc, and also a test in the regression test suite that we get the right line information even if we have very unusual newlines in and around the code block.

dwightguth:
One other thing I notice is that you don't seem to support conditionally including blocks.

ehildenb:
Seconded that we shouldn't merge this until we know it works with KEVM.

radumereuta:
@ehildenb I've been looking at how pandonc-tangle works and I can implement the same thing in kompile. But I have a slight hesitation. Other tools do not recognize the `css-selector`. For example, IntelliJ can highlight the K syntax even if in markdown if it has the `k` tag for a codeblock.

If this is something we want to standardize in K, I can spend a few more hours and add the rest of the functionality, but right now it looks like something custom and each project has its own little quirks.

Thoughts?

ehildenb:
Having selectors is necessary because there are several semantics that we compile in different ways depending on whether we need a certain codeblock compiled into the definition or not. The projects all use the same tangler, so there shouldn't be any quirks to worry about, but I would prefer better tangling syntax on the command line than what we have now. Certainly having to all the positive _and_ negative cases is quite annoying.

For example (from KEVM):

```
concrete_tangle := .k:not(.node):not(.symbolic):not(.nobytes):not(.memmap),.standalone,.concrete,.bytes,.membytes
java_tangle     := .k:not(.node):not(.concrete):not(.bytes):not(.memmap):not(.membytes),.standalone,.symbolic,.nobytes
haskell_tangle  := .k:not(.node):not(.concrete):not(.nobytes):not(.memmap),.standalone,.symbolic,.bytes,.membytes
node_tangle     := .k:not(.standalone):not(.symbolic):not(.nobytes):not(.memmap),.node,.concrete,.bytes,.membytes
```

could conceivably be just (without the `not` cases, because we assume in K that any codeblocks labeled with _only_ `k` are selected by default, and we put instead the `not` on the codeblocks themselves):

```

concrete_tangle := .concrete,.bytes,.membytes
java_tangle     := .standalone,.symbolic
haskell_tangle  := .standalone,.symbolic,.bytes,.membytes
node_tangle     := .node,.concrete,.bytes,.membytes
```

So I think we should keep the same css selectors in the markdown files itself with the small modification that we can put `:not(bytes)` directly on a codeblock for example.

I'm hesitant about this whole approach though. I would prefer that we just have the Java frontend call pandoc directly (and add pandoc to our optional dependencies). This would make it possible to tangle definitions without invoking the entirety of the Java frontend (just expose a `ktangle` executable script), and gives us more flexibility to use this functionality in alternative frontends directly (eg. `pyk`). I would really prefer a solution based on just writing a bash script which calls pandoc in the correct ways.

One argument I could see swaying my opinion on this is if this library gives us an exact source map (this codeblock on these lines in `target.k` came directly from these lines in `source.md`). That's something that's sorely missing from the existing tangling infrastructure.

@dwightguth what are your thoughts?

ehildenb:
To be clear, I'm fine with a non-pandoc solution, beacuse it seems more straightforward based on this work. I don't think it's a blocker though.

dwightguth:
Correct me if I'm wrong, but unless I'm mistaken, the .node target is the only one of our selectors which is not selecting for a specific set of backends, and that target is pretty much deprecated in favor of the web3 target. Is it possible that we could keep the idea of having tags that toggle on/off certain code blocks, but simply make them be some combination of the `kore`, `concrete`, `symbolic`, and `kast` attributes which are used already to guard module imports? You would still need boolean operators to combine the tags, but it would be a lot more user friendly experience.

dwightguth:
That being said, I think pandoc is a non starter because it doesn't preserve line information the right way.

radumereuta:
I come up with a selector expression based on `& | !`. Ex:`'k&(!a|b)'`. I can also implement the grammar for CSS-Selector if needed, I just thought this is a bit simpler. I don't do anything for dots, they are part of the tag. Here is a translation from the evm semantics:
```
.k:not(.node):not(.symbolic):not(.nobytes):not(.memmap),.standalone,.concrete,.bytes,.membytes
'(.k|k)&!.node&!.symbolic&!.nobytes&!.memmap&(.concrete|.bytes|.membytes)'
```
Notice `(.k|k)`. Suggestions?

Because I use `!`, the single quote is required when using it in command line.
I added a unit test with different weird spaces around code blocks.

Requiring a file now needs to specify the exact extension found on the disk: `require "test.k"` becomes `require "test.md"`. 

If you want to see the .k files that would be generated, just `kompile` with  `--debug`. You can find the files in `X-kompiled/.md2.k`. Again, I'm not doing anything fancy to reproduce the structure of the original files. I just print the list of files. If you have multiple files with the same name, they will get overridden. They are not actually used for compilation, just to check them out, so let me know if I should do something more - might get more complicated.

I tested the evm semantics locally. I looked at the diffs of the generated files and the selected blocks are identical to the pandoc version, but they are exactly in place this time. I didn't do a PR into evm because kompile fails with some LLVM error - I need more time to figure out what dependency I'm missing. So let me know if you need the PR before merging this..

ehildenb:
@radumereuta in CSS selector syntax for Markdown (at least what pandoc and other tools support), you _either_ do `{.TAG}`, or `TAG` on its own.

So "\`\`\`k" is equivalent to "\`\`\`{.k}".

So you can choose to either remove the `.` from the ones that are in curly brackets, or add it to those that aren't.

I'm also not sold on the `.`, so if you want to remove them entirely from the semantics that sounds good to me too. But then we lose compatibility with tools like `pandoc`, which I would like to avoid.

ehildenb:
So actually, I'd like to see the `.` removed from the selector too I suppose. Also, is there a chance you can allow spaces in the selectors?

I'd like to write:

```
k & !node & !symbolic & !nobytes & !memmap & (concrete | bytes | membytes)
```

So maybe the pre-processing on the tags should just remove the `.` from the ones defined in `{ ... }` when reading them, which gives the simpler selector syntax (because we're already breaking away from CSS selectors), but leaves it in the markdown sources (to maintain compatibility with other tools).

radumereuta:
I'm removing the leading dot from tags and from the selector expression. It should work either way.
Whitespaces are supported: `--md-selector 'k & !symbolic & !nobytes & !memmap | concrete | bytes | membytes'`.
I guess now @dwightguth you can also have a look since I'm a bit happier with the result.

ehildenb:
@radumereuta can we still use parentheses for clarity?

radumereuta:
Yeah, you can put parenthesis anywhere.

ehildenb:
Ok. I am happy with the end result then, but I'll leave @dwightguth to comment on the code structure and see if he has any ideas.

Just to make sure @radumereuta , do we get an accurate source map for markdown sources? I would like you to add at least 2 tests:

-   One test which kompiles and kruns fine in `k-distribution/tests/regression-new` but uses `imp.md` instead of `imp.k`.
-   One test which does not kompile fine (has some structural error in the definition), and outputs accurate line/column information about where the error is in the markdown file.

kernel/src/main/java/org/kframework/parser/inner/ExtractFencedKCodeFromMarkdown.java
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2018-2019 K Team. All Rights Reserved.
+package org.kframework.parser.inner;
```
dwightguth: This should be in the outer parsing package.

kernel/src/main/jjtree/TagSelector.jjt
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2013-2020 K Team. All Rights Reserved.
+options {
+  MULTI=true;
+  KEEP_LINE_COLUMN = false;
+  STATIC = false;
+}
+PARSER_BEGIN(TagSelector)
+package org.kframework.parser.tagSelector;
```
dwightguth: Let's make this package named outer.markdown.

kernel/src/main/jjtree/TagSelector.jjt
```diff
@@ -0,0 +1,148 @@
...
+    return tags;
+  }
+
+  public static ASTExpressionStart parseSelectorExp(String input) {
+    TagSelector t = new TagSelector(new StringReader(input));
+    try {
+      ASTExpressionStart result = t.ExpressionStart();
+      return result;
+    } catch (ParseException e) {
+      throw KEMException.compilerError("Could not parse " + input + " as a selector expression.", e);
```
dwightguth: This needs better error message. Take a look at how we format the error messages from javacc in the outer parser.

dwightguth (changes requested):
This looks really good! I agree we need integration tests like Everett suggested, and I have a couple nitpicks, but there's nothing really major wrong that I can see.

The tests I would suggest are:

1. Change tutorial/1_k/2_imp/lesson_5 to use markdown instead of the old style of literate comments. We can change the rest later, but I'd like that lesson at least to showcase everything modern k can offer.

2. Add a test to regression-new that tests that if you have two different rules in two different code blocks for the same function, the css selector selects the right one.

3. A test in regression-new/checks that tests that you get the right line information from the markdown file even in a case that pandoc couldn't handle.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.md
```diff
@@ -1,37 +1,37 @@
...
 
+```k
   syntax AExp  ::= Int | Id
                  | "-" Int                    [format(%1%2)]
                  | AExp "/" AExp              [left, strict, color(pink)]
                  > AExp "+" AExp              [left, strict, color(pink)]
                  | "(" AExp ")"               [bracket]
   syntax BExp  ::= Bool
-                 | AExp "<=" AExp             [seqstrict, latex({#1}\leq{#2}), color(pink)]
+                 | AExp "<=" AExp             [seqstrict, color(pink)]
```
ehildenb: We should keep the `\latex...` output attribute, because that can still affect the `--output latex` option to KAST in the future (does not right now, but will).

ehildenb: Note that this is a separate thing than the sources being in LaTeX.

radumereuta:
I think I got all the changes.
I changed the location of the .md2.k files to the temporary kompile directory, and if I get name collisions, I just append it with some index. This way you can check all files. You can find the complete path inside, at the end.

Makefiles now recognize both .k and .md source files.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -283,7 +308,7 @@ private String loadDefinitionText(File definitionFile) {
         Set<File> requiredFiles = new HashSet<>();
         Context context = new Context();
         if (autoImportDomains)
-            previousModules.addAll(loadModules(new HashSet<>(), context, Kompile.REQUIRE_PRELUDE_K, source, currentDirectory, lookupDirectories, requiredFiles, kore, preprocess, leftAssoc));
+            previousModules.addAll(loadModules(new HashSet<>(), context, Kompile.REQUIRE_PRELUDE_K, Source.apply("auto imported prelude"), currentDirectory, lookupDirectories, requiredFiles, kore, preprocess, leftAssoc));
```
ehildenb: Not sure what's going on here.

ehildenb: Was this a bug before?

ehildenb (approved):
This looks great to me! I'm approving, but we should also get @dwightguth 's review.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
radumereuta: When using -v, it would show that you are including the main file twice.
So yeah... a visual bug.

kernel/src/main/jjtree/TagSelector.jjt
```diff
@@ -0,0 +1,169 @@
...
+      ASTTagListStart result = t.TagListStart();
+      for (Node n : result.children)
+        if (n instanceof SimpleNode) {
+            String tag = ((SimpleNode) n).value.toString();
+            if (tag.charAt(0) == '.')
+                tag = tag.substring(1);
+            tags.add(tag);
+        }
+    } catch (Exception e) {
+      // ignore: e.printStackTrace();
```
dwightguth: This looks bad.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.md
```diff
@@ -1,30 +1,30 @@
 // Copyright (c) 2014-2019 K Team. All Rights Reserved.
```
dwightguth: If you view this file on GitHub you will see that this shows up at the top of the document. Please change this to a markdown comment.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.md
```diff
@@ -43,148 +43,170 @@ argument, because we want to give it a short-circuit semantics. */
...
+Note that we preferred to make the rule below structural.
 
+```k
   rule while (B) S => if (B) {S while (B) S} else {}  [structural]
+```
 
-/*@ \subsection{Programs}
+### Programs
 The semantics of an IMP program is that its body statement is executed
 in a state initializing all its global variables to 0.  Since \K's
```
dwightguth: Looks like you forgot one

dwightguth (changes requested):
Looks great, but there are still a couple minor issues. Please reduce the scope of the exceptions you are catching in that one spot, and do something other than just ignore them (unless the exception can be safely ignored and the function will still return the correct value).

Also, I spotted two very minor issues in imp.md.

radumereuta:
@dwightguth I think tests will pass this time.
It was a bit trickier to give error messages from Markdown. The parser I chose doesn't give line-column error messages but does give accurate offset info. So I had two choices:
1. use the location mapper algorithm found in Parser.java and add an try to add an offset to JavaCC. The downside was I would make this PR even bigger with refactorings to the kernel of the K parser, and I'm not sure if I can even change the offset of the JavaCC parser.
2. I created another string where I replace every character with whitespace except for the code-block tag. I believe this was less invasive and it keeps the error message consistent in the front-end.

dwightguth (approved):
I believe there may still be room for improvement in the error messages relating to malformed code tags, but we can let that be driven by actual use cases. This looks great! thanks for the contribution!

PR #1119, opened 2020-02-25
fix bug in AddEmptyLists where lub sort cannot be determined

dwightguth:
Previously we were getting an error in certain sort graphs saying that the least upper bound sort could not be determined, even though only one lub sort could actually appear in that location based on its expected sort. Here we correctly filter the lub by the expected sort, which we were not previously doing.

### DIFF
--- a/k-distribution/tests/regression-new/lub/test.k
+++ b/k-distribution/tests/regression-new/lub/test.k
@@ -1,0 +1,15 @@
+module TEST
+  imports DOMAINS
+
+  syntax IntPtr ::= PtrValue
+  syntax PtrValue ::= SymLoc
+  syntax SymLoc
+  syntax IntPtr ::= EncodedPtr
+  syntax BitValue ::= EncodedPtr
+  syntax Encodable ::= PtrValue
+  syntax BitValueOrEncodable ::= BitValue | Encodable
+
+  syntax EncodedPtr ::= encodedPtr(PtrValue)
+  syntax Bool ::= sameBase(IntPtr, IntPtr) [function]
+  rule sameBase(_, (encodedPtr(S:SymLoc) => S))
+endmodule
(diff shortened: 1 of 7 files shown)


traiansf (approved):
It looks reasonable to me

k-distribution/tests/regression-new/lub/test.k
```diff
@@ -0,0 +1,15 @@
...
+  syntax PtrValue ::= SymLoc
+  syntax SymLoc
+  syntax IntPtr ::= EncodedPtr
+  syntax BitValue ::= EncodedPtr
+  syntax Encodable ::= PtrValue
+  syntax BitValueOrEncodable ::= BitValue | Encodable
+
+  syntax EncodedPtr ::= encodedPtr(PtrValue)
+  syntax Bool ::= sameBase(IntPtr, IntPtr) [function]
+  rule sameBase(_, (encodedPtr(S:SymLoc) => S))
```
radumereuta, line 14: I guess the rewrite sort here is IntPtr?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AddEmptyLists.java
```diff
@@ -63,6 +63,24 @@ public AddEmptyLists(Module m, Sort expectedSort) {
         this.expectedSort = expectedSort;
     }
 
+    private Tuple2<Either<Set<KEMException>, Term>, Set<KEMException>> superApply(TermCons tc) {
```
radumereuta, line 70: The name of this function looks funny to me. Could you at least add some documentation?

radumereuta (approved):
Is there an issue that should be referenced? How did you find this example?
If you add the documentation you can merge it directly.

denis-bogdanas:
Fixes #1121

denis-bogdanas:
Fixes #1121 
Tested.

PR #1124, opened 2020-02-26
reset z3 process after every query on older z3 versions

dwightguth:
We were getting some segfaults from z3 on z3 4.4.1. I would prefer to support this version since it is the version installed by default on ubuntu bionic. However, we really can't do anything about the bug that causes the crash. That being said, we can make it significantly less likely by creating a fresh z3 process after every rule when the z3 version is old enough to encounter the bug.

I tested and this seems to resolve the issue with the beacon chain, but note that I can't guarantee it won't still occur elsewhere. If it does, we will probably have to tell people to install a newer z3.


radumereuta (approved):
lgtm

traiansf (approved):

PR #1138, opened 2020-03-03
Patch for Issue #1137

### LINKED ISSUE #1137: Haskell backend integration: `kprove` should save a `compiled.bin` for the generated `vdefinition.kore`
This is required to ensure that there would be a definition which can be used to unparse the output of the haskell backend.

Currently kprove does not save the composed definition (including the main module to be used for proofs).  However, the output (esp. if not `#True`), might include syntax defined in that module.

This is the case, for example with the KWASM Lemmas ( https://github.com/kframework/wasm-semantics/blob/master/kwasm-lemmas.md ) which define helping predicates such as
```
     syntax Bool ::= #isByteMap ( ByteMap ) [function, functional, smtlib(isByteMap)]
```

[test.zip](https://github.com/kframework/k/files/4276349/test.zip)

steps to reproduce:
```
kompile  --backend haskell  verif.k -d .
kprove test-spec.k   -d . 
```

Output:
```
file a bug report at https://github.com/kframework/k/issues (key not found:                              
isA(_)_VERIFICATION_Bool_S)
```

traiansf:
Fixes #1137 

@denis-bogdanas I just saw your  comment https://github.com/kframework/k/issues/1137#issuecomment-593923139
However, my PR is already done.  Since you mention that the other PRs are already stalling for quite some time, could  we merge this, as it's a small change, and revert it / adapt it when the other PRs are ready to be merged?


denis-bogdanas (changes requested):
This patch won't work correctly when you run `kprove` for 2 specifications based on same definition. The 2nd spec will load the extended definition produced by 1st spec. It will break VSC for example. I guess you need this for unparsing with `kast`. The correct fix would be to merge the 2 PRs I mentioned, and modify `kast` to identify the correct extended definition in definition cache created there.

Until then I suggest to keep this commit outside master and use it by need.

traiansf:
> This patch won't work correctly when you run `kprove` for 2 specifications based on same definition. The 2nd spec will load the extended definition produced by 1st spec. 

@denis-bogdanas Are you sure? If you take a look at my PR you would see that the definition is saved in the temporary `.kprove` directory, so it would be unique for any spec.  of course, in order to use it with `kast` one would need to specify the `.kprove` directory as the directory to use for the definition, but that would work well for debugging sessions.

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -55,6 +62,19 @@ public int run() {
 
         Tuple2<Definition, Module> compiled = proofDefinitionBuilder
                 .build(kproveOptions.specFile(files), kproveOptions.defModule, kproveOptions.specModule);
+
+        // Saving compiled definition to disk to be usable by other tools (e.g., kast)
+        CompiledDefinition compiledDefinition = new CompiledDefinition(
+                new KompileOptions(), compiled._1(), compiled._1(), files, kem, null);
```
denis-bogdanas: KompileOptions should be taken from original CompiledDefinition.

denis-bogdanas (commented):
Sorry I missed that. If initial `compiled.bin` is not overwritten then the change is sound. But it should be disabled by default and activated by an additional command-line option. Because in most cases, for example on Jenkins, you don't need it and it will degrade performance.

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -55,6 +62,19 @@ public int run() {
...
+        Path proveKompiledDir = files.resolveTemp("prove-spec-kompiled").toPath();
+        try {
+            Files.createDirectory(proveKompiledDir);
+            loader.saveOrDie(proveKompiledDir.resolve("compiled.bin").toFile(), compiledDefinition);
+        } catch (IOException e) {
+            kem.registerCriticalWarning(
+                    "Could not create prove-spec-kompiled directory. Definition will not be saved.", e);
+        }
+
```
denis-bogdanas: this code should be separated to a method.

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -55,6 +62,19 @@ public int run() {
...
+        // Saving compiled definition to disk to be usable by other tools (e.g., kast)
+        CompiledDefinition compiledDefinition = new CompiledDefinition(
+                new KompileOptions(), compiled._1(), compiled._1(), files, kem, null);
+        Path proveKompiledDir = files.resolveTemp("prove-spec-kompiled").toPath();
+        try {
+            Files.createDirectory(proveKompiledDir);
+            loader.saveOrDie(proveKompiledDir.resolve("compiled.bin").toFile(), compiledDefinition);
+        } catch (IOException e) {
+            kem.registerCriticalWarning(
+                    "Could not create prove-spec-kompiled directory. Definition will not be saved.", e);
```
denis-bogdanas: It will be useful to log the directory, because I expect there will be multiple dirs there with similar names.

dwightguth: if this directory is not writeable it's probably better to raise an error instead of a warning.

traiansf:
@denis-bogdanas : I've added a new option `--save-proof-definition-to <dir>` option, which addresses your issue, while also allowing the user to save the definition in a given directory.

denis-bogdanas:
@traiansf What is the status of this? I didn't review so far the build is failing.

traiansf:
I believer that if I remove the test, this will pass. The functionality seems to work fine on my machine, so I guess it's due to some interaction with the testing machinery.  As this is supposed to be used for debugging, I think the fact that it interacts badly with integration testing should be less relevant.

I don't know how to fix it any other way, so I see the following possibilities:
1. Remove the test
2. Have somebody with more experience on OS and concurrent testing-related issues take over
4. Leave it as-is and people needing this particular type of debugging could use it.
3. Close the PR alltogether.

Any suggestions?

denis-bogdanas:
@traiansf Looks like a Makefile issue. Try `include ../../../include/kframework/ktest.mak`

traiansf:
> @traiansf Looks like a Makefile issue. Try `include ../../../include/kframework/ktest.mak`

That was an issue, but not the main one. Apparently there are some issues with creating directories; perhaps Jenkins would need elevated rights for this?

```
     [exec] kore-exec /home/jenkins-slave/workspace/k_PR-1138/k-distribution/tests/regression-new/issue-1137/.kprove-2020-06-25-11-59-59-180-77e25113-c48c-4433-9c8d-6aedf5796fbc/vdefinition.kore --module VERIFICATION --prove /home/jenkins-slave/workspace/k_PR-1138/k-distribution/tests/regression-new/issue-1137/.kprove-2020-06-25-11-59-59-180-77e25113-c48c-4433-9c8d-6aedf5796fbc/spec.kore --spec-module TEST-SPEC --output /home/jenkins-slave/workspace/k_PR-1138/k-distribution/tests/regression-new/issue-1137/.kprove-2020-06-25-11-59-59-180-77e25113-c48c-4433-9c8d-6aedf5796fbc/result.kore
     [exec] mv .kprove* .kprove
     [exec] /home/jenkins-slave/workspace/k_PR-1138/k-distribution/bin/kore-exec  .kprove/vdefinition.kore --module VERIFICATION --prove .kprove/spec.kore --spec-module TEST-SPEC | /home/jenkins-slave/workspace/k_PR-1138/k-distribution/bin/kast -i kore -o pretty -d .prove /dev/stdin | diff - test-spec.k.out
     [exec] 0a1,15
     [exec] >   #Not ( {
     [exec] >     X
     [exec] >   #Equals
     [exec] >     b
     [exec] >   } )
     [exec] > #And
     [exec] >   <k>
     [exec] >     X:S ~> DotVar1 ~> .
     [exec] >   </k>
     [exec] > #And
     [exec] >   {
     [exec] >     isA ( X )
     [exec] >   #Equals
     [exec] >     true
     [exec] >   }
     [exec] [Error] Critical: Does not exist or not a directory:
     [exec] /home/jenkins-slave/workspace/k_PR-1138/k-distribution/tests/regression-new/issue-1137/.prove
```

I'll remove the test.

denis-bogdanas:
@traiansf I really like this test. Might use it as model. Can you move it to a different PR? And we'll ping Dwight there.

traiansf:
> @traiansf I really like this test. Might use it as model. Can you move it to a different PR? And we'll ping Dwight there.

Thanks for the suggestion. I've opened PR #1374 .

traiansf:
@denis-bogdanas test pass now. Are you ok with it as-is?

denis-bogdanas (approved):

PR #1139, opened 2020-03-03
mark Set2List as functional

yzhang90:


dwightguth (commented):
Is this functional? Technically, it might actually be a relation.

ttuegel:
It is definitely defined, but if it's a relation then it's not even `function`. This is a common problem: we don't have a separate `total` attribute, just `functional` = `function` + `total`.

ttuegel (approved):

dwightguth (approved):

PR #1146, opened 2020-03-06
Changed default claim type from one-path to all-path

### LINKED ISSUE #1638: Default reachability claim type should be all-path
See also: #1615

traiansf:
Fixes https://github.com/kframework/kore/issues/1638


ttuegel (approved):
It looks like this causes the output of `max-broken-spec.k` to change trivially. That test is pretty unstable, it seems to break whenever you look at it. Anyway, it should be fine to update the output, but you should double-check.

traiansf:
> It looks like this causes the output of `max-broken-spec.k` to change trivially. That test is pretty unstable, it seems to break whenever you look at it. Anyway, it should be fine to update the output, but you should double-check.

Maybe we should figure out what is crucial that we want to test about max-broken-spec.k (making sure we don't just return true for everything?) and maybe write a custom test for it? @dwightguth @ttuegel what do you think?

PR #1149, opened 2020-03-08
haskell-backend: Release build by default

ttuegel:
Build a release build by default, but add a profile for a debug build. The
profile can be activated with `mvn -P haskell.backend.debug`.


dwightguth:
It would be nice if we unified the stuff so you didn't have to pass multiple flags for both haskell and llvm if you want to do a full debug build of k.

ttuegel:
@dwightguth I made the `project.build.type` property affect both backends.

dwightguth:
Since it affects both backends, can we make the profile ids not have "Haskell backend" in the name?

dwightguth (approved):

PR #1151, opened 2020-03-09
Specialize ==K to sort-specific equality functions

### LINKED ISSUE #1645: Beacon chain: g-spec.k
The haskell-backend kprove hangs for [g-spec](https://github.com/runtimeverification/beacon-chain-verification/blob/master/dynamic/tests/symbolic/g-spec.k).  The REPL hangs even before starting the REPL command prompt.

The thing is that the g-spec is quite similar to [f2-spec](https://github.com/runtimeverification/beacon-chain-verification/blob/master/dynamic/tests/symbolic/f2-spec.k) in terms of the spec structure and the complexity, but f2-spec is passed.

To reproduce:
```
$ git clone git@github.com:runtimeverification/beacon-chain-verification.git
$ git checkout bd82ad69a2a8a414209149bb63bd70d5938bba30
$ cd dynamic
$ make deps
$ make build-static
$ kprove -d .build/defn/static -m VERIFICATION tests/symbolic/g-spec.k
```

ttuegel:
Reverse the direction of three rules to reduce `==K` to `==Int`, `==String`, or
`==Bool` (as appropriate) instead of lifting those three functions to `==K`.

Fixes kframework/kore#1645


ehildenb:
This should be tested with KEVM before merging, I am testing it locally and will report here.

ehildenb:
The only error I get:

```
kore-exec: [2020-03-10 15:37:59.951506822] Error (ErrorException): Error: missing hook
Symbol
    Lbl'UndsEqlsEqls'String'UndsUnds'STRING-COMMON'Unds'Bool'Unds'String'Unds'String{}
declared with attribute
    hook{}("STRING.eq")
We don't recognize that hook and it was not given any rules.
Please open a feature request at
    https://github.com/kframework/kore/issues
and include the text of this message.
Workaround: Give rules for Lbl'UndsEqlsEqls'String'UndsUnds'STRING-COMMON'Unds'Bool'Unds'String'Unds'String{}
CallStack (from HasCallStack):
  error, called at src/Kore/Step/Simplification/Simplify.hs:439:6 in kore-0.0.1.0-6sgypCPNOjr3ZT67ZaZodF:Kore.Step.Simplification.Simplify
```

Looks like we don't have `String.eq` hook implemented in Haskell backend?

I don't think it's a reason to block this PR, but if it's easy to do then maybe we should wait.

ehildenb:
Otherwise all the proofs passed (except that one, which was `make tests/specs/functional/merkle-spec.k.prove`)

dwightguth (approved):
I'm going to approve so this isn't blocked on me, but please get @ehildenb 's approval before merging.

ehildenb (approved):
I'm approving assuming that https://github.com/kframework/kore/pull/1660/files gets merged and the update job goes through.

PR #1153, opened 2020-03-09
remove remaining instances of kil.Attribute and kil.Attributes classes

dwightguth:
Some long-awaited cleanup. I deferred the move of the static final string variables to another PR as that will be a very noisy change to the codebase and it will be much easier to understand this diff if it's done separately. But the next PR I will make after this is merged will delete the remaining class and move the constants over to Att.

@ttuegel FYI.


kore/src/main/java/org/kframework/builtin/Hooks.java
```diff
@@ -25,6 +25,7 @@
     public static final String STRING = "STRING";
     public static final String SUBSTITUTION = "SUBSTITUTION";
     public static final String UNIFICATION = "UNIFICATION";
+    public static final String JSON = "JSON";
 
-    public static final Set<String> namespaces = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(ARRAY, BOOL, BUFFER, BYTES, FFI, FLOAT, INT, IO, KEQUAL, KREFLECTION, LIST, MAP, MINT, SET, STRING, SUBSTITUTION, UNIFICATION)));
+    public static final Set<String> namespaces = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(ARRAY, BOOL, BUFFER, BYTES, FFI, FLOAT, INT, IO, KEQUAL, KREFLECTION, LIST, MAP, MINT, SET, STRING, SUBSTITUTION, UNIFICATION, JSON)));
```
ehildenb: This PR doesn't seem to add anything for JSON parsing handling, should we wait for the JSON module to add this?

ehildenb (approved):

kore/src/main/java/org/kframework/builtin/Hooks.java
dwightguth: Whoops, that's my mistake, it's not intended to be part of this pr. I will fix.

PR #1155, opened 2020-03-10
use smt-hook with expression instead of prelude function for abs, min,

### LINKED ISSUE #1133: kprove: no `int_max` in Z3
I tried to run `kprove` on the following:
```
module LEFTPAD-SPEC
    imports LAMBDA

    rule
        (leftpad (< N:Int , < C:String , Str:String > >)) 
        =>
        ?NewStr:String
    requires lengthString(C) ==Int 1
    ensures lengthString(?NewStr) ==Int maxInt(N, lengthString(Str))
endmodule
```

`LAMBDA` is the tutorial langauge with a couple of minor extensions (strings, pairs). I put it on pastebin [here](https://pastebin.com/z5rJz5GG), but I don't think it's actually relevant to the error. 

K gives me the following error:
```
[Error] Critical: Z3 crashed on input query:
(declare-fun andInt (Int Int) Int)
(declare-fun shlInt (Int Int) Int)
(declare-fun shrInt (Int Int) Int)
(declare-fun xorInt (Int Int) Int)
(declare-fun orInt (Int Int) Int)
(declare-fun notInt (Int) Int)
(declare-fun |N_131| () Int)
(declare-fun |_257| () Int)
(declare-fun |_258| () Int)
(assert (and
  (and
	(= (<= |N_131| _257) true)
	(= _258 1))
  (not (and
	(= _257 (int_max |N_131| _257))))
))
result:
(error "line 17 column 30: unknown function/constant int_max")
```

I assume this is because of the definition of `maxInt` in `domains.k`. I resolved my problem by using the `#if_#then_#else_#fi`, not sure if there's a better way.

Also, there's another issue with `lengthString` not having an SMT translation at all, not sure if that's a limitation or just something that just never got defined.

dwightguth:
max

@ttuegel @ehildenb do one of you know a set of proofs that uses min, max, and abs that I can test this on? or am I going to need to write manual tests myself?

Fixes #1133


ttuegel:
@dwightguth I'm not aware of any proofs; we don't have any in the Haskell backend's integration tests, certainly.

ehildenb:
Well you can use the proof from the linked issue. I think you can probably adjust the example to also make a test for `minInt`.

Otherwise, KWasm doesn't really use an `minInt` or `maxInt` in a significant way. KEVM has a few lemmas which use it based on taking ranges of integers from local memory (see `tests/specs/lemmas.k` and look for `minInt` and `maxInt`).

For `abs` you could have a simple proof like:

```
rule <k> X:Int *Int X:Int => abs(X:Int) ^Int 2 ... </k>
```

which would need to be discharged whether `X` is positive or negative. My guess is Z3 should be able to handle this, but no guarantees.

ehildenb (approved):

PR #1161, opened 2020-03-12
add in_keys function for cell maps

dwightguth:
Note that per my discussion with @ttuegel , I can't guarantee that this will work when written in rules, and we are not planning on supporting that case until we have an unambiguous need for it.


ttuegel (commented):
How hard is it to generate axioms for the generated symbol? After kframework/kore#1675 is merged, we will want axioms for each Map sort that reduce `#Ceil` problems to `in_keys` expressions.

dwightguth:
It shouldn't be that hard to generate extra axioms that make use of this symbol during the ModuleToKORE phase.

ttuegel (approved):

PR #1174, opened 2020-03-20
Concrete haskell

traiansf:
Added support for parameterized concrete and symbolic attributes.


ttuegel (commented):
Did you mean to update the LLVM backend?

traiansf:
> Did you mean to update the LLVM backend?

Hm... no; thanks for catching that!  using master's version now

ttuegel (approved):

PR #1185, opened 2020-03-25
Use FastBuild for K packaging

gtrepta:
This will change the K releases to use a project.build.type of FastBuild. This will fix the packaging for Firefly, which currently suffers from a bug in LTO of the llvm backend for Release builds of K.

I have tested and confirmed that this change fixes the Firefly package.


ehildenb (approved):
I'm approving this so we can get it merged and have a release out to Firefly by Monday.

I'm not sure if we want to change the default globally like this, or just change the Jenkinsfile to set `-Dproject.build.type=FastBuild`. But I can't think of any obvious benefits to either approach, so if @dwightguth has an opinion we can have a follow-up PR.

dwightguth:
I think it's reasonable to set this as the default, since it represents the shortest kompile times. Users can specify optimization flags when that behavior is appropriate.

PR #1188, opened 2020-03-26
k-distribution/substitution: move syntax for KVAR to KVAR-SYNTAX module

### LINKED ISSUE #1186: !X:KVar only works on Java backend
The freshGenerator function is part of KVAR-SYMBOLIC by mistake instead of being part of KVAR. This means that !X:KVar does not work on the llvm backend. Moving the sentences to the other module should be sufficient to fix the issue.

ehildenb:
Fixes: #1186

### DIFF
--- a/k-distribution/include/builtin/substitution.k
+++ b/k-distribution/include/builtin/substitution.k
@@ -11,22 +11,22 @@ endmodule
 
 module KVAR-SYNTAX
   syntax KVar [token, hook(KVAR.KVar)]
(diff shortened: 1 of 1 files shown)


ehildenb:
I think I should add a unit test, but I don't know if this is quite what you intended, @dwightguth, so please let me know.

dwightguth:
I'm going to close this and remake my own PR because that will be easier for me to get it the way I want it. I will add a test though.

PR #1200, opened 2020-04-03
In keys collection fix

### LINKED ISSUE #1199: `in_keys` for cell maps has unintuitive syntax
Currently, the syntax generated for a configuration like this:

```
    <map>
      <entry multiplicity="*" type="Map">
        <key> .K </key>
        <value> .K </value>
      </entry>
    </map>
```

Looks like this:

```
  syntax Bool ::= EntryCell "in_keys" "(" EntryCellMap ")" [function, functional, hook(MAP.in_keys)]
```

Where I would expect it to look like this:

```
  syntax Bool ::= KeyCell "in_keys" "(" EntryCellMap ")" [function, functional, hook(MAP.in_keys)]
```

ehildenb:
Fixes #1199 

This adds tests of the `in_keys` construct, but the tests are not passing right now because of #1199.

### DIFF
--- a/k-distribution/tests/regression-new/cell_map/Makefile
+++ b/k-distribution/tests/regression-new/cell_map/Makefile
@@ -1,6 +1,5 @@
 DEF=test
 EXT=test
 TESTDIR=.
-KOMPILE_BACKEND?=java
 
 include ../../../include/ktest.mak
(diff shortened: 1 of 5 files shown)


ehildenb:
Re-open if needed.

PR #1202, opened 2020-04-07
Cell maps ceil rules

traiansf:


k-distribution/include/builtin/domains.k
```diff
@@ -177,11 +178,20 @@ module MAP-SYMBOLIC [symbolic]
   // potential infinite loop
   // rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP [ K2 <- V2 ] [ K1 <- V1 ] requires K1 =/=K K2
 
-  rule K1 in_keys(M K2 |-> _) => true          requires K1  ==K K2 orBool K1 in_keys(M)
-  rule K1 in_keys(M K2 |-> _) => K1 in_keys(M) requires K1 =/=K K2
+  rule K in_keys(M K |-> _) => true
+  rule K in_keys(_) => false [owise]
+
+  rule K1 in_keys(M K2 |-> _) => true          requires K1  ==K K2 orBool K1 in_keys(M) [simplification]
+  rule K1 in_keys(M K2 |-> _) => K1 in_keys(M) requires K1 =/=K K2 [simplification]
```
ttuegel, line 205: Why are we retaining these rules for the Haskell backend?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -499,6 +498,98 @@ private void genUnitAxiom(Production prod, StringBuilder sb) {
...
+    private void genMapCeilAxioms(Production prod, Collection<Rule> rules) {
+        Sort mapSort = prod.nonterminal(1).sort();
+        scala.collection.Set<Production> mapProds = module.productionsForSort().apply(mapSort.head());
+        Production concatProd = mapProds.find(p -> hasHookValue(p.att(), "MAP.concat")).get();
+        Production elementProd = mapProds.find(p -> hasHookValue(p.att(), "MAP.element")).get();
+        List<K> args = new ArrayList<>();
+        List<K> setArgs = new ArrayList<>();
+        Seq<NonTerminal> nonterminals = elementProd.nonterminals();
+        K restMap = KVariable("Rest", Att.empty().add(Sort.class, mapSort));
+        K restMapSet = KVariable("@Rest", Att.empty().add(Sort.class, mapSort));
```
ttuegel: What's going on with the pairs of set and element variables?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -499,6 +498,98 @@ private void genUnitAxiom(Production prod, StringBuilder sb) {
...
+        Sort sortParam = Sort(AddSortInjections.SORTPARAM_NAME, Sort("Q"));
+        KLabel ceilMapLabel = KLabel("#Ceil", mapSort, sortParam);
```
ttuegel: Should the `#Ceil` and `#And` constants be defined somewhere?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -499,6 +498,98 @@ private void genUnitAxiom(Production prod, StringBuilder sb) {
...
+                                                        restMapSet
+                                                ),
+                                                BooleanUtils.FALSE
+                                        ),
+                                        setArgsCeil
+                                )
+                        )
+                        , BooleanUtils.TRUE
+                        , BooleanUtils.TRUE
+                        , Att.empty().add("anywhere").add("simplification")
```
ttuegel, line 598: What is the `anywhere` for?

k-distribution/include/builtin/domains.k
traiansf: Well, I figured that since the Java engine still needs to work, we must keep these around.
One way to exclude them from the Haskell backend all-together would be to put them into a module which has attributes `symbolic` and `kast`, which would make it acceptable to the Java backend but excluded from the Haskell backend.  I'll try and do that.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: the element variables are used for the in_keys rules. The set variables are used for the ceil rules.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: That's awkward, but since #Ceil is not a function, we usually need to annotate `#Ceil` rules with `anywhere` to prevent them from being completed with `<k>` and stuff. Maybe here I didn't need to add them, as we're late in the compilation process, but I added them for consistency.

k-distribution/include/builtin/domains.k
traiansf: I've completely separated the existing symbolic part from the Haskell part. We can decide later whether we want to share some of the rules or not.

dwightguth (approved):

k-distribution/include/builtin/domains.k
```diff
@@ -158,7 +158,32 @@ module MAP
...
+    {(@K in_keys(@M)) #Equals false} #And #Ceil(@M) #And #Ceil(@K) #And #Ceil(@V)
+  [anywhere, simplification]
+
+   rule
+     #Ceil(@M:Map [@K:KItem])
+   =>
+     {(@K in_keys(@M)) #Equals true} #And #Ceil(@M) #And #Ceil(@K)
+   [anywhere, simplification]
+
+  rule
```
ttuegel, line 177: Why isn't `_[_ <- _]` simply `functional`?

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ttuegel: I find it difficult to follow how the rules are constructed because the construction of the rules is interleaved in this way, but it's not difficult enough that I'm going to hold up this pull request.

ttuegel (approved):

k-distribution/include/builtin/domains.k
traiansf: You're right. `_[_ <- _]` is already functional. the rule is not necessary.

traiansf:
@dwightguth : this is the error I'm getting from `llvm-kompile`:
```
/home/traian/Documents/k/k-distribution/lib/../bin/llvm-kompile: line 27: 29891 Segmentation fault      (core dumped) "$(dirname "$0")"/llvm-kompile-codegen "$definition" "$dt_dir"/dt.yaml "$dt_dir" $debug > "$mod"
```

I tried to run `llvm-kompile-codegen` stand-alone, but I get the same Segmentation fault error, and I don't know where to find the core dump.
```
/home/traian/Documents/k/k-distribution/bin/llvm-kompile-codegen imp-kompiled/definition.kore imp-kompiled/dt/dt.yaml imp-kompiled/dt 1
Segmentation fault (core dumped)
```

traiansf:
Given the above, I'm going to merge this as-is (once the tests pass), but if you can take a look and help me tackle the llvm backend issue, I'd like to remove the haskell backend guard in a subsequent PR.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
traiansf: I've refactored the code and added comments with the rules. hope it is clearer now.

PR #1211, opened 2020-04-10
LLVM backend unparser changes in K frontend

dwightguth:
Mostly this consists of generating some flat text files that are used to avoid having to parse the entire K definition in order to do unparsing. we also add the integration test for the functionality to the K tutorial, and include some helper code that was used to generate ANSI terminal color codes to be fed into the llvm backend.

Second PR is here: https://github.com/kframework/llvm-backend/pull/315


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -350,7 +381,47 @@ private void translateSymbols(Map<String, Boolean> attributes, SetMultimap<KLabe
...
+            if (koreAtt.contains("color")) {
+              colors.append("Lbl");
+              convert(prod.klabel().get().name(), colors);
+              colors.append('\n');
+              for (int i = 0; i < prod.items().size(); i++) {
+                if (prod.items().apply(i) instanceof Terminal) {
+                  colors.append(koreAtt.get("color"));
+                  colors.append('\n');
+                }
+              }
```
ehildenb: This isn't really about symbol translation directly, and may confuse future developers about where the logic for symbol translation ends and writing the extra files begins. You could consider separating out the new functionality into a function `writeSymbolUnparsingInformation` (or something similar) and just call that method here, to make the distinction clearer.

Your call.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1357,6 +1428,23 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
...
+          } else if (prod.items().apply(i) instanceof Terminal) {
+            format = format.replace("%" + (i+1), "%c" + ((Terminal)prod.items().apply(i)).value() + "%r");
+          } else {
+            hasFormat = false;
+          }
+        }
+        if (hasFormat) {
+          att = att.add("format", format);
+        } else {
+          att = att.remove("format");
```
ehildenb, line 1468: Similarly here.

ehildenb (changes requested):
Please add documentation for `format`, `color`, `colors`, `assoc`, and `comm` attributes to either an `Unparsing` section or the `Syntax Declaration` section of `pending-documentation.md`. Otherwise looks good to me. Yay for the performance boost!

ehildenb (approved):
I don't see documentation for `assoc` and `comm`, but it can wait for another time, because this is already a huge help for the unparsing documentation.

PR #1216, opened 2020-04-17
add new hook names for crypto hooks

dwightguth:
@ttuegel @sskeirik FYI

This is done in order to separate the crypto hooks out based on which native library they depend on in the llvm backend, in order to simplify the build process for new semantics. The old hooks will remain valid for the time being for backwards-compatibility.


ehildenb:
Have these hooks been added to the LLVM backend? Are we planning on incorporating the hooks from blockchain-k-plugin directly into the LLVM backend?

dwightguth:
No, these hooks are not part of the llvm backend because that would involve making developers install obscure Bitcoin and cryptography libraries in order to build the k frontend. It's not an issue with java and haskell because their package managers pull in the right libraries automatically.

ehildenb:
Ah sorry, I misread, forgot that this is only for the Java backend. Approving.

ehildenb (approved):

ehildenb:
Though I guess someone who needs this should approve as well.

sskeirik:
EDIT: TL;DR version --- I vote we change `HASH.sha3` hook to `HASH.sha3_256`.

Since we are changing hook names anyway, I propose the LLVM backend migrates from the name `sha3` to `sha3256` like the Haskell and Java backends or else something like `sha3_256`. I didn't think about changing that before since it wasn't a hook that I was adding, but I do think it makes sense to keep the digest size in bits after each hash function, for consistency of naming and also clarity. Thus, under this naming scheme we have either:

hash function name + digest size

hash function name + (separator when hash name ends with number) + digest size

dwightguth:
As far as I can tell, the build failed because github was experiencing problems. I restarted it.

PR #1217, opened 2020-04-17
performance improvements to prover frontend

dwightguth:


ehildenb:
@dwightguth please provide some high-level explanation here of what changes are made to improve performance. It's much easier to follow the code if I know the high-level reasoning ahead of time.

dwightguth:
Essentially, we were doing a bunch of work involving creating scanners that we didn't actually need to be doing. Part of the problem was that ResolveConfig had been made into a separate class in a manner that caused us to be calling certain expensive methods multiple times when we only needed to be calling them once, and part of the problem was that we were creating multiple scanners when we only needed to create one. Another problem was that we were parsing modules that were not referenced by the current spec, but were in the same file as a file that was required by the spec.

Thus, roughly speaking, it includes the following changes:
* We correctly compute which modules are entry point modules (ie the definition module and specification module) and don't parse bubbles in modules that are not referenced by these modules. 
* The ResolveConfig class was deleted and its logic moved into a resolveConfigBubbles sequence of functions within DefinitionParsing, allowing us to avoid duplicating calls to loadCache and getCombinedGrammar
* We move the construction of the scanner out of the parallel portion of the code, meaning it only runs once rather than once per thread.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -111,23 +113,36 @@ public DefinitionParsing(
...
+        if (!def.getModule(entryPointModule).isDefined()) {
+          throw KEMException.criticalError("Module " + entryPointModule + " does not exist.");
+        }
+        Stream<Module> modules = Stream.of(def.getModule(mainModule).get());
+        modules = Stream.concat(modules, stream(def.getModule(mainModule).get().importedModules()));
+        modules = Stream.concat(modules, Stream.of(def.getModule(entryPointModule).get()));
+        modules = Stream.concat(modules, stream(def.getModule(entryPointModule).get().importedModules()));
+        modules = Stream.concat(modules,
+                stream(def.entryModules()).filter(m -> !stream(m.sentences()).anyMatch(s -> s instanceof Bubble)));
+        def = Definition(def.mainModule(), modules.collect(Collections.toSet()), def.att());
```
ehildenb, line 128: I don't think I fully understand this, so want to make sure I do.

We're grabbing the entryPoint and main modules, as well as any modules they transitively import. We're also grabbing any entryModule (which I guess would be something like the proving module?), but _not_ the transitive imports of the entryModules? And we're specifically only grabbing entryModules which have bubbles which need parsing.

I guess my questions are:

-   Why not get transitive imports of entryModules()?
-   Why not apply the filter for only modules with bubbles across _all_ the modules, not just the entryModules()?

dwightguth: so, Definition.entryModules is a bit of a misnomer insofar as it contains every module that was included in the K files you parsed. that's why we're not doing a transitive closure there. And specifically we are only grabbing those modules which /don't/ contain bubbles. And the only reason we are doing this is because some of them are required to exist in the parser, like some of the ones in kast.k.

ehildenb: Ok, so it's an over-approximation of the modules needed to do parsing.

ehildenb (approved):

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
denis-bogdanas: 1. I recommend a comment explaining the above logic. Lack of comments in such places is a reason why code is hard to maintain. 
2. I don't see any advantage to build a Stream then to convert it to Set. You should create a Set from start. Code will look simpler. If you want the preserve the iterating order then use `LinkedHashSet`.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -247,15 +260,90 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+            } else {
+                throw KEMException.compilerError("Module Map must be visible at the configuration declaration, in module " + module.name());
+            }
+            return Module(module.name(), (Set<Module>) module.imports().$bar(Set(mapModule)),
+                    (Set<Sentence>) module.localSentences().$bar(configDeclRules),
+                    module.att());
+        }
+
+    }
+
```
denis-bogdanas, line 346: Please add comments where you see fit.

PR #1218, opened 2020-04-18
java-backend: fix SMT translation of [smt-lemma] rules with requires

daejunpark:
This fixes a (known) bug in the SMTLIB translation of [smt-lemma] rules.

Previously, for a [smt-lemma] rule with a non-empty requires clause, e.g.,
```
rule f(X) => g(X) requires X >Int 0 [smt-lemma]
```
the require clause was omitted in the generated SMTLIB formula:
```
(forall ((x Int)) (= (f x) (g x)))
```

Now, it generates a correct one:
```
(forall ((x Int)) (=> (> x 0) (= (f x) (g x))))
```


yzhang90 (approved):
LGTM

PR #1225, opened 2020-04-22
Simplify handling #Ceil(_Map_)

ttuegel:
This pull request implements two related changes which simplify how `#Ceil(_Map_)` is handled:

1. The generated axioms for the `MAP.in_keys` hook are removed. These interact poorly with the `#Ceil(_Map_)` axioms; it's hard to apply them without looping. Because this requires some meta-logic, we will implement these hooks fully symbolically in the backend.
1. Some redundant conditions are removed from the generated `#Ceil(_Map_)` axioms.


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -598,69 +598,33 @@ private void genMapCeilAxioms(Production prod, Collection<Rule> rules) {
...
-                BooleanUtils.TRUE,
-                Att.empty().add("owise")
-        );
-        rules.add(inKeysRule2);
+        // rule
+        //   #Ceil(MapItem(K1, K2, ..., Kn) Rest:Map)
+        // =>
+        //  {(@K1 in_keys(@Rest)) #Equals false} #And ... #And #Ceil(@Kn)
+        // Note: The {_ in_Keys(_) #Equals false} condition implies
+        // #Ceil(@K1) and #Ceil(@Rest).
```
traiansf: I agree with you that the `in_keys` condition would require the variables to be defined anyway.
Why then keep the definedness conditions?

traiansf (approved):

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ttuegel: Only the definedness conditions `#Ceil(@K2) #And ... #And #Ceil(@Kn)` are present; I will update the comment to clarify.

PR #1227, opened 2020-04-22
Packaging subdirectory [brew-staging: directory-change]

ehildenb:
Blocked on: #1231 

This PR applies several cleanups to the K repo structure.

-   Creates `package` subdirectory where packaging scripts are moved (currently except Homebrew scripts).
-   Moves `Dockerfile.*` into the appropriate packaging subdirectory, and has just one generic Dockerfile at the top level for normal CI steps.
-   Makes sure that when there is a `brew-staging: ` directive in the PR title, that the branch is merged into `master` for the `homebrew-k` repository before continuing.


ehildenb:
Needs kframework/homebrew-k#3 merged before this is merged.

ehildenb:
@dwightguth it really does not make sense to me that the K Platform Independent Binary is still failing on this PR. I'm not changing the Dockerfile it uses, or anything about any of the tests it's running. Everything that is changed has to do with the other packages, and the only thing changed is that scripts are moved to the `package` subdirectory to organize the repo a bit.

Any ideas what's going on?

ehildenb:
Evidence that this is working: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1227/29/pipeline/

ehildenb:
@dwightguth for some reason, when the Arch build was running before the Platform Independent K Binary build _in the same directory_, it was causing test18 to fail with a unicode error. fed0f80 runs the arch build in a subdirectory, which seems to fix the problem. This is ready for review.

Jenkinsfile
```diff
@@ -302,23 +309,24 @@ pipeline {
                         git show-ref --verify refs/remotes/origin/$brew_base_branch
                         git push -d origin brew-release-$PACKAGE || true
                         git checkout -b brew-release-$PACKAGE "origin/$brew_base_branch"
-                        ${WORKSPACE}/src/main/scripts/brew-update-to-local
+                        git merge origin/master
```
dwightguth, line 312: Why is this here? Hasn't it already done a fresh clone of master?

ehildenb: If there is not a `brew-staging: ` tag in the PR title, then it will do nothing (because `brew-release-*` branch will already be based on `origin/master`).

If there is a `brew-staging: ` tag, then it will make sure it's up to date. This is important because we don't always know the order that PRs will get merged in (so another K job may go through which updates `origin/master` in the `homebrew-k` repo), so we want to make sure that if the brew-staging branch is set, it merges cleanly into master before applying the later changes.

dwightguth (approved):

PR #1228, opened 2020-04-22
Dockerhub images

ehildenb:
This will push a built Ubuntu Bionic docker image with the K package installed on it to DockerHub under `runtimeverificationinc/kframework-k:ubuntu-bionic-COMMIT_ID`.

The naming scheme is `runtimeverificationinc/REPO_IDENTIFIER:DISTRO_IDENTIFIER-COMMIT_ID`, where `REPO_IDENTIFIER` is the full GitHub path to the repo with `/` replaced by `-`.

This will make it so that repos which depend on a K submodule can just grab the corresponding Docker image by inspecting the submodule's K commit ID. That way, we get to:

-   Still test downstream repos (eg KEVM) with custom versions of K to be sure our changes to K will fix our KEVM problems.
-   Use a pre-built docker image with K installed on CI jobs to shave off 8-12 minutes every semantics job we have (the `make deps` steps, usually).


ehildenb:
I have the changes to make it so that we also do Debian Buster and Arch docker images, but I don't think it's necessary. The primary target here is to make CI faster so adding different distros as examples seems lower priority.

ehildenb:
Here is an example of it working: https://github.com/kframework/wasm-semantics/pull/325

ehildenb:
One thing I'm not sure about is the tagging scheme. I like having the commit-id in the tag name so that we have the ability to peg to specific versions of K, and there is no limit on the number of tags in a given public Docker Hub repo.

But it may also be nice to provide a `master` tag for ease of use.

ehildenb:
It may be easier to review this one with `git diff origin/master origin/dockerhub-images -b`

dwightguth:
Having both specific and rolling tags is fairly normal for docker, so I +1 that idea.

ehildenb:
@dwightguth this should also update the `ubuntu-bionic-master` tag each time master branch changes too.

I also publish the docker image _before_ the github release, so that the release notes will point to a valid Docker Hub image.

Dockerfile
```diff
@@ -1,12 +1,16 @@
 FROM ubuntu:bionic
 
+ARG USER_ID=1000
```
dwightguth: please move this back where it was, immediately before its use. moving it to the top reduces caching in the dockerfile.

Jenkinsfile
```diff
@@ -364,15 +364,20 @@ pipeline {
       }
     }
     stage('Deploy') {
+      when { branch 'master' }
```
dwightguth: why did you remove the beforeAgent line?

Jenkinsfile
```diff
@@ -364,15 +364,20 @@ pipeline {
...
       agent {
         dockerfile {
           additionalBuildArgs '--build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g)'
+          args '-v /var/run/docker.sock:/var/run/docker.sock'
```
dwightguth: what is this doing?

Jenkinsfile
```diff
@@ -381,54 +386,70 @@ pipeline {
...
+        stage('DockerHub Images') {
+          environment { DOCKERHUB_TOKEN = credentials('dockerhub-ehildenb') }
+          steps {
+            dir('bionic') { unstash 'bionic' }
             sh '''
-              git checkout master
-              git merge brew-release-$PACKAGE
-              git push origin master
-              git push origin -d brew-release-$PACKAGE
+                sudo docker login --username ehildenb --password "${DOCKERHUB_TOKEN}"
```
dwightguth: is this storing your password on Jenkins? we probably should not use your user account.

Jenkinsfile
```diff
@@ -381,54 +386,70 @@ pipeline {
...
+
+                bionic_commit_tag="ubuntu-bionic-${SHORT_REV}"
+                kframework_k_docker_repo="runtimeverificationinc/kframework-k"
+                mv bionic/kframework_${VERSION}_amd64.deb kframework_amd64_bionic.deb
+                sudo docker image build . --file package/docker/Dockerfile.ubuntu-bionic --tag "${kframework_k_docker_repo}:${bionic_commit_tag}"
```
dwightguth: running docker as root should not be necessary if you use an unprivileged container.

package/docker/Dockerfile.ubuntu-bionic
```diff
@@ -0,0 +1,36 @@
+FROM ubuntu:bionic
+
+ARG USER_ID=1000
```
dwightguth: the same problem with ARG exists here.

dwightguth (changes requested):
Overall this looks okay ,but I have a bunch of minor concerns.

Dockerfile
ehildenb: I don't think this will reduce any cacheing, actually may improve it since these lines are more stable than the next lines (installing a bunch of stuff).

I mostly just like having all the arguments at the top for scanability. I can move it back if you really want, but don't see the need.

Jenkinsfile
ehildenb: This allows this docker container to speak to the LXC container docker daemon directly. Launching a docker daemon inside a docker container is strongly discouraged (see https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/), instead you are supposed to launch it using the hosts docker daemon, which this enables.

Note this is better for cacheing the images that are build inside this container too.

package/docker/Dockerfile.ubuntu-bionic
ehildenb: Same response. I like arguments at the top of the file, and don't think this will reduce cacheing.

Jenkinsfile
ehildenb: I put it back.

Jenkinsfile
ehildenb: I added the user to the `docker` group, and removed use of `sudo`. Need to test that it works.

ehildenb: It appears not to work. I think the issue is that the docker group inside the container does not map to the same gid as the docker group on the lxc host. See this thread for a similar issue seen with Azure containers https://github.com/microsoft/azure-pipelines-agent/issues/2056.

I don't think it's really worth it investigating how to get the group maps to work, since I don't think using `sudo` in the main CI file is that much of a problem.

ehildenb:
It worked, ready for re-review @dwightguth https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1228/36/pipeline/

Jenkinsfile
ehildenb: Switched to rvdockerhub.

Dockerfile
dwightguth: Everything after the ARG line depends on the value of the argument, so if the user id changes, we will have to reinstall all the debian packages and stack.

Jenkinsfile
dwightguth: Doesn't running docker with sudo mean that the root user inside the container has root access outside the container? Isn't that a security problem?

Dockerfile
ehildenb: Ok, I moved it back.

package/docker/Dockerfile.ubuntu-bionic
ehildenb: Moved it back.

ehildenb:
Evidence that it works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1228/49/pipeline/152

So now I am just running the `docker` commands directly on the LXC container, which at the very least is much simpler than all other solutions and means that the main CI docker container doesn't need `docker` or `sudo` privileges.

@dwightguth ready for re-review.

dwightguth (approved):

PR #1229, opened 2020-04-22
Better debugging support

dwightguth:
This includes the K frontend changes necessary to use and test the llvm backend debugging support implemented in kframework/llvm-backend#318.

I would like @Andrew-Miranti and @nishantjr to test this out and let me know  if they encounter problems during real-world use.

To summarize, the commands are:

* `k start` - Starts executing a program but does not take any steps, stopping at the first call to 
the step function
* ` k step [N]` - takes N top-level rewrite steps (by default 1)
* `k match MODULE.label subject` tells you why a particular rule does or doesn't apply when stopped at a breakpoint after running `k step`.


sskeirik:
Some usability considerations:

1. If you accidentally run `k step` before `k start`, you get:

```
(gdb) k step
Python Exception <class 'gdb.error'> The program is not being run.:
Error occurred in Python: The program is not being run.
```

2. If you run `k match` on a symbol that isn't defined, you get this:

```
(gdb) k match MICHELSON.#DoHash subject
Traceback (most recent call last):
  File "/home/stephen/work/michelson-semantics/ext/k/llvm-backend/target/build/install/lib/kllvm/gdb/kllvm.py", line 739, in invoke
    gdb.lookup_global_symbol(argv[0] + '.match').value()(gdb.parse_and_eval(argv[1]))
AttributeError: 'NoneType' object has no attribute 'value'

Python Exception <class 'AttributeError'> 'NoneType' object has no attribute 'value':
Error occurred in Python: 'NoneType' object has no attribute 'value'
```

Also, how do I find the label for a rule?

EDIT:

3. I get inconsistent results when I use `k step N`. If I run `k step` 11 times, my example program terminates. However, if I run `k step N` with a number larger than 11, it doesn't terminate. For some reason, it doesn't always take the requested number of steps.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 export PATH=$PATH:"$(dirname "$0")"/../../../../bin
-llvm-krun -d "$(dirname "$0")/imp-kompiled" -c PGM "$1" Pgm prettyfile -p
+pgm=$1
+shift
+llvm-krun -c PGM "$pgm" Pgm prettyfile -p "$@" -d "$(dirname "$0")/imp-kompiled"
```
ehildenb, line 5: This is a little confusing to me as an example, because it looks like `$@` is maybe all arguments to `-p`? Or no?

If not, can you put it at the end in a position where it's less confusing as an example?

```suggestion
llvm-krun -c PGM "$pgm" Pgm prettyfile -p -d "$(dirname "$0")/imp-kompiled" "$@"
```

ehildenb (approved):

k-distribution/tutorial/1_k/2_imp/lesson_5/imp
dwightguth: I would but then if you pass -d to `imp` as the K makefile does, the script will break.

ehildenb (approved):
@nishantjr and @sskeirik should still chime in on if this is usable for them.

But I don't think the issue that @sskeirik already pointed out needs to be fixed now, since if it's documented correctly and people know how to use it for now it's better than it was. We can iron out issues like that in a future PR.

dwightguth:
I spoke with Stephen and he was unable to reproduce the issue, otherwise I would have fixed it.

PR #1236, opened 2020-04-26
Jenkinsfile: use RELEASE_TAG job instead of RELEASE_URL

ehildenb:
This changes from the RELEASE_URL dependency updater for Kore to the RELEASE_TAG dependency updater, which is much simpler and more flexible.

Instead of committing the entire URL to the file `deps/k_release`, it will _only_ commit the tag (the end of the URL). This allows the same job to be used for repos that are not kframework/k.

When it makes the next commit to the `_update-deps_deps_k_release` job on the Haskell backend repo, it needs the logic adjusted, which I have done here: https://github.com/kframework/kore/pull/1764.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -394,6 +394,10 @@ pipeline {
         dir('mojave') { unstash 'mojave' }
         sshagent(['2b3d8d6b-0855-4b59-864a-6b3ddf9c9d1a']) {
           sh '''
+            git remote add release 'ssh://github.com/kframework/k.git'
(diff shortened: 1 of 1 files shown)


ehildenb:
I've also added the logic which makes sure that the release tag exists ahead of time, so that if people want to test the github release on a version _not_ on master, it does the "sensible" thing. Previously it would always make the release peg to whatever current `master` is, even if the release was happening on a commit on a PR branch. But now it will actually peg it to the commit that the release is made on.

ttuegel (approved):

PR #1238, opened 2020-04-26
Simpler Dockerfile, per-distro packaging Dockerfiles

ehildenb:
This:

-  Moves the dockerfiles for each distro into a subdirectory of `package`.
-  Simplifies the main Dockerfile for CI container (removes the parametric bits).
-  Uses the Dockerfile to setup the git credentials rather than throughout the Jenkinsfile.
-  Makes sure that `hub` is installed in the main dockerfile so we can use it directly for doing GitHub releases.
-  Switches the smoke-test `test-in-container` to use llvm backend, _not_ ocaml backend.

This is pulled out of #1227 , because that one is failing and I figured we try getting it in one change at a time.


ehildenb:
This passed the tests with everything except the actual deploy and MacOS packaging enabled https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1238/1/pipeline/225/.

src/main/scripts/test-in-container
```diff
@@ -9,5 +9,5 @@ echo 'Testing tutorial in user environment...'
 make -j`nproc`
 cd ~
 echo "module TEST imports BOOL endmodule" > test.k
-kompile test.k --backend ocaml
+kompile test.k --backend llvm
```
dwightguth: The reason we are building with the ocaml backend here is because we have already tested the other backends. Changing this means we will not get informed if something relating to the ocaml backend breaks, but it doesn't actually add any new test coverage. Please undo this change.

dwightguth (changes requested):

src/main/scripts/test-in-container
ehildenb:
```suggestion
kompile test.k --backend ocaml
```

ehildenb: Done @dwightguth

dwightguth (approved):

PR #1239, opened 2020-04-26
Kool update

nishantjr:
These are some fixes I had made. It doesn't work on the Hakell backend (See issue #1721) and I've not tested the LLVM backend. But since I've done the work, these can be merged.


ehildenb:
Re-open if needed.

PR #1241, opened 2020-04-28
Jenkinsfile: add quietPeriod for all triggered builds

ehildenb:
When a K PR gets merged, a lot of builds are triggered. These almost always completely saturate our CI machines. This should stagger them a bit so that their updates happen a little less intensely.

The quietPeriod for this job is set to 120s. I would like to set it to 1200s, but then they will never get scheduled because the build-killer runs ever 240s. So I think setting it to 120s should have them staggering by roughly 240s at a time, which will at least let some of the faster semantics test-suites get most of the way through before the slower ones start.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,31 +37,31 @@ pipeline {
     stage('Update Submodules (non-release)') {
       when { branch 'master' }
       steps {
-        build job: 'rv-devops/master', propagate: false, wait: false                                   \
+        build job: 'rv-devops/master', propagate: false, wait: false, quietPeriod: 120                 \
             , parameters: [ booleanParam(name: 'UPDATE_DEPS_SUBMODULE', value: true)                   \
                           , string(name: 'PR_REVIEWER', value: 'ehildenb')                             \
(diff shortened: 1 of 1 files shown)


dwightguth:
I don't understand. why would the build killer be killing pending builds for which no newer build exists? That sounds dangerous to me.

ehildenb:
The build killer runs every 4 minutes, so that maximum quiet period this triggered job ever sees is 4 minutes. So setting it to a quiet period of 20 minutes would mean that these jobs never get scheduled.

Note that the jobs being scheduled here are `rv-devops/master` jobs, which is the same job which runs the build killer.

The build killer will _not_ kill builds that no newer build exist for. The build kill will also not ever kill an `rv-devops/master` job.

ehildenb:
Staggering even by 2 minutes minimum here (more realistically by 4 minutes in most cases, and up to 6 minutes in others) will already be a big improvement.

ehildenb:
Hmmm, actually, it would appear that builds with different parameters on the same job are given different quiet periods: https://github.com/eclipse/codewind/issues/2408

So this will be useless anyway.

PR #1246, opened 2020-04-30
Codeowners on Jenkinsfile

ehildenb:


CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 /k-distribution/include/builtin/* @dwightguth
 /pending-documentation.md @dwightguth
+/Jenkinsfile @kframework/admin
```
ehildenb, line 3: @dwightguth I don't know if this will work, because we need the rv/admin group don't we?

dwightguth:
@kframework/admin is the correct way to @ the team. The corresponding runtimeverification team is @runtimeverification/admin.

dwightguth (approved):

PR #1247, opened 2020-05-01
GLR Bison parser

dwightguth:
This contains a number of improvements to the K bison parser generator, including the ability to generate GLR parsers powerful enough to parse C.

### DIFF
--- a/k-distribution/tests/regression-new/parse-c/c18-syntax.k
+++ b/k-distribution/tests/regression-new/parse-c/c18-syntax.k
@@ -23,7 +23,7 @@ endmodule
 
 module C18-COMMON
   imports C-LEXICAL-SORTS
-  
+
   // \section(A.2.1)
 
   syntax Exp ::= Identifier
@@ -108,7 +108,7 @@ module C18-COMMON
                 | Exps "," Exp [assignExp]
 
   // \section(A.2.2)
-  
+
   syntax Decl ::= DeclSpecifiers InitDecls ";"
                 | DeclSpecifiers ";"
                 | StaticAssertDecl
@@ -272,7 +272,7 @@ module C18-COMMON
   syntax StaticAssertDecl ::= "_Static_assert" "(" Exp "," StringLiteral ")" ";"
 
   // \section(A.2.3)
-  
+
   syntax Stmt ::= Identifier ":" Stmt
                 | "case" Exp ":" Stmt
                 | "default" ":" Stmt
@@ -319,7 +319,7 @@ module C18-COMMON
                    | DeclSpecifiers FunctionDeclarator CompoundStmt [format(%1 %2 %3)]
 
   syntax Decls ::= NeList{Decl,""}
-  
+
 endmodule
 
 module C18-BASE
--- a/k-distribution/tests/regression-new/parse-c/c18-syntax.k
+++ b/k-distribution/tests/regression-new/parse-c/c18-syntax.k
@@ -1,0 +1,340 @@
+module C-LEXICAL-SORTS
+  syntax Identifier
+  syntax Constant ::= Char
+  syntax StringLiteral
+  syntax Char
+endmodule
+
+module C-LEXICAL
+  imports C-LEXICAL-SORTS
+
+  imports ID-SYNTAX
+  imports FLOAT-SYNTAX
+  imports STRING-SYNTAX
+
+  syntax Identifier ::= Id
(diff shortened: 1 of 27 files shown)


k-distribution/tests/regression-new/parse-c/c18-syntax.k
```diff
@@ -0,0 +1,340 @@
...
+                | "for" "("     ";" Exp ";" Exp ")" Stmt
+                | "for" "("     ";"     ";" Exp ")" Stmt
+                | "for" "(" Exp ";"     ";"     ")" Stmt
+                | "for" "(" Exp ";" Exp ";"     ")" Stmt
+                | "for" "(" Exp ";" Exp ";" Exp ")" Stmt
+                | "for" "(" Exp ";"     ";" Exp ")" Stmt
+                | "for" "(" Decl     ";"     ")" Stmt
+                | "for" "(" Decl Exp ";"     ")" Stmt
+                | "for" "(" Decl Exp ";" Exp ")" Stmt
+                | "for" "(" Decl     ";" Exp ")" Stmt
```
radumereuta, line 298: The for doesn't look very nice. What happened?

kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java
```diff
@@ -201,8 +207,8 @@ private void writeAction(StringBuilder flex, TerminalLike key) {
     private void writeStandaloneAction(StringBuilder flex, TerminalLike key) {
         flex.append(" {\n" +
             "  int kind = ").append(tokens.get(key)._1()+1).append(";\n" +
-            "  *((char **)&yylval) = malloc(strlen(yytext) + 1);\n" +
-            "  strcpy(*((char **)&yylval), yytext);\n" +
+            "  *((char **)yylval  ) = malloc(strlen(yytext) + 1);\n" +
```
radumereuta: Why the extra space? :)

radumereuta (approved):
I had a quick look at the grammar and looks mostly fine.
I'm not that familiar with Bison, maybe Everet should also have a look before merging.

k-distribution/tests/regression-new/parse-c/c18-syntax.k
dwightguth: Combinatory explosion. The reference grammar has three optional norterminals, so we generate 2^3=8 productions. The alternative would be to introduce a new nonterminal that has an empty production, but bison dislikes this immensely and you end up with a parser that splits a lot more often.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -135,6 +137,9 @@ public boolean isKore() {
         @Parameter(names="--gen-bison-parser", description="Emit bison parser for the PGM configuration variable within the syntax module of your definition into the kompiled definition.")
         public boolean genBisonParser;
 
+        @Parameter(names="--gen-glr-bison-parser", description="Emit GLR bison parser for the PGM configuration variable within the syntax module of your definition into the kompiled definition.")
+        public boolean genGlrBisonParser;
```
ehildenb, line 141: We should emit a parser for each `$` variable that shows up in the configuration, not just `$PGM`.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -300,9 +303,15 @@ public void applyUserList(Set<org.kframework.definition.Sentence> res,
         org.kframework.definition.Production prod1, prod3;
 
         // Es ::= E "," Es
-        prod1 = Production(KLabel(p.getKLabel(kore), immutable(p.getParams())), sort,
-                Seq(NonTerminal(elementSort), Terminal(userList.getSeparator()), NonTerminal(sort)),
-                attrs.add("right"));
+        if (bisonLists) {
+          prod1 = Production(KLabel(p.getKLabel(kore), immutable(p.getParams())), sort,
+                  Seq(NonTerminal(sort), Terminal(userList.getSeparator()), NonTerminal(elementSort)),
+                  attrs.add("right"));
```
ehildenb: Should this be `left` instead?

ehildenb (commented):
We should generate parsers for _all_ `$..` vars in the initial configuration, not just `$PGM`, for both the LR and GLR parsers.

ehildenb (approved):

PR #1250, opened 2020-05-02
Fixes to published docker images to make useful for CI

ehildenb:
Currently the Jenkins CI jobs which use the Docker images get confused by how this images is setup.

Instead of adding the user `user:user` here, we let each individual job do that, so that they can create it with the correct uid and gid. This one now just installs K package and that's it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -191,7 +191,10 @@ pipeline {
               }
             }
             stage('Build and Package on Debian Buster') {
-              when { branch 'master' }
+              when {
+                branch 'master'
+                beforeAgent true
+              }
               stages {
                 stage('Build on Debian Buster') {
                   agent {
@@ -247,7 +250,10 @@ pipeline {
               }
             }
             stage('Build and Package on Arch Linux') {
-              when { branch 'master' }
+              when {
+                branch 'master'
+                beforeAgent true
+              }
               stages {
                 stage('Build on Arch Linux') {
                   agent {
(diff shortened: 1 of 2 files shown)


ehildenb:
This is now working with every job that the Jenkinsfile auto-updates K for:

- https://github.com/kframework/evm-semantics/pull/785
- https://github.com/kframework/wasm-semantics/pull/331
- https://github.com/runtimeverification/michelson-semantics/pull/27
- https://github.com/runtimeverification/beacon-chain-spec/pull/156
- https://github.com/runtimeverification/beacon-chain-verification/pull/75

dwightguth (commented):
You realize that a separate RUN layer that deletes the file after it's been used won't actually remove it from the docker image, right? You need a separate stage for that.

Also, it looks like this container runs as root, are you sure that's what you want?

ehildenb:
@dwightguth the impression I had was that the debian file will be in the images for two of the layers, but will not be in the image that users pull down with `FROM runtimeverificationinc/kframework-k:ubuntu-bionic-master`, for instance.

And yes, I mean for everything to run as root. I figure any user-setup should be up to the project themselves, and all we're doing is literally installing one package, so the image should be as bare-bones as possible.

The problem I was running into is that setting up the `user` account here was messing up other CI jobs, where we tell Jenkins to "Run as user `user`, but with whatever UID you are currently using for user `jenkins-slave` on the host machine.". This creates an image with K pre-installed without making any commitments to how the end user of the image will setup the linux user accounts.

dwightguth (approved):

PR #1252, opened 2020-05-04
Function encoding

dwightguth:
This is not ready to be merged yet, but it is ready to get initial feedback. I don't expect the tests will pass on the haskell or llvm backends, but I want to get reviews to make sure it looks like what we agreed on to people before I start working on getting it working in the llvm backend.

Sample input: 

```
syntax Int ::= foo(Bar, Baz) [function]

syntax Bar ::= "bar" | "bar2"
syntax Baz ::= "baz"

rule foo(bar, baz) => 0
  requires 0 >Int 0
rule foo(B:Bar, baz) => 1
rule foo(...) => 2
  requires 1 >Int 1 [owise]
```

Sample output:

```
  axiom{R} \implies{R} (
    \and{R}(
      \top{R}(),
      \and{R} (
          \in{SortBar{}, R} (
            X0:SortBar{},
            VarB:SortBar{}
          ),\and{R} (
          \in{SortBaz{}, R} (
            X1:SortBaz{},
            Lblbaz'Unds'TEST'Unds'Baz{}()
          ),
          \top{R} ()
        ))),
    \and{R} (
      \equals{SortInt{},R} (
        Lblfoo'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Bar'Unds'Baz{}(X0:SortBar{},X1:SortBaz{}),
        \dv{SortInt{}}("1")),
      \top{R}())) [...]

  axiom{R} \implies{R} (
    \and{R} (
      \not{R} (
        \or{R} (
          \exists{R} (Var'Unds'4:SortBar{},
            \and{R} (
              \top{R}(),
              \and{R} (
                \in{SortBar{}, R} (
                  X0:SortBar{},
                  Var'Unds'4:SortBar{}
                ),\and{R} (
                \in{SortBaz{}, R} (
                  X1:SortBaz{},
                  Lblbaz'Unds'TEST'Unds'Baz{}()
                ),
                \top{R} ()
              ))
          )),
          \or{R} (
            \and{R} (
              \equals{SortBool{},R}(
        Lbl'Unds-GT-'Int'Unds'{}(\dv{SortInt{}}("0"),\dv{SortInt{}}("0")),
        \dv{SortBool{}}("true")),
              \and{R} (
                \in{SortBar{}, R} (
                  X0:SortBar{},
                  Lblbar'Unds'TEST'Unds'Bar{}()
                ),\and{R} (
                \in{SortBaz{}, R} (
                  X1:SortBaz{},
                  Lblbaz'Unds'TEST'Unds'Baz{}()
                ),
                \top{R} ()
              ))
          ),
          \bottom{R}()
        ))
      ),
      \and{R}(
        \equals{SortBool{},R}(
        Lbl'Unds-GT-'Int'Unds'{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("1")),
        \dv{SortBool{}}("true")),
        \and{R} (
          \in{SortBar{}, R} (
            X0:SortBar{},
            Var'Unds'0:SortBar{}
          ),\and{R} (
          \in{SortBaz{}, R} (
            X1:SortBaz{},
            Var'Unds'1:SortBaz{}
          ),
          \top{R} ()
        ))
    )),
    \and{R} (
      \equals{SortInt{},R} (
        Lblfoo'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Bar'Unds'Baz{}(X0:SortBar{},X1:SortBaz{}),
        \dv{SortInt{}}("2")),
      \top{R}())) [...]

  axiom{R} \implies{R} (
    \and{R}(
      \equals{SortBool{},R}(
        Lbl'Unds-GT-'Int'Unds'{}(\dv{SortInt{}}("0"),\dv{SortInt{}}("0")),
        \dv{SortBool{}}("true")),
      \and{R} (
          \in{SortBar{}, R} (
            X0:SortBar{},
            Lblbar'Unds'TEST'Unds'Bar{}()
          ),\and{R} (
          \in{SortBaz{}, R} (
            X1:SortBaz{},
            Lblbaz'Unds'TEST'Unds'Baz{}()
          ),
          \top{R} ()
        ))),
    \and{R} (
      \equals{SortInt{},R} (
        Lblfoo'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Bar'Unds'Baz{}(X0:SortBar{},X1:SortBaz{}),
        \dv{SortInt{}}("0")),
      \top{R}())) [...]
```


yzhang90:
@dwightguth according to https://github.com/kframework/kore/pull/1789/files,
in the first sample rule,
```
 Lblfoo'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Bar'Unds'Baz{}(VarB:SortBar{},Lblbaz'Unds'TEST'Unds'Baz{}())
```
 should be
```
 Lblfoo'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Bar'Unds'Baz{}(X0:SortBar{},X1:SortBaz{})
```
, right?

dwightguth:
Whoops, good catch. I thought I fixed that but apparently I only fixed half of the issue.

dwightguth:
Updated the PR and the descripton. @yzhang90 take another look.

traiansf:
@dwightguth Can you confirm that this PR does not change the encoding of simplification rules?

dwightguth:
You're right that I forgot to preserve the old behavior for simplification rules. Do we want to forbid the owise attribute on simplification rules?

ttuegel:
> Do we want to forbid the owise attribute on simplification rules?

I think we should forbid `owise` and `simplification` on the same rule. With the Haskell backend, it wouldn't make sense: each simplification rule is attempted independently of any others.

Aside from the issues already pointed out, this looks good to me!

yzhang90:
LGTM

dwightguth:
Assuming the tests pass, I believe work on the llvm backend is done for this PR. It is still waiting on the llvm backend PR to be merged, and for the haskell backend to work. I am not going to merge the llvm backend PR until the haskell backend is working, in case it requires further changes based on information learned while making the haskell backend work. But the haskell backend support is the only thing left to support in this PR.

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    private void check(Att att, HasLocation loc) {
+        if (att.contains(Att.OWISE()) && att.contains(Att.SIMPLIFICATION())) {
+          errors.add(KEMException.compilerError("owise attribute is not supported on simplification rules.", loc));
+        }
+        if (att.contains(Att.PRIORITY()) && att.contains(Att.SIMPLIFICATION())) {
+          errors.add(KEMException.compilerError("priority attribute is not supported on simplification rules.", loc));
+        }
```
ana-pantilie, line 37: We should allow priority on simplification rules. For example, this is needed for the arithmetic simplification lemmas.

traiansf: @ana-pantilie could you provide an example?

ana-pantilie: @traiansf [These](https://github.com/kframework/k/blob/a34d4d155bb4bfa58194a892f36845b411f373b1/k-distribution/include/kframework/builtin/domains.k#L451) rules need to apply after any of the less general rules below.

traiansf: @ana-pantilie I agree it makes sense to try some simplification rules only after other simplification rules have been attempted, but I'm wondering whether using `priority` would not create confusion.
The reason I'm saying this is that for other cases (rules, function definitions), priority has semantic meaning (forces a check that rules with less priority cannot be applies), while priority for simplification axioms probably shouldn't, as it is more of a preference to attempt it before other simplification axioms. Maybe we could allow the `simplification` attribute to take an extra argument?

ana-pantilie: @traiansf Yes, you're right, it could be confusing. @ttuegel what do you think?

ttuegel: > Maybe we could allow the simplification attribute to take an extra argument?

That sounds fine.

ttuegel: @dwightguth I'm trying to add an optional argument to the `simplification` attribute so we can write (for example) `simplification(51)`. When I add arguments like that, the frontend starts to encode `simplification` (without arguments) as `simplification("")` in the generated Kore. I want similar behavior to the `concrete` attribute: `concrete` becomes `concrete()`, `concrete(X)` becomes `concrete("X")`, etc. Can you advise how I might do that?

dwightguth: I have no idea how that happens the way it does now with the concrete attribute. If anything that sounds suspicious to me and indicates that the code is probably doing something unintended. The way it is supposed to work in the current code is to decide whether the attribute has an argument based on whether a nonempty string appears anywhere as the argument of the attribute. So in that sense what you describe is the intended behavior. The k frontend doesn't actually know how to distinguish between an attribute `foo` and an attribute `foo()`. This could in theory be changed, but it's not remotely trivial and would likely be a very intrusive change to the frontend.

ttuegel:
Blocked on kframework/kore#2029.

ttuegel (approved):

ttuegel:
I am going to proceed with merging this _after_ we track down a performance regression in the Haskell backend. I expect that to be completed early next week. I want to delay merging this pull request so that we aren't contending with several different changes that may all impact performance.

PR #1254, opened 2020-05-05
domains.k: #Ceil rules for padLeftBytes, padRightBytes

denis-bogdanas:
#Ceil rules are based on K implementation for `padLeftBytes`, `padRightBytes`. Hooked rules must have same definedness requirements, e.g. should accept negative Length (2nd argument).

Required for ERC20 verifier.

### DIFF
--- a/k-distribution/include/kframework/builtin/domains.k
+++ b/k-distribution/include/kframework/builtin/domains.k
@@ -830,9 +830,6 @@ module BYTES-IN-K
   rule padLeftBytes(BS, LEN, VAL) => BS requires lengthBytes(BS) >=Int LEN andBool 0 <=Int LEN
   rule padLeftBytes(BS, LEN, VAL) => padLeftBytes(VAL : BS, LEN, VAL) requires lengthBytes(BS) <Int LEN andBool 0 <=Int LEN
 
-  rule #Ceil(padRightBytes(_, LEN, VAL)) => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
-  rule #Ceil(padLeftBytes(_, LEN, VAL))  => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
-
   syntax Bytes ::= reverseBytes(Bytes) [function, functional]
   syntax Bytes ::= reverseBytes(Bytes, Bytes) [function, klabel(reverseBytesAux)]
   rule reverseBytes(BS) => reverseBytes(BS, nilBytes)
@@ -860,6 +857,10 @@ endmodule
 
 module BYTES-SYMBOLIC [symbolic, kast]
   imports BYTES-IN-K
+
+  rule #Ceil(padRightBytes(_, LEN, VAL)) => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
+  rule #Ceil(padLeftBytes(_, LEN, VAL))  => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
+
 endmodule
 
 module BYTES
--- a/k-distribution/include/kframework/builtin/domains.k
+++ b/k-distribution/include/kframework/builtin/domains.k
@@ -864,8 +864,8 @@ module BYTES-SYMBOLIC-CEIL [symbolic, kore]
   imports BYTES-HOOKED
   imports INT
 
-  rule #Ceil(padRightBytes(_, LEN, VAL)) => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [simplification]
-  rule #Ceil(padLeftBytes(_, LEN, VAL))  => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [simplification]
+  rule #Ceil(padRightBytes(_, LEN, VAL)) => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
+  rule #Ceil(padLeftBytes(_, LEN, VAL))  => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
 endmodule
 
 module BYTES
--- a/k-distribution/include/kframework/builtin/domains.k
+++ b/k-distribution/include/kframework/builtin/domains.k
@@ -827,11 +827,11 @@ module BYTES-IN-K
   syntax Bytes ::= padRightBytes(Bytes, Int, Int) [function]
                  | padLeftBytes(Bytes, Int, Int) [function]
   rule padRightBytes(BS, LEN, VAL) => reverseBytes(padLeftBytes(reverseBytes(BS), LEN, VAL))
-  rule padLeftBytes(BS, LEN, VAL) => BS requires lengthBytes(BS) >=Int LEN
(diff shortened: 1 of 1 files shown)


k-distribution/include/builtin/domains.k
```diff
@@ -836,6 +836,9 @@ module BYTES-IN-K
   rule padLeftBytes(BS, LEN, VAL) => BS requires lengthBytes(BS) >=Int LEN
   rule padLeftBytes(BS, LEN, VAL) => padLeftBytes(VAL : BS, LEN, VAL) requires lengthBytes(BS) <Int LEN
 
+  rule #Ceil(padRightBytes(_,_, VAL)) => {(0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere]
+  rule #Ceil(padLeftBytes(_,_, VAL))  => {(0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere]
```
ehildenb: Shouldn't this also require that the middle argument be non-zero? What do the backend implementations of this assume? We should strive to make sure that this is an accurate description of what the users can expect the backends to be able to handle.

k-distribution/include/builtin/domains.k
```diff
@@ -836,6 +836,9 @@ module BYTES-IN-K
   rule padLeftBytes(BS, LEN, VAL) => BS requires lengthBytes(BS) >=Int LEN
   rule padLeftBytes(BS, LEN, VAL) => padLeftBytes(VAL : BS, LEN, VAL) requires lengthBytes(BS) <Int LEN
```
denis-bogdanas: Judging by K implementation there are no restrictions on middle argument. Hooked implementation must be reviewed and adapted so that it matches K implementation. If this is not practical then the discrepancy must be commented.

dwightguth: We should change the K implementation so that the len requirement has to be nonnegative.

denis-bogdanas:
@dwightguth Can you look why test is failing? The test is from `prelude-warnings` directory and it complains about `padLeftBytes` symbol inside `#Ceil`. Test fails on my computer too.
```
[Error] Compiler: Illegal function symbol on LHS of rule.
        Source(/mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k)
        Location(839,14,839,37)
[Error] Compiler: Illegal function symbol on LHS of rule.
        Source(/mnt/d/k5/k-distribution/target/release/k/include/builtin/domains.k)
        Location(840,14,840,36)
```
See rules quoted by Everett.

denis-bogdanas:
Ok, I found where warning comes from. It's a hack in front-end code that allows `#Ceil` rules for a hardcoded set of functions. Will add `padLeftBytes` and `padRightBytes` to the list.

dwightguth:
That's not the right solution... that's not a hard coded list of symbols that can appear under Ceil on the lhs, that's a list of functions that can appear on the lhs of patterns in concrete backends, of which padRightBytes and padLeftBytes are not. The solution is just to put it in a module with the `symbolic` attribute, like all the other `#Ceil` rules.

denis-bogdanas:
@dwightguth ready

k-distribution/include/builtin/domains.k
```diff
@@ -863,6 +863,10 @@ endmodule
 
 module BYTES-SYMBOLIC [symbolic, kast]
   imports BYTES-IN-K
+
+  rule #Ceil(padRightBytes(_, LEN, VAL)) => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [anywhere, simplification]
```
dwightguth: I have the same comment I have on the other PR. If these are intended for the haskell backend, they should not be in the module tagged [kast].

dwightguth (changes requested):

k-distribution/include/builtin/domains.k
denis-bogdanas: @dwightguth Should I create a separate module just for these rules?

ehildenb: They should be in a module tagged `kore` or `kore, symbolic`

denis-bogdanas:
I've put the rules in a new module. Anything else needed?

ehildenb (approved):
LGTM

k-distribution/include/builtin/domains.k
```diff
@@ -830,11 +830,11 @@ module BYTES-IN-K
   rule replaceAtBytes(B : BS, N, BS') => B : replaceAtBytes(BS, N -Int 1, BS') requires N >Int 0
   rule replaceAtBytes(_ : BS, 0, B : BS') => B : replaceAtBytes(BS, 0, BS')
 
-  syntax Bytes ::= padRightBytes(Bytes, Int, Int) [function]
-                 | padLeftBytes(Bytes, Int, Int) [function]
+  syntax Bytes ::= padRightBytes(Bytes, Int, Int) [function, klabel(padRightBytes)]
```
dwightguth: I don't understand this change; why was it necessary?

k-distribution/include/builtin/domains.k
```diff
@@ -863,6 +863,14 @@ endmodule
 
 module BYTES-SYMBOLIC [symbolic, kast]
   imports BYTES-IN-K
+  imports BYTES-SYMBOLIC-CEIL
```
dwightguth: this entire module is not going to get imported on the haskell backend, thus, neither will this import.

dwightguth (changes requested):

k-distribution/include/builtin/domains.k
denis-bogdanas: I don't remember. Probably for debugging. Can be removed.

k-distribution/include/builtin/domains.k
denis-bogdanas: Can you explain how module attributes work? Will `[symbolic, kast]` get imported by union or intersection of backends marked with [symbolic] and [kast]? If it's union, I guess hashell backend will import `BYTES-SYMBOLIC`.

dwightguth: A module tagged `symbolic, kast` will be imported only by the java backend.

k-distribution/include/builtin/domains.k
dwightguth: I'd rather we remove it because as it stands it's not going to do anything, which makes it redundant.

k-distribution/include/builtin/domains.k
denis-bogdanas: Ok, so what's your solution, where to put those rules? I have no idea based on your explanation. Feel free to close the PR/fix/suggest an actionable fix.

ehildenb: It's the intersection of the `kore` and `symbolic` which makes it the Haskell backend.

So you may need to make a new module.

denis-bogdanas:
@dwightguth all done.

denis-bogdanas:
@ttuegel Can you check why this PR fails? It adds the rules for `#Ceil(padRightBytes()), #Ceil(padLeftBytes())`

Haskell backend throws: `Expected function-like left-hand side of rule, but found...`

ttuegel:
The frontend is doing cell completion on the `#Ceil` rules; this causes the error seen above.

@denis-bogdanas You should be able to work around this by adding the `anywhere` attribute to the rules.

@dwightguth Do you know why the frontend is doing this when there are no constructors on the left-hand side of the rule?

dwightguth:
@ttuegel it is /supposed/ to look at the symbol underneath the connective to determine whether the rule should be completed to the top, but something must be going wrong. You are welcome to investigate: the code in question is at AddTopCellToRules.

dwightguth (approved):
Looks good finally. Please make sure to get @ttuegel to approve before merging.

ttuegel (approved):

PR #1257, opened 2020-05-05
Implement set union operation per discussion with Tom

Andrew-Miranti:
The Michelson semantics needs a Set union operation, but `_Set_` is not Set union, but rather set concatenation ( See https://github.com/kframework/k/issues/1134 for details ).  So I've written Set union in terms of set difference and concatenation.


dwightguth:
Please make this a hooked operation SET.union. I will update the llvm backend to add support for this hook.

k-distribution/include/builtin/domains.k
```diff
@@ -249,6 +251,10 @@ module SET
 
   /*@ Get an arbitrarily chosen element of the Set */
   syntax KItem ::= choice(Set)                [function, hook(SET.choice), klabel(Set:choice)]
+
+  /*@ Get the union of two sets ($A \cup B$) */
+  syntax Set ::= Set "USet" Set            [function, functional, hook(SET.union)]
```
dwightguth: let's make this |Set.

dwightguth (changes requested):

dwightguth (approved):

k-distribution/include/kframework/builtin/domains.k
```diff
@@ -242,6 +244,10 @@ module SET
 
   /*@ Get an arbitrarily chosen element of the Set */
   syntax KItem ::= choice(Set)                [function, hook(SET.choice), klabel(Set:choice)]
+
+  /*@ Get the union of two sets ($A \cup B$) */
+  syntax Set ::= Set "|Set" Set            [function, functional, hook(SET.union)]
```
nishantjr:
```suggestion
  syntax Set ::= Set "|Set" Set            [left, function, functional, hook(SET.union)]
```

nishantjr:
@dwightguth Reminder about this. I added the `left` attribute to the `|Set` operator after the review.

dwightguth:
Looks like the test is failing because of the OCAML backend. It shouldn't be hard to fix; just add the hook_union function to the SET module in hooks.ml and make it call hook_concat.

dwightguth:
It looks like the fix to DefinitionParsing.java in this file might be a fix for the same issue as #1388. At the very least, the fix is related, making the fix for that issue blocked on this PR. I am going to take this PR over and fix the remaining issues so it can be merged.

dwightguth (approved):

dwightguth:
Not sure why the map-symbolic test failed, but this PR is no longer blocking me, so I'm going to unassign myself. It might just be some kind of flaky test, or it might be some odd interaction with the changes to DefinitionParsing.java...

PR #1259, opened 2020-05-05
Add more support for Bison and add documentation

### LINKED ISSUE #1256: Add option to generate parsers for all configuration variables
Currently the generated Bison parsers (LR(1) and GLR) only generate a parser for `$PGM`.

We should make it easier to generate parsers for _all_ configuration variables in `kompile` which will make it more useful.

dwightguth:
Fixes #1256.


k-distribution/tests/regression-new/parseNonPgm/1.test.out
```diff
@@ -0,0 +1,8 @@
+11<generatedTop>
+  <k>
+    .Pgm ~> .
+  </k>
```
ehildenb, line 4: What's going on here?

pending-documentation.md
```diff
@@ -1665,6 +1665,72 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+   `--gen-bison-parser` or `--gen-glr-bison-parser` flags to `kompile`.
+   `kompile` will decide the sorts to use as start symbols based on the sorts
+   in the configuration declaration for the configuration variables. The $PGM
+   configuration variable will be generated based on the main syntax module
+   of the definition. The user must explicitly annotate other modules with the
+   configuration variables that they should be used to parse as attributes.
+   For example, `module TEST [parser(FOO)]` will tell `kompile` to parse the 
+   $FOO configuration variable using the syntax in the TEST module. If the user
+   forgets to annotate the modules with the configuration variables, only the 
+   $PGM parser will be generated.
```
ehildenb: I don't particularly like this way of specifying the module to parse in, because it isn't local to where the variable is declared. For example, what if you want to use the `INT-SYNTAX` module to parse a given configuration variable?

I think it's better to either,

1.   have the annotation on the cell itself (eg. `<foo parseModule="FOO-SYNTAX"> $FOO:Foo </foo>`), but this isn't perfect because what if the cell is like `<foo> bar($FOO:Foo) </foo>`.
2.   as an attribute on the module which declares the configuration
3.   don't directly generate parsers for configuration variables, instead for pairs of `(Sort, Module)`. So we would generate `parser_Pgm_MAIN-SYNTAX-MODULE`, `parser_Pgm_SOME-OTHER-MODULE`, `parser_Foo_FOO-SYNTAX`, `parser_Foo_MAIN-SYNTAX-MODULE`, where `Pgm` and `Foo` are sorts, not configuration variables. We can still always generate the default `parser_Pgm_MAIN-SYNTAX-MODULE`. Then `kompile/kast` can take a variadic number of Sort/Module pairs (default just being `[(sort(PGM), MAIN-SYNTAX-MODULE)]`) to generate the parsers.

In order of preference, I would say (3), (1), (2). Mostly I like (3) because it makes it exposes what the actual functionality is, which has little to do with the configuration variables, and more to do with parsers for sorts in modules. The only reason I see it getting conflated with configuration variables is because those are the most useful parsers to generate.

k-distribution/tests/regression-new/parseNonPgm/1.test.out
ehildenb: NVM, the `11` comes from the `#write()`

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -154,10 +154,24 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
...
-            for (Module mod : iterable(def.getParsedDefinition().entryModules())) {
-              if (mod.att().contains("parser")) {
-                new KRead(kem, files, InputModes.PROGRAM).createBisonParser(def.programParsingModuleFor(mod.name(), kem).get(), def.configurationVariableDefaultSorts.getOrDefault("$" + mod.att().get("parser"), Sorts.K()), files.resolveKompiled("parser_" + mod.att().get("parser")), kompileOptions.experimental.genGlrBisonParser);
-              }
+            for (Production prod : iterable(kompiledDefinition.mainModule().productions())) {
+                if (prod.att().contains("cell") && prod.att().contains("parser")) {
+                    String att = prod.att().get("parser");
+                    String[] parts = att.trim().split(";");
+                    for (String part : parts) {
+                        String[] subparts = part.trim().split(",");
```
ehildenb: I believe we use this same strategy for getting subparts of many axioms. Can you look and see if it's reasonable to factor out a method directed at it?

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+   in the configuration declaration for the configuration variables. The $PGM
+   configuration variable will be generated based on the main syntax module
+   of the definition. The user must explicitly annotate the configuration
+   declaration with the other modules to use to parse the other configuration
+   variables as attributes. For example, if I have the following cell in the
+   configuration declaration: `<cell> foo($FOO:Foo, $BAR:Bar) </cell>`,
+   One might annotate it with the attribute pair `parser="FOO, TEST; BAR, TEST2"`
+   to indicate that configuration variable `$FOO` should be parsed in the
+   `TEST` module, and configuration variable `$BAR` should be parsed in the
+   `TEST2` moducle. If the user forgets to annotate the declaration with the
```
```suggestion
   `TEST2` module. If the user forgets to annotate the declaration with the
```

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+In addition to on-the-fly parser generation using `kast`, K is capable of
+ahead-of-time parser generation of LR(1) or GLR parsers using Flex and Bison.
+This can be done one of two different ways.
+
+1. You can explicitly request for a particular parser to be generated by
+   invoking `kast --gen-parser <outputFile>` or
+   `kast --gen-glr-parser <outputFile>` respectively. `kast` will then create a
+   parser based on the same command line flags that govern on-the-fly parsing,
+   like `-s` to specify the starting sort, and `-m` to specify the module to
+   parse under. By default, this generates a parser for the sort of the $PGM
```
```suggestion
   parse under. By default, this generates a parser for the sort of the `$PGM`
```

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+   configuration variable in the main syntax module of the definition.
+2. You can request that a specific set of parsers be generated for all the
+   configuration variables of your definition by passing the
+   `--gen-bison-parser` or `--gen-glr-bison-parser` flags to `kompile`.
+   `kompile` will decide the sorts to use as start symbols based on the sorts
+   in the configuration declaration for the configuration variables. The $PGM
```
```suggestion
   in the configuration declaration for the configuration variables. The `$PGM`
```

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+   configuration variable will be generated based on the main syntax module
+   of the definition. The user must explicitly annotate the configuration
+   declaration with the other modules to use to parse the other configuration
+   variables as attributes. For example, if I have the following cell in the
+   configuration declaration: `<cell> foo($FOO:Foo, $BAR:Bar) </cell>`,
+   One might annotate it with the attribute pair `parser="FOO, TEST; BAR, TEST2"`
+   to indicate that configuration variable `$FOO` should be parsed in the
+   `TEST` module, and configuration variable `$BAR` should be parsed in the
+   `TEST2` module. If the user forgets to annotate the declaration with the
+   parser attribute, only the $PGM parser will be generated.
```
```suggestion
   parser attribute, only the `$PGM` parser will be generated.
```

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+
+Bison-generated parsers are extremely fast compared to `kast`, but they have
+some important limitations:
+
+* Bison parsers will always output kore. You can then pass the resulting AST
```
```suggestion
* Bison parsers will always output Kore. You can then pass the resulting AST
```

pending-documentation.md
```diff
@@ -1665,6 +1665,75 @@ Since this feature is not clearly needed by K users at the moment, it is only
...
+  few conflicts as possible if they are using GLR parsing. Bison's GLR support
+  is quite primitive, and in the worst case it can use exponential space and
+  time to parse a program, which generally leads the generated parser to report
+  "memory exhausted", indicating that the parse could not be completed within
+  the stack space allocated by Bison. It's best to ensure that the grammar is 
+  as close to LR(1) as possible and only utilizes conflicts where absolutely
+  necessary. One tool that can be used to facilitate this is to pass
+  `--bison-lists` to kompile. This will disable support for the `List{Sort}`
+  syntax production, and it will make `NeList{Sort}` left associative, but the
+  resulting productions generated for NeList will be LR(1) and use bounded
```
```suggestion
  resulting productions generated for `NeList{Sort}` will be LR(1) and use bounded
```

kore/src/main/java/org/kframework/utils/StringUtil.java
```diff
@@ -822,4 +699,22 @@ public static String decodeKoreString(String encoded) {
...
+
+    public static String[] splitOneDimensionalAtt(String att) {
+        String[] splitted = att.trim().split(",");
+        for (int i = 0; i < splitted.length; i++) {
+            splitted[i] = splitted[i].trim();
+        }
+        return splitted;
+    }
+
+    public static String[][] splitTwoDimensionalAtt(String att) {
```
ehildenb, line 711: I know we went over this, but if you can look again and see if this is re-usable in the places where we are doing this type of splitting, that would be good. I'm going to approve anyway.

ehildenb (approved):

PR #1268, opened 2020-05-07
Support priority attribute in macros

### LINKED ISSUE #1263: Macros respect priority attribute
Currently macros only respect the `owise` attribute, but it can be convenient to have a chain of macros which each pattern match differently, and have the user specify the order to  try them with `priority`.

Maybe a good way to implement this is to pull the priority-completion (adding the `priority` attribute to rules) out of [here](https://github.com/kframework/k/blob/cdc484531cf3e0149839c531d58f55f20bbb6cb1/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L285) in ModuleToKore, and making it an explicit `kompile` pass run before `ExpandMacros`. Then the comparator in `ExpandMacros` [here](https://github.com/kframework/k/blob/2e58f08ef9084ff12e3ef77d870a5640a6f50605/kernel/src/main/java/org/kframework/compile/ExpandMacros.java#L85) can be modified to just read off the `priority` attribute.

dwightguth:
Fixes #1263


kernel/src/main/java/org/kframework/compile/ExpandMacros.java
```diff
@@ -107,6 +107,19 @@ public ExpandMacros(ResolveFunctionWithConfig transformer, Module mod, FileUtil
...
+            try {
+                return Integer.parseInt(att.get(Att.PRIORITY()));
+            } catch (NumberFormatException e) {
+                throw KEMException.compilerError("Invalid value for priority attribute: " + att.get(Att.PRIORITY()) + ". Must be an integer.", e);
+            }
+        } else if (att.contains(Att.OWISE())) {
+            return 200;
+        }
+        return 50;
+    }
```
ehildenb: This same function should be re-used for `ModuleToKORE` to make sure that we keep any logic about how to compute the priority of a rule in sync.

ehildenb (approved):

PR #1269, opened 2020-05-07
Check for illegal function in lhs after applying macros

### LINKED ISSUE #1262: Bad error when rule pattern has function symbols not at head
If a user defines a macro that is not fully evaluated on the RHS (has function symbols), then the macro is used in the LHS of a rule, it will result in a rule with function symbols in the LHS.

Instead of being caught at build-time, this is caught at run-time by the LLVM backend (in some cases by halt and some cases by crash).

K Definition

```
module FIXED-INT
    imports INT
    imports BOOL

    syntax FInt ::= FInt ( value: Int , one: Int ) [klabel(FInt), symbol]
 // ---------------------------------------------------------------------

    syntax Wad = FInt
    syntax Wad ::= wad ( Int )
 // --------------------------
    rule wad(I) => FInt(I *Int WAD, WAD) [macro]

    syntax Int ::= "WAD"
 // --------------------
    rule WAD => 1000000000000000000 [macro]

    syntax Pgm ::= "go" FInt | Pgm ";" Pgm [left]
 // ---------------------------------------------
    rule <k> P ; Q => P ~> Q ... </k>
    rule <k> go wad(0) => . ... </k>
endmodule
```

Input program:

```
go wad(0) ; go wad(1)
```

Expected output: Fail at kompile time.
Actual output (kompile succeeds, stuck at runtime):

```
<k>
  go FInt ( 0 , 1000000000000000000 ) ~> go FInt ( 1000000000000000000 , 1000000000000000000 )
</k>
```

On a more complicated example, the LLVM backend actually crashes at this point in execution (the entire MKR semantics).

dwightguth:
Fixes #1262


ehildenb:
@dwightguth I'm somewhat confused about the architecture of this.

The `CheckFunctions` pass is already called in one place AFAICT, which is _after_ parsing but _before_ the kompile pipeline is applied (via `checkDefinition -> structuralChecks -> CheckFunctions` here: https://github.com/kframework/k/blob/fae52587395497ac83938c51b8793e7a61ae7677/kernel/src/main/java/org/kframework/kompile/Kompile.java#L135).

But here we're adding another run immediately after Macro expansion of this routine, but _directly_ in the MacroExpander.

It seems simpler to (in terms of understanding the dependencies in the Kompile pipeline) to go with one of two choices:

1.  We factor out CheckFunctions into its own Kompile pipeline step, and just call it immediately after ExpandMacros.
2.  We have a post-kompile-pipeline check step here https://github.com/kframework/k/blob/fae52587395497ac83938c51b8793e7a61ae7677/kernel/src/main/java/org/kframework/kompile/Kompile.java#L138 where we add this check in.

But right now, it's confusing to me that we are calling CheckFunctions in two places, the first place being _before_ Macro Expansion happens, which in my thinking should _always_ fail because ExpandMacros needs to remove some function symbols from rules where they shouldn't be.

ehildenb (approved):

PR #1272, opened 2020-05-08
Reorganize K distribution to remove dependence on /usr/lib/kframework sysroot [brew-staging: reorg]

### LINKED ISSUE #1260: Clarify distinction between installs at `/usr/{lib,bin}` and `/usr/lib/kframework/{lib,bin}`
Which files are installed where here? Is it possible to just have them all in one place? Does it differ between distros?

dwightguth:
Fixes #1260 

Basically, everything is now under /usr or /usr/local rather than /usr/lib/kframework. We create subdirectories under /usr/lib, /usr/include, and /usr/share in order to avoid conflicts.

Finally ready for review.

This encompasses three changes that must be merged largely in sync, so please do not merge this or the other two PRs until after all three have been approved. There are also the changes to the homebrew package that you should consider, but I did not create a PR because Jenkins is the one that merges those changes.

Here are the other diffs:

https://github.com/kframework/llvm-backend/pull/331
https://github.com/kframework/k-exercises/pull/19
https://github.com/kframework/homebrew-k/compare/master...reorg


Jenkinsfile
```diff
@@ -50,7 +50,7 @@ pipeline {
                     }
                   }
                   stages {
-                    stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git' } } }
+                    stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git', branch: 'reorg' } } }
```
ehildenb: Make sure this gets changed back before merge.

Jenkinsfile
```diff
@@ -318,15 +318,16 @@ pipeline {
                   agent { label 'anka' }
                   steps {
                     unstash 'src'
+                    dir('kframework') {
+                      checkout scm
+                    }
```
```suggestion
                    dir('kframework') { checkout scm }
```

Sticking to the formatting where we use less lines has made the Jenkinsfile more readable on many projects.

k-distribution/include/kframework/kore
```diff
@@ -0,0 +1 @@
+../../../haskell-backend/src/main/native/haskell-backend/src/main/kore/
```
ehildenb, line 1: What is this file used for? Searching for `kframework/kore` in this state-delta doesn't reveal anything where it's using this file for something AFAICT.

k-distribution/src/main/scripts/bin/k-configure-opam-common
```diff
@@ -1,18 +1,19 @@
...
 case "$(opam --version)" in
 1.*)
+        OCAML_VERSION="4.06.1+k$1"
 	echo Detected Opam version 1
 	opam init $INIT_ARGS
 	add_repo
 	opam switch "$OCAML_VERSION"
 	;;
 2.*)
+        OCAML_VERSION="4.10.0$1"
```
ehildenb: Spacing of these lines looks weird.

k-distribution/tutorial/find-k.mak
```diff
@@ -1,10 +1,12 @@
...
+else ifneq (,$(wildcard ${MAKEFILE_PATH}/../include/kframework/ktest.mak))
 export K_HOME?=${MAKEFILE_PATH}/..
-else ifneq (,$(wildcard /usr/lib/kframework/include/ktest.mak))
-export K_HOME?=/usr/lib/kframework
-else ifneq (,$(wildcard /usr/local/lib/kframework/include/ktest.mak))
-export K_HOME?=/usr/local/lib/kframework
+else ifneq (,$(wildcard /usr/include/kframework/ktest.mak))
+export K_HOME?=/usr
+else ifneq (,$(wildcard /usr/local/include/kframework/ktest.mak))
+export K_HOME?=/usr/local
```
ehildenb, line 9: Why do we need to check so many places? I thought that this was to make the installation point standard across all distributions? Is the `/usr/include/kframework` vs `/usr/local/include/kframework` a difference between Linux/MacOS.

I guess the point is more to make it so that we're not doing everything under `/usr/lib/kframework/...`, but instead directly inside `/usr/...` (on Linux) or `/usr/local/...` (on MacOS)?

ehildenb (commented):
Looks pretty good overall, a few nits inline, and one general comment.

It looks like we're removing all explicit Windows support, but searching for Windows still reveals a few places in the code that we should probably remove things from:

```
[dev@dev-ubuntu k]% git grep -i windows
README.md:Mac/Windows:
README.md:Windows:
README.md:## Installing on fresh Windows Subsystem for Linux
README.md:1. Install the Ubuntu package from the Windows Store, which as of now is an alias for the Ubuntu LTS 18.04 package. During installation you will be asked to create a new user. 
java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java:     * and `cmd /c 'cmd'` on windows.
java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java:        if (OS.current() == OS.WINDOWS) {
k-distribution/INSTALL.md:**NOTE**: We do not currently support running K on native Windows. To use K on
k-distribution/INSTALL.md:Windows, you are encouraged to install
k-distribution/INSTALL.md:[Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
k-distribution/INSTALL.md:    You should install on Windows by following the instructions for the Linux
k-distribution/INSTALL.md:    distribution you installed for Windows Subsystem for Linux.
k-distribution/include/kframework/builtin/domains.k:  // `cmd /c cmd:String` on windows and returns #systemResult
k-distribution/src/main/scripts/lib/setenv:  OS=windows
k-distribution/src/main/scripts/lib/setenv:  export PATH="$K_LIB_DIR/native/windows$ARCH":"$K_LIB_DIR/native/$ARCH":$PATH
k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java:            // fixing Windows line endings (git autocrlf=auto generates Windows line endings on checkout)
k-distribution/tutorial/NOTES.md:Windows user currently need to replace "kast" with "kast.bat" everywhere in their
kernel/pom.xml:      <classifier>windows64</classifier>
kernel/pom.xml:      <classifier>windows32</classifier>
kernel/pom.xml:      <classifier>windows</classifier>
kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java:                realScanner.close();//required for Windows.
kernel/src/main/java/org/kframework/kserver/KServerFrontEnd.java:        return (OS.current() != OS.WINDOWS) || (options.socket != null);
kernel/src/main/java/org/kframework/parser/inner/disambiguation/TypeInferencer.java:      File NULL = new File(OS.current() == OS.WINDOWS ? "NUL" : "/dev/null");
kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java:    private static final String EXE_EXTENSION = OS.current().equals(OS.WINDOWS) ? ".exe" : "";
kore/src/main/java/org/kframework/utils/OS.java:    OSX(true), LINUX(true), UNKNOWN(false), WINDOWS(false);
kore/src/main/java/org/kframework/utils/OS.java:            return OS.WINDOWS;
kore/src/main/java/org/kframework/utils/OS.java:        if (this == WINDOWS) {
pom.xml:    <skipCheckstyleOnWindows>true</skipCheckstyleOnWindows>
```

```
[dev@dev-ubuntu k]% git grep -i mingw
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java:            //WIN32 fix for line terminator issue: https://sourceforge.net/p/mingw/mailman/message/11374534/
```

```
[dev@dev-ubuntu k]% git grep -i cygwin
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
k-distribution/src/main/scripts/lib/setenv:export CYGWIN=nodosfilewarning
```

```
[dev@dev-ubuntu k]% git grep -i MSYS
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
k-distribution/src/main/scripts/lib/setenv:elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
```

ehildenb:
Note that I still think we should keep the documentation that says how to use WSL for Windows users.

k-distribution/src/main/scripts/bin/k-configure-opam-common
virgil-serbanuta: Mixing spaces with tabs, it seems... Should we pick one of the two (at least, one per language) and convert all sources?

k-distribution/include/kframework/kore
dwightguth: This is a symlink to where the kore prelude is declared which is used in source builds.

k-distribution/src/main/scripts/bin/k-configure-opam-common
ehildenb: No need to convert all sources, we can just make sure we don't introduce new errors and fix ones we find along the way.

k-distribution/tutorial/find-k.mak
dwightguth: These are the cases for if the tutorial is run from its github sources, its maven target directory, or from a debian package, or from mac os.

dwightguth:
I kept most of the references in the code relating to windows because if we ever decide to support windows later, they will end up saving us a bunch of headaches, but I removed all the references to native windows in the documentation (keeping all references to WSL), and removed all references to batch scripts,cmd, mingw, msys, and cygwin.

ehildenb (approved):

PR #1282, opened 2020-05-13
Squash docker image

ehildenb:
This should reduce the size of the resulting docker images by squashing them after building them.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -129,7 +129,7 @@ pipeline {
                     sh '''
                         mv bionic/kframework_${VERSION}_amd64.deb kframework_amd64_bionic.deb
                         docker login --username "${DOCKERHUB_TOKEN_USR}" --password "${DOCKERHUB_TOKEN_PSW}"
-                        docker image build . --file package/docker/Dockerfile.ubuntu-bionic --tag "${DOCKERHUB_REPO}:${BIONIC_COMMIT_TAG}"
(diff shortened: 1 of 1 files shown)


ehildenb:
@dwightguth I am testing locally with `--squash` to see the difference in image size:

```
runtimeverificationinc/kframework-k   squash              4d417e7d9836        4 seconds ago       2.1GB
runtimeverificationinc/kframework-k   no-squash           8a746df1db36        23 hours ago        2.18GB
```

Build commands:

```
docker image build --squash --file package/docker/Dockerfile.ubuntu-bionic --tag runtimeverificationinc/kframework-k:squash .
docker image build --file package/docker/Dockerfile.ubuntu-bionic --tag runtimeverificationinc/kframework-k:no-squash
```

It would appear that we don't really get much benefit from squashing intermediate layers, so I don't think it's worth it to try and reduce the docker image size. Note that to enable `--squash` we have to add:

```
{
    "experimental": true
}
```

to our `/etc/docker/daemon.json` on each of the slave machines. I think this shows that the benefit we get is pretty minimal though, so I don't think it's worth it.

If you agree, go ahead and close this one, or I can investigate more.

dwightguth:
Can we try to measure the size without --squash but with a separate stage that adds just the files that are needed? I'm curious whether it's any smaller.

ehildenb:
I haven't had time to investigate this and likely won't in the near future. I'm pretty sure `--squash` is achieving what we try with the multi-stage build anyway, but don't have time to investigate more and we've already seen that the 2GB download is much faster on CI than building dependencies every time.

PR #1287, opened 2020-05-14
Changes to K frontend to support new C++ unparsing

dwightguth:
This is not ready to be merged yet, but feel free to review it as a companion to kframework/llvm-backend#334.


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -198,6 +197,26 @@ public String convert(boolean heatCoolEq, StringBuilder sb) {
         }
         translateSymbols(attributes, functionRules, impurities, overloads, sb);
 
+        // print syntax definition
+        int length = sb.length();
+        for (Tuple2<Sort, scala.collection.immutable.List<Production>> sort : iterable(module.bracketProductionsFor())) {
+            for (Production prod : iterable(sort._2())) {
+                translateSymbol(attributes, functionRules, impurities, overloads, prod.att().get("bracketLabel", KLabel.class), prod, sb);
+            }
+            for (Production prod : iterable(module.sortedProductions())) {
```
virgil-serbanuta: Why is this `for` loop inside the above one? It does not seem to depend on `sort`, which, I guess, would mean that `genSubsortAxiom` below generates multiple identical axioms (assuming that the `bracketProductionsFor` can have more than one entry).

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1452,17 +1404,62 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
...
+                colors.append(conn).append(att.get("color"));
+                conn = ",";
+              }
+              if (format.charAt(i) == '%') {
+                escape = true;
+              } else {
+                escape = false;
+              }
+            }
+            att = att.add("colors", colors.toString());
```
virgil-serbanuta, line 1399: It would be nice if these new attributes would be documented somewhere.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: good catch. I did not notice that.

virgil-serbanuta (approved):

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1416,17 +1370,62 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
         String format = att.getOptional("format").orElse(Formatter.defaultFormat(prod.items().size()));
         int nt = 1;
         boolean hasFormat = true;
+        int terminals = 0;
```
ehildenb: I don't see this used anywhere. It's just incremented in the below loop, but then the value doesn't seem to be used.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -356,87 +361,34 @@ private void translateSymbols(Map<String, Boolean> attributes, SetMultimap<KLabe
...
+        for (NonTerminal nt : iterable(prod.nonterminals())) {
+            Sort sort = nt.sort();
+            sb.append(conn);
+            convert(sort, prod, sb);
+            conn = ", ";
         }
+        sb.append(") : ");
+        convert(prod.sort(), prod, sb);
+        sb.append(" ");
+        Att koreAtt = addKoreAttributes(prod, functionRules, impurities, overloads);
```
ehildenb, line 387: So now instead of writing separate files with the unparsing information, we're adding attributes to the generated Kore file?

I like the approach more, just want to make sure I understand.

dwightguth: I ended up needing extra information to support the bracket pass, so I decided to reevaluate the approach. The only reason I had used separate files originally was because it proved to be significantly faster. But I was able to optimize the code sufficiently that it became less of an issue, which is why I went back to my original idea, which was the kore file attributes.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ehildenb: Seconded. I see that `color`, `colors`, `format` are already documented in `pending-documentation.md`.

`priorities`, `left`, `right` are not, please add these to `pending-documentation.md`. I also see that `bracket` is folded into the paremetric productions section of `pending-documentation.md`, maybe it should be made into it's own section though?

I also see `terminals` here as an attribute and `nat`, but I don't think that's something we want the users setting, so maybe we shouldn't document that. If these are supposed to be user exposed, also add to `pending-documentation.md`. If not, then maybe we should either (i) make sure they don't exist in the attributes before adding them in, or (ii) have a check in the frontend that they do not occur.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ehildenb: @dwightguth I still don't see any uses of this variable.

ehildenb (approved):

PR #1292, opened 2020-05-16
Upstream symbolic integer reasoning

ehildenb:
In KEVM and KWasm we often use the same body of lemmas which do arithmetic normalization/simplification for the Haskell backend.

This adds these lemmas to `domains.k` in the `INT-SYMBOLIC` module, and sets up a testing harness for these lemmas, both with and without SMT solver enabled, for these lemmas (and any other lemmas in domains.k we want to test in the future).

The arithmetic simplification lemmas come from KEVM directly, and were supplied by @ana-pantilie . The tests come from there as well, and I added a few.

I'm making this PR because now I need these same lemmas in KWasm.


daejunpark:
Why do you need tests running with `--smt none`?

ehildenb:
@daejunpark if you run a test like `(X +Int 1) +Int 1 => X +Int 2` with SMT turned on, you're not sure that you're testing that the actual K term gets simplified, because the SMT solver can just tell you that `X + 1 + 1 == X + 2`.

So it's to make sure that the K rules themselves are what control the simplification.

daejunpark (approved):

daejunpark:
Hold on. Is it possible to not include some of these lemmas? Not all of these are always needed and necessarily useful. Sometimes we will need to exclude some of them.

daejunpark:
If it is not possible, I think at least we will need to have IMP-SYMBOLIC _not_ be imported to IMP by default.

daejunpark (dismissed):
Need a way to exclude (some of) these simplification rules.

ehildenb:
@daejunpark can you please provide a concrete example?

ehildenb:
These lemmas seem universally useful to me.

daejunpark:
I cannot recall an interesting concrete example, but generally, these
lemmas (esp. associativity ones) are by no means complete, but ad-hoc just
enough for some known cases. Some users may want to build their own
simplification rule set, in which case premature application of these
lemmas may disrupt it and they may want to start from scratch. (The rule
priority may help in that case, but lower priority rules can still be
applied to sub-terms, so it could be hard to control.)

ehildenb:
@daejunpark I don't think that's a good enough reason to block this. These lemmas are generally useful, and I made sure to not include ones which are specific to situations in KEVM or KWasm, just generic reasoning lemmas which reduce term size.

If you can find a specific lemma here which could be problematic, let me know, but AFAICT all of these are stricly simplifying the term. I would rather get it merged, then wait for examples from several semantics to come back and say that they are replacing this module all the time with a different one to re-asses.

ehildenb:
If you want, we can pull it out into `INT-SIMPLIFICATION`, but I do think these lemmas are generally useful across the board, so it makes sense to me to include in `INT-SYMBOLIC`. Note that `INT-SYMBOLIC` is not sent to the Java backend.

daejunpark:
I think it is better to pull it out into `INT-SIMPLIFICATION`, as you suggested. The only overhead is to add a single extra line of `imports INT-SIMPLIFICATION`.

Strictly speaking, `rule I +Int B => B +Int I` and `rule A -Int I => A +Int (0 -Int I)` do not reduce the term size. It normalizes terms in a specific way, which I'm not convinced to be _universally_ useful. (Indeed, it was inspired by common simplification tasks of KEVM verification. I do remember when they were first introduced by @denis-bogdanas.) I'd rather wait until we have sufficient evidence and requests from other semantics that they are useful and should be included by default. (I don't think KEVM and KWASM alone are representative enough.)

Also, as I mentioned, the following simplification rules over associativity is not complete. For example, it cannot simplify `(X +Int I) -Int X`, where `X` is symbolic and `I` is concrete, which I found recently in the newer version of Solidity-compiled bytecode. I'm not positive that it would scale to just add more rules whenever we found new cases. (So far, most of these rules were added to simplify the gas expression, of a specific form.) 
```
  rule (A +Int I2) +Int I3 => A +Int (I2 +Int I3) [concrete(I2, I3), symbolic(A), simplification]
  rule I1 +Int (B +Int I3) => B +Int (I1 +Int I3) [concrete(I1, I3), symbolic(B), simplification]
  rule I1 -Int (B +Int I3) => (I1 -Int I3) -Int B [concrete(I1, I3), symbolic(B), simplification]
  rule I1 +Int (I2 +Int C) => (I1 +Int I2) +Int C [concrete(I1, I2), symbolic(C), simplification]
  rule I1 +Int (I2 -Int C) => (I1 +Int I2) -Int C [concrete(I1, I2), symbolic(C), simplification]
  rule (I1 -Int B) +Int I3 => (I1 +Int I3) -Int B [concrete(I1, I3), symbolic(B), simplification]
  rule I1 -Int (I2 +Int C) => (I1 -Int I2) -Int C [concrete(I1, I2), symbolic(C), simplification]
  rule I1 -Int (I2 -Int C) => (I1 -Int I2) +Int C [concrete(I1, I2), symbolic(C), simplification]
  rule (C -Int I2) -Int I3 => C -Int (I2 +Int I3) [concrete(I2, I3), symbolic(C), simplification]

  rule I1 &Int (I2 &Int C) => (I1 &Int I2) &Int C [concrete(I1, I2), symbolic(C), simplification]
```

So far, we have been very strict when adding something in the default domain modules, making sure that they are comprehensive. These simplification rules are a good start, but not comprehensive yet to me.

I guess it is fine to add some trivial simplification rules (i.e., idempotent rules) in the `INT-SYMBOLIC` module, like:
```
  rule I +Int 0 => I [simplification]
  rule I -Int 0 => I [simplification]

  rule X modInt N => X requires 0 <=Int X andBool X <Int N [simplification]
  rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]

  rule X <<Int 0 => X [simplification]
  rule 0 <<Int X => 0 [simplification]
  rule X >>Int 0 => X [simplification]
  rule 0 >>Int X => 0 [simplification]
```
but I'd not be comfortable for the other rules that are quite "directional".

daejunpark:
That's my thought. But since it is quite subjective to judge whether these are universally useful, at this point without having sufficient examples and counter-examples, I'm OK to merge if most of the others are positive about these.

daejunpark (commented):
I lift the request change to just a comment:
> Need a way to exclude (some of) these simplification rules.

ehildenb:
@daejunpark I've added a test of `(X +Int I) -Int X` to the domains-no-smt test, and it seems to pass. I'm surprised that it passes too actually, so now I'm wondering how the Haskell backend is able to do it. None of the rules should apply because of the `symbolic` and `concrete` attributes.

@ttuegel is there a way we can see which function rules applied to simplify these claims?

dwightguth (approved):
This PR makes sense to me, but as always with PRs relating to symbolic reasoning, please get Tom's approval also.

ehildenb:
@ttuegel can you take a look?

k-distribution/include/builtin/domains.k
```diff
@@ -444,39 +444,40 @@ module INT-SYMBOLIC [symbolic, kore]
...
-    #Ceil(@I1:Int <<Int @I2:Int)
-  =>
-    {(@I2 >=Int 0) #Equals true} #And #Ceil(@I1) #And #Ceil(@I2)
-  [anywhere]
+  // Definability Conditions
+  rule #Ceil(@I1:Int /Int   @I2:Int) => {(@I2 =/=Int 0) #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) [anywhere]
+  rule #Ceil(@I1:Int %Int   @I2:Int) => {(@I2 =/=Int 0) #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) [anywhere]
+  rule #Ceil(@I1:Int modInt @I2:Int) => {(@I2 =/=Int 0) #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) [anywhere]
+  rule #Ceil(@I1:Int >>Int  @I2:Int) => {(@I2 >=Int 0)  #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) [anywhere]
+  rule #Ceil(@I1:Int <<Int  @I2:Int) => {(@I2 >=Int 0)  #Equals true} #And #Ceil(@I1) #And #Ceil(@I2) [anywhere]
```
ttuegel: These must have attribute `simplification`. I don't think `anywhere` is necessary: it should have no effect because `#Ceil` isn't a constructor. However, it shouldn't hurt anything.

k-distribution/include/builtin/domains.k
```diff
@@ -444,39 +444,40 @@ module INT-SYMBOLIC [symbolic, kore]
...
+
+  // Arithmetic Normalization
+  rule I +Int 0 => I [simplification]
+  rule I -Int 0 => I [simplification]
+
+  rule I +Int B => B +Int I          [concrete(I), symbolic(B), priority(51), simplification]
+  rule A -Int I => A +Int (0 -Int I) [concrete(I), symbolic(A), priority(51), simplification]
```
ttuegel: What is this `priority` attribute intended to do?

k-distribution/include/builtin/domains.k
ehildenb: I've incorporated the update #1305 so this shouldn't be an issue anymore.

k-distribution/include/builtin/domains.k
ehildenb: It's to make sure the rules preceding it are applied first:

```
  rule I +Int 0 => I [simplification]
  rule I -Int 0 => I [simplification]

  rule I +Int B => B +Int I          [concrete(I), symbolic(B), priority(51), simplification]
  rule A -Int I => A +Int (0 -Int I) [concrete(I), symbolic(A), priority(51), simplification]
```

For example, if you have `X -Int 0`, then you want it to go directly to `X` (and not to `X +Int (0 -Int 0)`).

ttuegel (approved):

PR #1304, opened 2020-05-26
A few minor features

dwightguth:
This PR has several small, independent changes that it seemed to me like it would be overkill to make into 3 different PRs, but if you want me to, let me know.

Roughly:

1. Add some code to the k include directory for calling some hooks implemented in the llvm backend
2. Fix a bug where adding injections to context sentences caused a crash, by disabling this functionality since it is not actually needed.
3. Add some new attributes relating to evaluation strategy which allow finer-tuned control.

### DIFF
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -401,7 +401,7 @@ rule isKResult(BE1:BExp || BE2:BExp) => true requires isKResult(BE1)
 ```
 
 Sometimes you wish to declare a production hybrid with respect to a predicate
-other than isKResult. You can do this by specifying a sort as the body of the
+other than `isKResult`. You can do this by specifying a sort as the body of the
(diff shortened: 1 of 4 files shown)


pending-documentation.md
```diff
@@ -400,6 +400,20 @@ being added to the definition of `isKResult`:
 rule isKResult(BE1:BExp || BE2:BExp) => true requires isKResult(BE1)
 ```
 
+Sometimes you wish to declare a production hybrid with respect to a predicate
+other than isKResult. You can do this by specifying a sort as the body of the
```
```suggestion
other than `isKResult`. You can do this by specifying a sort as the body of the
```

ehildenb (approved):
@dwightguth approved with one minor formatting suggestion.

Issue #1296, opened 2020-05-19
Warn about unused productions and unused matches at kompile time


hjorthjort:
@dwightguth came up with a suggestion for how to detect certain classes of mistakes at `kompile` time.

An unused production in a program is likely unintended. For example, a cell may become redundant in a semantics and fall out of use, but not get removed from the configuration.

Similarly, a variable that is used only once in a rule is likely a mistake. It may have fallen out of use due to refactoring, or reflect that the developer forgot to make some RHS update.

# Suggestion

At `kompile` time:

- [ ] Give a warning if some production is declared, but not mentiond in any rule
- [x] Give a warning if a variable is used exactly once in a rule. Possibly do not warn if the variable starts with `_`

# Example from elsewhere

Haskell will warn you if you have unused LHS variables that do not start with `_`.

```hs
foo a b = b
bar _ b = b
baz _a b = b
```

```
$ ghci -Wunused-matches bad.hs
...
/home/hjort/tmp/bad.hs:4:5: warning: [-Wunused-matches]
    Defined but not used: ‘a’
  |
1 | foo a b = b
```

No warning for `bar` and `baz`.

hjorthjort:
User story for this:

Had a bug creep in during refactoring. Iterating over a list applying changes, I forgot to carry the `DS` variable over to the right-hand-side. Cost about 10 mins to find this fix:

```
     rule #unfoldDefns(( table EXPO:InlineExport SPEC:TableSpec ) DS, I)
-      => #unfoldDefns(( table #freshId(I) EXPO SPEC ), I +Int 1)
+      => #unfoldDefns(( table #freshId(I) EXPO SPEC ) DS, I +Int 1)
```

PR #1310, opened 2020-05-28
Warn if non-anonymous variable appears only once.

dwightguth:
Fixes part of #1296.

### DIFF
--- a/k-distribution/tests/regression-new/checkWarns/checkUnusedVar.k
+++ b/k-distribution/tests/regression-new/checkWarns/checkUnusedVar.k
@@ -1,0 +1,13 @@
+module CHECKUNUSEDVAR-SYNTAX
+
+endmodule
+
+module CHECKUNUSEDVAR
+  imports INT
+
+  syntax KItem ::= foo(Int)
+
+  rule foo(X) => 0
+  rule foo(0) => !_:Int
+
+endmodule
--- a/k-distribution/tests/regression-new/checkWarns/checkUnusedVar.k
+++ b/k-distribution/tests/regression-new/checkWarns/checkUnusedVar.k
@@ -2,12 +2,20 @@ module CHECKUNUSEDVAR-SYNTAX
 
 endmodule
 
+module MOD
+  imports INT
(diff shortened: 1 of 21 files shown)


k-distribution/tests/regression-new/checkWarns/checkUnusedVar.k
```diff
@@ -0,0 +1,21 @@
...
+endmodule
+
+module CHECKUNUSEDVAR
+  imports INT
+  imports MOD
+
+  syntax KItem ::= foo(Int)
+
+  rule foo(X) => 0
+  rule foo(0) => !_:Int
```
ehildenb, line 17: Please add some tests that also include requires clauses, and ensures clauses.

Specifically, add a test which defines the variable in the LHS, but only uses in the requires clause or ensures clause (and make sure no warning is output).

Also add one which has a trivial requires clause which does not use the variable.

kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -202,11 +209,31 @@ public void apply(KToken k) {
...
+                            if (initializerProduction.isDefined()) {
+                                Set<Production> realProds = stream(initializerProduction.get())
+                                        .filter(p -> !p.att().contains(Att.RECORD_PRD(), Production.class))
+                                        .collect(Collections.toSet());
+                                if (realProds.size() == 1) { // should be only a single initializer
+                                    if (realProds.head().items().size() == 4) {
+                                      hasConfigVar = true;
+                                    }
+                                }
+                            }
```
ehildenb, line 231: What is this change going on here? Is it related to the variable declaration issues?

Or is it related to the new issue @hjorthjort posted #1318.

dwightguth: This is fixing a false positive that arose because we were generating a rule during parsing that had an unused variable.

ehildenb (approved):

PR #1313, opened 2020-05-30
Bundle opam 2.0 with docker image

ehildenb:
Currently we're installing the default version of opam with Ubuntu docker image, but it's 1.2 version. We should be using 2.0 because more projects rely on it (in this example, Tezos), and we don't have any projects relying on v1.2.


ehildenb:
@dwightguth I'm not sure this one is worth it, because the Tezos dockerfile is already forcibly putting in opam 2.0 anyway and we'll be removing our dependency on OCaml soon anyway.

Your call, feel free to close and delete this branch if you don't think so.

PR #1327, opened 2020-06-09
generalize strictness to multiple aliases

dwightguth:
This is a feature I added to make the new C semantics be able to express its evaluation order entirely using strictness.

Note that included here is also the fix for the infamous "attributes are wrong after modifying them but not changing the body of a rule" bug, which turned out to be much easier to fix than I thought it would be. It was actually causing the test I wrote to fail!


kernel/src/main/java/org/kframework/compile/ResolveStrict.java
```diff
@@ -193,6 +159,55 @@ public K apply(KVariable var) {
...
+        }
+        if (attribute.isEmpty()) {
+            for (int i = 1; i <= arity; i++) {
+                strictnessPositions.add(i);
+            }
+            aliases.add(DEFAULT_ALIAS);
+            resolve(sequential, sentences, arity, strictnessPositions, allPositions, aliases, production);
+            allPositions.addAll(strictnessPositions);
+        } else {
+            String[] components = attribute.split(";");
```
ehildenb, line 184: Do we have a wrapper for 1-dimensional attribute arrays? I know it would just be `attribute.split(";")` still, but maybe it would be good to call that function if it exists for semantic readability.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -487,21 +489,17 @@ private boolean equalsSyntax(Module _this, Module that) {
...
+                    if (kem.options.warnings.includesExceptionType(err.exception.getType())) {
+                        errors.add(KEMException.asError(err));
                     }
-                } else
-                    kem.addAllKException(parse.getWarnings().stream().map(e -> e.getKException()).collect(Collectors.toList()));
-                return Stream.of(parse.getParse());
-            }
+                }
+            } else
+                kem.addAllKException(parse.getWarnings().stream().map(e -> e.getKException()).collect(Collectors.toList()));
```
ehildenb:
```suggestion
            } else {
                kem.addAllKException(parse.getWarnings().stream().map(e -> e.getKException()).collect(Collectors.toList()));
            }
```

kernel/src/main/java/org/kframework/compile/ResolveStrict.java
```diff
@@ -193,6 +159,55 @@ public K apply(KVariable var) {
...
+                for (int i = 0; i < components.length; i+=2) {
+                    setAliases(components[i].trim(), aliases, production);
+                    setPositions(components[i+1].trim(), strictnessPositions, arity, production);
+                    resolve(sequential, sentences, arity, strictnessPositions, allPositions, aliases, production);
+                    aliases.clear();
+                    allPositions.addAll(strictnessPositions);
+                    strictnessPositions.clear();
+                }
+            } else {
+                throw KEMException.compilerError("Invalid strict attribute containing multiple semicolons.", production);
```
radumereuta, line 207: This error message doesn't really reflect the code, let alone the user input.
See my other comment.

pending-documentation.md
```diff
@@ -459,6 +459,20 @@ each argument of the production, whereas `strict(c; 1)` will instantiate it
 only for the first argument. The special variable `HERE` is used to tell the
 compiler where you want to place the production that is to be heated or cooled.
 
+You can also specify multiple context aliases for different parts of a production,
+for example:
+
+```k
+syntax Exp ::= foo(Exp, Exp) [strict(left; 1; right; 2)]
```
radumereuta, line 466:
```
syntax Exp ::= foo(Exp, Exp) [strict(1(left); 2(right)]
```
Wouldn't this be more intuitive? In this case, you might need to add a new parsing entry in `Outer.jj`, but it shouldn't be that hard.

ehildenb: Seconded, I was wondering why we aren't using the 2D array for this actually, `left, 1; right, 2` (since that is functionality we already support)

kernel/src/main/java/org/kframework/compile/ResolveStrict.java
dwightguth: We cannot call that function because this is actually a 2d attribute array. Calliing the 2d function could theoretically work, but it would require substantially refactoring this method, so I'd rather just leave the code like this.

pending-documentation.md
dwightguth: @ehildenb it's because `strict(left; 1, 2, 3)` is the existing syntax for a single alias label. Commas separate strictness positions, semicolons separate labels from positions and delimit multiple labels. We could have done what you describe previously, but it would probably not be backwards compatible at this point.

dwightguth: As for the other syntax, it is inconsistent with how we specify attributes elsewhere and would probably require a grammar to parse, which I would prefer to avoid when we can use a regular language instead.

radumereuta: In what way they are inconsistent? I may misunderstand what you are trying to do.
If by creating a parser you get better results, I think it's worth it. The JJTree part of JavaCC makes it quite easy.

ehildenb: I agree with @dwightguth that it's inconsistent, in other places where we need lists or 2D lists of data we always use `_;_` and `_;_,_,_;_`. If we want improved ways of specifying structured data in attributes, that can be another PR.

k-distribution/tests/regression-new/context-alias-2/test.k
```diff
@@ -0,0 +1,24 @@
...
+  imports INT
+  imports ID
+  imports MAP
+
+  syntax Exp ::= foo(Exp, Exp) [seqstrict(left; 1; right; 2)]
+               | Id "=" Int
+               | Id
+               | Int
+               | l(Exp) [symbol] | r(Exp) [symbol]
+  syntax KResult ::= Int
```
ehildenb: Please write some more tests:

-   One which uses 3 context aliases (eg `left`, `right`, and `center`)
-   One which uses the same alias for multiple argument positions: (eg `left; 1, 3; right; 2`)

ehildenb (approved):

ehildenb (approved):

PR #1329, opened 2020-06-11
scala version bump

dwightguth:
I have been wanting to update to a more recent version of scala for a while now, but whenever I try scala 2.13, I get a bunch of compiler errors. I did not realize that scala 2.12 was still being updated with bugfix releases, however, and this is just as good of a solution. Thus, we update to the latest release of scala 2.12.


dwightguth:
Encountering unexpected problems here involving serialization. I will look into it at some point, but it's not an urgent enough change for me to want to spend more time on it.

PR #1338, opened 2020-06-12
Replacement for krun

dwightguth:
This is not ready for full production use yet; in particular it does not support the haskell backend. However, it is designed to be an enormously faster drop-in replacement for krun when doing concrete execution on the llvm backend. I have modified the test harness for CI to use it everywhere the llvm backend is used, so it's reasonably well tested. But I suggest people try it out and report issues, because it is vaaaaaastly faster than krun. It still relies on the java implementation of `kast` if you have not implemented a bison parser or if you request more advanced output modes, but everything else, including unparsing, is done using tools that have essentially no overhead. As a result, I can do something like:

```
kompile --gen-bison-parser foo.k
kx pgm.foo
```

On most definitions that include an LR(1) syntax module, and get a pretty-printed output configuration with less than 100ms of overhead.

Eventually this will replace the java implementation of krun entirely, but feature parity isn't yet where we want it to be.


k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,438 @@
+#!/bin/bash
+set -e
```
ehildenb, line 2: I think we should `set -eo pipefail` at least, so any commands added in pipes also fail the script if they fail (instead of just the last command in the pipe).

We probably also should do `set -euo pipefail`, so that we fail on unitialized variables (and so changes to variables here must provide defaults), but that may be trickier.

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,438 @@
...
+  shift
+done
+
+if $search; then
+  io=false
+fi
+
+if [[ "${#ARGV[@]}" -gt 1 ]]; then
+  error 'Cannot pass more than one configuration variable as a positional argument.'
+fi
```
ehildenb: Just a styling argument, any single-line `if CONDITION; then COMMAND; fi` can be rewritten as `! CONDITION || COMMAND`. Up to you.

dwightguth:
There are no pipes in the current script, but I'm happy to future proof it by adding `set -o pipefail`. I will try out `set -u` tomorrow, but I can't guarantee it will work, necessarily. We are trying to maintain support for bash 3, which doesn't have associative arrays, so we implement associative arrays with string keys via a prefixed domain of variables; a variable being undefined thus specifically means that that key is not in the map, so there may end up being some difficulty there. But I'll try it.

kernel/src/main/java/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -68,6 +68,7 @@ public int run() {
         Kompile kompile = new Kompile(options, files.get(), kem, sw, !options.profileRules);
         Backend backend = koreBackend.get();
         CompiledDefinition def = kompile.run(options.outerParsing.mainDefinitionFile(files.get()), options.mainModule(files.get()), options.syntaxModule(files.get()), backend.steps(), backend.excludedModuleTags());
+        files.get().saveToKompiled("mainModule.txt", def.executionModule().name());
         sw.printIntermediate("Kompile to kore");
```
ehildenb, line 72: This information isn't already saved somewhere else?

dwightguth: It is not saved anywhere in any place that is convenient for a shell script to parse. Mostly the java code gets this information from the deserialized K definition, which we are not making use of in the new script.

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,442 @@
...
+      --output-file FILE   Print final configuration to FILE
+      --parser VALUE       Use VALUE as parser to parse \$PGM. For example,
+                           if the user says "$KX --parser cat foo.kore", then
+                           \`cat foo.kore\` is invoked and the result on stdout
+                           is used as the value of \$PGM as a KORE term.
+      --pattern PAT        Use PAT as the search pattern with --search. Can
+                           also be used without --search to match the output
+                           configuration against a pattern and print the
+                           substitution. Only supported on Haskell backend.
+      -save-temps          Do not delete temporary files when $KX terminates.
```
ehildenb:
```suggestion
      --save-temps         Do not delete temporary files when $KX terminates.
```

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,442 @@
...
+        on)
+        color=true
+        ;;
+  
+        off)
+        color=false
+        ;;
+  
+        *)
+        error 'Invalid value for --color. Should be "on" or "off".'
```
ehildenb, line 160: I don't see the definition of `error` anywhere...

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,442 @@
...
+  
+      -v|--verbose)
+      verbose=true
+      flags="$flags -v"
+      ;;
+  
+      -save-temps)
+      trap - INT TERM EXIT
+      flags="$flags -save-temps"
+      ;;
```
ehildenb:
```suggestion
      --save-temps)
      trap - INT TERM EXIT
      flags="$flags -save-temps"
      ;;
```

Why do we have just a single dash here?

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,442 @@
...
+    else
+      unescaped=$(</dev/stdin)
+      stdinText=$(awk 'BEGIN {for(n=0;n<256;n++)ord[sprintf("%c",n)]=n} {text = text $0 ORS} END {split(text, chars, ""); for (i=1; i<=length(text); i++) { if (chars[i] == "\"") { printf("%s", "\\\"") } else if (chars[i] == "\\") { printf("%s", "\\\\") } else if (chars[i] == "\n") { printf("%s", "\\n") } else if (chars[i] == "\t") { printf("%s", "\\t") } else if (chars[i] == "\r") { printf("%s", "\\r") } else if (chars[i] == "\f") { printf("%s", "\\f") } else if (ord[chars[i]] >= 32 && ord[chars[i]] < 127) { printf("%s", chars[i]) } else { printf("\\x%02x", ord[chars[i]]) } } }' <<< "$unescaped")
+    fi
+    stdinFile="$(mktemp tmp.stdin.XXXXXXXXXX)"
+    tempFiles+=("$stdinFile")
+    printf '\dv{SortString{}}(\"%s\")\n' "$stdinText" > "$stdinFile"
+    configVars="$configVars -c STDIN $stdinFile String korefile"
+  fi
+  execute llvm-krun $configVars -d "$kompiledDir" $flags --dry-run -o "$input_file"
```
ehildenb, line 430: It looks like `flags` is only intended to be for the llvm backend, maybe it should be named `llvm_flags` instead?

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,442 @@
...
+      ;;
+    esac
+  fi
+  shift
+done
+
+if $search; then
+  io=false
+fi
+
```
ehildenb: This should be more like:

```
if $search && $io; then
    error "Cannot specify both --search* and --io on"
fi
```

k-distribution/src/main/scripts/bin/kx
dwightguth: Well, this code is going to be used by the haskell backend code path as well to construct the initial configuration, so I'm not sure that makes sense.

k-distribution/src/main/scripts/bin/kx
```diff
@@ -0,0 +1,456 @@
...
+      ioText=off
+    fi
+    configVars="$configVars -c IO \dv{SortString{}}(\"$ioText\") String kore"
+  fi
+  if [ "$(cat "$kompiledDir"/sort_STDIN.txt 2>/dev/null)" = "String" ]; then
+    if $io; then
+      stdinText=
+    else
+      unescaped=$(</dev/stdin)
+      stdinText=$(awk 'BEGIN {for(n=0;n<256;n++)ord[sprintf("%c",n)]=n} {text = text $0 ORS} END {split(text, chars, ""); for (i=1; i<=length(text); i++) { if (chars[i] == "\"") { printf("%s", "\\\"") } else if (chars[i] == "\\") { printf("%s", "\\\\") } else if (chars[i] == "\n") { printf("%s", "\\n") } else if (chars[i] == "\t") { printf("%s", "\\t") } else if (chars[i] == "\r") { printf("%s", "\\r") } else if (chars[i] == "\f") { printf("%s", "\\f") } else if (ord[chars[i]] >= 32 && ord[chars[i]] < 127) { printf("%s", chars[i]) } else { printf("\\x%02x", ord[chars[i]]) } } }' <<< "$unescaped")
```
ehildenb, line 423: Where is this coming from? Is this a heuristic or definitely correct?

dwightguth: This is the awk script to escape an ascii string into a K string literal.

ehildenb:
@dwightguth there seem to be several places where the formatting is not correct (including a mix of tabs and spaces).

Can you audit the entire `kx` file and fix formatting everywhere? I don't want to have to point out every case. If you don't want to I can do it locally and push up formatting fixes, but I'm fairly opinionated about `case` statements which you may not like. Either way at least indentation should be fixed globally.

dwightguth:
@ehildenb I have fixed all the inconsistencies in formatting and indentation that I have spotted. If there are others you feel need to be addressed, you will need to point them out.

ehildenb (approved):

PR #1340, opened 2020-06-12
Warn if a symbol is declared but not used anywhere

dwightguth:


ehildenb (changes requested):
Can you add a test for a symbol that is _not_ a cell production? All the warnings I see are for productions which are cell productions.

ehildenb (approved):

PR #1342, opened 2020-06-14
Updates to pyk library for processing rule merges

ehildenb:
-   Add `timeout` option to `_teeProcessStdout` so that we can stop the rule merger if it's hanging.
-   Cleans up and adds to boolean predicate simplification when minimizing rules.
-   Pushes down rewrites in KSequences further.
-   Removes trivial requires/ensures clauses when minimizing rules.


k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -84,27 +84,41 @@ def rewriteAnywhereWith(rule, pattern):
...
-                }
-    return replaceKLabels(k, klabelMap)
+    if k is None:
+        return None
+    mlPredToBoolRules = [ (KApply('#True', [])  , KToken('true', 'Bool'))
+                        , (KApply('#False', []) , KToken('false', 'Bool'))
+                        , (KApply('#And'    , [KVariable('#V1'), KVariable('#V2')]) , KApply('_andBool_' , [KVariable('#V1'), KVariable('#V2')]))
+                        , (KApply('#Or'     , [KVariable('#V1'), KVariable('#V2')]) , KApply('_orBool_'  , [KVariable('#V1'), KVariable('#V2')]))
+                        , (KApply('#Not'    , [KVariable('#V1')])                   , KApply('notBool_'  , [KVariable('#V1')]))
+                        , (KApply('#Equals' , [KVariable('#V1'), KVariable('#V2')]) , KApply('_==K_'     , [KVariable('#V1'), KVariable('#V2')]))
```
hjorthjort, line 99: Is this really true? I thought the `#` values were used for matching logic operators and terms, not for the regular boolean operators.

ehildenb: No, it's technically not true, `#Equals` is just `#Equals`, not `_==K_`. But `_==K_` usually gets translated to `#Equals` for the Haskell backend, so this just translates it back for printing out.

We could make it more specific, but saying that `#Equals` is only translated back for some specific cases, but it's a lot of cases that we would have to list.

hjorthjort: My only worry is that we'll get a rule that doesn't work because `#And` and `andBool` work differently, for example, where one expresses matching over sets. And if an `#And` is introduced for ML specific purposes it would be very wrong to translate it back. So I'm a bit skeptical. I would rather have it behind a flag or option or something, since it's a little bit dangerous I don't think it should be the default.

ehildenb: I can rename the function `mlPredToBool => mlPredToBoolUnsafe` if you want?

hjorthjort: Sure, that would do it. And maybe add a docstring explaining why it's unsafe: all boolean operators translate to the ML operators, but not vice versa.

hjorthjort (approved):
LGTM with the `unsafe` name.

k-distribution/src/main/scripts/lib/pyk/kastManip.py
ehildenb: Done, I've made the renaming and some documentation.

PR #1346, opened 2020-06-16
don't create scanner for STDIN-STREAM and STDOUT-STREAM if they are not

dwightguth:
needed

This ought to significantly decrease incremental parsing time on semantics that do not use these modules.


dwightguth:
This didn't really seem to substantially improve the performance of parsing, so I'm going to close it.

PR #1349, opened 2020-06-17
Add ways to suppress unused var and symbol warnings

### LINKED ISSUE #1296: Warn about unused productions and unused matches at kompile time

dwightguth:
Fixes #1296


k-distribution/tests/regression-new/checkWarns/checkUnusedSymbol.k.out
```diff
@@ -1,5 +1,5 @@
 [Error] Compiler: Had 1 structural errors.
 [Error] Compiler: Symbol 'foo(_)_CHECKUNUSEDSYMBOL_Foo_Int' defined but not
-used.
+used. Add the 'unused' attribute if this is intentional.
 	Source(checkUnusedSymbol.k)
```
ehildenb, line 4: Add a test where we _do_ have the `unused` attribute on a klabel that is _not_ a cell klabel, and it is unused (and there should be no warning).

ehildenb (changes requested):
Add the test I ask for, and also add documentation to `pending-documentation.md` for this new attribute.

pending-documentation.md
```diff
@@ -185,6 +185,18 @@ have any constructors declared.
 syntax Bytes [hook(BYTES.Bytes), token]
 ```
 
+### `unused` attribute
+
+K will warn you if you declare a symbol that is not used in any of the rules of your
+definition. Sometimes this is intentional, however; In this case, you can suppress the
```
```suggestion
definition. Sometimes this is intentional, however; in this case, you can suppress the
```

ehildenb (approved):

PR #1351, opened 2020-06-17
Fix bug involving record production on rhs

### LINKED ISSUE #1318: Error for unbound variable doesn't specify source location
The following `kompile` error should report the offending line if possible.

```
[Error] Compiler: Found variable _ on right hand side of rule, not bound on left hand side. Did you mean "?_"?
```

Command:
```
kompile --backend llvm                                                    \
    --directory .build/defn/llvm -I .build/defn/llvm                                \
    --main-module WASM-TEST --syntax-module WASM-TEST-SYNTAX .build/defn/llvm/test.k
```

dwightguth:
Fixes #1318

### DIFF
--- a/kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
+++ b/kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
@@ -3,7 +3,7 @@ package org.kframework.parser
 import java.util
 import java.util.Optional
 
-import org.kframework.attributes.{Att,Location,Source}
+import org.kframework.attributes.{Att,HasLocation,Location,Source}
 import org.kframework.builtin.Sorts
 import org.kframework.definition.{NonTerminal, Production}
 import org.kframework.{kore => k}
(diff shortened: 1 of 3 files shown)


kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
```diff
@@ -24,18 +24,18 @@ class TreeNodesToKORE(parseSort: java.util.function.Function[String, Sort], stri
...
 
   def termConsToKApply(t: TermCons): K = {
     if (t.production.att.contains("recordPrd", classOf[Production])) {
       val realProd = t.production.att.get("recordPrd", classOf[Production])
       val map = new util.ArrayList(t.items).asScala.reverse.zipWithIndex.map { case (item, idx) => (t.production.nonterminal(idx).name.get, apply(item))} toMap
       val realItems = realProd.nonterminals.map {
-        case NonTerminal(sort, None) => anonVar(sort)
-        case NonTerminal(sort, Some(x)) => map.getOrElse(x, anonVar(sort))
+        case NonTerminal(sort, None) => anonVar(sort, t)
+        case NonTerminal(sort, Some(x)) => map.getOrElse(x, anonVar(sort, t))
```
ehildenb: It's a little confusing to me that you're passing the entire `t: Term` into `anonVar`, because why would it need a term? Are you able to pass just the `Location` in for it to use? I'm approving anyway, but something to think about.

ehildenb (approved):

kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
dwightguth: I changed the signature of the method to t: HasLocation to make it clearer what is happening.

PR #1357, opened 2020-06-19
change how kseq is pretty printed in llvm and haskell backends 

dwightguth:
This pulls as much of the output and test suite changes of #1338 into its own PR as possible while keeping out as much of the code changes as possible. The remaining changes to the output files and test suites are directly as a result of the changes surrounding the new execution tool. These changes focus solely on how unparsing works and on removing the regression tests specific to the ocaml backend.

The reason I deleted those tests is that there is only one semantics remaining using the ocaml backend, the C semantics, and even this is not in active development, as a replacement using the llvm backend is being built from the ground up. The only reason I have not yet removed the ocaml backend entirely from the code base is because I don't want to completely cut the support tether between the old C semantics and the K frontend, but I don't intend for us to do any further work on the ocaml backend if it can be at all helped. Thus, since the changes to output files were interfering with the ocaml backend, I simply removed these tests. Note that we are still testing the ocaml backend via the smoke test associated with the debian packages.


k-distribution/tutorial/1_k/3_lambda++/lesson_1/exercises/from-call-CC-to-callcc/Makefile
```diff
@@ -1,4 +1,3 @@
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 include $(MAKEFILE_PATH)/../../Makefile
 KOMPILE_FLAGS=--main-module CALLCC
-KOMPILE_BACKEND?=java
```
ehildenb, line 4: Why was this removed? Are we using LLVM backend for this one now intentionally?

k-distribution/tutorial/1_k/3_lambda++/lesson_1/exercises/from-callcc-to-call-CC/Makefile
```diff
@@ -1,4 +1,3 @@
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 include $(MAKEFILE_PATH)/../callCC/Makefile
 KOMPILE_FLAGS=--main-module CALLCC --syntax-module CALLCC-SYNTAX
-KOMPILE_BACKEND?=java
```
ehildenb, line 4: Intentional?

k-distribution/tutorial/1_k/3_lambda++/lesson_1/exercises/from-call-CC-to-callcc/Makefile
dwightguth: It was already using the llvm backend, because of a bug in the way the makefile was written. Since the underlying lesson also uses the llvm backend, I simply made this explicit.

ehildenb (approved):

PR #1378, opened 2020-06-25
Additional documentation for: cell fragments, the `prec` attribute, `#Layout` and converting between token sorts.

### LINKED ISSUE #1390: Request: Documentation for #Layout
I looked through `pending-document.md` and couldn't find an entry for `#Layout`.

nishantjr:
Fixes: #1390


pending-documentation.md
```diff
@@ -185,6 +185,99 @@ have any constructors declared.
...
 
+### Converting between `[token]` sorts
+
+You can convert between tokens of one sort via `String`s by defining functions
+implemented by builtin hooks.
+The hook `STRING.token2string` allows conversion of any token to a string:
+
+```k
+syntax String ::= FooToString(Foo)  [function, functional, hook(STRING.token2string)]
+```
```
ehildenb, line 196: Should this be indented consistently with the other code blocks?

pending-documentation.md
```diff
@@ -185,6 +185,99 @@ have any constructors declared.
...
+```k
+  syntax Bar ::= FooToBar(Foo) [function]
+  rule FooToBar(F) => StringToBar(FooToString(F))
+```
+
+### Parsing comments, and the `#Layout` sort
+
+Productions for the `#Layout` sort are used to describe tokens that are
+considered "whitespace". For example, below, we use it to define lines begining
+with `;` (semicolon) as comments.
```
ehildenb: Clarify here that they will be thrown away by the scanner, so the parser will not even see these tokens.

pending-documentation.md
```diff
@@ -185,6 +185,99 @@ have any constructors declared.
...
+```k
+syntax #Layout ::= r"(;[^\\n\\r]*)"    // Semi-colon comments
+                 | r"([\\ \\n\\r\\t])" // Whitespace
+```
+
+### `prec` attribute
+
+Consider the following naive attempt at creating a language whos syntax
+allows two types of variables: Names that contain underbars, and names that
+contain sharps/hashes/pound-signs:
```
ehildenb: `whos syntax allows` => `with syntax that allows`.

Also lowercase `Names` => `names`

pending-documentation.md
```diff
@@ -185,6 +185,99 @@ have any constructors declared.
...
+general tokens. We add the token attribute to this production so that all tokens
+of a particular sort are marked with the sort it is parsed as, and not a subsort
+thereof. e.g. we get `underbar(#token("foo", "NameWithUnderbar"))` instead of
+`underbar(#token("foo", "#LowerId"))`
+
+*TODO:* `#UpperId` and `#LowerId` have `prec(2)` while `KLabel` and `#KVariable`
+have `prec(1)`. That does not leave much room for other priorities. Even if
+`KLabel`, `#KVariable` aren't a problem, since we can't have negative
+precedences, only `prec(0)` and `prec(1)` are available to users. Perhaps we
+should multiply these by 100?
```
ehildenb: This should not be in the documentation, but made into an issue.

pending-documentation.md
```diff
@@ -1321,6 +1414,43 @@ Note that in the case of Set and Map, one guarantee is that K1, K2, K3, and K4
...
+
+```k
+syntax Set ::= collectOddFoos(FoosCell) [function]
+rule collectOddFoos(<foos> <foo> I </foo> REST </foos>)
+  => SetItem(I) collectOddFoos(<foos> REST </foos>)
+  requires I %Int 2 ==Int 1
+rule collectOddFoos(<foos> <foo> I </foo> REST </foos>)
+  => collectOddFoos(<foos> REST </foos>)
+  requires I %Int 2 ==Int 0
+rule collectOddFoos(<foos> .Bag </foos>) => .Set
```
```suggestion
    syntax Set ::= collectOddFoos(FoosCell) [function]
 // -----------------------------------------
    rule collectOddFoos(<foos> .Bag </foos>) => .Set
    rule collectOddFoos(<foos> <foo> I </foo> REST </foos>) => SetItem(I) collectOddFoos(<foos> REST </foos>) requires I %Int 2 ==Int 1
    rule collectOddFoos(<foos> <foo> I </foo> REST </foos>) => collectOddFoos(<foos> REST </foos>) requires I %Int 2 ==Int 0
```

This is just a suggestion, but try to format this one so it's more readable/scannable. I like to try and make it obvious what the differences between the multiple rules are.

pending-documentation.md
nishantjr: Different code blocks in the document are formatted differently. The most common was unindented, so I was using that. But, I'll change it to 4 spaces.

nishantjr: Actually, the document is overwhelmingly unindented, so I'll move the couple I forgot to left indented to that format

ehildenb:
It looks good to me, though the `Matching Cell Fragments` section seems maybe unneeded, but it could be helpful. @dwightguth will have to review.

dwightguth (commented):
This looks mostly okay, but I'm not happy with the  text of the section relating to `prec`. In particular, it does not simply try rules with higher precedence first. What happens is that it will always choose the longest token (ie, the most characters), it can. and if there is more than one possible token that is the longest, THEN it will choose the one out of that set with the highest precedence. unless more than one has the highest precedence, in which case it is unspecified which will be chosen.

dwightguth (approved):

PR #1395, opened 2020-07-02
fix z3 algorithm for sort inference

dwightguth:


ehildenb (changes requested):
Does this fix a bug visible to the K user? If so, please add a test.

ehildenb (approved):

PR #1401, opened 2020-07-08
make symbolic map lemmas available for haskell backend as well

daejunpark:
Only the last commit contains meaningful changes. Others are reorg changes.


k-distribution/tests/regression-new/map-symbolic-tests/Makefile
```diff
@@ -1,7 +1,7 @@
 DEF=map-tests
 EXT=mtest
 TESTDIR=.
-KOMPILE_BACKEND=java
+KOMPILE_BACKEND=haskell
 KPROVE_FLAGS=--restore-original-names --no-alpha-renaming --warnings none
 
 include ../../../include/kframework/ktest.mak
```
ehildenb: We should probably run these tests with both backends for now.

daejunpark: Yep, I know. I just wanted to test quickly, but doesn't seem to be a good idea, making Jenkins too loaded. Will not do that later.

ehildenb: Well you can also test locally instead. Just run `make` in this directory after building.

daejunpark: Yeah, I thought that K Jenkins build is quicker than compiling the haskell backend in my machine, which was wrong.

daejunpark:
I think we need to make this test similar to the KEVM lemmas-spec.k.  The haskell backend output contains too verbose.

daejunpark:
Hmm. `in_keys` doesn't seem to work as expected in the haskell backend..

daejunpark:
OK. @ssaha6 will transform this test to make it similar to the lemmas-spec.k test.

PR #1403, opened 2020-07-09
module INT-SYMBOLIC: marking just [symbolic] what can be common for Java and Haskell

denis-bogdanas:


dwightguth (approved):

k-distribution/include/kframework/builtin/domains.k
```diff
@@ -433,7 +433,25 @@ module INT-COMMON
                 | Int "dividesInt" Int    [function]
 endmodule
 
-module INT-SYMBOLIC [symbolic, kore]
+module INT-SYMBOLIC [symbolic]
+  imports INT-COMMON
+  imports INT-SYMBOLIC-KORE
```
ehildenb, line 438: The module import structure here seems a bit weird to me.

We have `INT-SYMBOLIC` import `INT-SYMBOLIC-KORE` and `INT-COMMON`, but `INT-SYMBOLIC-KORE` already imports `INT-COMMON`. Is this intentional @denis-bogdanas .

ehildenb (dismissed):
Looks good except for some module import weirdness.

k-distribution/include/kframework/builtin/domains.k
denis-bogdanas: @ehildenb  `INT-SYMBOLIC-KORE` has different attributes and is not imported all the time. So all imports are needed.

PR #1407, opened 2020-07-09
Update issue templates

ehildenb:


dwightguth:
I would also mention that we expect k definitions shared in bug reports to be minimized to as few files/sentences as possible while still reproducing the error.

ehildenb:
@dwightguth I've updated the template, please have another look.

dwightguth (approved):

PR #1409, opened 2020-07-09
report error if concrete backend has existentially-qualified variable

### LINKED ISSUE #1399: LLVM backend crashes when compiling rules with existential variables
Given input file `error.k`

```
module ERROR-SYNTAX
  syntax KItem ::= "Foo"
endmodule

module ERROR
  imports ERROR-SYNTAX
  rule Foo => ?_:KItem
endmodule
```

and using command:

`kompile --backend llvm error.k`

produces the following kind of error output:

```
/home/stephen/work/kframework/master/k-distribution/target/release/k/lib/kframework/../../bin/llvm-kompile: line 44: 353892 Aborted                 (core dumped) "$(dirname "$0")"/llvm-kompile-codegen "$definition" "$dt_dir"/dt.yaml "$dt_dir" $debug > "$mod"
[Error] Critical: llvm-kompile returned nonzero exit code: 134
Examine output to see errors.
```

Obviously, the LLVM backend doesn't support this kind of rule, but it would be helpful to inform the user about this case --- I ran into this problem when using the Markdown code tag feature so that the concrete and symbolic semantics of a language could live together in one file.

dwightguth:
Fixes #1399


k-distribution/tests/regression-new/checkWarns/Makefile
```diff
@@ -1,4 +1,4 @@
 KOMPILE_FLAGS=-w2e
-KOMPILE_BACKEND=llvm
+KOMPILE_BACKEND=haskell
```
ehildenb, line 2: Hmmmm, why is this here? Is there a test in this directory which has an existential on the RHS (so it was now failing for the wrong reason?)

Maybe we shouldn't be failing until we collect _all_ the warnings then.

ehildenb: Or maybe we should be running these checks tests with both backends all the time.

It makes me nervous that we'll be missing some warnings/errors because we're going through the slightly different `kompile` pipelines.

kernel/src/main/java/org/kframework/compile/checks/CheckRHSVariables.java
```diff
@@ -29,33 +30,37 @@
...
+        this.errorExistential = errorExistential;
     }
     private void check(Rule rule) {
         resetVars();
         Set<String> unboundVariableNames = getUnboundVarNames(rule);
-        gatherVars(true, rule.body());
-        gatherVars(false, rule.requires());
-        gatherVars(false, rule.ensures());
+        boolean isMacro = ExpandMacros.isMacro(rule);
+        boolean errorExistential = !isMacro && this.errorExistential && !(rule.att().contains(Att.LABEL()) && rule.att().get(Att.LABEL()).equals("STDIN-STREAM.stdinUnblock"));
```
ehildenb: Why are `macro` allowed to have existential variables on the RHS?

Please add test-cases exercising each of the clauses in this big conjunct.

ehildenb (approved):

PR #1413, opened 2020-07-10
map-symbolic-tests: update assignment specs to have an explicit RHS

daejunpark:
This is the first step toward running the map-symbolic-tests for both java and haskell backends.

The existing tests rely on the kprove failure output (even for passed tests), which the two backends do not agree on.  So, it is better to make the test specs to have an explicit RHS, so that the *.out file of passed tests could be simply `#True`.  For failed tests, the *.out file will still have the failure output.  This way, we can easily identify which tests are expected to pass or fail.  This also allows us later to reorg this test directory to run the haskell backend as well, and have different *.out files for different failed tests.

The explicit RHS of all assignment specs is manually written carefully, and I confirmed that the _haskell_ backend passes all but two (assignment-{7,11}-spec), where the two are failed because the haskell backend failed to infer `Z =/=K X` from both `Z in_keys(MAP)` and `notBool X in_keys(MAP)`.   The java backend, unfortunately, failed more tests because of its unsound behaviors and certain unification bugs, but I don't think we need to take care of them now, because we will eventually switch to the haskell backend.

The diff is quite big, but all the spec files (and *out files) are auto-generated by test-gen.sh, so the only meaningful file to review is: https://github.com/kframework/k/compare/map-symbolic-tests?expand=1#diff-25df3c38373f6337ee94dc799e6f7f42

But still the diff is big, and I'd like to have a separate PR for each test group (inkeys, lookup, and remove).  But please let me know if you want to see the changes of all the four test groups in a single PR.


k-distribution/tests/regression-new/map-symbolic-tests/assignment
```diff
@@ -1,31 +1,31 @@
...
+rule <k> assignment ( (MAP:Map X:MyId |-> 1) [ Y:MyId <- 2 ] [ Z:MyId <- 3 ] ) => assignmentResult (   MAP [ Y <- 2 ] [ Z <- 3 ]  X |-> 1   )   </k> requires Y =/=K X andBool Z =/=K Y andBool Z =/=K X
+rule <k> assignment ( (MAP:Map X:MyId |-> 1) [ Y:MyId <- 2 ] [ Z:MyId <- 3 ] ) => assignmentResult (   MAP [ Y <- 2 ] [ Z <- 3 ]  X |-> 1   )   </k> requires Y =/=K X andBool Z =/=K Y andBool Z =/=K X andBool Y in_keys(MAP) andBool Z in_keys(MAP)
+rule <k> assignment ( (MAP:Map X:MyId |-> 1) [ Y:MyId <- 2 ] [ Z:MyId <- 3 ] ) => assignmentResult (   MAP [ Y <- 2 ] [ Z <- 3 ]  X |-> 1   )   </k> requires Y =/=K X andBool Z =/=K Y andBool Z =/=K X andBool notBool Y in_keys(MAP) andBool notBool Z in_keys(MAP)
+rule <k> assignment ( (MAP:Map X:MyId |-> 1) [ Y:MyId <- 2 ] [ Z:MyId <- 3 ] ) => assignmentResult (   MAP  X |-> 3                         )   </k> requires Y ==K X andBool Z  ==K Y andBool Z  ==K X
+rule <k> assignment ( (MAP:Map X:MyId |-> 1) [ Y:MyId <- 2 ] [ Z:MyId <- 3 ] ) => assignmentResult (  .Map                                  )   </k> requires Y ==K X andBool Z  ==K Y andBool Z  ==K X andBool X in_keys(MAP) // unsatisfiable pre-condition
```
daejunpark, line 18: I'd like to have the RHS of this spec to be `#False`, but the java backend couldn't parse it.  `.Map` should be not bad, because it is unlikely to be proved accidentally.

k-distribution/include/kframework/builtin/domains.k
```diff
@@ -190,9 +190,7 @@ module MAP-JAVA-SYMBOLIC [kast, symbolic]
   rule ((K1 |-> V1) MAP) [ K2 <- V2 ] => (K1 |-> V1) (MAP [ K2 <- V2 ]) requires K1 =/=K K2
 
   rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP              [ K1 <- V2 ] requires K1  ==K K2
-
-  // potential infinite loop
-  // rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP [ K2 <- V2 ] [ K1 <- V1 ] requires K1 =/=K K2
+//rule (MAP:Map [ K1 <- V1 ]) [ K2 <- V2 ] => MAP [ K2 <- V2 ] [ K1 <- V1 ] requires K1 =/=K K2 // potential infinite loop even if `K2 in_keys(MAP)`
```
daejunpark, line 193: Note that I also found that this rule could lead to an infinite loop, even if we have the additional condition of `K2 in_keys(MAP)`.  A counter example: `M [ K1 <- _ ] [ K2 <- _ ]` where `M` is a variable, and we have both `K1 in_keys(M)` and `K2 in_keys(M)`.

(It is possible to make it finite by having a custom `in_keys` rules, but its performance overhead could be problematic for a large map. So, such rules are better to be put in project-specific lemmas.)

denis-bogdanas (approved):
Very neat, thanks for this cleanup! All seems correct.

dwightguth (approved):
I haven't looked at this in detail, but the changes to k itself seem fine and it looks like Denis has reviewed the tests

ehildenb:
Looks good to me too, but one thing you may consider @daejunpark . All the proofs that return `#True` could be put into a single `*-spec.k` file (maybe like `assignment-true-spec.k`), to consolidate the number of files. Then each failing one can get its own file. Up to you.

daejunpark:
@ehildenb That would not be that simple. Later we want to share the same *-spec.k file(s) for both java and haskell backends, but the set of passed tests by each backend would be different. We can still have a single spec file that contains all tests passed by both backends, but I'm not sure if it looks necessarily better. I'd like to go with the current form, if you're ok with it.

ehildenb (approved):

PR #1414, opened 2020-07-10
preparations for llvm backend PR relating to warnings

dwightguth:
We merge this PR first, then we will merge a PR which gives the llvm backend code the appropriate warning enumeration values, then we will merge the frontend PR that does the rest of the work.


ehildenb:
@dwightguth to be clear, we're only adding these two warnings right now because these two are reported by the LLVM backend right? In the follow up PR, you will add the remaining warnings to the enum?

dwightguth:
Correct. Specifically, these need to be added now so we can make the changes to the llvm backend without introducing problems because of the cross-repository dependency of k(k-distribution) -> llvm-backend(matching) -> k(kore). We can't make the rest of the changes to the frontend until we have modified the llvm backend, but we can't modify the llvm backend without at least this much of a change to the frontend.

ehildenb (approved):

PR #1420, opened 2020-07-13
debugger flag to kprove

### LINKED ISSUE #1289: Expose `--repl` option to Frontend
The haskell backend repl is one of the main ways to debug Haskell backend proofs.

Currently though, we need to override the `--haskell-backend-command` when calling `kprove ...`, which make involve setting up paths manually and such.

The `--repl` is a common enough option now that it should be exposed directly to the frontend so that we can just do either `kprove --repl ...` or `kprove --debugger ...` (to imitate `krun --debugger ...`).

dwightguth:
Fixes #1289


ttuegel (approved):

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -305,6 +312,10 @@ private String saveKoreSpecToTemp(ModuleToKORE converter, Module rules) {
                     args.add("--smt-prelude");
                     args.add(smtOptions.smtPrelude);
                 }
+                if (kProveOptions.debugScript != null) {
+                    args.add("--repl-script");
+                    args.add(files.resolveWorkingDirectory(kProveOptions.debugScript).getAbsolutePath());
+                }
```
ehildenb: Should we only enable this option if `--debugger` is passed? The only other use-case I can think of is if they pass `--haskell-backend-command` directly, and also want to use this option, but then tehy can just add `--repl-script` to their `--haskell-backend-command`.

ehildenb (approved):

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
dwightguth: I mean, if anything it should be an error to pass this flag but not --debugger. I can implement that behavior if you'd like.

ehildenb: Up to you, I don't think it's worth blocking the merge. But yes, that's what I meant, this flag should only be passed if `--debugger` is passed. If it's simple, we might as well do it I think.

PR #1421, opened 2020-07-13
map-symbolic-tests: updated lookup, inkeys and remove specs to have an explicit RHS

ssaha6:


ehildenb:
This needs to be updated.

PR #1424, opened 2020-07-13
documentation for attributes

### LINKED ISSUE #1293: Document attributes
- [x] `priority` on rules
- [x] `owise` on rules
- [x] `priorities` on syntax
- [x] `left` on syntax
- [x] `assoc` on syntax
- [x] `comm` on syntax

dwightguth:
Fixes #1293

### DIFF
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -382,7 +382,7 @@ syntax Exp ::= left:
 
 This indicates that multiplication and division are left-associative, ie, after
 symbols with higher priority are parsed as innermost, symbols are nested with
-the rightmost on top. Addition and subtrasction are right associative, which 
+the rightmost on top. Addition and subtraction are right associative, which 
 is the opposite and indicates that symbols are nested with the leftmost on top.
 Note that this is similar but different from evaluation order, which also
 concerns itself with the ordering of symbols, which is described in the next
--- a/pending-documentation.md
+++ b/pending-documentation.md
@@ -321,9 +321,9 @@ parse that is chosen. If all the parses were rejected, it is a parse error. If
 multiple parses remain, they might be resolved by further disambiguation such
 as via the `prefer` and `avoid` attributes, but if multiple parses remain after
 disambiguation finishes, this is an ambiguous parse error, indicating there is
-not a unique parse for that  term. In this case, you can get all possible
-parses that were valid by means of the `--gen-glr-parser` command line flag.
-Usually, however, this indicates a problem in your grammar.
+not a unique parse for that term. In the vast majority of cases, this is
+an error and indicates that you ought to either change your grammar or add
+brackets to the term in question. 
 
 Priority is specified in K grammars by means of one of two different
 mechanisms. The first, and simplest, simply replaces the `|` operator in a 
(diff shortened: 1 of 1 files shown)


pending-documentation.md
```diff
@@ -276,16 +277,193 @@ syntax Pgm ::= underbar(NameWithUnderbar)
...
+
+Priority is applied individually to each possible ambiguous parse of a term. It
+then either accepts or rejects that parse. If there is only a single remaining
+parse (after all the other disambiguation steps have happened), this is the
+parse that is chosen. If all the parses were rejected, it is a parse error. If
+multiple parses remain, they might be resolved by further disambiguation such
+as via the `prefer` and `avoid` attributes, but if multiple parses remain after
+disambiguation finishes, this is an ambiguous parse error, indicating there is
+not a unique parse for that  term. In this case, you can get all possible
+parses that were valid by means of the `--gen-glr-parser` command line flag.
```
ehildenb: I would remove this sentence about `--gen-glr-parser` flag. It makes it sound like the purpose of this flag is for returning all ambiguous parses, which is not the purpose of that flag. Instead say that K will print out the ambiguous parses and the user must fix it by adding brackets where needed or changing their grammar.

pending-documentation.md
```diff
@@ -276,16 +277,193 @@ syntax Pgm ::= underbar(NameWithUnderbar)
...
+ambiguities in the above grammar like so:
+
+```k
+syntax Exp ::= left:
+               Exp "*" Exp
+             | Exp "/" Exp
+             > left:
+               Exp "+" Exp
+             | Exp "-" Exp
+```
```
ehildenb: I would have one of these examples be `left`, and the other `right`, to show that you can use both.

pending-documentation.md
```diff
@@ -1213,6 +1391,72 @@ Here we know that `foo(I +Int 1)` will return an Int, but the return sort of
...
+consider the following function:
+
+```k
+syntax Int ::= foo(Int) [function]
+rule foo(0) => 0
+rule foo(_) => 1 [owise]
+```
+
+Here `foo(0)` is defined explicitly as 0. Any other integer yields the integer
+1. In particular, the second rule above will only be tried after the first
```
```suggestion
Here `foo(0)` is defined explicitly as `0`. Any other integer yields the integer
`1`. In particular, the second rule above will only be tried after the first
```

pending-documentation.md
```diff
@@ -1213,6 +1391,72 @@ Here we know that `foo(I +Int 1)` will return an Int, but the return sort of
...
+rule has been shown not to apply.
+
+This is because the first rule has a lower number assigned for its priority
+than the second rule. In practice, each rule in your semantics is implicitly
+or explicitly assigned a numerical priority. Counterintuitively, numbers which
+are lower correspond to **higher** priority, i.e., they will be tried first.
+
```
ehildenb: I would remove the "Counterintuitively", and instead say something to the effect that "rules are tried in increasing order of priority, starting at priority `0` and ending at priority `200`". It's only counterintuitive when you think of the number as how "prioritous" a rule is, instead of "where the rule goes in teh priority list".

pending-documentation.md
```diff
@@ -1213,6 +1391,72 @@ Here we know that `foo(I +Int 1)` will return an Int, but the return sort of
...
+You can specify the priority of a rule with the `priority` attribute. For
+example, I could equivalently write the second rule above as:
+
+```k
+rule foo(_) => 1 [priority(200)]
+```
+
+The number 200 is not chosen at random. In fact, when you use the `owise`
+attribute, what you are doing is implicitly setting the priority of the rule
+to 200. This has a couple of implications:
```
```suggestion
The number `200` is not chosen at random. In fact, when you use the `owise`
attribute, what you are doing is implicitly setting the priority of the rule
to `200`. This has a couple of implications:
```

pending-documentation.md
```diff
@@ -1213,6 +1391,72 @@ Here we know that `foo(I +Int 1)` will return an Int, but the return sort of
...
+
+1. Multiple rules with the owise attribute all have the same priority and thus
+   can apply in any order.
+2. Rules with priority higher than 200 apply **after** all rules with the 
+   `owise` attribute have been tried.
```
```suggestion
2. Rules with priority higher than `200` apply **after** all rules with the 
   `owise` attribute have been tried.
```

pending-documentation.md
```diff
@@ -276,16 +277,195 @@ syntax Pgm ::= underbar(NameWithUnderbar)
...
+               Exp "*" Exp
+             | Exp "/" Exp
+             > right:
+               Exp "+" Exp
+             | Exp "-" Exp
+```
+
+This indicates that multiplication and division are left-associative, ie, after
+symbols with higher priority are parsed as innermost, symbols are nested with
+the rightmost on top. Addition and subtrasction are right associative, which
```
```suggestion
the rightmost on top. Addition and subtraction are right associative, which 
```

ehildenb (approved):

PR #1426, opened 2020-07-13
Creating separate save option

### LINKED ISSUE #1330: Add `--md-selector` option to `kprove`
KProve currently calls the kompile pipeline, which can take either markdown or k file sources. So we should also add the `--md-selector` option to `kprove` to make sure that people can use tangling on proofs they want to perform as well.

radumereuta:
Fixes #1330


radumereuta:
This should be fixed when we redesign kprove.

PR #1428, opened 2020-07-15
Fine-tuned warning control

### LINKED ISSUE #1405: Framework for enabling/disabling specific warnings
We need to design/implement a system for enabling different warn levels.

Examples are OCaml and GCC.

-   GCC has a bunch of `--WWARNING-GROUP-NAME`, or `-Wno-warning-group-name`. 
-   OCaml has each warning be asssigned a number (warn-level), and sets the list of numbers which are default warn levels. 
-   GHC has syntax similar to GCC but meaning is more like OCaml.

### Implementation:

Warnings names toggleable by `-W` and `-Wno`.
3 specific warn groups `-Wall`, `-Wdefault`, and `-Wnone`.

dwightguth:
Basically, there's a lot of underlying machinery, but the only user-facing changes are that a couple of warnings that sometimes have false positives were made hidden, a few warnings that were very rare and in deprecated code were made into errors, and command line flags were added that allow fine tuned control over exactly which warnings are displayed.

Fixes #1405


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -105,7 +105,7 @@ public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager
         this.sw = sw;
 
         if (kompileOptions.backend.equals("ocaml")) {
-            kem.registerCriticalWarning("The OCaml backend is in the process of being deprecated (final date May 31, 2020). Please switch to the LLVM backend.");
+            kem.registerCriticalWarning(ExceptionType.DEPRECATED_BACKEND, "The OCaml backend is in the process of being deprecated (final date May 31, 2020). Please switch to the LLVM backend.");
         }
```
ehildenb: Is there much use in distinguishing `DEPRECATED_BACKEND` from `FUTURE_ERROR`? Seems to me that `DEPRECATED_BACKEND` will only be used twice, and equally well could be just thrown into "these are features that we will not support in the future.

kore/src/main/java/org/kframework/utils/errorsystem/KException.java
```diff
@@ -103,7 +95,24 @@ public int hashCode() {
...
+        FUTURE_ERROR,
+        UNUSED_VAR,
+        PROOF_LINT,
+        FIRST_HIDDEN, // warnings below here are hidden by default
+        MISSING_HOOK_JAVA,
+        USELESS_RULE,
+        UNRESOLVED_FUNCTION_SYMBOL,
+        MALFORMED_MARKDOWN,
+        INVALIDATED_CACHE,
+        UNUSED_SYMBOL
```
ehildenb: I think backend-specific things should not be in this list. So we should just have `MISSING_HOOK`, instead of `MISSING_HOOK_OCAML` or `MISSING_HOOK_JAVA`. And we should get rid of `DEPRECATED_BACKEND` and just use `FUTURE_ERROR`.

dwightguth: I actually originally had these separate, but they had to be split because most java backend definitions report a lot of missing hook warnings just because a bunch of hooks were never implemented, and it makes the output extremely verbose whenever you call krun.

dwightguth: So I had to make a separate category so I could make it disabled by default.

ehildenb: Ah OK, makes sense.

ehildenb (approved):

PR #1430, opened 2020-07-15
Update issue templates

ehildenb:


dwightguth (commented):
This feels very vague, when my feeling is that we should ask for precise specification when it comes to changes to the K language. We can always amend the specification as it is written by the original person, but we can't amend it into something precise if the template it starts from is not precise.

ehildenb:
@dwightguth do you have a suggestion about where it can be less vague? It contains all the needed components I believe, and having them write initial documentation of the feature I think should give us a precise enough grasp on what is going on with the new feature (as well as a psuedo-code example and test-case).

ehildenb:
But keep in mind that we can't predict what new features people will want will look like, so we need to keep it not too precise I think. Sometimes they'll be new attributes, other times new outer syntax.

dwightguth (approved):

PR #1432, opened 2020-07-16
Make new krun replacement work with haskell backend

dwightguth:


ttuegel (commented):
I encountered some errors while trying to test this locally.

- The script's interpreter is `/bin/bash`. There is no guarantee that `bash` is installed in `/bin`. The cross-platform way to call bash is `/usr/bin/env bash`.
- After correcting for that error, the script fails with `llvm-krun: command not found`.

dwightguth:
The script uses `llvm-krun` and `kprint`, which are installed by the llvm backend, to construct the initial configuration and perform unparsing. Did you build k with `-Dllvm.backend.skip`?

k-distribution/src/main/scripts/bin/kx
```diff
@@ -11,26 +11,30 @@ depth=
 params=()
 allPath=true
 dir=.
-cmd=
+cmd=kore-exec
```
ehildenb: This variable should be named a little better, because it appears it's only for the Haskell backend? Or at least, it's only for when doing search or prove?

I would say it should be `haskellCmd`, instead of just `cmd`. If later we change LLVM backend to handle things like search or prove, then we can change the variable name.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1374,7 +1374,7 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
           if (prod.items().apply(i) instanceof NonTerminal) {
             format = format.replaceAll("%" + (i+1) + "(?![0-9])", "%" + (nt++));
           } else if (prod.items().apply(i) instanceof Terminal) {
-            format = format.replaceAll("%" + (i+1) + "(?![0-9])", "%c" + ((Terminal)prod.items().apply(i)).value().replace("\\", "\\\\").replace("$", "\\$") + "%r");
+            format = format.replaceAll("%" + (i+1) + "(?![0-9])", "%c" + ((Terminal)prod.items().apply(i)).value().replace("\\", "\\\\").replace("$", "\\$").replace("%", "%%") + "%r");
```
ehildenb, line 1377: This doesn't result in too many `%` being added? It looks something like `%3` could get turned into `%%cSOME_TERMINAL`, where maybe it should be `%cSOME_TERMINAL`.

Also, why use `replace` for the last two instead of `replaceAll`? Just want you to double-check that this is correct.

kernel/src/main/java/org/kframework/ksearchpattern/KSearchPatternFrontEnd.java
```diff
@@ -0,0 +1,122 @@
...
+          KompileOptions kompileOptions = this.kompileOptions.get();
+          Rule pattern = compiledDef.compilePatternIfAbsent(files, kem, options.pattern(), Source.apply("<command line>"));
+          K patternTerm = RewriteToTop.toLeft(pattern.body());
+          K patternCondition = pattern.requires();
+          org.kframework.definition.Module mod = compiledDef.executionModule();
+          ModuleToKORE converter = new ModuleToKORE(mod, files, compiledDef.topCellInitializer, kompileOptions);
+          StringBuilder sb = new StringBuilder();
+          ExpandMacros macroExpander = ExpandMacros.forNonSentences(mod, files, kompileOptions, false);
+          K withMacros = macroExpander.expand(patternTerm);
+          K kWithInjections = new AddSortInjections(mod).addInjections(withMacros);
```
ehildenb, line 102: This looks like a very short `kompile` pipeline? Could it be framed as such? Mostly curious.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: %c and %r are not part of the string passed to `replace`. We use `replace` because `replace` replaces a string everywhere with another string, whereas `replaceAll` replaces a regular expression with a string containing backreferences. We need the regular expressions in one case, but not the other, thus, we use replaceAll only in that one place.

kernel/src/main/java/org/kframework/ksearchpattern/KSearchPatternFrontEnd.java
dwightguth: What do you mean by "framed as"? Like, what change are you suggesting? It is correct that we are essentially taking a single pattern with a single optional side condition and parsing and kompiling it to kore.

ehildenb: Not suggesting a change, just want to understand (should have made that clear). The main reason I'm thinking is because we've discussed changes to the kompile pipeline (ie framing it more declaratively using dependencies/checks interface), and if that happens the question is should this also happen here, or not?

Nothing actionable, just want to make sure it's been noted.

k-distribution/include/kframework/ktest.mak
```diff
@@ -45,6 +45,9 @@ SOURCE_EXT?=$(or $(and $(wildcard $(DEF).k), k), $(or $(and $(wildcard $(DEF).md
 ifeq ($(KOMPILE_BACKEND),llvm)
 KRUN=$(abspath $(MAKEFILE_PATH)/../../bin/kx)
 endif
+ifeq ($(KOMPILE_BACKEND),haskell)
+KRUN=$(abspath $(MAKEFILE_PATH)/../../bin/kx)
+endif
```
ehildenb, line 50: Are we testing the old `krun` codepath at all at this point then?

ehildenb: We still are falling back to the old tools if we don't have direct generated parsers, so it should be OK.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ehildenb: Ah sorry, I misread.

k-distribution/include/kframework/ktest.mak
dwightguth: Well, we are still testing the code path via the java backend in places...

ehildenb (approved):

ttuegel (approved):
LGTM! I was able to get it working once the LLVM backend was installed.

PR #1437, opened 2020-07-17
Report error if user uses cell bags on unsupported backend

### LINKED ISSUE #1419: Modify tutorial to not use type Bag (so kore backends work)
There's a test for bag cells for Java backend, which doesn't work on LLVM:
To enable on LLVM:
File `regression-new/cell-sort-llvm/Makefile`
```
DEF=../cell-sort/test
EXT=test
TESTDIR=../cell-sort
KOMPILE_BACKEND=llvm

include ../../../include/ktest.mak
```

Result: either test should pass or throw a meaningful error that cell bags are not supported.
Actual: no idea, Jenkins is down at the time of editing, but none of the above.

dwightguth:


k-distribution/tests/regression-new/cell-bag-sort-llvm/test.k.out
```diff
@@ -0,0 +1,7 @@
+[Error] Compiler: Cell bags are not supported except on the Java backend. If
+you want this feature, file an issue at https://github.com/kframework/k  As a
+workaround, you can add the attribute type="Set" and add a unique identifier to
+each element in the set.
```
ehildenb: Why is this `Java` backend? Isn't this test run with LLVM backend?

dwightguth: Except on.

ehildenb: Ah, sorry I misread. Ok then, I think we should:

1.  Change the sentence which says `If you want this feature, file an issue ...` to have a period at the end.
2.  Instead link them directly to this issue: https://github.com/kframework/k/issues/1419 and tell them to comment on it instead of filing a new issue.

ehildenb: Also, I would say instead "Cell bags are only supported on the Java backend." It's a bit more natural wording.

ehildenb (approved):

PR #1438, opened 2020-07-17
Report location of import on error importing something

### LINKED ISSUE #1429: [Bug] kompile - line number or source module not indicated when imported module not found.
Input file:

```
module FOO
    imports BAR
endmodule
```

Reproduction steps:

```
$  ./ext/k/k-distribution/target/release/k/bin/kompile test.k
[Error] Compiler: Could not find module: BAR$SYNTAX
```

Expected Behaviour:

```
[Error] Compiler: Could not find module: BAR$SYNTAX  imported by FOO at Source(foo.k, 12, 23)
```

dwightguth:
Fixes #1429

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/json/JsonParser.java
+++ b/kernel/src/main/java/org/kframework/parser/json/JsonParser.java
@@ -8,6 +8,7 @@ import org.kframework.definition.Context;
 import org.kframework.definition.Configuration;
 import org.kframework.definition.Definition;
 import org.kframework.definition.FlatModule;
+import org.kframework.definition.Import;
 import org.kframework.definition.NonTerminal;
 import org.kframework.definition.Module;
 import org.kframework.definition.Production;
@@ -155,8 +156,8 @@ public class JsonParser {
         String name = data.getString("name");
 
         JsonArray jsonimports = data.getJsonArray("imports");
-        Set<String> imports = new HashSet<>();
-        jsonimports.getValuesAs(JsonString.class).forEach(i -> imports.add(i.getString()));
+        Set<Import> imports = new HashSet<>();
+        jsonimports.getValuesAs(JsonString.class).forEach(i -> imports.add(Import.apply(i.getString(), Att.empty())));
 
         JsonArray sentences = data.getJsonArray("localSentences");
         Set<Sentence> localSentences = new HashSet<>();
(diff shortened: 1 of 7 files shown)


kernel/src/main/java/org/kframework/parser/json/JsonParser.java
```diff
@@ -155,8 +156,8 @@ public static FlatModule toFlatModule(JsonObject data) throws IOException {
         String name = data.getString("name");
 
         JsonArray jsonimports = data.getJsonArray("imports");
-        Set<String> imports = new HashSet<>();
-        jsonimports.getValuesAs(JsonString.class).forEach(i -> imports.add(i.getString()));
+        Set<Import> imports = new HashSet<>();
+        jsonimports.getValuesAs(JsonString.class).forEach(i -> imports.add(Import.apply(i.getString(), Att.empty())));
```
ehildenb, line 161: Hmmmm, shouldn't we be correctly encoding/decoding the imports to JSON?

ehildenb: As in, shouldn't we change the JSON encoding to include the attributes on imports instead of assuming it's empty?

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -61,7 +63,32 @@ object FlatModule {
...
+  def noSyntax(name: String): String =
+    if (isSyntax(name))
+      name.dropRight(syntaxString.length)
+    else
+      name
+}
+
+
+
+case class FlatModule(name: String, imports: Set[Import], localSentences: Set[Sentence], @(Nonnull@param) val att: Att = Att.empty)
```
ehildenb, line 91: Because the JSON encoding is directly an encoding of the `FlatModule`, I think it should be a faithful encoding of the actual `Import`, not just the name.

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
dwightguth: I mean, those attributes don't actually get kept. You are converting from a Module back to a FlatModule and the attributes will have been lost by that point anyway. We would need a much larger change that tracked the attributes of imports through the entire compilation process if we were going to be able to do what you suggest here. But since imports can't actually have attributes and the attributes are just placeholders for the location, and the location information isn't really needed by the json, I figured this was good enough.

ehildenb (approved):

PR #1456, opened 2020-07-23
Disable #isConcrete on haskell backend

### LINKED ISSUE #1436: [Bug] krun - Haskell backend cannot use stdout stream
**Describe the bug**

Running a program with the Haskell backend using a definition which is kompiled with the `stdout` stream results in a runtime error because Haskell backend does not support `KREFLECTION.isConcrete` hook (and won't).

This error comes from Stack Overflow issue: https://stackoverflow.com/questions/62943995/kframework-issues-with-isconcrete-and-function-global-configuration-lookup

**Input Files**

`test.k`

```
module TEST-SYNTAX
    imports DOMAINS-SYNTAX
    syntax Prog ::= run(Int)
endmodule

module TEST
    imports TEST-SYNTAX
    imports DOMAINS

    syntax KResult

    configuration
        <k> $PGM:Prog </k>
        <bar> 6 </bar>
        <log stream="stdout"> .List </log>

    syntax Int ::= foo(Int) [function]
 
    rule [[ foo(0) => I ]]
         <bar> I </bar>

    rule run(I) => foo(I)
endmodule
```

`test.in`

```
run(0)
```

**Reproduction Steps**

On version 1bd3a49 of K:

Kompile the definition:

```
```

Running on Haskell backend causes runtime error:

```
$ krun test.in
kore-exec: [2020-07-16 22:18:37.052578181] Error (ErrorException):
    Error:
      module 'TEST':
      hooked-symbol 'Lbl'Hash'isConcrete'LParUndsRParUnds'K-REFLECTION-SYMBOLIC'Unds'Bool'Unds'K' declaration (/home/dev/src/k/./test-kompiled/definition.kore 174:17):
        found KREFLECTION.isConcrete hook
    CallStack (from HasCallStack):
      error, called at app/share/GlobalMain.hs:499:36 in main:GlobalMain
[Error] Critical: Haskell Backend execution failed with code 1 and produced no
output.
```

**Expected behavior**

-   If this is truly not supported, we should fail at `kompile` time instead.
-   Otherwise, the `stdout` support should be refactored to not use `isConcrete` (perhaps using tangling to send a difference version of rules to Java/Haskell backends, if we need to support this on Java backend).

dwightguth:
Fixes #1436

I am tagging @ttuegel just to make sure he agrees nothing in this module is used by the haskell backend.


ehildenb (approved):

ttuegel (approved):
Looks good. The Haskell backend cannot use anything in that module.

PR #1457, opened 2020-07-23
Publish and link to builtins and tutorial on website

### LINKED ISSUE #1452: Provide link to standard library/builtin tutorial directory on Github pages rendered website
This will enable us to see these pages on the website.

ehildenb:
Fixes: #1452


ehildenb:
What it looks like here: https://kframework.github.io/k/

radumereuta:
Some of the links from https://kframework.github.io/k/tutorial/README.html don't work.
I don't think it's the scope of this PR but we should at least track them with an issue.

k-distribution/include/kframework/builtin/README.md
```diff
@@ -0,0 +1,22 @@
+---
+copyright: Copyright (c) 2020 K Team. All Rights Reserved.
+---
+
+K Builtins
+==========
+
+The K Builtins (also referred to as the K Prelude or the K Standard Library)
+consists of several files which contain definitions that make working with K
+simpler.
```
radumereuta: Maybe also add something like: These files can be found under `include/kframework/builtin/` in your installation directory.

ehildenb: Added this: 

```
These files can be found under `include/kframework/builtin` in your K
installation directory, and can be imported with `requires "FILENAME"` (without
the path prefix).
```

ehildenb:
Addressed the text of the builtin readme, but fixing all the links/text of the tutorial is another task, I agree. This is to just to get them in place on the website so we can begin iterating on them.

dwightguth (approved):

PR #1461, opened 2020-07-24
Jenkinsfile: timeout on mac build step

ehildenb:
To keep the K master build from absorbing an executor forever if Anka goes down.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -366,7 +366,7 @@ pipeline {
             branch 'master'
             beforeAgent true
           }
-          options { timeout(time: 120, unit: 'MINUTES') }
+          options { timeout(time: 150, unit: 'MINUTES') }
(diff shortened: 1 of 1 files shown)


dwightguth:
Are you sure this timeout is long enough?

Jenkinsfile
```diff
@@ -366,6 +366,7 @@ pipeline {
             branch 'master'
             beforeAgent true
           }
+          options { timeout(time: 120, unit: 'MINUTES') }
```
```suggestion
          options { timeout(time: 150, unit: 'MINUTES') }
```

ehildenb: @dwightguth you can apply this change if you want a longer timeout. 2hrs should be enough in general, but padding it with 30min should help.

dwightguth (approved):

PR #1464, opened 2020-07-27
Fix deploy process

ehildenb:
Deploy process is currently failing for some silly reason, AFAICT it's because we're using a reference repository for the original clone and then the git objects dir doesn't exist for it to make tags based on.

This updates it using improvements I've learned from the Firefly web deploy process too.


ehildenb:
Successful release: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1464/11/pipeline

dwightguth (approved):

PR #1466, opened 2020-07-28
Convert lambda to markdown

### LINKED ISSUE #1410: Convert `1_k/1_lambda` to markdown
All lessons but the last are `*.k` files, and the last lesson is a `*.md` file (the goal would be just to make sure that no latex definitions are used anymore).

radumereuta:
Fixes: #1410 

And use the bison parser for programs.
I run the tests on my machine before and after the switch to the bison parser and the runtimes are exactly the same.
The input files are too small to see a difference here.

There are a few other issues with the content of the comments (regarding latex usage), but they might require a complete overhaul of the entire tutorial so I didn't touch them.


ehildenb:
Adding `--gen-bison-parser` isn't enough to make the switch. See in the `IMP` semantics that it actually has a little script called `imp` which handles manually calling the bison parser and `kprint` after the fact @radumereuta .

dwightguth:
This is not the case anymore actually. The kx script ought to pick up the bison parser automatically and uses kprint by default. I'm quite surprised you don't see o performance impact. Are you sure you're not using the java k frontend?

radumereuta:
@dwightguth you are right. I wasn't calling the bison parser in kompile because I miswrote the Makefile, but now kompile crashes:
`[Error] Internal: bison returned nonzero exit code: 139`
Tried to run bison manually and I get `Segmentation fault (core dumped)`.

Also, I think this might be a problem:
```
module LAMBDA // main syntax module
  imports DOMAINS
  imports SUBSTITUTION
```
Thoughts?

dwightguth:
I forgot that lambda didn't have a syntax module. Go ahead and skip bison for lambda and we will do it for the other definitions only.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.md
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2019 K Team. All Rights Reserved.
+<!--- Copyright (c) 2014-2020 K Team. All Rights Reserved. --->
```
ehildenb: Use the format used in other markdown files for the copyright instead. Look at `k-distribution/tutorial/1_k/README.md` for an example.

radumereuta: I used the format from imp.md. I changed both files.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.md
```diff
@@ -1,4 +1,6 @@
-// Copyright (c) 2014-2019 K Team. All Rights Reserved.
+---
+copyright: Copyright (c) 2014-2020 K Team. All Rights Reserved.
+---
 
 // C/Java-like comments like this one are completely ignored by the K tool.
```
dwightguth: let's remove this comment and replace it with something talking about how markdown works, because this is not accurate otherwise.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.md
```diff
@@ -13,103 +15,103 @@ several lines, are ignored, too.  */
...
-programming} approach.  The various semantic features defined in a \K
+Note:
+ **K** follows the
+[literate programming](https://en.wikipedia.org/wiki/Literate_programming)
+approach. The various semantic features defined in a **K**
 module can be reordered at will and can be commented using normal
 comments like in C/C++/Java.  If those comments start with
-`\texttt{@}' preceded by no space (e.g.,
-`\texttt{//{@} {\textbackslash}section\{Variable declarations\}}'
+`\texttt{@}` preceded by no space (e.g.,
```
dwightguth: let's adjust the language here too.

dwightguth (changes requested):
Mostly I just want us to remove references to `/*@` and Latex.

radumereuta:
There still is one more reference that explains the latex attributes.
When I converted imp you asked me not to remove them. Not sure what to do with them now.

I've updated the description regarding C-like comments and Markdown integration with K.
I guess we can deprecate KPP now since it's no longer relevant. You can still get the K file without the comments by running `kompile --debug <def>`.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.md
```diff
@@ -0,0 +1,145 @@
...
+```k
+  rule (lambda X:KVar . E:Exp) V:Val => E[V / X]
+```
+
+### Integer and Boolean Builtins
+The LAMBDA arithmetic and Boolean expression constructs are simply rewritten
+to their builtin counterparts once their arguments are evaluated.
+The operations with subscripts in the right-hand sides of the rules below are
+builtin and come with the corresponding builtin sort; they are actually
+written like `+Int` in ASCII, but they have LaTeX attributes to be
```
dwightguth: this is another reference that should probably be adjusted

radumereuta: This is referring to the latex attributes.
When I converted imp for the first time, you (or Everett) asked me not to remove them.
So what do we want to do with them? At this moment I don't see how they are useful.

dwightguth (approved):

PR #1468, opened 2020-07-29
Update documentation of concrete and symbolic attributes

ttuegel:


pending-documentation.md
```diff
@@ -1157,45 +1198,31 @@ rule X +Int (Y +Int Z) => (X +Int Z) +Int Y
...
 
--   the rule lhs _matches_ the subterm of interest, and
--   the side condition has no remainder given the current top-level predicate
-    (that is, the current top-level predicate _implies_ the side condition of
-    the rule).
+Sometimes you only want a rule to apply if some or all arguments are concrete
+(not symbolic). This is done with the `concrete` attribute. Conversely, the
+`symbolic` attribute will allow a rule to apply only when some arguments are not
+concrete. These attributes should only be given with the `simplification`
+attribute.
```
ehildenb, line 1207: Is it currently an error to specify both `simplification` and `concrete/symbolic`? If not, can we open an issue for that and add it to the backlog?

pending-documentation.md
```diff
@@ -1157,45 +1198,31 @@ rule X +Int (Y +Int Z) => (X +Int Z) +Int Y
...
 
-Note that the `simplification` attribute can be applied to _any_ rule, not just
-function rules, and that `simplification` rules are tried _before_ rules from
-the semantic definition.
-
-For example, for the following definition:
+For example, the following will only re-associate terms when all arguments
+are concrete:
```
ehildenb, line 1210: I think use a better example, which won't cause an infinite loop if you can. Just a minor annoyance on my part.

pending-documentation.md
```diff
@@ -1157,45 +1198,31 @@ rule X +Int (Y +Int Z) => (X +Int Z) +Int Y
...
+And the following rules will only re-associate terms when it will end up
+grouping concrete sub-terms:
 
 ```k
-    rule sizeWordStackAux(WS, N) => N +Int sizeWordStackAux(WS, 0)
-      requires N =/=Int 0
-      [simplification]
-```
+rule X +Int (Y +Int Z) => (X +Int Y) +Int Z
+  [simplification, concrete(X), concrete(Y), symbolic(Z)]
```
ehildenb: I thought we supported something like `concrete(X, Y), symbolic(Z)` instead of having them all be separate?

dwightguth (approved):
Looks fine to me, but please get Everett to approve before merging.

pending-documentation.md
ehildenb: Yes, looking at `domains.k` we do support grouping them. Let's use the grouped version, to encourage better styling for people.

pending-documentation.md
ttuegel: I don't see why this will cause a loop. It is the same example used for the Java backend. Anyway, I took two examples from `INT-SYMBOLIC-KORE` instead, as I thought that would be more productive.

pending-documentation.md
ttuegel: Added #1470

ehildenb (approved):

PR #1476, opened 2020-07-31
convert domains.k to markdown

dwightguth:
This is quite a big one because of all the documentation that was involved.


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -533,10 +1082,43 @@ module INT
...
+`M` bits of exponent. The number of bits of precision is assumed to include
+any optional `1` that precedes the IEEE 754 mantissa. In other words, `p24x8`
+is equal to the IEEE `binary32` format, and `p53x11` is equal to the IEEE
+`binary64` format.
+
+```k
 module FLOAT-SYNTAX
   syntax Float [hook(FLOAT.Float)]
-  syntax Float ::= r"([\\+-]?[0-9]+(\\.[0-9]*)?|\\.[0-9]+)([eE][\\+-]?([0-9]+(\\.[0-9]*)?|\\.[0-9]d+))?([fFdD]|([pP][0-9]+[xX][0-9]+))?" [token, prec(1)]
+  syntax Float ::= r"([\\+-]?[0-9]+(\\.[0-9]*)?|\\.[0-9]+)([eE][\\+-]?[0-9]+)?([fFdD]|([pP][0-9]+[xX][0-9]+))?" [token, prec(1)]
```
ehildenb, line 1121: What's going on here? Are we removing support for some class of floating point parses?

ehildenb (approved):

k-distribution/include/kframework/builtin/domains.md
dwightguth: Removed here is the case when the exponent has a decimal point, which I'm quite certain was never actually working and should never have been part of this syntax.

PR #1477, opened 2020-07-31
Check attributes (symbolic/concrete/simplification)

### LINKED ISSUE #1470: Complain when `concrete` or `symbolic` used without `simplification`
It should be an error to use the `concrete` or `symbolic` attributes without the `simplification` attribute.

dwightguth:
Fixes #1470


k-distribution/tests/regression-new/checks/concrete.k
```diff
@@ -0,0 +1,26 @@
...
+  rule bar2(1) => 1
+
+  syntax Int ::= baz(Int) [function]
+  rule baz(0) => 0 [concrete]
+
+  syntax Int ::= baz2(Int) [function]
+  rule baz2(0) => 0 [concrete]
+  rule baz2(1) => 1
+
+endmodule
```
ehildenb, line 28: Just to make sure, can you add a test that uses `concrete(X)` or `symbolic(X,Y)` as well? Because that's another form that these show up in.

k-distribution/tests/regression-new/checks/concrete.k
```diff
@@ -0,0 +1,26 @@
+module CONCRETE
```
ehildenb, line 1: Also add a test that does both `concrete(X), symbolic(X), simplification`, because it looks like you added a check for this case too.

kernel/src/main/java/org/kframework/compile/checks/CheckKLabels.java
```diff
@@ -149,6 +150,32 @@ public void check() {
...
+                    (rule.att().getOptional(Att.SYMBOLIC()).orElse("foo").equals("") &&
+                    rule.att().contains(Att.CONCRETE()))) {
+                    errors.add(KEMException.compilerError("Rule cannot be both concrete and symbolic in the same variable.", rule));
+                }
+                if (!rule.att().getOptional(Att.CONCRETE()).orElse("foo").equals("")) {
+                    allConcrete = false;
+                }
+                if (!rule.att().getOptional(Att.SYMBOLIC()).orElse("foo").equals("")) {
+                    allSymbolic = false;
+                }
```
ehildenb: I really am not a fan of this way of selecting `"foo"` as the default to compare too. Would prefer something like:

```
Set<String> concreteVars = computeConcreteVars();
Set<String> symbolicVars = computeSymbolicVars();
if ((! (concreteVars.isEmpty() && symbolicVars.isEmpty())) && hasSimplification()) {
   error("Can't have concrete/symbolic without simplification");
}
if (concreteVars.intersects(symbolicVars)) {
   error("Can't have the same variable both concrete/symbolic: " concreteVars.intersection(symbolicVars));
}
```

The logic you have here is pretty hard to read.

k-distribution/tests/regression-new/checks/concrete.k
dwightguth: I did not.

kernel/src/main/java/org/kframework/compile/checks/CheckKLabels.java
dwightguth: I will refactor out a function with a more human-readable name, but note that we're not doing anything quite this complicated, nor do I want to implement that now when it was not specifically requested as part of this issue. We can create another issue and prioritize it next week if we want to add more advanced checks. I mostly just implemented this check as part of implementing the other one.

ttuegel (approved):

ehildenb (approved):

PR #1482, opened 2020-08-03
Location info in k

### LINKED ISSUE #1358: Access to source location information in semantics
Source location is a nice to have information for printing error messages, and possibly for defining a semantics for whitespace sensitive languages. My immediate use case is so that I can generate output closer to Boogies, and can check my semantics against their test suite.

We'd like to be able to access the source information in rules. This means that
the source information would need to be represented as a term.

We cannot mark up every single term with source information as that would
increase the size of the kore term unnecessarily.

### One possible interface

Have a special sort `#SourceLocation` with the following constructor.
This constructor is not used in the grammar for program parsing.

```k                                                                                                                    
syntax #SourceLocation ::= #SourceLocation( file: String, line: Int, column: Int)
```

Consider the production:

```k
syntax Foo ::= #SourceLocation "foo" "(" ")"
```

When parsing a program `#SourceLocation` is parsed as an empty token, but is
populated with the source location information for that location.
We may also have multiple occurances if we want the begin and end information
for a particular token.

```k                                                                                                                    
syntax Foo ::= #SourceLocation "foo" "(" ")" #SourceLocation
```

Note that the meaning of 

```k                                                                                                                    
syntax Foo ::= "foo" "(" #SourceLocation ")"
```

is not obvious. Does it correspond to the position of the open or the close paren?

dwightguth:
Fixes #1358


radumereuta:
How is the new functionality being tested exactly?
I was expecting a new test where you would get the location information from a file and access it during krun.

nishantjr:
Besides the test, will rules over parametric productions work with the haskell backend? Or would we need to define our own productions with the `#location` klabel? In that case, what happens if we need multiple sorts to have location information? They cannot both have the `#location` klabel

dwightguth:
@nishantjr I am in conversation with Tom about how to support the type of parametric constructors this needs in the Haskell backend. I would like it to be spec'd out as a separate issue, but the consensus is it is doable with a certain amount of modification of ModuleToKORE, and a small amount of additional support from the Haskell backend.

dwightguth:
I added a pretty basic test. It doesn't test the functionality of the line markers, because most of the actual functionality associated with that code is specific to C and thus lives in the c repo, but it tests the basic locations.

k-distribution/include/kframework/builtin/kast.k
```diff
@@ -295,6 +295,15 @@ module K-AMBIGUITIES
 
 endmodule
 
+module K-LOCATIONS
+  imports STRING-SYNTAX
+  imports INT-SYNTAX
+
+  // filename, startLine, startCol, endLine, endCol
+  syntax {Sort} Sort ::= #location(Sort, String, Int, Int, Int, Int) [symbol, format(%3)]
```
nishantjr, line 305:
```suggestion
  syntax {Sort} Sort ::= #location(Sort, file: String, startLine: Int, startColumn: Int, endLine: Int, endColumn: Int) [symbol, format(%3)]
```

radumereuta:
This looks great, but can you add another rule that modifies the input so in the final output contains terms from both `1.test` and `test.k`?

k-distribution/tests/regression-new/locations/test.k
```diff
@@ -0,0 +1,13 @@
+module TEST-SYNTAX
+  imports UNSIGNED-INT-SYNTAX
+  syntax Exp [locations]
+  syntax Exp ::= Exp "+" Exp | Int [klabel(IntConst)]
```
nishantjr, line 4: What is the purpose of the `IntConst` label?

radumereuta: I think, instead of an injection, it adds the label `IntConst`.
But I have a different question, why do you need to add `[locations]` to the sort declaration?
I would expect location information to be accessible to every term everywhere.
Even if the backend doesn't preserve it currently, we can say `N/A` and fix it later.

nishantjr:
Looks good. Suggested addition to `pending-documentation`

### Source line information

K is able to insert source metadata into the parse tree on a per-sort basis.
To enable this mark the sort with the `locations` attribute.

```
  syntax Exp [locations]
  syntax Exp ::= Exp "/" Exp | Int
```

K implicitly wraps productions of these sorts in a `#location` term (see the `K-LOCATIONS` module in `kast.k`).
The metadata can thus be accessed with ordinary rewrite rules:

```
  rule #location(I / 0, File, L1, C1, L2, C2) => "Error: Division by zero at location" +String File +String ":" Int2String(I) 
```

k-distribution/tests/regression-new/locations/test.k
nishantjr: Yes, by why do we need that label? Won't an injection work just as well?

nishantjr: If we added it to every term everywhere, it would greatly increase the size of the kore representation. We'd also need to unwrap the source information for every sort.

k-distribution/include/kframework/builtin/kast.k
dwightguth: Believe me, I wanted to do this. But the accessor functions that this would generate would be parametric functions, which we don't support yet.

k-distribution/tests/regression-new/locations/test.k
dwightguth: I'm pretty sure you're right that an injection would also work for the location information; I didn't think of this when I was writing the test, though. Do you want me to change it back? Both should work so there's nothing wrong with the test as it is currently written.

dwightguth: As far as why we don't add it everywhere, the answer is it will greatly complexify the implementation of the semantics which would then need to know how to deal with location information being wrapped around every single term.

dwightguth:
> This looks great, but can you add another rule that modifies the input so in the final output contains terms from both `1.test` and `test.k`?

I'm not sure I follow. The rewriter doesn't do anything special with the location annotations; they're just regular terms that you then have in your AST. I don't see how adding more rules to the semantics would actually test anything new.

nishantjr (approved):
Looks good.

radumereuta:
> > This looks great, but can you add another rule that modifies the input so in the final output contains terms from both `1.test` and `test.k`?
> 
> I'm not sure I follow. The rewriter doesn't do anything special with the location annotations; they're just regular terms that you then have in your AST. I don't see how adding more rules to the semantics would actually test anything new.

I wanted the output to be something like this:
```
input: ab
def: rule a => x
output: #location(x, "test.k", 1, 1, 1, 1) ~> #location(b, "1.test", 2, 2, 2, 2)
```
If I have a definition with no rules, then the output of krun should contain only terms from `1.test`, but if you start rewriting parts of that term, some of them would come from the definition, and I want to make sure we're tracking those correctly.
If the user wants to keep track of original input file, then they should specify it by adding the location wrappers.

radumereuta:
This makes krun crash:
```
module TEST-SYNTAX
  syntax Exp [locations]
  syntax Exp ::= Exp "+" Exp [klabel(plus)] | "a" | "b" | "c" | "d" | "e" | "f" | "g"
endmodule

module TEST
  imports TEST-SYNTAX
  imports K-LOCATIONS

  //syntax Exp ::= error(String, String, Int, Int, Int, Int) [klabel(Loc)]
  //rule a => e [anywhere] // e points to "test.k"
  //rule #location(b, _File, _L1, _C1, _L2, _C2) => #location(f, "ff", 100, 101, 102, 103) [anywhere] // create new location for f
  //rule #location(c, File, L1, C1, L2, C2) => #location(g, File, L1, C1, L2, C2) [anywhere] // copy location from c to g
  //rule #location(d, File, L1, C1, L2, C2) => error("message here", File, L1, C1, L2, C2) [anywhere] // convert the term completely
endmodule
```
Input: `a + b`
Other questions are:
* what happens when you have `a + b + c + d` w/o left, or right associative attribute (basically ambiguities in the tree)
* what happens when you have `[anywhere]` on rewrite rules that match on `#location`.
* what happens when you uncomment all the other rules

dwightguth:
The feature in question is designed to track locations in the program being interpreted. Inserting locations from the semantics itself is quite counterproductive to that aim, and not at all what this feature is intended for.

I am aware of the issue where parametric constructors cause the llvm backend to crash if they are not referenced somewhere in the semantics, but I don't view this as terribly important because there's very little reason why you wouldn't reference them /somewhere/.

In response to your other questions:

1. Each parse in the parse forest will be annotated with the location information from that parse
2. The anywhere rule you wrote does not reference locations anywhere, thus it does not match or rewrite over locations. As I said before, the parametric `#location` symbol is just a regular symbol and is not implicitly inserted or used anywhere in the semantics; it is the responsibility of the developer of the semantics to manipulate it appropriately in explicit rules. This being said, anywhere rules over parametric symbols are not supported yet, so I wouldn't expect the later anywhere rules to work. Again, this is a separate issue.

ehildenb (approved):

PR #1499, opened 2020-08-07
fix seqstrict to use result attribute correctly

dwightguth:


ehildenb (approved):
@dwightguth I'm approving but please add a test.

PR #1500, opened 2020-08-07
use clang 10 on ubuntu bionic

dwightguth:


dwightguth:
This turned out to not be blocking me, so I will close this.

PR #1507, opened 2020-08-11
don't add location to amb nodes

dwightguth:


ehildenb:
@dwightguth what's the motivation for this change?

dwightguth:
It's a bugfix; right now we are not initializing these struct fields and so the parser exercises undefined behavior by reading bits that are not initialized.

ehildenb (approved):

PR #1510, opened 2020-08-12
Update simple-untyped.md with scrapable headers

### LINKED ISSUE #1496: Change headers to include citation metadata
From here: https://github.com/kframework/k/pull/1486#discussion_r466503142

Using YAML metadata makes the tools that process these files more able to produce relevant documentation.

We should do this for all files that have sensible "authors" or "institutions".

radumereuta:
Fixes: #1496
Don't really like how GitHub renders them though.
- the table is very long horizontally.
- the copyright column is too narrow.
- to follow the bib and pandoc style for specifying authors I have to split first name of last name

@ehildenb @dwightguth Thoughts on this? Should I continue and do all of them like this, or do we find another format?
I think the GitHub rendering is the most important since this is where most people look at the definitions so we should make them look as good as possible.

I also found a list of supported languages on GitHub for which they offer syntax highlighting: https://github.com/github/linguist/blob/master/vendor/README.md
If we could do something similar it could improve K code readability in the browser.

### DIFF
--- a/k-distribution/tutorial/2_languages/1_simple/1_untyped/simple-untyped.md
+++ b/k-distribution/tutorial/2_languages/1_simple/1_untyped/simple-untyped.md
@@ -1,5 +1,15 @@
 ---
 copyright: Copyright (c) 2014-2020 K Team. All Rights Reserved.
+title: SIMPLE - Untyped
+author:
+  - given: Grigore
+    family: Roșu
+    email: grosu@illinois.edu
+    organization: University of Illinois at Urbana-Champaign
+  - given: Traian Florin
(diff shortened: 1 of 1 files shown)


dwightguth:
I agree that it looks really ugly on github. Do we have the same problem with the kframework.github.io site? or does it look better there? If it looks better there, that's probably acceptable. But if it looks bad both places, that's a bigger problem.

radumereuta:
By looking at #1512 it's not even rendered.

ehildenb:
Wait at the very least until @shd101wyy 's website updates are in.

PR #1522, opened 2020-08-19
package/docker/Dockerfile.ubuntu-bionic: update to z3 4.8.7

ehildenb:
4.8.8 has been out for 3 months, 4.8.7 has been out longer. There is a proof on KEVM that takes absurdly long on 4.8.6 vs 4.8.7, so I don't see any reason not to upgrade.

4.8.8 does not have a NixOS version, so that's why I'm keeping it on 4.8.7.

### DIFF
--- a/package/docker/Dockerfile.ubuntu-bionic
+++ b/package/docker/Dockerfile.ubuntu-bionic
@@ -10,7 +10,7 @@ RUN    apt-get update                   \
                         git             \
                         python
 
-RUN    git clone 'https://github.com/z3prover/z3' --branch=z3-4.8.6 \
+RUN    git clone 'https://github.com/z3prover/z3' --branch=z3-4.8.7 \
     && cd z3                                                        \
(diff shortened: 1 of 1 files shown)


ttuegel:
> 4.8.8 does not have a NixOS version, so that's why I'm keeping it on 4.8.7.

NixOS also didn't have z3-4.8.6, so I don't see why this should be a blocker.

ehildenb:
Hmmm, that's true. Not sure though, I've tested 4.8.7 and it makes a proof I want to go through quickly on KEVM go through quickly, haven't tested 4.8.8. I can also just use 4.8.7 on KEVM directly, without changing this.

ehildenb:
I removed `automerge`, I'm trying 4.8.7 on KEVM on this PR (where the proof that behaves better on 4.8.7 is introduced): https://github.com/kframework/evm-semantics/pull/899

PR #1534, opened 2020-09-01
Resolve overloads in bison parser

dwightguth:
This is necessary in order for the glr disambiguation algorithm to correctly identify two ambiguous parses as having the same AST when overload resolution would lead to that fact being the case.

A happy side effect of this change is that the ASTs generated by the bison parser are now in normal form.


k-distribution/include/kframework/cparser/main.c
```diff
@@ -77,6 +61,20 @@ bool equalsNode(node *x0, node *x1) {
   }
 }
 
+char *injSymbol(char *lesser, char *greater) {
+  char *prefix = "inj{";
+  char *infix = ", ";
+  char *suffix = "}";
+  size_t len = strlen(prefix) + strlen(suffix) + strlen(lesser) + strlen(greater) + strlen(infix) + 1;
+  char *symbol = malloc(len);
+  strcpy(symbol, prefix);
```
virgil-serbanuta, line 70: Half joking: How about caching the lengths, then:
```
  strcpy(strcpy(strcpy(strcpy(strcpy(symbol, prefix) + len(prefix), lesser) + len(lesser), infix) + len(infix), greater) + len(greater), suffix);
```

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -215,6 +217,83 @@ private static void encode(Sort sort, StringBuilder sb) {
...
 
+  private static void appendOverloadCondition(StringBuilder bison, Module module, Production greater, Production lesser, List<Integer> nts) {
+    bison.append("true");
+    for (int i = 0; i < nts.size(); i++) {
+      boolean hasSameSort = lesser.nonterminals().apply(i).sort().equals(greater.nonterminals().apply(i).sort());
+      if (!hasSameSort) {
+        bison.append(" && strncmp($").append(nts.get(i)).append(".nterm->symbol, \"inj{\", 4) == 0 && (false");
+        Sort greaterSort = lesser.nonterminals().apply(i).sort();
+        for (Sort lesserSort : iterable(module.subsorts().elements())) {
+          if (module.subsorts().lessThanEq(lesserSort, greaterSort)) {
```
virgil-serbanuta, line 229: I assume that "less" means "included in", i.e. here `lesserSort` is included in `greaterSort` and `lesser`'s sort is included in `greater`'s, right?

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -215,6 +217,83 @@ private static void encode(Sort sort, StringBuilder sb) {
...
+        bison.append("\";\n" +
+            "    n->hasLocation = " + (hasLesserLocation ? "1" : "0") + ";\n");
+        for (int i = 0; i < nts.size(); i++) {
+          boolean hasSameSort = lesser.nonterminals().apply(i).sort().equals(greater.nonterminals().apply(i).sort());
+          if (hasSameSort) {
+            bison.append(
+                "    n->children[").append(i).append("] = $").append(nts.get(i)).append(".nterm;\n");
+          } else {
+            bison.append(
+                "    node *origChild = $").append(nts.get(i)).append(".nterm;\n" +
```
virgil-serbanuta: Would this work if there are multiple terminals? I.e. wouldn't `origChild` be redeclared in the same block?

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -215,6 +217,83 @@ private static void encode(Sort sort, StringBuilder sb) {
...
+        }
+        bison.append(")");
+      }
+    }
+  }
+
+  private static void appendOverloadChecks(StringBuilder bison, Module module, Production greater, List<Integer> nts, boolean hasLocation) {
+    for (Production lesser : iterable(module.overloads().elements())) {
+      if (module.overloads().lessThan(lesser, greater)) {
+        bison.append("  if (");
```
virgil-serbanuta, line 243: This is a repeated `if( ... ) { ... } else` section, which means that only the first `lesser` production that passes the test is processed. Why is it good enough to use only the first one?

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
dwightguth: A <= B in this context means either A = B, or A is a subsort of B.

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
dwightguth: This is a good catch; I meant to topologically sort the entries, but forgot. This should be fixed now; because of the topological sorting, a lesser overload will always appear before a greater overload, so we will always end up choosing an overload such that there is no lesser overload than the overload we chose that would also satisfy the constraints. As a result, any remaining ambiguities will be real ambiguities.

virgil-serbanuta (approved):
This looks reasonable to me, it might be good to also get an approval from someone that understands the code better.

PR #1535, opened 2020-09-01
Java Backend: Applying "expand macros" step for spec rules earlier, before concretize config

denis-bogdanas:


ehildenb:
@denis-bogdanas please provide a test for this which fails on the previous version, but works on this version. I think the following should work.

file `main.k`:

```
module DATA
    imports INT
    imports BOOL

    syntax Bool ::= #range ( Int, Int )
    rule #range(5, I) => 0 <=Int I andBool I <Int 32  [macro]
    rule #range(8, I) => 0 <=Int I andBool I <Int 256 [macro]
endmodule

module MAIN
    imports DATA

    syntax Pgm ::= "a" | "b"

    rule <k> a => b ... </k>
endmodule
```

file `test-spec.k`:

```
requires "main.k"

module VERIFICATION
    imports MAIN

    rule #range(8, X) => true requires #range(5, X) [smt-lemma]
endmodule

module TEST-SPEC
    imports VERIFICATION

    rule <k> _:Pgm => a ... </k>
endmodule
```

Then call `kompile --boundary-cells k test-spec.k --def-module VERIFICATION`. This should trigger the weird check before this change which causes the `smt-lemma` to incorrectly be interpreted as a proof obligation and fails the entire proof, but shouldn't trigger this after this change.

ehildenb:
But you should double-check that it triggers it.

denis-bogdanas:
Test added, with some modifications. It fails on K-master.

ehildenb (approved):

PR #1537, opened 2020-09-03
Claim keyword

### LINKED ISSUE #1434: [KIP] - kprove should use the `claim` keyword
Introduce a new keyword in K called `claim` which is used to mark things that are to be proved, in anticipation of having claims not just in one module.

```
claim <k> LHS => RHS ... </k>
```

How should `kompile` treat `claim`:

-  Ignore them: 
-  Prove them at `kompile` time: 1
-  Treat them as semantic (trusted) rules: 
-  Error: 11111

Will treat as error.

How should `kprove` treat `rule` declared _outside_ the main module and _not_ on a simplification rule:

-   Ignore: 
-   Error: 1
-   Deprecation warning: 1111
-   Trusted claim: 

We'll use a deprecation warning then, saying that they should replace `rule` with `claim` (haskell backend already knows the `trusted` keyword, so that can continue working with the `claim` keyword).

radumereuta:
Fixes #1434


radumereuta:
@dwightguth please review.
I finally managed to find all the transformation steps used to transform the claims and all tests pass now.
I converted some of the tests, but not all of them.
I'm still not sure about removing `specificationSteps`. Can you explain a bit why you want it removed? It makes it much easier to follow which steps are being applied.
I see that at one point `Tag.java` was removed. Do you recall why? Because I used some of those tags. Did you replace it with something else?

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -297,9 +298,37 @@ private void checkDefinition(Definition parsedDef, Set<String> excludedModuleTag
...
     }
 
+    // checks that are not verified in the prover
+    public void definitionChecks(Set<Module> modules) {
+        modules.forEach(m -> stream(m.localSentences()).forEach(s -> {
+            // Check that the `claim` keyword is not used in the definition.
+            if (s instanceof Claim)
+                errors.add(KEMException.compilerError("Claims are not allowed in the definition.", s));
+        }));
+    }
```
ehildenb, line 312: Add a test to `k-distribution/tests/regression-new/checks` which exercises this case.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -297,9 +298,37 @@ private void checkDefinition(Definition parsedDef, Set<String> excludedModuleTag
...
+            if (m.name().equals(mainDefModule.name()) || mainDefModule.importedModuleNames().contains(m.name()))
+                return s;
+            if (s instanceof Rule && !s.att().contains("simplification")) {
+                kem.registerCompilerWarning(KException.ExceptionType.FUTURE_ERROR, errors, "Deprecated: use claim instead of rule to specify proof objectives.", s);
+                return new Claim(((Rule) s).body(), ((Rule) s).requires(), ((Rule) s).ensures(), s.att());
+            }
+            return s;
+        }, "rules to claim");
+        return mt.apply(specModule);
+    }
```
ehildenb, line 330: Add a test to `k-distribution/tests/regression-new/checkWarns` which exercises this case.

radumereuta: It was a bit more complicated to modify the Makefile in that directory, so I made a new test in `regression-new` with a comment that it should be removed once the transition to claim is done.

ehildenb:
Can you try a KEVM proof (eg. `make tests/specs/mcd/vat-move-diff-spec.k.prove`) with this change? https://github.com/kframework/evm-semantics

kernel/src/main/java/org/kframework/kompile/Kompile.java
radumereuta: This is actually even more complicated than I originally thought. I modified `ktest.mak` to take into consideration stderr for the kprove tests. I modified all tests to use claim instead of the rule keyword, but there are other warnings inside those tests which were ignored until now, stuff like:
```
< [Warning] Compiler: Variable 'M' defined but not used. Prefix variable name
< with underscore if this is intentional.
< 	Source(./a1-spec.k)
< 	Location(12,19,12,20)
```
Or:
```
< kore-exec: [927827] Warning (WarnFunctionWithoutEvaluators):
<     (InfoReachability) while simplifying the claim:
<     No evaluators for function symbol:
<         Lbln'Unds'A'Unds'Int{}
```
^This last one I assume is from the haskell backend. I'm not sure I can fix this, and even if I fix it now, probably it's going to be unstable in the future.

So... thoughts?

ehildenb: You can fix both of these I believe. The first one can eb fixed by changing `M => _M`. The second one can be fixed by changing the production `n_A_` to have the attribute `no-evaluators`.

But I don't think it should be necessary to catch all of `stderr`, should it? Maybe it is necessary.

PR #1553, opened 2020-09-22
Use `claim` keyword for proof obligations

### LINKED ISSUE #1434: [KIP] - kprove should use the `claim` keyword

radumereuta:
Fixes #1434

This is supposed to be a simplified version of #1537 without all the tests so it makes it easier to review.


radumereuta:
Ok, second try. The first PR ended up way too big, so I'm avoiding updating the tests. They will return warnings for using rule instead of claim, but that can be changed in a future PR dedicated for tests only.

This PR contains mostly the changes to the code, so I'm reiterating the message from last time:
@dwightguth please review.
I finally managed to find all the transformation steps used to transform the claims and all tests pass now.
I converted some of the tests, but not all of them.
I'm still not sure about removing `specificationSteps`. Can you explain a bit why you want it removed? It makes it much easier to follow which steps are being applied.
I see that at one point `Tag.java` was removed. Do you recall why? Because I used some of those tags. Did you replace it with something else?

radumereuta:
Tests have passed, @dwightguth you can review this now.

k-distribution/include/kframework/ktest-fail.mak
```diff
@@ -24,7 +24,7 @@ kompile: $(TESTS)
 dummy:
 
 %.k %.md: dummy
-	$(KOMPILE) $(KOMPILE_FLAGS) --backend $(KOMPILE_BACKEND) $(DEBUG_FAIL) $@ -d $(DEFDIR) 2>&1 | sed 's!'`pwd`'/\(\./\)\{0,1\}!!g' $(CHECK) $@.out $(CHECK2)
+	$(KOMPILE) $(KOMPILE_FLAGS) --backend $(KOMPILE_BACKEND) $(DEBUG_FAIL) $@ -d $(DEFDIR) 2>&1 | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' $(CHECK) $@.out $(CHECK2)
```
ehildenb, line 27: What's the explanation behind this one?

ehildenb:
@radumereuta the tests look good to me, but please have @dwightguth review the overall architecture.

k-distribution/include/kframework/ktest-fail.mak
radumereuta: On my machine, the paths from reported warnings include a `/././` and on Jenkins they only include `/./`.
I'm not sure why. It might be a java thing. This change takes into consideration both cases.

kernel/src/main/java/org/kframework/utils/Tag.java
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2012-2020 K Team. All Rights Reserved.
+package org.kframework.utils;
+
+public class Tag {
+
+    public final static String require = "require";
```
dwightguth: Let's not bring this whole file back when we're just using it to mark the constant strings that are part of the Bubble class. Just put them as static fields on that class.

dwightguth (commented):
What you have here mostly looks good, but I thought the plan was to remove the specificationSteps method?

dwightguth:
To clarify a little: I would rather we apply the same set of steps to all sentences and just have those steps do different things depending on whether something is a rule or a claim, than have a different set of steps during kprove than during kompile. Without making a change like this, it will be impossible to eventually in the long run have claims just be compiled by `kompile` and have kprove not compile anything, just perform proving.

kore/src/main/scala/org/kframework/compile/MergeRules.scala
```diff
@@ -30,6 +30,9 @@ class MergeRules(val automatonAttribute: String, filterAttribute: String) extend
   val isRulePredicate = KLabel("isRule")
 
   def apply(m: Module): Module = {
+    // TODO: find a better way to avoid adding merged rules to the spec module
+    if (m.claims.nonEmpty) return m
```
radumereuta: @dwightguth I need your input on this.
Since adding all steps to the kprove pipeline, `MergeRules` is being called and it creates a rule which the backend doesn't expect. The most simple solution I could find was to avoid this step on modules that contain claims.

radumereuta:
This should pass now.
All claims are being processed by the same main pipeline as used by kompile.
I'm curious about the performance impact since more steps are being applied over more modules.

dwightguth (approved):

PR #1554, opened 2020-09-23
support ffi_type_pointer in ffi.md

dwightguth:


ehildenb (approved):

ehildenb:
@dwightguth you may consider giving these `ffi.md` klabels that are prefixed with `ffi_` so that we aren't polluting the klabel namespace too much.

PR #1579, opened 2020-10-01
Check that constructors of builtin sorts are not declared

### LINKED ISSUE #1568: [Bug] [kompile] - LLVM backend crashes with non-intuitive error message when non-function/non-builtin symbols are used in side-conditions
K Version
---------

```
kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: 39291c7
Git branch: master
Build date: Wed Sep 30 14:09:24 CDT 2020
```

Description
-----------

Running `kompile` causes the LLVM backend to crash when non-function/non-builtin symbols are used in side-conditions.

The error message returned to the user talks about LLVM IR types and is unable to be linked to the root cause without experimental verification.

Input Files
-----------

`test.k`

```
module TEST-SYNTAX
  syntax Foo ::= "Foo" | "Bar"
endmodule

module TEST
  import TEST-SYNTAX
  import BOOL

  configuration <k> $PGM:Foo </k>

  syntax Bool ::= "myCond" // [function]

  rule <k> Foo => Bar </k>
    requires myCond
endmodule
```

Note: the author forgot to add the `[function]` attribute to `myCond` 

Reproduction Steps
------------------

```
kompile test.k --backend llvm
```

Expected Behavior
-----------------

The user should receive an error message which states that a non-function/non-builtin symbol was used in a side-condition. This lets the user know how to fix the problem.

This error should be generated for any backend that does not support this functionality.

dwightguth:
Fixes: #1568


drumenbassik (approved):

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -3,28 +3,57 @@
...
+            if (sortAtt.contains(Att.HOOK()) && !sortAtt.get(Att.HOOK()).equals("ARRAY.Array") && !(sortAtt.get(Att.HOOK()).equals("KVAR.KVar") && isSymbolicKast)) {
+                if (!prod.att().contains(Att.FUNCTION()) && !prod.att().contains(Att.BRACKET()) &&
+                    !prod.att().contains("token") && !(prod.klabel().isDefined() && macros.contains(prod.klabel().get()))) {
+                    if (!(prod.sort().equals(Sorts.K()) && ((prod.klabel().isDefined() && (prod.klabel().get().name().equals("#EmptyK") || prod.klabel().get().name().equals("#KSequence"))) || prod.isSubsort()))) {
+                        if (!(sortAtt.contains("cellCollection") && prod.isSubsort())) {
+                            errors.add(KEMException.compilerError("Cannot add new constructors to hooked sort " + prod.sort(), prod));
+                        }
+                    }
+                }
+            }
```
ehildenb, line 56: Can we instead have a pass which runs before this and does:

-  The same check, and if it passes adds the `constructor` attribute if it determines that the production is an attribute.
-  Then this check is changed to loop over all productions with `constructor` attribute and make sure none of them are targeting builtin sorts.

This has the benefit of (i) clearly defining what we call a "constructor", and (ii) adding the attribute to productions in case it's useful to the other various tools/backends.

dwightguth: There are several problems with this approach:

1. The checks normally happen before any transformation happens to the k definition. We might be able to change this, but it would be a substantial architectural change and I'd rather not make such a fundamental change on a completely ad-hoc basis. If we're going to do something like this, it would be better to have an explicit semantic analysis phase and lay oiut exactly how it should work and convey information.
2. Transformers that only add attributes to sentences don't really work. I've tried repeatedly to try to bypass this restriction and it's never gone well. It's well outside the scope of this change.
3. Your proposed solution would disallow sentences that aren't constructors that the user put the constructor attribute on in built-in sorts, which would be a breaking change (albeit one probably not exercised by anything). But it would be very confusing.

ehildenb (approved):

PR #1583, opened 2020-10-02
Warn if user imports module with not-lr1 attribute when doing LR(1) parsing

### LINKED ISSUE #1572: Add warning for importing non-LR-1 grammars into bison parser
This can be done by marking certain modules as non-LR-1.

This should be a _warning_, because users should still have flexibility to try things.

This comes from #1570

dwightguth:
Fixes #1572


ehildenb (approved):

ehildenb (changes requested):
Please add text to `pending-documentation.md` about this attribute in the section about parsing and the bison parser.

ehildenb (approved):

PR #1596, opened 2020-10-15
Apply minor fixes to kx

sskeirik:
kx command `-o` option is broken when not `pretty` or `kore` because of incorrect `kast` command invocation.


dwightguth (approved):
This looks good in terms of the changes itself, but it looks like this isn't the only place where kast is called in this script without a -d and you may also want to fix those.

dwightguth (approved):

PR #1597, opened 2020-10-15
Priority filter fix.

### LINKED ISSUE #1593: [Bug] kprove - Cannot use #Forall as a child of #And
K Version
---------

```
$ /home/andrei/kore/.build/k/bin/kompile  --version
RV-K version 1.0-SNAPSHOT
Git revision: 5e4e023
Git branch: UNKNOWN
Build date: Wed Sep 09 17:51:33 EEST 2020
```

Description
-----------

Running `kprove` with the following files causes a parsing error.

Input Files
-----------

`test.k`

```
module TEST
    imports BOOL
    imports INT

    syntax Exp ::= Int | Bool
                 | pair(Int, Int) [klabel(pair), symbol]

    syntax KItem ::= "#assume" Exp [klabel(assume), symbol]
    rule #assume true => .K
    rule #assume false => #Bottom

    syntax KItem ::= "#assert" Exp  [klabel(assert), symbol]
                   | "#fail"
    rule #assert true => .K
    rule #assert false => #fail

// Uninterpreted functions
    syntax Int ::= i(Int, Int)   [function, functional, no-evaluators, smtlib(fi), klabel(i)]
    syntax Int ::= p(Exp, Int)   [function, functional, no-evaluators, smtlib(fp), klabel(p)]
endmodule
```

`issue-2138.spec.k`

```
module VERIFICATION
  import TEST
endmodule

module ISSUE-2138-SPEC
  import VERIFICATION

  rule
    <k> #assert i ( 1 , 0 ) ==Int 0 => . </k>
      requires true #And #Forall X. { i ( X, 0 ) #Equals 0 }

endmodule
```

Reproduction Steps
------------------

Running the spec produces an error:

```
$ /home/andrei/kore/.build/k/bin/kprove --haskell-backend-command /home/andrei/kore/.build/kore/bin/kore-exec -d . -m VERIFICATION issue-2138-spec.k
[Error] Inner Parser: Priority filter exception. Cannot use #Forall as a child
of #And
	Source(/home/andrei/kore/test/issue-2138/././issue-2138-spec.k)
	Location(10,28,10,63)
```

Expected Behavior
-----------------

The spec should be run successfully.

radumereuta:
Fixes: #1593
Allow for left guarded and right guarded operators under binary operators

### DIFF
--- a/k-distribution/tests/regression-new/checks/ambError.k
+++ b/k-distribution/tests/regression-new/checks/ambError.k
@@ -1,4 +1,4 @@
-// issue #1593
+// Issue #1593
 // testing that priority between left guarded and
 // right guarded operators are properly disambiguated
 
--- a/k-distribution/tests/regression-new/checks/ambError.k
+++ b/k-distribution/tests/regression-new/checks/ambError.k
@@ -1,0 +1,21 @@
+// issue #1593
+// testing that priority between left guarded and
+// right guarded operators are properly disambiguated
+
+module AMBERROR-SYNTAX
+
+endmodule
+
+module AMBERROR
+  imports AMBERROR-SYNTAX
+    imports INT
+
+    syntax Exp ::= Int
+
+    syntax Exp ::= Exp "+" Exp [klabel(plus), symbol]
+                 > "l" Exp     [klabel(lguard), symbol]
+                 | Exp "r"     [klabel(rguard), symbol]
+    rule . => 1 + l 2 // non ambiguous, should parse
+    rule . => 1 r + 2 // non ambiguous, should parse
+    rule . => l 1 r   // ambiguous - error
+endmodule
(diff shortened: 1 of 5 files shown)


dwightguth:
I'm going to need to see some kind of peer reviewed source indicating that this is a safe change to make and won't introduce ambiguities before I agree to it.

radumereuta:
https://homepages.cwi.nl/~jurgenv/papers/SLE2013-1.pdf
Page 5, first paragraph you have a small example.
In the paper, they have a much more general approach which allows them to disambiguate more, but the algorithm is more complex. They duplicate productions and I think it might slow down our parser even further. I only picked up the part where left and right guarded productions are safe to use with binary operators.

I was a bit surprised by this issue since I thought this was the default behavior, but apparently, a long time ago I ran into some issues with FUN and forgot about it.

dwightguth:
So, the paragraph you refer me to tells me why the grammar we are encountering is not ambiguous and why the priority mechanism as written is problematic on this example, but it doesn't tell me that the solution that you provide here is safe.

radumereuta:
I'm not sure why this is failing, let's try again.

PR #1610, opened 2020-10-20
Pyk updates for gas expression manipulation

ehildenb:
These updates are inspired by the MKR regression test suite, where we are rewriting things like gas expression optimization directly in the pyk library instead of the external javascript library.

-   Print out the variable's `originalName` when it's available in `prettyPrintKast`.
-   Switch a bunch of double-quotes to single-quotes for readability.
-   Add a few routines to `kastManip`, including `whereMatchingTopDown`, `whereMatchingBottomUp`, `replaceWith`, and `replaceAnywhereWith`.


k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -66,6 +66,24 @@ def replace(k):
...
+def whereMattchingBottomUp(effect, matchPattern, pattern):
+    def _effect(k):
+        matchingSubst = match(matchPattern, k)
+        newK = k
+        if matchingSubst is not None:
+            newK = effect(matchingSubst)
+        return newK
+    return traverseBottomUp(_effect, pattern)
+
+def whereMattchingTopDown(effect, matchPattern, pattern):
```
gtrepta:
```suggestion
def whereMatchingBottomUp(effect, matchPattern, pattern):
    def _effect(k):
        matchingSubst = match(matchPattern, k)
        newK = k
        if matchingSubst is not None:
            newK = effect(matchingSubst)
        return newK
    return traverseBottomUp(_effect, pattern)

def whereMatchingTopDown(effect, matchPattern, pattern):
```

gtrepta (approved):

PR #1619, opened 2020-10-29
Fine tune function checks

### LINKED ISSUE #1591: [Bug] kompile rejects non-concrete simplification rule in the presence of concrete function rule
I am using K from fec13504826f02ea16cfd5638dc8b72a8b000340.

Description
-----------

`test.k`:

```.k
module TEST
  import INT

  // An uninterpreted function to trigger the simplification rule below.
  syntax Int ::= i(Int, Int) [function]

  syntax Bool ::= fun ( Int ) [function]

  // A concrete definition rule for our function. The content of this rule is
  // irrelevant, but being concrete forces all other definition rules to be
  // concrete.
  rule fun(X) => X >Int 0 [concrete]

  // A simplification rule defined over the uninterpreted function. This rule
  // should not be subject to the restriction on the concrete attribute because
  // it is not part of the definition of fun(_).
  rule fun(i(_, _)) => false [simplification]
endmodule
```

Running `kompile --backend haskell test.k` gives:

```
[Error] Compiler: Found concrete attribute without simplification attribute on
function with one or more non-concrete rules.
        Source(/home/ttuegel/evm-semantics/./test.k)
        Location(12,8,12,26)
```

Expected behavior: `kompile` should accept this definition.

Having the `concrete` attribute on one function definition rule constrains all other definition rules for the same function to have the `concrete` attribute, but there should be no constraint on `simplification` rules for the function.

See also: kframework/kore#2190

radumereuta:
Fixes: #1591


ehildenb:
@ttuegel the tests look like the correct behavior to me, but I think you should double-check them too to make sure the correct behavior was captured.

ttuegel (approved):
Looks good to me!

PR #1621, opened 2020-10-30
Anka cache builder

dwightguth:
Fixes runtimeverification/devops#24


dwightguth:
This does not seem sufficient to get rid of the intermittent errors, but it is still a marked improvement, so I'd like to get it merged.

package/macos/brew-install-deps
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash -ex
 
 brew update
-brew install maven cmake boost libyaml jemalloc llvm gmp mpfr z3 opam pkg-config flex bison
-
-curl -sSL https://get.haskellstack.org/ | sh
-
-k-distribution/src/main/scripts/bin/k-configure-opam-dev
+brew install maven cmake boost libyaml jemalloc llvm@8 gmp mpfr z3 opam pkg-config flex bison haskell-stack
```
ehildenb, line 4: Will this actually work with building the Mac build? When will ocaml be configured thenN

ehildenb: I'm worried that when the release goes tries to go through, building K will not work.

ehildenb (approved):

ehildenb:
I put an approval on this, @dwightguth . Please consider my concerns above about the OCaml backend tests not working on `master` builds, and if there are big changes (in addition to the above), ping me again.

PR #1623, opened 2020-10-30
Allow user to configure max stack depth of bison parser

### LINKED ISSUE #1602: Allow specifying max memory usage  for the Bison generated parser
Motivation
----------

I'm moving the KWasm semantics to the bison parser. I'm experimenting with some primitive versions of the grammar. When using parametric `List{Sort, Delimiter}` in the grammar and feeding the parser a large input file (~2Mb), it crashes with a "memory exhausted" error. Inspection with Valgrind reveals the total heap allocation is around ~12Mb. I'm willing to offer much more memory to the parser. The file contains a list of ~6000 elements, which I assume is causing the problem.

Using `--bison-lists`, `NeList` and adding an extra production for the empty list case everywhere the list is used seems to get around the issue.
```
syntax TypeDecls ::= NeList{TypeDecl, ""}
syntax TypeDefn ::= "(" "type" OptionalId "(" "func" ")" ")" // Empty case
                  | "(" "type" OptionalId "(" "func" TypeDecls ")" ")"
```

But this will be very invasive in the syntax and semantics.

According to the Bison documentation:

> By defining the macro YYMAXDEPTH, you can control how deep the parser stack can become before memory is exhausted. Define the macro with a value that is an integer. This value is the maximum number of tokens that can be shifted (and not reduced) before overflow.
> ...
> The default value of YYMAXDEPTH, if you do not define it, is 10000. 

https://www.gnu.org/software/bison/manual/html_node/Memory-Management.html

Example K Code
--------------

The easiest way would probably be to allow adding extra macros on the command line, rather than introducing one flag per relevant option.

```
kompile --gen-bison-parser --extra-bison-macros "YYMAXDEPTH=2000000,YYINITDEPTH=2000" myfile.k
```

Documentation
-------------

"You can specify extra macros to the generated Bison by passing the `--extra-bison-macros` with a comma-separated list of macros and values."

I can try to optimize the grammar further and avoid long cons-lists or deeply nested structures.
I can easily modify the Bison genarator locally and maintain my own K fork.

Testing Approach
----------------

Use a simple list grammar:
test.k:
```
module TEST
   imports DOMAINS

   syntax Pgm ::= List{Int, ""}
```

Then generate a parser:
```
kompile --gen-bison-parser test.k
```

Then one could (dynamically) feed the number `1` 99000 (or other suitable number) times into the parser expecting success, and again with 11000 numbers, verifying failure.

Then generate a new parser with new macros:
```
kompile --gen-bison-parser --extra-bison-macros "YYMAXDEPTH=20000" test.k
```

Ensure this succeeds with 11000 numbers, and fails with 21000 numbers.

dwightguth:
Fixes #1602


ehildenb (approved):

kernel/src/main/java/org/kframework/kast/KastOptions.java
```diff
@@ -106,6 +106,9 @@ public File outputFile() {
     @Parameter(names="--bison-file", description="C file containing functions to link into bison parser.")
     public String bisonFile;
 
+    @Parameter(names="--bison-stack-max-depth", description="Maximum size of bison parsing stack.")
+    public long bisonStackMaxDepth = 10000;
```
ehildenb:
```suggestion
    @Parameter(names="--bison-stack-max-depth", description="Maximum size of bison parsing stack (default: 10000).")
    public long bisonStackMaxDepth = 10000;
```

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -143,6 +143,9 @@ public boolean isKore() {
         @Parameter(names="--bison-file", description="C file containing functions to link into bison parser.")
         public String bisonFile;
 
+        @Parameter(names="--bison-stack-max-depth", description="Maximum size of bison parsing stack.")
+        public long bisonStackMaxDepth = 10000;
```
ehildenb:
```suggestion
        @Parameter(names="--bison-stack-max-depth", description="Maximum size of bison parsing stack (default: 10000).")
        public long bisonStackMaxDepth = 10000;
```

PR #1624, opened 2020-10-30
Remove mini kore

radumereuta:
Removing dead code.


kernel/src/test/java/org/kframework/unparser/KPrintTest.java
```diff
@@ -48,11 +39,13 @@ private String bytes2String(byte[] input) {
     }
 
     private String asKast(K term) {
-        return bytes2String(new KPrint().serialize(term, OutputModes.KAST));
+        new KPrint();
+        return bytes2String(KPrint.serialize(term, OutputModes.KAST));
```
ehildenb: Why is this change made? It actually doesn't make sense to me, because you're using an `static` method of `KPrint()`, so it seems that we shouldn't need to declare a `new` one.

ehildenb (approved):

kernel/src/test/java/org/kframework/unparser/KPrintTest.java
radumereuta: I used IntelliJ's autofix and didn't notice it added the new instantiation.
I've pushed a fix so please have another look.

PR #1637, opened 2020-11-10
Better defined ARRAY Lookup

gtrepta:
From the discussion in https://github.com/kframework/kore/issues/2253


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -159,7 +159,8 @@ of any of the syntax defined in any of these modules.
 
   rule makeArray(I::Int, D::KItem) => arr(.List, I, D)
 
-  rule arr(L::List, _, D::KItem) [ IDX::Int ] => #if IDX >=Int size(L) #then D #else L[IDX] #fi
+  rule arr(L::List, _, _       ) [ IDX::Int ] => L[IDX] requires IDX <Int size(L)
```
ehildenb: I think for completeness you need to say `0 <=Int IDX andBool IDX <Int size(L)`

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -159,8 +159,8 @@ of any of the syntax defined in any of these modules.
 
   rule makeArray(I::Int, D::KItem) => arr(.List, I, D)
 
-  rule arr(L::List, _, _       ) [ IDX::Int ] => L[IDX] requires IDX <Int size(L)
-  rule arr(_      , _, D::KItem) [ _        ] => D      [owise]
+  rule arr(L::List, _, _       ) [ IDX::Int ] => L[IDX] requires 0 <=Int IDX andBool IDX  <Int size(L)
+  rule arr(L::List, _, D::KItem) [ IDX::Int ] => D      requires 0 <=Int IDX andBool IDX >=Int size(L)
```
ehildenb: These rules don't cover the case that `IDX <Int 0` though, which we should cover I think to keep it total. I would stick with the `owise` rule.

ehildenb:
@gtrepta please also add some tests for this. A simple K definition in `k-distribution/tests/regression-new/array-haskell` should do the trick. Just add a simple semantics that uses arrays, and a handful of tests which are run with indices ranging from negative to zero to positive to out of bounds.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -159,7 +159,8 @@ of any of the syntax defined in any of these modules.
 
   rule makeArray(I::Int, D::KItem) => arr(.List, I, D)
 
-  rule arr(L::List, _, D::KItem) [ IDX::Int ] => #if IDX >=Int size(L) #then D #else L[IDX] #fi
+  rule arr(L::List, _, _       ) [ IDX::Int ] => L[IDX] requires 0 <=Int IDX andBool IDX  <Int size(L)
```
ttuegel, line 162: Should we look up the size from the second argument of `arr` instead? Computing `size(L)` is `O(log n)`.

gtrepta: I don't think the second argument represent the length of the list, it's the length of the array.

ehildenb: If it's the logical length of the array, then it should probably be taken into account? Because the list may be longer than the logical length of the array.

Maybe if you store 10 elements, then remove the last 5, it just doesn't bother removing the last 5 but instead just makes the second parameter 5 less?

Not sure, probably need some clarity from @dwightguth on how the Array works

dwightguth: so, the idea is that you can specify a length for an array but the List only creates elements up to the highest offset that was actually accessed. It's a lazy initialization. So the length of the array and the length of the list are not necessarily the same, but the list ought to never be longer than the array.

dwightguth (approved):

PR #1641, opened 2020-11-11
Lexical identifiers

### LINKED ISSUE #1633: [KIP] - Enable more modular token definitions using name expansion
Motivation
----------

It would be really useful to be able to use names of sorts of which tokens are defined within the regexes defining these tokens, in the same way a scanner allows name definitions and then name expansions in lexical rules. This would allow more modular token definitions and more readable specs, which can be especially helpful for semantics using multiple complex token definitions.

Example K Code
--------------

The following is a very simple example. In a language where the size of an integer is indicated by a prefix `n` for 64-bit and `u` for 128-bit integers, using name expansion, we would be able to do the following:

```
module TEST-SYNTAX
  syntax Num ::= r"[0-9]+" [token]
  syntax UInt64Token ::= r"n{Num}" [token]
  syntax UInt128Token ::= r"u{Num}" [token]
endmodule
```

Documentation
-------------

A syntactic production for a token type associates the sort's name with the regular expression defining that token, so that this name may be used in the regular expression defining other tokens. For example, suppose we have the following token definitions:

```k
  syntax Digit ::= r"[0-9]" [token]
  syntax Alpha ::= r"[a-zA-Z]" [token]
```

Then, the sort names `Digit` and `Alpha` may be used in the regex defining other tokens by enclosing the name with curly braces: `{Digit}` and `{Alpha}`, and these names will be expanded into their regexes above. For example, we can define a new token type:

```k
  syntax Alnum ::= r"{Alpha}|{Digit}" [token]
```
This process can be repeated. For example, using these definitions, an identifier token in a language can now be defined as:

```k
  syntax Identifier ::= r"{Alpha}({Alnum})*" [token]
```

I can think of at least two ways to implement this:

- Direct expansion at the level of K module processing, where the regexes themselves are expanded in place (so that the generated scanner does not have to be made aware of name abstractions)

- Injections of name definitions in the scanner (so that the regexes won't have to be expanded) 

Testing Approach
----------------

The above examples may be used for testing.

dwightguth:
Fixes #1633


k-distribution/tests/regression-new/issue-1633/lexical.k
```diff
@@ -0,0 +1,62 @@
...
+  syntax lexical UniversalCharacterName = r"(\\\\u{HexQuad})|(\\\\U{HexQuad}{2})"
+  syntax lexical HexQuad = r"{HexDigit}{4}"
+
+  syntax IntConstant ::= r"({DecConstant}|{OctConstant}|{HexConstant})({IntSuffix}?)" [token]
+  syntax lexical DecConstant = r"{NonzeroDigit}({Digit}*)"
+  syntax lexical OctConstant = r"0({OctDigit}*)"
+  syntax lexical HexConstant = r"{HexPrefix}({HexDigit}+)"
+  syntax lexical HexPrefix = r"0x|0X"
+  syntax lexical NonzeroDigit = r"[1-9]"
+  syntax lexical OctDigit = r"[0-7]"
```
ehildenb, line 26: Can these have attributes? If so, can you just put some dummy attributes on some of these to demonstrate that it parses correctly?

ehildenb (approved):

k-distribution/tests/regression-new/issue-1633/lexical.k
dwightguth: They cannot currently have attributes, similar to some of the other auxiliary syntax sentences like `syntax priorities`. We can change this if we want, but right now no attributes on such a sentence have any meaning whatsoever, so it seems premature.

PR #1649, opened 2020-11-13
fix concrete function test to not expand builtin sort

daejunpark:
Fix the test to not violate the restriction that new constructors not be added to builtin sorts.


dwightguth (approved):

daejunpark:
Added a comment describing the intention of tests.

We want to test if the (java) backend to distinguish the following two cases:
- functional terms with no variables: _not_ concrete
- constructor terms with no variables: concrete

(Of course, any terms with variables are _not_ concrete.)

PR #1652, opened 2020-11-16
Fix the dark mode and switch the syntax highlighter from highlight.js to prism.js

shd101wyy:
Demo: https://shd101wyy.github.io/k/

Screenshot from 2020-11-16 21-30-47


web/prismjs/k.js
```diff
@@ -0,0 +1,25 @@
...
+        lookbehind: true,
+        greedy: true,
+      },
+    ],
+    string: {
+      pattern: /(["'])(?:\\(?:\r\n|[\s\S])|(?!\1)[^\\\r\n])*\1/,
+      greedy: true,
+    },
+    "class-name": /\b(?:strict|avoid|prefer|bracket|non-assoc|seqstrict|left|right|macro|token|notInRules|autoReject|structural|latex|binder|klabel|format)\b/,
+    keyword: /\b(?:syntax|rule|Id|Int|Bool|String|configuration|imports|require|Token|Lexer|Float|requires|Kresult|context|module|endmodule)\b/,
```
ehildenb:
```suggestion
    keyword: /\b(?:syntax|syntax lexical|syntax priorities|rule|Id|Int|Bool|String|configuration|imports|require|Token|Lexer|Float|requires|Kresult|context|module|endmodule)\b/,
```

You need to inspect the bison outer parser for all the keywords (do `find ./ -name Outer.jj` in the K repo).

dwightguth (approved):
I'm approving because the changes to the documentation look fine, but it would be good if someone with more experience with the web stuff could review also.

ehildenb (approved):

PR #1666, opened 2020-11-20
Add option --emit-json to kprove as well

ehildenb:
This enables automatically generating (in `pyk`) the proof obligations needed to show that lemmas used for verification are sound.

### DIFF
--- a/kernel/src/main/java/org/kframework/kprove/KProve.java
+++ b/kernel/src/main/java/org/kframework/kprove/KProve.java
@@ -2,7 +2,6 @@
 package org.kframework.kprove;
 
 import com.google.inject.Inject;
-import org.kframework.RewriterResult;
 import org.kframework.attributes.Source;
 import org.kframework.definition.Definition;
 import org.kframework.definition.Module;
@@ -10,8 +9,10 @@ import org.kframework.definition.Rule;
 import org.kframework.kompile.CompiledDefinition;
 import org.kframework.kompile.KompileOptions;
 import org.kframework.krun.KRun;
+import org.kframework.RewriterResult;
 import org.kframework.rewriter.Rewriter;
 import org.kframework.unparser.KPrint;
+import org.kframework.unparser.ToJson;
 import org.kframework.utils.BinaryLoader;
 import org.kframework.utils.errorsystem.KEMException;
 import org.kframework.utils.errorsystem.KExceptionManager;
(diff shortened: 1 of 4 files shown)


kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -82,4 +82,7 @@ public synchronized File specFile(FileUtil files) {
 
     @Parameter(names="--debug-script", description="Run script passed in specified file when the debugger starts. Used with --debugger.")
     public String debugScript;
+
+    @Parameter(names="--emit-json", description="Emit JSON serialized def module and spec module.")
```
ehildenb:
```suggestion
    @Parameter(names="--emit-json", description="Emit JSON serialized main definition for proving.")
```

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -72,6 +74,14 @@ public int run() {
         Module specModule = compiled._2();
         Rule boundaryPattern = buildBoundaryPattern(compiledDefinition);
 
+        if (kproveOptions.emitJson) {
+            try {
+                files.saveToKompiled("def-module.json", new String(ToJson.apply(compiled._1()), "UTF-8"));
```
radumereuta: I have a few issues here:
1. `compiled._1()` returns an entire definition, not only a module, so the name of the file is not really correct.
2. Right now, each time you call `kprove` you may have a different set of claims, so maybe it would be better if you save this in `kprove.temp`. I know we want to eventually optimize this step, so it might not matter then. Thoughts?

ehildenb: Hmmmm, saving in `kprove.temp` may be difficult, because I need to access the file afterward. How about I let the user specify the location with `--emit-json-file some_file_name`, and otherwise default to `*-kompiled/kprove-definition.json`?

radumereuta: That sounds even more complicated. I'm happy with where it saves right now if we are to fix kprove eventually.

ehildenb: Ok, it works for me for now anyway then.

radumereuta: But you should still change the name of the file to reflect that you have a definition inside, not just a module.

ehildenb: done

radumereuta (approved):

PR #1671, opened 2020-11-27
Fix #1646 - error message for simplification attribute

### LINKED ISSUE #1646: Error message for missing simplification attribute is confusing
With the following definition:

```
require "domains.md"

module TEST
    imports INT
    imports BOOL

    syntax Int ::= bool2Word ( Bool ) [function]
 // --------------------------------------------
    rule bool2Word(true)  => 1
    rule bool2Word(false) => 0

    rule I <=Int bool2Word(_) => false requires 1 <Int I
endmodule
```

I get the following error message:

```
[Error] Compiler: Illegal function symbol on LHS of rule.
        Source(/home/dev/src/k/./test.k)
        Location(14,18,14,30)
```

Which can be improved by adding "consider adding `simplification` attribute to the rule if this is intended".

With the following definition:

```
require "domains.md"

module TEST
    imports INT
    imports BOOL

    syntax Int ::= bool2Word ( Bool ) [function]
 // --------------------------------------------
    rule bool2Word(true)  => 1
    rule bool2Word(false) => 0
endmodule
```

We can call `kompile --backend haskell test.k`, and then try to prove the following specification:

```
require "test.k"

module VERIFICATION
    imports TEST

    rule I <=Int bool2Word(_) => false requires 1 <Int I
endmodule

module TEST-SPEC
    imports VERIFICATION
 
    claim <k> 3 => 5 ... </k>
endmodule
```

With `kprove --def-module VERIFICATION test-spec.k`.

It gives the following error message:

```
[Error] Compiler: Illegal function symbol on LHS of rule.
        Source(/home/dev/src/k/././test-spec.k)
        Location(6,18,6,30)
[Warning] Critical: Illegal function symbol on LHS of rule.
Note: this warning will become an error in subsequent releases.
        Source(/home/dev/src/k/././test-spec.k)
        Location(6,18,6,30)
[Error] Compiler: Had 1 structural errors after macro expansion.
  while executing phase "expand macros" on sentence at
        Source(/home/dev/src/k/././test-spec.k)
        Location(6,10,6,57)
```

Which is confusing for several reasons:

-  It mentions `macro expansion` as an Error, but there are no macros present in the definition.
-  It lists the illegal function symbol on the LHS as both an error and a warning.

radumereuta:
Fixes: #1646


k-distribution/tests/regression-new/kprove-java/rhs-vars-broken-spec.k.out
```diff
@@ -1,15 +1 @@
...
-	Location(10,20,10,21)
-[Warning] Compiler: Variable 'A' defined but not used. Prefix variable name
-with underscore if this is intentional.
-	Source(rhs-vars-broken-spec.k)
-	Location(12,20,12,21)
-[Warning] Critical: Found variable A on right hand side of rule, not bound on
-left hand side. Did you mean "?A"?
-Note: this warning will become an error in subsequent releases.
-	Source(rhs-vars-broken-spec.k)
-	Location(12,20,12,21)
```
ehildenb: It seems to me this should be part of the test? Seems it's testing broken variables?

kernel/src/main/java/org/kframework/compile/checks/CheckRewrite.java
```diff
@@ -130,7 +131,7 @@ public void apply(KApply k) {
                 }
             }
         }.accept(body);
-        if (!h.hasRewrite) {
+        if (!h.hasRewrite && !isClaim) {
             errors.add(KEMException.compilerError("Rules must have at least one rewrite.", body));
         }
```
ehildenb, line 136: Are we not enabling this check on `claim`? They should have at least one rewrite as well, I would think?

radumereuta: There are several claims which don't contain rewrites:
https://github.com/kframework/k/blob/master/k-distribution/tests/regression-new/kbmc/success-test1-spec-bmc.k#L13
There are a couple more but I can't find them right now.

k-distribution/tests/regression-new/kprove-java/rhs-vars-broken-spec.k.out
radumereuta: @traiansf you are the author of this test. Is this what you had in mind?

traiansf:
Yes, my main thought here was to check that the correct errors are actually being reported.

ehildenb (approved):

PR #1680, opened 2020-12-01
`kast -` reads from stdin

### LINKED ISSUE #582: Have `kast` default to reading from `stdin` if no term is passed.
This would make it easier to get the K frontend to do unparsing for the Haskell backend without having to dump the results to an intermediate file, for example.

dwightguth:
Fixes #582


ehildenb:
Will this also allow `krun -`? I'd like that as well.

dwightguth:
It doesn't appear so, primarily because stdin from krun isn't piped to stdin from kast. That would probably be difficult to do while maintaining support for the existing behavior of piping stuff to krun. I'm not sure it's a change we can make without breaking the tutorial.

This is ready for review, though.

ehildenb (approved):

PR #1681, opened 2020-12-02
Warn for rogue syntax in proof module

### LINKED ISSUE #1471: Add warning for syntax declared between main module and verification module
In preparation for #1333, we should have a period where the check for syntax declared between main module and verification module is warned for, so that people have time to adjust their definitions.

Currently this is already a problem for the `kore-repl`, because if you unparse a term which uses syntax declared in such a way, it will give a `NoSuchElementException`

radumereuta:
Fixes: #1471


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -313,6 +313,14 @@ public void definitionChecks(Set<Module> modules) {
 
     // Extra checks just for the prover specification.
     public Module proverChecks(Module specModule, Module mainDefModule) {
+        // check rogue syntax in spec module
+        Set<Sentence> toCheck = mutable(specModule.sentences().$minus$minus(mainDefModule.sentences()));
+        for (Sentence s : toCheck)
+            if (s instanceof Production || s instanceof SyntaxSort || s instanceof SyntaxLexical ||
+                    s instanceof SyntaxPriority || s instanceof SyntaxAssociativity)
```
ehildenb: There isn't some `Syntax` super-class you can use instead of listing all these out?

@dwightguth should probably check that this check is being done in the correct place, output looks OK to me.

radumereuta:
```
case class Production(klabel: Option[KLabel], params: Seq[Sort], sort: Sort, items: Seq[ProductionItem], att: Att)
  extends Sentence with ProductionToString
```
They all extend `Sentence` directly, I had to list them all.

radumereuta: Would this message be better?
`Found syntax declaration in proof module. This can cause parsing issues since it's not visible from the main module.`

dwightguth: rather than using a `instanceof`, you ought to be able to say `sentence.isSyntax()`

ehildenb (approved):

PR #1707, opened 2020-12-14
Issue1682 kore pp

### LINKED ISSUE #1682: [Bug] krun unable to unparse symbolic ints to kore `(NoSuchElementException: key not found: #SemanticCastToInt)`
K Version
---------

```
RV-K version 1.0-SNAPSHOT
Git revision: ada1ae8
Git branch: UNKNOWN
Build date: Sun Nov 29 17:43:14 CST 2020
```

Description
-----------

See title

Reproducing
-----------

Running this definition with `--output kore` crashes.

```
module TEST
    imports INT
    configuration <k> .K </k>
    rule <k> .K => ?_:Int </k> 
endmodule
```

```
$ krun  
<k>
  'QuesUnds'0:Int ~> .
</k>
$ krun  --output kore
file a bug report at https://github.com/kframework/k/issues
(NoSuchElementException: key not found: #SemanticCastToInt)
```

Expected Behavior
-----------------

Correct kore output

radumereuta:
Fixes: #1682


dwightguth (changes requested):
I would really like to see variables still be annotated by their sort during unparsing; you can just add the cast during KOREToTreeNodes instead of KOREToK

radumereuta:
@dwightguth I've done the change, and there are over 700 files that need to be updated with sort annotations.
That was the only file that had them, not sure why.
Are you still sure we want this?

ehildenb:
Let's preserve the old behavior with the simple annotation (so that we (i) are not making kprove output more verbose, and (ii) we can still correctly call ModuleToKore).

We can do this by making a separate attribute to indicate that a variable needs to be annotated at unparse time because of an injection that lives above it.

dwightguth (approved):

PR #1723, opened 2020-12-18
Fix Nix job on pull requests from forks

ttuegel:
It turns out that the git-commit action works just fine with forks if we check out the pull request HEAD, instead of the merge commit. That's what we want to do anyway! :tada:


ttuegel:
The trick is: the Action to automatically update dependencies is triggered by a push event on the fork. This works as long as the user hasn't disabled Actions.

ttuegel:
I have also disabled the installation test temporarily, until I have time to deal with the caching issue.

dwightguth (approved):

PR #1726, opened 2021-01-04
Expand macros in kx script

### LINKED ISSUE #1648: Separated macro expander
Currently we cannot use the bison parser for semantics which use macros because macro expansion is very closely tied to the Java parser. So we need to produce a new tool which just does macro expansion which the Bison parser can use directly.

There are two main approaches we have discussed:

-   Build a standalone tool over the LLVM (or Haskell) backend data-structures that takes a list of Kore rules (the macro rules) and directly expands them over an input kore term.
-   Build a K definition which does macro expansion, and feed that to whichever backend to do the expansion.

dwightguth:
This ought to be the last change needed in order to be able to change kx -> krun and krun -> krun-legacy. I will do that in a separate PR after this is merged, though.

Fixes #1648

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -271,13 +271,22 @@ public class ModuleToKORE {
         }
 
         sb.append("\n// rules\n");
+        StringBuilder macrosSb = new StringBuilder();
+        macrosSb.append("// macros\n");
         int ruleIndex = 0;
         ListMultimap<Integer, String> priorityToAlias = ArrayListMultimap.create();
         for (Rule rule : sortedRules) {
-            convertRule(rule, ruleIndex, heatCoolEq, topCellSortStr, attributes, functionRules,
-                    priorityToPreviousGroup, priorityToAlias, sentenceType, sb);
+            if (ExpandMacros.isMacro(rule)) {
+                convertRule(rule, ruleIndex, heatCoolEq, topCellSortStr, attributes, functionRules,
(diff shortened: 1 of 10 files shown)


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1168,6 +1183,24 @@ private void convertRule(RuleOrClaim rule, int ruleIndex, boolean heatCoolEq, St
...
+            sb.append("} ");
+            sb.append("\\equals{");
+            sb.append(ruleInfo.productionSortStr);
+            sb.append(",R} (\n    ");
+            convert(left, sb);
+            sb.append(",\n    ");
+            convert(right, sb);
+            sb.append(")\n  ");
+            convert(consideredAttributes, rule.att(), sb, freeVarsMap, rule);
+            sb.append("\n\n");
```
ehildenb, line 1203: It looks like here we're including the macro rules in the final definition.kore (in addition to putting them in macros.kore). Is that correct? What's the reasoning?

ehildenb (approved):

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: No, if you look at the above if condition, we are controlling which file we emit rules to based on whether that rule is a macro.

ehildenb: Ah sorry, I missed that this code was inside `convertRule`. Thanks!

PR #1661, opened 2020-11-18
Haskell symbolic lookup rule

ana-pantilie:
Fixes https://github.com/kframework/kore/issues/2247

PR #1732, opened 2021-01-06
Re-apply "Haskell symbolic lookup rule (#1661)"

ttuegel:
This re-applies commit 39d8802.


ttuegel:
@dwightguth I marked the new tests with `.NOTPARALLEL` as you suggested. I ran the tests several times on Jenkins without being able to reproduce the error. I think that we started to see problems just because we added so many new Haskell backend tests.

dwightguth (approved):

dwightguth:
Looks good. Sounds like you weren't able to reproduce the failure we got, so it makes sense to wait and see if it happens again with this tweak.

PR #1708, opened 2020-12-14
Warnings and errors for cmd line vars

radumereuta:
Fixes: #1683

PR #1733, opened 2021-01-06
make kx the default krun on llvm and haskell backends

dwightguth:
This consists of essentially 3 changes:
1. We rename krun -> krun-legacy and kx -> krun
2. We modify the new krun so that it works on any backend by means of calling to the old krun when on a backend it doesn't support
3. We modify the test suite to make everything work and use krun as the command invoked by the test harness.

The main user-visible changes:
1. People using options supported by krun-legacy but not krun (primarily very old, deprecated options) will need to switch to using krun-legacy instead of krun. If you think the option needs to be supported by the new script, you should let me know your use case.
2. People using the llvm backend or haskell backend will switch to the new runner, which should behave exactly the same as the old runner, but is generally faster, especially if you have a bison parser. If the behavior is different, this is probably a bug. You can always switch back to the old runner by invoking krun-legacy instead.
3. People on the llvm or haskell backend that use external parsers should generate KORE instead of KAST, JSON, or binary. You can always invoke `kast -i kast|json|binary -o kore` to perform this conversion, but this will negate the performance improvements you would otherwise gain.


dwightguth:
This is a potentially breaking change so I'd like to get as much feedback as possible about how else we can improve things, and we should probably also do some kind of announcement and maybe like a 5.0.1 tag before it's merged so people know how to stay productive if this breaks something for them.

I'd welcome thoughts on what else we can do to smooth the rollout of this feature.

ttuegel:
I tried the new `krun` with the Haskell backend's integration test suite. Some tests are failing with the following message when the `configuration` does not contain the `$PGM` variable:

```
$ /home/ttuegel/kore/.build/k/bin/krun \
    --haskell-backend-command /home/ttuegel/kore/.build/kore/bin/kore-exec \
    -d . 1.test --output-file 1.test.out
cat: ./test-kompiled/sort_PGM.txt: No such file or directory
```

Obviously, we can change the definitions to include `$PGM`, but I thought this should be documented somewhere.

ttuegel:
I will need to make a small change to the Nix package: the frontend package currently has an optional dependency on the LLVM backend, but now it will be a required dependency. Should I push that change to this branch, or do you want me to make a separate pull request that we will merge before this one?

radumereuta:
`kx` should be able to run even if the configuration doesn't have a `$PGM`. Don't change the test.
Can you post here a simple test that fails?

dwightguth:
I have updated the PR so that an error is reported if a configuration variable is passed that does not exist, like was done here. This really should have been an error since the user is passing in something that is garbage and gets ignored, but it was just silently ignoring the variable in the past. The solution is to just not pass in the positional argument if the $PGM variable doesn't exist.

radumereuta:
@dwightguth I think your last commit overlaps with my changes: #1708

ehildenb:
Let's get #1708 and #1746 merged before this one.

ehildenb:
@dwightguth you can pull out the changes to the DisambModule and the KSyntax2Bison files, as well as the changes to `imp` in the tutorial as a separate PR, so that this one becomes pure KX => KRUN changes.

dwightguth:
Apologies for force-pushing to this branch; it turns out that `git merge` apparently doesn't support the diff options necessary to intelligently merge my changes with Radu's pull request, so I had to do a rebase instead. I went ahead and removed the changes to the Java code from this PR. The changes to the krun script themselves will be a lot harder to separate out because of the limitations of git; I would probably have to manually convert each commit to a patch, manually edit the patch, and then manually reapply the commits to a different branch. I can do it if you want, but note that reviewing this PR one commit at a time ought to allow you to see those changes easily as separate from the rename itself.

ehildenb (approved):

PR #1743, opened 2021-01-11
update tutorial 1_lambda to use Bison

dwightguth:


dwightguth:
Needs to be merged in tandem with kframework/k-exercises#22

k-distribution/tutorial/1_k/3_lambda++/lesson_5/Makefile.det
```diff
@@ -3,6 +3,7 @@ TESTDIR=tests/deterministic
 DEFDIR=det
 KOMPILE_BACKEND=java
 include $(MAKEFILE_PATH)/../../1_lambda/lesson_1/Makefile
+KOMPILE_FLAGS=
```
radumereuta, line 7: Is this needed?

k-distribution/tutorial/1_k/3_lambda++/lesson_6/lambda.md
```diff
@@ -43,6 +43,7 @@ We move all the LAMBDA++ syntax here.
                | Exp Exp              [strict, left]
                | "(" Exp ")"          [bracket]
 // Arithmetic
+               > "-" Int
```
radumereuta, line 46: Why did you decide to add unary minus? The original definition didn't have it.
Dito for the other lessons.

radumereuta (commented):
Two questions.
Looks good otherwise.

k-distribution/tutorial/1_k/3_lambda++/lesson_5/Makefile.det
dwightguth: Yes; it disables the bison parser on a lesson that compiles with the Java backend and thus cannot use it.

k-distribution/tutorial/1_k/3_lambda++/lesson_6/lambda.md
dwightguth: The reason is that we are importing DOMAINS-SYNTAX instead of DOMAINS, which imports UNSIGNED-INT-SYNTAX instead of INT. As a result, there is no other way of expressing negative integers aside from this production. While it may have seemed to work previously, you would have still ended up with parser errors resulting from the lexer/parser ambiguity the old grammar had. The tests simply didn't exercise this case. So I elected to fix all the definitions so that they appropriately describe negative integers in an unambiguous fashion.

k-distribution/tutorial/1_k/3_lambda++/lesson_5/Makefile.det
radumereuta: Can you add a simple comment to this line to state that? It's not obvious and might cause headaches in the future.
Something like:
`# using the java backend so disable bison parser`

radumereuta (approved):
lgtm, other than the single comment

Jenkinsfile
```diff
@@ -54,7 +54,7 @@ pipeline {
                     }
                   }
                   stages {
-                    stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git' } } }
+                    stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git', branch: 'lambda' } } }
```
ehildenb: reminder to fix this once ready for merge

k-distribution/tutorial/1_k/3_lambda++/lesson_5/Makefile.det
ehildenb: Seconded adding the comment

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1199,7 +1200,7 @@ private void convertRule(RuleOrClaim rule, int ruleIndex, boolean heatCoolEq, St
             sb.append(",\n    ");
             convert(right, sb);
             sb.append(")\n  ");
-            convert(consideredAttributes, rule.att(), sb, freeVarsMap, rule);
+            convert(consideredAttributes, rule.att().add(Att.PRIORITY(), Integer.toString(getPriority(rule.att()))), sb, freeVarsMap, rule);
```
ehildenb, line 1203: This means  that the backends no longer need to calculate priority themselves anymore, right?

ehildenb (approved):

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
dwightguth: This is just for macros. The other code for non macros is still in place for now, primarily because the llvm backend computes priorities slightly different than the haskell backend and the k frontend currently.

PR #1744, opened 2021-01-12
replaced the values of unit and element attributes from K labels to KORE labels

### LINKED ISSUE #1614: Kore symbol attributes should refer to Kore symbols, not K labels
Certain Kore symbol attributes produced by the frontend refer to K label names instead of Kore symbol names. For example, the declaration for the built-in `_Map_` symbol is:

```.kore
hooked-symbol Lbl'Unds'Map'Unds'{}(SortMap{}, SortMap{}) : SortMap{}
    [unit{}(".Map"), element{}("_|->_"), hook{}("MAP.concat"), /* elided */ ]
```

Note that the `unit` and `element` symbols are referenced by their K labels, which are not canonical in Kore. It would be more useful to the backend if these were listed by their Kore symbols, which are canonical, i.e.

```
    [unit{}(Lbl'Stop'Map{}), element{}(Lbl'UndsPipe'-'-GT-Unds'{}), /* elided */]
```

@dwightguth Is this reasonable?

MirceaS:
Fixes #1614


kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -368,6 +368,15 @@ private void translateSorts(Set<SortHead> tokenSorts, Map<String, Boolean> attri
 
     private void translateSymbols(Map<String, Boolean> attributes, SetMultimap<KLabel, Rule> functionRules,
                                   Set<KLabel> impurities, Set<Production> overloads, StringBuilder sb) {
+        HashMap<String, String> KLabelToKORE = new HashMap<String, String>();
```
dwightguth: I don't understand why we're using this instead of just calling `production` on the klabel to get the production corresponding to the klabel and then calling convert directly in the function for converting attributes. It seems simpler to me than constructing this map.

MirceaS: @dwightguth 
From what I can see, `production` takes as argument a `KApply`. All we have in this scenario is a string denoting the KLabel. I couldn't figure out a way to extract a `Production` from that alone...

dwightguth: So, you can do something like `KApply(KLabel(str))`

dwightguth (approved):

PR #1750, opened 2021-01-15
Bugfix for bison parser

dwightguth:


ehildenb:
@dwightguth is there a reasonable regression test you can include for this bugfix?

radumereuta:
Is this supposed to fix an issue? Or is it just something you noticed while you were working on something else?
Looks good otherwise.

k-distribution/tests/regression-new/bison-glr-bug/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+test: test.kore
+	cat test.kore | diff - test.ref
+
+test.kore: forwarder.iele kompile
+	./iele-kompiled/parser_PGM forwarder.iele > test.kore
+include ../../../include/kframework/ktest.mak
+
+clean:
+	rm -rf test.kore test-kompiled
```
radumereuta: This is missing some file patterns.
I've updated the list in my warnings PR:
`rm -rf $(KOMPILED_DIR) .depend-tmp .depend .krun-* .kprove-* kore-exec.tar.gz`
Is it possible to delete test.kore and still call the command from ktest.mak?

radumereuta (changes requested):

k-distribution/tests/regression-new/bison-glr-bug/Makefile
dwightguth: You're welcome to suggest how to do that, but I don't really know myself.

radumereuta (approved):

PR #1755, opened 2021-01-19
Fix warnings

### LINKED ISSUE #1758: [Bug] wrong warning for HOLE in contexts
```
context alias [foo]: HERE
    requires isFoo(HOLE)
/home/radu/work/k/k-distribution/bin/kompile --syntax-module TEST --backend llvm  test.k -d .
[Warning] Compiler: Variable 'HOLE' defined but not used. Prefix variable name
with underscore if this is intentional.
	Source(/home/radu/work/k/k-distribution/tests/regression-new/context-alias-3/./test.k)
	Location(7,20,7,24)
```
dwight.guth  01:07
HOLE in contexts and context aliases and HERE in context aliases should probably be ignored for the purposes of this check

radumereuta:
Fixing some of the warnings from the integration tests.
Fixes: #1758
There are still a few issues that are not being handled here.
- [Warning] Compiler: Found syntax declaration in proof module. This will not be visible from the main module.
  I will leave these messages for kprove fixes
- I added a few issues to the haskell backend with missing location information on some warnings
- #1760 not being handled by this PR either - waiting for feedback there


kernel/src/main/java/org/kframework/compile/checks/CheckAnonymous.java
```diff
@@ -94,11 +94,12 @@ public void check(Sentence s) {
                     if (s instanceof ContextAlias && entry.getElement().equals("HERE")) {
                         continue;
                     }
-                    if (s instanceof Context && entry.getElement().equals("HOLE")) {
+                    if (s instanceof Context || s instanceof ContextAlias && entry.getElement().equals("HOLE")) {
```
dwightguth: parentheses here please

dwightguth (changes requested):
looks quite good. I had one minor suggestion.

dwightguth (approved):

PR #1763, opened 2021-01-23
map-symbolic-tests-haskell: Run tests in parallel

### LINKED ISSUE #1748: [Bug] regression tests not parallel are slow
Recently have been added a new set of tests which have the option:
```
# There are too many memory-intensive tests to run in parallel.
.NOTPARALLEL:
```
https://github.com/kframework/k/blob/master/k-distribution/tests/regression-new/map-symbolic-tests-haskell/Makefile#L2

I run the integration tests a lot and this makes it really slow for my workflow.

Is there an option to disable parallelization only on jenkins?
@ttuegel said there are some problems with some of the machines.

I can comment that line out, but doing so each time I switch to a new branch is just going to be frustrating and error-prone.

ttuegel:
Fixes #1748.


ttuegel:
I set my workstation to trigger a build on this branch every six hours, so we should get plenty of logs to examine.

ttuegel:
Because I am very clever, I forgot that, in order to automatically push commits, I have to actually be at my desk to enter my password. Still, I can kick off a new build periodically.

radumereuta:
What's very interesting is that it's java that runs out of memory:
`java.lang.OutOfMemoryError: GC overhead limit exceeded`
Does this pass if run with the new `kx` as the default script?

ehildenb:
Can we try increasing the max heap size available to kserver when running the test-suite.

Something like `export K_OPTS=--Xmx16G` before launching kserver (play with the 16G amount, that may be high).

Change either in the pom.xml (more permanent), or on the Jenkinsfile.

Maybe this can be set to only run on `mvn verify` or directly the `make`? This depends on the number of threads, so maybe it's Jenkins specific.

Let's test just by modifying the Jenkinsfile first, and see if it fixes it.

ttuegel:
We were already setting `-Xmx8G` for the K tutorial tests. I moved that setting earlier in the Jenkinsfile so that it applies to the regular test suite, too.

ttuegel:
I'm now getting consistent failures in the tutorial tests, and I have no idea why.

dwightguth (approved):

ehildenb (approved):

PR #1769, opened 2021-01-28
Adding error message for malformed bracket productions

### LINKED ISSUE #1764: Check that the bracket attribute is used correctly
Productions with `bracket` attribute have to:

-  Be productions with a single non-terminal.
-  Have a return sort the same as the non-terminal sort.

Are allowed to be parametric (in which case you need to check that the parameter sorts are the same).

No restrictions on terminals in the production.

radumereuta:
Fixes: #1764
Cleaned up some dead code also

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkBracket.k
+++ b/k-distribution/tests/regression-new/checks/checkBracket.k
@@ -1,0 +1,10 @@
+module CHECKBRACKET-SYNTAX
+  syntax {A} A ::= "(" A ")" [bracket]
+  syntax Exp ::= "(" Exp ")" [bracket]
+  syntax Exp ::= "(" Exp Exp ")" [bracket]
+endmodule
+
+module CHECKBRACKET
+  imports CHECKBRACKET-SYNTAX
+
+endmodule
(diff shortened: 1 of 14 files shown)


radumereuta:
The tests failed because of `regression-new/parse-c`:
```
  syntax DirectDeclarator ::= Identifier
                            | "(" Declarator ")" [bracket]
```
Apparently, there are legitimate uses of `bracket` even when the inner sort is different from the production sort.
The problem is they break typing.
Thoughts?

k-distribution/tests/regression-new/checks/checkBracket.k
```diff
@@ -0,0 +1,10 @@
+module CHECKBRACKET-SYNTAX
+  syntax {A} A ::= "(" A ")" [bracket]
+  syntax Exp ::= "(" Exp ")" [bracket]
+  syntax Exp ::= "(" Exp Exp ")" [bracket]
+endmodule
```
ehildenb: Please also add some examples of bracket productions that are ill-formed in the following ways:

-   The thing inside the bracket is a subsort of the final sort.
-   The thing inside the bracket is a supersort of the final sort.
-   The thing inside the bracket is an unrelated sort to the final sort.

Also add these tests which should not be errors:

-  The bracket is not a surrounding production like above, but some weird left only, right only, or mixed operator. Add a few of these.
-  The bracket is not traditional bracket characters like `(` or `[`, but something ilke `%`.

dwightguth (approved):
lgtm

PR #1772, opened 2021-01-28
optimize memory usage of kompile

dwightguth:


radumereuta:
Can you explain how this is supposed to work?

ehildenb:
Agree a brief explanation would be nice. But I also want to know how much of an affect this has on memory usage?

ehildenb (approved):

dwightguth:
I haven't measured memory usage exactly, but basically, my experience with Java is that the JVM doesn't necessarily mark local method variables as dead immediately after their last use in the stack frame; instead, it waits until the end of the stack frame to mark them as dead. This means that the CompiledDefinition was still considered "live" during the decision tree generation process, even though it was never used after that point. By explicitly marking the relevant references as null, it allows the garbage collector to free all that memory sooner, which means that peak memory usage of the decision tree generator is less.

Issue #1333, opened 2020-06-11
Low overhead for `kprove` (call `kompile` ahead of time)


denis-bogdanas:
File `demo.k`
```
module DEMO
  imports DOMAINS
  configuration <k> .K </k>
endmodule

module EXTENDED
  imports DEMO
  configuration <extended> <k/> </extended>
endmodule
```

File `extended-bad-module-spec.k`
```
requires "demo.k"

module VERIFICATION-DEMO
    imports DEMO
endmodule

module VERIFICATION-EXTENDED
    imports EXTENDED
    imports VERIFICATION-DEMO
endmodule

module EXTENDED-BAD-MODULE-SPEC
    imports VERIFICATION-EXTENDED

    rule <k> .K => ?_ </k>
endmodule
```
File `Makefile`
```
DEF=extended
EXT=test
TESTDIR=.
KOMPILE_BACKEND=haskell
KPROVE_FLAGS=--def-module VERIFICATION-DEMO

export KOMPILE_BACKEND
export KPROVE_FLAGS

include ../../../include/kframework/ktest.mak
```

ehildenb:
@denis-bogdanas we need more information on how to reproduce. Is this a `kompile` time error? And what is the error message? What do you expect to happen.

I see you opened the PR with the test, but please just post that info here.

denis-bogdanas:
We have a spec that imports `VERIFICATION-EXTENDED`, which imports `VERIFICATION-DEMO`. And kprove argument is `--def-module VERIFICATION-DEMO` instead of `--def-module VERIFICATION-EXTENDED`. The 2 verification modules have different configuration.

I expect an error telling that `--def-module` argument is incorrect, because the definition top cell is different from the spec rule top cell.

Current output:
```
java.lang.AssertionError: Unexpected non-rule claim syntax KItem ::= Cell
     [exec] 	at org.kframework.backend.kore.ModuleToKORE.convertSpecificationModule(ModuleToKORE.java:845)
```

This is low priority.

dwightguth:
the entire way definitions are constructed in the kprove frontend is nasty, and it's the reason for this type of problem. We need to have a serious discussion about how to change the way the kprove and kompile interact, which will probably introduce breaking changes to how specs are written, but I don't see a better option.

denis-bogdanas:
[...]

PR #1778, opened 2021-02-02
Changes to `kompile` and `kprove` to allow standalone installations

### LINKED ISSUE #1754: Investigate improving ship-abilitly of K definitions
Motivation
----------

Packaging K definitions is somewhat difficult right now, which is exasperated by the fact that many of K's functionalities require (i) an install of K, an (ii) a `*-kompiled` directory that we can point at. This makes it hard to install standalone semantics as tools, except for those specific tools that are compiled to binaries directly with the LLVM backend.

Proposals
-------------

This is brainstorming ideas for improving this situation, not necessarily concrete proposals about what to do.

- Serialize entire `*-kompiled` directory into a single file which can be unpacked into the correct place.
- Support generating runner scripts which wrap `krun ....` and `kprove ....` with the correct arguments like `--directory ...` already applied to them for each semantics.
- Produce standalone binaries for the kompiled definition with Haskell backend (build the definition.kore into the ELF file?).
- Provide a way to package K directly into other packages (so KEVM would come with its own local K install in the users installation directory). This is space-intensive, but avoids the problem where a user has to first download the K deb file and install it, then can download and install the KEVM deb file.

ehildenb:
Fixes #1754 

The goal is to be able to have a standalone semantics installation which bundles the correct version of K with it. This is already possible for concrete backends, but it's important that the frontend that is bundled not attempt to write to the `*-kompiled` directories (because it will be in an unwriteable location).

Changes needed are:

-   Parse cache is not writeable (`cache.bin`)

### DIFF
--- a/kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
+++ b/kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
@@ -56,7 +56,7 @@ public class ProofDefinitionBuilder {
      * @param specModuleName Module containing specifications to prove
      */
     public Tuple2<Definition, Module> build(File specFile, String defModuleName, String specModuleName) {
-        return build(specFile, defModuleName, specModuleName, true);
+        return build(specFile, defModuleName, specModuleName, false);
     }
 
     public Tuple2<Definition, Module> build(File specFile, String defModuleName, String specModuleName, boolean readOnlyCache) {
--- a/kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
+++ b/kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
@@ -56,6 +56,10 @@ public class ProofDefinitionBuilder {
      * @param specModuleName Module containing specifications to prove
      */
     public Tuple2<Definition, Module> build(File specFile, String defModuleName, String specModuleName) {
+        return build(specFile, defModuleName, specModuleName, true);
+    }
+
+    public Tuple2<Definition, Module> build(File specFile, String defModuleName, String specModuleName, boolean updateCaches) {
         String defModuleNameUpdated =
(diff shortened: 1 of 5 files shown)


radumereuta (commented):
I'm neutral on this one.
Would this still be a problem if we fix #1333?
Can you also add some references in the code to this issue?
There is a lot of code for just one specific scenario. Having some references to the motivation in the code could be helpful in the future.

ehildenb:
@radumereuta most of the code change is just a refactoring of the NumberSentences step in kompilation to be IO pure, which I think is arguably a good thing anyway.

The only things added specifically for this are the checks about whether we should update the cache or not (just if statements around those steps).

So I guess I'm surprised to see you say "a lot of code", because there is really very little new code here.

And this wouldn't be a problem if we fix #1333, but that doesn't really matter to me, because that has been de-prioritized for whatever reasons.

radumereuta:
You are right, there isn't a lot of new code because you moved it around, but the diff is not small either.
I'm reiterating my first statement. I'm neutral on this one. Should be properly fixed when we overhaul kprove.
Ask for my review again when tests pass.

ehildenb:
I've pulled out the NumberSentence refactor because it doesn't seem to cause problems, and willl amke this diff smaller: https://github.com/kframework/k/pull/1787

radumereuta (approved):

kernel/src/main/java/org/kframework/kprove/ProofDefinitionBuilder.java
```diff
@@ -56,14 +56,18 @@ public ProofDefinitionBuilder(CompiledDefinition compiledDefinition, Backend bac
      * @param specModuleName Module containing specifications to prove
      */
     public Tuple2<Definition, Module> build(File specFile, String defModuleName, String specModuleName) {
+        return build(specFile, defModuleName, specModuleName, true);
+    }
```
ehildenb:
```suggestion
        return build(specFile, defModuleName, specModuleName, false);
    }
```

PR #1784, opened 2021-02-02
fix bison generation bug when nonterminal has no productions

dwightguth:
Basically, we were accidentally inserting a production from epsilon to every sort that didn't have any productions. This comes about because I misunderstood the syntax of Bison. The PR should fix this and thus no longer emit these spurious bison rules. I have tested that this fixes the memory exhuastion bug in Algorand and substantially reduces the number of conflicts in definitions with nonterminals that have no productions.


ehildenb (approved):
LGTM. If there is a reasonable test you can add, please add it before adding automerge, otherwise feel free to automerge.

PR #1792, opened 2021-02-04
Fix auto delete files when krun --dry-run

### LINKED ISSUE #1775: [Bug] krun - dry-run option does not save intermediate files
K Version: 95b467bdfceb61e51c67b9d4a3f11d7698c9ac86

`krun --dry-run` prints the command that would be used to run the backend, but it deletes the temporary files that the backend would use!

radumereuta:
Fixes: #1775
Weird thing, there are two input files after execution, even though the extra file is in `tempFiles`, so it should have been deleted.

I also moved the sort information for each config variable into a single file to avoid leftovers when recompiling a definition w/o removing *-kompiled directory. Less clutter and fewer issues in krun.


ehildenb:
@radumereuta is surprised that `input_file` continues to exist after the execution, he only expects `expanded_input_file` to exist.

radumereuta:
Apparently, slash in bash removes prefixes, not entire elements and this resulted in `tempFiles` getting corrupted.

dwightguth (approved):

PR #1803, opened 2021-02-09
Handle parametric productions correctly

### LINKED ISSUE #1780: [Bug] ModuleToKore should handle parametric productions correctly
This is an intermediate step in trying to fix #1415.
After fixing:
```
lazy val productionsForSort: Map[SortHead, Set[Production]] =
    productions
      .filter(prd => !prd.isSortVariable(prd.sort)) // line added
      .groupBy(_.sort.head)
      .map { case (l, ps) => (l, ps) }
```
things started to crash in ModuleToKore.java, specifically 'no junk' and  'no confusion different constructors' axiom generation don't take into consideration parameters and it results in `key not found exception`.

I will post more questions here when I find more issues.

radumereuta:
Fixes: #1780
The problem was that ModuleToKore didn't take into consideration parameters when generating axioms for constructors:
- constructors that return a parameter are not supported and should not be included in any of the axioms
- modified `convertParams` to take a list of parameters instead of a single one
- *no confusion different constructors* didn't handle parameters at all
- added `production.renameParams(prefix:String, suffix:String):Production` to aid with renaming parameters inside a production. If we generate axioms that contain two or more parametric productions we risk having accidental name collisions.
- *no junk axiom* had the same issue with not handling parameters at all

This is not necessarily blocking the implementation of parametric functions but by the time I realized it the work was almost done and it's worth merging in.


k-distribution/tests/regression-new/issue-1803-paramProds/Makefile
```diff
@@ -0,0 +1,3 @@
+KOMPILE_BACKEND?=haskell
+
+include ../../../include/kframework/ktest-fail.mak
```
dwightguth: I don't understand this test. You're importing ktest-fail, so it's supposed to not kompile, but the output file doesn't show any error. That seems like a bug to me. We should never output nothing if compilation fails.

dwightguth (commented):
I'd like to get Tom's opinion, but this code feels messy to me, particularly the way renaming of sort variables is happening. If the functionality that we're modifying doesn't work yet anyway, I'd rather we just leave this alone for now and come back later, because we have no idea what we might decide later about parametric constructors.

k-distribution/tests/regression-new/issue-1803-paramProds/Makefile
radumereuta: I copied the files from another test. I wanted to check that it kompiles correctly, with no errors or warnings.

ttuegel:
I thought we decided to postpone supporting parametric constructors at all. To support parametric constructors, we will have to decide how (or whether) to handle constructors that are less parametric than their sort (generalized algebraic data types). As far as I can tell, this pull request doesn't attempt to solve that problem, so from my perspective it's still incomplete.

PR #1804, opened 2021-02-09
Krun config sorts in single file

radumereuta:
This is to clean up a bit the *-kompiled directory.

I'm using Visual Studio Code for running and debugging bash, so I added it to gitignore also.


dwightguth (changes requested):
declare -A will not work on Mac OS.

dwightguth (approved):

PR #1809, opened 2021-02-11
Re-enable Arch Build

### LINKED ISSUE #1805: Re-enable Arch Linux
The Arch dockerfile is failing to build, because of some change in the Arch upstream signing keys. Somehow this needs to be fixed, because it's blocking K releases.

ehildenb:
Fixes: #1805


sskeirik:
According to [this StackOverflow answer from 02/11/2021](https://serverfault.com/questions/1052963/pacman-doesnt-work-in-docker-image/1053273#1053273), there is a problem where the latest versions of Arch Linux depend on a newer glibc which now requires new Linux kernel features that older host distros (like Ubuntu) don't support.

What distro is running our CI servers?

If you look at the other answer on the page, there appears to a kludgy solution that gets around this where the host is able to somehow ignore this problem --- I'm not sure if it is worth it to do that though.

ehildenb:
Evidence it works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/PR-1809/11/

dwightguth (approved):

PR #1812, opened 2021-02-13
Keep #Or patterns which appear only on the RHS of a rewrite

### LINKED ISSUE #1789: [Bug] [kompile] - Broken #Or translation to kore in function equation
K Version
---------

```sh
$ kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: c305b58
Git branch: master
Build date: Sat Jan 30 02:51:41 EET 2021
```

Description
-----------

Rule whose right-hand-side consist of an `#Or` expression replaces the entire RHS with a variable during translation.

Input Files
-----------

tmp.k
```k
module TMP
  imports INT

  syntax Int ::= split(Int)  [function, functional]

  rule  split(V:Int)
        =>
          (
            (
              {true #Equals (V >=Int 0)}
              #And V
            )
            #Or
            (
              {false #Equals (V >=Int 0)}
              #And (V +Int 1)
            )
          )
    [label(xyzzy)]

endmodule
```

Reproduction Steps
------------------

```sh
$ kompile --backend haskell tmp.k
```

The `xyzzy` rule above is translated to the following, which replaces the entire right-hand side with a variable.
```kore
  axiom{R} \implies{R} (
    \and{R}(
      \top{R}(),
      \and{R} (
          \in{SortInt{}, R} (
            X0:SortInt{},
            VarV:SortInt{}
          ),
          \top{R} ()
        )),
    \and{R} (
      \equals{SortInt{},R} (
        Lblsplit'LParUndsRParUnds'TMP'Unds'Int'Unds'Int{}(X0:SortInt{}),
        Var'Unds'0:SortInt{}),
      \top{R}()))
```

Expected Behavior
-----------------

The RHS should be translated to Kore instead of being replaced with a variable.

radumereuta:
Fixes: #1789
@dwightguth please pay extra attention to this PR. I followed the instructions in the issue, but I don't fully understand all the implications behind the reasoning.


k-distribution/tests/regression-new/issue-1789-rhsOr/1.test
```diff
@@ -0,0 +1 @@
+1
```
ehildenb: Can we add a test for `0` as well, to make sure the `split` is working as intended @radumereuta ?

dwightguth:
This seems needlessly complicated to me. Shouldn't it be sufficient to just not call apply in the KRewrite overload of applp inside transform?

ehildenb:
Let's make sure we have a test which exercises the `#Or` above a `=>` case, something like this:

```
rule <k> a => b ... </k>
     <foo> bar #Or baz </foo>
```

We want this to become `(bar #Or baz) #as X`, to make it executable for the LLVM backend.

But we don't want the rewrite itself to split (`#Or` pushed through the `=>`).

dwightguth (approved):

k-distribution/tests/regression-new/issue-1789-rhsOr/haskell/0.test.out
```diff
@@ -0,0 +1,3 @@
+<k>
+  0 ~> .
+</k>
```
ehildenb: Sorry, we should instead use `-2` instead of `0`, so that we can test that both sides of the `#Or` are tested in the `split`

PR #1813, opened 2021-02-13
Use a global Stack cache

ttuegel:
This pull request would use a global Stack cache for macOS builds. I'm not quite sure how to test it before merging; I think it would have to be merged to rebuild the Anka template. And, this is all a moot point if Homebrew clears the environment.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -339,9 +339,7 @@ pipeline {
               stages {
                 stage('Build Homebrew Bottle') {
                   agent { label 'anka' }
-                  environment {
-                    STACK_ROOT = '/opt/stack'
-                  }
+                  environment { STACK_ROOT = '/opt/stack' }
                   steps {
                     unstash 'src'
                     dir('kframework') { checkout scm }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -339,6 +339,9 @@ pipeline {
               stages {
                 stage('Build Homebrew Bottle') {
                   agent { label 'anka' }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -339,6 +339,9 @@ pipeline {
               stages {
                 stage('Build Homebrew Bottle') {
                   agent { label 'anka' }
+                  environment {
+                    STACK_ROOT = '/opt/stack'
+                  }
```
```suggestion
                  environment { STACK_ROOT = '/opt/stack' }
```

Jenkinsfile.anka
```diff
@@ -13,6 +13,9 @@ pipeline {
     stage('Build Anka VM') {
       options { timeout(time: 150, unit: 'MINUTES') }
       agent { label 'anka-k-master' }
+      environment {
+        STACK_ROOT = '/opt/stack'
+      }
```
ehildenb:
```suggestion
      environment { STACK_ROOT = '/opt/stack' }
```

ehildenb (approved):
Feel free to `automerge` if you want to try it, we can roll it back

dwightguth:
Looks like Homebrew filters the environment unless you set the HOMEBREW_NO_ENV_FILTERING variable.

ttuegel:
Do you know if it's possible to disable filtering for certain variables only?

dwightguth:
It seems that Homebrew has a whitelist of variables, but I don't see any way to modify that whitelist as a user.

PR #1826, opened 2021-02-22
Allowing calling ModuleToKORE in file-IO pure way

ehildenb:
-   Removes all calls to file-IO out of ModuleToKORE by hoisting `convert` function out of `ModuleToKORE` into `KoreBackend`.
-   Provides a function with the same signature as the previous `convert` function which does the reads and writes.
-   Makes the file writes in KoreBackend only happen if the tool calling it is `KOMPILE`, so that other tools are not updating `syntaxDefinition.kore` or `macros.kore` erroneously.

This is needed for two things:

-   Having standalone installations of semantics. It's important that `kprove` does not try to write back to the `*-kompiled` directory when executing, so this is needed.
-   Fixing the KEVM K update to use the latest kx script, because the KEVM test-suite first builds (`kompile`), then proves, (`kprove`), then searches (`krun --search`). The issue here is hard to explain, but I'll try:
    -   Calling `kompile` leads to the generation of `syntaxDefinition.kore`.
    -   Subsequent calls to `krun` use `kore-expand-macros`, which uses `syntaxDefinition.kore` in the new `kx` codepath (used to use the serialized definition in `compiled.bin` directly instead).
    -   Calling `kprove` then _overwrites_ `syntaxDefinition.kore`, with contents relevant for the proof being done, because it calls `ModuleToKORE` via `KoreBackend`.
    -   Then a subsequent call to `krun` causes a crash, because the `syntaxDefinition.kore` is not the correct one from the original `kompile` call.

### DIFF
--- a/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
+++ b/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
@@ -7,6 +7,7 @@ import org.kframework.backend.kore.KoreBackend;
 import org.kframework.compile.Backend;
 import org.kframework.kompile.CompiledDefinition;
 import org.kframework.kompile.KompileOptions;
+import org.kframework.main.Tool;
 import org.kframework.utils.errorsystem.KExceptionManager;
 import org.kframework.utils.errorsystem.KEMException;
 import org.kframework.utils.file.FileUtil;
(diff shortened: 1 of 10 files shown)


ehildenb:
The main point of this branch is probably obsolete given https://github.com/kframework/k/pull/1829, but maybe this refactoring is still useful.

radumereuta:
This looks fine, but it's touching some files that are used by the llvm tools so I would prefer if Dwight has a look also.

radumereuta (approved):
lgtm, but as I said previously, I think Dwight should also have a look.

dwightguth (approved):

PR #1828, opened 2021-02-22
brackets obey priority rules

dwightguth:
Currently there is no way to disambiguate using priorities if the ambiguity results from a parse where one of the parses has a bracket and the other parse has a non-bracket production containing parentheses. What we do to fix this is to make the priority relation be applicable to bracket productions. We also modify the way priority is applied so that it is applied to bracket productions even if the nonterminal in the production is a middle nonterminal.

This should not affect the semantics of any definitions that did not express a priority relation with respect to bracket productions, and in the event that such definitions exist, it should probably be seen as having been parsed erroneously previously. However, I will monitor for failed tests and see whether existing definitions are going to be affected by this.


radumereuta:
The purpose of `bracket` is to override priorities. What are you trying to do?

ehildenb:
See discussion here: https://github.com/runtimeverification/algorand-sc-semantics/issues/134#issuecomment-783674470

radumereuta:
I just added a comment in the main issue on why I think this is not the correct solution: https://github.com/runtimeverification/algorand-sc-semantics/issues/134#issuecomment-785376444

ehildenb:
I believe that @radumereuta 's concerns here https://github.com/runtimeverification/algorand-sc-semantics/issues/134#issuecomment-785376444 need to be addressed.

When we discussed fixing this issue, I think the rational we came up with for making `bracket` special for the priority filter is that `bracket` productions always only have exactly a single non-terminal, so of course we want to consider that non-terminal position for the priority filter, rather than edge non-terminals.

But Radu suggests (as Dwight has in the past) that we should instead be renaming the production for `"(" K ")"` in rules. Dwight suggested just applying a patch to `kast.md`, and Radu suggests using module signatures.

My thinking is that we should be able to handle s-expression like languages, _without_ that particular semantics having to rename the K parens. So I'm not a huge fan of the module signatures or patch to `kast.md` approach.

radumereuta:
But this is a fundamental problem in parsing. If you have two productions that are this similar, you will have ambiguities.
If you have a way to avoid them from ever existing, then it's the best approach. This is why you have LL and LR as a subspace of CFG. I view the renaming/hiding of the default bracket as the best approach. It results in fewer ambiguities that the user has to handle. Less error-prone. Parsing time is reduced.

radumereuta:
Another proposed approach (a long time ago) was to have custom disambiguation rules written in K and which would apply at disambiguation time. That didn't come to life because it would have been 1. expensive at runtime, 2. pretty complex to implement.

dwightguth:
I agree that telling the priority filter to visit the nonterminal in the middle of the production when the production is a bracket production is a bit of a hack. Everett considered it to be okay, so I went ahead with it since it was his suggestion, but my original suggestion was to follow the same path that SDF did and provide a mechanism by which the user can specify which nonterminals in the outer production they want the priority filter to apply to, and then just apply it to the outermost nonterminals by default. I can go that route, at which time this code becomes more or less exactly the same as before with the exception that there is an added generalization which was already present in the parser generation process K was originally based on.

With that being said, you are fundamentally misunderstanding what a bracket does. A bracket is a production that does not generate any symbol in the AST. There is no reason why a bracket should not obey the priority relation when the user explicitly asks for it to do so. Indeed, it is highly counterintuitive that I might explicitly write:
```
syntax Exp ::= "(" Exp [bracket , foo]
syntax Exp ::= Exp "+" Exp [plus] | Int
syntax priority foo > plus
rule (I1 + I2 => I1 +Int I2
```

and somehow NOT get a parse error. That, to me, is the hack, that we introduced because we previously had no way of tracking brackets as part of the priority relation simply because they had no KLabel.

I agree that you can solve the parsing ambiguities in Algorand using some kind of renaming. However, you can also solve them using this PR. Given that one is a change that will require potentially weeks of work and the other took me... today and yesterday? I know which one I want to implement.

My proposed solution is very simple: I will simply implement the solution to this problem that I originally proposed to Everett in the Algorand meeting the other day. This will have the following effects:

1. No existing definitions change their semantics
2. We fix a bug that prevented the user from expressing priority relations with respect to bracket productions themselves
3. We introduce a feature relating to priority relations that was present in SDF (which our implementation of priorities is based on) but never made its way into K

I see no cons to any of these items.

radumereuta:
The priorities and bracket interaction in K are supposed to be a mirror of Rascal (https://www.rascal-mpl.org/) which is the successor of SDF. They also realized the limitations and changed a few things: priorities apply only on the outermost NTs, and subsorts are not included in priorities (and others - see my linked paper in my previous comment).

That being said. I have no counter argument against *we have a deadline soon and we need a fix now*. The best approach would have been to avoid the problem in the first place. Production renaming at module inclusion is my preferred solution. We needed this functionality for a very long time. We had this issue with S expression languages before. We wanted to rename the rewrite arrow, we got requests to offer alternatives to the XML-like notation for cells. Plus, less ambiguities at parsing time means faster kompile.

The latest update is a bit better though. There are still a few kinks to iron out. For example:
```
  syntax Exp ::= Int
             | "[" Exp "]" [bracket, myBracket, applyPriority(1)]
             > Exp "+" Exp [klabel(plus)]
  syntax priorities myBracket > plus // this is needed, inline doesn't work
  rule [1 + 2] => .K
```
inline priorities for bracket doesn't work. Doesn't reject my rule.
```
  syntax priorities defaultBracket > plus
  rule (1 + 2) => .K
```
The error message would be valid normally but now it contradicts itself:
```
[Error] Inner Parser: Priority filter exception. Cannot use
_+__TEST_Exp_Exp_Exp as an immediate child of (_)_KSEQ_Sort_Sort. Consider
using parentheses around _+__TEST_Exp_Exp_Exp
```

I wouldn't spend time fixing them though, since I don't consider this a go-to solution. It's going to be *just do this, and it's going to work for you*.

k-distribution/pl-tutorial/1_k/4_imp++/lesson_6/imp.k
```diff
@@ -6,14 +6,14 @@ module IMP-SYNTAX
                  | "++" Id
                  | "read" "(" ")"
                  | "-" Int
+                 | "(" AExp ")"               [bracket]
                  > AExp "/" AExp              [left, strict, division]
                  > AExp "+" AExp              [left, strict]
-                 | "(" AExp ")"               [bracket]
```
radumereuta, line 11: I don't understand these changes. Why did you rearrange them?

pending-documentation.md
```diff
@@ -367,6 +367,18 @@ nonterminal. If the rightmost nonterminal is followed by a terminal (or
 respectively the leftmost preceded), priority and associativity do not apply.
 Thus we can generalize these concepts to arbitrary context-free grammars.
 
+Note that in some cases, this is not the behavior you want. You may actually
+want to reject parses even though the leftmost and rightmost item in a
+production are terminals. You can accomplish this by means of the
+`applyPriority` attribute. When placed on a production, it tells the parser
+which nonterminals of a production the priority filter ought to reject children
+under, overriding the default behavior. For example, I might have a production
```
radumereuta, line 375: Can you make it impersonal?

kernel/src/main/java/org/kframework/parser/inner/disambiguation/PriorityVisitor.java
```diff
@@ -35,22 +38,55 @@ public PriorityVisitor(POSet<Tag> priorities, Set<Tuple2<Tag, Tag>> leftAssoc, S
...
-            }
-            if (tc.production().items().apply(tc.production().items().size() - 1) instanceof NonTerminal) {
-                int last = tc.items().size() - 1;
-                Either<java.util.Set<KEMException>, Term> rez =
-                        new PriorityVisitor2(tc, PriorityVisitor2.Side.RIGHT, priorities, leftAssoc, rightAssoc).apply(tc.get(last));
-                if (rez.isLeft())
-                    return rez;
-                tc = tc.with(last, rez.right().get());
+            if (tc.production().att().contains("applyPriority")) {
+              String[] pieces = StringUtil.splitOneDimensionalAtt(tc.production().att().get("applyPriority"));
```
radumereuta, line 44: Doing this at every step might be slow.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/PriorityVisitor.java
```diff
@@ -77,11 +113,10 @@ public PriorityVisitor2(TermCons parent, Side side, POSet<Tag> priorities, Set<T
         }
 
         public Either<java.util.Set<KEMException>, Term> apply(TermCons tc) {
-            if (tc.production().att().contains(Att.BRACKET())) return Right.apply(tc);
             //if (Side.RIGHT  == side && !(tc.production().items().apply(0) instanceof NonTerminal)) return Right.apply(tc);
             //if (Side.LEFT == side && !(tc.production().items().apply(tc.production().items().size() - 1) instanceof NonTerminal)) return Right.apply(tc);
```
radumereuta, line 117: Since we're open to changing the behavior of priorities, can we add these now? We could finally fix some long standing limitations.

k-distribution/pl-tutorial/1_k/4_imp++/lesson_6/imp.k
dwightguth: Think about it. The previous order was saying that `()` couldn't appear under a `/`, which we obviously want to allow. Once I fixed the bug that was preventing the priority relation the definition had expressed from being applied, suddenly you had parsing errors because the grammar had been incorrectly specified. So I fixed it.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/PriorityVisitor.java
dwightguth: I'm still not persuaded we won't introduce new ambiguities that are difficult to solve into existing definitions if we do this. It's a much more intrusive change than what I'm suggesting. Also, unless I'm mistaken, if we were to naively uncomment these lines, the test I wrote would fail.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/PriorityVisitor.java
dwightguth: This is not happening inside a loop. The only way we could precompute this value would be to do some kind of preprocessing pass in another visitor and then store the values in structured form as attributes and read them here. That would be a lot more complicated to code, because currently we don't have any kind of semantic analysis phase of compilation. Also, you will note we do the same thing with other attributes later on in the compilation process in places where it doesn't seem to have much effect on performance, so I'm not really worried about it. Let's optimize based on data, not based on what we assume will be slow.

radumereuta (approved):
Let's see how this goes.

PR #1831, opened 2021-02-24
docs: note substrString may throw exceptions in the LLVM backend

### LINKED ISSUE #1820: [Bug] [krun] - krun fails with an uncaught exception when substrString is passed out-of-bounds indices
K Version
---------

```
$ kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: bfcf61f
Git branch: master
Build date: Wed Feb 17 11:11:17 CST 2021
```

Description
-----------

A call to `substrString(S, start, end)` with `start` or `end` being greater than `lengthString(S)` causes `krun` to fail at runtime with an uncaught exception.

When `start >Int lengthString(S)`, the error is:
```
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: Invalid string slice
../../k/k-distribution/target/release/k/bin/krun: line 48: 91470 Abort trap: 6           "$@"
[Error] Critical: Backend crashed during rewriting with exit code 134
```

When `end >Int lengthString(S), the error is:

```
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: Invalid string slice
../../k/k-distribution/target/release/k/bin/krun: line 48: 91697 Abort trap: 6           "$@"
[Error] Critical: Backend crashed during rewriting with exit code 134
```

Input Files
-----------

`test.k`

```
module TEST-SYNTAX
  imports STRING-SYNTAX

  syntax Pgm ::= "go"
endmodule

module TEST
  imports TEST-SYNTAX
  imports STRING

  configuration
    <k> $PGM:Pgm </k>
    <s> "test string" </s>

  rule <k> go => . </k>
       <s> S => substrString(S, lengthString(S) +Int 1, lengthString(S)) </s> // invalid! -- start index out of bounds
//       <s> S => substrString(S, 0, lengthString(S) +Int 1) </s>               // invalid! -- end index out of bounds
endmodule
     
```

`input.test`

```
go
```

Reproduction Steps
------------------

`kompile test.k` succeeds, but `krun` fails with:  

```
krun input.test
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: Invalid string slice
../../k/k-distribution/target/release/k/bin/krun: line 48: 92713 Abort trap: 6           "$@"
[Error] Critical: Backend crashed during rewriting with exit code 134```
```

Expected Behavior
-----------------

- I don't think that this is normal behavior. I think `krun` should fail "gracefully" by catching this exception, producing an a more descriptive error message to the user, and if possible pointing out where in the file the invalid index is used. 
- This should also be documented so that users of `substrString` are reminded to guard against this problem when strings are dynamically loaded (bounds are not known a priori).

sskeirik:
Fixes: #1820 

To ensure the added documentation was correct, I manually compiled examples of substrString which could induce failures and checked that they threw exceptions in the LLVM backend as expected. 

@ttuegel I'm assuming the Haskell backend does not crash when `substrString` is passed invalid indices, but if does, please let me know so I can update this PR documentation to include that fact.


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1399,8 +1399,14 @@ a single code point) and an integer in O(1) time.
 ### String substring
 
 You can compute a substring of a string in O(N) time (where N is the
-length of the substring). Note that the range generated is 
-`[startIndex..endIndex)`
+length of the substring). There are two important facts to note:
+
+1.  the range generated includes the character at `startIndex` but excludes the
+    character at `endIndex`, i.e., the range is `[startIndex..endIndex)`.
+2.  when using the LLVM backend, if the substring indices are invalid (i.e.,
```
dwightguth: I imagine on the Haskell backend, in this case it returns #Bottom, which is not what you want, so you probably just want to say "This function only returns valid results if" rather than singling out a specific backend and explaining that it causes it to crash.

sskeirik: #1820 which was the inspiration for this change was about the frustration caused by not knowing that builtin functions may throw exceptions when passed invalid input. If the comment doesn't mention that part, then I don't think it will address the issue very well. I can remove the part about being specific to the LLVM backend though without any harm though (since I just "may throw"). What do you think?

There are other ways to achieve a similar effect that we can try, e.g. add a `may-throw(...)` attribute parameterized by a list of backend names that we can attach to symbols with the `function` attribute. Then we can add a new warning class about functions which throw exceptions.

dwightguth: I mean, the point is, if it doesn't have the functional attribute, it's not defined on all inputs. The documentation can describe what inputs it's defined on, and we can explain at the top of the documentation what happens on various backends if you call a function on inputs it's not defined on, but we shouldn't be going into detail on individual backends in this particular place in the documentation, because then we'd have to do it everywhere.

sskeirik: @dwightguth I updated the documentation according to your recommendation. Would you mind taking a look at this again?

dwightguth (approved):

PR #1835, opened 2021-02-27
Fix krun --dry-run warnings

radumereuta:
`$value != $tempDir && newArray+=($value)`
This would try to execute as a command line and gave an error message because it couldn't.
So wrap it inside an if.

### DIFF
--- a/k-distribution/src/main/scripts/bin/krun
+++ b/k-distribution/src/main/scripts/bin/krun
@@ -215,7 +215,9 @@ do
       # slash removes prefixes so build a new array with the files to keep
       newArray=()
       for value in "${tempFiles[@]}"; do
-          $value != $tempDir && newArray+=($value)
+          if [ $value != $tempDir ]; then 
+            newArray+=($value)
+          fi
       done
       tempFiles="${newArray[@]}"
       unset newArray


k-distribution/src/main/scripts/bin/krun
```diff
@@ -215,7 +215,9 @@ do
       # slash removes prefixes so build a new array with the files to keep
       newArray=()
       for value in "${tempFiles[@]}"; do
-          $value != $tempDir && newArray+=($value)
+          if [ $value != $tempDir ]; then 
+            newArray+=($value)
+          fi
```
ehildenb, line 220: Just something to note (for posterity), you should never use `&&` to chain together commands (for the reason you noticed). 

Instead `||` should be used. So you wolud instead say:

```
$value == $tempDir || newArray+=($value)
```

This makes sure that the line never throws an error.

But it looks good!

ehildenb (approved):

dwightguth (approved):

k-distribution/src/main/scripts/bin/krun
MirceaS: @ehildenb I thought that if you use `||`, the command on the right only gets executed when the first one fails. If that's the case, then I'm not sure `||` is appropriate for chaining commands in general, unless I'm missing something...

PR #1847, opened 2021-03-09
lessons 3/4: parsing and disambiguation

dwightguth:


k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+A non-terminal, conversely, represents a sort name, where the syntax of that
+production accepts any valid term of that sort at that position.
+
+This is why, when we wrote the program `colorOf(Banana())`, `krun` was able to
+execute that program: because it represented a term of sort `Color` that was
+parsed and interpreted by K's interpreter. Seen this way, you can see how that
+string represents a sentence in the formal language defined by the above
+grammar. K then parses it automatically into an AST of that term, and then,
+since the symbol `colorOf` is a function, it evaluates that function using
+the rules that define it.
```
ehildenb:
```suggestion
parsed and interpreted by K's interpreter. K parses this string into an AST of the term, then evaluates the `colorOf` function using the functional rules provided in the definition.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+    Lblfalse'Unds'LESSON-03-A'Unds'Boolean{}()
+  )
+)
+```
+
+Don't worry about what exactly this means yet, just understand that it
+represents the AST of the program that you just parsed. You ought to be able
+to recognize the basic shape of it by seeing the words `true`, `false`, and
+`And` in there. This is `KORE`, the intermediate representation of K, and we
+will cover it in detail later.
```
ehildenb:
```suggestion
`And` in there. This is **Kore**, the intermediate representation of K, and we
will cover it in detail later.
```

ehildenb: You may also want to consider showing `--output kast` here, because later in the lesson the parsing ambiguities are given in Kast form.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+Essentially what this error is saying is, `kast` was unable to parse this
+program because it is ambiguous. K's just-in-time parser is what is technically
+known as a GLL parser, which in practice means it handles the full generality
+of context-free grammars, including those grammars which are ambiguous. An
+ambiguous grammar is one where one or more sentences in the grammar might be
+able to be parsed into more than one distinct AST. In this example, it can't
+decide whether it should be parsed as `(true && false) || false` or as
+`true && (false || false)`. As a result, it reports the error to the user.
+
```
ehildenb:
```suggestion
This error is saying that `kast` was unable to parse this
program because it is ambiguous. K's just-in-time parser is a GLL parser, which means it can handle the full generality
of context-free grammars, including grammars which are ambiguous. An
ambiguous grammar is one where the same string (program) can be validly parsed as multiple distinct ASTs. In this example, it can't
decide whether it should be parsed as `(true && false) || false` or as
`true && (false || false)`. As a result, it reports the error to the user.

```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+true && (false || false)
+```
+
+If you parse these programs with `kast`, you will once again get a single
+unique AST with no error. If you look, you may spot that the bracket itself
+does not appear in the AST. In fact, this is a property unique to brackets:
+productions with the bracket attribute are not represented in the parsed AST
+of a term, and the child of the bracket is folded immediately into the parent
+term. This is the reason for the requirement that a bracket production have
+a single non-terminal of the same sort as the production itself.
```
ehildenb:
```suggestion
If you parse these programs with `kast`, you will once again get a single
unique AST with no error. If you look, you will notice that the bracket itself
does not appear in the AST. In fact, this is a property unique to brackets:
productions with the bracket attribute are not represented in the parsed AST
of a term, and the child of the bracket is folded immediately into the parent
term. This is the reason for the requirement that a bracket production have
a single non-terminal of the same sort as the production itself.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+## Priority blocks
+
+Of course, in practice, very few formal languages outside the domain of
+natural language processing are ambiguous. The main reason for this is that
+parsing unambiguous languages is asymptotically faster than parsing ambiguous
+languages. Programming language designers instead prefer to use a notion of
+operator precedence and associativity to make expression grammars unambiguous
+by automatically choosing certain parses over other parses in cases where the
+grammar would otherwise be ambiguous.
```
ehildenb:
```suggestion
Of course, in practice, very few formal languages outside the domain of
natural language processing are ambiguous. The main reason for this is that
parsing unambiguous languages is asymptotically faster than parsing ambiguous
languages. Programming language designers use the notions of
operator precedence and associativity to make expression grammars unambiguous.
These tools work by instructing the parser which ASTs to select in the case of ambiguities; often times it's possible to remove _all_ the ambiguities in a grammar with these techniques.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+While it is sometimes possible to explicitly rewrite the grammar to remove
+these parses, because K's grammar specification and AST generation are
+inextricably linked, this is generally discouraged. Instead, we use the
+approach of explicitly expressing the relative precedence of different
+operators in different situations in order to resolve the ambiguity.
+
```
ehildenb: This block has the tone of "contradicting" the previous block of text, even though it agrees with it. Doesn't seem to add much either, so I would just remove it.

```suggestion
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+For example, in C, `&&` binds tighter in precedence than `||`, meaning that
+the expression `true && false || false` would be parsed as
+`(true && false) || false`.
```
ehildenb:
```suggestion
For example, in C, `&&` binds tighter in precedence than `||`, meaning that
the expression `true && false || false` has only one valid parse:
`(true && false) || false`.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+next `>` operator or the end of the production block. Thus, we can see that the
+second, third, fourth, and fifth priority groups in this grammar all consist
+of a single production.
+
+The meaning of these priority groups becomes apparent when parsing programs:
+A symbol with a **lesser priority**, (i.e., one that **binds looser**), cannot
+appear as the **direct child** of a symbol with a **greater priority** (i.e.,
+one that **binds tighter**. In this case, the `>` operator can be seen as a
+**less-than** operator describing a transitive partial ordering on the 
+productions in the production block, expressing their relative priority.
```
ehildenb:
```suggestion
one that **binds tighter**. In this case, the `>` operator can be seen as a
**greater-than** operator describing a transitive partial ordering on the 
productions in the production block, expressing their relative priority.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+`||`, we **reject** the parse where `||` is under the `&&` operator. As a
+result, we are left with the unambiguous parse `(true && false) || false`.
+Similarly, `true || false && false` parses unambiguously as
+`true || (false && false)`. Conversely, if the user explicitly wants the other
+parse, they can express this parse using brackets. In other words, they can
+explicitly write the program `true && (false || false)`. This still parses
+successfully because the `||` operator is no longer the **direct** child of the
+`&&` operator, but is instead the direct child of the `()` operator, and the
+`&&` operator is an **indirect** parent. Indirect parents are completely
+ignored when disambiguating; only the direct parent/child relation matters.
```
ehildenb:
```suggestion
To see this more concretely, let's look again at the program
`true && false || false`. As noted before, previously this program was
ambiguous because the parser could either choose the `&&` to be a child of the `||`
or vice versa. However, because a symbol with lesser priority (in this case `||`)
cannot appear as the direct child of a symbol with greater priority (in this case `&&`),
the parser can now **reject** the parse where `||` is under the `&&` operator. As a
result, we are left with the unambiguous parse `(true && false) || false`.
Similarly, `true || false && false` parses unambiguously as
`true || (false && false)`. Conversely, if the user explicitly wants the other
parse, they can express this using brackets by explicitly writing `true && (false || false)`. This still parses
successfully because the `||` operator is no longer the **direct** child of the
`&&` operator, but is instead the direct child of the `()` operator, and the
`&&` operator is an **indirect** parent.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+In fact, however, we do not reject this program as a parse error. Why is that?
+Well, the rule for priority is slightly more complex than previously described.
+In actual fact, it applies only conditionally. Specifically, it applies in
+cases where the child is either the first or last production item in the
+parent's production. For example, in the production `Bool "&&" Bool`, the
+first `Bool` non-terminal is not preceded by any terminals, and the last `Bool`
+non-terminal is not followed by any terminals. As a result of this, we apply
+the priority rule to both children of `&&`. However, in the `()` operator, 
+the sole non-terminal is both preceded by and followed by a terminal. As a
```
ehildenb:
```suggestion
the sole non-terminal is both preceded by and followed by terminals (`"("` and `")"`, respectively). As a
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+grammar is still ambiguous. We can see this if we try to parse the following
+program (`assoc.bool`):
+
+```
+true && false && false
+```
+
+Priority blocks will not help us here: the problem comes between two parses
+where both possible parses have a direct parent and child which is within a
+single priority block.
```
ehildenb:
```suggestion
Priority blocks will not help us here: the problem comes between two parses
where both possible parses have a direct parent and child which is within a
single priority block (in this case, the `&&` is in the same priority block as itself).
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+This is where the notion of associativity comes into play. Essentially,
+associativity applies the following additional rule to parses: a
```
ehildenb, line 330:
```suggestion
This is where the notion of associativity comes into play.
Associativity applies the following additional rules to parses: a
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+left-associative symbol cannot appear as a direct rightmost child of a symbol
+with equal priority; a right-associative symbol cannot appear as a direct
+leftmost child of a symbol with equal priority; and a non-associative symbol
+cannot appear as a direct leftmost **or** rightmost child of a symbol with
+equal priority.
```
ehildenb:
```suggestion
This is where the notion of associativity comes into play. Associativity applies the following additional rule to parses:

-   a left-associative symbol cannot appear as a direct rightmost child of a symbol
with equal priority,
-   a right-associative symbol cannot appear as a direct leftmost child of a symbol with equal priority, and
-   a non-associative symbol cannot appear as a direct leftmost **or** rightmost child of a symbol with equal priority.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+endmodule
+```
+
+In this example, unlike the one above, `&&`, `^`, and `||` have the same
+priority. However, viewed as a group, the entire group is left associative.
+This means that none of `&&`, `^`, and `||` can appear as the right child of
+any of `&&`, `^`, or `||`. As a result of this, this grammar is also not
+ambiguous. However, it expresses a different grammar, and you are encouraged
+to think about what the differences are in practice.
```
ehildenb: Please provide some example programs here that users can call `kast` on and see the difference, rather than just saying "think about it".

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+which you wish to express a priority or associativity relation over are
+co-located in the same block of productions. However, in practice this is not
+always feasible or desirable, especially as a definition grows in size across
+multiple modules.
+
+As a result of this, K provides a second way of declaring priority and
+associativity relations.
+
+Consider the following grammar, which we will name `lesson-03-h.k` and which
+will express the exact same grammar relation as `lesson-03-f.k`
```
ehildenb:
```suggestion
Consider the following grammar, which we will name `lesson-03-h.k` and which
will express the exact same grammar as `lesson-03-f.k`
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+Consider the following definition (`lesson-03-i.k`):
+
+```k
+module LESSON-03-I
+
+  syntax Exp ::= "true" | "false"
+  syntax Stmt ::= "if" "(" Exp ")" Stmt
+                | "if" "(" Exp ")" Stmt "else" Stmt
+                | ";"
```
ehildenb:
```suggestion
                | "{" "}"
```

The empty block `{ }` may read better in the examples below.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+`if(_)__LESSON-03-I_Stmt_Exp_Stmt`(`true_LESSON-03-I_Exp`(.KList),`if(_)_else__LESSON-03-I_Stmt_Exp_Stmt_Stmt`(`false_LESSON-03-I_Exp`(.KList),`;_LESSON-03-I_Stmt`(.KList),`;_LESSON-03-I_Stmt`(.KList)))
+2: syntax Stmt ::= "if" "(" Exp ")" Stmt "else" Stmt
+
+`if(_)_else__LESSON-03-I_Stmt_Exp_Stmt_Stmt`(`true_LESSON-03-I_Exp`(.KList),`if(_)__LESSON-03-I_Stmt_Exp_Stmt`(`false_LESSON-03-I_Exp`(.KList),`;_LESSON-03-I_Stmt`(.KList)),`;_LESSON-03-I_Stmt`(.KList))
+        Source(./dangling-else.if)
+        Location(1,1,1,30)
+```
+
+Roughly, we see that the ambiguity is between an `if` with an `else` or an `if`
+without an `else`. Siknce we want to pick the first parse, we can tell K to
```
ehildenb:
```suggestion
without an `else`. Since we want to pick the first parse, we can tell K to
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+                | ";"
+endmodule
+```
+
+Here we have added the `avoid` attribute to the `else` production. As a result,
+when an ambiguity occurs and one or more of the possible parses has that symbol
+at the top of the ambiguous part of the parse, we remove those parses from
+consideration and consider only those remaining. The `prefer` attribute behaves
+similarly, but instead removes all parses which do not have that attribute.
+In both cases, no action is taken if the parse is not ambiguous.
```
ehildenb:
```suggestion
Here we have added the `avoid` attribute to the `else` production. Now,
when an ambiguity occurs, any of the possible parses which has the `avoid` symbol
at the top will be removed from consideration. The `prefer` attribute behaves
similarly, but instead removes all parses which do not have a `prefer` symbol at the top.
In both cases, no action is taken if the parse is not ambiguous.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+a production consists of a regular expression followed by the `token`
+attribute, and the resulting AST consists of a typed string containing the
+value recognized by the regular expression.
+
+For example, we might define the integer token in C with the following
+production:
+
+```k
+syntax IntConstant ::= r"(([1-9][0-9]*)|(0[0-7]*)|(0[xX][0-9a-fA-F]+))(([uU][lL]?)|([uU]((ll)|(LL)))|([lL][uU]?)|(((ll)|(LL))[uU]?))?" [token]
+```
```
ehildenb: This is much too overwhelming of an example to start with. Let's begin with just something like `[1-9][0-9]*` or the like, and then show them this one.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+identifiers for use in your semantics later. For example:
+
+```k
+syntax Id ::= "main" [token]
+```
+
+Here we declare that `main` is a token of sort `Id`. Instead of being parsed
+as a symbol, it gets parsed as a token instead, generating a typed string
+in the AST. This is useful because the parser generally does not treat the
+`main` function in C specially; only the semantics treats it specially.
```
ehildenb:
```suggestion
Here we declare that `main` is a token of sort `Id`. Instead of being parsed
as a symbol, it gets parsed as a token, generating a typed string
in the AST. This is useful because the parser generally does not treat the
`main` function in C specially; only the semantics treats it specially.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,627 @@
...
+
+## Ahead-of-time parser generation
+
+Finally, so far we have been entirely focused on K's support for just-in-time
+parsing, where the parser is generated on the fly prior to being used.
+This benefits from being faster to generate the parser, but it suffers in
+performance if you have to repeatedly parse strings with the same parser.
+For this reason, it is generally encouraged that when parsing programs,
+you use K's ahead-of-time parser generation instead. K makes use of
+[GNU Bison](https://www.gnu.org/software/bison/) to generate parsers.
```
ehildenb:
```suggestion
So far we have been entirely focused on K's support for just-in-time
parsing, where the parser is generated on the fly prior to being used.
This benefits from being faster to generate the parser, but it suffers in
performance if you have to repeatedly parse strings with the same parser.
For this reason, it is generally encouraged that when parsing programs,
you use K's ahead-of-time parser generation. K makes use of
[GNU Bison](https://www.gnu.org/software/bison/) to generate parsers.
```

ehildenb:
I've gone through and provided copy tweaks here and there, but this really needs a larger overhaul, including:

-  Split into at least 2 lessons (it's way too long as is, easy to lose focus). One suggestion would be a single lesson which explains BNF, bracket, and token, and a second lesson which explains priority, assoc, and prefer/avoid.
-  There are many sections which are _only_ explanatory, without offering any real things for the user to try. These section serve better as a reference manual than a tutorial. Every section needs to have something explicit for the user to try at the end.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
dwightguth: This paragraph actually does serve an important role. It explains to anyone who is previously familiar with other parser generators an important way in which parser generation in K is usually done differently than they might be used to.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
dwightguth: It is actually a clearer explanation the way I originally left it. Your wording doesn't adequately explain that what matters is where the ambiguous part of the parse begins.

dwightguth:
I'm fine splitting the lesson in half, but I don't understand what you mean in your second bullet point. The general structure of the tutorial is that the lesson explains concepts, and then we provide exercises to synthesize and practice that knowledge at the end. Are you saying you would rather have exercises after each section?

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,362 @@
...
+Input the following program into your editor as `and.bool` in the same
+directory:
+
+```
+true && false
+```
+
+We cannot interpret this program yet, because we have not given rules defining
+the meaning of the `&&` function yet, but we can parse it. To do this, you can
+run (from the same  directory):
```
ehildenb:
```suggestion
run (from the same directory):
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,362 @@
...
+    Lblfalse'Unds'LESSON-03-A'Unds'Boolean{}()
+  )
+)
+```
+
+Don't worry about what exactly this means yet, just understand that it
+represents the AST of the program that you just parsed. You ought to be able
+to recognize the basic shape of it by seeing the words `true`, `false`, and
+`And` in there. This is **Kore**, the intermediate representation of K, and we
+will cover it in detail later.
```
ehildenb, line 131: I think you should have them do `--output kast` as well, and have them compare the Kore and the KAST, make sure they can map the `_&&__LESSON-03-A_Boolean_Boolean_Boolean` symbol to the munged version. You could just say "Now try `--output kast` and see how they compare" here as something they can try.

ehildenb: Just giving them something to do like this will force them to actually read the output.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,362 @@
...
+## Ambiguities
+
+Now let's try a slightly more advanced example. Input the following program
+into your editor as `and-or.bool`:
+
+```
+true && false || false
+```
+
+When you try and parse this program, you ought to see the following error:
```
ehildenb, line 165:
```suggestion
Try parsing this program (`kast and-or.bool`), you ought to see the following error:
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,362 @@
...
+true && (false || false)
+```
+
+If you parse these programs with `kast`, you will once again get a single
+unique AST with no error. If you look, you might notice that the bracket itself
+does not appear in the AST. In fact, this is a property unique to brackets:
+productions with the bracket attribute are not represented in the parsed AST
+of a term, and the child of the bracket is folded immediately into the parent
+term. This is the reason for the requirement that a bracket production have
+a single non-terminal of the same sort as the production itself.
```
ehildenb, line 243:
```suggestion
of a term, and the child of the bracket is folded immediately into the parent
term. This is the reason for the requirement that a bracket production have
a single non-terminal of the same sort as the production itself.

Confirm using `kast` that indeed, the bracket production defined above does
not appear in the resulting output kore on these two examples.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,362 @@
...
+reduce-reduce conflict. In this case, you can either modify your grammar to be
+LR(1), or you can enable use of Bison's GLR support by instead passing 
+`--gen-glr-bison-parser` to `kompile`. Note that if your grammar is ambiguous,
+the ahead-of-time parser will not provide you with particularly readable error
+messages at this time.
+
+If you have a K definition named `foo.k`, and it generates a directory when
+you run `kompile` called `foo-kompiled`, you can invoke the ahead-of-time
+parser you generated by running `foo-kompiled/parser_PGM <file>` on a file.
+
```
ehildenb, line 374:
```suggestion

Time the difference in parse time for the program `and-or.bool` above using K's
just-in-time parser (`time kast and-or.bool --output kore`), and the
ahead-of-time parser (`time lesson-03-kompiled/parser_PGM and-or.bool`).
Note that the ahead-of-time parser can _only_ output kore, so no need for the
`--output kore` flag.
```

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,296 @@
+# Lesson 1.4: Disambiguating Parses
+
+The purpose of this lesson is to teach how to use K's builtin features for
+disambiguation to transform an ambiguous grammar into an unambiguous one that
+expresses the intended the intended ASTs.
```
ehildenb:
```suggestion
expresses the intended ASTs.
```

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,296 @@
...
+Well, the rule for priority is slightly more complex than previously described.
+In actual fact, it applies only conditionally. Specifically, it applies in
+cases where the child is either the first or last production item in the
+parent's production. For example, in the production `Bool "&&" Bool`, the
+first `Bool` non-terminal is not preceded by any terminals, and the last `Bool`
+non-terminal is not followed by any terminals. As a result of this, we apply
+the priority rule to both children of `&&`. However, in the `()` operator, 
+the sole non-terminal is both preceded by and followed by terminals. As a
+result, the priority rule is not applied when `()` is the parent. Because of
+this, the program we mentioned above successfully parses.
```
ehildenb, line 95:
```suggestion
this, the program we mentioned above successfully parses.

Inspect the kore output by parsing the program `true && true || false`, to make
sure it selects the correct parse. Switch the priority of `&&` and `||` and re-parse,
and inspect that the alternate parse is selected.
```

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+endmodule
+```
+
+In this example, unlike the one above, `&&`, `^`, and `||` have the same
+priority. However, viewed as a group, the entire group is left associative.
+This means that none of `&&`, `^`, and `||` can appear as the right child of
+any of `&&`, `^`, or `||`. As a result of this, this grammar is also not
+ambiguous. However, it expresses a different grammar, and you are encouraged
+to think about what the differences are in practice.
```
ehildenb, line 175:
```suggestion
to think about what the differences are in practice.

Try again to parse the program `true && true || false`, and then the program
`true || true && false` using the left-associative grammar in `LESSON-04-C`.
Inspect the kore output for each, to understand how it selects each parse.
You can also try the same with `left` replaced by `right` in the grammar.
```

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+```k
+module LESSON-04-D
+
+  syntax Boolean ::= "true" [literal] | "false" [literal]
+  syntax Boolean ::= "(" Boolean ")" [atom, bracket]
+  syntax Boolean ::= "!" Boolean [not, function]
+  syntax Boolean ::= Boolean "&&" Boolean [and, function]
+  syntax Boolean ::= Boolean "^" Boolean [xor, function]
+  syntax Boolean ::= Boolean "|" Boolean [or, function]
```
ehildenb:
```suggestion
  syntax Boolean ::= "true"               [literal]
                   | "false"              [literal]
                   | "(" Boolean ")"      [atom, bracket]
                   | "!" Boolean          [not, function]
                   | Boolean "&&" Boolean [and, function]
                   | Boolean "^" Boolean  [xor, function]
                   | Boolean "|" Boolean  [or, function]
```

Might as well get them in the habit of seeing/writing more legible K early.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,296 @@
...
+                | ";"
+endmodule
+```
+
+Here we have added the `avoid` attribute to the `else` production. As a result,
+when an ambiguity occurs and one or more of the possible parses has that symbol
+at the top of the ambiguous part of the parse, we remove those parses from
+consideration and consider only those remaining. The `prefer` attribute behaves
+similarly, but instead removes all parses which do not have that attribute.
+In both cases, no action is taken if the parse is not ambiguous.
```
ehildenb, line 297:
```suggestion
In both cases, no action is taken if the parse is not ambiguous.

Parse the program `if (true) {} if (false) {} else {}` using the grammar of
`LESSON-04-F`, to see how it builds the kore AST. Now place the `avoid`
attribute on the other `"if"` production instead, and see how this changes the
kore AST. Alternatively, just swap in `prefer` for the existing `avoid`.
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
dwightguth: This is redundant; we already talk about using kast and noticing that the bracket doesn't appear in the resulting AST.

dwightguth:
I addressed some of your comments here. It seems a number of them seem to be trying to insert miniature exercises into the body of the lesson, though. This is not the (current) structure of the lesson, and it breaks up the flow considerably, not to mention it's a considerable deviation in the narrative style of the text. It's also an anomaly when considering the structure of previous lessons. I'm not necessarily opposed to having this type of miniature exercise in the text (although I wonder if we are overdoing things), but we need to come up with a much better way of embedding it into the document if we are going to.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -1 +1,397 @@
...
+)
+```
+
+Note how the first output is largely a name-mangled version of the second
+output. The one difference is the presence of the `inj` symbol in the KORE
+output. We will talk more about this in later lessons.
+
+### Exercise
+
+Parse the expression `false || true` with `--output kastr`. See if you can
```
ehildenb:
```suggestion
Parse the expression `false || true` with `--output kast`. See if you can
```

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ehildenb: We talk about it, but this is about giving the user an exercise (so you could put the `### Exercise` block here). I think many people learn better if they are given little tasks to confirm their understanding, so that they have to figure out "what do I need to type in the terminal to confirm that the brackets do not show up?" and "how do I read the output and make sure that the bracket does not show up?"

ehildenb: Ah I see you gave exactly this exercise below. Sounds good!

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,312 @@
...
+the priority rule to both children of `&&`. However, in the `()` operator, 
+the sole non-terminal is both preceded by and followed by terminals. As a
+result, the priority rule is not applied when `()` is the parent. Because of
+this, the program we mentioned above successfully parses.
+
+### Exercise
+
+Parse the program `true && false || false` yourself, and confirm that the AST
+places `||` as the top level symbol. Then modify the definition so that you
+will get the opposite parse.
```
ehildenb:
```suggestion
will get the alternative parse.
```

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -0,0 +1,312 @@
...
+This means that none of `&&`, `^`, and `||` can appear as the right child of
+any of `&&`, `^`, or `||`. As a result of this, this grammar is also not
+ambiguous. However, it expresses a different grammar, and you are encouraged
+to think about what the differences are in practice.
+
+### Exercise
+
+Parse the program `true && false && false` yourself, and confirm that the AST
+places the rightmost `&&` at the top of the expression. Then modify the 
+definition to generate the opposite parse.
```
ehildenb:
```suggestion
definition to generate the alternative parse.
```

ehildenb (approved):

PR #1853, opened 2021-03-11
Fix1844 krun crash

### LINKED ISSUE #1844: [Bug] [krun/kompile] - Can no longer run definitions that don't use `$PGM`
K Version
---------

```
RV-K version 1.0-SNAPSHOT
Git revision: 7ac809a
Git branch: UNKNOWN
Build date: Fri Mar 05 00:15:13 CST 2021
```

Description
-----------

Running `krun` with a configuration that does not use `$PGM` results in a unhelpful error.
This used to work previously.

Reproducing
-----------

```
$ cat test.k 
module TEST
    imports BOOL

    syntax FOO ::= "foo"
    configuration <k> foo </k>
endmodule
$ krun 
kore-exec: [88663] Error (ErrorVerify):
    Error:
      symbol or alias 'LblinitGeneratedTopCell' (.krun-2021-03-05-00-18-36-U7im2v8NSF/tmp.in.1FEhIcKWgc 1:1):
        Expected 0 operands, but got 1.
[Error] Critical: Backend crashed during rewriting with exit code 1
```

Expected Behavior
-----------------

Either:

1. A clear and concise error message saying that the `$PGM` variable is required.
2. Successfully running the configuration, resulting in a stuck configuration `<k> foo </k>`

radumereuta:
Fixes: #1844
When https://github.com/kframework/llvm-backend/pull/400 gets merged into master, this should pass.

Added:
- `krun.nopgm` - if such a file exists in the testing directory, then it's going to call krun with no arguments
- tests for all backends for `krun.nopgm`
- krun detects which backend was used to kompile from a file now, to avoid caching issues when compiling the same definition with llvm then haskell.


k-distribution/src/main/scripts/bin/krun
```diff
@@ -572,6 +572,8 @@ else
     result=$?
     set -e
   fi
+else
+  error "Backend $backendName not supportd. Use krun-legacy."
```
ehildenb:
```suggestion
  error "Backend $backendName not supported. Use krun-legacy."
```

ehildenb (approved):

PR #1856, opened 2021-03-13
Use a global Stack cache on macOS

ttuegel:
Set HOMEBREW_NO_ENV_FILTERING=1 so that we can point Stack at the global cache
in the Anka template. Otherwise, Stack-inside-Homebrew will download and compile
all dependencies on every CI job.

See also: #1813

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -336,6 +336,7 @@ pipeline {
                 stage('Build Homebrew Bottle') {
                   agent { label 'anka' }
                   environment {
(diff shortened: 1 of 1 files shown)


ttuegel:
Homebrew warns about using `HOMEBREW_NO_ENV_FILTERING`:

```
Warning: HOMEBREW_NO_ENV_FILTERING is undocumented, deprecated and will be removed in a future Homebrew release (because it breaks many things)!
```

Also, the Anka VM is missing the cache at `/opt/stack`.

PR #1857, opened 2021-03-15
Semantic versioning

### LINKED ISSUE #1785: [Bug] kompile --version does not show git commit ID when kompile was installed via brew
This is an issue that was reported to us via the Algorand folks. Apparently when you do `brew install kframework/k/kframework` followed by `kompile --version`, it doesn't know what the branch and commit hash are that it should show. This makes it difficult for K developers to determine what version of K our customers have installed, so it should be viewed as an urgent issue. I have not reproduced locally yet; it's possible that this may affect only certain versions of Mac OS.

ehildenb:
Fixes: https://github.com/kframework/k/issues/1785

- `master` branch is no longer where releases happen from, rather from `release` branch.
- `master` branch job now only does one thing: bumps version numbers and pushes to `release` branch.
- Both submodule/release-tag versions are bumped from `release` job, and packaging is done there.


dwightguth:
I wonder if it would be better to have some kind of release-5.0, release-5.1, etc naming structure for release branches. This is more in alignment with how other projects with major-minor versions deal with releases, would get around the issue of having to deal with changes to the major or minor version, and would also have the benefit that we could in theory push bugfixes to old releases if it ultimately proves necessary.

ehildenb:
This is an example versioned bump that went through successfully using this workflow: https://github.com/kframework/k/releases/tag/v5.0.0

ehildenb:
Evidence this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/release/7/pipeline/351

Here is an example where the auto-updater worked, but the repo needs to adjust how it reads the version tag: https://github.com/kframework/kore/pull/2490

dwightguth (commented):
I don't understand how you can commit a file containing a git commit hash. The act of doing so would render the hash out of date. Why not just remove the commit hash from the version message entirely? It should be simple enough to look at the tags we push to git and find the tag that corresponds to a particular patch version.

ehildenb:
The file is committed on teh `release` branch, but the release is applied to the correct commit on teh `master` branch. So for the release, it is correct @dwightguth .

I'm also happy to remove it altogether, I just figured people wanted it in there because it was there previously. I tried to match the previous version message as close as possible.

dwightguth (approved):
Looks fine assuming you tested the version date thing.

PR #1862, opened 2021-03-16
Include Haskell backend with Nix packages

ttuegel:
This bundles the Haskell backend with the frontend by default, as users expect. At one time, both backends were optional in the Nix package and users could choose, but the LLVM backend later became required for `krun`. At this point, it's just confusing to have one backend be privileged over the other.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,7 +31,7 @@ let
   inherit (llvm-backend-project) clang llvm-backend;
 
   k = callPackage ./nix/k.nix {
-    inherit llvm-backend mavenix;
+    inherit haskell-backend llvm-backend mavenix;
   };
 
   haskell-backend-project = import ./haskell-backend/src/main/native/haskell-backend {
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -31,7 +31,7 @@ let
   inherit (llvm-backend-project) clang llvm-backend;
 
   k = callPackage ./nix/k.nix {
```
andreiburdusa, line 34: Can you give me some details about what `callPackage` does?

andreiburdusa (approved):

default.nix
ttuegel: `callPackage` calls a function in another file with default arguments, like this:

1. `callPackage` `import`s the file (first argument). The file should be a Nix function taking a set of arguments (`{a, b, ...}:`).
2. `callPackage` applies the argument set (second argument) to the imported function.
3. If any of the arguments of the imported function are still missing, `callPackage` looks them up in `pkgs`.

andreiburdusa: Now I see it's presented in the 13th Nix Pill, but after the 12th I thought I don't have a good enough understanding even though I completed the exercises, so I started reading Eelco Dolstra's thesis. Maybe I should resume reading the pills, now that I read some chapters from the thesis.

ttuegel: I don't know... I've never read his thesis myself. I know that some of the internal details are only documented ther, but I don't know if it's necessary to understand the internals.

PR #1863, opened 2021-03-16
Add a Nix lock file to releases

ttuegel:
This is necessary to automatically update repositories which depend on kframework/k but do not include it as a submodule.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: "Release"
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  nix-release:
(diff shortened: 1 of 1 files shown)


andreiburdusa (commented):
My understanding is that the purpose of this pull request is to update a nix file that tells the repository to update k. Is it correct?

ttuegel:
> My understanding is that the purpose of this pull request is to update a nix file that tells the repository to update k.

Something like that: it attaches a file to the release page (on GitHub) that tells the other repository where to find the new source code.

andreiburdusa (approved):

PR #1868, opened 2021-03-18
Add definedness conditions to some symbolic map rules

ana-pantilie:
This is needed for https://github.com/kframework/k/pull/1830

### DIFF
--- a/k-distribution/tests/regression-new/map-symbolic-tests-haskell/remove-9-spec.k.out
+++ b/k-distribution/tests/regression-new/map-symbolic-tests-haskell/remove-9-spec.k.out
@@ -27,6 +27,12 @@
   }
 #And
   {
+    false
+  #Equals
+    X:MyId in_keys ( MAP [ Y:MyId <- undef ] )
+  }
+#And
+  {
     true
   #Equals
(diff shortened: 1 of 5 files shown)


k-distribution/tests/regression-new/map-symbolic-tests-haskell/remove-9-spec.k.out
```diff
@@ -25,6 +25,12 @@
   #Equals
     X:MyId in_keys ( MAP )
   }
+#And
+  {
+    false
+  #Equals
+    X:MyId in_keys ( MAP [ Y:MyId <- undef ] )
```
ttuegel, line 32: Can we add rules to simplify this more? (If so, we can do it in another pull request.)

ttuegel (approved):

k-distribution/tests/regression-new/map-symbolic-tests-haskell/remove-9-spec.k.out
ana-pantilie: I noticed this as well. We should look into it as part of another issue I think.

dwightguth (approved):

PR #1870, opened 2021-03-19
Forbid macro att on function rules

### LINKED ISSUE #1852: [kompile] - Having `function` on `macro` symbols should be a kompile-time error
K Version
---------

Example:

```
$ kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: aa04b03
Git branch: UNKNOWN
Build date: Sat Mar 06 18:12:48 UTC 2021
```

Description
-----------

Running `kompile` on a definition which contains a `macro` rule for a `function` symbol produces no error, but makes it so that `kore-expand-macros` fails every time with a terrible error message.

Input Files
-----------

`test.k`

```
module TEST
    imports INT

    syntax Int ::= foo ( Int ) [function]
    rule foo(I) => 3 +Int I [macro]
endmodule
```

Reproduction Steps
------------------

Kompiling produces no error:

```
$ kompile test.k
```

Expected Behavior
-----------------

Calling `kompile` on this definition would result in an error about the production `foo` being both a `function` and a `macro`.

radumereuta:
Fixes: #1852

I'm assuming all of these attributes are not allowed on function rules: macro|macro-rec|alias|alias-rec
How about claims. What attributes are not allowed?

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckFunctions.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckFunctions.java
@@ -90,7 +90,7 @@ public class CheckFunctions {
         new RewriteAwareVisitor(true, errors) {
             @Override
             public void apply(KApply k) {
-                if (isRHS() || k.klabel() instanceof KVariable || CheckKLabels.isInternalKLabel(k.klabel().name(), m) || !m.attributesFor().contains(k.klabel())) {
+                if (isRHS() || k.klabel() instanceof KVariable || !m.attributesFor().contains(k.klabel())) {
                     return;
                 }
                 if (k.klabel().name().equals("#withConfig")) {
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckFunctions.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckFunctions.java
@@ -40,12 +40,19 @@ public class CheckFunctions {
                 check(rl.body());
         } else if (sentence instanceof Claim) {
             // functions are allowed on LHS of claims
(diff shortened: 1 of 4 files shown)


radumereuta:
This is failing: https://github.com/kframework/k/blob/master/k-distribution/tests/regression-new/proof-tests/deposit/spec/deposit-symbolic.k#L8
Are we sure we want to enforce this restriction?

dwightguth:
Claims can't be macros so you can forbid all of those attributes on claims.

kernel/src/main/java/org/kframework/compile/checks/CheckFunctions.java
```diff
@@ -83,4 +85,26 @@ public void apply(KApply k) {
             }
         }.apply(body);
     }
+
+    public void checkFuncAtt(RuleOrClaim r) {
+        new RewriteAwareVisitor(true, errors) {
+            @Override
+            public void apply(KApply k) {
+                if (isRHS() || k.klabel() instanceof KVariable || !m.attributesFor().contains(k.klabel())) {
```
dwightguth: Instead of checking isRHS, which will return true outside of any rewrite, you probably want to check `isRHS() && !isLHS()`

dwightguth (changes requested):
Please add a test that this check correctly handles rules with function context. You don't seem to have implemented this the usual way, so I'd like to be certain.

radumereuta:
@dwightguth I added more tests.
I also added the check to `context` and `context alias`. Please confirm which attributes are allowed because I couldn't find hard evidence of what is allowed where.

dwightguth (approved):

PR #1874, opened 2021-03-23
Pyk updates

ehildenb:
Updates `krunJSON` to use the newer `krun` instead of `krunLegacy`. Renames the existing `krunJSON` as `krunJSONLegacy`, so that people can continue using it as needed.
Also quiet down the rule coverage converter (used for rule merging).

This requires an explicit call to `kast`.

If you are passing in an entire configuration to `krunJSON`, make sure to add `krunFlags = ['--term']`, and `kastFlags = ['--sort', 'GeneratedTopCell']`. If you are just parsing some initial statement, this isn't needed.


yzhang90 (approved):
LGTM

PR #1884, opened 2021-03-29
Nix: Fix kserver

ttuegel:
The `kserver` that comes with the Nix package is broken, but this pull request would fix it.


ttuegel:
The tests fail because of the lines get wrapped differently, so we need to disable line wrapping in the test suite. There is a command-line option for `kompile` and `krun` to do this.

andreiburdusa (approved):
The only documentation I could find about `buildMaven` is this https://github.com/fzakaria/nixpkgs/blob/b9321adf1ea91bca9e2127ce4f0b428391745474/doc/languages-frameworks/maven.section.md and it doesn't give details about `preFixup`, `doInstallCheck` or `installCheckPhase`. Is there another place that explains all its parameters?

ttuegel:
Run tests individually: `nix-build test.nix --argstr test NAME`

PR #1894, opened 2021-04-02
Fix for `kompile --coverage` when rhs of rule contains parametric function

dwightguth:
There was a bug where the semantics would get stuck if you enable coverage and it tries to apply a function rule where the top symbol on the rhs is parametric in the return sort. This was because the instrumentation with the coverage information caused incorrect sort information to be inferred about the rule during AddSortInjections. The solution turns out to be to infer the sort a little differently and also ensure that the instrumented code still expects the same sort in the context of the rhs.

The test I added fails without these changes. I also tested that the Algorand semantics can correctly generate a coverage xml file when built with this version of K.


kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -303,6 +303,10 @@ private void match(Production prod, Sort declaredSort, Sort actualSort, Map<Sort
                         .collect(Collectors.toSet());
     }
 
+    public Sort topSort(K term) {
+      return sort(term, freshSortParam());
+    }
+
```
radumereuta, line 312: Why is this called `topSort`?
Please add some comments because it's not clear what top sort means in this context.

k-distribution/tests/regression-new/coverage-poly/test.k
```diff
@@ -0,0 +1,7 @@
+module TEST
+  imports K-REFLECTION
+
+  syntax Foo ::= parseFoo() [function]
+  rule parseFoo() => #parseKORE("Lblfoo{}()")
+  syntax Foo ::= foo() [symbol]
+endmodule
```
radumereuta, line 7: Is this test helpful other than 'it doesn't crash'?

radumereuta (changes requested):
Just some questions, and a request for comments.
Other than that it looks good.

k-distribution/tests/regression-new/coverage-poly/test.k
dwightguth: Yes. The test got stuck before I made these changes, because when --coverage was passed to kompile, the rule for `parseFoo` was being instrumented in a manner that caused the rule to not apply.

radumereuta (approved):

PR #1895, opened 2021-04-02
Keep only the files required by dry-run

### LINKED ISSUE #1850: [Bug] krun - --dry-run doesn't save temporary files
K Version
---------

```
$ kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: 4c3c90e
Git branch: add-ref-default.nix
Build date: Tue Mar 30 14:29:27 EEST 2021
```

Description
-----------

Running `krun` with `--dry-run` prints a command that can't be executed because some temporary files are missing.
But those files are present after executing the command with `--save-temps`.

Input Files
-----------

`test.k`

```
module TEST
    imports INT
    syntax FOO ::= "foo"
    configuration <k> foo ~> $PGM:Int </k>
    rule <k> foo => #Bottom ... </k>
endmodule
```

`2.test`

```
1
```

Reproduction Steps
------------------

```
$ kompile --backend haskell -d . test.k
$ krun --search-final --dry-run --haskell-backend-command /home/andrei/kore/.build/kore/bin/kore-exec -d . 2.test
/home/andrei/kore/.build/kore/bin/kore-exec ./test-kompiled/definition.kore --module TEST --pattern .krun-2021-03-11-17-48-49-xUJZdjQeBL/tmp.in.zXImruBIo3 --output .krun-2021-03-11-17-48-49-xUJZdjQeBL/result.kore --searchType FINAL --search .krun-2021-03-11-17-48-49-xUJZdjQeBL/tmp.pattern.PmSzl0lJTD
```
Then execute the resuled command. In this case it's
```
$ /home/andrei/kore/.build/kore/bin/kore-exec ./test-kompiled/definition.kore --module TEST --pattern .krun-2021-03-11-17-48-49-xUJZdjQeBL/tmp.in.zXImruBIo3 --output .krun-2021-03-11-17-48-49-xUJZdjQeBL/result.kore --searchType FINAL --search .krun-2021-03-11-17-48-49-xUJZdjQeBL/tmp.pattern.PmSzl0lJTD
.krun-2021-03-11-17-48-49-xUJZdjQeBL/tmp.pattern.PmSzl0lJTD: copyFile:atomicCopyFileContents:withReplacementFile:copyFileToHandle:openBinaryFile: does not exist (No such file or directory)
```

Expected Behavior
-----------------

The command printed by using `--dry-run` should give the same result as
```
$ krun --search-final  --haskell-backend-command /home/andrei/kore/.build/kore/bin/kore-exec -d . 2.test
```
which is
```
#Bottom
```

radumereuta:
Fixes: #1850
Keeping better track of which files are required by --dry-run and excluding them from the tempFiles list.


dwightguth (changes requested):
This doesn't seem to be correctly handling the case where the script exits abnormally.

radumereuta:
@dwightguth can you be more specific?
What I've done is to keep track of what files are being used to run the backend and remove them from the list of files that are to be deleted. If the script exits abnormally, then the behavior is just as before. Delete all temp files, or keep all temp files if you specified `--save-temps`.
`--dry-run` is an option that should work when the definition normally goes through.

ehildenb:
Let's make sure we continue to handle this case "incrementally" (any intermediate produced files should be kept on abnormal exit).

radumereuta:
I made a function that decides which files to keep if `--dry-run` was provided to keep things tidy.
This is called before constructing the final execute command, which is also the soonest I know for sure which files to keep.

> Let's make sure we continue to handle this case "incrementally" (any intermediate produced files should be kept on abnormal exit).

I disagree with this exact statement. None of our other tools behave in this way. The temporary files are kept only if the `--debug` option is provided. In the krun script, the best equivalent would be `--save-temps`. I still believe it should be renamed to `--debug` because we get to confusing situations like this one.

k-distribution/src/main/scripts/bin/krun
```diff
@@ -45,6 +45,21 @@ tempFiles+=("$input_file")
 tempFiles+=("$expanded_input_file")
 tempFiles+=("$kore_output")
 
+keepTempsIfDryRun () {
+  # kepp files used by dry-run
+  if $dryRun; then
+    local newArray=()
+    for value in "${tempFiles[@]}"; do
+        if [[ ! $@ =~ ${value} ]]; then
```
dwightguth: $@ should always be quoted

dwightguth (approved):
One minor change, otherwise seems okay. I'm not convinced that the amount of quoting of filenames is adequate, but appropriate testing to ensure that the script works on filenames containing spaces is out of scope for this change, so we can ignore it for now. Please address the comment I made before merging though.

k-distribution/src/main/scripts/bin/krun
```diff
@@ -45,6 +45,21 @@ tempFiles+=("$input_file")
 tempFiles+=("$expanded_input_file")
 tempFiles+=("$kore_output")
 
+keepTempsIfDryRun () {
+  # kepp files used by dry-run
+  if $dryRun; then
+    local newArray=()
+    for value in "${tempFiles[@]}"; do
```
ehildenb, line 52: Will this work on MacOS? I thought that it only worked in Bash 4+ to have an empty array and then append to it later.

radumereuta: I think the problem was with associative arrays. Anyways, I've tested it with the mac version and it works:
```
radu@radu:~/work/t e s t$ cat test.sh
krun --dry-run 1.test
radu@radu:~/work/t e s t$ ../bash/bash-3.2.57/bash test.sh
kore-exec ./test-kompiled/definition.kore --module TEST --pattern .krun-2021-04-15-21-51-20-Z2RhjNb9Yx/tmp.in.C1RvjO4LLV --output .krun-2021-04-15-21-51-20-Z2RhjNb9Yx/result.kore
radu@radu:~/work/t e s t$ ls -a
.  ..  1.test  .krun-2021-04-15-21-51-20-Z2RhjNb9Yx  test.k  test-kompiled  test.sh
```

PR #1899, opened 2021-04-05
Add definedness conditions to simplification rules in domains.md

### LINKED ISSUE #2487: Audit simplification rules in the frontend
Some of the simplification rules we have added to the frontend are not valid because they have missing definedness conditions. In general, the `ensures` clause of the rule should be used to ensure that the left-hand side of the simplification rule is defined. We should audit all the simplification rules in `domains.md` to find any other problems.

I suspect that `simplification` rules with partial functions should never have anonymous variables, but we will have to see if that holds for the other rules. If possible, we should come up with a check that the backend can perform to warn users when they may have written unsound simplification rules.

andreiburdusa:
Fixes https://github.com/kframework/kore/issues/2487


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -419,15 +419,15 @@ module MAP-KORE-SYMBOLIC [kore,symbolic]
   rule #Ceil(@M:Map [@K:KItem]) => {(@K in_keys(@M)) #Equals true} #And #Ceil(@M) #And #Ceil(@K) [anywhere, simplification]
 
   // Symbolic update
-  rule (K |-> _ M:Map) [ K <- V ] => (K |-> V M) [simplification]
+  rule (K |-> _ M:Map) [ K <- V ] => (K |-> V M) ensures notBool (K in_keys(M)) [simplification]
```
ehildenb: Is this not already implied by `K |-> _ M` being a well-defined map?

ttuegel (changes requested):
These `ensures` clauses are redundant because these simplification rules already preserve `#Bottom`. Let's add comments at each rule explaining why the `ensures` is not needed.

ttuegel (changes requested):
Please _remove_ the `ensures` clauses and leave a comment explaining that _adding_ them would be redundant.

ttuegel (approved):

dwightguth (approved):

PR #1900, opened 2021-04-06
Remove un-needed rule

MirceaS:
The rule `rule .Bytes => String2Bytes("")` is causing some problems when validating kore definitions in the Haskell backend as the resulting pattern doesn't follow the format that we impose on function rules. We suggest removing this rule as it doesn't currently serve a purpose from what we can tell.


MirceaS:
Also the comment above where the rule used to be refers to the following module so I also moved the `endmodule` terminator accordingly.

dwightguth:
I have no objection to this, but @ttuegel needs to review since he was the one who originally added this rule.

dwightguth (approved):
Approving, but please don't merge without Tom's approval.

ttuegel (approved):
I have no recollection of adding this rule.  :shrug:

MirceaS:
@dwightguth could it have been someone else that added that rule, that you might want to wait for a review from, or am I good to merge?

ttuegel:
I did add that rule, I just don't have any recollection of why it was necessary. It doesn't make sense because that function is hooked.

PR #1904, opened 2021-04-07
Front end cleanup

radumereuta:
Splitting the rule refactoring into more pieces.
This is solving:
- interleaving in Jenkins output. This should make it easier to identify the output of each file.
- one test was not cleaning the temp files correctly
- quadratic behavior in FlatModule creation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -81,7 +81,7 @@ pipeline {
                           echo 'Starting kserver...'
                           k-distribution/target/release/k/bin/spawn-kserver kserver.log
                           cd k-exercises/tutorial
-                          make -j`nproc` ${MAKE_EXTRA_ARGS}
+                          make -j`nproc` --output-sync ${MAKE_EXTRA_ARGS}
                         '''
                       }
                       post {
@@ -398,7 +398,7 @@ pipeline {
                       spawn-kserver $WD/kserver.log
                       cd pl-tutorial
                       echo 'Testing tutorial in user environment...'
-                      make -j`sysctl -n hw.ncpu` ${MAKE_EXTRA_ARGS}
+                      make -j`sysctl -n hw.ncpu` --output-sync ${MAKE_EXTRA_ARGS}
                       cd ~
                       echo 'module TEST imports BOOL endmodule' > test.k
                       kompile test.k --backend ocaml
--- a/kernel/src/main/java/org/kframework/parser/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/ParserUtils.java
@@ -4,6 +4,7 @@ package org.kframework.parser;
 import org.apache.commons.io.FileUtils;
 import org.kframework.attributes.Att;
 import org.kframework.attributes.Source;
+import org.kframework.definition.FlatModule;
 import org.kframework.definition.Module;
 import org.kframework.definition.ModuleTransformer;
 import org.kframework.kil.Definition;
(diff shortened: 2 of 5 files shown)


kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -104,6 +104,9 @@ case class FlatModule(name: String, imports: Set[Import], localSentences: Set[Se
       throw KEMException.compilerError(msg)
     }
 
+    if (koreModules.contains(mainModName))
+      return koreModules(mainModName)
+
```
ehildenb: We should be returning the `FlatModule` associated with `name`, not with `mainModName`.
Shouldn't this read:

```suggestion
    if (koreModules.contains(name))
      return koreModules(name)

```

ehildenb: Ah I see above we say that `mainModName = this.name`, but I would still prefer we use `this.name` here, as it expresses the intent of this function better (and thus may behave better for later changes which may introduce or change code here).

Jenkinsfile
```diff
@@ -81,7 +81,7 @@ pipeline {
                           echo 'Starting kserver...'
                           k-distribution/target/release/k/bin/spawn-kserver kserver.log
                           cd k-exercises/tutorial
-                          make -j`nproc` ${MAKE_EXTRA_ARGS}
+                          make -j`nproc` --output-sync ${MAKE_EXTRA_ARGS}
```
ehildenb, line 84: Did you try `--output-sync=recurse` instead?

kore/src/main/scala/org/kframework/definition/outer-ext.scala
radumereuta: `var mainModName = this.name`
It's the same thing. I'm going to remove `mainModName` since it's redundant and confusing.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -220,11 +218,16 @@ private String loadDefinitionText(File definitionFile) {
...
+        previousModules.stream().map(m -> koreModules.put(m.name(), m)).collect(Collectors.toList()); // populate koreModules with previousModules
         HashSet<org.kframework.kil.Module> kilModulesSet = new HashSet<>(kilModules);
 
-        Set<Module> finalModules = kilModules.stream().map(m -> kilToKore.apply(m, kilModulesSet, koreModules)).flatMap(m -> Stream.concat(Stream.of(m), Stream.of(koreModules.get(m.name() + "$SYNTAX")))).collect(Collectors.toSet());
+        java.util.List<FlatModule> flatModules = kilModulesSet.stream().map(kilToKore::toFlatModule).sorted(Comparator.comparing(FlatModule::name)).collect(Collectors.toList());
+
+        // accumulate in koreModules the new modules. Use an ordered set to stabilize the error message for circular imports
+        flatModules.stream().map(m -> m.toModule(immutable(new LinkedHashSet<>(flatModules)), koreModules, Seq())).collect(Collectors.toSet());
+        Set<Module> finalModules = mutable(koreModules.values().toSet());
+
```
ehildenb, line 230: I don't feel comfortable reviewing this piece of code, because I don't understand the intricacies of Java vs. Scala datastructures and mutable vs immutable datastructures in each. @dwightguth can you have a look at this one?

Jenkinsfile
radumereuta: Yes. This way is easier to read. Every call is surrounded by `enter` and `leaving` directory messages. The output is longer, but it makes it easier to track down failing tests.

ehildenb (approved):

kernel/src/main/java/org/kframework/parser/ParserUtils.java
ehildenb: @radumereuta just looking at this code, it looks like we'll be calling `m.toModule` on each module in `flatModules`, but it seems to me that this may repeat work because perhaps the first module you call it on ends up already converting the second module you call it on. Is this the case?

radumereuta: Yes, that's why I added the caching step below.
```
if (koreModules.contains(this.name))
      return koreModules(this.name)
```

ehildenb: But will each subsequent invocation of `toModule` have the updated map from the previous invocation present? I can't quite follow the control flow here, but are you saying that each call of `toModule` will see the updated `koreModules` from the previous call?

ehildenb (approved):

kernel/src/main/java/org/kframework/parser/ParserUtils.java
ehildenb: From discussions on Slack, it looks like `koreModules` is mutable and updated in place, so subsequent invocations will see the prior updated `koreModules` map.

Issue #1867, opened 2021-03-17
Better scanner handling


radumereuta:
`kompile` and `kprove` both need to parse rules, but when running a second time, even if none of the rules were changed, they still generate some new scanners.
Here is the output of the second run of kompile on a simple test.k file.
```
Outer parsing [124 modules]                                  =  1.128s
New scanner:TEST$SYNTAX-CONFIG-CELLS                         =  0.117s
New scanner:TEST-CONFIG-CELLS                                =  0.147s
New scanner:STDOUT-STREAM$SYNTAX-CONFIG-CELLS                =  0.211s
New scanner:STDOUT-STREAM-CONFIG-CELLS                       =  0.224s
New scanner:STDIN-STREAM$SYNTAX-CONFIG-CELLS                 =  0.215s
New scanner:STDIN-STREAM-CONFIG-CELLS                        =  0.223s
Parse configurations [0/6 declarations]                      =  1.449s
New scanner:TEST-RULE-CELLS                                  =  0.143s
New scanner:K-EQUAL-RULE-CELLS                               =  0.129s
New scanner:INT-SYMBOLIC-RULE-CELLS                          =  0.128s
New scanner:INT-SYMBOLIC-KORE-RULE-CELLS                     =  0.133s
New scanner:INT-KORE-RULE-CELLS                              =  0.153s
New scanner:K-EQUAL-KORE-RULE-CELLS                          =  0.154s
New scanner:STRING-KORE-RULE-CELLS                           =  0.162s
New scanner:INT-RULE-CELLS                                   =  0.181s
New scanner:STRING-COMMON-RULE-CELLS                         =  0.186s
New scanner:K-IO-RULE-CELLS                                  =  0.231s
New scanner:STDIN-STREAM-RULE-CELLS                          =  0.247s
New scanner:STDOUT-STREAM-RULE-CELLS                         =  0.252s
Parse rules [0/96 rules]                                     =  0.576s
```
Even though 0 rules were parsed, there are 18 scanners being generated.
This is the output of running kprove twice on a KEVM spec file:
```
New scanner:EVM$SYNTAX-CONFIG-CELLS                          =  0.494s
New scanner:EVM-CONFIG-CELLS                                 =  1.053s
New scanner:DAI-SYMBOL-PASS-SPEC-RULE-CELLS                  =  1.270s
Parse spec modules [0/1586 rules]                            =  7.210s
```
Again, 0 rules were parsed, but we're still spending time generating scanners.

[...]

PR #1914, opened 2021-04-12
Refactor rule parsing

radumereuta:
Part 1 in fixing: #1867
It is slightly faster:
```
          this branch (master)
test.k:          4.0s (4.0s)
time make -j 32 -O
regression-new: 8m24s (8m46s)
    kserver:    2m45s (2m52s)
pl-tutorial:    8m19s (8m13s)
    kserver:    1m33s (1m40s)
```
Here is the output of kompile on a test file with one rule and a configuration:
```
Outer parsing [133 modules]                                  =  0.884s
New scanner: TEST$SYNTAX-CONFIG-CELLS                        =  0.139s
New scanner: TEST-CONFIG-CELLS                               =  0.163s
New scanner: STDOUT-STREAM$SYNTAX-CONFIG-CELLS               =  0.227s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.232s
New scanner: STDIN-STREAM$SYNTAX-CONFIG-CELLS                =  0.211s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.236s
Parse configurations [6/6 declarations]                      =  1.695s
New scanner: TEST-RULE-CELLS                                 =  0.151s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.234s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.226s
Parse rules [101/101 rules]                                  =  0.672s
```
second time:
```
Outer parsing [133 modules]                                  =  0.858s
New scanner: TEST$SYNTAX-CONFIG-CELLS                        =  0.130s
New scanner: TEST-CONFIG-CELLS                               =  0.161s
New scanner: STDOUT-STREAM$SYNTAX-CONFIG-CELLS               =  0.211s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.227s
New scanner: STDIN-STREAM$SYNTAX-CONFIG-CELLS                =  0.237s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.234s
Parse configurations [0/6 declarations]                      =  1.528s
Parse rules [0/101 rules]                                    =  0.081s
```
The diff was getting pretty big so I'm limiting the changes to rules only.
I'm going to start working on the configuration after this gets merged. I will deal with code duplication at the same time.

Just as a comparison. Here is the output of -v on master:
```
Outer parsing [133 modules]                                  =  0.836s
New scanner: TEST$SYNTAX-CONFIG-CELLS                        =  0.148s
New scanner: TEST-CONFIG-CELLS                               =  0.166s
New scanner: STDOUT-STREAM$SYNTAX-CONFIG-CELLS               =  0.218s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.227s
New scanner: STDIN-STREAM$SYNTAX-CONFIG-CELLS                =  0.209s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.242s
Parse configurations [6/6 declarations]                      =  1.650s
New scanner: TEST-RULE-CELLS                                 =  0.146s
New scanner: STRING-KORE-RULE-CELLS                          =  0.177s
New scanner: STRING-COMMON-RULE-CELLS                        =  0.179s
New scanner: K-IO-RULE-CELLS                                 =  0.279s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.287s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.287s
Parse rules [101/101 rules]                                  =  0.710s
```
second time:
```
Outer parsing [133 modules]                                  =  0.829s
New scanner: TEST$SYNTAX-CONFIG-CELLS                        =  0.139s
New scanner: TEST-CONFIG-CELLS                               =  0.152s
New scanner: STDOUT-STREAM$SYNTAX-CONFIG-CELLS               =  0.215s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.226s
New scanner: STDIN-STREAM$SYNTAX-CONFIG-CELLS                =  0.221s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.245s
Parse configurations [0/6 declarations]                      =  1.544s
New scanner: TEST-RULE-CELLS                                 =  0.153s
New scanner: STRING-COMMON-RULE-CELLS                        =  0.157s
New scanner: STRING-KORE-RULE-CELLS                          =  0.166s
New scanner: K-IO-RULE-CELLS                                 =  0.239s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.249s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.256s
Parse rules [0/101 rules]                                    =  0.587s
```

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -345,7 +345,8 @@ public class DefinitionParsing {
                 throw KEMException.compilerError("Module Map must be visible at the configuration declaration, in module " + module.name());
             }
             return Module(module.name(), (Set<Module>) module.imports().$bar(Set(mapModule)),
-                    (Set<Sentence>) module.localSentences().$bar(configDeclSyntax),
+                    (Set<Sentence>) module.localSentences().$bar(configDeclSyntax)
+                            .filter(s -> !(s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration))),
                     module.att());
         } else {
             Module mapModule;
@@ -355,7 +356,8 @@ public class DefinitionParsing {
                 throw KEMException.compilerError("Module Map must be visible at the configuration declaration, in module " + module.name());
             }
             return Module(module.name(), (Set<Module>) module.imports().$bar(Set(mapModule)),
-                    (Set<Sentence>) module.localSentences().$bar(configDeclRules),
+                    (Set<Sentence>) module.localSentences().$bar(configDeclRules)
+                            .filter(s -> !(s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration))),
                     module.att());
         }
 
@@ -380,48 +382,19 @@ public class DefinitionParsing {
             boolean needNewScanner = !scanner.getModule().importedModuleNames().contains(module.name());
             final Scanner realScanner = needNewScanner ? parser.getScanner(options.global) : scanner;
 
-            Set<Sentence> claimSet = stream(module.localSentences())
+            Set<Sentence> parsedSentences = stream(module.localSentences())
                     .parallel()
                     .filter(s -> s instanceof Bubble)
                     .map(b -> (Bubble) b)
-                    .filter(b -> b.sentenceType().equals(claim))
-                    .flatMap(b -> performParse(cache.getCache(), parser, realScanner, b))
-                    .map(this::upClaim)
-                .collect(Collections.toSet());
-
-            Set<Sentence> ruleSet = stream(module.localSentences())
-                    .parallel()
-                    .filter(s -> s instanceof Bubble)
-                    .map(b -> (Bubble) b)
-                    .filter(b -> b.sentenceType().equals(rule))
-                    .flatMap(b -> performParse(cache.getCache(), parser, realScanner, b))
-                    .map(this::upRule)
-                .collect(Collections.toSet());
-
-            Set<Sentence> contextSet = stream(module.localSentences())
-                    .parallel()
-                    .filter(s -> s instanceof Bubble)
-                    .map(b -> (Bubble) b)
-                    .filter(b -> b.sentenceType().equals(context))
-                    .flatMap(b -> performParse(cache.getCache(), parser, realScanner, b))
-                    .map(this::upContext)
-                .collect(Collections.toSet());
-
-            Set<Sentence> aliasSet = stream(module.localSentences())
-                    .parallel()
-                    .filter(s -> s instanceof Bubble)
-                    .map(b -> (Bubble) b)
-                    .filter(b -> b.sentenceType().equals(alias))
-                    .flatMap(b -> performParse(cache.getCache(), parser, realScanner, b))
-                    .map(this::upAlias)
-                .collect(Collections.toSet());
+                    .flatMap(b -> performParse(cache.getCache(), parser, realScanner, b).map(p -> upSentence(p, b.sentenceType())))
+                    .collect(Collections.toSet());
 
             if (needNewScanner) {
                 realScanner.close();//required for Windows.
             }
 
             return Module(module.name(), module.imports(),
-                    stream((Set<Sentence>) module.localSentences().$bar(ruleSet).$bar(claimSet).$bar(contextSet).$bar(aliasSet)).filter(b -> !(b instanceof Bubble)).collect(Collections.toSet()), module.att());
+                    stream((Set<Sentence>) module.localSentences().$bar(parsedSentences)).filter(b -> !(b instanceof Bubble)).collect(Collections.toSet()), module.att());
         }
     }
 
@@ -439,6 +412,16 @@ public class DefinitionParsing {
             }
             return upRule(res.iterator().next());
         }
+    }
+
+    private Sentence upSentence(K contens, String sentenceType) {
+        switch (sentenceType) {
+        case claim:         return upClaim(contens);
+        case rule:          return upRule(contens);
+        case context:       return upContext(contens);
+        case alias:         return upAlias(contens);
+        }
+        throw new AssertionError("Unexpected sentence type: " + sentenceType);
     }
 
     private Claim upClaim(K contents) {
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -411,19 +411,19 @@ public class DefinitionParsing {
         // scanners can be reused so find the top modules which include all other modules
         java.util.Set<Module> topMods = getTopModules(defWithCaches.modules()).stream().filter(m -> m.sentences().filter(s -> s instanceof Bubble).size() != 0).collect(Collectors.toSet());
         // prefer modules that import the main module. This way we avoid using the main syntax module which could contain problematic syntax for rule parsing
-        java.util.Set<Module> orderedBotMods = new java.util.LinkedHashSet<>();
+        java.util.Set<Module> orderedTopMods = new java.util.LinkedHashSet<>();
         for (Module m : topMods) {
             if (m.name().equals(defWithCaches.mainModule().name()) || m.importedModuleNames().contains(defWithCaches.mainModule().name()))
-                orderedBotMods.add(m);
+                orderedTopMods.add(m);
         }
-        orderedBotMods.addAll(topMods);
+        orderedTopMods.addAll(topMods);
 
         // map the module name to the scanner that it should use when parsing
         java.util.Map<String, Module> donorModule = new HashMap<>();
         for (Module m : mutable(defWithCaches.modules())) {
             if (stream(m.localSentences()).anyMatch(s -> s instanceof Bubble)) {
-                Module scannerModule = orderedBotMods.stream().filter(bm -> m.equals(bm) || bm.importedModuleNames().contains(m.name())).findFirst()
-                        .orElseThrow(() -> new AssertionError("Expected at least one bottom module to have a suitable scanner: " + m.name()));
+                Module scannerModule = orderedTopMods.stream().filter(bm -> m.equals(bm) || bm.importedModuleNames().contains(m.name())).findFirst()
+                        .orElseThrow(() -> new AssertionError("Expected at least one top module to have a suitable scanner: " + m.name()));
                 donorModule.put(m.name(), scannerModule);
             }
         }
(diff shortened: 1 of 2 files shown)


ehildenb:
I don't understand the performance numbers you post at the top. It looks like you only include numbers for `master`, and not for this branch?

Also, you should still not be taking performance numbers with `-j32`. It should be done with something like `-j4` at most, and `-j1` for best accuracy. Otherwise you'll be including a significant amount of thread/memory thrashing in the kernel in your timing numbers, which throws it off.

radumereuta:
Here are the results for running make with 4 threads on the pl-tutorial folder with the kserver active:
```
time make -j 4 -O pl-tutorial
this branch  (master)
      3m19s  (3m18s)
```
I should clarify that I'm using my local machine with 32 logical cores. This is why I usually use `-j32`
kserver actually handles the threads and spawns exactly the number of logical CPUs available. So when running `make -j32` and spawn 32 instances of kompile, the kserver will not allow more than 32 internal threads in total on all instances of kompile. This means parsing is no longer parallel, but I can always run the compilation pipeline which is not parallel.

When using only 4 instances of kompile, then kserver can parallelize the creation of scanners and the difference is not significant anymore.

The performance numbers above are for this branch on the left and for master on the right (between braces).

The differences are not that big, but I'm also doing only half of the work. To keep this diff as short as possible, I'm not including the refactoring to config parsing. There should be some improvement there as well because we're building config scanners in kprove, which is really odd.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
                 throw KEMException.compilerError("Module Map must be visible at the configuration declaration, in module " + module.name());
             }
             return Module(module.name(), (Set<Module>) module.imports().$bar(Set(mapModule)),
-                    (Set<Sentence>) module.localSentences().$bar(configDeclRules),
+                    (Set<Sentence>) module.localSentences().$bar(configDeclRules)
+                            .filter(s -> !(s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration))),
```
ehildenb, line 350: We're returning a module here that only has configuration bubbles? What are we doing with this module? What about all the other sentences?

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
+    // Find the entry modules (not included in any other module)
+    private static java.util.Set<Module> getBotModules(Set<Module> allModules) {
+        java.util.Set<Module> included = new HashSet<>();
+        for (Module m : mutable(allModules)) {
+            included.addAll(mutable(m.importedModules()));
+        }
+        java.util.Set<Module> rez = mutable(allModules);
+        rez.removeAll(included);
+        return rez;
+    }
```
ehildenb: Should we call this `getEntryModules` instead, since that's the terminology we use everywhere else?

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: The other way around. After config parsing, I'm removing all the bubbles that contain configurations.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: `Definition.entryModules` is where we store the modules visible by the user.
Other modules not visible by the user include helper modules for parsing (*-CONFIG-CELLS), the syntax split (*$SYNTAX). I guess my comment is not that intuitive then. I will change it.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
+        // prepare scanners for remaining bubbles
+        // scanners can be reused so find the bottom modules which include all other modules
+        java.util.Set<Module> botMods = getBotModules(defWithCaches.modules()).stream().filter(m -> m.sentences().filter(s -> s instanceof Bubble).size() != 0).collect(Collectors.toSet());
+        // prefer modules that import the main module. This way we avoid using the main syntax module which could contain problematic syntax for rule parsing
+        java.util.Set<Module> orderedBotMods = new java.util.LinkedHashSet<>();
+        for (Module m : botMods) {
+            if (m.name().equals(defWithCaches.mainModule().name()) || m.importedModuleNames().contains(defWithCaches.mainModule().name()))
+                orderedBotMods.add(m);
+        }
+        orderedBotMods.addAll(botMods);
```
ehildenb: Once again, should we call these `entryMods` instead, to have consistent terminology? `botMods` to me sound like modules that are at the "bottom" of the import chain, meaning they import no other modules.

ehildenb: Also, why does `botMods` not include any modules with bubbles?

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
+
+        // map the module name to the scanner that it should use when parsing
+        java.util.Map<String, Module> donorModule = new HashMap<>();
+        for (Module m : mutable(defWithCaches.modules())) {
+            if (stream(m.localSentences()).anyMatch(s -> s instanceof Bubble)) {
+                Module scannerModule = orderedBotMods.stream().filter(bm -> m.equals(bm) || bm.importedModuleNames().contains(m.name())).findFirst()
+                        .orElseThrow(() -> new AssertionError("Expected at least one bottom module to have a suitable scanner: " + m.name()));
+                donorModule.put(m.name(), scannerModule);
```
ehildenb: Do we already expect at this point that all the scanners we need are generated?

ehildenb: Or I guess here we just expect all the modules to be present, and we generate the scanners later. Makes sense.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
-
-            return Module(module.name(), module.imports(),
-                    stream((Set<Sentence>) module.localSentences().$bar(ruleSet).$bar(claimSet).$bar(contextSet).$bar(aliasSet)).filter(b -> !(b instanceof Bubble)).collect(Collections.toSet()), module.att());
         }
+        // create scanners
+        Map<Module, ParseInModule> donorScanners = new HashSet<>(donorModule.values()).parallelStream().map(x -> {
+            ParseInModule pim = RuleGrammarGenerator.getCombinedGrammar(gen.getRuleGrammar(x), isStrict, profileRules, files);
+            pim.getScanner(options.global);
+            return new Tuple2<>(x, pim);
+        }).collect(Collectors.toMap(Tuple2::_1, Tuple2::_2));
```
ehildenb: Perhaps should be called `donorParseInModules`?

ehildenb: Also, do we know at this point that every scanner we generate here is actually needed? We haven't actually checked (as far as I can tell) that each generate scanner here will be used.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
+                                    parsedBubbles.getAndIncrement();
+                                    if (kem.options.warnings2errors && !result._2().isEmpty()) {
+                                        for (KEMException err : result._2()) {
+                                            if (kem.options.includesExceptionType(err.exception.getType())) {
+                                                errors.add(KEMException.asError(err));
+                                            }
+                                        }
+                                    } else {
+                                        kem.addAllKException(result._2().stream().map(e -> e.getKException()).collect(Collectors.toList()));
+                                    }
```
ehildenb, line 467: This same logic is used here and above to deal with both `warnings2Errors` and warnings themselves. Can that be factored out to a method to make this shorter? (line 384 above)

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
ehildenb: I guess I would call them `topModules` then, not `botModules`. They are at the top of the import chains.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: Should call them `topModules` then?
I'm using visual cues when deciding what to call top and what is bottom, because that's how you write a definition. But semantically I guess the main module is a top module.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: The previous step minimizes the number of scanners needed:
`// map the module name to the scanner that it should use when parsing`
`donorParseInModules` does sound better.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -441,6 +508,16 @@ public Rule parseRule(CompiledDefinition compiledDef, String contents, Source so
...
 
+    private Sentence upSentence(K contens, String sentenceType) {
+        switch (sentenceType) {
+        case claim:         return upClaim(contens);
+        case rule:          return upRule(contens);
+        case context:       return upContext(contens);
+        case alias:         return upAlias(contens);
+        }
+        throw new AssertionError("Unexpected sentence type: " + sentenceType);
+    }
```
ehildenb: `contens` to `contents`, typo?

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -355,74 +346,150 @@ private Module resolveConfigBubbles(Definition def, Module inputModule, RuleGram
...
+                orderedBotMods.add(m);
+        }
+        orderedBotMods.addAll(topMods);
+
+        // map the module name to the scanner that it should use when parsing
+        java.util.Map<String, Module> donorModule = new HashMap<>();
+        for (Module m : mutable(defWithCaches.modules())) {
+            if (stream(m.localSentences()).anyMatch(s -> s instanceof Bubble)) {
+                Module scannerModule = orderedBotMods.stream().filter(bm -> m.equals(bm) || bm.importedModuleNames().contains(m.name())).findFirst()
+                        .orElseThrow(() -> new AssertionError("Expected at least one bottom module to have a suitable scanner: " + m.name()));
```
ehildenb: "bottom module" => "top module"?

PR #1917, opened 2021-04-15
Cleanup DefinitionParsing

radumereuta:
- Mostly done with auto refactor IntelliJ suggestions
- pushed down RuleGrammarGenerator since it's cheap to build and it makes function names more readable
- remove config bubbles from the definition at the end of resolveConfigBubbles, which allows the next item on the list
- refactor nonConfig bubbles to be parsed at the same time

These changes are just surface refactorings and do not change how the underlying algorithm works.
Done some quick tests with regression-new, kserver, and evm-semantics and the time and memory usage are the same.

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -136,7 +136,7 @@ public class DefinitionParsing {
         modules = Stream.concat(modules, Stream.of(def.getModule(entryPointModule).get()));
         modules = Stream.concat(modules, stream(def.getModule(entryPointModule).get().importedModules()));
         modules = Stream.concat(modules,
-                stream(def.entryModules()).filter(m -> !stream(m.sentences()).anyMatch(s -> s instanceof Bubble)));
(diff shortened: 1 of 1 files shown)


ehildenb:
Looks good to me, @dwightguth should probably review though. Did you double check that this doesn't lead to the memory/thread leak we saw before in the other PR?

radumereuta:
Yes. I ran the regression-new tests locally with and without the kserver and the memory and times we're the same (withing margin of error).
The memory/thread leak that I got before was because I forgot to `close()` ParseInModule objects. I'm not touching any of those things here, just some code shuffle.

dwightguth:
Can you please test this revision against the Haskell backend integration test suite? It looks like your change will change the scanner used to parse some modules during kprove and I want to make sure there are no regressions.

dwightguth:
Otherwise it looks fine. You are also free to just revert the change that removes the `Module` parameter from `parseNonConfigBubbles`, if you would rather.

radumereuta:
After getting some help from Tom, I finally managed to run the kore integration tests successfully.

dwightguth (approved):
I am approving. There does still seem to be an issue with the scanner changing during kprove, but if you can run the kore integration tests, it's clearly either minor or hard to trigger, so there's no point worrying about it unless we get a crash later.

PR #1921, opened 2021-04-20
Version 5.1

ehildenb:
Fixes: #1559

- [x] File `./package/version` has been bumped to `NEW_MAJOR.NEW_MINOR.0`, and `./package/version.sh sub` has been invoked.
- [x] `CHANGELOG.md` has been updated with new features.
- [x] `krun` help script is correctly reflecting the available options
- [x] All `*Options.java` files are up-to-date.
- [x] K active projects page is up-to-date (receiving commits in the last 3 months).
- [x] Audit of warnings in regression test-suite and tutorial.
- [x] Audit package dependencies in `package/`.

PR #1927, opened 2021-04-21
Remove `imports syntax` to clean out the code

### LINKED ISSUE #1851: [KIP] - Deprecate `imports syntax` ($SYNTAX modules)
Do we still want this feature? It creates an extra level of complexity in the kompile data structures. If this is not needed, or could be replaced with something more general, we should consider deprecating it.

Dwight: `imports syntax` directive, was created because there was a situation where you wanted to be able to import another configuration fragment for configuration composition, but not import the initializers for those cells. I'm not sure it's really used in any of our semantics currently, so I'm not hugely opposed to deprecating it somehow.

This is currently being used in `lambda++`:
https://github.com/kframework/k/blob/master/k-distribution/pl-tutorial/1_k/3_lambda++/lesson_1/lambda.k#L38
https://github.com/kframework/k/blob/master/k-distribution/pl-tutorial/1_k/3_lambda++/lesson_5/lambda.k#L14
and domains.md, DEFAULT-STRATEGY:
https://github.com/kframework/k/blob/master/k-distribution/include/kframework/builtin/domains.md

I also searched evm-semantics, c-semantics, k-exercises, wasm-semantics, firefly, 
michelson-semantics, elrond-semantics, iele-semantics and pending-documentation and couldn't find any reference for this feature.

radumereuta:
Fixes: #1851


ehildenb:
Let's wait for #1921 to get merged so that this goes into #1697

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -57,90 +24,44 @@ case class Bubble(sentenceType: String, contents: String, att: Att = Att.empty)
...
-        }
-        if (Import.isSyntax(_import.name)) {
-          result = Some(koreModules.get(_import.name).get)
-        }
-          result.get
-      } else if (koreModules.contains(_import.name))
-          koreModules.get(_import.name).get
-        else
-          throw KEMException.compilerError("Could not find module: " + _import.name, _import)
+  def toModule(allModules:Seq[FlatModule], previousModules:Set[Module]):Set[Module] = {
```
ehildenb: Please call this `toModules` instead, it's confusing me.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -57,90 +24,44 @@ case class Bubble(sentenceType: String, contents: String, att: Att = Att.empty)
...
+        val newM = new Module(
+          m.name,
+          m.imports.map(i => memoization.getOrElse(i.name,
+            toModuleRec(allModules.find(f => f.name.equals(i.name))
+              .getOrElse(throw KEMException.compilerError("Could not find module: " + i.name, i)), visitedModules :+ m))),
+          m.localSentences,
+          m.att
+        )
+        newM.checkSorts()
+        newM
```
ehildenb, line 62: Let's use explicit `return` statements, if you don't mind.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -57,90 +24,44 @@ case class Bubble(sentenceType: String, contents: String, att: Att = Att.empty)
...
-        if (Import.isSyntax(_import.name)) {
-          result = Some(koreModules.get(_import.name).get)
-        }
-          result.get
-      } else if (koreModules.contains(_import.name))
-          koreModules.get(_import.name).get
-        else
-          throw KEMException.compilerError("Could not find module: " + _import.name, _import)
+  def toModule(allModules:Seq[FlatModule], previousModules:Set[Module]):Set[Module] = {
+    val memoization:TrieMap[String, Module] = collection.concurrent.TrieMap[String, Module]()
```
ehildenb: Why `concurrent`?

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -57,90 +24,44 @@ case class Bubble(sentenceType: String, contents: String, att: Att = Att.empty)
...
+    previousModules.map(m => memoization.put(m.name, m))
+    allModules.map(m => toModuleRec(m, Seq()))
+    def toModuleRec(m:FlatModule, visitedModules: Seq[FlatModule]):Module = {
```
ehildenb: It's a little confusing to me to use `toModuleRec` here on the line before it's declared. Can we just do the `allModules.map ...` line after the declaration of `toModuleRec`?

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
```diff
@@ -124,16 +125,16 @@ public static Definition toDefinition(JsonObject data) throws IOException {
...
-        Map<String,Module> koreModules = new HashMap<>();
-        FlatModule mainFlatMod = entryModules.stream().filter(mod -> mod.name().equals(mainModuleName)).findFirst().get();
-        Module mainMod = mainFlatMod.toModule(immutable(entryModules), JavaConverters.mapAsScalaMapConverter(koreModules).asScala(), Seq());
-
-        return new Definition(mainMod, immutable(new HashSet<>(koreModules.values())), toAtt(data.getJsonObject("att")));
+        scala.collection.Set<Module> koreModules = FlatModule.toModule(immutable(flatModules), Set());
+        return Constructors.Definition(
+                koreModules.find(x -> x.name().equals(mainModuleName))
+                        .getOrElse(() -> { throw new AssertionError("Could not find main module name: " + mainModuleName); }),
+                koreModules, toAtt(data.getJsonObject("att")));
```
ehildenb: Will this assertion error print out the JSON file that was being parsed or other source location information? Is it possible to?

IIRC, assertion errors only produce stack dumps?

ehildenb: Also, isn't this somehow exactly the same code on line 107 of KILtoKORE.java? Should it be factored out to a common place? Maybe in `outer.scala`?

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -86,35 +82,16 @@ public FlatModule toFlatModule(Module m) {
...
-
-        HashMap<String, org.kframework.definition.Module> koreModules = new HashMap<>();
-
-        apply(mainModule, kilModules, koreModules);
-
-        // Set<org.kframework.definition.Module> modules = kilModules.map(i ->
-        // apply((Module) i))
-        // .collect(Collectors.toSet());
+        List<FlatModule> flatModules = kilModules.stream().map(this::toFlatModule).sorted(Comparator.comparing(FlatModule::name)).collect(Collectors.toList());
+        scala.collection.Set<org.kframework.definition.Module> koreModules = FlatModule.toModule(immutable(flatModules), Set());
```
ehildenb: This is where I'm confused by calling it `toModule` instead of `toModules`.

ehildenb:
Looks pretty good, as far as I can tell it's just ripping out the `$SYNTAX` stuff. The only thing that looks a little off is the changes to `toModule` in a `FlatModule`, but the old code there was heavily intertwined with the `$SYNTAX` stuff, so it makes sense that it changed a lot.

Just a few nits.

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
radumereuta: This class doesn't have access to the source file. It receives only the contents.
I've chosen to throw AssertionError because I don't expect this to be thrown unless something very wrong is happening.
I've updated the error message so it points out that it happened when loading from a JSON file.
The instance from KILtoKORE is only used in UnitTests.

The code does look similar, but I don't think it's worth extracting a function. Even less now that I changed the error message.
There is no special self-contained logic in there, just an object creation with a very specific context.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -8,9 +8,8 @@ import org.kframework.attributes._
 import org.kframework.utils.errorsystem.KEMException
 
 import scala.annotation.meta.param
-import scala.collection.Set
+import scala.collection.{Set, mutable}
```
ehildenb: I think we agreed stylistically at one point that we don't use this type of import, always one import per line? Not sure.
```suggestion
import scala.collection.Set
import scala.collection.mutable
```

radumereuta: Ok. Didn't know about that. IntelliJ did that by default.
I changed my settings so it doesn't happen in the future.

dwightguth (commented):
There seems to be some significant changes to the logic of constructing modules from KIL in this PR, especially with regards to the kprove code path, so I would again like to know for sure that the Haskell backend integration tests still pass as well.

radumereuta:
Just ran the haskell integration tests.
All tests pass, and the time is pretty much identical.
I also used VisualVM to visually compare the memory, threads, and cpu usage with master, and they look similar.

ehildenb (approved):

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -57,90 +24,50 @@ case class Bubble(sentenceType: String, contents: String, att: Att = Att.empty)
...
+        val newM = new Module(
+          m.name,
+          m.imports.map(i => memoization.getOrElse(i.name,
+            toModuleRec(allModules.find(f => f.name.equals(i.name))
+              .getOrElse(throw KEMException.compilerError("Could not find module: " + i.name, i)), visitedModules :+ m))),
+          m.localSentences,
+          m.att
+        )
+        newM.checkSorts()
+        newM
```
ehildenb: This reads a little gross to me, hard to understand the control flow because of the nested getOrElse* and the internal `map`.

Does it read better if you first compute the imports in a loop then construct the new module? Something like (psuedocode):

```suggestion
        val imports = Seq()
        for (String iName: m.imports) {
            if (! memoization.contains(iName) && ! allModules.contains(iName)) {
              throw KEMException.compilerError("Could not find module: " + i.name, i)
            } else {
              imports.add(memoization.getOrElseUpdate(iName, toModuleRec(allModules.find(iName), visitedModule :+ m))
            }
        }
        val newM = new Module(m.name, imports, m.localSentences, m.att)
        newM.checkSorts()
        newM
```

I dunno, it's just hard for me to unwrap the code as written.

ehildenb:
Feel free to add `automerge` if you're satisfied @radumereuta

PR #1930, opened 2021-04-21
Documentation update for symbol, klabel, simplification, and Z3 version

### LINKED ISSUE #1706: Document version requirements for dependencies
Hi,

I run into test failures coming from z3 trying to build k on os x (both release 5.0 and master). I suspect this may be due to a z3 version mismatch. Could you please document versions of all dependencies used per platform for successful (major) releases for people installing from source? Thanks!

### LINKED ISSUE #1777: Clarify the `symbol` attribute in the pending document
After reading the pending document, it is still unclear to me what does `symbol` attribute do. In the pending document, it says that symbol and klabel are used together. However, I found that in the evm semantics, some places only use klabel.

### LINKED ISSUE #1926: Add documentation for restrictions on simplification rules
Rules with the `simplification` attribute must preserve definedness; that is, if the left-hand side refers to any partial function then:

- the right-hand side must be `#Bottom` when the left-hand side is `#Bottom`, _or_
- the rule must have an `ensures` clause that is `false` when the left-hand side is `#Bottom`, _or_
- the rule must have a `requires` clause that is `false` when the left-hand side is `#Bottom`.

These conditions are in order of decreasing preference: the best option is to preserve `#Bottom` on the right-hand side, the next best option is to have an `ensures` clause, and the least-preferred option is to have a `requires` clause.

The most preferred option is to _write total functions_.

ehildenb:
Fixes: #1777 
Fixes: #1926 
Fixes: #1706


pending-documentation.md
```diff
@@ -29,27 +29,56 @@ This syntax can be used anywhere in a K definition that expects a non-terminal.
...
+    as the symbol name.
+-   In frontend K, you can refer to "symbol3" as
+    `#Baz(_,_)_MYMODULE_FooBarBaz_Int_Int` (from auto-generated klabel), and
+    the backend will see
+    `'Hash'Baz'LParUndsCommUndsRParUnds'MYMODULE'Unds'FooBarBaz'Unds'Int'Unds'Int`
+    as the symbol name.
+
+The `symbol` provided *must* be unique to this definition, unless you are
+intentionally to identify two pieces of concrete syntax as producing the same
+symbol. Uniqueness is not enforced by default for backwards compatibility.
```
dwightguth: Uniqueness actually is enforced now.

pending-documentation.md
```diff
@@ -29,27 +29,56 @@ This syntax can be used anywhere in a K definition that expects a non-terminal.
...
+
+In general, it's recommended to use `symbol` attribute whenever you use `klabel`
+unless you explicitely have a reason not to (sometimes this messes with
```
dwightguth: this is definitely not true. For example, you specifically want `klabel` but not `symbol` when you are using the klabel attribute to indicate overloaded productions.

dwightguth (changes requested):
Needs a couple changes; a few lines of this new copy are not accurate.

pending-documentation.md
```diff
@@ -69,16 +69,17 @@ Here, we have that:
 
 The `symbol` provided *must* be unique to this definition, unless you are
 intentionally to identify two pieces of concrete syntax as producing the same
-symbol. Uniqueness is not enforced by default for backwards compatibility.
+symbol (in which case you use `overload` as well). This is enforced by K.
```
dwightguth: Let's just remove the entire "unless" clause because the user will not actually be allowed to do this.

pending-documentation.md
```diff
@@ -69,16 +69,17 @@ Here, we have that:
...
 
 In general, it's recommended to use `symbol` attribute whenever you use `klabel`
-unless you explicitely have a reason not to (sometimes this messes with
-deprecated backends). It can be very helpful use the `symbol` attribute, as many
-debugging messages are printed in Kast format, which will be more readable with
-the `symbol` names you explicitely declare. In addition, if you are
-programatically manipulating definitions via the JSON Kast format, building
-terms using the user-provided pretty `symbol, klabel(...)` is easier and less
-error-prone when the auto-generation process for klabels changes.
+unless you explicitely have a reason not to (eg. you want to `overload` symbols,
```
dwightguth: **overload**, not `overload`. Overload is not a piece of K syntax.

dwightguth (changes requested):
still needs a little more tweaking

pending-documentation.md
ehildenb: TIL, fixed!

pending-documentation.md

dwightguth (approved):

ttuegel (approved):

PR #1931, opened 2021-04-21
kast tool should respect --output-file

### LINKED ISSUE #1088: `kast` ignores the parameter `--output-file`

ehildenb:
Fixes: #1088


k-distribution/tests/regression-new/issue-1088/kast-in.k
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2019 K Team. All Rights Reserved.
+
+module KAST-IN-SYNTAX
+  imports INT-SYNTAX
+
+  syntax Pgm ::= Int
+endmodule
+
+module KAST-IN
```
radumereuta, line 9: What I usually do when I name these files is:
I name the directory `issue-<number>-<short-description>`. This way I can easily distinguish between the tests when I'm looking for one. For `<short-description>` I usually use the branch name.
I always use `test.k` because when I get a failure, it's easy to just copy the file into my test setup. All the files are named test.k.
Not necessarily a convention, but it speeds up my debugging process.

radumereuta (commented):
@dwightguth is more familiar with the bison parser. I think he should have a deeper look at this.
Looks fine otherwise.

dwightguth (approved):

k-distribution/tests/regression-new/issue-1088/kast-in.k
ehildenb: I got these files by copy-pasting another similar test I've written in the past. Agree that we should have a general cleanup of the regression test-suite though.

PR #1935, opened 2021-04-22
Fix and simplify release trigger

ehildenb:
-   Hoists logic out of `package/version.sh` into the Jenkinsfile.
-   Correctly takes the version on `master` if the user has changed it, only bumps the version if it's unchanged on master (reasoning being that changes to `version` file on `master` mean "this is the version I want to set to").

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -661,8 +661,14 @@ pipeline {
             cd k-release
             git fetch --all
             git checkout -B release origin/release
-            git merge origin/master
-            ./package/version.sh bump
+            old_master="$(git merge-base origin/master origin/release)"
+            new_master="$(git rev-parse origin/master)"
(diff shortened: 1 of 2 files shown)


ehildenb:
I've tested this locally, to generate this release commit: https://github.com/kframework/k/commit/2972663f3aff8e3e699ec6e38917ececa5f1f79d which triggered this release job: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/k/detail/release/39/

I also tested locally that when there is _no_ change to the version file in a PR, it continues to bump the patch number correctly (the behavior it had before).

This should be ready to go.

dwightguth (approved):
I don't fully understand the problem this PR is trying to solve, but as long as you're satisfied that it's working correctly, it correctly fixes the build failure we had on CI, and you don't think there's any significant risk of broken releases, go ahead and merge it.

PR #1960, opened 2021-05-06
Add flag to build Nix package with LTO

andreiburdusa:
See https://github.com/kframework/llvm-backend/pull/407#issuecomment-832832707


nix/k.nix
```diff
@@ -37,7 +39,9 @@ let
       "-Dllvm.backend.skip=true"
       "-Dhaskell.backend.skip=true"
     ];
-
+    
+    cmakeBuildType = if release then "Release" else "FastBuild";
+
```
ttuegel: This won't do anything here because the frontend is built with Maven instead of CMake. Also, the frontend is all in Java, so there is no LTO anyway.

```suggestion
```

ttuegel (changes requested):

nix/k.nix
```diff
@@ -37,7 +37,7 @@ let
       "-Dllvm.backend.skip=true"
       "-Dhaskell.backend.skip=true"
     ];
-
+
```
ttuegel:
```suggestion

```

ttuegel (approved):

PR #1961, opened 2021-05-06
Refactor rule and config parsing

### LINKED ISSUE #1867: Better scanner handling

radumereuta:
Fixes: #1867
These are incremental changes to the inner parser mostly based on refactoring and doing caching before even getting close to parsing.

Waiting for feedback. Then I can continue with the actual refactoring to scanner generation.

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -432,7 +432,7 @@ public class DefinitionParsing {
         RuleGrammarGenerator gen = new RuleGrammarGenerator(compiledDef.kompiledDefinition);
         try (ParseInModule parser = RuleGrammarGenerator
                 .getCombinedGrammar(gen.getRuleGrammar(compiledDef.executionModule()), isStrict, profileRules, files)) {
-            java.util.Set<K> res = performParse(new HashMap<>(), parser, parser.getScanner(options.global),
+            java.util.Set<K> res = parseBubble(parser, new HashMap<>(),
                     new Bubble(rule, contents, Att().add("contentStartLine", Integer.class, 1)
                             .add("contentStartColumn", Integer.class, 1).add(Source.class, source)))
                     .collect(Collectors.toSet());
@@ -545,18 +545,17 @@ public class DefinitionParsing {
     }
 
     private Stream<? extends K> performParse(Map<String, ParsedSentence> cache, ParseInModule parser, Scanner scanner, Bubble b) {
+        parser.setScanner(scanner);
+        return parseBubble(parser, cache, b);
+    }
+
+    private Stream<? extends K> parseBubble(ParseInModule pim, Map<String, ParsedSentence> cache, Bubble b) {
         int startLine = b.att().get("contentStartLine", Integer.class);
         int startColumn = b.att().get("contentStartColumn", Integer.class);
         Source source = b.att().get(Source.class);
-        Tuple2<Either<java.util.Set<KEMException>, K>, java.util.Set<KEMException>> result;
-        if (cache.containsKey(b.contents())) {
-            ParsedSentence parse = cache.get(b.contents());
-            cachedBubbles.getAndIncrement();
-            registerWarnings(parse.getWarnings());
-            Att att = parse.getParse().att().addAll(b.att().remove("contentStartLine").remove("contentStartColumn").remove(Source.class).remove(Location.class));
-            return Stream.of(new AddAtt(a -> att).apply(parse.getParse()));
-        }
-        result = parser.parseString(b.contents(), START_SYMBOL, scanner, source, startLine, startColumn, true, b.att().contains(Att.ANYWHERE()) || b.att().contains(Att.SIMPLIFICATION()) || ExpandMacros.isMacro(b));
+        boolean isAnywhere = b.att().contains(Att.ANYWHERE()) || b.att().contains(Att.SIMPLIFICATION()) || ExpandMacros.isMacro(b);
+        Tuple2<Either<java.util.Set<KEMException>, K>, java.util.Set<KEMException>> result =
+                pim.parseString(b.contents(), START_SYMBOL, pim.getScanner(), source, startLine, startColumn, true, isAnywhere);
         parsedBubbles.getAndIncrement();
         registerWarnings(result._2());
         if (result._1().isRight()) {
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -14,6 +14,7 @@ import org.kframework.compile.ExpandMacros;
 import org.kframework.compile.GenerateSentencesFromConfigDecl;
 import org.kframework.definition.Bubble;
 import org.kframework.definition.Claim;
+import org.kframework.definition.Configuration;
 import org.kframework.definition.Context;
 import org.kframework.definition.ContextAlias;
 import org.kframework.definition.Definition;
@@ -296,18 +297,7 @@ public class DefinitionParsing {
                     .map(b -> (Bubble) b)
                     .filter(b -> b.sentenceType().equals(configuration))
                     .flatMap(b -> performParse(cache.getCache(), parser, parser.getScanner(options.global), b))
-                    .map(contents -> {
-                        KApply configContents = (KApply) contents;
-                        List<K> items = configContents.klist().items();
-                        switch (configContents.klabel().name()) {
-                        case "#ruleNoConditions":
-                            return Configuration(items.get(0), BooleanUtils.TRUE, configContents.att());
-                        case "#ruleEnsures":
-                            return Configuration(items.get(0), items.get(1), configContents.att());
-                        default:
-                            throw KEMException.compilerError("Illegal configuration with requires clause detected.", configContents);
-                        }
-                    })
+                    .map(this::upConfiguration)
                     .flatMap(
                             configDecl -> stream(GenerateSentencesFromConfigDecl.gen(configDecl.body(), configDecl.ensures(), configDecl.att(), parser.getExtensionModule(), kore)))
                     .collect(Collections.toSet());
@@ -391,6 +381,7 @@ public class DefinitionParsing {
         case rule:          return upRule(contents);
         case context:       return upContext(contents);
         case alias:         return upAlias(contents);
+        case configuration: return upConfiguration(contents);
         }
         throw new AssertionError("Unexpected sentence type: " + sentenceType);
     }
@@ -452,6 +443,19 @@ public class DefinitionParsing {
             return ContextAlias(items.get(0), items.get(1), ruleContents.att());
         default:
             throw KEMException.criticalError("Illegal context alias with ensures clause detected.", contents);
+        }
+    }
+
+    private Configuration upConfiguration(K contents) {
+        KApply configContents = (KApply) contents;
+        List<K> items = configContents.klist().items();
+        switch (configContents.klabel().name()) {
+        case "#ruleNoConditions":
+            return Configuration(items.get(0), BooleanUtils.TRUE, configContents.att());
+        case "#ruleEnsures":
+            return Configuration(items.get(0), items.get(1), configContents.att());
+        default:
+            throw KEMException.compilerError("Illegal configuration with requires clause detected.", configContents);
         }
     }
 
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -271,34 +271,7 @@ public class DefinitionParsing {
     }
 
     private Definition resolveConfigBubbles(Definition def) {
-        Definition defWithCaches = resolveCachedBubbles(def, false);
-        RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
-
-        Definition defWithParsedConfigs = DefinitionTransformer.from(m -> {
-            if (stream(m.localSentences()).noneMatch(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration)))
-                return m;
-            Module ruleParserModule = gen.getConfigGrammar(m);
-
-            ParseCache cache = loadCache(ruleParserModule);
-            try (ParseInModule parser = RuleGrammarGenerator.getCombinedGrammar(cache.getModule(), isStrict, profileRules, files)) {
-                if (stream(m.localSentences()).filter(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration))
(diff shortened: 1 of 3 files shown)


kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -269,69 +271,76 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+            Set<Sentence> importedConfigurationSortsSubsortedToCell = stream(m.productions())
+                  .filter(p -> p.att().contains("cell"))
+                  .map(p -> Production(Seq(), Sorts.Cell(), Seq(NonTerminal(p.sort())))).collect(toSet());
+
+            Module module = Module(m.name(), m.imports(),
+                  (Set<Sentence>) m.localSentences().$bar(importedConfigurationSortsSubsortedToCell),
+                  m.att());
+
+            Set<Sentence> configDeclProductions;
+            ParseCache cache = loadCache(gen.getConfigGrammar(module));
```
dwightguth, line 289: This and the following line are expensive steps, and you're doing them twice. It would be better if we could avoid redoing this work.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -369,12 +378,61 @@ private Module resolveNonConfigBubbles(Module module, Scanner scanner, RuleGramm
...
+     * @param def    The Definition with Bubbles.
+     * @param isRule true if it should target non config Bubbles, false if it should parse only config bubbles
+     * @return A new Definition object with Bubbles replaced by the appropriate Sentence type.
+     */
+    private Definition resolveCachedBubbles(Definition def, boolean isRule) {
+        RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
+        return DefinitionTransformer.from(m -> {
+            if (stream(m.localSentences()).noneMatch(s -> s instanceof Bubble && (isRule || ((Bubble) s).sentenceType().equals(configuration))))
+                return m;
+            ParseCache cache = isRule ? loadCache(gen.getRuleGrammar(m)) : loadCache(gen.getConfigGrammar(m));
```
dwightguth, line 324: Here's a third place we're calling loadCache

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -473,36 +545,19 @@ private boolean equalsSyntax(Module _this, Module that) {
     }
 
     private Stream<? extends K> performParse(Map<String, ParsedSentence> cache, ParseInModule parser, Scanner scanner, Bubble b) {
+        parser.setScanner(scanner);
```
dwightguth: I don't understand why there are now two functions here.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -186,6 +186,10 @@ public Scanner getScanner() {
         return scanner;
     }
 
+    public void setScanner(Scanner scanner) {
```
dwightguth, line 189: This seems suspicious to me.

dwightguth (changes requested):
It definitely looks better than your last stab at refactoring, but I'm concerned about certain relatively expensive steps being executed multiple times now, as well as some oddness surrounding parseBubble vs performParse

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: Removed and replaced with `resolveBubbles` which computes the minimum amount of scanners.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
radumereuta: I needed this because I'm now precomputing the scanners before I have all the parsers.
When I get to create the parser, I just lookup for the appropriate scanner and do parsing.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: There are now only two places, once when looking for cached bubbles, and another when doing parsing.
I've done some profiling and it seems that the most expensive step is actually `ParseInModule.getExtensionModule()`
`loadCache` didn't even register. I could avoid calling it twice by combining `resolveBubbles` and `resolveCachedBubbles` but it doesn't seem to be a problem, so at the moment I prefer to have them separate for clarity.

radumereuta:
I did some preliminary testing and here are some numbers:
master: make -j 32 regression-new -> 2m36s
current: make -j 32 regression-new -> 2m32s
With the kserver on.

I ran `kompile -v --backend haskell` over a simple test definition (no kserver), and here are the outputs:
```
$ cat test.k
module TEST
  imports TEST-SYNTAX
  imports INT

  rule 0 => 1
endmodule
```
this branch first time:
```
Outer parsing [57 modules]                                   =  0.751s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.189s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.193s
New scanner: TEST-CONFIG-CELLS                               =  0.123s
Parse configurations [3/3 declarations]                      =  0.636s
New scanner: TEST-RULE-CELLS                                 =  0.120s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.194s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.186s
Parse rules [101/101 rules]                                  =  0.750s
Validate definition                                          =  0.058s
Apply compile pipeline                                       =  0.245s
Kompile to kore                                              =  0.002s
Save to disk                                                 =  0.024s
Backend                                                      =  0.287s
Cleanup                                                      =  0.000s
Total                                                        =  2.765s
```
this branch second time:
```
Outer parsing [57 modules]                                   =  0.725s
Parse configurations [0/3 declarations]                      =  0.270s
Parse rules [0/101 rules]                                    =  0.071s
Validate definition                                          =  0.073s
Apply compile pipeline                                       =  0.290s
Kompile to kore                                              =  0.001s
Save to disk                                                 =  0.022s
Backend                                                      =  0.338s
Cleanup                                                      =  0.000s
Total                                                        =  1.801s
```
master branch first time:
```
Outer parsing [57 modules]                                   =  0.736s
New scanner: TEST-CONFIG-CELLS                               =  0.126s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.173s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.181s
Parse configurations [3/3 declarations]                      =  0.802s
New scanner: TEST-RULE-CELLS                                 =  0.123s
New scanner: STRING-COMMON-RULE-CELLS                        =  0.177s
New scanner: STRING-KORE-RULE-CELLS                          =  0.182s
New scanner: K-IO-RULE-CELLS                                 =  0.248s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.252s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.258s
Parse rules [101/101 rules]                                  =  0.699s
Validate definition                                          =  0.063s
Apply compile pipeline                                       =  0.255s
Kompile to kore                                              =  0.001s
Save to disk                                                 =  0.032s
Backend                                                      =  0.291s
Total                                                        =  2.890s
```
master branch second time:
```
Outer parsing [57 modules]                                   =  0.714s
New scanner: TEST-CONFIG-CELLS                               =  0.127s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.176s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.185s
Parse configurations [0/3 declarations]                      =  0.738s
New scanner: TEST-RULE-CELLS                                 =  0.119s
New scanner: STRING-COMMON-RULE-CELLS                        =  0.137s
New scanner: STRING-KORE-RULE-CELLS                          =  0.136s
New scanner: K-IO-RULE-CELLS                                 =  0.201s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.200s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.216s
Parse rules [0/101 rules]                                    =  0.473s
Validate definition                                          =  0.081s
Apply compile pipeline                                       =  0.275s
Kompile to kore                                              =  0.001s
Save to disk                                                 =  0.029s
Backend                                                      =  0.315s
Total                                                        =  2.637s
```

dwightguth:
I fail to see how the changes you pushed address my concerns with the original pull request. We asked for incremental changes, and adding more non incremental changes onto something that mostly looks good isn't an adequate way to address the concerns that I had with the pull request as it originally existed.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -269,69 +271,76 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+        Definition defWithCaches = resolveCachedBubbles(def, false);
+        RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
+
+        Definition defWithParsedConfigs = DefinitionTransformer.from(m -> {
+            if (stream(m.localSentences()).noneMatch(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration)))
+                return m;
+            Module ruleParserModule = gen.getConfigGrammar(m);
+
+            ParseCache cache = loadCache(ruleParserModule);
+            try (ParseInModule parser = RuleGrammarGenerator.getCombinedGrammar(cache.getModule(), isStrict, profileRules, files)) {
```
dwightguth: Here is the first ParseInModule being created. It gets `initialize` and `getScanner` called on it so that the object is fully computed prior to the split into multiple threads, which is important, otherwise the work gets done once on each thread.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -269,69 +271,76 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+                  .filter(p -> p.att().contains("cell"))
+                  .map(p -> Production(Seq(), Sorts.Cell(), Seq(NonTerminal(p.sort())))).collect(toSet());
+
+            Module module = Module(m.name(), m.imports(),
+                  (Set<Sentence>) m.localSentences().$bar(importedConfigurationSortsSubsortedToCell),
+                  m.att());
+
+            Set<Sentence> configDeclProductions;
+            ParseCache cache = loadCache(gen.getConfigGrammar(module));
+            ParseInModule parser = RuleGrammarGenerator.getCombinedGrammar(cache.getModule(), isStrict, profileRules, files);
```
dwightguth: Here a ParseInModule is being created a second time, when before this change, the same ParseInModule was being reused in both places.

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -269,69 +271,76 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+                configDeclProductions = stream(module.localSentences())
+                      .parallel()
+                      .filter(s -> s instanceof Configuration)
+                      .map(b -> (Configuration) b)
+                      .flatMap(configDecl -> stream(GenerateSentencesFromConfigDecl.gen(configDecl.body(), configDecl.ensures(), configDecl.att(), parser.getExtensionModule(), kore)))
```
dwightguth: Here getExtensionModule is called in parallel from each thread, which means each thread does the work of computing the extension module, even though we could in theory be using the version computed on line 286.

dwightguth (changes requested):

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
radumereuta: This is a code readability choice. The alternatives would be to either collapse two functions together (which make them way too big), or pass parameters to those functions, which would make the parameter list bigger than I would like. I took this decision because generating a `ParseInModule` object is cheap.

Issue #1838, opened 2021-03-02
[Bug] Kompile pipeline duplicates modules


radumereuta:
Git revision: 061c0ba (latest master)

Replicate with just command line:
```
radu@radu:~/work/k/k-distribution/tests/regression-new/kprove-haskell$ kompile verif.k --backend haskell
radu@radu:~/work/k/k-distribution/tests/regression-new/kprove-haskell$ kprove sum-spec.k --save-proof-definition-to save
#Top
radu@radu:~/work/k/k-distribution/tests/regression-new/kprove-haskell$ kprove test-spec.k -d save
[Error] Internal: Uncaught exception thrown of type IllegalStateException.
Please rerun your program with the --debug flag to generate a stack trace, and
file a bug report at https://github.com/kframework/k/issues
(IllegalStateException: Duplicate key UNSIGNED-INT-SYNTAX$SYNTAX (attempted
merging values module UNSIGNED-INT-SYNTAX$SYNTAX
[org.kframework.attributes.Location(Location(878,1,881,9)),
...
```
The other way I replicate it is by checking the `Definition` object directly after applying the compilation steps:
```
Kompile.java 146: org.kframework.Collections.mutable(kompiledDefinition.modules()).stream().collect(java.util.stream.Collectors.toMap(org.kframework.definition.Module::name, m -> m));
```
I tested for both java and haskell backends and I get the same problem.
The problem is pretty old. I tested the build from one year ago and it's still present.
Most likely there are multiple problematic steps since I tried removing some and I get various failure states.

dwightguth:
I can confirm this bug affects the latest master. Also, I found a good commit in the history that predates this regression: 325fb9d75780716ede721683f2cd36e1f818d250

I am starting my bisect now and will provide the first bad commit as soon as I have it.

radumereuta:
I found the first commit that introduces the duplicates for the java backend: 2178532dcd45dada33efcbc67e2146c6638bf91a
Leaving it here for reference.
The easiest way to track the issue is to add an assert in outer.scala in the Definition constructor:
```
case class Definition(
                       mainModule: Module,
                       entryModules: Set[Module],
                       att: Att)
  extends DefinitionToString with OuterKORE {
  assert(modules.map(x => x.name).size == modules.size) // add this line
```

radumereuta:
I tried to enforce a more strict assert in the Module constructor:
```
assert(importedModulesIds.size == importedModuleNames.size, "Module imports duplicate modules.")
[...]

PR #1965, opened 2021-05-12
Add MAP import to all modules containing configs

radumereuta:
This simple test fails on master.
The `MAP` module should be added to every module where a configuration is present.

### DIFF
--- a/kore/src/main/scala/org/kframework/definition/outer.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer.scala
@@ -349,7 +349,7 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
   override lazy val hashCode: Int = name.hashCode
 
   override def equals(that: Any) = that match {
-    case m: Module => m.name == name && m.sentences == sentences
+    case m: Module => m.name == name && m.sentences == sentences && m.imports.map(x => x.name) == imports.map(x => x.name)
   }
 
   def flattened()   : FlatModule                = new FlatModule(name, imports.map(m => Import(m.name, Att.empty)), localSentences, att)
--- a/kore/src/main/scala/org/kframework/definition/outer.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer.scala
@@ -348,10 +348,6 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
 
   override lazy val hashCode: Int = name.hashCode
 
-  override def equals(that: Any) = that match {
-    case m: Module => m.name == name && m.sentences == sentences && m.imports.map(x => x.name) == imports.map(x => x.name)
-  }
-
   def flattened()   : FlatModule                = new FlatModule(name, imports.map(m => Import(m.name, Att.empty)), localSentences, att)
   def flatModules() : (String, Set[FlatModule]) = (name, Set(flattened) ++ imports.map(m => m.flatModules._2).flatten)
 }
--- a/k-distribution/tests/regression-new/checks/mapImportInConfigModules.k
+++ b/k-distribution/tests/regression-new/checks/mapImportInConfigModules.k
@@ -1,0 +1,19 @@
+module MAPIMPORTINCONFIGMODULES-SYNTAX
+  imports INT-SYNTAX
+endmodule
+
+module CONFIG
+  imports INT
+  syntax E ::= "balance"
+  configuration <k> $PGM:Int </k> <balance> .Map </balance>
+
+endmodule
+
+module MAPIMPORTINCONFIGMODULES
+  imports MAPIMPORTINCONFIGMODULES-SYNTAX
+  imports INT
+  imports CONFIG
+
+  rule <balance> .Map => .Map </balance>
+  rule <k> 0 => 1 </k>
+endmodule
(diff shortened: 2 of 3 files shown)


dwightguth (changes requested):
Can you explain to me why this change is needed in order to fix Algorand on your PR when it works fine on master already? If you can't explain why it's needed, I really can't justify this.

ehildenb:
@dwightguth this seems like a bugfix to me (or at least, the intent is one, based on my conversations with @radumereuta on Slack).

He discovered (while doing other work), that if the `mainModule` does not `import MAP` but contains a `configuration` (either locally or in one of its imports), then `import MAP` will be added to the `mainModule`.

This works if your `configuration` declaration is in the `mainModule`, but if it's in a module that is imported by it which _does not_ already `import MAP`, then `MAP` will be missing from the module with the `configuration` declaration in it, where it is needed.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -349,7 +349,7 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
   override lazy val hashCode: Int = name.hashCode
 
   override def equals(that: Any) = that match {
-    case m: Module => m.name == name && m.sentences == sentences
+    case m: Module => m.name == name && m.sentences == sentences && m.imports.map(x => x.name) == imports.map(x => x.name)
```
ehildenb: What is going on here? This seems like it deserves a test of its own?

radumereuta:
@dwightguth ignore Algorand for now. I encounter errors like this sometimes, and I prefer to fix that if it's isolated instead of trying to work around it to push another PR. (in this case the scanner changes)
Instead, look at the test I added. On master that fails with a parse error in the configuration because it can't parse token `Map`.

@ehildenb I've removed the equals method completely. Now the default behavior checks for:
`Module(name, localSentences, imports, att)`
Previously it would check only for `name` and the cumulated `sentences` which would ignore completely the import structure. For a single pass it may be fine, but when you apply multiple transformers and change connections between modules can lead to inconsistent states.
I found about this because k-exercises/tutorial/1_k/3_lambda++/lesson_1/exercises/from-call-CC-to-callcc failed. So we already have a test. Not sure how to test it even further, because what I did was to remove an optimization. I also tested regression-new and Algorand and the times are exactly the same.

This is related partially to #1838. It mitigates it a bit, but not completely. I have a bit more confidence that at least it no longer duplicates modules semantically.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -348,10 +348,6 @@ case class Module(val name: String, val imports: Set[Module], localSentences: Se
 
   override lazy val hashCode: Int = name.hashCode
 
-  override def equals(that: Any) = that match {
```
dwightguth, line 351: If we're going to be using the default equals for case classes, let's use the default hashCode as well.

dwightguth (commented):
This seems an acceptable change if I view it as purely unrelated to the scanner changes, and if it can resolve some of the issues with duplicated modules later in the pipeline, so much the better. We should have a discussion about what exactly we want the behavior of K to be here, though, before we merge this. We automatically imported MAP into the main module prior because the definition doesn't work without it. Forcing the other modules to explicitly import MAP if they use map productions makes the import system more consistent, but the configuration declaration semantics inconsistent. Whereas, allowing the other modules to automatically import MAP makes the semantics of configuration declarations more consistent, but the semantics of imports less consistent. I don't have a strong opinion on which is better, but it seems there are pros and cons to both, so we should probably discuss this as a design change rather than treating it purely like a bugfix.

kore/src/main/scala/org/kframework/definition/outer.scala
radumereuta: I thought about that. But I think this is faster and still safe.
It's only used to quickly determine if two modules are different. And in our case, that happens most of the time since a Definition has unique Module names.

radumereuta:
Master behavior:
This parses fine:
```
module CONFIG1
configuration <k> .K </k>
endmodule

module CONFIG2
imports CONFIG1
configuration <x> <k/> <y> .Map </y> </x>
endmodule

module TEST
imports CONFIG2
endmodule
```
This gives an error message:  `Parse error: unexpected token 'Map'.`
```
module CONFIG
configuration <x> <k> .K </k> <y> .Map </y> </x>
endmodule

module TEST
imports CONFIG
endmodule
```
So I view this branch as actually a bug fix. It makes the behavior more consistent.
Now both of the examples parse w/o errors.

dwightguth:
I mean, it makes it so that an anomaly is more consistently anomalous, if that's what you mean. That's not the same thing as just being unqualifiedly more consistent. As I said, I want to see us discuss this as a group, even if it's brief because everyone has consensus.

ehildenb (approved):

radumereuta:
Timings:
make build -j 1
EVM master: 1m07.004s, this branch 1m07.344s
make build -j 10
Algorand master: 2m16s, 2m19s. This branch 2m42s, 2m25s, 2m26s.
These numbers are consistent with what I usually get and in the same range.

I tested EVM with and w/o hashCode:
hashCode deleted: 3m47s, 1m28s, 1m40s
with hashCode again: 1m06s
It seems it's more consistent with precomputed hashCode.

dwightguth (approved):

PR #1978, opened 2021-05-19
Search in llvm backend

dwightguth:
This implements the frontend functionality surrounding search in the llvm backend. We still don't support search modes other than --search-final, or search patterns. I tried to update imp++ and lambda++ in the pl-tutorial, but ran into problems getting the makefiles for the exercises to work, and decided it wasn't worth the trouble. Thus, we use a modified version of imp++ lesson 8 to integration test this functionality. In theory, it should be possible to update lambda++ and imp++ assuming the issues with the build system for the exercises can be resolved.


k-distribution/src/main/scripts/bin/krun
```diff
@@ -521,10 +521,23 @@ if [ "$backendName" = "llvm" ]; then
...
+    if [ -n "$pattern" ]; then
+      error "Backend llvm does not support search patterns yet."
+    fi
+    if [[ $searchType != "FINAL" ]]; then
+      error "Backend llvm does not support --search-all, --search-one-or-more-steps, or --search-one-step yet"
+    fi
+    set +e
+    execute $cmdprefix "$kompiledDir/search" "$expanded_input_file" "${depth:--1}" "$kore_output" $interpreter_flags
+    result=$?
+    set -e
```
ehildenb, line 534:
```suggestion
    result=0
    execute $cmdprefix "$kompiledDir/search" "$expanded_input_file" "${depth:--1}" "$kore_output" $interpreter_flags \
      || result=$?
```

Or just set `result=0` at the top of the file, since it seems to only be used here.

Can you also change the remaining cases where this pattern is used? We sohuld just leave `-e` set, and use `|| result=$?` if we want to record a bad exit-code _and_ prevent failure.

k-distribution/tests/regression-new/imp++-llvm/imp.md
```diff
@@ -0,0 +1,677 @@
...
+IMP++
+=====
+
+Author: Grigore Roșu (grosu@illinois.edu)  
+Organization: University of Illinois at Urbana-Champaign
+
+### Abstract
+This is the **K** semantic definition of the IMP++ language.
+IMP++ extends the IMP language with the features listed below. We
+strongly recommend you to first familiarize yourself with the IMP
```
ehildenb, line 14: What are the differences between the tutorial version of this and this one?

It seems to me the only difference is `type="Set"` on the `<thread>` cell. Can we fix this in the tutorial instead to just use the LLVM backend? Maybe by adding a unique-thread-id cell to the `<thread>` sub-configuration, and having every new thread spawned used `!X:Int` as that id?

I know you asked about this on Slack, but I didn't realize it was such a small change between the tutorial and this version.

dwightguth: It's not the definitions that are the problem. The changes to the definitions are manageable. The problem is the changes to the Makefiles

ehildenb (approved):

PR #1983, opened 2021-05-21
Exclude simplification rules in concrete backend

### LINKED ISSUE #1940: [Bug] kompile - suggests to add existing attribute (simplification)
First fill in the title (template provided).
K Version
---------

```
$ kompile --version
K version:    5.1.0
Build date:   Fri Apr 23 20:39:29 EEST 2021
```

Description
-----------

Running `kompile` without `--backend haskell` produces errors for simplification rules. The error message suggests adding the `simplification` attribute.

Input Files
-----------

`tmp.k`

```
module TMP

syntax Int ::= f(Int)  [function, functional]
syntax Int ::= g(Int)  [function, functional]

rule g(f(X)) => X [simplification]

endmodule
```

Reproduction Steps
------------------

Kompiling produces an error suggesting to add the `simplification` attribute:

```
$ kompile tmp.k
[Error] Compiler: Illegal function symbol f(_)_TMP_Int_Int on LHS of rule.
Consider adding `simplification` attribute to the rule if this is intended.
        Source(/mnt/data/runtime-verification/verified-smart-contracts/multisig/protocol-correctness/proof/functions/tmp/tmp.k)
        Location(6,8,6,12)
```

Expected Behavior
-----------------

Either `tmp.k` should just kompile, or the error message should say something like "Simplification rules not supported by default. Kompile for a backend that supports them."

radumereuta:
Fixes: #1940
Exclude sentences depending on backend type before parsing starts

I'm not sure if there was a step to exclude simplification rules for concrete backends. If I missed it please point it out.
The best place is before parsing anyway.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
@@ -10,7 +10,6 @@ import org.kframework.builtin.Sorts;
 import org.kframework.compile.*;
 import org.kframework.definition.*;
 import org.kframework.definition.Module;
-import org.kframework.kompile.CompiledDefinition;
 import org.kframework.kompile.Kompile;
 import org.kframework.kompile.KompileOptions;
 import org.kframework.kore.ADT;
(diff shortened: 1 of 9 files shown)


dwightguth (changes requested):
This is not an appropriate solution. The llvm backend already filters out simplification rules inside the backend, but we still ought to parse them and compile them. You just need to modify the check that reports the error in question so that it correctly handles the simplification attribute.

radumereuta:
> This is not an appropriate solution. The llvm backend already filters out simplification rules inside the backend, but we still ought to parse them and compile them. You just need to modify the check that reports the error in question so that it correctly handles the simplification attribute.

Why do we need to parse them if they are excluded anyway? It seems to me like missing an opportunity to make kompile faster.

dwightguth:
I mean, in order to report parser and compilation errors?

dwightguth:
Besides, in general, the more similar the kore is between the llvm and haskell backend, the better.

radumereuta:
But you don't care about that rule on that backend. It's the same idea when we exclude entire modules based on the same criteria.

dwightguth:
I mean, that was done out of necessity. This isn't necessary.

radumereuta:
I'm not entirely convinced. Let's discuss this in the next front-end meeting.

ehildenb:
I agree with @dwightguth here, I do think we should keep the backends seeing as close to the same Kore as possible.

radumereuta:
@dwightguth @ehildenb I've changed it to how you wanted.
I'm leaving the simplification rule in, and the backend should interpret it as it sees fit.
It might be easier to review the commits.

ehildenb:
It appears that now the change just makes it so that `simplification` rules are also sent to the LLVM Backend? I support that on principle (I think ideally both backends should see the exact same kore), but just want to make sure that's what's intended here @radumereuta

radumereuta:
That was not my original intent, but you and Dwight seem to prefer it.
With this implementation I expect the simplification rule to go through the entire kompile pipeline, and the LLVM backend will exclude it.
>Besides, in general, the more similar the kore is between the llvm and haskell backend, the better.

dwightguth (approved):

PR #1984, opened 2021-05-21
Normalize paths instead of canonicalize

### LINKED ISSUE #1906: [Bug] kompile - Inconsistent symlink handling in require
First fill in the title (template provided).
Examples:

K Version
---------

```
$ kompile --version
RV-K version 1.0-SNAPSHOT
Git revision: 0cfac90
Git branch: master
Build date: Wed Mar 17 01:46:12 EET 2021
```

Description
-----------

If the main file being kompiled is a symlink, its `require`d files are resolved relative to the directory containing the symlink. However, the `require`d files for everything else are resolved relative to the directory containing the actual file.

Input Files
-----------

In a directory, let's call it `original`:

x.k:
```
require "y.k"
module X
endmodule
```

y.k:
```
require "z.k"
module Y
endmodule
```

z.k:
```
module Z
endmodule
```

Reproduction Steps
------------------

In the parent directory containing the `original` dorectory mentioned above:

```
mkdir symlink
cd symlink
ln -s ../original/x.k x.k
kompile x.k
ln -s ../original/y.k y.k
kompile y.k
```

For the first kompile, you should get an error: ```[Error] Critical: Could not find file: y.k```. I think that this is the correct behaviour. The second kompile, however, works.

Expected Behavior
-----------------

I think that all `require`d files should be resolved relative to the directory containing the symlink. If that's not reasonable, the `kompile` behaviour should at least be consistent (error in both cases above or no error in either case).

radumereuta:
Fixes: #1906
This should keep symlinks but still eliminate '..' and '.'
Not sure how to add a test for this because the error message contains absolute paths that are different on each machine.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/ParserUtils.java
@@ -141,10 +141,7 @@ public class ParserUtils {
                         .filter(file -> file.exists()).findFirst();
 
                 if (definitionFile.isPresent()) {
-                    File canonical = definitionFile.get().getAbsoluteFile();
-                    try {
-                        canonical = canonical.getCanonicalFile();
-                    } catch (IOException e) {}
+                    File canonical = definitionFile.get().toPath().toAbsolutePath().normalize().toFile();
                     if (!requiredFiles.contains(canonical)) {
                         requiredFiles.add(canonical);
                         results.addAll(slurp(loadDefinitionText(canonical),
@@ -152,10 +149,10 @@ public class ParserUtils {
                                 canonical.getParentFile(),
                                 lookupDirectories, requiredFiles));
                     }
-                }
-                else
+                } else {
                     throw KEMException.criticalError("Could not find file: " +
                             finalDefinitionFile + "\nLookup directories:" + allLookupDirectories, di);
+                }
             }
         }
         return results;
@@ -257,7 +254,7 @@ public class ParserUtils {
             boolean preprocess,
             boolean leftAssoc) {
         return loadDefinition(mainModuleName, syntaxModuleName, definitionText,
-                Source.apply(source.getAbsolutePath()),
+                Source.apply(source.toPath().toAbsolutePath().normalize().toString()),
                 currentDirectory, lookupDirectories, autoImportDomains, kore, preprocess, leftAssoc);
     }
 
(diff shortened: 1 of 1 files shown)


kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -141,21 +141,18 @@ else if (di instanceof Require) {
                         .filter(file -> file.exists()).findFirst();
 
                 if (definitionFile.isPresent()) {
-                    File canonical = definitionFile.get().getAbsoluteFile();
-                    try {
-                        canonical = canonical.getCanonicalFile();
-                    } catch (IOException e) {}
+                    File canonical = definitionFile.get().toPath().toAbsolutePath().normalize().toFile();
```
virgil-serbanuta: This seems reasonable to me, but I'd like to point this out, just in case other people think differently:
"Eliminating ".." and a preceding name from a path may result in the path that locates a different file than the original path."
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#normalize()

virgil-serbanuta (approved):

kernel/src/main/java/org/kframework/parser/ParserUtils.java
radumereuta: I did notice that. What do other people think?
I chose to eliminate `..` because it makes the reported error messages more readable.
I saw people getting confused by kompile because they were calling different files in different locations. And it happened to me too. Having clean source paths can help with that.

dwightguth:
Uh, Virgil's concern seems quite serious to me. If normalized, but not canonicalized, paths, are not guaranteed to refer to the same file as the original path, then it's probably best to keep code the way it is, which canonicalizes it if it can and otherwise leaves it as a non-normalized path. I guess we could always just keep everything as a non-normalized absolute path... But the current approach seems better to me in the vast majority of cases.

radumereuta:
I think I found the option that seems most intuitive. I replaced `normalize()` with `toRealPath`
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#toRealPath(java.nio.file.LinkOption...)
>When not resolving symbolic links and the preceding name is a symbolic link then the names are only removed if it guaranteed that the resulting path will locate the same file as this path.

@dwightguth please re-review

dwightguth (approved):
looks good, thanks

PR #1987, opened 2021-05-25
Don't wrap exception messages to 80 chars in integration tests

andreiburdusa:
Needed for #1884


radumereuta:
Do you know about `make update-results`?
It should just override the output files with the new output. Just make sure to inspect the files before committing to make sure you are getting the expected results.

andreiburdusa:
> Do you know about `make update-results`?
> It should just override the output files with the new output. Just make sure to inspect the files before committing to make sure you are getting the expected results.
> 

I didn't know about it, thanks

ttuegel (approved):

PR #1993, opened 2021-05-26
Lesson 7: Side conditions and rule priority

dwightguth:
Includes a couple minor fixes to previous lessons


k-distribution/k-tutorial/1_basic/07_side_conditions/README.md
```diff
@@ -1 +1,171 @@
...
+```k
+module LESSON-07-B
+  imports BOOL
+  imports INT
+
+  syntax Bool ::= isPositive(Int) [function]
+
+  rule isPositive(I) => true
+    requires I >Int 0	
+  rule isPositive(_) => false [owise]
```
ehildenb:
```suggestion
  rule isPositive(I) => true  requires I >Int 0	
  rule isPositive(_) => false [owise] 
```

I like to write function definitions with `[owise]` so that it lines up with the requires clauses, somehow it reads better in my mind.

ehildenb: Your choice though.

k-distribution/k-tutorial/1_basic/07_side_conditions/README.md
```diff
@@ -1 +1,171 @@
...
+  rule isPositive(I) => true
+    requires I >Int 0	
+  rule isPositive(_) => false [owise] 
+endmodule
+```
+
+Note that we have introduced a new piece of syntax here: `_`. This is actually
+just a variable. However, as a special case, when a variable is named `_`, it
+does not bind a value that can be used on the right hand side of the rule, or
+in a side condition.
```
ehildenb:
```suggestion
Note that we have introduced a new piece of syntax here: `_`. This is actually
just a variable. However, as a special case, when a variable is named `_`, it
does not bind a value that can be used on the right hand side of the rule, or
in a side condition. Effectively, `_` is a placeholder which means "don't care".
```

k-distribution/k-tutorial/1_basic/07_side_conditions/README.md
```diff
@@ -1 +1,171 @@
...
+
+As a final note, remember that if you assign a rule a priority higher than 200,
+it will be tried **after** a rule with the `owise` attribute, and if you assign
+a rule a priority less than 50, it will be tried **before** a rule with no
+explicit priority.
+
+## Exercises
+
+1. Write a function `isEven` that returns whether an integer is an even number.
+Use only two rules and one side condition.
```
ehildenb:
```suggestion
1. Write a function `isEven` that returns whether an integer is an even number.
Use only two rules and one side condition. You may need to refer to [domains.md](LINK) to see the available `Int` operations.
```

ehildenb (approved):

PR #1996, opened 2021-05-31
Parse configs in parallel

### LINKED ISSUE #1867: Better scanner handling

radumereuta:
Fixes: #1867
The most efficient way is with nested parallel streams

There is a slight speed improvement. I'm including the config parsing time reported by kompile -v
EVM: 1.0s, down from 1.6s (3 configs)
Algorand: 6.1s, down from 8.6s

I've tried creating FutureTasks directly with the exact dependencies to send to the ThreadPool, but that made the code look a lot uglier and no speed improvement.

This doesn't entirely fix the problem with extra scanners being generated, but I'm fine with that. The extra scanners are not that expensive anyway. Here is the output for a simple test file:
First time
```
Outer parsing [58 modules]                                   =  0.728s
New scanner: TEST-CONFIG-CELLS                               =  0.138s
New scanner: STDIN-STREAM-CONFIG-CELLS                       =  0.211s
New scanner: STDOUT-STREAM-CONFIG-CELLS                      =  0.206s
Parse configurations [3/3 declarations]                      =  0.452s
New scanner: TEST-RULE-CELLS                                 =  0.121s
New scanner: STRING-KORE-RULE-CELLS                          =  0.197s
New scanner: STRING-COMMON-RULE-CELLS                        =  0.195s
New scanner: K-IO-RULE-CELLS                                 =  0.281s
New scanner: STDOUT-STREAM-RULE-CELLS                        =  0.285s
New scanner: STDIN-STREAM-RULE-CELLS                         =  0.311s
Parse rules [101/101 rules]                                  =  0.575s
```
second time:
```
Outer parsing [58 modules]                                   =  0.750s
Parse configurations [0/3 declarations]                      =  0.263s
New scanner: TEST-RULE-CELLS                                 =  0.121s
Parse rules [0/101 rules]                                    =  0.251s
```

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -283,6 +283,10 @@ public class DefinitionParsing {
         RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
 
         // parse config bubbles in parallel
+        // step 1 - use scala parallel streams to generate parsers
+        // step 2 - use java parallel streams to parse sentences
+        // this avoids creation of extra (costly) threads at the cost
+        // of a small thread contention between the two thread pools
         Map<String, Module> parsed = defWithCaches.parMap(m -> {
             if (stream(m.localSentences()).noneMatch(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration)))
                 return m;
--- a/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
+++ b/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
@@ -283,37 +283,31 @@ public class DefinitionParsing {
         RuleGrammarGenerator gen = new RuleGrammarGenerator(def);
 
         // parse config bubbles in parallel
-        Map<String, java.util.Set<Sentence>> parsed = stream(defWithCaches.modules())
-                .filter(m -> stream(m.localSentences()).anyMatch(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration)))
-                .parallel()
-                .flatMap(m -> {
-                    Module configParserModule = gen.getConfigGrammar(m);
-                    ParseCache cache = loadCache(configParserModule);
-                    try (ParseInModule parser = RuleGrammarGenerator.getCombinedGrammar(cache.getModule(), isStrict, profileRules, files)) {
-                        // each parser gets its own scanner because config labels can conflict with user tokens
-                        parser.getScanner(options.global);
-                        parser.initialize();
-
-                        java.util.Set<Sentence> parsedSet = stream(m.localSentences())
-                                .filter(s -> s instanceof Bubble && ((Bubble) s).sentenceType().equals(configuration))
-                                .map(b -> (Bubble) b)
-                                .parallel()
-                                .flatMap(b -> parseBubble(parser, cache.getCache(), b)
-                                        .map(p -> upSentence(p, b.sentenceType())))
-                                .collect(Collectors.toSet());
-                        return Stream.of(new Tuple2<>(m.name(), parsedSet));
-                    }
-                }).collect(Collectors.toMap(Tuple2::_1, Tuple2::_2));
-
-        Definition defWithParsedConfigs = DefinitionTransformer.from(m -> {
+        Map<String, Module> parsed = defWithCaches.parMap(m -> {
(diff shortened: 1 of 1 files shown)


dwightguth (commented):
I don't understand why we didn't structure this the same way the rule parser is structured

radumereuta:
I think the code looks simpler this way.
For rule parsing, you have scala threads and java threads, and auto detection if you need to build a new scanner.
Config parsing is simpler, one scanner per module and rarely multiple configs per module.

I've tried multiple implementations and found that this performs the best, and it's the shortest piece of code I could come up with.

dwightguth:
I mean, if this really performs better than the alternative, then that's fine, but I need more persuasion. I need to see the code modeled after the rule parsing code that you tried, so I can make sure it looks right to me, and then I need to see the performance numbers of both versions on the Algorand semantics.

dwightguth:
It would also be good to try to test a contrived definition with like, 50 modules, each containing 50 configurations, to make sure it doesn't deadlock. I could easily see this doing exactly that, theoretically speaking.

radumereuta:
@dwightguth I concede. Using two thread pools seems to work well enough for our situation.
There is some thread contention but the 'correct' solution would be to create tasks that never have to wait for another thread to finish. Then handle all the synchronization from the main thread. The implementation will look ugly.

dwightguth (approved):
Looks good, thanks!

PR #1999, opened 2021-06-01
automatically remove `not-lr1` modules from Bison grammars

dwightguth:
I tested this and it ought to fix the tutorial, lesson 3, problem 1. It does this by removing the problematic imports from the `MAP` module and thus allowing Bison to successfully be able to process that module. It's possible some definitions might now behave slightly differently, but I don't think the problem should break anything, and in the worst case, if it does, we can just add an `imports BASIC-K` somewhere to get back the old behavior.


kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -117,10 +117,8 @@ public static Module transformByPriorityAndAssociativity(Module module) {
   }
 
   public static void writeParser(Module module, Module disambModule, Scanner scanner, Sort start, File path, boolean glr, long stackDepth, KExceptionManager kem) {
-    if (!glr && module.att().contains("not-lr1")) {
-        kem.registerInnerParserWarning(ExceptionType.NON_LR_GRAMMAR, "Importing a module that is tagged as not being LR(1) when using Bison's LR(1) parser generator: " + module.att().get("not-lr1"));
-        // print so it appears before we call bison which may not terminate
-        kem.print();
+    if (module.att().contains("not-lr1")) {
+        kem.registerInnerParserWarning(ExceptionType.NON_LR_GRAMMAR, "Importing a module that is tagged as not being LR(1) when using Bison's parser generator: " + module.att().get("not-lr1") + " Module will not be included in Bison grammar.");
```
radumereuta: `kem.registerBisonParserWarning(ExceptionType.NON_LR_GRAMMAR, "Skipping module X tagged as not-lr1")`
The error message is way too wordy. It should give you some information on what happened. The full explanation should be somewhere in the documentation (maybe pending-documentation.md)

radumereuta (changes requested):
One issue with the warning message.
The other thing I don't like is that ParseInModule keeps accumulating parameters. But I don't have any improvement suggestions for now.

radumereuta:
You missed the part where I asked for a new `KExceptionGroup`. Right now it's `Inner parser`, but this is from the Bison parser which is made for programs. I find that slightly misleading. You could then make the error message read as:
`[Warning] Bison Parser: Skipping modules [ML-SYNTAX] tagged as not-lr1.`

Also, I suggest removing `-w2e` flag from the Makefile and add a test for kast that checks the parser actually works.

radumereuta (approved):

PR #2001, opened 2021-06-01
Update Location of cached Bubbles

### LINKED ISSUE #1873: [Bug] duplicate rules breaks kompile error messages
```
module TEST
  imports INT
  claim 1 => 0
  claim 1 => 0
  claim 1 => 0
endmodule
	Source(/home/radu/work/test/./test.k)
	Location(4,9,4,15)
	Source(/home/radu/work/test/./test.k)
	Location(5,9,5,15)
```
Only 2 error messages (instead of 3). Second kompile only one:
```
	Source(/home/radu/work/test/./test.k)
	Location(4,9,4,15)
```
It got very confusing when I was trying to add tests for rules with different attributes.
Should we disallow them?
How about rules with the same contents but different attributes?
How about a rule and a claim which has the same contents?

Also. Error messages don't get updated with the new location if I add new lines before the rule.

radumereuta:
Fixes: #1873

This also fixes the location information for error messages on a second compilation if a Bubble has moved.
I've tested it manually adding rules, removing, moving.
Attributes for rules are still a problem. If you changed some important attributes, you should clean and build again.


kore/src/main/java/org/kframework/utils/errorsystem/KEMException.java
```diff
@@ -125,7 +125,7 @@ public String getMessage() {
         return exception.toString();
     }
 
-    private static KEMException create(ExceptionType type, KExceptionGroup group, String message,
+    public static KEMException create(ExceptionType type, KExceptionGroup group, String message,
```
dwightguth: this was private for a reason. why did you make it public?

dwightguth (changes requested):
This looks good, mostly, but does it also address the issue where if you change just the attributes of a rule, the change doesn't propagate because we end up loading the attributes from the cached rule? It would be good to make sure that that also works correctly if we're going to be touching this code.

kore/src/main/java/org/kframework/utils/errorsystem/KEMException.java
radumereuta: I need to recreate a copy of KExceptions thrown by the parser but with the Source and Location replaced.
I had to expose `exceptionGroup` through a getter and make the general constructor public.
Alternatively, if you want, I could make a function `KMException.updateLocation`. But that seemed custom to me.

radumereuta:
Regarding attributes in rules. That's a bit more complicated. For example, `anywhere` has an effect on parsing (for type inference purposes). So even if I copy the attributes from Bubble to Sentence, the term may not be correct.
Alternatively, I could compare the old attributes, with the new ones, and invalidate the cache for that rule, and let the parser run again. But then, do I do it for all attributes? Or just some? Which ones? If you have any input on these questions, I could update the PR to include these changes as well.

dwightguth:
You're right that some attributes have an effect on the parser. I had forgotten that particular detail. Rather than try to exhaustively enumerate all such attributes, a list which may become out of date, let's just invalidate the cache if the attributes of a rule have changed. Just remember not to invalidate it if the only attributes that change are the location.

dwightguth:
Rather than directly exposing the constructor for KEMException, let's just add a `withLocation` method to KEMException that returns a new exception that is identical except the location. This is in keeping with what we do when we need to promote an object from a warning to an error.

dwightguth (approved):

Issue #1998, opened 2021-06-01
K Tutorial typos/clarifications/issues


JKTKops:
I'm going through the new K tutorial and will use this issue to document findings per @ehildenb 's recommendation.

## lesson 2
* [x] missing word: "... as `banana.color` in the **same** directory ..."
* [x] I found the tutorial smoother after automating passing code to krun without an intermediate file. I'd suggest, immediately before the section header "Productions, Constructors, and Functions," adding "We could also run this program with `krun -cPGM='colorOf(Banana())'`."

## lesson 3
* [x] missing period after "semantics treats it specially"

* [x] lesson-03-d.k cannot kompile with the `--gen-bison-parser` flag. Attempting it produces:
```
[Warning] Inner Parser: Importing a module that is tagged as not being LR(1)
when using Bison's LR(1) parser generator: [SET, LIST]
[Error] Internal: bison returned nonzero exit code: 139
```

@dwightguth says:
"I tested that it works to have a definition that didn't import BOOL , so I assumed that it should work to generate a Bison grammar for such a definition, but actually, it turns out that MAP gets imported implicitly because of the implicit initial configuration in the DEFAULT-CONFIGURATION module"

## lesson 5

* [x] "Some example~~s~~ modules include..."

## lesson 6

* [x] misspelled "functions" after module C

## lesson 7

* [x] clarify praticality of side conditions:

When working with simple K definitions, a reader will wonder why
```
rule isPositive(I) => true requires I >Int 0
rule isPositive(_) => false [owise]
```
wouldn't be written as
```
rule isPositive(I) => I >Int 0
```
In fact, it could be. To better motivate side conditions I think it would be helpful if the first couple of example rules were not predicates, e.g building towards an example like this:
```
syntax Grade ::= A() | B() | F() // harsh school
               | gradeFrom(Int) [function]

rule gradeFrom(I) => A() requires I >Int 90
rule gradeFrom(I) => B() requires I >Int 80
rule gradeFrom(_) => F() [owise]
```

* [x] Additionally, it may help to clarify why side conditions on predicates can be better than inlining the condition. However, since this is the 1-basic section, maybe that should be left for later altogether.

ehildenb: 
"If you have a rule which looks like this:
```
rule SOME_PREDICATE(X, Y, Z) => SOME_REALLY_LARGE_TERM(X, Y, Z)
```
Then the haskell backend will apply the rule when SOME_PREDICATE is around, which may make your term unnecesarily large (and thus slower to reason about).
[...]

PR #2006, opened 2021-06-03
Misc tutorial improvements

dwightguth:
This addresses the remaining issues currently present in #1998. I suggest you review one commit at a time as each contains separate fixes.


k-distribution/k-tutorial/1_basic/07_side_conditions/README.md
```diff
@@ -68,42 +68,42 @@ What this means, in practice, is that this rule has lower priority than other
...
-don't care about this term."
-
-This rule is saying, "if the first rule does not apply, then the result of the
-function is false." Note, however, that `owise` can be used in more complicated
-ways. For example, you can perform additional matching in an `owise` rule. You
-can also have multiple higher-priority rules, or multiple `owise` rules. What
-this means in practice is that all of the non-`owise` rules are tried first, in
-any order, followed by all the `owise` rules, in any order.
+This rule is saying, "if the first rule does not apply, then the grade is a B
+if the percentile is greater than or equal to 80." Note here that we use both
```
ehildenb:
```suggestion
This rule is saying, "if the all other rules which match do not apply, then the grade is a B
if the percentile is greater than or equal to 80." Note here that we use both
```

k-distribution/k-tutorial/1_basic/07_side_conditions/README.md
```diff
@@ -68,42 +68,42 @@ What this means, in practice, is that this rule has lower priority than other
...
+a side condition and an `owise` attribute on the same rule. This is not
+required (as we will see later), but it is allowed. What this means is that the
+side condition is only tried if the previous rules did not apply **and** the
```
ehildenb:
```suggestion
side condition is only tried if the other rules did not apply **and** the 
```

I would prefer to say "other" instead of "previous" in these cases, because "previous" might imply that it means "earlier in the file", which we do not mean to imply.

ehildenb (approved):
Please look at my comments.

PR #2010, opened 2021-06-04
Attributes have restricted value type

dwightguth:
Here we drastically restrict what can be stored in an `Att`. This should make it massively easier later on for us to switch serialization libraries.

Here's what we allow in the value of an attribute now:

* String
* int
* Source
* Location
* Definition
* Module
* Sentence
* K
* KLabel
* Sort
* KList

Here are (most of) the classes that are implicitly exposed as serialized because of the above list:
* scala.collection.Set
* Context
* ContextAlias
* Configuration
* Bubble
* Rule
* Claim
* scala.collection.Seq
* Tag
* Associativity
* Option
* Production
* ProductionItem
* NonTerminal
* RegexTerminal
* Terminal
* KToken
* SortHead
* KApply
* KSequence
* KVariable
* KAs
* KRewrite
* InjectedKLabel

I don't see any reason why we shouldn't be able to get a serialization library to work with all of these, with a bit of effort.

### DIFF
--- a/kore/src/main/scala/org/kframework/attributes/Att.scala
+++ b/kore/src/main/scala/org/kframework/attributes/Att.scala
@@ -5,6 +5,14 @@ import java.util.Optional
 import org.kframework.Collections._
 import scala.collection.Set
 
+/**
+ * Marker class for objects that can be stored as the value of an attribute.
+ *
+ * So far this trait implements no methods, but in the future it may
+ * include some methods relating to serialization/deserialization that
+ * all inheritors must implement. It may depend on what serialization library
+ * we choose to use going forward.
+ */
 trait AttValue
 
 /**
--- a/kore/src/main/scala/org/kframework/attributes/Att.scala
+++ b/kore/src/main/scala/org/kframework/attributes/Att.scala
@@ -3,6 +3,7 @@ package org.kframework.attributes
 import java.util.Optional
 
 import org.kframework.Collections._
+import scala.collection.Set
 
 trait AttValue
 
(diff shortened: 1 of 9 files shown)


kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -108,9 +118,21 @@ object Att {
...
+  private val intClassName = classOf[java.lang.Integer].getName
 
   def from(thatAtt: java.util.Map[String, String]): Att =
     Att(immutable(thatAtt).map { case (k, v) => ((k, Att.stringClassName), v) }.toMap)
+
+  private def apply(thatAtt: Map[(String, String), Any]) = {
+    new Att(thatAtt)
+  }
+
+  def mergeAttributes(p: Set[Att]) = {
```
radumereuta, line 138: Isn't there a standard operator that can be overridden? Maybe `:+`?

kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -3,13 +3,20 @@ package org.kframework.attributes
 import java.util.Optional
 
 import org.kframework.Collections._
+import scala.collection.Set
+
+trait AttValue
```
radumereuta, line 16: Documentation.

radumereuta (changes requested):

kore/src/main/scala/org/kframework/attributes/Att.scala
dwightguth: Maybe, but I see no reason to make it harder to call from Java code.

radumereuta (approved):

PR #2011, opened 2021-06-07
Lesson 10: Strings

dwightguth:


k-distribution/k-tutorial/1_basic/10_strings/README.md
```diff
@@ -1 +1,106 @@
...
+Write a string literal containing the following block of text:
+
+```
+This is an example block of text.
+Here is a quotation: "Hello world."
+	This line is indented.
+ÁÉÍÓÚ
+```
+
+Use the `\t` and `\x` character escapes when doing so.
```
ehildenb: Is the idea that they should kompile a definition which defines this string literal?

Or make a language which allows strings in teh K cell, and `krun ...` it?

dwightguth: I mean, the idea is just to write the string literal, so they get familiar with the syntax of string literals in K.

ehildenb: Well, I guess that doesn't seem like a very fulfilling exercise, to just write down a string literal but have no way to check that it's written correctly. I would either just remove this entirely, or instead just say "Here is an example of a string literal", without asking them to do anything with it.

I guess you want to test that they can figure out how to use the `\t` and the `\x`? I'm not sure it's a useful skill. At the very least you should indicate what the code-points for the given UTF-8 strings there are, or point them to somewhere they can look it up easily.

dwightguth: Well, we can simply remove the exercise entirely, but there really /isn't/ a way for them to test correctness at this point because we haven't taught I/O.

ehildenb (approved):

PR #2012, opened 2021-06-08
Parse record productions in linear time

### LINKED ISSUE #1942: Record productions have exponential behavior
Old title: `computeOverloadPOSet` worst case takes 81s

This worst-case gets activated in one of the definitions @virgil-serbanuta is working on.
Test definition:
https://github.com/radumereuta/verified-smart-contracts/tree/multisig.new/multisig/protocol-correctness/proof/functions
Run as:
`kompile -I ../../../ --backend haskell functions-execute.k -v  --profile-rule-parsing --debug`
`kprove proof-count-can-sign.k -v`

I'm currently investigating, and I'm making this issue to keep track of progress.
Update 1:
```
syntax StateCell ::= invariantStateFull(
      numUsers:Usize,
      userIdToAddress:Map,
      addressToUserId:Map,
      numBoardMembers:Usize,
      numProposers:Usize,
      userIdToRole:Map,
      quorum:Usize,
      actionLastIndex:Usize,
      actionData:Map,
      actionSigners:Map,
      callerAddress:KItem,
      stack:Stack,
      variables:Map,
      performedActions:List)  [function, functional]
```
This is one of the productions that takes a very long time to run `computeOverloadPOSet`.
Update 2:
This slowdown is because of how we parse record productions.
At parsing time there are 16385 overloads for `invariantStateFull` (arity 14), 8193 for `invariantStateStack` (arity 13), 2049 for `invariantState` (arity 11), 1025 for `invariantMultisigState` (arity 10), and many others with mere hundreds.

radumereuta:
Fixes: #1942

This generates one production for each parameter and a list pattern that is able to parse them in linear time.
Immediately after parsing, the terms are replaced with their expanded/collapsed variant. This is done really early to make sure type inference and the other disambiguation steps are being applied on all terms. This minimizes the work of this step and ensures it follows the other parsing rules/restrictions.


radumereuta:
Timings for this branch and (master)
EVM: 1m08s (1m06s)
C: 3m26s (3m20s)
Algorand: 2m31s (2m39s)
regression-new: 1m53s (1m55s)
**the example from the issue: 8.5s (56s)**

The timings I've seen are pretty much the same.
The example that Virgil found now finishes a lot faster.

There are still a few issues with `computeSubsortPOSet`. It can take up to 1s per module on the C semantics, but I can investigate those after this gets merged.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/CollapseRecordProdsVisitor.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+            // construct expanded term with provided items and anonymous variables for missing ones
+            PStack<Term> collapsedItems = ConsPStack.empty();
+            for (NonTerminal nt : mutable(origPrd.nonterminals())) {
+                if (nt.name().isDefined() && children.containsKey(nt.name().get()))
+                    collapsedItems = collapsedItems.plus(children.get(nt.name().get()));
+                else {
+                    Production anonVarPrd = Production.apply(Seq(), Sorts.KVariable(), Seq(Terminal.apply("_[A-Z][A-Za-z0-9'_]*")), Att.empty());
+                    // The name is required so disambiguation doesn't collapse the variables into the same term.
+                    collapsedItems = collapsedItems.plus(Constant.apply("_" + nt.name().getOrElse(() -> "Gen") + uid++, anonVarPrd, tc.location(), tc.source()));
```
dwightguth: This is not a sound way of generating names for variables, because the user could write a variable named `_Gen0` in their rule. You need to either change it back to naming the variable "_" so the anonymous variables pass will give it a fresh name, or else you need to collect all the variables mentioned in the sentence and increment the counter inside a loop until you find one that isn't already used.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -593,6 +607,8 @@ object Production {
   }
 
   val kLabelAttribute = "klabel"
+  private var uid = 1
```
dwightguth: I don't like the insertion of statefulness here. It means that the grammar you generate is going to be different each time you run something, depending on what has been run previously. Can we please modify this so that the grammar generated is deterministically determined by the input module?

ehildenb:
Does this update work identically for example where the record production is _not_ a function applied to arguments separated by commas? For example: https://github.com/kframework/k/blob/master/k-distribution/include/kframework/builtin/domains.md#map-update

Can you add a test of a very large misfix syntax with record productions similar to the one you added? Also, in your test have several rules, which test different uses of record.

kore/src/test/scala/org/kframework/definition/OuterTest.scala
```diff
@@ -33,24 +33,37 @@ class OuterTest {
...
-    val newAtt = Att.add("recordPrd", classOf[Production], prod).add("unparseAvoid")
     val records = prod.recordProductions
-    Assert.assertEquals(4, records.size)
-    Assert.assertEquals(Set(
-      Production(Seq(), sort1, Seq(Terminal("foo"), Terminal("("), Terminal("..."), Terminal(")")), newAtt),
-      Production(Seq(), sort1, Seq(Terminal("foo"), Terminal("("), Terminal("..."), Terminal("baz"), Terminal(":"), nt2, Terminal(")")), newAtt),
-      Production(Seq(), sort1, Seq(Terminal("foo"), Terminal("("), Terminal("..."), Terminal("bar"), Terminal(":"), nt1, Terminal(")")), newAtt),
-      Production(Seq(), sort1, Seq(Terminal("foo"), Terminal("("), Terminal("..."), Terminal("bar"), Terminal(":"), nt1, Terminal(","), Terminal("baz"), Terminal(":"), nt2,Terminal(")")), newAtt)
-    ), records)
+    Assert.assertEquals(7, records.size)
```
dwightguth, line 45: these tests need to assert what the actual generated productions are.

k-distribution/tests/regression-new/checks/checkRecordProdDuplicateLabels.k
```diff
@@ -0,0 +1,48 @@
...
+      stack:Int,
+      variables:Map,
+      performedActions:List) // 30 params
+ rule invariantStateFull(... p1:X, p1:Y) => 1 // duplicate keys here, rest should be fine
+ rule invariantStateFull(... numUsers:_X, stack:_Y, quorum:_Z, variables:_P) => 1
+ rule invariantStateFull(... numUsers:_X) ~> invariantStateFull(...) => 1
+
+ syntax StateCell ::= abc(one:Int)
+ rule abc(... one:X) => 1
+
```
ehildenb: How about a test like `p1, numUsers, p1`, or one with the direction flipped, like `variables, quorum, stack, numUsers`.

Also, please add a second record production that uses misfix syntax, and some example uses of that one too.

dwightguth (changes requested):
Functionally, it looks mostly correct. Overall this is very good work. The code quality needs to be slightly raised in a couple places though, and the tests need to assert more about their outputs.

kore/src/test/scala/org/kframework/definition/OuterTest.scala
radumereuta: This is kind of hard because I generate fresh sorts each time.

k-distribution/tests/regression-new/checks/checkRecordProdDuplicateLabels.k
dwightguth: This feature is only automatically added to the grammar for prefix productions, so your comments about mixfix are out of scope of this change.

kore/src/test/scala/org/kframework/definition/OuterTest.scala
dwightguth: See my comment above about making it deterministic.

dwightguth (approved):
I'm not 100% convinced by the idea of generating increasing integers based on the module name, but it seems to be sufficient to be able to consistently generate the same productions regardless of when the function is called, so it's probably good enough, even though it's not quite how I would have done it. This way is more concise also, which is a positive.

PR #2015, opened 2021-06-08
use default java serialization again

dwightguth:
FST doesn't work with Java 16, nor is it likely that it can be easily
fixed. The same problems apply to kryo. We might be able to make a
serialization library work if we specify how to serialize each type of
object that we serialize, but that would be a lot of work. I'm not
saying we shouldn't do that, but I wanted to present this alternate
solution so we can consider the pros and cons and do some performance
measurements.

Mostly with this PR I want to start a discussion about what the right path forward is. I'm not specifically advocating this solution. That being said, any other solution is likely to be significantly more complicated, so let's keep that in mind.


dwightguth:
Given that this issue is affecting people who run `brew upgrade` on MacOS, what say we merge this early while we continue to work on improving performance?

dwightguth:
Here is the performance data Radu asked for:

| Task | Time Before | Time After |
| ---- | -------------- | ----------- |
| Time to `kompile imp.md` full | 10.8s | 10.5s |
| Time to `kompile imp.md` incremental | 7.8s | 8.3s |
| Time to `kast collatz.imp` | 1.3s | 1.3s |
| Time to kompile KEVM full | 1m22s | 1m8s |
| Time to kompile KEVM incremental | 52.4s | 49.9s |
| Time to kast `add11.json` | 5.4s | 7s |

It looks like the time spent in kompile serializing and deserializing is not a major factor in the performance of kompile. The impact on performance in kast seems very small on small definitions, but is definitely somewhat noticable on larger ones. All in all, this isn't too bad, I don't think. Especially considering it doesn't impact the hot path.

radumereuta (approved):
It doesn't seem to be that bad.
I'm approving this, but maybe let Everett have a look also.

PR #2018, opened 2021-06-09
Lesson 11: Casts

dwightguth:


k-distribution/k-tutorial/1_basic/11_casts/README.md
```diff
@@ -0,0 +1,188 @@
...
+  syntax Exp2 ::= Exp | Exp2 "+" Exp2
+endmodule
+```
+
+This grammar is a little ambiguous and contrived, but it serves to demonstrate
+how a semantic cast might be insufficient to disambiguate a term. If I were 
+to write the term `(I1:Int + I2:Int):Exp2`, the term would be ambiguous,
+because the cast is not sufficiently strict to determine whether you mean
+the derivation `Exp2 -> Exp -> Exp "+" Exp` or the derivation
+`Exp2 -> Exp2 "+" Exp2`.
```
ehildenb: Are these what you meant to write? Neither makes sense to me. Perhaps you should just annotate the productions with names `syntax Exp ::= Int | Exp "+" Exp [exp]`, and the say "whether you mean the derivation using the `exp` production or the `exp2` production.

I guess the `_ -> _` notation is confusing to me, because it's not defined anywhere and maybe users won't be familiar with it from other contexts. Whereas we've already explained putting user-defined attributes on productions and using them to identify them.

k-distribution/k-tutorial/1_basic/11_casts/README.md
```diff
@@ -0,0 +1,188 @@
...
+`1 && true` or `false + true`, but these will cause errors at runtime, because
+the projection cast will fail.
+
+## Exercises
+
+1. Extend the `eval` function in `LESSON-11-D` to include Strings and add a `.`
+operator which concatenates them.
+
+2. Modify your solution from lesson 1.9, problem 2 by using an `Exp` sort to
+express the integer and boolean expressions that it supports, in the same style
```
hjorthjort:
```suggestion
express the integer and Boolean expressions that it supports, in the same style
```

hjorthjort (approved):

k-distribution/k-tutorial/1_basic/11_casts/README.md
```diff
@@ -0,0 +1,188 @@
...
+from the perspective of the grammar and from the perspective of rewriting,
+they are in fact identical. However, the second variant has a unique meaning
+in the **type system** of K: namely, the term inside the cast cannot be a
+**subsort**, i.e., a term of another sort `S2` such that the production
+`syntax S ::= S2` exists.
+
+As a result, if we were to write in the above grammar the term
+`(I1:Int + I2:Int)::Exp2`, then we would know that the first derivation above
+should be chosen, whereas if we want the second derivation, we could write
+`(I1:Int + I2:Int)::Exp`.
```
ehildenb, line 121:
```suggestion

As a result, if we were to write in the above grammar the term
`(I1:Int + I2:Int)::Exp2`, then we would know that the first derivation above
should be chosen, whereas if we want the second derivation, we could write
`(I1:Int + I2:Int)::Exp`.

#### Exercise

Add the (ambiguous) rules `rule I1:Int + I2:Int => I1 +Int I2` and
`rule I1:Int + I2:Int => I1 -Int I2` to the module `LESSON-11-C` above, and inspect
the parse ambiguity you get when you `kompile`. Use strict casts to say that the
`exp` derivation leads to integer addition (first rule), and the `exp2` derivation
leads to integer subtraction (second rule).
```

k-distribution/k-tutorial/1_basic/11_casts/README.md
dwightguth: It's a fairly intuitive way of describing parse derivations, but I went ahead and changed it just in case it confused other people as well.

ehildenb (approved):

PR #2020, opened 2021-06-09
kore-print script

dwightguth:
Blocked on kframework/llvm-backend#411. It is ready for review though.


k-distribution/src/main/scripts/bin/kunparse
```diff
@@ -0,0 +1,245 @@
...
+
+error () {
+  printf "[Error] Critical: $@\n" | $fold_lines 1>&2
+  exit $result
+}
+
+warning () {
+  printf "[Warning] Compiler: $@\n" | $fold_lines 1>&2
+}
+
```
ehildenb: This isn't used anywhere and I'm not convinced it would make sense as a warning message for this script.

In general, I would say both `error()` and `warning()` should probably have `kunparse` (or `$0`) somewhere in their output, so that we know where the message is coming from.

k-distribution/src/main/scripts/bin/kunparse
```diff
@@ -0,0 +1,245 @@
...
+  if [[ "$outputFile" == "-" && -t 1 ]]; then
+    color=true
+  else
+    color=false
+  fi
+fi
+
+if [[ "${#ARGV[@]}" -ne 1 ]]; then
+  error 'Must pass exactly one KORE file as a positional argument.'
+fi
```
ehildenb:
```suggestion
if [[ "${#ARGV[@]}" -ne 1 ]]; then
  print_usage
  error 'Must pass exactly one KORE file as a positional argument.'
fi
```

k-distribution/src/main/scripts/bin/kunparse
```diff
@@ -0,0 +1,245 @@
...
+      outputFile="$2"
+      shift
+      ;;
+
+      -v|--verbose)
+      verbose=true
+      ;;
+
+      --save-temps)
+      trap - INT TERM EXIT
```
ehildenb: I like that! Did not know about `trap - ...`

ehildenb:
I like this script, but I really wish it was called `kprint` and the underlying command used was called `kunparse`. It's so much less of a mouthful to talk about `kprint` than it is to talk about `kunparse`. Basically I wish the names of the `kprint` and `kunparse` commands were switched, or at the very least the underlying `kprint` was changed to be named `kore-print` or something, and this was called `kprint`.

dwightguth:
I mean, I don't disagree with your assessment of the binary tool names, but it's going to be difficult to change without breaking a bunch of stuff.

ehildenb:
I think we discussed that this one should be called `kore-print`. But otherwise looks good.

k-distribution/src/main/scripts/bin/kore-print
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+set -u
+
+export PATH="$(cd "$(dirname "$0")"; pwd):$PATH"
+
+# initialize flags
+ARGV=()
```
ehildenb, line 9: Bash associative arrays don't work on default MacOS bash install.

k-distribution/src/main/scripts/bin/kore-print
```diff
@@ -0,0 +1,241 @@
...
+  -o, --output MODE        Select output mode to use when unparsing. Valid
+                           values are pretty, program, kast, binary, json,
+                           latex, kore, and none.
+      --output-file FILE   Print converted term to FILE
+      --save-temps         Do not delete temporary files when $KORE_PRINT
+                           terminates.
+  -v, --verbose            Print significant commands executed by $KORE_PRINT.
+      --version            Print version of K
+      --                   Interpret all arguments following as positional
+                           arguments.
```
ehildenb: Periods at end of sentences.

k-distribution/src/main/scripts/bin/krun
```diff
@@ -126,19 +126,21 @@ present; see below).
                            This assumes that the initial configuration contains
                            no macro terms. Behavior is undefined if this is not
                            true.
+      --no-substitution-filtering  Don't filter conjuncts with anonymous
+                                   variables from substitution output
```
ehildenb, line 130: Periods.

k-distribution/src/main/scripts/bin/krun
```diff
@@ -612,7 +618,7 @@ if ! $dryRun; then
 
     case "$outputMode" in
       pretty)
-      execute kprint "$kompiledDir" "$kore_output" $color > "$outputFile"
+      execute kprint "$kompiledDir" "$kore_output" $color $filterSubst > "$outputFile"
       ;;
 
       kore)
```
ehildenb, line 624: This whole block should probably be replaced with just a call to `kore-print`, I would think? So that we don't duplicate the logic of selecting `kast` or `kprint` in two places?

dwightguth: I mean, I do intend to refactor the krun script to make better use of these types of things, but I wanted to do that in a separate PR. The changes in this PR are just to make things stylistically consistent and bring krun up to feature parity with kore-print.

ehildenb (approved):

ehildenb (approved):

PR #2022, opened 2021-06-10
kparse-gen

dwightguth:
This is the second of the three scripts I am creating to factor out a new interface for kast. This one's not very interesting, since it basically just invokes the old java kast under the hood. But it provides a consistent interface for all of the tools. The really interesting PR is going to be the next one, which is going to try to create a new Java entry point that takes a K definition with just the syntax, and does outer parsing and then jit parsing of a sentence. Hopefully it will perform at least as well as the current `kast`.

I'm making this a draft PR because there won't be much point in merging this if the next PR doesn't work out, but this ought to be ready for review.


k-distribution/src/main/scripts/bin/kparse-aot
```diff
@@ -0,0 +1,194 @@
...
+
+error () {
+  printf "[Error] Inner Parser: $@\n" | $fold_lines 1>&2
+  exit $result
+}
+
+print_usage () {
+cat <<HERE
+Usage: $KPARSE_AOT <OUTPUT_FILE> [-d DIR] [-m MODULE] [-s SORT]
+Generate an ahead-of-time parser from a K definitino.
```
ehildenb:
```suggestion
Generate an ahead-of-time parser from a K definition.
```

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,194 @@
...
+execute () {
+  (
+  if $verbose; then
+    set -x
+  fi
+  "$@"
+  )
+}
+
+KPARSE_AOT=$(basename "$0")
```
ehildenb:
```suggestion
KPARSE_GEN=$(basename "$0")
```

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,194 @@
...
+
+fold_lines="fold -s"
+
```
ehildenb: Can we define this with the other initialize flags? Or is it here because this is the only use? I still may prefer to have it with the other flags.

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,193 @@
...
+source "$kompiledDir/configVars.sh"
+if [ -z "${sort+unset}" ]; then
+  if [ -z "${declaredConfigVar_PGM+unset}" ]; then
+    sort=KItem
+  else
+    sort="$declaredConfigVar_PGM"
+  fi
+fi
+
+execute kast ${bisonFiles[@]} --bison-stack-max-depth "$stackDepth" -d "$dir" $glr -m "$module" -s "$sort" "${ARGV[0]}"
```
ehildenb:
```suggestion
execute kast ${bisonFiles[@]} --bison-stack-max-depth "$stackDepth" --directory "$dir" $glr --module "$module" --sort "$sort" "${output_file}"
```

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,193 @@
...
+      ;;
+    esac
+  fi
+  shift
+done
+
+if [[ "${#ARGV[@]}" -ne 1 ]]; then
+  print_usage
+  error 'Must pass exactly one KORE file as a positional argument.'
+fi
```
ehildenb:
```suggestion
if [[ "${#ARGV[@]}" -ne 1 ]]; then
  print_usage
  error 'Must pass exactly one output file as a positional argument.'
fi
output_file="${ARGV[0]}"
```

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+set -u
+
+export PATH="$(cd "$(dirname "$0")"; pwd):$PATH"
+
+# initialize flags
+ARGV=()
+bisonFiles=()
```
ehildenb, line 10: Initializing arrays as empty then appending to them later will fail on default Bash on MacOS unless we instruct users to update their Bash. It gives a weird error that we've had to debug for users several times.

ehildenb: One trick I've used is to just initialize the array to something like `ARGV=(--)`, and then when using the array skip the first argument.

k-distribution/src/main/scripts/bin/kparse-gen
```diff
@@ -0,0 +1,195 @@
...
+  -m, --module MODULE      Module to use to generate grammar
+      --no-exc-wrap        Do not wrap exception messages to 80 chars. Keep
+                           long lines.
+      --save-temps         Do not delete temporary files when $KUNPARSE
+                           terminates.
+  -s, --sort               Sort to use as start symbol of parser
+  -v, --verbose            Print significant commands executed by $KUNPARSE.
+      --version            Print version of K
+      --                   Interpret all arguments following as positional
+                           arguments.
```
ehildenb: Either use periods at the end of all sentences here, or none of them.

k-distribution/src/main/scripts/bin/kparse-gen
ehildenb: You missed the update to the error text as well.

k-distribution/src/main/scripts/bin/kparse-gen
dwightguth: As I said in the other PR, you are thinking of associative arrays; this is just a regular array and it works fine on Bash 3.

ehildenb (approved):

PR #2025, opened 2021-06-14
Lessons 13 and 14: K rewriting and evaluation order

dwightguth:
I put these two together into a single pull request because they are structurally connected in terms of both lessons gradually explaining the syntactic sugar that allows you to easily express the semantics of expressions in K.


k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+```
+  syntax KItem ::= S
+```
+
+In other words, every sort is a subsort of the sort `KItem`, and thus a term
+of any sort can be injected as an element of a term of sort `K`, also called
+a **K sequence**.
+
+By default, when you `krun` a program, the AST of the program is inserted as
+ the sole element of a K sequence into the `<k>` cell. This explains why we
```
hjorthjort:
```suggestion
the sole element of a K sequence into the `<k>` cell. This explains why we
```

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+of those rules, apply it, and continue rewriting. However, it is
+**nondeterministic** which rule applies. In theory, it could be any of them.
+By passing the `--search` flag to `krun`, you are able to tell `krun` to 
+explore all possible nondeterministic choices, and generate a complete list of
+all possible final configurations reachable by each nondeterminstic choice that
+can be made.
+
+### Exercise
+
+Pass a program containing no functions to `krun`. Observe the output and try
```
hjorthjort: Any suggestion which semantics you should do this on?

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+
+  syntax KItem ::= freezer1(Val) | freezer2(Exp)
+                 | freezer3(Val) | freezer4(Exp)
+
+  rule <k> E1:Val + E2:Exp ~> K:K </k> => <k> E2 ~> freezer1(E1) ~> K </k> [priority(51)]
+  rule <k> E1:Exp + E2:Exp ~> K:K </k> => <k> E1 ~> freezer2(E2) ~> K </k> [priority(52)]
+  rule <k> E1:Val && E2:Exp ~> K:K </k> => <k> E2 ~> freezer3(E1) ~> K </k> [priority(51)]
+  rule <k> E1:Exp && E2:Exp ~> K:K </k> => <k> E1 ~> freezer4(E2) ~> K </k> [priority(52)]
+
+  rule <k> E1:Val ~> freezer1(E2) ~> K:K </k> => <k> E2 + E1 ~> K </k>
```
hjorthjort: Maybe keep the naming from above, i.e. switch `E1` and `E2`?

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+```
+
+This is of course rather cumbersome currently, but we will soon introduce
+syntactic convenience which makes writing definitions of this type considerably
+easier. For now, notice that there are roughly 3 types of rules here: the first
+matches a K cell in which the first element of the K sequence is an `Exp` whose
+arguments are values, and rewrites the first element of the sequence to the
+result of that expression. The second also matches a K cell with an `Exp` in
+the first element of its K sequence, but it matches when one or both arguments
+ of the `Exp` are not values, and replace the first element of the K sequence
```
hjorthjort:
```suggestion
of the `Exp` are not values, and replace the first element of the K sequence
```

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+
+We can equivalently write it like following:
+
+```
+  rule <k> (I1:Int + I2:Int => I1 +Int I2) ~> _:K </k>
+```
+
+When you put a rewrite inside a term like this, in essence, you are telling
+the rule to only rewrite **part** of the left-hand-side to the right-hand-side.
+In practice, this is implemented by lifting rewrite operator to the top of
```
hjorthjort:
```suggestion
In practice, this is implemented by lifting the rewrite operator to the top of
```

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -1 +1,282 @@
...
+the rule by means of duplicating the surrounding context.
+
+There is a second way that the above rule can be simplified, however. K
```
hjorthjort:
```suggestion
There is a way that the above rule can be further simplified, however. K
```
"second" makes it sound like it's another way to accomplish the same thing.

k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
```diff
@@ -0,0 +1,297 @@
...
+`HOLE`.
+3. For each integer offset prior in the list to the one currently being
+processed, the predicate `isKResult(E)` is conjuncted together and included
+as a side condition, where `E` is the child of the production term with that
+offset, starting from 1. For example, if the attribute lists `1, 2`, then
+the rule generated for the `2` will include `isKResult(E1)` where `E1` is the
+first child of the production.
+
+As you can see if you work through the process, the above code will ultimately
+generate the same contexts present in `LESSON-14-B`. Finally, note that there
```
hjorthjort: I would suggest putting the last sentence in it's own paragraph, so it's clear it is introducing the next two paragraphs.

hjorthjort (changes requested):

hjorthjort (approved):

PR #2026, opened 2021-06-15
Various fixes

radumereuta:
A few changes that don't merit their own PR
- Normalize starting file path (sometimes the first file would contain `/././` in the reported path)
- Moved the message of no of rules parsed to a more consistent place


dwightguth (changes requested):
The first two changes look fine, but I would prefer users not have to specify --debug in order to be able to access the timing information they are collecting.

radumereuta:
Ah, I see. About that. I would like to change it so you have to provide `--debug` in order to generate the timing info. I am also working on a change that puts everything in a single file, with entries ordered by location to make it easier to compare between runs. But that's not quite ready yet.

I just find the `--debug` name more appropriate in this case. `--profile-rules` also disables caches for parsing. I would like to add an option `--disable-cache` or `--no-cache`. I find that quite useful when I do profiling.

radumereuta:
I've removed the changes to `--profile-rules`. That option needs more consideration.

dwightguth (approved):

PR #2029, opened 2021-06-15
change the way we store parsers in the kompiled directory

dwightguth:
This is a refactoring of the structure of the kompiled directory. Instead of naming parsers `parser_FOO` where `FOO` is a configuration variable name, we name them `parser_Foo_BAR` where `Foo` is a sort and `BAR` is a module name. We also create a symlink to the old location when we run `kompile --gen-bison-parser` or `kompile --gen-glr-bison-parser` so that `krun` will still work as expected.

We also modify `kast --gen-parser` and `kast --gen-glr-parser` so that they create a copy of their output in the kompiled directory if the kompiled directory is writeable. If the copy fails (e.g. because the kompiled directory is read-only), then we just ignore the error.

The goal of doing this is so that each time we create a bison parser, it gets cached in the kompiled directory where our `kparse` script can pick it up. This will allow us to create a script that by default, uses the bison parser if one exists, and falls back on the jit parser if it can't find one. That script will be in the next PR once this one is merged.


kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -125,6 +127,9 @@ public int run() {
             KRead kread = new KRead(kem, files.get(), options.input);
             if (options.genParser || options.genGlrParser) {
                 kread.createBisonParser(parsingMod, sort, options.bisonOutputFile(), options.genGlrParser, options.bisonFile, options.bisonStackMaxDepth);
+                try {
+                  Files.copy(options.bisonOutputFile().toPath(), files.get().resolveKompiled("parser_" + sort.name() + "_" + options.module).toPath());
+                } catch (IOException e) {}
```
ehildenb, line 132: I just want to double-check. If I re-call `kast ....` to generate a parser, it will overwrite any existing parser that is there, correct?

ehildenb (approved):

radumereuta (approved):

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
dwightguth: Well, yeah. That's not very likely to happen in practice though unless you're doing it deliberately.

PR #2037, opened 2021-06-28
kparse script

dwightguth:
This PR is marked as draft because this script won't function until #2022 is merged. However, it is ready for review.

### DIFF
--- a/k-distribution/src/main/scripts/bin/kparse
+++ b/k-distribution/src/main/scripts/bin/kparse
@@ -46,28 +46,29 @@ Generate an on-the-fly parser from a K definition and use it to parse a term.
 Mandatory arguments to long options are mandatory for short options too.
 
 $KPARSE expects exactly 1 positional argument. It must be either a readable
-file, or '-'. By default, this file will be parsed using the main syntax
-module of the K definition and the sort of the \$PGM configuration variable.
-The resulting AST, in KORE format, will be output on stdout. This can be
-customized; see below. If no \$PGM configuration variable exists, the default
-sort will be the KItem sort.
+file, or '-' (for stdin). By default, this file will be parsed using the main
+syntax module of the K definition and the sort of the \$PGM configuration
+variable. The resulting AST, in KORE format, will be output on stdout. The
+sort, module, and input format used for parsing can be customized; see below.
+If no \$PGM configuration variable exists, the default sort will be the KItem
+sort.
 
   -d, --directory DIR      Look for a kompiled directory ending in "-kompiled"
-                           under the directory DIR.
+                           under the directory DIR
   -h, --help               Display this help and exit
   -i, --input MODE         Select input mode to use when parsing. Valid values
-                           are program, kast, binary, json, and kore. 
+                           are program, kast, binary, json, and kore.
   -m, --module MODULE      Module to use to generate grammar
       --no-exc-wrap        Do not wrap exception messages to 80 chars. Keep
-                           long lines.
+                           long lines
       --output-file FILE   Print resulting AST to FILE
       --save-temps         Do not delete temporary files when $KPARSE
-                           terminates.
+                           terminates
   -s, --sort               Sort to use as start symbol of parser
-  -v, --verbose            Print significant commands executed by $KPARSE.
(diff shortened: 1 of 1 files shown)


k-distribution/src/main/scripts/bin/kparse
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+set -u
+
+export PATH="$(cd "$(dirname "$0")"; pwd):$PATH"
+
+# initialize flags
+ARGV=()
```
ehildenb, line 9: Bash associative arrays on MacOS issue.

k-distribution/src/main/scripts/bin/kparse
```diff
@@ -0,0 +1,230 @@
...
+
+print_usage () {
+cat <<HERE
+Usage: $KPARSE <INPUT_FILE> [-d DIR] [-m MODULE] [-s SORT]
+Generate an on-the-fly parser from a K definition and use it to parse a term.
+
+Mandatory arguments to long options are mandatory for short options too.
+
+$KPARSE expects exactly 1 positional argument. It must be either a readable
+file, or '-'. By default, this file will be parsed using the main syntax
```
ehildenb:
```suggestion
file, or '-' (for `stdin`). By default, this file will be parsed using the main syntax
```

k-distribution/src/main/scripts/bin/kparse
```diff
@@ -0,0 +1,230 @@
...
+      --no-exc-wrap        Do not wrap exception messages to 80 chars. Keep
+                           long lines.
+      --output-file FILE   Print resulting AST to FILE
+      --save-temps         Do not delete temporary files when $KPARSE
+                           terminates.
+  -s, --sort               Sort to use as start symbol of parser
+  -v, --verbose            Print significant commands executed by $KPARSE.
+      --version            Print version of K
+      --                   Interpret all arguments following as positional
+                           arguments.
```
ehildenb: Periods on sentences.

k-distribution/src/main/scripts/bin/kparse
```diff
@@ -0,0 +1,230 @@
...
+Generate an on-the-fly parser from a K definition and use it to parse a term.
+
+Mandatory arguments to long options are mandatory for short options too.
+
+$KPARSE expects exactly 1 positional argument. It must be either a readable
+file, or '-'. By default, this file will be parsed using the main syntax
+module of the K definition and the sort of the \$PGM configuration variable.
+The resulting AST, in KORE format, will be output on stdout. This can be
+customized; see below. If no \$PGM configuration variable exists, the default
+sort will be the KItem sort.
```
ehildenb: This is a bit confusing to me. Are you saying just that the output on stdout can be customized? Then I would say "will be output on stdout (or to a file, as specified, see below).". If you mean that the output format AST can be changed (to KAST or something), then I thought that was a non-goal of this tool?

k-distribution/src/main/scripts/bin/kparse
dwightguth: This isn't an associative array. Just a regular array, which is fine.

k-distribution/src/main/scripts/bin/kparse
dwightguth: I couldn't tell if you were saying that there should be or should not be periods, but I looked at other --help messages for existing linux tools and they seem to generally not use periods, so I've made it consistent that way.

k-distribution/src/main/scripts/bin/kparse
```diff
@@ -0,0 +1,231 @@
...
+      case "$2" in
+        program) ;;
+        kast) ;;
+        binary) ;;
+        json) ;;
+        kore) ;;
+
+        *)
+	error 'Invalid value for --input. Should be one of "program", "kast", "binary", "json", or "kore".'
+        ;;
```
ehildenb: Formatting here seems messed up. `error` here should be indented one more level right?

Please audit all three of yoru PRs for uses of tabs/spacse and consistent spacing.

ehildenb (approved):

PR #2049, opened 2021-07-02
README: add minimum version for clang, lldb, and llvm

SchmErik:
Newer Ubuntu versions (21.04) no longer support version 8 of these
tools in apt by default. Write a note to clarify the version
requirement for users who are using newer Ubuntu operating systems.


dwightguth (approved):

dwightguth:
Thanks!

PR #2050, opened 2021-07-02
Fix lock contention during serialization and deserialization when using kserver

dwightguth:
FST serialization required us to use a global lock during serialization and deserialization because the FSTConfiguration object was not thread safe. However, when we switched to Java serialization, this lock was no longer necessary, and actually introduced significant slowdown in the performance of krun when run using the kserver, because multiple different K commands were competing for the lock, and all the `kast` processes had to wait to read a serialized definition while another `kompile` process was writing to a completely unrelated file.

We shouldn't actually need this lock. The only time two k processes might actually be writing to the same serialized file at the same time (or reading from a file while another is writing) is when you run kprove, which writes to the parse cache. However, since this is just a cache, we can remove the need for locking simply by updating the cache on the file system as an atomic move operation. By doing this, we ensure that the cache is always readable in a valid state, while allowing it to be updated by kprove. The only cost is that if two kprove processes write to the cache at the same time, one set of writes will be permanently lost. However, since it is only a cache, this is an acceptable behavior.

I have tested that the pl-tutorial, regression-new, and haskell backend test suites all still pass when run with significant parallelism both with and without kserver, and Radu can confirm that this addresses the lock contention issue.

### DIFF
--- a/kernel/src/main/java/org/kframework/utils/BinaryLoader.java
+++ b/kernel/src/main/java/org/kframework/utils/BinaryLoader.java
@@ -43,17 +43,15 @@ public class BinaryLoader {
             throw KEMException.criticalError("Could not create directory " + dir);
         }
         try {
-            saveSynchronized(file, o);
+            saveImpl(file, o);
         } catch (IOException e) {
             throw KEMException.criticalError("Could not write to " + file.getAbsolutePath(), e);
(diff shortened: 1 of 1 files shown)


radumereuta (approved):
Tested and it's back to normal. With a small slowdown, as we expected.

PR #2054, opened 2021-07-05
Rename `pending-documentation.md` to `USER_MANUAL.md`

sskeirik:
This PR does several things:

- it renames `pending-documentation.md` to `USER_MANUAL.md`
- updates links to the former file name to point to the latter file name
- it slightly reorganizes the top-level sections of the file so that they group together well (but not the nested sections).

Elaborating on point 3, the former organization was:

- syntax declaration
- evaluation strategy
- configuration declaration
- rule declaration
- ...

The new organization is:

- syntax declaration
- configuration declaration
- rule declaration
- evaluation strategy
- ...

These, in my mind, are some first steps towards having a proper user manual that we want. I figure we can start small, like this PR, and gradually organize the information we have accumulated over time.

I would love feedback on this. @dwightguth @ehildenb I have added you both as reviewers, since I wasn't sure who should look. I also added @shd101wyy because of the fact that web links need to be updated, so we can double-check that no web link was missed.


shd101wyy (approved):
Tried running the website locally and it looks good to me :+1: 
Thank you @sskeirik !

nishantjr:
Instead of manually adding a ToC we should either use pandoc or Javascript to generate one. I can see a manually added one getting out of sync quickly.

sskeirik:
@nishantjr @dwightguth So, I don't have the bandwidth to add a proper ToC sync mechansim, so I removed that part from this PR. I figure we can find a way to generate that in a separate PR.

EDIT: I updated the my summary comment at the top of the file to reflect this.

dwightguth:
If ever there were a great example of why GitHub should be using the patience diff algorithm, here it is. But this looks good.

dwightguth (approved):

PR #2059, opened 2021-07-07
relicense as BSD 3 clause

dwightguth:


ehildenb (approved):

h0nzZik:
Will there be a similar PR for the LLVM backend?

dwightguth:
Good point. I will do that next.

PR #2065, opened 2021-07-08
Enable bison by default

dwightguth:


radumereuta:
Tested this on my machine with the kserver on and I got:
Master: 1m47s, 
Here is the cpu usage dropoff:
image

This branch: 2m57s
image
It seems there are 1 or 2 processes that take a lot longer.
This looks like a significant performance drop.

But I have another, **bigger,** issue with making the bison parser the default. The bison parser only accepts a restricted set of grammars. This has the potential to negatively impact the new user experience. The java parser kind of just works. At least it's consistent with the semantics parser. You get similar error messages, you can parse the same language.

I think the bison parser should remain an advanced feature. An optimization step that can be applied to certain definitions when it makes sense.

ehildenb:
I agree that this performance drop means we should not make it default. But it also sounds like we have a reasonable explanation for that? Namely that we are spending a bunch of time building the parsers for the tests, but not using them.

In which case, I guess we need a separate argument for why it should be default.

kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -387,60 +388,79 @@ private static void processProduction(Production prod, Module module, Module dis
...
+            "  node *n2 = malloc(sizeof(node));\n" +
+            "  n2->symbol = \"dotk{}\";\n" +
+            "  n2->str = false;\n" +
+            "  n2->location = @$;\n" +
+            "  n2->hasLocation = false;\n" +
+            "  n2->nchildren = 0;\n" +
+            "  n2->sort = \"SortK{}\";\n" +
+            "  n->children[1] = n2;\n" +
+            "  value_type result = {.nterm = n};\n" +
+            "  $$ = result;\n" +
```
ehildenb, line 408: Are these changes (to KSyntax2Bison.java) related to the goals of this PR (making Bison default?) If not, can we plul it out into its own PR?

PR #2071, opened 2021-07-11
Check cell sorts

### LINKED ISSUE #2066: Z3 captured errors should include location information about the context
Type inference, in particular, should include some information about what rule it was trying to parse when it got the error message.

This is so we can better debug definitions or if there is something wrong with K it can be used to get a minimized example.

### LINKED ISSUE #2069: [Bug] [kompile] - Fails with Z3 warning on simple configuration composition
K Version
---------

```
% kompile --version
K version:    5.1.93
Build date:   Thu Jul 08 21:16:33 UTC 2021
% z3 --version
Z3 version 4.8.7 - 64 bit
% lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.5 LTS
Release:        18.04
Codename:       bionic
```

Description
-----------

Kompile will not handle a very simple configuration composition, where the cell-sorts of the sub-configurations are mentioned directly in the super-configuration module.

Input Files
-----------

File `web3.k`

```
requires "domains.md" 
 
module EVM 
    imports INT 
 
    configuration 
      <kevm> 
        <k> $PGM:Int </k> 
        <account> 0 </account> 
      </kevm> 
 
endmodule 
 
module WEB3 
    imports EVM 
 
    configuration 
      <kevm-client> 
        <kevm/> 
        <foo> $FOO:KItem </foo> 
      </kevm-client> 
 
    syntax AccountItem ::= AccountCell 
 // ---------------------------------- 
 
endmodule 
```

Reproduction Steps
------------------

```
% kompile web3.k --backend llvm --no-llvm-kompile --main-module WEB3 --syntax-module WEB3
[Error] Internal: Unexpected result from z3: (error "line 35 column 13: unknown
constant SortAccountCell")
```

Expected Behavior
-----------------

Either this should `kompile` correctly, or the error message it outputs should be more informative about (i) the location of the problematic sentences, and (ii) what exactly the problem is.

radumereuta:
Fixes: #2066 
Fixes: #2069
It turns out that we disabled the definedness checks for cell sorts completely and only Z3 complained about this.
If you mistyped a cell sort, this problem would be present even before the parsing refactoring that triggered the Firefly error.

To fix it, I add a temporary SortSyntax with the cell sort found on the RHS and delete it after we parse the configurations, then check for sorts soundness.

You can now use the sort of a cell in the same module with the configuration declaration.

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkCellSortDeclFail.k
+++ b/k-distribution/tests/regression-new/checks/checkCellSortDeclFail.k
@@ -1,0 +1,14 @@
+// Copyright (c) 2016-2021 K Team. All Rights Reserved.
+
+module CHECKCELLSORTDECLFAIL-SYNTAX
+    syntax Pgm
+endmodule
+
+module CHECKCELLSORTDECLFAIL
+    imports CHECKCELLSORTDECLFAIL-SYNTAX
+
+configuration <T> <k> $PGM:Pgm </k> </T>
+
+syntax Pgm ::= MisTypedCell
+
+endmodule
--- a/k-distribution/tests/regression-new/checks/checkCellSortDeclFail.k.out
+++ b/k-distribution/tests/regression-new/checks/checkCellSortDeclFail.k.out
@@ -1,0 +1,4 @@
+[Error] Compiler: Could not find sorts: [MisTypedCell]
+	Source(checkCellSortDeclFail.k)
+	Location(12,16,12,27)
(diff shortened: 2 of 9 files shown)


dwightguth (changes requested):
looks alright, if a bit messy, but please add a test for the original bug.

dwightguth (approved):

PR #2082, opened 2021-07-14
add #let A = B #in C syntax

dwightguth:
Turns out to be pretty simple.


radumereuta (changes requested):
This could use a test.
If you can piggyback on an existing one that would be even better.
Other than that, the changes to the code look fine.

dwightguth:
Whoops. Good catch. I wrote a test and apparently forgot to `git add` it

radumereuta (approved):
I will let @ehildenb add the automerge if he is happy with the functionality.

k-distribution/tests/regression-new/let-test/test.k
```diff
@@ -0,0 +1,6 @@
+module TEST
+  imports INT
+
+  rule 0 => #let FOO = 1 #in FOO
```
ehildenb, line 6: Let's add a few more rules, specifically one with 1 nested binding, one with 2. If you are happy with the output they give, then feel free to add automerge @dwightguth

kernel/src/main/java/org/kframework/compile/RewriteAwareVisitor.java
```diff
@@ -69,6 +69,18 @@ public void apply(KApply k) {
                 isRHS = wasRHS;
                 isLHS = wasLHS;
                 apply(k.items().get(1));
+            } else if (k.klabel().name().equals("#fun3")) {
```
ehildenb: Should this be `#fun3` or `#let`? I guess the next case was handling `#fun3` before, and now this case is handling `#fun3` and the next case is handling `#let`.

ehildenb: Ah, actually, the cases are identical. Why don't we just do `.equals('#fun3') or .equals('#let')` and have one case?

nishantjr:
@dwightguth Could we add some documentation please.

nishantjr:
@dwightguth @ehildenb Are we also thinking about syntax for allowing let bindings that works across cells as well?
Looking at this I don't think we can do:

```
rule <foo> I => Foo </foo>
      <bar> _ => Foo </bar>
      where Foo :=K foo(I)
```

PR #2090, opened 2021-07-19
Lesson 16: collections

dwightguth:


k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,345 @@
...
+## Maps
+
+The most frequently used type of data structure in K is the map. The sort
+provided by K for this purpose is the `Map` sort, and it is provided in
+[domains.md](../../../include/kframework/builtin/domains.md) in the `MAP`
+module. This type is not (currently) polymorphic. All `Map` terms are maps that
+map terms of sort `KItem` to other terms of sort `KItem`. Of course, a `KItem`
+can contain any other sort except a `K` sequence. If you need to store such
+a term in a map, you can always use a wrapper such as
+`syntax KItem ::= kseq(K)`.
```
ehildenb:
```suggestion
map terms of sort `KItem` to other terms of sort `KItem`. A `KItem`
can contain any other sort except a `K` sequence. If you need to store such
a term in a map, you can always use a wrapper such as
`syntax KItem ::= kseq(K)`.
```

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,345 @@
...
+       <functions>... X |-> S ...</functions>
+       <fstack> .List => ListItem(stackFrame(K)) ...</fstack>
+
+  // return statement
+  rule <k> return I:Int ; ~> _ => I ~> K </k>
+       <fstack> ListItem(stackFrame(K)) => .List ...</fstack>
+
+  syntax Bool ::= isKResult(K) [function, symbol]
+  rule isKResult(_:Int) => true
+  rule isKResult(_) => false [owise]
```
ehildenb, line 222: What's the reason we're using this approach rather than `syntax KResult ::= Int`?

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,345 @@
...
+`<fstack>` cell and pretty prints the current stack trace. Test this function
+by creating a new expression that returns the current stack trace as a string.
+Make sure to update `isKResult` and the `Exp` sort as appropriate to allow
+strings as values.
+
+## Sets
+
+The final primary data structure sort in K is a set, i.e., an idempotent
+unordered collection where elements are deduplicated. The sort provided by K
+for this purpose is the `Set` sort and it is provided in `domains.md` in the
```
ehildenb, line 275: Should we make `domains.md` a link like in the rest of the file, or are you trying to get them familiar with navigating there on their own?

k-distribution/k-tutorial/1_basic/16_collections/README.md
dwightguth: I would rather introduce people to this syntax first since KResult isn't really intended to be a sort, and we want to discourage people from casting terms to sort KResult, since doing so interacts badly with some features of strictness. It's better than they think of KResult as a function rather than a sort.

k-distribution/k-tutorial/1_basic/16_collections/README.md
dwightguth: Stylistically, I have been linking the first reference to domains.md in each lesson but not future occurrences.

k-distribution/k-tutorial/1_basic/16_collections/README.md
ehildenb: A total function into Bool is isomorphic to a set, which is what a Sort is in the model theory of matching logic. So I guess I assumed the different definitions were semantically identical. But it sounds like there is some operational difference?

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,344 @@
...
+
+  syntax Bool ::= isKResult(K) [symbol, function]
+  rule isKResult(_:Int) => true
+  rule isKResult(_) => false [owise]
+endmodule
+```
+
+There are several new features in this definition. First, note we import
+the module `ID-SYNTAX`. This module is defined in `domains.md` and provides a
+basic syntax for identifiers. This syntax is only imported when parsing
```
ehildenb:
```suggestion
basic syntax for identifiers (which we use for variable names in this programming language). This syntax is only imported when parsing
```

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,344 @@
...
+help make the output more readable.
+
+### Exercise
+
+Create a sort `Stmt` that is a subsort of `Decl`. Create a production of sort
+`Stmt` for variable assignment. Feel free to use the syntax
+`syntax Stmt ::= Id "=" Exp ";"`. Write a rule that implements variable
+assignment using a map update function. Then write the same rule using a
+map pattern. Test your implementations with some programs to ensure they behave
+as expected.
```
ehildenb: This one is a bit confusing to me. Should we be adding this in addition to the existing `Decl` assignment operator?

If it's meant that the `Decl` operator is variable declaration, and this is variable update, maybe we should refer to this as `variable update` or `variable assignment update` instead of just `variable assignment`?

I'm just confused by reading this if the intention is that it's added in addition to the existing assignment/declaration operator, or instead of.

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,344 @@
...
+
+### Exercise
+
+Add a term of sort `Id` to the `stackFrame` operator to keep track of the
+name of the function in that stack frame. Then write a function
+`syntax String ::= printStackTrace(List)` that takes the contents of the
+`<fstack>` cell and pretty prints the current stack trace. Test this function
+by creating a new expression that returns the current stack trace as a string.
+Make sure to update `isKResult` and the `Exp` sort as appropriate to allow
+strings as values.
```
ehildenb: Tell them teh module name in `domains.md` to find the string operators (eg. `_+String_`) and to find the hook they need (I think `Id2String`).

k-distribution/k-tutorial/1_basic/16_collections/README.md
```diff
@@ -1 +1,344 @@
...
+## Exercises
+
+1. Modify your solution to Lesson 1.14, Problem 2 and introduce the sorts
+`Decls`, `Decl`, and `Stmt` which include variable and function declaration
+(without function parameters), and return and assignment statements, as well
+as call expressions. Use `List` and `Map` to implement these operators, making
+sure to consider the interactions between components, such as saving and
+restoring the environment of variables at each call site. Don't worry about
+local function definitions or global variables for now. Make sure to test the
+resulting interpreter.
```
ehildenb, line 348: This seems like a bit of a stretch in difficulty. Are you intending that they store previous call-states? Like are you intending something like:
```
<k> call f(...) ~> REST => FBODY </k>
<callState> .List => ListItem(REST | MEM | DECL) ... </callState>
<state> MEM => .Map </state>
<declared> DECL => .Set </declared>
```

Have we introduced the concept of storing the continuation (the `~> REST` portion)?

I think we'll need to provide more guidance about how to complete this. I think you'll have to at least mention something like "Use a new cell called `<callStack>` which is a list of call-states, where a call-state consists of the current continuation, the current state, and the current declared variable." or the like.

k-distribution/k-tutorial/1_basic/16_collections/README.md
dwightguth: I mean, if I say something like:

```
syntax Exp ::= Int | Exp "+" Exp
syntax KResult ::= Int

rule I:KResult + I2:KResult => whatever
```

This will not parse, whereas if I say

```
rule I + I2 => whatever
requires isKResult(I) andBool isKResult(I2)
```

That's fine. Furthermore, if I say:

```
syntax Exp ::= Int | Exp "+" Exp [strict, hybrid]
syntax KResult ::= Int

rule I:KResult => whatever
```

`I` will not match `1 + 2`, even though `isKResult(1 + 2)` will return true.

Thus it's pretty much always better to use isKResult rather than `:KResult`, and I'd prefer to introduce the material in a fashion that encourages that behavior.

k-distribution/k-tutorial/1_basic/16_collections/README.md
dwightguth: I'm not sure I agree with the claim that this will be too hard. It's essentially just the combination of the previous examples in this lesson. The only really new component involves saving and restoring certain cells when a new call frame is created, something that we already explained how to do earlier in this lesson.

k-distribution/k-tutorial/1_basic/16_collections/README.md
ehildenb: Hmmm, so we're treating `KResult` specially, in the sense that it's not a proper sort, but a declaration for a function `isKResult`. I guess I don't like that very much, but makes sense.

ehildenb (approved):

Issue #2033, opened 2021-06-15
Clarity on tutorial lessons


ehildenb:
Issues found by the folks going through the tutorial. I short check-list is provided here, more details for each issues are provided in comments.

@dwightguth I've gone through and put (in bold following each bullet point) the things to check for that bullet point.

ehildenb:
You need to `kompile --enable-search --backend llvm` to have the binary generated that will do a search on LLVM backend. This should be made clearer.

rybla:
Some minor notes:

- [x] Lesson 9: Some formatting errors in the table, because of how "paragraphs" within table cells are handled. **Check that tables are formatted correctly on github.com and on kframework.org**.
- [x] Lesson 9: It would be nice to see example of how the color attribute is used, since its a little unclear if/what the color attribute takes as an arguments (and what are valid colors? I see that certain named colors work). **Provide an example the user can run on the command-line with the color attribute.**

- [x] Lesson 10: Formatting errors in tables; some of the escaped need double-backslashes. **Check formatting of tables on github.com and kframework.org**.
- [x] Link to lesson 11 is broken. **Check end-of-document links**.

- [x] Lesson 11: typo: “provies” => “provides”. **Fix**.

- [ ] Lesson 12: what a "freezer" is is not clearly explained. **Make the link between the user-defined freezers (with manual strictness) and generated ones more explicit, maybe explaining the naming scheme for freezers.**

JKTKops:
- [x] Lesson 10: typo: `end => start` => ` end >= start`. **Fix**

- [x] Lesson 11: It's sort of mentioned off-hand, but it might help to state clearly that casts/strict casts are a compile-time feature, while projection casts are a runtime feature that can cause runtime errors. **Make explicit that casts/strict casts are for parsing, and projection casts will cause a runtime halt (but nothing at kompile time). Maybe make sure there is an example of such a runtime halt in the exercises.**

- [x] Lesson 12: unbalanced paren before the last code block, I'm not sure what was intended there. **Fix typo**.

- [x] Lesson 13: "be aware" => "remember"? **Change wording**.

[...]

PR #2093, opened 2021-07-20
Fix tutorial issues

dwightguth:
This fixes most of the issues present in #2033. There were two that I chose not to address:

* The comment about freezers in lesson 13 is not in need of clarification because for the purposes of lesson 13, freezers are just regular productions, and in lesson 14, we explain exactly how freezer productions are autogenerated by contexts.
* The concern about how we presented strictness and contexts was in direct contradiction to the opinion of someone else who took the tutorial, so I'm inclined to think it's probably fine as is.

The rest I tried to address in some form or other, whether it be a simple wording or spelling fix, or a more complex explanation or rewrite of a section.

I chose to make the table in lesson 9 a `<table>` element because it would have looked very bad with a 700-character line in it, but in lesson 10, the longest line was only 108 characters, so I decided to leave it as markdown. Let me know if you think it's better to consistently use `<table>` for both.

Here are the links to the tables as rendered by github:
* https://github.com/kframework/k/tree/tutorialfix/k-distribution/k-tutorial/1_basic/09_unparsing#custom-unparsing-of-terms
* https://github.com/kframework/k/tree/tutorialfix/k-distribution/k-tutorial/1_basic/10_strings#the-string-sort


k-distribution/k-tutorial/1_basic/11_casts/README.md
```diff
@@ -125,19 +125,23 @@ should be chosen, whereas if we want the second derivation, we could write
 Write a function `eval` which takes a term of sort `Exp2` and returns an `Int`.
 The function should be capable of evaluating any term of sort `Exp2` to the
 integer value which it expresses. You will need to write strict casts in order
-to disambiguate these rules.
+to disambiguate these rules. You will not be able to directly write terms of
+sort Exp2 directly in a program, so you will have to create new function
+symbols that rewrite constant symbols to terms of sort Exp2 and use strict
```
ehildenb:
```suggestion
sort `Exp2` directly in a program, so you will have to create new function
symbols that rewrite constant symbols to terms of sort `Exp2` and use strict
```

k-distribution/k-tutorial/1_basic/13_rewrite_rules/README.md
```diff
@@ -70,23 +70,24 @@ saw the output we did in Lesson 1.2.
...
 of those rules, apply it, and continue rewriting. However, it is
-**nondeterministic** which rule applies. In theory, it could be any of them.
+**non-deterministic** which rule applies. In theory, it could be any of them.
 By passing the `--search` flag to `krun`, you are able to tell `krun` to 
-explore all possible nondeterministic choices, and generate a complete list of
+explore all possible non-deterministic choices, and generate a complete list of
 all possible final configurations reachable by each nondeterminstic choice that
-can be made.
+can be made. Note that the `--search` flag to krun only functions if you pass
+`--enable-search` to kompile first.
```
ehildenb, line 90:
```suggestion
can be made. Note that the `--search` flag to krun only functions for the LLVM backend if you pass
`--enable-search` to kompile first.
```

ehildenb (approved):

Issue #2100, opened 2021-07-27
Remove redundant exception information from some error messages


ehildenb:
See https://github.com/kframework/k/issues/2064#issuecomment-887111404 for context.

PR #2101, opened 2021-07-27
Outer: remove redundancy in parser error message

SchmErik:
Consider the following files:
primay.k:

  requires "secondary.k"

  module PRIMARY

    syntax Foo ::= "bar"

  endmodule

secondary.k:

  module SECONDARY

    syntax Foo ::= | "baz"

  endmodule

When compiling primary.k, the following error message is generated:

  18.

  Was expecting one of:

  "non-assoc" ...
      <STRING> ...
      <STRING_REGEX> ...
      "(" ...
      "left" ...
      "right" ...
      "List" ...
      "NeList" ...
      <UPPER_ID> ...
      <LOWER_ID> ...
      <NAT> ...
      <UPPER_ID> ...
      "List" ...
      "NeList" ...
      <LOWER_ID> ...
      )
  	Source(/home/erik/samps-k/secondary.k)
  	Location(3,18,3,18)

This error message's "ParseException: ..." message is redundant. Remove
this redundancy by not passing the ParseException object when
registering the exception so that the error message looks like the
following:

  "NeList", <LOWER_ID>]
  	Source(/home/erik/samps-k/secondary.k)
  	Location(3,18,3,18)

### DIFF
--- a/kernel/src/main/javacc/Outer.jj
+++ b/kernel/src/main/javacc/Outer.jj
@@ -227,7 +227,7 @@ public class Outer {
                   sb.deleteCharAt(sb.length() - 1);
                   return sb.toString();
                 }
-              }).collect(Collectors.toList()).toString(), e, source,
+              }).collect(Collectors.toList()).toString(),  source,
             new Location(e.currentToken.next.beginLine, e.currentToken.next.beginColumn, e.currentToken.next.endLine, e.currentToken.next.endColumn));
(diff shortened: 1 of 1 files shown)


SchmErik:
I know this is not exactly the solution we discussed in #2100 but it's less code so I thought I'd submit for feedback...

SchmErik:
This not the correct solution. @radumereuta @dwightguth please review #2102 instead

PR #2102, opened 2021-07-27
Outer: remove redundancy in parser error message

SchmErik:
Consider the following files:
primay.k:

```k
requires "secondary.k"

module PRIMARY

  syntax Foo ::= "bar"

endmodule
```

secondary.k:

```k
module SECONDARY

  syntax Foo ::= | "baz"

endmodule
```

When compiling primary.k, the following error message is generated:
```
  18.

  Was expecting one of:

  "non-assoc" ...
      <STRING> ...
      <STRING_REGEX> ...
      "(" ...
      "left" ...
      "right" ...
      "List" ...
      "NeList" ...
      <UPPER_ID> ...
      <LOWER_ID> ...
      <NAT> ...
      <UPPER_ID> ...
      "List" ...
      "NeList" ...
      <LOWER_ID> ...
      )
  	Source(/home/erik/samps-k/secondary.k)
  	Location(3,18,3,18)
```

This error message's "ParseException: ..." message is redundant. Remove
the redundancy by adding a boolean field to the KException class and
overloading several other methods to create KException objects that do
not print exception messages that are redundant to the actual error
messages. Doing so shortens the previous error message to look like this:
```
  "NeList", <LOWER_ID>]
  	Source(/home/erik/samps-k/secondary.k)
  	Location(3,18,3,18)
```

### DIFF
--- a/k-distribution/tests/regression-new/checks/missingBar.k
+++ b/k-distribution/tests/regression-new/checks/missingBar.k
@@ -1,0 +1,9 @@
+module MISSINGBAR
+
+  syntax Foo ::= foo ()
+                 bar ()
+
+  rule foo () => bar ()
+
+endmodule
+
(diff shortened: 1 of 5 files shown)


ehildenb:
Please add a test it `k-distribution/tests/regression-new/checks` for this new error message.

dwightguth (commented):
Looks great! Good work! I agree we should add a test though.

SchmErik:
I added a test in a separate commit just now

radumereuta (approved):

radumereuta:
Fixes: #2100

PR #2108, opened 2021-07-29
Use module signatures to clean up the K prelude

dwightguth:
The goal of this PR is to make use of the module signatures feature in order to clean up the syntax exported by each module in the K prelude. Roughly speaking, we have applied two major heuristics when deciding how to determine what imports and syntax are private and what is public:

1. Modules that import other modules import them privately unless one module is clearly an explicit superset of the other. For  example, INT imports INT-SYNTAX publicly, but imports BOOL privately.
2. If a module has syntax that is intended to be part of the implementation of the module, but should not be referenced directly by users, that syntax is made private. The exception is that we have not done this in 100% of the cases involving the OCAML and Java backends.


k-distribution/tests/regression-new/array-haskell/1.test
```diff
@@ -1 +1 @@
-arr ( .List, 0, 0 ) [ 0 ]
+makeArray(0, 0) [ 0 ]
```
ehildenb, line 1: What is going on here? Why doesn't it simplify down to the constructor anymore?

k-distribution/tests/regression-new/rat/defined/java/4.a.out
```diff
@@ -1,5 +1,5 @@
 <T>
   <k>
-    < 29 , 18 > ;  < 1 , 18 > ;  < 35 , 54 > ;  < 15 , 14 > ;  < 52 , 9 > ;  < 38 , 9 > ;  < 35 , 9 > ;  < 45 , 7 > ;  < 47 , 6 > ;  < -37 , 6 > ;  < 35 , 6 > ;  < 5 , 42 > ;  12 ;  -2 ;  35 ;  < 5 , 7 > ;  .Pgm
+    29 /Rat 18 ;  1 /Rat 18 ;  35 /Rat 54 ;  15 /Rat 14 ;  52 /Rat 9 ;  38 /Rat 9 ;  35 /Rat 9 ;  45 /Rat 7 ;  47 /Rat 6 ;  -37 /Rat 6 ;  35 /Rat 6 ;  5 /Rat 42 ;  12 ;  -2 ;  35 ;  5 /Rat 7 ;  .Pgm
```
ehildenb, line 3: Once again, why do we not simplify down to the constructor anymore?

k-distribution/tests/regression-new/array-haskell/1.test
dwightguth: This is the input, not the output. What is going on is we have made the internal implementation of arrays private, so we now have to go through the (public) array constructor.

k-distribution/tests/regression-new/rat/defined/java/4.a.out
dwightguth: This is similar to the other one, but a little different. We made the internal constructor private, so the rule that was present in the K definition that rewrote the internal construtor into the constructor seen here in the output file was no longer able to be written. I just deleted that rule since it wasn't really necessary, and as a result, the output contains just the internal constructor, which gets written via the format attribute as /Rat.

k-distribution/tests/regression-new/equals-pattern/test.k
```diff
@@ -3,6 +3,8 @@
 module TEST-SYNTAX
   imports INT
   imports K-EQUAL
+  imports BOOL
+  imports BASIC-K
```
nishantjr, line 7: I'm a little concerned that having a long list of modules that most semantics would need to import would lead to semantics just importing `DOMAINS`. I'm not sure if that would negate the modularity benefits of having them private in the first place?

Could we merge `K-EQUAL` into `BOOL`?
Could we have `BASIC-K` imported by default?

nishantjr (changes requested):
I feel that this change would break most semantics.
We should thus be a bit cautious about it. 

Ideally, I'd like to:

1. allow a transition period (6 weeks?) where users can migrate from the previous version to the current one. I'm not sure if there is an easy way to implement this. A new flag seems like an overkill. 
2. wait for  #2115 to be completed, so that if we find any further improvements that that PR enables, we can apply them too.

k-distribution/tests/regression-new/equals-pattern/test.k
dwightguth: BASIC-K actually is imported everywhere by default except the program parser, where we deliberately import nothing implicitly. The import exists here because we actually want to include it in the program parser for this definition.

We could maybe make `BOOL` import K-EQUAL publicly though. What do others think about this?

dwightguth:
I really don't think any kind of transition period is necessary here. Yeah, I expect a bunch of stuff to break, but I also expect it to be very easy to fix. Any given broken semantics ought to be able to be fixed in less than an hour. It's not like these are going to be complicated fixes that take a lot of time to get right.

As far as the partial imports go, we can wait if you want, but I really wouldn't change anything about this PR as a result of that change, so I don't see much point.

radumereuta (commented):
lgtm
I did some performance testing in regression-new and it's the same as master.
For peace of mind. Can you test one of the big definitions and see what it takes to make it work again? (I think you are most familiar with the c-semantics).

nishantjr:
LGTM then.

nishantjr (approved):
LGTM

radumereuta (approved):

Issue #1585, opened 2020-10-07
[KIP] - We should have a comprehensive, user-facing reference manual for K.


malturki:
[KIP] - We should have a comprehensive, user-facing reference manual for K.

Motivation
----------

It can be difficult for K users to hop around tutorials and multiple references (online documents and academic papers) every time they need to learn about specific features that they want to use in their projects. Furthermore, having one resource to which we can refer people can significantly improve their user experience while working with K, and can make our lives easier as well (think of the various support channels we have --- riot channel, stackoverflow, ... etc).

Suggestion
-----------

I’d like to suggest having a comprehensive, user-facing reference manual for K. The objective would be to have one reference that we can point people to to learn about K or look up information about any specific feature of K, and from which they may be referred to other resources (existing tutorials, online projects, academic papers, .. etc). 

Unlike existing resources,  the reference manual will be
  - comprehensive: describing all features of K and K tools designed for K users — the whole spectrum of users, and
  - cohesive: describing related features and feature design decisions in a structured way.

I think this will complement existing resources quite well. The pending documentation we have right now can focus more on recent and experimental features that haven’t made their way to the reference manual yet. Also, the manual will not be a guide to defining different semantic styles in K, like the K tutorials, although it will probably include many smaller examples to illustrate concepts.

Example Structure
------------------

The following is a suggested general structure of the manual (Note: we probably already have material for most of these topics):

  1. Introduction (what K is, its philosophy, highlights of why K, perhaps also a quick overview of its foundations, manual structure)
  2. Using K (a quick tutorial on getting K, installing it, highlighting its tools and the different K backends and how they are used with a simple hello-world-like example)
  3. Syntax and Parsing (everything that has to do with defining syntax in K and generating language parsers, with examples)
  4. Rules and Rewriting (everything that has to do with defining semantics in K - configurations, continuations, matching, types of rules, rule attributes, … etc, and krun, with examples)
  5. Modules and Predefined Data Modules (syntax vs. semantics modules, importing modules, modules included in K’s distribution- Sets, Lists, Maps, …, with examples)
  6. Analyzing Programs (exploring the reachable state space of a program, proving invariants, model checking…, a quick overview of the theory behind it with appropriate pointers and with examples)
  7. Advanced Features (optional)
  8. Debugging K specs (methods and tools that help in debugging K specs)
  9. Reference (A complete list of commands and their supported flags)
  10. Index

ehildenb:
[...]

PR #2112, opened 2021-08-04
Add introduction to the user manual

sskeirik:
This is a first attempt at writing an introduction to the user manual that we now have.

Note that I am using this issue (#1585) as an overall reference for what to write and in what order.


USER_MANUAL.md
```diff
@@ -1,12 +1,81 @@
...
+collection of language-generic command-line interface (CLI) tools which revolve
+around K specifications. These tools cover a broad range of uses, but typically
+fall into one of the following categories:
+
+1.  Transforming K Specs (e.g. compilation, visual presentation)
+2.  Running K Specs (e.g. concrete and symbolic execution)
+3.  Analyzing K Specs (e.g. theorem proving, bounded model checking)
+
+Different tools in the K Framework have different levels of maturity.
+The most well-supported and broadly used tools are the following:
```
nishantjr:
```suggestion
The main *user-facing* K tools include:
```

USER_MANUAL.md
```diff
@@ -1,12 +1,81 @@
...
+
+-   `komile` - the K compiler driver
```
nishantjr:
```suggestion
-   `kompile` - the K compiler driver
```

USER_MANUAL.md
```diff
@@ -1,12 +1,81 @@
...
+
+This user manual is designed to be a tool reference. In particular, it is not
+desgined to be a tutorial on how to write K specifications or to teach the
+logical foundations of K. New K users should consult our dedicated
+[K tutorial](https://kframework.org/k-distribution/k-tutorial/).
+Researchers seeking to learn more about the logic underlying K are encouraged
+to peruse the
+[growing literature on K and matching logic](https://fsl.cs.illinois.edu/projects/pl/index.html).
+We will consider the manual complete when it provides a complete description of
+all well-supported K tools and features.
```
nishantjr:
```suggestion
This user manual is designed to be a tool reference. In particular, it is not
desgined to be a tutorial on how to write K specifications or to teach the
logical foundations of K. New K users should consult our dedicated
[K tutorial](https://kframework.org/k-distribution/k-tutorial/),
or the more language-design oriented [PL tutorial](https://kframework.org/k-distribution/pl-tutorial/).
Researchers seeking to learn more about the logic underlying K are encouraged
to peruse the
[growing literature on K and matching logic](https://fsl.cs.illinois.edu/projects/pl/index.html).
We will consider the manual complete when it provides a complete description of
all well-supported K tools and features.
```

nishantjr (changes requested):

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+
+The K Framework is a programming language and system design toolkit made for
+practioners and researchers alike.
+
+**K For Practioners:**
+*K is a framework for deriving programming languages tools from their semantic
+specifications.*
+
+Typically, programming language tools are developed as follows: for each
+language *L*, a separate team develops tool *T* (e.g. a compiler, interpreter,
```
dwightguth: I don't know that we want to use this type of capital letter style here for something like this. It's better just to use plain English

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+parser, symbolic execution engine, etc). Thus, for *N* different languages and
+*M* different tools, we have *N x M* different teams and implementations.
+
+K approaches the problem differently, by imagining how each of the *M*
```
dwightguth: same here

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+typically fall into one of the following categories:
+
+1.  Transforming K Specs (e.g. compilation)
+2.  Running K Specs (e.g. concrete and symbolic execution)
+3.  Analyzing K Specs (e.g. theorem proving)
+
+The main *user-facing* K tools include:
+
+-   `kompile` - the K compiler driver
+-   `kast` - the stanadlone K parser and abstract syntax tree (AST)
```
dwightguth: let's replace this with `kparse`

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+1.  `syntax` declarations encode the *system primitives*;
+2.  `configuration` declarations encode the *system state*;
+3.  `context` and `rule` declarations encode the *system behavior*.
+
+K sentences are then organized into one or *modules* which are stored in one or
+more *files*. In this scheme, files may *require* other files and modules may
+*import* other modules, giving rise to a hierarchy of files and modules. We
+give an intuitive sketch of the two levels of grouping in the diagram below:
+
+```k
```
dwightguth: this is not k code so the `k` label is not appropriate on this code fence

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+        contexts
+    -   **rule declarations:** define how the system transitions from one state
+        to the next
+
+### K Syntax Overview
+
+Here we provide an overview of the top-level K syntax using a K-style BNF
+notation. Note that this is not a full definition; we focus instead on the
+overall structure of the specification. Syntactic categories which are not
+defined are denoted by names surrounded by curly braces (`{}`). In particular,
```
dwightguth: if we're going to be giving a grammar, we shouldn't make it incomplete like this. It's possible to define the syntax of all of these identifiers and such fairly readily.

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+attributes (`{Attribute}`), a highly flexible mechanism for modifying the
+interpretation of modules and various kinds of sentences.
+
+Due to the highly generic nature of user-defined K syntax, the parsing process
+of K specifications roughly proceeds in three stages, corresponding to the
+three kinds of declarations shown above.
+
+1.  In the first stage, system primitive declarations are gathered and a parser
+    is constructed.
+2.  In the second stage, the system primitive parser is used to parse
```
dwightguth: better to talk about outer and inner parsing as that is a much more endemic nomenclature to the project.

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+    configuration declarations; in our BNF grammar below, we use `{TermBubble}`
+    to describe things which can be parsed using the primitive syntax.
+3.  Finally, a third stage parser is constructed containing both primitive and
+    configuration syntax is constructed which is used to parse contexts,
+    context aliases, and rules. In our BNF grammar below, we use the words
+    `{Bubble}` and `{BubbleWithRewrites}` to describe parsers which contain the
```
dwightguth: this is incorrect; there is only one type of bubble. What may appear inside that bubble is not part of the outer syntax of K; it's part of a separate grammar that is generated from the user-defined syntax.

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+  File         ::= RequiresList ModuleList
+
+  RequiresList ::= "" | Requires RequiresList
+  Requires     ::= RequireCmd {String}
+  RequireCmd   ::= "requires" | "require"
+
+  Module       ::= "module" {ModuleId} AttributeSet ImportList SentenceList "endmodule"
+
+  AttributeSet ::= "" | "[" Attributes "]"
+  Attributes   ::= Attributes "," {Attribute}
```
dwightguth: this is not correct as your grammar here does not allow for the possibility of a sentence with only a single attribute.

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+
+  FunctionNotation   ::= {Id} "(" ParameterList ")"
+  ParameterList      ::= "" | NeParameterList
+  NeParameterList    ::= Parameter | Parameter "," NeParameterList
+  Parameter          ::= OptParamName {SortId}
+  OptParamName       ::= "" | {Id} ":"
+
+  GenericNotation    ::= StringOrSortIdList
+  StringOrSortIdList ::= StringOrSortId | StringOrSortId StringOrSortIdList
+  StringOrSortIdList ::= {String} | {SortId}
```
dwightguth: this is also not correct as you are not allowing for regular expression terminals

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+
+  LexicalDeclaration ::= "syntax" "lexical" {Id} "=" {String}
+
+  AssociativityDeclaration ::= "syntax" Associativity NeIdList
+  NeIdList                 ::= {Id} | {Id} NeIdList
+  Associativity            ::= "left" | "right" | "non-assoc"
+
+  PriorityDeclaration ::= "syntax" "priorities" PriorityList
```
dwightguth: this can also be written as "syntax priority"

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+  PriorityList        ::= NeIdList ">" PriorityList
```
dwightguth: this is also not correct as your grammar does not allow a sentence with only two priority levels

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+                        | NeIdList ">" NeIdList
+
+  ConfigDeclaration ::= "configuration" NeCellDeclList
```
dwightguth: this is not really how configuration parsing works.

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+  StartCellDecl     ::= "<" {CellId} CellAttributeList ">"
+  InnerCellDecl     ::= {TermBubble} | NeCellDeclList
+  EndCellDecl       ::= "</" {CellId} ">"
+  CellAttributeList ::= "" | CellAttribute CellAttributeList
+  CellAttribute     ::= {Id} "=" {String}
+
+  RuleDeclaration   ::= "rule" OptName {BubbleWithRewrites} RequiresCondition EnsuresCondition AttributeSet
+  OptName           ::= "" | Name
+  Name              ::= "[" {Id} "]" ":"
+  RequiresCondition ::= "" | "requires" {Bubble}
```
dwightguth: this is also not how bubbles in rules work

USER_MANUAL.md
```diff
@@ -1,12 +1,355 @@
...
+  EnsuresCondition  ::= "" | "ensures" {Bubble}
+
+  ContextDeclaration      ::= "context" {BubbleWithRewrites} RequiresCondition AttributeSet
+  ContextAliasDeclaration ::= "context" "alias" Name {Bubble} AttributeSet
```
dwightguth: context aliases can have requires clauses.

dwightguth (changes requested):
I have some small low-level concerns with wording, but the biggest problem is that your grammar is simply not a good description of how outer parsing works, even if we set aside the individual minor bugs in how it's defined.

sskeirik:
@dwightguth Thanks for your review comments. Re: bubble parsing: I don't fully understand how this works, as you saw, what I wrote was my best attempt. Two obvious possible paths forward on this:

1.  throw out the grammar section and keep the other stuff
2.  try and fix the grammar section by fixing the bubble description

If we want to pursue option (2), I would appreciate some help on:

1.  pointers on where I can find the syntax of valid K identifiers for sorts, cells, record selectors, modules, etc.
2.  help understanding how bubble parsing works

dwightguth:
Generally speaking all the syntax for the outer parsing can be found in `kernel/src/main/javacc/Outer.jj` and all the syntax for inner parsing can be found in `k-distribution/include/kframework/builtin/kast.k` and `kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java`. It shouldn't be too hard to find the syntaxes for the identifiers since these should all be inside Outer.jj or kast.k in definitions of formal grammars.

Bubble parsing is more complex. Right now, here is (roughly) how it works:

A bubble starts immediately following the `rule`, `configuration`, `context`, `context alias`, or `claim` keywords and continues until the next occurrence of the `syntax`, `endmodule`, `rule`, `claim`, `configuration`, or `context` keywords. A bubble can contain any characters; parsing of that content happens later.

The first pass of bubble parsing takes place inside the outer parser itself; we look for a rule label in the form of the `"(?s)\\s*\\[\\s*([^\\[\\]\\_\\ \\n\\r\\t]+)\\s*\\]\\s*:\\s*(.*)"` Java regular expression, and extract it out from the bubble and put it in the attributes of the sentence as the `label` attribute. We also try to find a spot in the sentence starting with a `[` at which it is possible to derive a set of attributes immediately followed by the end of the bubble. If we find one, that becomes the attributes for that sentence, and everything left over after this first pass of bubble parsing becomes the final bubble, ready for inner parsing. If we don't find one, the bubble extends to the very end and the sentence has no explicit attributes.

Once this pass of bubble parsing is complete, we have the attributes of the sentence, the type of the sentence, and a string containing the body of the sentence itself. We then generate a scanner and parser in order to parse that bubble, and everything inside gets parsed as a term of sort `#RuleContent`. The grammar used to generate this scanner and parser is augmented with the syntax of configurations (as defined by the `CONFIG-CELLS` module in the case of configurations and the `RULE-CELLS` modules in the case of contexts, context aliases, rules, and claims) and the syntax of K (as defined by the `K` module). We also use the `#Layout` sort to decide what characters the scanner should treat as whitespace or comments. If the `#Layout` sort is not defined by the user, we use the `DEFAULT-LAYOUT` module.

One final component gets mixed into the grammar generation, and that is the code mentioned above in `RuleGrammarGenerator.java`. This includes various pieces of logic that munge the grammar or add certain productions that cannot be easily expressed in K, like casts or sort predicates. RuleGrammarGenerator also takes care of any necessary internal munging in order to make the grammar into a form that our implementation can handle; you shouldn't need to worry about those details when describing how parsing works though.

dwightguth:
All that being said, I have no objection if you want to remove that part of the manual and handle it later.

USER_MANUAL.md
sskeirik: I rewrote this section to use plain english.

USER_MANUAL.md
sskeirik: I rewrote this section to use plain english.

USER_MANUAL.md
sskeirik: Done.

USER_MANUAL.md
sskeirik: Removed.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

USER_MANUAL.md
sskeirik: I removed this section from this PR because it needs more polish.

sskeirik:
@dwightguth I believe I have addressed your review comments on wording and deleted the syntax section because it needs more polish. I would appreciate a re-review. Thanks.

dwightguth (approved):

USER_MANUAL.md
```diff
@@ -1,12 +1,242 @@
...
+new language, the same basic tools and patterns are re-implemented again and
+again.
+
+K approaches the problem differently, by imagining how different tools could be
+derived generically from a common specification. In other words, for any
+language, a language-generic tool is a function that takes a language spec as
+input and produces a tool as output. The end result is that the exercise of
+designing new languages and their associated tooling is now reduced to
+developing a single language specfication from which we derive our tooling *for
+free*.
```
nishantjr: It's a little confusing that the tools are both "derived generically from a common specification" and that they take the "language spec as input".

```suggestion
K approaches the problem differently -- it generates each of these tools from a single language specification.
The work of programming language design and tool implementation are made separate concerns.
The end result is that the exercise of
designing new languages and their associated tooling is now reduced to
developing a single language specification from which we derive our tooling *for
free*.
```

USER_MANUAL.md
```diff
@@ -1,12 +1,242 @@
...
+|             |                                              |
+|             |       +---------+                            |
+|  K Term ----+-------|  krun   |--> K Term                  |
+|             |       +---------+                            |
+|             |                                              |
+|             |       +---------+                            |
+|  K Claims --+-------| kprove  |--> K Claims                |
+|                     +---------+                            |
+|                                                            |
++============================================================+
```
nishantjr, line 211: Kore has not been introduced yet, also `kparse` and `krun` take Programs as input.

```suggestion
 K Compilation Process
+============================================================+
|                     +---------+                            |
|  K Specification ---| kompile |--> Formal Specification +  |
|                     +---------+                         |  |
+=========================================================|==+
                                                          |
 K Execution Process                                      |
+=========================================================|==+
|                                                         |  |
|             +-------------------------------------------+  |
|             |                                              |
|             |       +---------+                            |
|  Program ---+-------| kparse  |--> K Term                  |
|             |       +---------+                            |
|             |                                              |
|             |       +---------+                            |
|  Program ---+-------|  krun   |--> K Term                  |
|             |       +---------+                            |
|             |                                              |
|             |       +---------+                            |
|  K Claims --+-------| kprove  |--> K Claims                |
|                     +---------+                            |
|                                                            |
+============================================================+
```

USER_MANUAL.md
```diff
@@ -1,12 +1,242 @@
...
+```
+
+where:
+
+-   process outlines are denoted by boxes with double-lined borders
+-   programs are denoted by boxes with single-lined borders
```
nishantjr: I think we can avoid confusion between programs in the programming language, and executables.

```suggestion
-   executables are denoted by boxes with single-lined borders
```

nishantjr (approved):
Looks good.

PR #2114, opened 2021-08-05
Module signatures

### LINKED ISSUE #1658: Module signatures supporting public/private syntax and imports
This would enable more incremental parsing.

We want to be able to:

-   Declare that a piece of syntax is `private`, so it cannot be used in either (i) modules which import the module it's declared in, or (ii) programs.
-   Declare that you are importing a module `A` into module `B` in a `private` manner, meaning that the module `B` may use the syntax of module `A`, but any module importing `B` may not use the syntax of module `A`.

This will cause problems with the existing lexer generation, because currently we only generate a single lexer for the entire definition. Roughly we can start with "the lexer should ignore the public/private status, while the parser should respect them".

For reference, we should profile the processes of (i) lexer generation and (ii) grammar generation. Currently we think that (i) takes on the order of seconds, and (ii) takes on the order of milliseconds (for "normal" definitions).

dwightguth:


nishantjr:
@dwightguth Could we add a test for this case:

```
module C1 [private]
  imports B
endmodule
```

I think this should be equivalent to:

```
module C1 [private]
  imports [private] B
endmodule
```

and we also need syntax for:

```
module C1 [private]
  imports public B
endmodule
```

dwightguth:
Currently my implementation treats `imports` as always public and `imports private` as always private. I guess we could make it so that we had both `imports private` and `imports public` and that `imports` was treated as having the same visibility as the module itself. I tend to agree that's more intuitive. On the other hand, imports, unlike other syntax, are always at the top of the module, so you would be unlikely to forget and give an import the wrong accessibility, which is the reason why we introduced the module attribute in the first place. I'm open to the change, but I'd like to hear what @radumereuta and @ehildenb think the intuitive behavior here is as well.

k-distribution/tests/regression-new/checks/signature.k.out
```diff
@@ -0,0 +1,28 @@
+[Error] Inner Parser: Parse error: unexpected token 'foo'.
+	Source(signature.k)
+	Location(24,8,24,11)
+[Error] Inner Parser: Parse error: unexpected token 'fu'.
+	Source(signature.k)
+	Location(26,8,26,10)
+[Error] Inner Parser: Parse error: unexpected token 'baz'.
+	Source(signature.k)
+	Location(28,8,28,11)
```
ehildenb, line 9: Can we make these error messages more informative easily? Something like also saying "Symbol 'baz' is part of the private signature of module 'A', and so cannot be used directly in module 'B'"?

dwightguth: Not easily. We would have to have some way of recognizing that it parsed as a "baz" and then rejecting that parse, which is going to be a lot less efficient than simply removing it from the grammar entirely.

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
```diff
@@ -129,9 +132,10 @@ public Module getRuleGrammar(Module mod) {
...
         // import CONFIG-CELLS in order to parse cells specific to configurations
         Module newM = new Module( mod.name() + "-" + CONFIG_CELLS
-                                , Set(mod, baseK.getModule(K).get(), baseK.getModule(CONFIG_CELLS).get(), baseK.getModule(DEFAULT_LAYOUT).get())
-                                , Set()
-                                , Att()
+                                , (scala.collection.Set<Module>) mod.publicImports().$bar(Set(baseK.getModule(K).get(), baseK.getModule(CONFIG_CELLS).get(), baseK.getModule(DEFAULT_LAYOUT).get()))
+                                , mod.privateImports()
+                                , mod.localSentences()
+                                , mod.att()
                                 );
```
ehildenb: This pattern than `getConfigGrammar` uses seems like it could be extracted into a single function `getGrammar`, which takes as input `Module mod, String cellName`, then `getConfigGrammar(Module mod) => getGramma(mod, mod.name() + "-" + CONFIG_CELLS)`, for instance. That would deduplicate `getRuleGrammar` and `getConfigGrammar`

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
```diff
@@ -179,15 +187,19 @@ public static boolean isParserSort(Sort s) {
...
     }
 
     public static ParseInModule getCombinedGrammar(Module mod, boolean strict, boolean timing, FileUtil files) {
-      return getCombinedGrammar(mod, strict, timing, false, files);
+      return getCombinedGrammar(mod, strict, timing, false, false, files);
+    }
+
+    public static ParseInModule getCombinedGrammar(Module mod, boolean strict, boolean timing, FileUtil files, boolean forGlobalScanner) {
+      return getCombinedGrammar(mod, strict, timing, false, forGlobalScanner, files);
     }
```
ehildenb: Can we check which of these overloads are used directly and remove the ones that aren't (if there are any)? Maybe we don't need the one without `boolean forGlobalScanner` anymore?

kernel/src/main/java/org/kframework/parser/inner/kernel/Grammar.java
```diff
@@ -58,6 +58,14 @@
     /** The set of "root" NonTerminals */
     private BiMap<String, NonTerminal> startNonTerminals = HashBiMap.create();
 
+    public NonTerminal nullNT() {
+      return get("_Null");
+    }
+
+    public Grammar() {
+      add(new NonTerminal("_Null"));
+    }
```
ehildenb, line 67: We're sure these "special" nonterminals cannot show up anywhere in a user's grammar?

dwightguth: I'm sure. Sort names can't begin with an underscore.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -56,12 +56,14 @@ object FlatModule {
...
+        val f = (i: Import) => memoization.getOrElse(i.name
           // if can't find the Module in memoization, build a new one
           , toModuleRec(allModules.find(f => f.name.equals(i.name))
               .getOrElse(throw KEMException.compilerError("Could not find module: " + i.name, i))
-            , visitedModules :+ m)))
-        val newM = new Module(m.name, newImports, m.localSentences, m.att)
+            , visitedModules :+ m))
+        val newPublicImports = m.imports.filter(_.isPublic).map(f)
+        val newPrivateImports = m.imports.filter(!_.isPublic).map(f)
+        val newM = new Module(m.name, newPublicImports, newPrivateImports, m.localSentences, m.att)
```
ehildenb, line 66: This construction pattern (split into public/private sentences, then build new Module type) seems common. Can we extract a "smart constructor" for `Module` which encapsulates that logic (so you just pass in the full set of sentences, and it does the split and constructs the module with the split sentences)?

k-distribution/tests/regression-new/checks/signature.k.out
radumereuta: I agree with Dwight. One benefit of hiding syntax is that the parser doesn't have to handle those cases, which might lead to fewer ambiguities and faster parsing.
It's doable, but it adds additional complexity and makes parsing equally slow.
Then raises the question, if you are doing it for imported private syntax, why not do it for parallel modules...

nishantjr:
@dwightguth Could you also add documentation to the manual please.

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
radumereuta: I'm confused about `forGlobalScanner` as well. Why is it needed?

radumereuta (changes requested):
This is very large, and I will have to go through it again in more detail. But for now, I found a test that fails to pass:
```
module TEST
  syntax Foo ::= foo() // <- B may not use this symbol

  syntax Foo ::= bam() [public] // <- B may use this symbol

  syntax Foo ::= fu() [private] // <- B may not use this symbol

  syntax Foo ::= "bar" [public, token] // <- B may use this token
               | r"ba." [token, prec(3)]      // <- B may not use this token

  rule foo() => .K
  rule bam() => .K
  rule fu() => .K
  rule bar => .K
  rule baz => .K
endmodule
```
```
$ kompile test.k

Error:
  module 'TEST':
definition.
Examine output to see errors.
```

radumereuta:
The issue fails on master also. Very strange.
If this ends up being outside the scope of this PR, please add a test that checks for the positive case, where you have public/private syntax, and compilation passes.

radumereuta:
> Currently my implementation treats `imports` as always public and `imports private` as always private. I guess we could make it so that we had both `imports private` and `imports public` and that `imports` was treated as having the same visibility as the module itself. I tend to agree that's more intuitive. On the other hand, imports, unlike other syntax, are always at the top of the module, so you would be unlikely to forget and give an import the wrong accessibility, which is the reason why we introduced the module attribute in the first place. I'm open to the change, but I'd like to hear what @radumereuta and @ehildenb think the intuitive behavior here is as well.

Sounds like a good idea. It gives you more control over syntax visibility. If it's not a big change, I'm in favor.

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
dwightguth: All of these overloads are used somewhere. In response to Radu's question, it's needed so that the scanner we generate to parse rules contains /all/ of the syntax and not just the syntax visible to the main module.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
dwightguth: In point of fact, while there are three places in this PR that do something similar like this, they are enough different that they would not benefit from refactoring out a common function.

dwightguth:
@radumereuta, please create a separate issue for the MatchError exception you encountered, as it is outside the scope of this PR.

dwightguth:
I have not yet written the manual segment, but the code itself has had all of the issues everyone raised addressed in some form or other. I still will want someone to review the documentation once I write it, but the code ought to be ready for re-review now.

nishantjr:
Looks good. Thanks!

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
radumereuta: If I understand correctly, you are trying to optimize the case when you are parsing `STDOUT-STREAM` and `STDIN-STREAM` by reducing the number of productions included in the scanner. IMO the possible gains in speed don't merit adding a new parameter here. So I would like either:
- remove the option. The benefits don't outweigh the added code complexity.
- add inline documentation explaining the functionality. I spent more than half an hour trying to understand why this is needed and how it works.

If I'm getting this wrong, please correct me.

radumereuta (changes requested):
2 changes
- `forGlobalScanner`  (see my other comment)
- please add a test with the new public/private syntax that kompiles w/o errors (a check with the empty output would be sufficient)

Question: you were talking about changes to domains.md to include public/private syntax. I don't see any changes to that file. Am I remembering wrong?

Other than that looks pretty good.

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
dwightguth: No see, if the main scanner for rules doesn't include private syntax, then we will need to generate a new scanner for every module containing any private syntax, which will be very expensive.

nishantjr:
> Question: you were talking about changes to domains.md to include public/private syntax. I don't see any changes to that file. Am I remembering wrong?

@radumereuta That would be a non-backward compatible change. We should perhaps do that when we bump major versions?

dwightguth:
@radumereuta the changes to domains.md are in a separate pr in order to simplify things, because as Nishant says, they are not backwards compatible. I have commented on your comments about the scanner, but I can't make that change without causing serious problems. I will add the test you ask for though.

kernel/src/main/java/org/kframework/parser/inner/generator/RuleGrammarGenerator.java
radumereuta: I understand this. Why isn't this the default for every other scanner? Why do you need to make it an option?

dwightguth: This is the only place where scanners are created that are used non-modularly for other modules. Every other place where a ParseInModule is being constructed, we either don't use the scanner that comes with that object, or else we use that scanner only to scan that module, in which case we don't /want/ to include the private syntax in the scanner.

radumereuta: That sounds reasonable. Please add that explanation as a code comment. It's going to be useful in the future.

USER_MANUAL.md
```diff
@@ -537,6 +588,24 @@ automatically. It will also automatically sort associative-commutative
...
+module in which it is declared; it is not visible from modules that import
+that module.
+
+You can also change the default visibility of a module with the `private`
+attribute, when it is placed directly on a module. A module with the `private`
+attribute has all syntax `private` by default; this can be overridden on
+specific sentences with the `public` attribute.
+
+Note that the `private` module attribute also changes the default visiblity
+of imports; please refer to the appropriate section elsewhere in the manual
```
radumereuta, line 606: Don't you have a link or reference instead of just _elsewhere_?
Sounds dismissive.

radumereuta (approved):
Just one comment. Otherwise looks pretty good. Let's start testing it.

USER_MANUAL.md
dwightguth: I'm open to including a link if you know how...

USER_MANUAL.md
```diff
@@ -537,6 +588,24 @@ automatically. It will also automatically sort associative-commutative
...
+for more details.
+
```
nishantjr, line 630: I'm not sure how to suggest code blocks, but I think we should add an example.

```suggestion
We give an example of possible usage below:

module WIDGET
  syntax Widget ::= gadget()
endmodule

module FOO [private]
  imports WIDGET                         // <- B may not use `gadget()`
  syntax Foo ::= foo()                   // <- B may not use this symbol
  syntax Foo ::= bam() [public]          // <- B may use this symbol
  syntax Foo ::= fu() [private]          // <- B may not use this symbol
  syntax Foo ::= "bar" [public, token]   // <- B may use this token
               | r"ba." [token, prec(3)] // <- B may not use this token
endmodule

module B
  imports private FOO                    // <- B may use the _public_ syntax of A, but C may not
  syntax Bar ::= a() [public]            // <- C may use this symbol
               | b() [private]           // <- C may not use this symbol
endmodule

module C
  imports B
  imports WIDGET                         // C may now use WIDGET's public interface
endmodule

```

ehildenb: I agree that we should add an example like this.

PR #2116, opened 2021-08-05
Verbose k prove

radumereuta:
This changes a bit the way `--profile-rule-parsing` generates timing.log files.
- collect everything into a single threadsafe queue
- print the collected messages into a single file specified at command line
- added more information to `kprovex -v`
- changed the timings from seconds with decimals to milliseconds (since I find them more readable)


ehildenb:
Make it so that `--profile-rule-parsing` takes a file parameter, and use that file for saving the info. That will make this fit better into some rule-parsing profile helpers I have written as well.

radumereuta:
Slightly updated how we print when `-v`. I've indented `New scanner` to hint that the times are included in the next print.
```
$kompile test.k -v --debug --backend haskell --profile-rule-parsing test.log --gen-bison-parser
Parse command line options                                   =  0.012s
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/prelude.md)
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/kast.md)
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/domains.md)
Importing: Source(/home/radu/work/test/test.k)
Outer parsing [68 modules]                                   =  0.744s
  New scanner: DEFAULT-CONFIGURATION-CONFIG-CELLS            =  0.104s
  New scanner: STDOUT-STREAM-CONFIG-CELLS                    =  0.210s
  New scanner: STDIN-STREAM-CONFIG-CELLS                     =  0.215s
Parse configurations [3/3 declarations]                      =  0.518s
  New scanner: TEST-RULE-CELLS                               =  0.149s
  New scanner: STRING-KORE-RULE-CELLS                        =  0.159s
  New scanner: STRING-BUFFER-IN-K-RULE-CELLS                 =  0.177s
  New scanner: STRING-COMMON-RULE-CELLS                      =  0.206s
  New scanner: TEST-RULE-CELLS                               =  0.247s
  New scanner: K-IO-RULE-CELLS                               =  0.288s
  New scanner: STDOUT-STREAM-RULE-CELLS                      =  0.294s
  New scanner: STDIN-STREAM-RULE-CELLS                       =  0.301s
Parse rules [112/112 rules]                                  =  0.603s
Validate definition                                          =  0.072s
Apply compile pipeline                                       =  0.332s
  New scanner: TEST-PROGRAM-GRAMMAR                          =  0.071s
  New Bison parser: TEST-PROGRAM-GRAMMAR                     =  0.510s
Kompile to kore                                              =  0.514s
Save to disk                                                 =  0.091s
Backend                                                      =  0.334s
Total                                                        =  3.221s
```
```
$kprovex spec.k --debug -v --profile-rule-parsing spec.log
Parse command line options                                   =  0.010s
Importing: Source(/home/radu/work/test/spec.k)
Importing: Source(/home/radu/work/test/test.k)
Outer parsing [72 modules]                                   =  0.396s
  New scanner: TEST-RULE-CELLS                               =  0.145s
  New scanner: SPEC-RULE-CELLS                               =  0.129s
Parse spec modules [1/1 rules]                               =  0.536s
Apply prover steps                                           =  0.180s
Executing [kore-exec, /home/radu/work/test/./test-kompiled/definition.kore, --module, TEST, --prove, /home/radu/work/test/./.kprove-2021-08-23-23-48-05-183-23d8e592-4f1b-4ca1-8d52-7a9a3c1b3979/spec.kore, --spec-module, SPEC, --output, /home/radu/work/test/./.kprove-2021-08-23-23-48-05-183-23d8e592-4f1b-4ca1-8d52-7a9a3c1b3979/result.kore]
Executing command: kore-exec /home/radu/work/test/./test-kompiled/definition.kore --module TEST --prove /home/radu/work/test/./.kprove-2021-08-23-23-48-05-183-23d8e592-4f1b-4ca1-8d52-7a9a3c1b3979/spec.kore --spec-module SPEC --output /home/radu/work/test/./.kprove-2021-08-23-23-48-05-183-23d8e592-4f1b-4ca1-8d52-7a9a3c1b3979/result.kore
kore-exec: [292220] Warning (WarnStuckClaimState):
    The claim cannot be rewritten further, and the claimed implication is not valid. /home/radu/work/test/spec.k:8:11-8:59
Backend                                                      =  0.462s
<k>
  doneLemma ( 1 ) ~> _DotVar1 ~> .
</k>
Cleanup                                                      =  0.032s
Total                                                        =  1.616s
```
Here is the output from `--profile-rule-parsing`. I find this alignment much easier to read and spot odd rules out. I've also included separate parse and type inference times. All in milliseconds.
```
   17 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:1052   parse:   6 typeInf:   1
   15 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:1053   parse:   1 typeInf:  12
   23 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:1055   parse:   6 typeInf:  10
   23 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:1056   parse:  16 typeInf:   2
   15 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:1059   parse:   2 typeInf:  10
```

ehildenb:
We need to make sure that the cost associated with each line is actually accurate. For example, is the cost associated with `  New scanner: ...` correctly reporting _just_ the time for producing that new scanner, or does it also include rule parsing time?

We need to keep track of these costs separately if we want to produce more informative output.

kernel/src/main/java/org/kframework/utils/TimingCollector.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Deque;
+import java.util.List;
+import java.util.concurrent.ConcurrentLinkedDeque;
+
+@ThreadSafe
+public class TimingCollector {
+    final Deque<Tuple2<String, String>> messages = new ConcurrentLinkedDeque<>();
```
ehildenb: We need to check if there is any chance this will block or not? @radumereuta please link us to the documentation.

ehildenb:
Let's switch back to a single file per thread. It should delete all existing timing files, then still collect to separate files, then combine at the end into the file supplied by the user.

dwightguth:
I can't find the exact stack overflow thread that I found when I was searching however long ago, but the essential idea is that if you have time-sensitive information that you are gathering in a bunch of highly parallel threads, and you need to collect it into a single record, but the order of records within the log doesn't matter, the best solution is simply to collect it into one record per thread, and then combine all the records after the fact. This will perform even better than a lock-free concurrent data structure, because even with a lock-free concurrent data structure, you are going to be paying the cost associated with whatever primitive concurrency operation the data structure uses to avoid needing locks. Because some threads will be spending time on these concurrent operations, the other threads that are parsing rules during that time period will be affected by the load on the other threads. This will skew the time being recorded, albeit less than if you were using a file lock.

If the main concern is that we simply want to be able to have access to a single shared record that is automatically compiled by the compiler, the easiest way to do this is simply to combine the files together for you after the fact. It could be as easy as running the following bash script from kompile after parsing finishes:

```
cat timing*.log | sort -k2 -n > "$1"
rm timing*.log
```

SchmErik:
> I can't find the exact stack overflow thread that I found when I was searching however long ago, but the essential idea is that if you have time-sensitive information that you are gathering in a bunch of highly parallel threads, and you need to collect it into a single record, but the order of records within the log doesn't matter, the best solution is simply to collect it into one record per thread, and then combine all the records after the fact. This will perform even better than a lock-free concurrent data structure, because even with a lock-free concurrent data structure, you are going to be paying the cost associated with whatever primitive concurrency operation the data structure uses to avoid needing locks. Because some threads will be spending time on these concurrent operations, the other threads that are parsing rules during that time period will be affected by the load on the other threads. This will skew the time being recorded, albeit less than if you were using a file lock.

I agree. The primitive concurrency operations can get really expensive. I don't think it's in our scope to find out how these data structures are using java bytecode instructions or how jvm map to x86-64 instructions. Hypothetically, these lock-free data structures could be using x86 atomic operations and those instructions do things like invalidate caches on other processors which creates a lot of noise on the bus. In this environment, I think the solution to use multiple files would give the most accurate measurements because it has the least amount of opportunities for threads to contend for resources.

PR #2123, opened 2021-08-10
fix krun script to use module specified by module attribute for kparse

dwightguth:
by default

Currently, if you specify the`parser` attribute for a configuration variable, krun will not actually store that module and use it to perform parsing, which leads to a performance regression because we use the jit parser even though a bison parser exists. The solution is simply to record the module we are supposed to use for parsing and use it by default when we call kparse.

### DIFF
--- a/k-distribution/src/main/scripts/bin/krun
+++ b/k-distribution/src/main/scripts/bin/krun
@@ -478,7 +478,9 @@ else
       fi
     fi
     indirectName="declaredConfigVar_$name"
+    indirectModuleName="declaredConfigVarModule_$name"
     sortName=${!indirectName}
+    moduleName=${!indirectModuleName:-unset}
     if [ -z "${!parser_name+unset}" ]; then
       if [ "$name" = "PGM" ]; then
         if $hasArgv; then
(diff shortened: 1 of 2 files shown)


k-distribution/src/main/scripts/bin/krun
```diff
@@ -478,16 +478,20 @@ else
...
+    moduleName=${!indirectModuleName:-unset}
     if [ -z "${!parser_name+unset}" ]; then
       if [ "$name" = "PGM" ]; then
         if $hasArgv; then
           parser=(kparse --directory "$dir")
         else
           parser=(kparse --directory "$dir" --module "$mainModuleName")
         fi
-      else
+      elif [ -z "${!indirectModuleName+unset}" ]; then
```
ehildenb, line 491: AFAICT, this check will never pass, because we're always defining `indirectModuleName` above. But I also don't understand the `${!...}` syntax..

radumereuta:
lgtm. I'll let Everett approve it when he's satisfied with his questions.
I see that I also added one such line, but I think I copy/pasted another instance and didn't pay attention to `!`.

k-distribution/src/main/scripts/bin/krun
dwightguth: It's an indirect variable lookup. We're not looking up indirectModuleName, we're expanding it as a string and looking up the variable with the name its value expands to. That variable might not be set if the user didn't specify the parser attribute for a particular configuration variable.

ehildenb (approved):

PR #2127, opened 2021-08-11
Add attribute reference to the user manual

sskeirik:


dwightguth (commented):
I'm fine with us describing the syntax of attributes themselves, but the details about each individual type of attribute are broken out into their own section elsewhere in the manual and this summary doesn't really appropriately address that.

sskeirik:
@dwightguth Thanks for your feedback. How could this PR be improved? If each attribute description linked to the section where the details on that attribute are described, would that be an improvement?

Just for clarify: my intention is to provide a quick reference of different attributes with summaries. There are several possible usages:

1. (for newcomers) seeing an overview of all attributes in one place, to get a flavor for what is possible
2. (for experts) helping recall some attribute that you heard about/saw before, but don't remember how to spell it

dwightguth (approved):

PR #2130, opened 2021-08-13
Lesson 19: debugging

dwightguth:


k-distribution/k-tutorial/1_basic/19_debugging/README.md
```diff
@@ -1 +1,359 @@
...
+
+Suppose we try to run the program `foo(bar(baz2()))`. It is obvious from this
+example why the rule in this definition will not apply. However, in practice,
+such cases are not always obvious. You might look at a rule and not immediately
+spot why it didn't apply on a particular term. For this reason, it can be
+useful to get the debugger to provide a log about how it tried to match that
+term. You can do this with the `k match` command. If you are stopped after
+having run `k start` or `k step`, you can obtain this log by running the
+command `k match MODULE.label subject` for a particular top-level rule label.
+
```
hjorthjort, line 322: I'm not 100% sure I understand this functionality. So if I'm anywhere in the execution, I can write `k match LESSON-19-D.baz subject` and see whether or not my current configuration matches a given rule? Or does it only work when the program is stuck?

Also, for clarity, I'd add the prompt here at the top of the output to show an example of what exactly the input is:
```
(gdb) k match LESSON-19-D.baz subject
Subject:
baz2 ( )
does not match pattern:
baz ( )
```

hjorthjort (changes requested):

hjorthjort (approved):

PR #2133, opened 2021-08-16
Remove unneeded provider

radumereuta:
Copy-paste error when I added kprovex.
`extraConcreteRuleLabels` is used in kprove to enhance the kompile pipeline. But when this step is applied in kompile is not needed in kprovex. It was also wrong because it was trying to extract the info from KProveOptions instead of KompileOptions.

Discovered this when I was working with @SchmErik. Now when you generate the dependency graph, you should see the KProveOptions instance disappear.


SchmErik:
I checked out your pull request and I see that the generated PDF from the dot file no longer contains the KproveOptions when using kompile.

SchmErik (approved):

dwightguth (approved):

PR #2138, opened 2021-08-17
Refactor Module class to use Import class

dwightguth:
This is a preliminary refactoring in preparation for adding partial imports that extracts out metadata about each import in a module into its own class in the AST. This will help us with the task of adding additional metadata about each import indicating which sentences are imported.


radumereuta (approved):
lgtm

PR #2139, opened 2021-08-17
update K deploy pipeline for Debian Bullseye

dwightguth:
Tested Debian Bullseye code against CI.


package/debian/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG LLVM_VERSION=8
 ENV TZ America/Chicago
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN if [ "$BASE_IMAGE" = "debian:buster" ]; then echo "Enabling backports..."; echo "deb http://deb.debian.org/debian buster-backports main" > /etc/apt/sources.list.d/buster-backports.list; fi
+# RUN if [ "$BASE_IMAGE" = "debian:bullseye" ]; then echo "Enabling backports..."; echo "deb http://deb.debian.org/debian bullseye-backports main" > /etc/apt/sources.list.d/bullseye-backports.list; fi
```
ehildenb, line 10: Can we just delete this line instead? We will have it in th eGit history.

ehildenb:
Is all the relevant documentation updated that we require LLVM-10 or greater now?

ehildenb (approved):

package/debian/Dockerfile
dwightguth: I want to leave it as a comment because I am anticipating needing to uncomment it sometime prior to Debian 12 being released. It will be easier to revive if we don't have to go digging around in multiple-year-old git history to find it again.

dwightguth:
I haven't updated the documentation about the minimum required version yet because we haven't yet agreed that llvm 10 should be the minimum required version. I asked about it on Slack but only Erik responded. I'd rather make that a separate change though that goes through when we actually stop supporting llvm 8

Issue #2115, opened 2021-08-05
Importing a set of sentences from a module


dwightguth:
We want to improve the module signatures feature by allowing the type of "import foo from BAR` syntax present in most other import systems. Here are three possible syntaxes for consideration:

1. Sentence label

```
module FOO
  syntax [foo]: Foo ::= foo(Foo) 
  syntax [bar]: Foo ::= foo(Bar)
endmodule

module BAR
  imports FOO.foo
  imports FOO.foo as bar(_)
endmodule
```

Potential problems: you can't import something without help from someone who can modify the module being imported from. You also have trouble with sentence labels and pipes mixing.

2. Sentence tag

```
module FOO
  syntax Foo ::= foo(Foo) [tag(foo)]
  syntax Foo ::= foo(Bar) [tag(bar)]
endmodule

module BAR
  imports FOO.foo
  imports FOO.foo as bar(_)
endmodule
```

Basically the same but uses the tag attribute instead of the sentence label. Allows multiple sentences to be given the same tag, which could introduce complexity when it comes to renaming.

3. Explicit syntax

```
module FOO
  syntax Foo ::= foo(Foo)
  syntax Foo ::= foo(Bar)
endmodule

module BAR
  imports 
    syntax Foo ::= foo(Foo)
    syntax Bar ::= foo(Bar) => bar(_)
  from FOO
endmodule
```

Requires you to basically redeclare what you import, but is the most flexible.

nishantjr:
# Use cases

## Allow mixing program syntax with semantics code

K definitions are routinely split into modules for program syntax and semantics.
For readability, it may be nicer program grammar close to the semantic rules that define them.
While For simpler semantics this may be achieved through the `public` and `private` attributes,
this does not work with module hierarchies with more than two levels (such as any semantics that uses proofs).

```
module IMP-SYNTAX
  imports IMP.program
endmodule

module IMP
  syntax [program] AExp  ::= Int | Id | "(" AExp ")" [bracket]

  syntax [program] AExp ::= AExp "/" AExp [left, strict]
  rule I1 / I2 => I1 /Int I2  requires I2 =/=Int 0

  syntax AExp ::= "#undefined"
  rule I1 / I2 => #undefined   // #undefined may not be used in program syntax
                               // but may be used in proofs etc.

  syntax [program] AExp ::= AExp "+" AExp [left, strict]
  rule I1 + I2 => I1 +Int I2

[...]

PR #2144, opened 2021-08-18
`imports FOO.bar`

dwightguth:
This does not contain any syntax for renaming yet; I am simply focusing on the task of importing specific syntax by its tag.


kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -90,7 +91,7 @@ public FlatModule toFlatModule(Module m) {
     }
 
     public org.kframework.definition.FlatImport apply(Import imp) {
-        return org.kframework.definition.FlatImport.apply(imp.getName(), imp.isPublic(), convertAttributes(imp));
+        return org.kframework.definition.FlatImport.apply(imp.getName(), imp.isPublic(), imp.getTag() == null ? immutable(Optional.empty()) : immutable(Optional.of(Tag(imp.getTag()))), convertAttributes(imp));
```
radumereuta: Line too long.

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
```diff
@@ -147,7 +149,7 @@ public static FlatModule toFlatModule(JsonObject data) throws IOException {
 
         JsonArray jsonimports = data.getJsonArray("imports");
         Set<FlatImport> imports = new HashSet<>();
-        jsonimports.getValuesAs(JsonObject.class).forEach(i -> imports.add(FlatImport.apply(i.getString("name"), i.getBoolean("public"), Att.empty())));
+        jsonimports.getValuesAs(JsonObject.class).forEach(i -> imports.add(FlatImport.apply(i.getString("name"), i.getBoolean("public"), immutable(Optional.ofNullable(i.getString("tag")).map(s -> Tag(s))), Att.empty())));
```
radumereuta: Line too long.

radumereuta (changes requested):
This could use some tests.
Some JUnit tests in `OuterParsingTests` for the new syntax and two tests in
`regression-new/checks` one with an error, and one that passes.
I will look into the semantics of this PR tomorrow when I have more energy.

nishantjr:
More tests, please, and documentation too.

radumereuta:
This is part of #2115

k-distribution/tests/regression-new/checks/partialImport.k
```diff
@@ -0,0 +1,21 @@
+module PARTIALIMPORT
+  imports A
+
+  rule foo() => .K
+  rule bar() => .K
+  rule alsoAFoo() => .K
```
radumereuta, line 7: Can you also add:
```
rule baz() => .K // should fail because we're not importing transitively
rule boo() => .K // should fail because of private import
```
and define it in module C? Just to make sure we're not including C transitively.
And then add a
```
module D [private]
  syntax Boo ::= boo()
endmodule
```
And import this in A.

USER_MANUAL.md
```diff
@@ -275,6 +275,16 @@ of the import with the `public` or `private` keyword immediately prior to the
 module name. A module imported privately does not export its syntax to modules
 that import the module doing the import.
 
+An import statement can also choose to explicitly hide some or all of the
+syntax being imported. If an import's module name is followed by a `.`
+character followed by an attribute name or klabel, then only sentences that
+have that attribute name or productions that have that klabel attribute will
+be imported by that declaration. For example, if I say `imports FOO.bar`, then
+only productiosn whose klabel attribute is `bar` or whose attribute list
```
radumereuta:
```suggestion
only production whose klabel attribute is `bar` or whose attribute list
```

radumereuta (changes requested):
Overall pretty good. Just the two small changes.

k-distribution/tests/regression-new/checks/partialImport.k
dwightguth: If you have a suggested code change, please make it as a suggestion; what you wrote here is ambiguous enough that I can't actually tell what you meant.

radumereuta (approved):

PR #2147, opened 2021-08-19
Lesson 20: Backends

dwightguth:


k-distribution/k-tutorial/1_basic/20_backends/README.md
```diff
@@ -1 +1,81 @@
...
+
+  rule I => I +Int 1
+    requires I <Int 100
+endmodule
+```
+
+Previously we compiled this definition using the LLVM Backend, but if we
+instead execute the command `kompile lesson-20.k --backend haskell`, we
+will get an interpreter for this K definition that is implemented in Haskell
+instead. Unlike the default LLVM Backend, the Haskell Backend is not a
```
ehildenb, line 55: I guess this isn't technically correct, but you clarify it in the next sentence. I would say something more like "will get the needed Kore files for the Haskell Backend to symbolically execute your language"

ehildenb (approved):

PR #2150, opened 2021-08-20
Update to Scala 2.13.5

dwightguth:
Not ready for review yet.


dwightguth:
This is proving to be more work getting all the tests passing than it is probably worth. I am going to try again with the latest version of scala 2.12, which came out in May.

PR #2156, opened 2021-08-24
Tighten outer syntax sort parsing

sskeirik:
Currently, the outer syntax admits sort declarations like:

```k
syntax A:B     ::= "A" // weird case 1
syntax 5       ::= "B" // weird case 2
syntax {2} Foo ::= MyConstructor(2) // another instance of weird case 2
```

But if you tried to use a case 1 sort specifically in a rule, parsing fails:

```k
rule foo(MyVar:A:B) => bar
```

Because, presumably, the variable name is understood to be `MyVar:A` with sort `B`.

This PR modifies `Outer.jj` to forbid the two weird cases listed above, that is, sort names may not be numbers and may not contain semi-colons.

However, it preserves:

-   the ability to use numbers as sort paramters, which is needed for, e.g., machine ints,

    ```k
    syntax Foo ::= myCoolFunc(MInt{32})
    ```

### DIFF
--- a/kernel/src/main/javacc/Outer.jj
+++ b/kernel/src/main/javacc/Outer.jj
@@ -238,7 +238,7 @@ public class Outer {
   public static Sort parseSort(String string) {
     Outer parser = new Outer(new StringReader(string));
     try {
-      return parser.SimpleSort();
+      return parser.SimpleSortParam();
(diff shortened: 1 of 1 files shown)


radumereuta:
Reopen if this is still needed.

PR #2157, opened 2021-08-24
Improve parser error messages

dwightguth:
This PR adds two major improvements to error messages of the form `unexpected token 'foo'`. The first is that the error message mentions the last token that parsed successfully, which can be useful in understanding why the error occurred in many, but not all, cases. The second is that we report a list of ProductionItems that could have appeared next. We show at most 5 by default, but you can get the complete list by passing --verbose to kompile or kast. The list is sorted first by type of production item and then alphabetically, and we automatically filter out non-maximal sorts.

This is not ready for review yet as I still need to test to ensure it doesn't introduce any meaningful performance regressions to the case when parsing succeeds.

### DIFF
--- a/k-distribution/tests/regression-new/checks/functionContextInRewrite.k.out
+++ b/k-distribution/tests/regression-new/checks/functionContextInRewrite.k.out
@@ -1,4 +1,4 @@
-[Error] Inner Parser: Parse error: unexpected token '['.
+[Error] Inner Parser: Parse error: unexpected token '[' following token '('. Expected one of: K, "...".
 	Source(functionContextInRewrite.k)
 	Location(8,12,8,13)
 [Error] Compiler: Had 1 parsing errors.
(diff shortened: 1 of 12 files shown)


dwightguth:
Closing because of performance issues. Will remake a smaller PR without the problematic parts of this change.

radumereuta:
This is some extra information from slack. It's good to keep track of it here since it provides some insight if we want to try similar things in the future.
dwight.guth  22:35
@channel To test whether there is a significant performance regression when parsing rules that successfully parse in my improved parser error message PR, I compiled KEVM both with and without this change using the --profile-rule-parsing flag and computed the min, max, mean, and median rule parsing times.
| Branch      | Run | Min | Max    | Mean  | Median |
| ----------- | --- | --- | ------ | ----- | ------ |
| master      |   1 | 3.8 | 2703.7 |  82.3 |   41.3 |
| master      |   2 | 3.1 | 2564.4 |  95.7 |   48.8 |
| parsererror |   1 | 4.7 | 4883.1 | 128.4 |   65.9 |
| parsererror |   2 | 4.4 | 4136.2 | 109.0 |   53.0 |

As you can see, there is somewhat of a slowdown here. This is because we have to keep track of failing parse states as we parse, which means there are some additional hash set insertions in the parser kernel code which insert into sets that gradually grow in size as the parse continues. It's possible some clever tricks could cut down on this overhead, but we probably can't easily eliminate it entirely unless we want to compute all the information about the error message after the fact when we know parsing has failed, which would likely involve invoking some of the parser code twice, which would then make parsing much slower on definitions that have many errors.

Issue #2154, opened 2021-08-24
[Bug] test case failure in proof-tests


Baltoli:
One test case still fails on my machine (macos11)

Running `make -C proof-tests`, the following command gets run, producing the diff output below:

```shell
/Users/brucecollie/code/k/k-distribution/bin/kprove deposit-spec.k --smt-prelude imap.smt2 --no-exc-wrap  -d .  | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' | diff - deposit-spec.k.out
```

```diff
1,17c1
< <T>
<   <k>
<     tn ( V0 , V1 , up ( V1 , V0 +Int 1 ) ) ~> V2
<   </k>
<   <zerohashes>
<     V3
<   </zerohashes>
<   <branch>
<     V4
<   </branch>
<   <depositCount>
<     V0
<   </depositCount>
<   <treeHeight>
<     V5
<   </treeHeight>
< </T>
---
> #Top
```

dwightguth:
I can reproduce this failure on my local machine. It seems to only affect certain versions of z3. When I install the version of z3 that is used by the CI, the test passes. So I would probably confirm that this is the only test failing on your machine and then ignore it. We don't really have the resources to deal with the fact that different versions of z3 are capable of solving different types of constraints.

ehildenb:
4.8.7 works.

ehildenb:
@Baltoli can you isolate the Z3 query that is actually causing the problem? You can do `--solver-transcript FILE_NAME` to produce a transcript of the Z3 queries that the Haskell backend is sending, and you can check which one behaves differently on 4.8.7 vs later versions. `--solver-transcript` is a `kore-exec` option, so you may need to use `--haskell-backend-command ...` argument to `kprove` to make this work (or to set env variable `KORE_EXEC_OPTS`).

Ask on slack for more info. Getting solver transcripts is worth learning how to do.

Baltoli:
This is a reduced version of the solver transcript that I see being generated:

```smt
(set-option :print-success false )
; success

(set-option :timeout 40 )
; success

(define-sort IMap () (Array Int Int ) )
; success

(declare-sort |HB_#String| 0 )
; success

(declare-sort |HB_SortBranchCellOpt| 0 )
; success

(declare-sort |HB_SortCmds| 0 )
; success

(declare-sort |HB_SortDepositCountCellOpt| 0 )
; success

(declare-sort |HB_SortGeneratedCounterCellOpt| 0 )
[...]

PR #2158, opened 2021-08-25
Demonstrate test failures with Haskell backend

Baltoli:
This PR just changes the `proof-tests/deposit/spec` test case to run with the Haskell backend, as a demo for #2154. Not for merging.


radumereuta:
When you make a new PR, you have the option to make it as a `draft`. It's not going to be announced to the #k channel and it's going to be greyed out so people know it's work in progress. You can still convert it after the fact (see top right).
When it's ready to be reviewed, you can mark it as `ready`, reviewers get notified and we get a message in #k.

Baltoli:
Missed the option on the dropdown - thanks Radu!

PR #2160, opened 2021-08-25
report better error if bison parsing fails

dwightguth:
Previously we had several minor issues that arose if the user passed a program that did not parse correctly to the bison parser. Now we correctly print a helpful error message and then immediately exit in all such cases.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java
+++ b/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java
@@ -102,6 +102,10 @@ public class Scanner implements AutoCloseable {
             "    yylloc->first_column = yycolumn; yylloc->last_column = yycolumn + yyleng - 1; \\\n" +
             "   yycolumn += yyleng; \\\n" +
             "   yylloc->filename = filename;\n" +
+            "#define ECHO do {\\\n" +
+            "  fprintf (stderr, \"%d:%d:%d:%d:syntax error: unexpected %s\\n\", yylloc->first_line, yylloc->first_column, yylloc->last_line, yylloc->last_column, yytext);\\\n" +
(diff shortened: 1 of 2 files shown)


kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java
```diff
@@ -102,6 +102,10 @@ public void writeStandaloneScanner(File path) {
             "    yylloc->first_column = yycolumn; yylloc->last_column = yycolumn + yyleng - 1; \\\n" +
             "   yycolumn += yyleng; \\\n" +
             "   yylloc->filename = filename;\n" +
+            "#define ECHO do {\\\n" +
+            "  fprintf (stderr, \"%d:%d:%d:%d:syntax error: unexpected %s\\n\", yylloc->first_line, yylloc->first_column, yylloc->last_line, yylloc->last_column, yytext);\\\n" +
+            "  exit(1);\\\n" +
+            "} while (0)\n" +
```
radumereuta, line 108: Why is this inside a do-while?
Also, do you have an example of how the error message looked before and after?

dwightguth: I mean, previously it would just silently discard unrecognized tokens from the input and treat the program as if it didn't contain those characters. so there was no error previously.

As far as the do-while issue, it's fairly idiomatic C when you have a function-like macro that you want to contain multiple C statements. Here's a pretty good explanation on stack overflow: https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

radumereuta (approved):

radumereuta:
lgtm
If you don't want to wait for Everett to review it, then go ahead and add automerge.

PR #2161, opened 2021-08-25
Update parser error messages, second draft

dwightguth:
This is a version of #2157 which has had the "expected" logic stripped out, leaving only the "following token" logic.


radumereuta (approved):
lgtm
Although Rikard had a comment on slack regarding the usefulness of having more informative error messages with a flag even though parsing might get slower.

ehildenb (approved):

PR #2136, opened 2021-08-17
Draft: implement -I command line option for Kprovex

SchmErik:
Extract -d from OuterParsingOptions

Signed-off-by: Erik Kaneda <erik.kaneda@runtimeverification.com>

PR #2164, opened 2021-08-26
More verbose information

radumereuta:
for kprove and kprovex
for scanners and bison parser


radumereuta:
Here is how `-v` looks now:
```
$ kompile test.k -v --debug --backend haskell --gen-bison-parser
Parse command line options                                   =  0.011s
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/prelude.md)
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/kast.md)
Importing: Source(/home/radu/work/k/k-distribution/include/kframework/builtin/domains.md)
Importing: Source(/home/radu/work/test/test.k)
Outer parsing [65 modules]                                   =  0.761s
  New scanner: TEST-CONFIG-CELLS                             =  0.122s
  New scanner: STDOUT-STREAM-CONFIG-CELLS                    =  0.181s
  New scanner: STDIN-STREAM-CONFIG-CELLS                     =  0.207s
Parse configurations [3/3 declarations]                      =  0.473s
  New scanner: TEST-RULE-CELLS                               =  0.129s
  New scanner: STRING-KORE-RULE-CELLS                        =  0.171s
  New scanner: TEST-RULE-CELLS                               =  0.169s
  New scanner: STRING-COMMON-RULE-CELLS                      =  0.190s
  New scanner: K-IO-RULE-CELLS                               =  0.259s
  New scanner: STDOUT-STREAM-RULE-CELLS                      =  0.258s
  New scanner: STDIN-STREAM-RULE-CELLS                       =  0.285s
Parse rules [101/101 rules]                                  =  0.544s
Validate definition                                          =  0.150s
Apply compile pipeline                                       =  0.261s
  New scanner: TEST-SYNTAX-PROGRAM-GRAMMAR                   =  0.045s
  New Bison parser: TEST-SYNTAX-PROGRAM-GRAMMAR              =  0.147s
Kompile to kore                                              =  0.150s
Save to disk                                                 =  0.030s
Backend                                                      =  0.277s
Cleanup                                                      =  0.000s
Total                                                        =  2.657s
```
```
$ kprovex spec.k --debug -v
Parse command line options                                   =  0.010s
Importing: Source(/home/radu/work/test/spec.k)
Importing: Source(/home/radu/work/test/test.k)
Outer parsing [72 modules]                                   =  0.399s
  New scanner: TEST-RULE-CELLS                               =  0.149s
  New scanner: SPEC-RULE-CELLS                               =  0.130s
Parse spec modules [1/1 rules]                               =  0.589s
Apply prover steps                                           =  0.212s
Executing [kore-exec, /home/radu/work/test/./test-kompiled/definition.kore, --module, TEST, --prove, /home/radu/work/test/./.kprove-2021-08-26-16-10-44-237-e089dcee-7ad0-4a7e-bb05-f9c9b7697286/spec.kore, --spec-module, SPEC, --output, /home/radu/work/test/./.kprove-2021-08-26-16-10-44-237-e089dcee-7ad0-4a7e-bb05-f9c9b7697286/result.kore]
Executing command: kore-exec /home/radu/work/test/./test-kompiled/definition.kore --module TEST --prove /home/radu/work/test/./.kprove-2021-08-26-16-10-44-237-e089dcee-7ad0-4a7e-bb05-f9c9b7697286/spec.kore --spec-module SPEC --output /home/radu/work/test/./.kprove-2021-08-26-16-10-44-237-e089dcee-7ad0-4a7e-bb05-f9c9b7697286/result.kore
kore-exec: [268587] Warning (WarnStuckClaimState):
    The proof has reached the final configuration, but the claimed implication is not valid. Location: /home/radu/work/test/spec.k:8:11-8:59
Backend                                                      =  0.399s
#Top
Cleanup                                                      =  0.038s
Total                                                        =  1.647s
```

dwightguth (changes requested):
Please don't access kompileOptions.global and mutate it like this. Instead, just inject GlobalOptions and access the injected value directly.

The rest of this seems fine though.

SchmErik:
> Please don't access kompileOptions.global and mutate it like this. Instead, just inject GlobalOptions and access the injected value directly.

@dwightguth I'm curious about your comment here. Could you tell me the reasoning behind why you're saying this?  I'm assuming that you're talking about accessing the global options in the actual constructor. Why is this preferred over overriding the value in the `@inject` method?

radumereuta:
@dwightguth we have quite a few references to `KompileOptions.GlobalOptions`. I will have to change all of those to reference the injected one. If I miss one (now, or in the future) then it's a potential for bugs.
This will only be made worse by #2136, where @SchmErik has the same exact problem with `OuterParsingOptions`

Can you be more specific? Because the way I understand your suggestion right now doesn't offer a very compelling solution.

SchmErik:
> @dwightguth we have quite a few references to `KompileOptions.GlobalOptions`. I will have to change all of those to reference the injected one. If I miss one (now, or in the future) then it's a potential for bugs.
> This will only be made worse by #2136, where @SchmErik has the same exact problem with `OuterParsingOptions`
> 
> Can you be more specific? Because the way I understand your suggestion right now doesn't offer a very compelling solution.

Hey @radumereuta, I think I have the answer... @dwightguth please chime in if something doesn't sound right... So for my `OuterParsingOptions` class, doing this:
```java
     @Inject
     public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw) {
        this(kompileOptions.updateOuterParsingOptions(outerParsing), outerParsing, files, kem, sw, true);

     }
```
means that I am unconditionally overriding `kompileOption`'s `OuterParsingOption` field. For a regular `kompile` invocation, this seems to clobber the value that's already contained in `kompileOptions.OuterParsing`. For programs like `kprove`, we have an `OuterParsingOptions` class inside of `KProveOptions` and the `kompileOption.OuterParsing` does not get populated. In this case, we want to override `kompileOption.OuterParsing` with the one contained in `KProveOptions`. It's much safer to handle this logic in the parameterized constructor by doing something like this:
```java
    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
        this.kompileOptions = kompileOptions;
        if (outerParsing != null) {
            kompileOptions.outerParsing = outerParsing;
        }

        //... more code here
```
For Radu's case, it might not cause anything to crash because it's the `GlobalOptions` object but I think having this code in the `@inject` constructor is more resilient to change. I could foresee someone placing the `@inject` annotation on a different constructor and causing something to break. I also think it looks more organized to handle something like this in the main constructor so these function calls aren't scattered around various constructors in this file.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -78,7 +78,8 @@
     public static final File BUILTIN_DIRECTORY = JarInfo.getKIncludeDir().resolve("builtin").toFile();
     public static final String REQUIRE_PRELUDE_K = "requires \"prelude.md\"\n";
 
-    public final KompileOptions kompileOptions;
+    private final KompileOptions kompileOptions;
+    private GlobalOptions globalOptions;
```
dwightguth: why isn't this final?

dwightguth (approved):

PR #2169, opened 2021-08-27
Fix duplicate module not being reported

### LINKED ISSUE #2146: [Bug] kompile - No error for duplicated module
K Version
---------

Example:

```
$ kompile --version
K version:    5.1.0
Build date:   Thu Aug 19 17:16:41 EEST 2021

$ git log | head -n 3
commit ed9b191c492b4d93a40a37a2db23589dfb2bd98b
Author: Dwight Guth <dwight.guth@runtimeverification.com>
Date:   Wed Aug 18 15:10:36 2021 -0500
```

Description
-----------

Declaring a module called LIST does not produce a conflict error with the LIST module in the `domains.md` file, and `kompile` silently uses the `domains.md` one.

Input Files
-----------

`tmp.k`
```
module TMP-SYNTAX
  syntax KItem ::= "a"
endmodule

module LIST
  imports TMP-SYNTAX
  rule a => .K
endmodule

module TMP
  imports LIST
  imports TMP-SYNTAX
endmodule
```

`tmp-proof.k`
```
module TMP-PROOF
  imports TMP-SYNTAX

  claim a => .K
endmodule
```

Reproduction Steps
------------------

No kompile error, proof fails:

```
$ kompile tmp.k --backend haskell && kprove tmp-proof.k
kore-exec: [247786] Warning (WarnStuckClaimState):
    The claim cannot be rewritten further, and the claimed implication is not valid. /mnt/data/runtime-verification/elrond-multisig/tmp2/tmp-proof.k:4:9-4:16
<k>
  a ~> _DotVar1 ~> .
</k>
```

However, if the module name and import are changed to, say, `LIS` instead of `LIST`, the proof works.

Expected Behavior
-----------------

`kompile` should produce an error about the `LIST` module being declared twice.

radumereuta:
Fixes: #2146

I tried to add a test in `regression-new/checks` but that includes paths from the include directory which isn't being handled by our script and jenkins fails. I tested it manually, and I also had to modify the pyk example to make it work.

### DIFF
--- a/k-distribution/tests/regression-new/itp/nth-ancestor/nth1-spec.k
+++ b/k-distribution/tests/regression-new/itp/nth-ancestor/nth1-spec.k
@@ -23,8 +23,6 @@
     by apply/connect1.
 */
 
-require "chain.k"
-
 // base case
 module NTH1-SPEC
     imports V
--- a/kernel/src/main/java/org/kframework/parser/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/ParserUtils.java
@@ -34,6 +34,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -207,6 +208,8 @@ public class ParserUtils {
                         previousModules.stream().map(m -> Triple.of(m.name(), m.att().get(Att.SOURCE(), Source.class),
                                 m.att().get(Att.LOCATION(), Location.class))),
                         kilModules.stream().map(m -> Triple.of(m.getName(), m.getSource(), m.getLocation())))
+                // make sure we have unique modules (double requires), and preserve order
+                .collect(Collectors.toCollection(LinkedHashSet::new)).stream()
                 .collect(Collectors.groupingBy(Triple::getLeft));
 
         List<String> duplicateModules = groupedModules
--- a/kernel/src/main/java/org/kframework/parser/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/ParserUtils.java
@@ -1,8 +1,11 @@
 // Copyright (c) 2015-2019 K Team. All Rights Reserved.
 package org.kframework.parser;
 
+import com.google.common.collect.Streams;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.tuple.Triple;
 import org.kframework.attributes.Att;
+import org.kframework.attributes.Location;
 import org.kframework.attributes.Source;
 import org.kframework.definition.FlatModule;
 import org.kframework.definition.Module;
@@ -27,7 +30,6 @@ import org.kframework.utils.options.OuterParsingOptions;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.LinkOption;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Comparator;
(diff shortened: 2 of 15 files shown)


k-distribution/tests/regression-new/itp/nth-ancestor/nth1-spec.k
```diff
@@ -23,8 +23,6 @@
     by apply/connect1.
 */
```
SchmErik, line 25: From my understanding, you're removing this chain.k so that the test suite won't fail. Could you add another testcase that will cause a failure and update the expected output so that we're exercising this new error?

SchmErik (dismissed):
From my understanding, you're removing this chain.k so that the test suite won't fail. Could you add another testcase that will cause the kompile error that you're adding and update the expected output so that we're exercising this new error?

k-distribution/tests/regression-new/itp/nth-ancestor/nth1-spec.k
radumereuta: That's what I did initially (see my first commit).
But our testing infrastructure is not smart enough to remove paths from the include directory, and Jenkins started to fail.
```
[Error] Outer Parser: Module LIST previously declared at Source(/home/radu/work/k/k-distribution/target/release/k/include/kframework/builtin/domains.md) and Location(631,1,746,9)
```
To add a test that exercises the reported error would involve significant changes to the makefiles, and I don't think they are warranted just for this test.

SchmErik: I thought that a script removed some of these absolute paths and replaced them with just the filenames before comparing against a saved diff

SchmErik: I just checked the test suite. When you're running something like `make -j8 -O -k` from the `regression-new/checks` directory, I see this command:
`kompile -w none --no-exc-wrap --backend llvm  badFunctionRuleWithContext.k -d . 2>&1 | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' | diff - badFunctionRuleWithContext.k.out`
but this removes the current directory. Maybe you could add a test that will fail from a duplicate module that's defined in the test suite's working directory rather than `domains.md`. You could even duplicate definition of the same module in the same file.

radumereuta: We already have such a test: https://github.com/kframework/k/blob/master/k-distribution/tests/regression-new/checks/duplicateModule.k
Again, check the first commit in this PR.
The problem was that we had separate checks for domains.md vs user definition.
To fix this I put together the previousModules with the current modules and run the check again.

SchmErik (approved):
With all of the things that we've talked about, I think this looks good. There's no need to add a new test because the existing regression test should exercise the code you wrote.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -199,8 +202,15 @@ private String loadDefinitionText(File definitionFile) {
 
         new CollectProductionsVisitor(kore, context).visit(def);
 
-        Map<String, List<org.kframework.kil.Module>> groupedModules = kilModules.stream()
-          .collect(Collectors.groupingBy(org.kframework.kil.Module::getName));
+        // Triple of moduleName, Source, Location
+        Map<String, List<Triple<String, Source, Location>>> groupedModules =
```
dwightguth: we generally have been using scala tuples rather than apache commons lang3 where possible in new code. So please make this a Tuple3 instead.

dwightguth (changes requested):
what exactly was going wrong here? I don't understand why the groupingBy we did before wouldn't have been sufficient  to identify this issue. Also, we need a test.

radumereuta:
@dwightguth regarding the test, please read my first message:
>I tried to add a test in `regression-new/checks` but that includes paths from the include directory which isn't being handled by our script and jenkins fails. I tested it manually, and I also had to modify the pyk example to make it work.

The problem was we were doing this check for each call, but we called this check twice. Once for domains.md and once for the user definition. So we would check correctly for duplicates in each subsection but not together. I added the tuple to unify the two groups and apply the check again.

dwightguth:
Ah I see. That makes sense to me. We do still need a test though, regardless of whether the test proves difficult to create.

dwightguth (approved):
seems fine

PR #2172, opened 2021-08-30
Implement -I, --md-selector, and --no-prelude in kprove, kprovex, keq, and kbmc

### LINKED ISSUE #1330: Add `--md-selector` option to `kprove`

### LINKED ISSUE #1909: [KIP] - Kprove should have configurable include paths
[KIP] - Use `foo` instead of `syntax` for declaring productions

Motivation
----------

This would be useful for better access to trusted claims.

Example K Code
--------------

I would like to be able to use project relative paths for including K files in proofs, e.g. the `trusted` directory mention below could live at root of the project, while the proof below could be in a random subfolder.

```
require "trusted/claim1.k"

module MY-PROOF
  imports CLAIM1

  claim ...
endmodule
```

Documentation
-------------

A user can specify an include path with -I, e.g.

```
$ kprove -I `git rev-parse --show-toplevel` my-proof.k
```

I guess users could use relative paths (which in some contexts is fairly hard to do, e.g. when generating trusted claims with bazel) or they could concatenate files instead of requiring them.

Testing Approach
----------------

I don't know much about the existing testing harness.

SchmErik:
This change implements command line options from OuterParsingOptions class for several tools other than kompile. This is broken down into the following changes:

Extract `-d` from `OuterParsingOptions`
----------------------------------------------------
This extraction prepares `OuterParsingOptions` class to be included in the
following tools: kprove, kprovex, keq, and kmbc.

KProve, KEq, KBMC: add OuterParsing options
----------------------------------------------------
Adding this object enables the `-I`, `--md-selector`, and `--no-prelude`
options to these tools. By doing so, each tool can configure their own
OuterParsing options rather than getting this information by de-serializing
the kompile object from the cache left by a previous invocation of the
`kompile` command.

pyk tests: remove -I from kompile add -I to kprove
----------------------------------------------------
This test case utilized -I in the kompile command in order to pass the
parameter of this option to kprove. This was done through serializing
the `kompile` object during the invocation of the `kompile` command and
de-serializing the `kompile` object to access the `outerParsingOption`
during the invocation of the `kprove` command. Since the -I option has
been added to `kprove`, drop this command from the `kompile` command and
move it to the `kprove` command.

kprove-markdown test: use --md-selector for testing
Add additional lines in a test file to test the new --md-selector
option for kprove. Before this feature was added, the default behavior was for
kprove to use `k` which is the default md-selector. In order to test that the --md-selector
option is functional, text has been added to the test case so that it will
cause a failure on the default md-selector.

Fixes: #1909
Fixes: #1330

### DIFF
--- a/kernel/src/main/java/org/kframework/kdep/KDepModule.java
+++ b/kernel/src/main/java/org/kframework/kdep/KDepModule.java
@@ -5,6 +5,7 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.google.inject.TypeLiteral;
 import com.google.inject.multibindings.Multibinder;
+import org.kframework.kompile.KompileOptions;
 import org.kframework.main.FrontEnd;
 import org.kframework.main.GlobalOptions;
 import org.kframework.main.Tool;
@@ -12,6 +13,7 @@ import org.kframework.utils.inject.Options;
 import org.kframework.utils.inject.OuterParsingModule;
 import org.kframework.utils.inject.RequestScoped;
 import org.kframework.utils.options.OuterParsingOptions;
+import org.kframework.utils.options.OutputDirectoryOptions;
 
 /**
  * Guice module for kdep tool. Binds the information needed to compute the kompiled directory as well as the options
@@ -36,7 +38,8 @@ public class KDepModule extends AbstractModule {
     }
 
     @Provides
-    OuterParsingOptions outerParsingOptions(KDepOptions options) {
-        return options.outerParsing;
-    }
+    OuterParsingOptions outerParsingOptions(KDepOptions options) { return options.outerParsing; }
+
+    @Provides
+    OutputDirectoryOptions outputDirectoryOptions(KompileOptions options) { return options.outputDirectory; }
 }
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -42,6 +42,7 @@ import org.kframework.utils.errorsystem.KExceptionManager;
 import org.kframework.utils.file.FileUtil;
 import org.kframework.utils.file.JarInfo;
 
+import org.kframework.utils.options.OuterParsingOptions;
 import scala.collection.JavaConverters;
 import scala.Function1;
 import scala.Option;
@@ -86,7 +87,7 @@ public class Kompile {
     java.util.Set<KEMException> errors;
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, boolean cacheParses) {
-        this(kompileOptions, files, kem, new Stopwatch(kompileOptions.global), cacheParses);
+        this(kompileOptions, null, files, kem, new Stopwatch(kompileOptions.global), cacheParses);
     }
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem) {
@@ -94,12 +95,24 @@ public class Kompile {
     }
 
     @Inject
-    public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, Stopwatch sw) {
-        this(kompileOptions, files, kem, sw, true);
+    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw) {
+        this(kompileOptions, outerParsing, files, kem, sw, true);
     }
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
+        this(kompileOptions, null, files, kem, sw, cacheParses);
+    }
+
+    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
         this.kompileOptions = kompileOptions;
+
+        /**
+         * Invoking the kompile command will inject outerParsing into kompileOptions.outerParsing and leave the outerParsing parameter as null.
+         * Skip this assignment to prevent kompileOptions.outerParsing from being clobbered for the kompile utility.
+         */
+        if (outerParsing != null) {
+            kompileOptions.outerParsing = outerParsing;
+        }
         this.files = files;
         this.kem = kem;
         this.errors = new HashSet<>();
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -171,7 +171,7 @@ public class Kompile {
         } else {
           rootCell = Sorts.GeneratedTopCell();
         }
-        CompiledDefinition def = new CompiledDefinition(kompileOptions, parsedDef, kompiledDefinition, files, kem, configInfo.getDefaultCell(rootCell).klabel());
+        CompiledDefinition def = new CompiledDefinition(kompileOptions, kompileOptions.outerParsing, parsedDef, kompiledDefinition, files, kem, configInfo.getDefaultCell(rootCell).klabel());
 
         if (kompileOptions.genBisonParser || kompileOptions.genGlrBisonParser) {
             if (def.configurationVariableDefaultSorts.containsKey("$PGM")) {
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -84,46 +84,36 @@ public class Kompile {
     private final ParserUtils parser;
     private final Stopwatch sw;
     private final DefinitionParsing definitionParsing;
+    private final OuterParsingOptions outerParsingOptions;
     java.util.Set<KEMException> errors;
 
-    public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, boolean cacheParses) {
-        this(kompileOptions, null, files, kem, new Stopwatch(kompileOptions.global), cacheParses);
+    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsingOptions, FileUtil files, KExceptionManager kem, boolean cacheParses) {
+        this(kompileOptions, outerParsingOptions, files, kem, new Stopwatch(kompileOptions.global), cacheParses);
     }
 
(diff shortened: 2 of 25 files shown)


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -86,20 +87,32 @@
...
+    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
         this.kompileOptions = kompileOptions;
+
+        /**
+         * Invoking the kompile command will inject outerParsing into kompileOptions.outerParsing and leave the outerParsing parameter as null.
+         * Skip this assignment to prevent kompileOptions.outerParsing from being clobbered for the kompile utility.
+         */
+        if (outerParsing != null) {
+            kompileOptions.outerParsing = outerParsing;
+        }
```
radumereuta: @dwightguth can you please have a look here?
On my PR you requested changes for something similar.

radumereuta (commented):
Overall very good. Just one comment where I would like @dwightguth to have a look.
Also, the comments in the first message seem to be a repeat of the commit messages.
I wouldn't include them there. If I need further explanations for each commit, then I can have a look at each one. Save the first message for an overall description of what you are trying to achieve in this PR. In this case I wouldn't even bother since the issue you are trying to fix already describes the problem pretty well.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -86,20 +87,32 @@
     java.util.Set<KEMException> errors;
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, boolean cacheParses) {
```
dwightguth: where is this constructor called? I would prefer not to construct an instance with an uninitialized field like this.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -86,20 +87,32 @@
...
 
     @Inject
-    public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, Stopwatch sw) {
-        this(kompileOptions, files, kem, sw, true);
+    public Kompile(KompileOptions kompileOptions, OuterParsingOptions outerParsing, FileUtil files, KExceptionManager kem, Stopwatch sw) {
+        this(kompileOptions, outerParsing, files, kem, sw, true);
     }
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
+        this(kompileOptions, null, files, kem, sw, cacheParses);
```
dwightguth: same question here

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth: I would prefer instead of mutating KompileOptions that we simply store the OuterParsingOptions in the Kompile object and reference it directly when we need to read its values. We try to avoid this type of mutability in the code base where possible.

dwightguth (changes requested):
seems mostly reasonable but I had a few specific comments

kernel/src/main/java/org/kframework/kompile/Kompile.java
SchmErik: I think the main problem that I see is that kompileOptions.outerParsing is initialized differently depending on the command that we're running. If I can find a way to construct an instance of kompile options with the appropriate outerParsing object, we can eliminate this change along with the constructor change

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -35,6 +36,9 @@
     @ParametersDelegate
     public OuterParsingOptions outerParsing = new OuterParsingOptions();
 
+    @ParametersDelegate
+    public OutputDirectoryOptions outputDirectory = new OutputDirectoryOptions();
```
radumereuta, line 40: I think this should be `transient`.

radumereuta, line 40:
```suggestion
    // WARNING: use Guice dependency injection to access this field.
    @ParametersDelegate
    public transient OutputDirectoryOptions outputDirectory = new OutputDirectoryOptions();
```

kernel/src/main/java/org/kframework/kompile/Kompile.java
SchmErik: kernel/src/main/java/org/kframework/kompile/KompileFrontEnd.java line 67

kernel/src/main/java/org/kframework/kompile/Kompile.java
SchmErik: this is being called from k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala line 41:
val compiler = new Kompile(options, files, kem, false)

SchmErik: @dwightguth I guess one option is to initialize the outerParsingOption with kompileOptions.OuterParsing but I am not sure if we should be doing that. This constructor is called from a test and the other one is called from kompile so I'm tempted to say that this would be fine...

radumereuta:
The final version should probably contain a test that exercises the `-I` and `--md-selector` options.

SchmErik:
> The final version should probably contain a test that exercises the `-I` and `--md-selector` options.

You helped me add the -I tests by helping me fix the pyk tests so I think we all that we have left is a good --md-selector test. Do you know of any good test cases to exercise this?

dwightguth (commented):
Looks good except you should probably use KProveOptions.outerParsing when constructing the CompiledDefinition object in kprove.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -38,6 +39,9 @@ GlobalOptions getGlobalOptions_UseOnlyInGuiceProvider() {
     @ParametersDelegate
     public OuterParsingOptions outerParsing = new OuterParsingOptions();
 
+    @ParametersDelegate
+    public transient OutputDirectoryOptions outputDirectory = new OutputDirectoryOptions();
+
```
radumereuta, line 44: I'm a bit confused. Wasn't OuterParsingOptions supposed to be transient, and OutputDirectoryOptions be serializable?

kernel/src/main/java/org/kframework/kdep/KDepModule.java
```diff
@@ -36,7 +38,8 @@ GlobalOptions globalOptions(KDepOptions options) {
     }
 
     @Provides
-    OuterParsingOptions outerParsingOptions(KDepOptions options) {
-        return options.outerParsing;
-    }
+    OuterParsingOptions outerParsingOptions(KDepOptions options) { return options.outerParsing; }
+
+    @Provides
+    OutputDirectoryOptions outputDirectoryOptions(KompileOptions options) { return options.outputDirectory; }
```
radumereuta, line 44: @dwightguth shouldn't these be `@RequestScoped`?

kernel/src/main/java/org/kframework/kompile/KompileModule.java
```diff
@@ -45,4 +46,7 @@ GlobalOptions globalOptions(KompileOptions options) {
 
     @Provides
     OuterParsingOptions outerParsingOptions(KompileOptions options) { return options.outerParsing; }
+
+    @Provides
+    OutputDirectoryOptions outputDirectoryOptions(KompileOptions options) { return options.outputDirectory; }
```
radumereuta, line 51: `@RequestScoped`?

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -286,18 +286,18 @@ public void initialize(CompiledDefinition def) {
...
                     this.rewriteThreadSet.requires(), this.rewriteThreadSet.ensures());
         }
         KLabel stratCell = KLabel("<s>");
         if (mainModule.definedKLabels().contains(stratCell)) {
             Rule makeStuck = Rule(IncompleteCellUtils.make(stratCell, false, KRewrite(KSequence(), KApply(KLabel("#STUCK"))), true), BooleanUtils.TRUE, BooleanUtils.TRUE);
             Rule makeUnstuck = Rule(IncompleteCellUtils.make(stratCell, false, KRewrite(KApply(KLabel("#STUCK")), KSequence()), true), BooleanUtils.TRUE, BooleanUtils.TRUE);
-            this.makeStuck = convert(new Kompile(kompileOptions, globalOptions, files, kem).compileRule(def.kompiledDefinition, makeStuck));
-            this.makeUnstuck = convert(new Kompile(kompileOptions, globalOptions, files, kem).compileRule(def.kompiledDefinition, makeUnstuck));
+            this.makeStuck = convert(new Kompile(kompileOptions, kompileOptions.outerParsing, globalOptions, files, kem).compileRule(def.kompiledDefinition, makeStuck));
+            this.makeUnstuck = convert(new Kompile(kompileOptions, kompileOptions.outerParsing, globalOptions, files, kem).compileRule(def.kompiledDefinition, makeUnstuck));
```
radumereuta, line 300: Probably these references should all use the injected outerParsing object instead of the reference from kompileOptions. That's going to be null.

radumereuta (changes requested):

kernel/src/main/java/org/kframework/kdep/KDepModule.java
dwightguth: It honestly doesn't matter. The only difference is that if it's request scoped, this function will be called only once instead of each time the options are requested. But this function is so simple it will be inlined by the jvm, so there's no significant difference in performance because of this.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
dwightguth: This doesn't actually matter because this method is always called by kompile, where it actually won't be null

dwightguth (approved):

radumereuta (approved):
Ok, I'm convinced by Dwight's answers.

PR #2181, opened 2021-09-07
Add git --describe output to --version

### LINKED ISSUE #2180: kompile --version should include git revision
When built locally kompile --version looks something like this:
```
$ kompile --version
K version:    5.1.0
Build date:   Fri Sep 03 21:34:37 EEST 2021
```
Which is wrong. Only the CI knows the exact version number, and the local build puts a 0 instead.
Since we introduced the version number this caused problems with local builds when reporting errors.
I think it should always include git revision. No matter what the version number says.

Baltoli:
The packaged version identifier doesn't properly capture the state of the repository when the package was built (only the most recent minor version - closes #2180). 

This PR adds and configures a build plugin that registers the output of `git --describe` to the kernel JAR manifest. The git revision info can then be read at runtime in the same way that the build time etc. are.

The plugin used is [here](https://github.com/git-commit-id/git-commit-id-maven-plugin) - it seems to be actively maintained and used by lots of other projects.

### DIFF
--- a/kernel/pom.xml
+++ b/kernel/pom.xml
@@ -209,6 +209,8 @@
           </execution>
         </executions>
         <configuration>
+          <useNativeGit>true</useNativeGit>
+          <failOnNoGitDirectory>false</failOnNoGitDirectory>
           <includeOnlyProperties>
             <includeOnlyProperty>^git.commit.id.describe$</includeOnlyProperty>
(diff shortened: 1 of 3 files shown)


Baltoli:
This approach wasn't going to be the right one to go with, so closing the PR and will revisit on a different tack.

PR #2184, opened 2021-09-08
Include git describe in output of kompile --version

### LINKED ISSUE #2180: kompile --version should include git revision

Baltoli:
When the project is compiled, the output of git describe is captured as
a Maven property; this property is then stored in the packaged .jar
file's manifest (similarly to existing properties like the build
timestamp).

This implementation doesn't rely on an external plugin, but does require
a git executable (other antrun tasks for copyright checks have similar
requirements). It will fail silently if the package is built from a
source distribution with no .git directory such as a tarball, and the
revision will not be printed.

Example output:
```
K version:    5.1.0
Build date:   Wed Sep 08 11:19:05 BST 2021
Git revision: v5.1.172-0-ge1a5b4f92-dirty
```

Fixes #2180

### DIFF
--- a/kernel/pom.xml
+++ b/kernel/pom.xml
@@ -232,6 +232,30 @@
         </configuration>
       </plugin>
       <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>1.7</version>
+        <executions>
+          <execution>
+            <id>get-git-revision</id>
+            <phase>compile</phase>
+            <configuration>
+              <exportAntProperties>true</exportAntProperties>
+              <target>
+                <exec executable="git" outputproperty="git.describe.out" failifexecutionfails="false">
+                  <arg value="describe" />
+                  <arg value="--tags" />
+                  <arg value="--dirty" />
+                  <arg value="--long" />
+                </exec>
+              </target>
+            </configuration>
+            <goals>
+              <goal>run</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
         <artifactId>maven-jar-plugin</artifactId>
         <version>2.5</version>
         <configuration>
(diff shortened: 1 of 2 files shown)


radumereuta:
One small thing that I don't like is `5.1.0` conflicts with `v5.1.172-0-ge1a5b4f92-dirty`
The first line says 0, the second one says 172. Could we reconfigure the output somehow?
How about replacing `String version` with `gitRevision` if not empty instead of printing a new line?
(I'm asking because I don't know what corner cases might be where this would not work).

Baltoli:
Replacing the existing version line makes sense to me on reflection - it would then reflect the most specific version information that can be obtained for the current source package. If getting the git description fails, we would always be able to fall back to the existing version information and be no worse off than we currently are.

I can't think of a corner case where this would break and end up worse off than we currently are.

Baltoli:
For clarity, the proposed output would be:

Building from tarball:
```
K version:    5.1.0
Build date:   Wed Sep 08 11:19:05 BST 2021
```

Building from a Git checkout:
```
K version:    v5.1.172-0-ge1a5b4f92-dirty
Build date:   Wed Sep 08 11:19:05 BST 2021
```

@ehildenb do you have a preference on this?

radumereuta:
Since we're at it, @ehildenb  what is the reason that stopped us from committing the full version into the master branch?

ehildenb:
Hmmm, this would probably be a nice thing to fix, yes. Well, the main reason is that versions get bumped on the `release` branch, not on the `master` branch, because of how our CI works. But we could very well have that `git describe` command be used to extract the version printed by the Java backend, and that should work, because the tag it gives should be the previous version tag I think.

Baltoli:
I've updated the format of the --version output to only print a single version string @radumereuta.

radumereuta (approved):
lgtm
Feel free to add `automerge` if you don't want to wait for Everett's review.

PR #2186, opened 2021-09-08
Refactor K-REFLECTION

### LINKED ISSUE #2122: Too many scanners being created
See https://github.com/kframework/k/issues/2119#issuecomment-895483245

Many of the scanners are useless, for example we're building scanners for INT but we aren't even importing that.

radumereuta:
Reduce the number of automatically imported modules.


radumereuta:
Too many issues keep surfacing from this, for too little gain. I'm shelving this for now.

PR #2191, opened 2021-09-09
Lesson 21: symbolic execution

dwightguth:


k-distribution/k-tutorial/1_basic/21_symbolic_execution/README.md
```diff
@@ -1 +1,174 @@
...
+that such variables begin with the `?` prefix. For example, here is a rule
+that rewrites a constructor `foo` to a symbolic integer:
+
+```
+rule foo => ?X:Int
+```
+
+When this rule applies, a variable is introduced to the configuration, which
+then is unified against the rules that might apply in order to symbolically
+execute that configuration.
```
ehildenb:
`````suggestion
```
rule <k> foo => ?X:Int ... </k>
```

When this rule applies, a fresh variable is introduced to the configuration, which
then is unified against the rules that might apply in order to symbolically
execute that configuration.
`````

We should wrap this in `<k>` tags, because it's not logically sound to introduce fresh variables in function rules (which this looks like).

File: k-distribution/k-tutorial/1_basic/21_symbolic_execution/README.md [hunk 2]
```diff
@@ -1 +1,174 @@
...
+has this type of variable on the right-hand side: the `ensures` clause.
+An `ensures` clause is similar to a `requires` clause and can appear after
+a rule body, or after a `requires` clause. The `ensures` clause is used to
+introduce constraints that might apply to the variable that was introduced by
+that rule. For example, we could write the rule above with the additional
+constraint that the symbolic integer that was introduced must be less than
+five, by means of the following rule:
+
+```
+rule foo => ?X:Int ensures ?X <Int 5
```
ehildenb:
```suggestion
rule <k> foo => ?X:Int ... </k> ensures ?X <Int 5
```

k-distribution/k-tutorial/1_basic/21_symbolic_execution/README.md
```diff
@@ -1 +1,174 @@
...
+`#Equals`. While `andBool`, `orBool`, and `==K` represent functions of sort
+`Bool`, `#And`, `#Or`, and `#Equals` are **matching logic connectives**. We
+will discuss matching logic in more detail later in the tutorial, but the basic
+idea is that these symbols represent Boolean operators over the domain of
+configurations and constraints, as opposed to over the `Bool` sort.
+
+Notice that the configuration listed above is a disjunction of conjunctions.
+This is the most common form of output that can be produced by the Haskell
+Backend. In this case, each conjunction consists of a configuration and a pair
+of constraints. What this conjunction describes, essentially, is a
```
ehildenb:
```suggestion
Backend. In this case, each conjunction consists of a configuration and a set
of constraints. What this conjunction describes, essentially, is a
```

Less confusing.

k-distribution/k-tutorial/1_basic/21_symbolic_execution/README.md
```diff
@@ -1 +1,174 @@
...
+it is only capable of rewriting concrete configurations.
+
+By contrast, the Haskell Backend performs **symbolic execution**, which is
+capable of rewriting any configuration, including those where parts of the
+configuration are **symbolic**, ie, contain variables or uninterpreted
+functions.
+
+## Unification
+
+Previusly, we have introduced the concept that K rewrite rules operate by means
```
ana-pantilie:
```suggestion
Previously, we have introduced the concept that K rewrite rules operate by means
```

ana-pantilie (approved):

k-distribution/k-tutorial/1_basic/21_symbolic_execution/README.md
ehildenb: @dwightguth this one as well.

bingle2000:
There is a mistake when explain ensures example at:"In the second case, we see that the second rule did not apply because ?X is less than 10. Moreover, because of the ensures clause on the first rule, we know that ?X is not zero, therefore **the second rule** will not apply a second time. If we had omitted this constraint, we would have ended up infinitely applying the first rule, leading to krun not terminating."

The bolded text **the second rule**  should be "the first rule".

PR #2195, opened 2021-09-10
kdep: fix broken -D option

### LINKED ISSUE #2194: [Bug] [kdep] - kompiled directory path ignored by kdep
The path to the timestamp file generated by `kdep` does not respect the `-d` option and is therefore incorrect.

Steps to reproduce:
```shell
$ touch mod.k
$ mkdir -p some/nested/path
$ kdep -d some/nested/path mod.k
mod-kompiled/timestamp : \
    /Users/brucecollie/code/k/k-distribution/target/release/k/include/kframework/builtin/kast.md \
    /Users/brucecollie/code/k/k-distribution/target/release/k/include/kframework/builtin/domains.md \

```

The expected behaviour from kdep is to respect the `-d` option:
```shell
$ kdep -d some/nested/path mod.k
some/nested/path/mod-kompiled/timestamp : \
    /Users/brucecollie/code/k/k-distribution/target/release/k/include/kframework/builtin/kast.md \
    /Users/brucecollie/code/k/k-distribution/target/release/k/include/kframework/builtin/domains.md \

```

I've run `git bisect` and found that the bug was introduced in the big refactor of frontend options in [this commit](https://github.com/kframework/k/commit/309f81b73c0c3a447e78f4fd2bb413203e519a1d).

@SchmErik is there an obvious reason for this? I'm happy to investigate but you may know more quickly what's going on.

The workaround if this is breaking your Makefiles is just to edit the first line of the `.depend` file manually to point to the correct location.

SchmErik:
Commit 309f81b73c introduced a spelling error in KDepModule.java that
broke kdep's -D option. Fix this by providing the correct `OutputDirectoryOption` in the provider.

Fixes #2194

### DIFF
--- a/k-distribution/include/kframework/ktest-kdep.mak
+++ b/k-distribution/include/kframework/ktest-kdep.mak
@@ -14,4 +14,4 @@ all: $(TESTS)
 dummy:
 
 %.k %.md: dummy
-	$(KDEP) $(OPTS) $@ $(CHECK) $@.out
+	$(KDEP) $(OPTS) $@ | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' $(CHECK) $@.out
--- a/k-distribution/include/kframework/ktest-kdep.mak
+++ b/k-distribution/include/kframework/ktest-kdep.mak
@@ -1,0 +1,17 @@
+# path to the current makefile
+MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+# path to the kdep binary of this distribuition
+KDEP=$(abspath $(MAKEFILE_PATH)/../../bin/kdep)
+# all tests in test directory with matching file extension
+TESTS?=$(wildcard ./*.md) $(wildcard ./*.k)
+
+CHECK=| diff -
+
+.PHONY: all
+
+all: $(TESTS)
+
+dummy:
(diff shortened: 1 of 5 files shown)


SchmErik:
I still need to add a test to this... Does anyone know if we have any tests for `kdep` in `regression-new`? @Baltoli has mentioned that we could do a simple test by looking at the output of `kdep -d somePath ...` and look for `somePath`. I'm not opposed to this but I looked at my test run of `regression-new` and found that we aren't really exercising `kdep` in any of the tests. I think there's benefit in creating a test that exercises kdep and would be great if we can exercise the -d option along with this new test case.

Baltoli:
I know less about the frontend tool options than you @SchmErik, but I think the important features of `kdep` that we'd like to test are:

* The correct root directory is specified for the timestamp file.
* The `-I` option is respected (this might be covered by other tool test suites but I think it's worth having here regardless)
* Transitive includes are computed properly
* Nothing gets dragged in that isn't in the inclusion graph

The test in this PR is enough to test for the specific regression that it fixes - in the interests of keeping PRs small I think the `kdep` tests can probably live in a separate PR to this.

Baltoli (approved):
LGTM - issue raised for more comprehensive kdep tests in #2196

PR #2202, opened 2021-09-14
Sort output from kdep tests

Baltoli:
This should fix the flakiness issue where these tests would
intermittently fail on some CI runs but not others.

### DIFF
--- a/k-distribution/include/kframework/ktest-kdep.mak
+++ b/k-distribution/include/kframework/ktest-kdep.mak
@@ -14,4 +14,7 @@ all: $(TESTS)
 dummy:
 
 %.k %.md: dummy
-	$(KDEP) $(OPTS) $@ | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' $(CHECK) $@.out
+	$(KDEP) $(KDEP_FLAGS) $@ | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' $(CHECK) $@.out
+
+clean:
+	rm -rf $(KOMPILED_DIR) .depend-tmp .depend .kompile-* .krun-* .kprove-* .kbmc-* kore-exec.tar.gz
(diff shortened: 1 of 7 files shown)


k-distribution/include/kframework/ktest-kdep.mak
```diff
@@ -14,4 +14,4 @@ all: $(TESTS)
 dummy:
 
 %.k %.md: dummy
-	$(KDEP) $(OPTS) $@ | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' $(CHECK) $@.out
+	$(KDEP) $(OPTS) $@ | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' | sort $(CHECK) $@.out
```
radumereuta: 1. Can we move this sort in the Java code? This would make it stable for other use cases in the future.
2. rename `OPTS` -> `KDEP_FLAGS` for consistency with other scripts.
3. Implement the `make clean` option. It doesn't do anything in this case, but it does return an error message and it stops the make clean from the parent directory. To be safe, copy the entire rule from ktest.mak.

Alternatively (for 2 and 3), could we include `ktest.mak`, but modify the local Makefile (like [here](https://github.com/kframework/k/blob/master/k-distribution/tests/regression-new/issue-2146-duplicateModules/Makefile)). If it's possible, I would prefer this option since it makes it clear it's a custom scenario and we don't need to create another .mak file.

radumereuta (changes requested):

k-distribution/include/kframework/ktest-kdep.mak
Baltoli: I've addressed these comments - as regards 2/3, I think the best way is to use a new `.mak` as Erik implemented it originally. The `ktest.mak` version does a lot of extra work (rather than just invoking kdep, which doesn't even need to `kompile` the source files). We already do this for `keq` tests as well.

radumereuta (changes requested):
You need to update the outputs since they are now sorted.
```
regression-new/kdep-options$ make update-results
make: *** No rule to make target 'update-results'.  Stop.
```
Also, add this option to the new .mak file so we can update all the tests at once if we need to.

radumereuta (approved):

PR #2203, opened 2021-09-14
Add check for search executable in krun

### LINKED ISSUE #2173: Better error message using `--search` with LLVM backend without having kompiled with `--enable-search`
Currently the LLVM backend with `--search` crashes with a horrible error instead of saying "You must pass `--enable-search` to `kompile` to be able to use `krun --search` with the LLVM backend". This should be improved.

See https://runtimeverification.slack.com/archives/C8DMP61DW/p1630379321002400

Baltoli:
This stops the script from trying to execute something that doesn't
exist, and suggests to the user how they could change their invocation
of krun to fix the problem.

Closes #2173

### DIFF
--- a/k-distribution/src/main/scripts/bin/krun
+++ b/k-distribution/src/main/scripts/bin/krun
@@ -437,12 +437,6 @@ if [ ! -f "$kompiledDir/definition.kore" ]; then
   exit $?
 fi
 
-if $search; then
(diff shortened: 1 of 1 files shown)


Baltoli:
@radumereuta does this need to be tested explicitly, or is the CI passing enough evidence that this doesn't blow up every use of `krun --search` in the test suite?

radumereuta (approved):

PR #2204, opened 2021-09-14
Rework timing info for --profile-rule-parsing

### LINKED ISSUE #1782: [KIP] - Profiling info for kompile
There have been some complaints recently from different users about kompile being slow. I'm trying to make an easy to use profiler that can give users hints about what makes their definition slow to kompile.

I'm proposing an overhaul of `--profile-rule-parsing`:
* [x] for performance reasons, each thread prints its rules in a separate file. There should be a step that aggregates all of them into a single file for easy reading. Maybe order them so we can diff different runs.
* [ ] move the log files from `*-kompiled` to `.kompile-*`. The files are specific to a run of kompile, not the definition.
* [ ] extend the logging to the kompile steps. Maybe there is a generic way for low maintenance.
* [ ] activate it when using `--debug`

Ideas for extra information we could print out:
* [x] Fix new scanner print when verbose (#1908)
* [x] Compact rule-profiling information into a single report (maybe even provide a consolidated summary of the hotspots).
* [ ] How many ambiguities each rule has.
* [ ] Which labels tend to cause ambiguities (by looking at productions present in rules with lots of ambiguities).
* [x] How long each rule takes to parse vs disambiguate (right now we're measuring total time).
* [ ] How long each kompile step tends to take.
* [ ] Warnings for rules that take longer than 1s to parse.

radumereuta:
This changes a bit the way `--profile-rule-parsing` generates timing.log files:
- print the collected messages into a single file specified at command line
- changed the timings from seconds with decimals to milliseconds (since I find them more readable)
- move the option to OuterParsingOptions to make it available in other tools as well

Here is a sample from the output file:
```
   18 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:868   parse:   2 typeInf:   7
    9 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:869   parse:   2 typeInf:   0
   28 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:870   parse:  10 typeInf:   5
   16 /home/radu/work/k/k-distribution/include/kframework/builtin/domains.md:872   parse:   2 typeInf:   0
   38 /home/radu/work/k/k-distribution/include/kframework/builtin/kast.md:523   parse:  12 typeInf:   9
   38 /home/radu/work/test/test.k:11   parse:  13 typeInf:  10
   38 /home/radu/work/test/test.k:5   parse:  12 typeInf:  10
```
All times in milliseconds.
The first column is total parsing time aligned to the right. If something is way bigger, it's going to pop out immediately.
The entries are ordered by the second column to provide a stable output. It's easy to compare different runs.
At the end, I added split timings for parsing time and type inference. If parsing time is large, you may want to add parentheses, if the type inference is large, then you may want to type your variables. If I find more relevant information, I can easily add it later.


k-distribution/tests/regression-new/profile/Makefile
```diff
@@ -1,5 +1,11 @@
 DEF=test
 EXT=test
-KOMPILE_FLAGS=--profile-rule-parsing --syntax-module TEST
+KOMPILE_BACKEND=java
```
dwightguth: Why was the backend of this test changed? In general we want to avoid adding new tests that run on the ocaml or java backends when possible, unless we are specifically testing those backends.

kernel/src/main/java/org/kframework/utils/options/OuterParsingOptions.java
```diff
@@ -51,4 +51,7 @@ public synchronized File mainDefinitionFile(FileUtil files) {
 
     @Parameter(names="--md-selector", description="Preprocessor: for .md files, select only the md code blocks that match the selector expression. Ex:'k&(!a|b)'.")
     public String mdSelector = "k";
+
+    @Parameter(names="--profile-rule-parsing", description="Generate time in seconds to parse each rule in the semantics. Found in -kompiled directory under timing.log.")
+    public String profileRules;
```
dwightguth: This is not an appropriate place for this flag because it does not relate to outer parsing.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -228,10 +228,7 @@ public Scanner getScanner() {
             parseStringTerm(String input, Sort startSymbol, Scanner scanner, Source source, int startLine, int startColumn, boolean inferSortChecks, boolean isAnywhere) {
         scanner = getGrammar(scanner);
 
-        long start = 0;
-        if (profileRules) {
-            start = System.nanoTime();
-        }
+        long start = System.currentTimeMillis(), endParse = 0, startTypeInf = 0, endTypeInf = 0;
```
dwightguth: I don't understand why this is being changed. Not only are we reducing the precision by a factor of 1000, you are now also calling the potentially somewhat expensive function even when its value doesn't matter.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -248,6 +245,7 @@ public Scanner getScanner() {
             } catch (KEMException e) {
                 return Tuple2.apply(Left.apply(Collections.singleton(e)), Collections.emptySet());
             }
+            endParse = System.currentTimeMillis();
```
dwightguth: Same issue throughout here

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -300,18 +300,15 @@ public Scanner getScanner() {
...
                     synchronized(t) {
-                        t.write(source.toString());
-                        t.write(':');
-                        t.write(Integer.toString(startLine));
-                        t.write(':');
-                        t.write(Integer.toString(startColumn));
-                        t.write(' ');
-                        t.write(Double.toString((stop - start) / 1000000000.0));
-                        t.write('\n');
+                        long stop = System.currentTimeMillis();
```
dwightguth: By moving the stop time measurement inside the lock, you have added the totally irrelevant time spent waiting on the lock to the measurement.

dwightguth (changes requested):
Overall, the logic to collate the data seems okay, but there are serious issues with the modifications you made to how the data is measured, and there's other problems with how the options parsing was modified.

kernel/src/main/java/org/kframework/utils/options/OuterParsingOptions.java
radumereuta: The only alternative would be to make an entirely new `InnerParsingOptions.java` class and add it to kompile and kprove (not sure if it's needed anywhere else). Is this what you have in mind?

k-distribution/tests/regression-new/profile/Makefile
radumereuta: That's fair. I needed to test kprove besides kompile. I will change it to haskell.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
radumereuta: According to this article: https://www.geeksforgeeks.org/java-system-nanotime-vs-system-currenttimemillis/
`System.nanoTime()`
- may not be thread safe (although I've read that might not be the case with modern java versions)
- `currentTimeMillis()` executes faster than `nanoTime()` 
- the precision is irrelevant. I did some tests with nanotime and the time variation for the same rule can be 30ms. Considering that we are running multi threaded, cache misses are non-deterministic, opportunistic CPU boost... this variation is understandable. So milliseconds is completely fine in our case. We are looking for big outliers (100ms or more).

I don't feel strongly about it, I can change it back to `nanoTime()`. But according to my research, `currentTimeMillis()` is a better fit for our case.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
radumereuta: That's fair. BUT, that's also confusing. On #2116 you argued that using a `ConcurrentLinkedDeque` (which is thread-safe but contentious when adding elements) is not ok. But you are using a lock here.

What is going on? How is this code working/supposed to work?

kernel/src/main/java/org/kframework/utils/options/OuterParsingOptions.java
dwightguth: that would probably be better. Right now, this flag is being included in KDep, which is completely inappropriate.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
dwightguth: Alright, this is fine then.

dwightguth: We should still only call the function if we are actually profiling though.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
dwightguth: You're right that something weird is going on. When I first wrote this code, each thread had its own ParseInModule instance so it didn't really matter that there was a lock there. At some point that changed so each thread uses the same ParseInModule instance, and we didn't address this inconsistency.  You probably want to change the writer so it's inside a ThreadLocal, I think, and remove the lock.

radumereuta: I see 4 possible solutions here:
1. Use ThreadLocal to print the timing info in a file per thread. First google link I found though warns about using ThreadLocal+ThreadPools: https://www.baeldung.com/java-threadlocal#threadlocalsand-thread-pools
2. Generate a new tempfile for each parsed rule. Basically, bypass the synchronization problem entirely. We're combining them into a single file anyway. But we create more IO traffic.
3. Use a `ConcurrentLinkedQueue` which (as I said above) is threadsafe but contentious when adding elements. But the implementation is based on volatility ([Compare-And-Swap](https://www.baeldung.com/java-queue-linkedblocking-concurrentlinked#concurrentlinkedqueue)). Meaning the variable it's writing into is _not_ stored in cache, only in the main memory, and it offers very low overhead and low probability of collision. Since we're doing this operation outside of the area we're measuring, it's fine. I see this solution as a lower cost than writing to disk.
4. `Tuple2<Errors, Term> parseInModule` becomes `Tuple3<Errors, Term, Timing> parseInModule`. In other words, bypass the thread synchronization issue, and let the stream collectors put everything together. I may even be able to add timing info for rules that fail to parse.

dwightguth: I like the second solution. We are only appending data so the I/O overhead should be pretty low since it will be sequential on disk. We can explore a different solution if it seems that the data changes too much based on this.

dwightguth (approved):

PR #2207, opened 2021-09-15
Updates to pyk

ehildenb:
- Minimizer is updated:
  - Better printing of `#Or` for minimized terms.
  - Addition of `--omit-labels` for not printing certain labels in the resulting config.
  - Special case the `#Top` so the minimizer doesn't crash.
  - Propogate up constraints when printing minimized terms.

- New/removed library functions:
  - `buildAssoc` and `buildCons` for building lists of items.
  - Remove unused `whereMatchingTopDown`
  - Factor out `countVarOccurances` from `uselessVarsToDots`
  - Add `splitConfigAndConstraints` for separating a term and path condition.
  - `propogateUpConstraints` can be used to dedupe constraints between `#Or` clauses.
  - Function `markUselessVars` for going through and putting (or removing) `_` from the front of variable names as needed.
  - `labelsToDots` which takes a list of klabels and replaces occurances of them with dots (used for `--omit-labels`).
  - `pushDownRewrites` is slightly more general.
 
- Misc:
  - Main functionality is abstracted into `main` function, which can be extended in sub-tools.
  - Lots of double-quoted strings are turned into single-quoted strings.
  - Many more functions documented.
  - Allow passing `debug = True` to `prettyPrintKast` to debug where printing fails (usually you built a malformed term).
  - Deduplicate code between `KClaim` and `KRule` cases in several places.
  - Better unparsing of `requires` and `ensures`, and `KSequence`.


k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -302,28 +318,31 @@ def prettyPrintKast(kast, symbolTable):
...
     if isKSequence(kast):
-        unparsedItems = [ prettyPrintKast(item, symbolTable) for item in kast['items'] ]
-        unparsedKSequence = '\n~> '.join(unparsedItems)
-        if len(unparsedItems) > 0:
-            unparsedKSequence = '    ' + unparsedKSequence
+        if len(kast['items']) == 0:
+            return prettyPrintKast(KConstant(klabelEmptyK), symbolTable, debug = debug)
+        if len(kast['items']) == 1:
+            return prettyPrintKast(kast['items'][0], symbolTable, debug = debug)
+        unparsedKSequence = '\n~> '.join([ prettyPrintKast(item, symbolTable, debug = debug) for item in kast['items'][0:-1] ])
```
radumereuta, line 341: You don't want to indent the `~>`?

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -165,12 +193,94 @@ def addOccurance(k):
...
+    for i in l2:
+        if not i in common:
+            newL2.append(i)
+    return (common, newL1, newL2)
+
+def propogateUpConstraints(k):
+    """Try to propogate common constraints up disjuncts.
+
+    -   Input: kast disjunct of constrained terms (conjuncts).
+    -   Output: kast where common constraints in the disjunct have been propogated up.
```
radumereuta: `propagated` Typo? Multiple locations, function names and comments.

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -206,22 +321,28 @@ def _collapseDots(_kast):
...
                 return lhs
-            if  isKApply(lhs) and isKApply(rhs) and lhs["label"] == rhs["label"] and isCellKLabel(lhs["label"]) \
-            and len(lhs["args"]) == len(rhs["args"]):
-                    newArgs = [ KRewrite(lArg, rArg) for (lArg, rArg) in zip(lhs["args"], rhs["args"]) ]
-                    return KApply(lhs["label"], newArgs)
+            if isKVariable(lhs) and isKVariable(rhs) and lhs['name'] == rhs['name']:
+                return lhs
+            if  isKApply(lhs) and isKApply(rhs) and lhs['label'] == rhs['label'] and len(lhs['args']) == len(rhs['args']):
+                    newArgs = [ KRewrite(lArg, rArg) for (lArg, rArg) in zip(lhs['args'], rhs['args']) ]
+                    return KApply(lhs['label'], newArgs)
```
radumereuta: Is this correctly indented? Seems a bit off from the rest.

radumereuta (approved):
1. Several questions (inline).
2. This was kind of long. I didn't go too deep into the logic. Even so, took me about 2 hours.
Overall looks good, but please check the inline comments before merging.

k-distribution/src/main/scripts/lib/pyk/kast.py
ehildenb: Whatever is surrounding the `~>` will take care of indenting it (at newlines) properly. We just need to insert the newlines to tell it where to indent.

So by default, the `~>` will line up on the next line with the first item. For example:

```
           <k>
             ( . => JUMPI 16 bool2Word( CALLVALUE_CELL ==Int 0 )
             ~> #pc[ JUMPI ] )
             ~> #execute
             ...
           </k>
```

Here the indent level is set by `<k>` to be indented once (2 spaces). Then that's used as the base indent level for all newlines produced inside (so both the nested `~>` and teh non-nested `~>` get the same indent).

This probably isn't ideal, but it's readable enough.

k-distribution/src/main/scripts/lib/pyk/kastManip.py
ehildenb: Fixed.

k-distribution/src/main/scripts/lib/pyk/kastManip.py
ehildenb: Fixed.

PR #2209, opened 2021-09-16
Add #kompiledDirectory hook to K-REFLECTION

### LINKED ISSUE #1889: Metadata about *-kompiled directory available as hooked functions in llvm backend
Currently, the Algorand semantics passes in the `*-kompiled` directory (and other arguments) as configuration variables in order to be able to load other test files etc...

The LLVM backend could conceivably provide a hook `#kompiledDirectory()` which returns this directory so it does not need to be passed in as a configuration variable.

Baltoli:
This PR implements and tests the `#kompiledDir` hook suggested in #1889. The changes to the frontend are minimal; most of the work has already been done in the LLVM backend:
* Adds a hooked function to `K-REFLECTION`, along with the associated documentation in `domains.md`.
* Integration tests to check that the correct directory path is returned when the function is evaluated on a kompiled definition.

Depends on #2208 for the matching LLVM backend feature

Closes #1889


radumereuta:
lgtm, but @dwightguth can you do the final review, since it involves the llvm backend also.

dwightguth (approved):

Issue #1957, opened 2021-05-04
Latest z3 version makes java kprove to fail


radumereuta:
I've updated to Ubuntu 21.04 which comes by default with z3-4.8.10, which crashes java kprove with
`ArrayIndexOutOfBoundsException: Index 15 out of bounds for length 15` in `FastRuleMatcher.java`
Failing test: https://github.com/kframework/k/tree/master/k-distribution/tests/regression-new/proof-tests/deposit/spec
I tried z3-4.8.11 (not yet released) and it doesn't crash, but it gives a different result.

As a workaround, I've manually built z3-4.8.7

ehildenb:
Use `--debug-z3-queries` to retrieve the query made to Z3 which causes the change in behavior. You can dump the Z3 query to a file `test.z3`, and call `z3 test.z3` on it. See what different versions of Z3 report on that query.

If Z3 does change its behavior on the query, we should report it.

Otherwise, the correct fix is to migrate this test to the Haskell backend.

radumereuta:
```
(define-sort IMap () (Array Int Int))
(declare-fun isValidBranch (IMap Int) Bool)
(declare-fun isValidZeros (IMap Int) Bool)
(declare-fun |K_138| () Int)
(declare-fun |TREE_HEIGHT_137| () Int)
(declare-fun |M_139| () Int)
(declare-fun |Branch_135| () IMap)
(declare-fun |Zeros_136| () IMap)
(assert (and
  (and
    (= (>= |K_138| |TREE_HEIGHT_137|) true)
    (= (<= |K_138| |TREE_HEIGHT_137|) true)
    (= (< |M_139| (^ 2 |TREE_HEIGHT_137|)) true)
    (= (isValidBranch |Branch_135| |M_139|) true)
    (= (isValidZeros |Zeros_136| |M_139|) true)
    (= (>= |K_138| 0) true)
    (= (>= |M_139| 0) true)
    (= (>= |TREE_HEIGHT_137| 0) true)
  )
  (not (and
    (= (<= (+ |M_139| 1) (^ 2 |K_138|)) true)
  ))
))

(check-sat)
```
z3-4.8.7 -> unsat
z3-4.8.8 -> unsat
z3-4.8.9 -> unsat
z3-4.8.10 -> unknown
z3-master -> unknown (0c6722f48b2814e44ed98ea33b2a356579fd90f4)

There are a bunch of queries like this one, which switches from `unsat` to `unknown`. This is the first one.

[...]

PR #2210, opened 2021-09-20
Add explicit to_int cast to ^Int SMT hook

Baltoli:
This was recommended as a workaround for #2154 and #1957 by the Z3
maintainers (https://github.com/Z3Prover/z3/issues/5243).

The original issue arises because Z3 cannot assume that the exponent is
non-negative, and so must infer that the result could potentially have
type Real after evaluating the exponentiation. By adding the to_int
cast, we opt in to integer semantics everywhere ^Int is used.

Additionally, this commit adds a simplification rule for integer
exponentiation that sends negative exponents to zero directly from K.

### DIFF
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -930,7 +930,7 @@ You can:
 ```k
   syntax Int ::= "~Int" Int                     [function, klabel(~Int_), symbol, functional, latex(\mathop{\sim_{\scriptstyle\it Int}}{#1}), hook(INT.not), smtlib(notInt)]
                > left:
-                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook((to_int (^ #1 #2))), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
+                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook((ite (< #2 0) 0 (to_int (^ #1 #2)))), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
                | Int "^%Int" Int Int            [function, klabel(_^%Int__), symbol, left, smt-hook((mod (^ #1 #2) #3)), hook(INT.powmod)]
                > left:
                  Int "*Int" Int                 [function, functional, klabel(_*Int_), symbol, left, smt-hook(*), latex({#1}\mathrel{\ast_{\scriptstyle\it Int}}{#2}), hook(INT.mul)]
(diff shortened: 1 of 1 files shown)


radumereuta:
Can you do some tests with different Z3 versions? We want to know which version to recommend in the future.
In my original issue, I found that z3-4.8.10 wasn't working, but after that, it might work.

Baltoli:
Yes, I'll verify that this works with different releases of Z3.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -930,7 +930,7 @@ You can:
 ```k
   syntax Int ::= "~Int" Int                     [function, klabel(~Int_), symbol, functional, latex(\mathop{\sim_{\scriptstyle\it Int}}{#1}), hook(INT.not), smtlib(notInt)]
                > left:
-                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook(^), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
+                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook((to_int (^ #1 #2))), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
```
ehildenb: We need to not use this encoding if the second argument is `<Int 0`. Because this is declared in a different module than the rule you added below, we are not guaranteed that a user will have the below simplification rule apply prior to sending to SMT.

If the second argument is `<Int 0`, then the overall result should be `0`. We could encode this to Z3 as `(ite (< #2 0) (0) (to_int (^ #1 #2)))`.

ehildenb: We also need to check that `to_int` is _truncating_, not _rounding_.

Baltoli: For positive arguments, the definition of `to_int` is truncating rather than rounding (see documentation [here](http://smtlib.cs.uiowa.edu/theories-Reals_Ints.shtml))

Baltoli:
As Radu identified in the original issue, the `to_int` workaround requires a newer version of Z3 (`>= 4.8.11` in my testing - see table below, where I tested the branch against source builds of Z3). The CI failure for this change seems to be because Ubuntu 20.04 supplies Z3 4.8.7 by default.

This means that this change as proposed would require an upgrade of the required Z3 version, which may not be practical.

| Z3 Version | Working? |
| -------------: | :-------------: |
| `4.8.7` | ✗ |
| `4.8.8` | ✗ |
| `4.8.9` | ✓ |
| `4.8.10` | ✗ |
| `4.8.11` | ✓ |
| `4.8.12` | ✓ |
| `HEAD` | ✓ |

radumereuta:
That's a problem because we need to decide which versions of Z3 we want to support.
And we need to synchronize with the haskell backend.
@dwightguth @ehildenb any thoughts? Or should we have this discussion in the front-end meeting?

ehildenb:
Hmmm, I guess we should test that all the existing semantics work with the bumped version, before going forward with it.

This will become a hard requirements for users then, because if they want to use the Haskell backend and the `^Int` symbol they'll build smt queries with that symbol present.

radumereuta:
Let's try to update to `z3-4.8.11`.
We need to update the docker files to the semantics to see if they all pass.
evm-semantics
wasm-semantics
Michelson-semantics
Algorand-sc-semantics
IELE-semantics
If the docker file doesn't specify the Z3 version, ping Everett.

Baltoli:
Superseded by #2225 via cherry-picked commits; closing

PR #2214, opened 2021-09-20
K-REFLECTION: add #unparseKORE hook

### LINKED ISSUE #2183: `#unparseKore` hook
We currently have `#parseKore`, which can take a Kore term and parse it into the current runtime, but we need a counterpart call `#unparseKore`:

```
syntax {Sort} String ::= #unparseKore ( Sort ) [function, hook(KREFLECTION.unparseKORE)]
```

Currently the only way to do this is via several calls to `#system` to invoke `kast` on the pretty-output version of the string.

SchmErik:
This hook takes a term and prints the AST as a string.

Fixes: #2183


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -2115,6 +2115,7 @@ unsupported in modern K. There are three exceptions:
   equal to the sort where the function appears.
 * `#kompiledDirectory` - Returns the path to the current compiled K definition
   directory.
+* `#unparseKORE` = Takes a K term and converts it to a string.
```
Baltoli, line 2118: Above it mentions three exceptions; this is the fourth

Baltoli (dismissed):

dwightguth (approved):
I'm approving, but you should address Bruce's comment before merging.

PR #2216, opened 2021-09-21

### LINKED ISSUE #2119: [kompile] should give more readable error message when rejecting token sorts with constructors
```
module TEST
  syntax Foo ::= foo()
  syntax Foo ::= bam()
  syntax Foo ::= fu()
  syntax Foo ::= "bar" [token]
               | r"ba." [token, prec(3)]

  rule foo() => .K
  rule bam() => .K
  rule fu() => .K
  rule bar => .K
  rule baz => .K
endmodule

$ kompile test.k

Error:
  module 'TEST':
definition.
Examine output to see errors.
```

SchmErik:
Token sorts should only contain productions that have `token` or
`function` labels. Any other label (including subsorts) are illegal. Add
a check to ensure that token definitions are correct.

Fixes: #2119

### DIFF
--- a/kore/src/main/scala/org/kframework/definition/outer-ext.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer-ext.scala
@@ -64,6 +64,7 @@ object FlatModule {
         val newImports = m.imports.map(f)
         val newM = new Module(m.name, newImports, m.localSentences, m.att)
         newM.checkSorts()
+        newM.checkTokenSorts()
         newM
       })
     }
--- a/kore/src/main/scala/org/kframework/definition/outer.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer.scala
@@ -363,6 +363,21 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     case _ =>
   }
 
+  def checkTokenSorts () = {
+    val localTokens = localProductions
+      .collect({ case p if p.att.contains("token") => p })
+      .groupBy(_.sort)
+      .map { case (s, ps) => (s, ps) }
+    // Token sorts should only contain productions with function or token labels. Any other label is incorrect.
+    val nontokens = productions.filter(p => !p.att.contains("function") && !p.att.contains("token"))
+    val conflicts = nontokens.filter(p => localTokens.contains(p.sort))
+    conflicts foreach {
+      case p: Production => throw KEMException.compilerError(
+        "Sort "+ p.sort.name + " was declared as a token. Productions of this sort can only contain [function] or [token] labels.", p)
+      case _ =>
+    }
+  }
+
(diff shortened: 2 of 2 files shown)


kore/src/main/scala/org/kframework/definition/outer-ext.scala
```diff
@@ -64,6 +64,7 @@ object FlatModule {
         val newImports = m.imports.map(f)
         val newM = new Module(m.name, newImports, m.localSentences, m.att)
         newM.checkSorts()
+        newM.checkTokenSorts()
```
radumereuta, line 67: There are two locations where we call `checkSorts()`. Why are we calling `checkTokenSorts()` only once?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -363,6 +363,21 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     case _ =>
   }
 
+  def checkTokenSorts () = {
+    val localTokens = localProductions
+      .collect({ case p if p.att.contains("token") => p })
+      .groupBy(_.sort)
+      .map { case (s, ps) => (s, ps) }
```
radumereuta: This can be compacted further. Haven't tested it but this might be equivalent:
```
localProductions.filter(p => p.att.contains("token")).groupBy(_.sort)
```

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -363,6 +363,21 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
...
+    // Token sorts should only contain productions with function or token labels. Any other label is incorrect.
+    val nontokens = productions.filter(p => !p.att.contains("function") && !p.att.contains("token"))
+    val conflicts = nontokens.filter(p => localTokens.contains(p.sort))
+    conflicts foreach {
+      case p: Production => throw KEMException.compilerError(
+        "Sort "+ p.sort.name + " was declared as a token. Productions of this sort can only contain [function] or [token] labels.", p)
+      case _ =>
+    }
```
radumereuta, line 379: This can also be compacted further.
```
productions
.filter(c1 & c2)
.map(p => throw)
```

radumereuta (changes requested):
This could use a test.

kore/src/main/scala/org/kframework/definition/outer-ext.scala
SchmErik: Oh right, thanks for the reminder. I'll extract these two lines in a new function and place that in both places that call checkSorts()

SchmErik:
What's the expected behavior in this case?
```
  syntax X ::= "a"
  syntax X ::= "b" [token]
```
This current PR, this results in an error on these two lines

SchmErik:
syntax X ::= "a"
  syntax X ::= "b" [token]

SchmErik:
I made several large modifications to my solution so I'm going to close this and open a different pull request.

PR #2218, opened 2021-09-22
Expose floating-point truncation hook

Baltoli:
This function mimics the style of the floor and ceiling operations, nd
is a direct passthrough to the MPFR implementation of floating-point
truncation.

There don't seem to be any integration tests for ceil / floor at the
moment; they're only unit-tested in the backend. If it would be useful
to have tests for all these functions, I can implement them.

Relies on https://github.com/kframework/llvm-backend/pull/429 being merged, for the corresponding LLVM backend hook implementation.


radumereuta (approved):
lgtm.
You should merge it after the llvm-backend updates though.

Baltoli:
Thanks - it won't even build in CI until the backend change is merged so no worries on that front.

dwightguth (approved):

PR #2225, opened 2021-09-24
Update Dockerfiles to bump standard Z3 version to 4.8.11

Baltoli:
We were using the package manager-supplied version of Z3 in several Debian / Ubuntu Dockerfiles. This PR is an attempt to bring these all up to 4.8.12 (to deal with #2154 and #2210); when apt supplies a new enough version we can revert to using that one.

As best I can tell, Arch and macOS supply 4.8.12 through pacman and Homebrew respectively already, so don't need the manual installation.

To get the CI passing, I've had to cherry-pick the commits from #2210 into this PR (#2210 is therefore superseded by this PR and should be closed once this is merged); upgrading Z3 exposes the test failure that led to this problem in the first place, and so it needs to happen at the same time as the `to_int` fix.


Baltoli:
Note that we probably want to wait until the downstream projects have bumped their versions of Z3 to merge this. My progress on doing so:

* [x] evm-semantics (https://github.com/kframework/evm-semantics/pull/1137)
* [x] wasm-semantics (https://github.com/kframework/wasm-semantics/pull/436)
* [x] michelson-semantics (https://github.com/runtimeverification/michelson-semantics/pull/331)
* [x] algorand-sc-semantics (https://github.com/runtimeverification/algorand-sc-semantics/pull/414)
* [x] IELE-semantics (https://github.com/runtimeverification/iele-semantics/pull/322)

Baltoli:
As it turns out, there's a transient prover bug in Z3 4.8.12 that breaks the Michelson test suite. To address this, we're dropping back to Z3 4.8.11.

package/docker/Dockerfile.ubuntu-bionic
```diff
@@ -10,13 +10,13 @@ RUN    apt-get update                   \
...
-    && python scripts/mk_make.py                                    \
-    && cd build                                                     \
-    && make -j8                                                     \
-    && make install                                                 \
-    && cd ../..                                                     \
+RUN    git clone 'https://github.com/z3prover/z3' --branch=z3-4.8.11 \
+    && cd z3                                                         \
+    && python scripts/mk_make.py                                     \
+    && cd build                                                      \
+    && make -j8                                                      \
```
ehildenb, line 17: I guess we actually need to adjust the packgaing rules...

ehildenb: @dwightguth how do we want to approach this? Should K bundle its own Z3, or we just have instructions for people that they must separately install the correct Z3.

Baltoli:
Marking this as a draft because Z3 `v4.8.11` is causing a significant performance regression in the EVM semantics. We don't want to go ahead with the version bump until we know what's causing this, and can put a fix in place.

The issue tracking this regression is https://github.com/kframework/k/issues/2247

Baltoli:
@ehildenb @dwightguth This is ready to go now that the individual semantics have all been merged. There's a comment above regarding the packaging strategy that we may want to address, however - let me know what the best thing to do there is.

ehildenb:
I think the only realistic solution is to just update all the documentation (so README and install guide) to mention that users are expected to separately install a newer version of z3 if they don't have the correct version.

Baltoli:
OK @ehildenb - I'll add a stanza to the README and install guide to document that we expect users to separately install z3 version 4.8.11.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1055,6 +1055,8 @@ module INT-SYMBOLIC [symbolic]
   rule X modInt N => X requires 0 <=Int X andBool X <Int N [simplification]
   rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]
 
+  rule _ ^Int N => 0 requires N <Int 0 [simplification]
+
```
ehildenb: I'd like to see a test that exercises this new Z3 encoding. Anything that sends the symbol to Z3 on the Haskell backend should do, because it was causing crashes previously.

```
requires "domains.md"

module TEST
    imports BOOL
    imports INT-SYMBOLIC

    syntax KItem ::= runLemma ( Bool ) | doneLemma ( Bool )
 // -------------------------------------------------------
    rule <k> runLemma(B) => doneLemma(B) ... </k>

    rule N ^Int N' <=Int M ^Int M' => true
      requires 0 <Int N andBool 0 <Int N'
       andBool 0 <Int M andBool 0 <Int M'
       andBool N <=Int M andBool N' <=Int M'
      [simplification, smt-lemma]
                     
endmodule
                     
module TEST-SPEC
    imports TEST
                     
    claim <k> runLemma(3 ^Int X <=Int 3 ^Int Y) => doneLemma(true) ... </k>
      requires 0 <Int X andBool 0 <Int Y
       andBool X <=Int Y
endmodule
```

I'd also like Daejun to double-check this new Z3 encoding of `_^Int_`, to make sure we are getting the correct semantics.

ehildenb: Ping @daejunpark

ehildenb:
@ACassimiro ping.

k-distribution/include/kframework/builtin/domains.md
daejunpark: @ehildenb @Baltoli  The simplification rule, `N < 0  \implies  X ^ N == 0`, doesn't seem correct, when X is 1 or -1.   When X = 1, it is 1.  When X = -1, it is alternating between 1 (if N is even) and -1 (if N is odd).

ehildenb: Hmmm, is `-1` the only special case? Maybe we should just remove this simplification rule altogether.

For `-2`, the result is `1/I`, which I guess should go to 0 for `1 <Int I`. I guess it's not worth adding these axioms yet until we see that we need them.

So let's remove this rule altogether.

daejunpark: @ehildenb Note that it is not only for the simplification rule, but also for the z3 encoding in the hook: `(ite (< #2 0) 0 (to_int (^ #1 #2)))`.   As I mentioned, this encoding is incorrect for `#1` being 1 or -1.

Baltoli: So do we want to outright remove the simplification rule, and add another check to the SMT hook that handles the cases for 1 / -1?

ehildenb: Well, I think we could also change the smt hook to say `(and (< 0 #1) (< 0 #2))`, so that we don't have to think about the 0/negative cases. Then we could add K rules for specifically the various 0 cases `N ^ 0 requires 0 <Int N`, `0 ^ N requires 0 <Int N`, `0 ^ 0`.

daejunpark: Let me clarify a bit, to make sure we are all on the same page:

Let `X` and `N` be an integer.  Let `Y` be `\trunc(X ^ N)`.  
(`\trunc` means the truncation rounding, e.g., `\trunc(-0.1) = 0`, `\trunc(0.1) = 0`, `\trunc(1.2) = 1`, etc.)

Then we have:
1. `X = 0` implies `( if N > 0, then Y = 0, else ( if N < 0, then Y = \infty, else Y = ??? ) )`   (see below for `0^0`)
2. `X = 1` implies `Y = 1`
3. `X = -1` implies  `( if N is even, then Y = 1, else Y = -1 )`
4. `(X > 1 or X < -1) and N < 0` implies `Y = 0`

Note that:
- `N = 0`  does _not_ necessarily imply  `Y = 1`  until we define `0^0 = 1`.
- `N < 0`  does _not_ imply  `Y = 0`.   (See (1), (2), and (3) above for the counter-examples.)

**Remark**:
Regarding `0^0`, many mathematicians seem to agree on `0^0 = 1` (see [here](https://www.maa.org/book/export/html/116806)), while WolframAlpha considers it [undefined](https://www.wolframalpha.com/input/?i=0%5E0).  We need to double-check what's the interpretation of (a specific version of) z3 for that.

daejunpark: My simple suggestion would be to just put a guard for `to_int`, like: `(implies (>= #2 0) (to_int (^ #1 #2)))`.

Those who want to reason about the negative exponent case, would need to come up with their own lemmas, simplifications, and/or custom z3 encodings, I think.

virgil-serbanuta: FWIW, I think that if the type system contains both `+\infty` and `-\infty`, then `0^N` should be undefined when `N < 0`

ACassimiro: Just adding one point here: as mentioned by @daejunpark, we should double-check the interpretation of `0 ^ 0`, and, if I'm not mistaken, ```**``` is underlying implementation of ```^``` for z3, correct? I've made a script to test `x=0`, `y=0`, `n=x**y` using different versions of z3's python module. Bellow is its output (version followed by the generated model, if can be achieved):
```

4.4.2.1.post1: [n = 0**0, y = 0, x = 0]
4.5.1.0: [n = 0**0, y = 0, x = 0]
4.5.1.0.post1: [n = 0**0, y = 0, x = 0]
4.5.1.0.post2: [n = 0**0, y = 0, x = 0]
4.8.0.0.post1: [n = 0**0, y = 0, x = 0]
4.8.5.0: [n = 0**0, y = 0, x = 0]
4.8.6.0: [n = 0**0, y = 0, x = 0]
4.8.7.0: [n = 0, y = 0, x = 0]
4.8.8.0: [n = 0, y = 0, x = 0]
4.8.9.0: [n = 0, y = 0, x = 0]
4.8.10.0: unknown
4.8.11.0: unknown
4.8.12.0: unknown
```

In this case, we may end up getting different outcomes for different versions of z3, so I think a more safe approach would be something on the lines of what @ehildenb said (adding K rules for the specific 0 and negative cases).

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1055,7 +1059,13 @@ module INT-SYMBOLIC [symbolic]
...
   rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]
 
-  rule _ ^Int N => 0 requires N <Int 0 [simplification]
+  rule  1 ^Int _ =>  1                           [simplification]
+  rule -1 ^Int N =>  1 requires N %Int 2 ==Int 0 [simplification]
+  rule -1 ^Int N => -1 requires N %Int 2 ==Int 1 [simplification]
+  rule  0 ^Int N =>  0 requires N >Int 0         [simplification]
+  rule  X ^Int N =>  0 requires absInt(X) >Int 0
+                        andBool N <Int 0         [simplification]
+  rule  X ^Int 0 =>  1 requires X =/=Int 0       [simplification]
```
Baltoli: @ehildenb @daejunpark I've implemented the special cases for integer exponentiation per the [document here](https://github.com/runtimeverification/security/blob/61e19e0830960dc048f7dd8e4f42b06be186e792/knowledge-base/exponentiation.md).

The SMT encoding will only send `_^Int_` to Z3 if both `X` and `N` are strictly positive. In the cases where that might not be the case, these simplification rules will apply if we can define the result.

Cases where the result of the exponentiation would be ambiguous (`0 ^ 0`) or infinite (`0 ^ -1`) are left undefined; if users need to define these cases they can supply their own axioms.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1055,6 +1059,14 @@ module INT-SYMBOLIC [symbolic]
   rule X modInt N => X requires 0 <=Int X andBool X <Int N [simplification]
   rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]
 
+  rule  1 ^Int _ =>  1                           [simplification]
+  rule -1 ^Int N =>  1 requires N %Int 2 ==Int 0 [simplification]
+  rule -1 ^Int N => -1 requires N %Int 2 ==Int 1 [simplification]
+  rule  0 ^Int N =>  0 requires N >Int 0         [simplification]
+  rule  X ^Int N =>  0 requires absInt(X) >Int 0
```
daejunpark: Shouldn't be `absInt(X) >Int 1`?   (Note that the simplification rules are not applied in the order.)

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1055,6 +1059,14 @@ module INT-SYMBOLIC [symbolic]
   rule X modInt N => X requires 0 <=Int X andBool X <Int N [simplification]
   rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]
 
+  rule  1 ^Int _ =>  1                           [simplification]
+  rule -1 ^Int N =>  1 requires N %Int 2 ==Int 0 [simplification]
+  rule -1 ^Int N => -1 requires N %Int 2 ==Int 1 [simplification]
```
daejunpark: I'm not sure if `N %Int 2` will return 1, for negative even numbers, e.g., -2, -4, ....   Some backends may return `-1`, depending on their semantics of modulo.  (Some programming languages do!)  You may want to say that it is _not equal_ to 0, to be more general.

k-distribution/include/kframework/builtin/domains.md
Baltoli: Good catch! You're absolutely right

k-distribution/include/kframework/builtin/domains.md
Baltoli: Got it - will address. Thanks!

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -930,7 +934,7 @@ You can:
 ```k
   syntax Int ::= "~Int" Int                     [function, klabel(~Int_), symbol, functional, latex(\mathop{\sim_{\scriptstyle\it Int}}{#1}), hook(INT.not), smtlib(notInt)]
                > left:
-                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook(^), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
+                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook((ite (and (< 0 #1) (< 0 #2)) 0 (to_int (^ #1 #2)))), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
```
daejunpark, line 937: I'm not sure what you're trying to encode here.  This encoding means `if (> #1 0) and (> #2 0), then 0, else (to_int (^ #1 #2))`, which doesn't sound correct to me.

daejunpark: ~~You might want to do this?~~ 
_**UPDATE: please ignore this.  it doesn't work, type-error.**_
```suggestion
                 Int "^Int" Int                 [function, klabel(_^Int_), symbol, left, smt-hook((implies (and (< 0 #1) (< 0 #2)) (to_int (^ #1 #2)))), latex({#1}\mathrel{{\char`\^}_{\!\scriptstyle\it Int}}{#2}), hook(INT.pow)]
```

daejunpark: Also, if you have `(< 0 #2)`, you don't need to add `(< 0 #1)`, right?  I mean, `X ^ N` will be well-defined for all X, if N > 0, right?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1055,6 +1059,14 @@ module INT-SYMBOLIC [symbolic]
   rule X modInt N => X requires 0 <=Int X andBool X <Int N [simplification]
   rule X   %Int N => X requires 0 <=Int X andBool X <Int N [simplification]
 
+  rule  1 ^Int _ =>  1                            [simplification]
+  rule -1 ^Int N =>  1 requires N %Int 2 ==Int  0 [simplification]
+  rule -1 ^Int N => -1 requires N %Int 2 =/=Int 0 [simplification]
+  rule  0 ^Int N =>  0 requires N >Int 0          [simplification]
+  rule  X ^Int N =>  0 requires absInt(X) >Int 1
+                        andBool N <Int 0          [simplification]
+  rule  X ^Int 0 =>  1 requires X =/=Int 0        [simplification]
```
ehildenb, line 1068: Have we double-checked that `_%Int_` is behaving correctly on negative numbers here?

I remember that `_%Int_` and `_modInt_` had different behaviors for negative inputs on the first number, I just want to make sure we either get the right one, or write this code so that it doesn't matter which one.

Baltoli:
Marking as a draft until #2307 and #2308 are resolved.

radumereuta:
Reopen when more SMT work is done. Maybe use a newer version also.

PR #2226, opened 2021-09-24
Record prods macros

### LINKED ISSUE #2217: [Bug] [kompile] - macro for record production is applied to itself
K Version
---------

Example:

```
$ kompile --version
K version:    v5.1.188-4-g6f2b343a8f
Build date:   Tue Sep 21 16:51:08 UTC 2021
```

Description
-----------

A `macro` rule for a record production leads to `kompile` trying to attempt macro expansion on the macro itself.

Input Files
-----------

File `test.k`:

```
requires "domains.md"

module TEST
    imports INT

    syntax Int ::= f ( a: Int, b: Int )
 // -----------------------------------
    rule f(X, Y) => X +Int Y [macro]

endmodule
```

Reproduction Steps
------------------

Kompile the definition:

```
$ kompile --backend haskell test.k --main-module TEST
[Error] Compiler: Illegal function symbol _+Int_ on LHS of rule. Consider
adding `simplification` attribute to the rule if this is intended.
        Location(10,21,10,29)
[Error] Compiler: Had 1 structural errors after macro expansion.
  while executing phase "expand macros" on sentence at
        <none>
        <none>
```

Expected Behavior
-----------------

I expect either (i) the kompiler gives us a better error, saying "macros are not allowed for record productions", or (ii) it handles this correctly, and does not attempt to apply the macro to itself.

radumereuta:
Fixes: #2217
- move isMacro to outer.Module
- don't generate sort projections for macro symbols
- tests


ehildenb:
Looks good to me, but @dwightguth and @SchmErik should review too.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -238,20 +238,22 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
...
-    r.body match {
-      case Unapply.KApply(Unapply.KLabel("#withConfig"), Unapply.KApply(s, _) :: _) => s
-      case Unapply.KApply(Unapply.KLabel("#withConfig"), Unapply.KRewrite(Unapply.KApply(s, _), _) :: _) => s
-      case Unapply.KApply(s, _) => s
-      case Unapply.KRewrite(Unapply.KApply(s, _), _) => s
-      case _ => KORE.KLabel("")
-    }
-  })
+  lazy val rulesFor: Map[KLabel, Set[Rule]] = rules.groupBy(r => matchKLabel(r))
+  lazy val macroKLables: Set[KLabel] = rules.filter(r => r.isMacro).map(r => matchKLabel(r))
```
SchmErik: I've been trying to integrate this code into a separate issue and this code isn't getting the expected result for me. The problem is that `rule` productions containing macros is a "bubble" rather than a "rule"...

SchmErik: nevermind - this bubble/rule issue is due to me trying to evaluate this logic too early in the kompile pipeline

SchmErik (approved):
Looks good to me. I ended up using the macro functions that you wrote in this PR and it works well

dwightguth (approved):

PR #2234, opened 2021-09-28

### LINKED ISSUE #2119: [kompile] should give more readable error message when rejecting token sorts with constructors

SchmErik:
Token sorts should only contain productions that have `token` or
`function` labels. Any other label (including subsorts) are illegal.

Fixes: #2119


kernel/src/main/java/org/kframework/compile/checks/CheckTokens.java
```diff
@@ -0,0 +1,43 @@
...
+import org.kframework.definition.Production;
+import org.kframework.definition.Sentence;
+import org.kframework.utils.errorsystem.KEMException;
+
+import java.util.Set;
+
+public class CheckTokens {
+    private final Set<KEMException> errors;
+    private final Module m;
+    private static final ImmutableSet<String> ignoredSortNames = ImmutableSet.of("Bag", "IOFile", "IOInt", "IOString", "K", "KBott", "KItem", "KLabel", "KList", "Sort");
```
radumereuta: Why `Sort`? Where is that coming from?
Not sure about some others also. `KItem` for example.
I just noticed that in the original issue, I complained about the LLVM and Haskell backend. But Java is a bit different. Probably we don't need to run this check there.
Could you check all these, please?

kernel/src/main/java/org/kframework/compile/checks/CheckTokens.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public CheckTokens(Set<KEMException> errors, Module m) {
+        this.errors = errors;
+        this.m = m;
+    }
+
+    // This check ensures that sorts containing token declarations only contain syntax declarations that are also tokens (or macros).
+    public void check(Sentence sentence) {
```
radumereuta, line 23: `private void check(Production p) {`
You can match directly on Production and do away with this check.

kernel/src/main/java/org/kframework/compile/checks/CheckTokens.java
```diff
@@ -0,0 +1,43 @@
...
+                || ignoredSortNames.contains(p.sort().name())) {
+            return;
+        }
+
+        if (!m.tokenProductionsFor().contains(p.sort()) // We only care about sorts that have been declared as tokens.
+                || p.klabel().isDefined() && m.macroKLabels().contains(p.klabel().get())) {
+            return;
+        }
+
+        errors.add(KEMException.compilerError("Sort " + p.sort().name() + " was declared as a token. Productions of this sort can only contain [function] or [token] labels.", p));
```
radumereuta: Line too long.

radumereuta (changes requested):

kernel/src/main/java/org/kframework/compile/checks/CheckTokens.java
SchmErik: ah, thanks that would make things more clear

kernel/src/main/java/org/kframework/compile/checks/CheckTokens.java
SchmErik: I came up with these by running kompile on the tokenCheck.k file and looking at the errors that were generated without ignoring special sort names and `Sort` was one of them

radumereuta: You should check all of them. They might indicate something else is wrong.

SchmErik: oh strange... Now I'm getting that I only need to ignore "KBott" and "KLabel"

SchmErik: I think this was written when I was still working on solidifying the logic. I removed all but KBott and KLabel and I'm rerunning the testsuite locally. I'll push another commit addressing this and the other feedback you have. Thanks!

SchmErik:
this PR got a little hectic with a bunch of commits that ended up getting reverted or were typo fixes. Now that the CI is passing, I'll open yet another PR with cleaner commits.

PR #2238, opened 2021-09-30
kprove: pass branching allowed to kore-exec when running proofs 

### LINKED ISSUE #2224: KProve does not pass `--branching-allowed` through to haskell backend, only java
The option `--branching-allowed` is there to allow limiting the number of concurrent execution paths. This can be useful for terminating proofs early.

Currently, this option is only passed through to the Java backend. But the haskell backend supports the `--breadth` option, so if `--branching-allowed N` is present on the `kprove` or `kprovex` command-line arguments, we should pass `--breadth N` to the Haskell backend.

This should probably work for krun as well I guess, but that's less important/useful.

Currently we have to do `--haskell-backend-command="kore-exec --breadth N"`

SchmErik:
fixes #2224


SchmErik:
@ehildenb I don't know kprove very well but I only added this capability to the `prove` method. Do you think I should add this code to other functions like `execute`?

ehildenb:
We really only need this functionality for proving right now, so I wouldn't bother @SchmErik . So if @radumereuta approves, the changes look good to me!

radumereuta (approved):

PR #2241, opened 2021-09-30

### LINKED ISSUE #2119: [kompile] should give more readable error message when rejecting token sorts with constructors

SchmErik:
Token sorts should only contain productions that have `token`,
`bracket`, or `function` labels or the RHS of the production is a macro
term. Any other label (including subsorts) are illegal.

The java backend is more forgiving when it comes to token sort
declarations. Avoid this check when kompile is running with the java
backend.

Fixes: #2119


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -437,6 +438,8 @@ public void structuralChecks(scala.collection.Set<Module> modules, Module mainMo
 
         stream(modules).forEach(m -> stream(m.localSentences()).forEach(new CheckHOLE(errors, m)::check));
 
+        stream(modules).forEach(m -> stream(m.localSentences()).forEach(new CheckTokens(errors, m, isSymbolic && isKast)::check));
```
radumereuta: I think this check `isSymbolic && isKast` can be done before even entering this line.
So we would avoid this entire step instead of looking at each sentence and decide if we need to do something.

radumereuta (dismissed):

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -437,6 +438,10 @@ public void structuralChecks(scala.collection.Set<Module> modules, Module mainMo
 
         stream(modules).forEach(m -> stream(m.localSentences()).forEach(new CheckHOLE(errors, m)::check));
 
+        if (!(isSymbolic && isKast)) {
```
radumereuta:
```suggestion
        if (!(isSymbolic && isKast)) { // if it's not the java backend
```

radumereuta (approved):

PR #2246, opened 2021-10-01
bug-report: update template

ehildenb:
Our bug report template is a bit verbose. This should make it a bit easier to user.


radumereuta (approved):
Looks better.

PR #2248, opened 2021-10-05
kompile: check rules with smt-lemma labels

### LINKED ISSUE #2193: [Bug] [kompile|kprove] - kompile/kprove should throw an error for smt-lemma rules that can't be encoded
K Version
---------

Example:

```
$ kompile --version
K version:    5.1.0
Build date:   Fri Sep 03 22:19:38 EEST 2021
```

Description
-----------

SMT Lemmas that can't be encoded because they use function symbols without `smtlib` attribute should produce an error.

Input Files
-----------

`tmp.k`

```
module TMP
  imports BOOL
  imports INT
  imports K-EQUAL

  syntax KItem ::= a(Int)
  syntax Int ::= f(Int)  [function, functional, smtlib(asdf)]

  rule f(0) => 0
  rule f(_) => 1  [owise]

  rule f(_) >Int -1 => true  [simplification, smt-lemma]

  rule a(X) => .K requires f(X) >=Int -1

endmodule
```

`proof-tmp.k`
```
module PROOF-TMP
  imports TMP

  claim a(_X) => .K

endmodule
```

Reproduction Steps
------------------

1. Check that the proof passes:
```
kompile tmp.k --backend haskell && kprove proof-tmp.k
```

2. Remove the `smtlib` attribute for `f`:

3. The proof no longer passes:
```
kompile tmp.k --backend haskell && kprove proof-tmp.k
```

Expected Behavior
-----------------

Either `kompile` or `kprove` should complain that the `smt-lemma` rule can't be sent to the SMT solver.

SchmErik:
Rules marked with the smt-lemma label must be structured in such a way
that all terms (aside from tokens) are labeled with `smtlib` or
`smt-hook` attributes. Without these labels, it's not possible to encode
these rules for Z3.

Fixes: #2193


kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
```diff
@@ -0,0 +1,49 @@
...
+        new VisitK() {
+            @Override
+            public void apply(KRewrite k) {
+                super.apply(k.left());
+                super.apply(k.right());
+            }
+
+            @Override
+            public void apply(KApply k) {
+                if (!k.att().toString().matches(".*smtlib.*|.*smt-hook.*")) {
```
radumereuta: I don't think that's how you're supposed to check if it contains those attributes.
https://github.com/kframework/k/blob/master/kore/src/main/scala/org/kframework/attributes/Att.scala#L92

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    private void check(Rule rule){
+        if (!(rule.body() instanceof KRewrite)) {
+            return;
+        }
+        new VisitK() {
+            @Override
+            public void apply(KRewrite k) {
```
radumereuta: Are you sure you have to `@Override` this? The base class seems to have this code already.
https://github.com/kframework/k/blob/master/kore/src/main/scala/org/kframework/kore/transformers.scala#L95

radumereuta (changes requested):

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
dwightguth: Yeah you can use `k.att().contains(Att.SMTLIB())`

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public void check(Sentence sentence) {
+        if ((sentence instanceof Rule) && sentence.att().contains("smt-lemma")) {
+            check((Rule) sentence);
+        }
+    }
+
+    private void check(Rule rule){
+        if (!(rule.body() instanceof KRewrite)) {
```
dwightguth: this seems weird. This is not going to correctly capture whether or not the rule has a rewrite in it. What are you trying to achieve here?

dwightguth (changes requested):
Radu's comments look appropriate to me, but I also had one more.

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
SchmErik: this was left over from a previous change before I found out about the visitors. Before I found out about visitors, I thought I had to do a bunch of these kinds of checks but this isn't required so I removed it. Thanks for catching this

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
SchmErik: removed this method all together. Thanks

kernel/src/main/java/org/kframework/compile/checks/CheckSmtLemmas.java
SchmErik: I ended up using
```
if (m.productionsFor().isDefinedAt(k.klabel()) && !m.productionsFor().get(k.klabel()).get()
         .exists(p -> p.att().contains(Att.SMT_HOOK()) || p.att().contains(Att.SMTLIB()))) {
```
instead. The `k.att().contains(Att.SMTLIB())` doesn't work because the SMTLIB is technically in k.att().att()

radumereuta (commented):
lgtm, but I'll let Dwight give the final approval.

radumereuta (approved):

dwightguth (approved):

PR #2251, opened 2021-10-05
Update LLVM backend and reorder tests

Baltoli:
This is a bit of history rewriting to get the collection-sorting PR
merged without requiring the new garbage collection work to also be
merged.

PR #2254, opened 2021-10-06
Un-revert: Update LLVM backend and reorder tests (#2251)

### LINKED ISSUE #2211: Stabilize pretty printed configurations
Is it possible to sort terms inside a pretty printed configuration from krun output?
Sometimes, when we make changes to seemingly unrelated parts, we get different [outputs](https://github.com/kframework/k/pull/2209/commits/ccff3618b1c548017b386975db7ff93dfad6f0f8).
We should be able to sort `Map`, `Set`, and cells with `multiplicity="*"`.

Baltoli:
This is a bit of history rewriting to get the collection-sorting PR
merged without requiring the new garbage collection work to also be
merged.

@dwightguth this is the remade PR; can be reviewed and merged once the nix / ld / macos issue is fixed.

### DIFF
--- a/k-distribution/tests/regression-new/imp++-llvm/div.imp.out
+++ b/k-distribution/tests/regression-new/imp++-llvm/div.imp.out
@@ -5,8 +5,8 @@
           .
         </k>
         <env>
-          y |-> 1
           x |-> 0
+          y |-> 1
         </env>
         <id>
           0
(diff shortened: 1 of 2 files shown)


dwightguth:
I'm going to go ahead and use my admin powers to merge this as the test updates in this PR are necessary in order to get my PR that updates the NixOS version passing. While MacOS Nix is currently not passing, this is already not passing on master, so it's not actually a regression.

PR #2262, opened 2021-10-08
kompile: allow macros on productions

SchmErik:
Before this change, macros were only allowed on rules like so:

syntax Int ::= "this"
rule this => 3 [macro]

This change allows macros on productions and removes the need for
the macro label on rules like so:

syntax Int ::= "this" [macro]
rule this => 3

This is done by propagating the macro label to the appropriate rule
before macro expansion.


SchmErik:
This was a draft PR and is created from my fork. Closing this and opening #2272

PR #2269, opened 2021-10-14
More complete shell.nix in README

mpetruska:
This proposition is made in an effort to help people using Nix getting started with new K projects. The modified shell.nix example file has all the dependencies, so no pre-installation of packages are needed when using Nix.


calin1304:
That section of the README states

> To create a development environment for a project that __depends on the K Framework,__ you can add a `shell.nix` based on this template

so the template `shell.nix` is just to have the K binaries available in a shell.

I think adding other packages to the shell is up to the user, so there is no need list them in this template.

README.md
```diff
@@ -262,12 +262,17 @@ Framework, you can add a `shell.nix` based on this template:
 
 ```.nix
 # shell.nix
+{ pkgs ? import <nixpkgs> {} }:
+
+with pkgs;
 let
   kframework = import ./path/to/k {};
   inherit (kframework) mkShell;
```
calin1304: I also noticed this template might fail because `clang` is mentioned in `buildInputs` but it is not inherited from the `kframework` Nix package.

```suggestion
  inherit (kframework) mkShell clang;
```

ehildenb:
@mpetruska we need some more context on your use cases to decide if we should include this. Specifically, as @calin1304 said, the current use case for `shell.nix` seems different for you than what we originally intended for it.

Would it be more appropriate to have a separate `env.nix` which includes `shell.nix` but adds these dependencies? Or can you link to some standard or documentation that says we should be using your approach for `shell.nix`?

ehildenb:
Pleas re-open if you provide more clarification.

Issue #2247, opened 2021-10-04
Type inference is very slow because of z3


radumereuta:
When parsing KEVM claims the time spent doing type inference is larger than usual. For example:
```
kprovex --directory tests/specs/erc20/verification/haskell tests/specs/erc20/ds/transferFrom-failure-1-d-spec.k -I /home/radu/work/evm-semantics/.build/usr/lib/kevm/include/kframework -I /home/radu/work/evm-semantics/.build/usr/lib/kevm/blockchain-k-plugin/include/kframework --profile-rule-parsing timings/tests/specs/erc20/ds/transferFrom-failure-1-d-spec.k.log --format-failures -v --dry-run
```
z3-4.8.7: Parse spec modules [1/1 rules]                               = 30.434s
26305 /home/radu/work/evm-semantics/tests/specs/erc20/ds/transferFrom-failure-1-d-spec.k:8   parse: 287 typeInf:25476
z3-4.8.11: killed after ~5 minutes

Parsing takes 287ms, and type inference 25s. When I switch to z3-4.8.11 it takes a huge amount of time. I killed it after ~5 minutes.
Times are collected using `--verbose` and `--profile-rule-parsing`.

ehildenb:
Two things that I think could help with this:

- Add an option to print out the z3 queries sent when doing type inference so we can inspect them.
- Timeout Z3 at 10s (it's commonly accepted that if Z3 doesn't come back in a few seconds, it will take forever), and if we do have to kill Z3 during type-inference, print out which variables that are having trouble being inferred so that we can make the rule faster. We can also allow the user to set this timeout, so that if I want to optimize the rules in a given semantics I can set it very low and annotate the sorts which need it to speed up parsing on that definition.

Baltoli:
This seems to still be a problem using a `HEAD` build of Z3, unfortunately (4.8.13).

radumereuta:
I will try to extract a query from Z3 and do a git bisect to see where the regression started.
Although, I'm not happy with the performance before that either.
I will post my findings when I know more for sure. Might take a while though.

radumereuta:
@Baltoli could you update us on the z3 bug now that we improved the runtime?!

radumereuta:
I used git bisect on z3 and found the commit that makes everything slow for the type inference: https://github.com/Z3Prover/z3/commit/b1606487f08cec8d950be078ac31881fc36a7bea
[...]

PR #2271, opened 2021-10-15
Anonymous variables get their type directly

radumereuta:
from the above restriction
Partial fix for: #2247

Tested with evm proofs: `time make test-prove KPROVE_OPTS=--dry-run TEST_SYMBOLIC_BACKEND=haskell -j 12 -O`
z3-4.8.7
Before:
```
real	6m34,792s
user	83m29,499s
sys	1m38,575s
```
Slowest rule was **3m44s** - dsvalue-read-pass-summarize-spec.k. A few other rules over 1m.
After this change:
```
real	3m0,959s
user	73m1,389s
sys	1m32,802s
```
Slowest rule was **15s**.

Updated to z3-4.8.13 (latest master). Same command.
```
real     3m56,514s
user	79m20,607s
sys      1m35,394s
```
Slowest rule was **1m1s** - transferFrom-failure-1-a-spec.k. A few others between 15s and 1m.

It seems this optimization works pretty well for the old version of Z3, and it makes it somewhat acceptable for the latest Z3. I wouldn't update yet though. I still have some changes in mind which I find tricky to implement for now.

I would like to get this into master since it may save people a lot of time.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/inner/disambiguation/TypeInferencer.java
+++ b/kernel/src/main/java/org/kframework/parser/inner/disambiguation/TypeInferencer.java
@@ -567,7 +567,7 @@ public class TypeInferencer implements AutoCloseable {
         Sort oldExpectedSort = expectedSort;
         Optional<ProductionReference> oldExpectedParams = expectedParams;
         TermCons tc = (TermCons)pr;
-        // traverse childrfen
+        // traverse children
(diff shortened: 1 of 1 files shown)


Baltoli (approved):

dwightguth (approved):
This looks good and I think it should be sound since anonymous variables cannot be subject to any constraints that would make this restriction untrue in the maximal solution.

PR #2272, opened 2021-10-15
kompile: allow macros on productions

### LINKED ISSUE #1584: Allow putting `macro` on productions
Rather than requiring `macro` on `rule`, we should allow `macro` on productions (and `macro-rec`), and then that will just be propogated down to the appropriate rules before macro expansion so that none of the frontend logic needs to change.

This makes it possible to:

-   Eventually disallow `macro` on rules, in favor of declaring the function itself a macro.
-   Easier to determine which declared symbols are constructors or not for the purpose of disallowing constructors declared in builtin sorts.

SchmErik:
This pull request enables macro labels to be used on productions rather than on rules.

Fixes: #1584


kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -26,6 +26,22 @@ class Att private (val att: Map[(String, String), Any]) extends AttributesToStri
...
+    }
+    if (contains(Att.ALIAS)) {
+      return Some(Att.ALIAS)
+    }
+    if (contains(Att.ALIAS_REC)) {
+      return Some(Att.ALIAS_REC)
+    }
+    return None;
+  }
+
```
radumereuta, line 44: Is it possible to have multiple labels from this list on the same production?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -234,16 +234,28 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
   lazy val rules: Set[Rule] = sentences collect { case r: Rule => r }
   lazy val rulesAndClaims: Set[RuleOrClaim] = Set[RuleOrClaim]().++(claims).++(rules)
   lazy val rulesFor: Map[KLabel, Set[Rule]] = rules.groupBy(r => matchKLabel(r))
-  lazy val macroKLabels: Set[KLabel] = rules.filter(r => r.isMacro).map(r => matchKLabel(r))
+  lazy val macroKLabels: Set[KLabel] = macroKLabelsFromRules++macroKLabelsFromProductions
+  lazy val macroKLabelsFromRules: Set[KLabel] = rules.filter(r => r.isMacro).map(r => matchKLabel(r))
+  lazy val macroKLabelsFromProductions: Set[KLabel] = productions.filter(p => p.isMacro).map(p => matchKLabel(p))
 
-  private def matchKLabel(r: Rule) = r.body match {
+  def matchKLabel(r: Rule) = r.body match {
```
radumereuta: IntelliJ: Type annotation required for public member.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -234,16 +234,28 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
...
     case Unapply.KRewrite(Unapply.KApply(s, _), _) => s
     case _ => KORE.KLabel("")
   }
 
+  private def matchKLabel(p: Production) = p.klabel match {
+    case Some(klabel) => klabel
+    case _ => KORE.KLabel("")
+  }
+
+  def ruleLhsHasMacroKLabel(r: Rule) = r.body match {
```
radumereuta: IntelliJ: Type annotation required for public member.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaBackend.java
```diff
@@ -123,6 +123,7 @@ public JavaBackend(KExceptionManager kem, FileUtil files, GlobalOptions globalOp
                 //.andThen(ModuleTransformer.fromSentenceTransformer(new NormalizeAssoc(KORE.c()), "normalize assoc"))
                 .andThen(AddBottomSortForListsWithIdenticalLabels.singleton())
                 .andThen(ModuleTransformer.from(new GenerateSortProjections(false)::gen, "adding sort projections"))
+                .andThen(m2 -> ModuleTransformer.fromSentenceTransformer((mod, s) -> new PropagateMacro(mod).propagate(s), "propagate macro labels from production to rules").apply(m2))
```
radumereuta: Line too long.

radumereuta (changes requested):
Just a few small issues.
I will go through this again tomorrow. Too tired now for deep logic.
Also, @dwightguth is the most familiar with macros, so he should be the one to do the final approval.

kore/src/main/scala/org/kframework/attributes/Att.scala
SchmErik: it is but it doesn't make sense to have multiple labels. I think this should be a separate check all together

radumereuta:
I managed to have another look. Plenty of tests and the code lgtm, other than those small surface comments.

radumereuta (approved):
@dwightguth can you give the final approval?

dwightguth (approved):

PR #2279, opened 2021-10-19
remove unneeded --depth 0 option

SchmErik:
There is a regression introduced in the haskell backend from this
commit: https://github.com/kframework/kore/commit/ed422630f9383b2774f1ec8ea04eee86ea0afe1e

The regression causes some calls to kprove to hang. In order to avoid
the hang, the `--depth 0` option was added to this test. This change
removes this option from the Makefile so that the test can execute
without this band aid. This change should be merged after the regression
has been resolved.

### DIFF
--- a/k-distribution/tests/regression-new/kprove-append/Makefile
+++ b/k-distribution/tests/regression-new/kprove-append/Makefile
@@ -2,6 +2,6 @@ DEF=def
 EXT=def
 KOMPILE_BACKEND=haskell
 TESTDIR=.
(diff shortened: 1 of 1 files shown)


SchmErik:
@ehildenb, It looks like this is resulting in an error. Long story short, I fixed a bug which resulted in this test case crashing but now it loops in a stuck state. Is this test case expected to be in a stuck state? See the thread below for more information...

https://github.com/kframework/k/issues/2245#issuecomment-974517618

SchmErik:
please refer to https://github.com/runtimeverification/k/pull/2422 for an updated pull request.

PR #2280, opened 2021-10-19
Kompile: add check for existential variable in LHS

### LINKED ISSUE #2264: [Bug] [kprove|kprovex] - Expecting failure when claim contains existential in LHS
K Version
---------

```
$ kompile --version
K version:    v5.1.226-0-g70dd86c5c
Build date:   Tue Oct 12 15:34:16 EEST 2021
```

Description
-----------

Our encoding of claims is the following:
```
∀ X . φ(X) → mod ∃ Y . ψ(X, Y)
```
where `X` and `Y` are sets of variables and `mod` is a reachability modality.

Therefore, it is incorrect to have existentially quantified variables on the LHS of the claim. This should be caught by the frontend and explained to the user. Right now, the frontend doesn't complain at all.

Input Files
-----------

File test.k
```
module TEST
    imports INT

endmodule
```

File test-spec.k
```
module TEST-SPEC
    imports TEST

    claim <k> ?X:Int => . </k>

endmodule
```

Reproduction Steps
------------------

```
```
The behavior is the same for both `kprove` and `kprovex`:
```
$ kprove test-spec.k
```
```
$ kprovex test-spec.k
```
The frontend succeeds resulting in the backend being run (the result of the backend is not important).

Expected Behavior
-----------------

I expect the frontend to throw an error signalling that `X` is existentially quantified, and that the LHS is not allowed to contain existentially quantified variables.

SchmErik:
Existential variables are not allowed in the LHS of rules/claims. They
are only allowed on the RHS.

Fixes: #2264

### DIFF
--- a/k-distribution/tests/regression-new/checks/existentialCheck.k
+++ b/k-distribution/tests/regression-new/checks/existentialCheck.k
@@ -1,0 +1,5 @@
+module EXISTENTIALCHECK
+
+    claim <k> ?X:Int => . </k>
+
+endmodule
--- a/k-distribution/tests/regression-new/checks/existentialCheck.k
+++ b/k-distribution/tests/regression-new/checks/existentialCheck.k
@@ -2,4 +2,6 @@ module EXISTENTIALCHECK
 
     claim <k> ?X:Int => . </k>
 
+    rule <k> ?X:Int => . </k>
+
 endmodule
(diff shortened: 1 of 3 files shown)


radumereuta:
Can you add a rule that produces the same error?

SchmErik:
> Can you add a rule that produces the same error?

Done

radumereuta (approved):

PR #2293, opened 2021-11-01
fix some minor issues with the k tutorial

dwightguth:
This fixes a number of the issues in #2033. I didn't fix all of them; just the ones that could easily be fixed with a few lines or fewer of modification. I will do another pass at some point in the future.


hjorthjort (approved):
LGTM

PR #2297, opened 2021-11-03
Kompile: improve macro propagation

### LINKED ISSUE #1584: Allow putting `macro` on productions

SchmErik:
This pull request contains improvements to the functionality that allows macro labels to be written on productions. One change is an adjustment to the macro propagation logic. Another change the addition of a kompile error that is emitted if rules contain macros that remain unresolved after macro expansion

fixes: #1584


k-distribution/tests/regression-new/macro_vars-productions/macro-vars.k
```diff
@@ -12,4 +12,7 @@ module MACRO-VARS
   syntax CPPType ::= underlyingType(CPPType) [function]
   syntax CPPType ::= "cppEnumType" [macro]
   rule cppEnumType => bar(_, scopedEnum() #Or unscopedEnum() ) [unboundVariables(_)]
+  syntax Int ::= A ( Int ) [macro]
+  rule A ( N ) => N +Int 1
+  rule A ( N ) => N [simplification]
```
ehildenb, line 17: This is a bit weird of a definition, you're basically saying that `N ==Int N +Int 1`. Not sure it matters because this is a test, but it looks weird to me.

radumereuta (changes requested):
lgtm except for the failing tests and Everett's comment.

k-distribution/tests/regression-new/macro_vars-productions/macro-vars.k
SchmErik: I agree. it's a really weird definition but this is also one of the simplest cases. There's also an error that occurs if I change N +Int 1 to N but I'll file a separate bug for that. The idea with this test is to exercise the macro propagation logic. The new logic is supposed to not propagate to simplification rules

radumereuta (approved):
Love the comments.
It explains the intent of the code, not the actual implementation.

PR #2314, opened 2021-11-09
Make `krun` compatible with bash >= 3.0

### LINKED ISSUE #2313: Config variables not parsed properly in `krun`
Some tests fail because the wrapper scripts don't pass config variables through properly.

Baltoli:
macOS ships an ancient version of bash by default, which breaks part of the `krun` config variable script [when handling empty arrays](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u).

This PR fixes the places in `krun` where the faulty `set -u` behaviour is relied on, and replaces them with a portable expansion.

Closes #2313

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,15 +80,13 @@ If you install this list of dependencies, continue directly to the [Build and In
 On macOS using [Homebrew](https://brew.sh/):
 ```shell
 git submodule update --init --recursive
-brew install bash bison boost cmake flex gcc gmp openjdk jemalloc libyaml llvm make maven mpfr pkg-config python stack zlib z3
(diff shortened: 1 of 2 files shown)


dwightguth (changes requested):
I'd really rather we fix the scripts so that either they use zsh on mac os, or else that they work on bash 3.

Baltoli:
Sure, I can rework the script to work with bash < 4.4 following that SO page I linked - it makes some of the code look a bit more arcane but otherwise isn't a huge change. I wasn't sure what the prioritisation of old bash support would be, so opted for the documentation change as a lower-impact fix.

dwightguth (approved):

PR #2322, opened 2021-11-10
Use GitHub Actions to test K on MacOS instead of Anka

dwightguth:
While Anka has served us very well as a way of having access to different versions of MacOS for testing purposes, it has been rather unreliable for us when used for CI. We don't use GitHub Actions for CI more broadly because we want more parallelism in our machines for the most part. However, we were not using a particularly beefy machine to test on MacOS previously, so using GitHub actions would not be any slower than Anka, overall. Since this is a public repository, we also don't have to worry about paying for a certain number of minutes; GitHub actions does have intrinsic limits, but we are likely well below those and below them, the service is free.

As such, it seemed appropriate to remove our dependence on Anka and make it so that our CI tests K on MacOS using GitHub Actions. This was possible via the following changes:

1. The Deploy phase of the Jenkinsfile now creates a prerelease rather than a full release. I would have preferred to use a draft release, but GitHub actions does not allow you to trigger workflows when a draft release is created for some inexplicable reason.
2. When a prerelease is created, we kick off a GitHub workflow which builds the homebrew bottle on MacOS and then tests it locally on a clean environment. If the tests pass, it attaches the bottle to the release, pushes the changes to the homebrew formula to the homebrew-k repo, and marks the release as no longer being a prerelease.

I have not actually tested the "push changes to homebrew-k" or "mark release as final" phases of the workflow yet, as it is impossible to do so without potentially disrupting the user experience. Particular care should be spent reviewing these sections, and when this is merged, I will monitor closely and push additional fixes if necessary in order to address any issues that arise.

Here is evidence that the rest of the workflow passed: https://github.com/kframework/k/actions/runs/1440945977


CODEOWNERS
```diff
@@ -1,3 +1,4 @@
 /k-distribution/include/kframework/builtin/* @dwightguth
 /USER_MANUAL.md @dwightguth
 /Jenkinsfile @kframework/admin
+/.github/workflows/* @kframework/admin
```
radumereuta, line 4: Who has access to admin?

radumereuta (commented):
Looks good as far as I can tell.
I don't have that much experience with setting up CI, so I leave the final approval to Everett.

SchmErik:
I think this looks good but I don't have a lot of experience with CI and home brew either...

Baltoli (approved):
Looks like a sensible translation of the Jenkinsfile to me; these things are tricky to review but it seems to be working in your testing so I'm happy.

CODEOWNERS
ehildenb: Dwight, me, Grigore, Patrick.

Jenkinsfile
```diff
@@ -519,26 +419,21 @@ pipeline {
...
             mv ../focal/kframework_${VERSION}_amd64.deb                 kframework_${VERSION}_amd64_focal.deb
             mv ../bullseye/kframework_${VERSION}_amd64.deb              kframework_${VERSION}_amd64_bullseye.deb
             mv ../arch/kframework-git-${VERSION}-1-x86_64.pkg.tar.zst   kframework-git-${VERSION}-1-x86_64.pkg.tar.zst
-            mv $LOCAL_BOTTLE_NAME                                       $BOTTLE_NAME
 
             echo "K Framework Release ${VERSION}"  > release.md
             echo ''                               >> release.md
             cat k-distribution/INSTALL.md         >> release.md
-            hub release create                                                                  \
+            hub release create --prerelease                                                     \
```
ehildenb, line 431: I want to make sure I understand how this works. So this release will be cut, and then the Mac package will be triggered (because of the `prereleased` tag) _after_ this release is cut, and it will turn it into a proper release?

ehildenb (approved):
If my question is answered, feel free to add automerge.

Jenkinsfile
ehildenb: So there is a chance that mac release fails, which leads to the release staying as a pre-release?

dwightguth: if the mac release fails, the github workflow will delete the release.

PR #2333, opened 2021-11-15
Fix2321 - refactor collection steps into the constructor of ModuleToKore 

### LINKED ISSUE #2321: [Bug] [kprovex] - Parse error in kore generated by kprovex
K Version
---------

```
$ kompile --version
K version:    v5.1.252-0-gea4fa02ed
```

Description
-----------
I'm getting the following kore parse error when I run a command with `kprovex` instead of `kprove`:

```
kore-exec: [127708] Error (ErrorParse):
    /Users/emarzion/koreissue/./.kprove-2021-11-10-11-40-18-595-7b2fdc5e-e231-441d-92eb-b8e3ceb7c569/spec.kore:13:67:
       |
    13 |         \and{SortBool{}}(\dv{SortBool{}}("true"),\forall{SortInt{}, SortBool{}}(VarX:SortInt{},\equals{SortInt{}, SortBool{}}(Lbli'LParUndsCommUndsRParUnds'TEST'Unds'Int'Unds'Int'Unds'Int{}(VarX:SortInt{},\dv{SortInt{}}("0")),\dv{SortInt{}}("0")))),
       |                                                                   ^
    unexpected ','
    expecting '}'

[Error] Critical: Haskell Backend execution failed with code 1 and produced no
output.
```

Input Files
-----------

`test.k`:

```
module TEST
    imports BOOL
    imports INT

    syntax Exp ::= Int | Bool
                 | pair(Int, Int) [klabel(pair), symbol]

    syntax KItem ::= "#assume" Exp [klabel(assume), symbol]
    rule #assume true => .K
    rule #assume false => #Bottom

    syntax KItem ::= "#assert" Exp  [klabel(assert), symbol]
                   | "#fail"
    rule #assert true => .K
    rule #assert false => #fail

// Uninterpreted functions
    syntax Int ::= i(Int, Int)   [function, functional, no-evaluators, smtlib(fi), klabel(i)]
endmodule 
```

`issue-2138-spec.k`:

```
requires "test.k"

module VERIFICATION
  import TEST
endmodule

module ISSUE-2138-SPEC
  import VERIFICATION

  claim
    <k> #assert i ( 1 , 0 ) ==Int 0 => . </k>
      requires true #And ( #Forall X. { i ( X, 0 ) #Equals 0 } )

endmodule
```

Reproduction Steps
------------------

`kompile --backend haskell -d . --main-module VERIFICATION issue-2138-spec.k`
`kprovex --haskell-backend-command kore-exec issue-2138-spec.k`

Expected Behavior
-----------------

I expect `kprovex` to successfully return `#Top`, which works when you replace `kprovex` with `kprove` in the command above.

radumereuta:
Fixes #2321
When I built kprovex, I eliminated the step which reprinted definition.kore.
In that step, there is a collection pass that gathers `ModuleToKORE.mlBinders`. I moved it to the constructor.
There is another step, `ModuleToKORE.impureFunctions` but if I move this to the constructor it makes kast crash [here](https://github.com/kframework/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1628). It tries to find the label for `<generatedTop>` inside the syntax module. @dwightguth how are `impureFunctions` used here? What is their purpose?


dwightguth:
I don't think impureFunctions is actually being used for anything meaningful. You might be able to remove it. You should check its usages.

radumereuta:
> I don't think impureFunctions is actually being used for anything meaningful. You might be able to remove it. You should check its usages.

```
if (impureFunctions.contains(k.klabel().name())) {
    throw KEMException.internalError("Cannot yet translate impure function to kore: " + k.klabel().name(), k);
}
```
Only this error message.

dwightguth:
Yeah, it's fine to remove that. It was originally intended to report some cases where the encoding to kore wasn't implemented yet, but it's pretty clear that check isn't working as intended currently, so there's no particular need to keep it.

radumereuta:
I think it might work correctly for kompile. What I was trying to do, was to call that for kprovex, and it ended up being called for kast also, which failed. And for good reasons, it wasn't designed to work there.

I'm trying to say that the current code is correct, so it's ready for review.

radumereuta:
Just remove that code path. Not being used.

radumereuta:
@dwightguth reminder, please re-review.

dwightguth (approved):

PR #2336, opened 2021-11-18
Check for existential variables in requires clause

### LINKED ISSUE #2324: [Bug] [kprove|kprovex] Existential variable is allowed on LHS, but shouldn't
K Version
---------

```
$ kompile --version
K version:    v5.1.250-0-g07918e142
Build date:   Sat Nov 06 22:18:15 CET 2021
```

Description
-----------

The frontend should produce an error if an existential appears on the LHS, see <https://github.com/kframework/k/issues/2264>. However, this is not always the case.

Consider the following definition:
```
module TEST
  imports K-EQUAL
  imports BOOL

  syntax Program ::= "stateA" | "stateB"
  syntax Bool ::= isDone(Program) [function]

  rule stateA => stateB

  rule isDone(stateB) => true
  rule isDone(_) => false [owise]
endmodule
```

Then the following claim is successfully proven in both `kprove` and `kprovex`:
```
module TEST-SPEC
  imports TEST

  claim <k> stateA => ?STATE </k>
    requires isDone(?STATE)
endmodule

```
However, since existentials are not allowed on the LHS, this should produce an error.

Output of `kprovex`:
```
kore-exec: [357438] Warning (WarnTrivialClaim):
    Claim proven without rewriting at: /home/david/Documents/RV/k-snippets/prove-nonexistential/test-spec.k:13:9-14:28
#Top
```

SchmErik:
The `requires` clause is a part of the LHS of the rule. Since existential
variables are not allowed on LHS of the rule, these variables are not
allowed in the clause. Add a check to catch rules like this.

Fixes: https://github.com/kframework/k/issues/2324

### DIFF
--- a/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/simple-spec.k
+++ b/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/simple-spec.k
@@ -43,7 +43,7 @@ claim
 </lambdaT>
 requires N1 ==Int N2
  andBool M1 ==Int M2
- andBool ?_Ks1 ==K ?_Ks2
+ andBool Ks1 ==K Ks2
 
 claim
 <k> check => . </k>
--- a/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/simple-spec.k
+++ b/k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/simple-spec.k
@@ -43,7 +43,7 @@ claim
 </lambdaT>
 requires N1 ==Int N2
  andBool M1 ==Int M2
- andBool Ks1 ==K Ks2
+ andBool _Ks1 ==K _Ks2
 
 claim
 <k> check => . </k>
--- a/k-distribution/tests/regression-new/spec-rule-application/def02-spec.k
+++ b/k-distribution/tests/regression-new/spec-rule-application/def02-spec.k
@@ -10,7 +10,7 @@ module DEF02-SPEC
 
     claim <k> mid Y => end ?Z </k>
          <var> _ </var>
-      requires ?Z ==Int incPos(Y)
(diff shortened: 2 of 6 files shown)


SchmErik:
converting to draft. Will set this as ready to review after a few tests have been fixed: https://github.com/kframework/k/issues/2324

k-distribution/tests/regression-new/spec-rule-application/def02-spec.k.out
```diff
@@ -1 +1,8 @@
-#Top
```
SchmErik: @radumereuta Changing the def02-spec.k resulted in a different test output. Is this the correct approach??

k-distribution/tests/regression-new/bisimulation/examples/program-equivalence/imp-lambda/simple-spec.k
```diff
@@ -43,7 +43,7 @@ claim
 </lambdaT>
 requires N1 ==Int N2
  andBool M1 ==Int M2
- andBool ?_Ks1 ==K ?_Ks2
+ andBool Ks1 ==K Ks2
```
radumereuta: `andBool _Ks1 ==K _Ks2`
Don't forget about underscore, otherwise you will get warnings.

k-distribution/tests/regression-new/spec-rule-application/def02-spec.k
```diff
@@ -10,7 +10,7 @@ module DEF02-SPEC
 
     claim <k> mid Y => end ?Z </k>
          <var> _ </var>
-      requires ?Z ==Int incPos(Y)
+      requires Z ==Int incPos(Y)
```
radumereuta, line 13: `?Z` appears in two locations. You need to update both. Otherwise, you will get two different variable names.

radumereuta (changes requested):

radumereuta (approved):

PR #2338, opened 2021-11-20
pyk frontend tools

### LINKED ISSUE #2268: pyk KDefinition should be an object
Currently, pyk is organized as a collection of functions, but does not take advantage of Python's object-oriented nature.

The most obvious candidate for where to start this is to make `KDefinition` an object, instead of a JSON data-structure. A `KDefinition` should contain fields:

- Path to `*-kompiled` directory (passed in as argument to constructor).
- JSON definition (read from `*-kompiled/compiled.json`).
- symbolTable (by called `buildSymbolTable` on the JSON definition: https://github.com/kframework/k/blob/master/k-distribution/src/main/scripts/lib/pyk/kast.py#L268).

and public methods:

- `krun`: Implemented like this, but where `definition` no longer needs to be passed in: https://github.com/kframework/k/blob/master/k-distribution/src/main/scripts/lib/pyk/__init__.py#L51
- `kprove`: Implemented like this, but where `definition` no longer needs to be passed in: https://github.com/kframework/k/blob/master/k-distribution/src/main/scripts/lib/pyk/__init__.py#L71
- `printKast`: Implemented as wrapper around `prettyPrintKast` method which already exists, but no longer needs `symbolTable`. Prints a given term that is psased in: https://github.com/kframework/k/blob/master/k-distribution/src/main/scripts/lib/pyk/kast.py#L302
- `print`: Prints the JSON definition using `prettyPrintKast`.

ehildenb:
Fixes: https://github.com/kframework/k/issues/2268

This PR does many things:

- Many formatting changes (" => ').
- Organization (move several utilities to `util.py`)
- Adds a new file `ktool.py` which contains new classes for running K definitions:
  - `KPrint` takes in a definition path, and produces an unparser for it.
  - `KProve > KPrint` takes in a definition path and main file name, and produces a prover instance for it which allows you to `prove` claim files, or `proveClaim` individual claims.
- Removes `__init__.py` contents, which are not relevant anymore with new class-based interface/organization.
- Makes the `kpyk` runner tool just a thin wrapper around `KProve` and `KPrint`, and makes sure that both are tested on CI.
- Removes the dedicated `graphviz.py` file and inlines it into `__main__.py`, and adds a test for `graph-imports`.
- Improves the pyk unparsing of requires/ensures clauses.


k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -216,11 +205,22 @@ def addAttributes(kast, att):
...
         sys.stderr.write(str(kast))
         sys.stderr.flush()
         sys.exit(1)
 
+def flattenLabel(label, kast):
+    """Given a tree of a label, return all the leaves.
+
+    -   Input: label of tree, and kast term.
+    -   Output: list of leaves of tree (singleton list for no occurance of label at top).
+    """
```
radumereuta, line 205: I don't get this explanation.
What is the tree of a label? A label is a single string. A tree is a structure.

k-distribution/src/main/scripts/lib/pyk/ktool.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        return prettyPrintKast(kast, self.symbolTable)
+
+class KProve(KPrint):
+    """Given a kompiled directory and a main file name, build a prover for it.
+    """
+
+    def __init__(self, kompiledDirectory, mainFileName, useDirectory = None):
+        super(KProve, self).__init__(kompiledDirectory)
+        self.directory      = '/'.join(self.kompiledDirectory.split('/')[0:-1])
```
radumereuta: You should use a library to get the parent directory here. What happens when you have `/my/path/test-kompiled/.`
Also, `directory` -> `parentDir`? I find it more descriptive.

k-distribution/src/main/scripts/lib/pyk/ktool.py
```diff
@@ -0,0 +1,102 @@
...
+        self.useDirectory   = self.directory + '/kprove' if useDirectory is None else useDirectory
+        if not os.path.exists(self.useDirectory):
+            os.makedirs(self.useDirectory)
+        self.mainFileName   = mainFileName
+        self.prover         = [ 'kprovex' ]
+        self.proverArgs     = [ ]
+        with open(self.kompiledDirectory + '/backend.txt', 'r') as ba:
+            self.backend    = ba.read()
+        with open(self.kompiledDirectory + '/mainModule.txt', 'r') as mm:
+            self.mainModule = mm.read()
```
radumereuta, line 42: Do we have tests for these?
In case we change the contents of those files.

radumereuta (changes requested):
Looks reasonable.
But I will go again over the entire thing once you address those minor concerns.

k-distribution/src/main/scripts/lib/pyk/kast.py
ehildenb: I updated this documentation to be clearer, instead referring to them as cons-lists.

k-distribution/src/main/scripts/lib/pyk/ktool.py
ehildenb: I modified the testing harness to use `kpyk prove ...` (https://github.com/kframework/k/pull/2338/commits/3a534ab0dc9da0c269dfa493835b7bbd3871b417).

`kpyk prove ...` is really not intended to be used directly, it's just a wrapper that's there so we can have tests like this. But because `kpyk prove` builds a `KProve` object, that test will fail if those files aren't present.

k-distribution/src/main/scripts/lib/pyk/ktool.py
ehildenb: I'm now using the `Path` object from `pathlib` for this. It does make it nicer.

Jenkinsfile
```diff
@@ -63,7 +63,7 @@ pipeline {
             }
           }
           stages {
-            stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git', credentialsId: 'rv-jenkins-access-token' } } }
+            stage('Checkout code') { steps { dir('k-exercises') { git url: 'git@github.com:kframework/k-exercises.git' } } }
```
ehildenb: This is just undoing a previous thing that did _not_ correctly solve a credentials issue. @dwightguth you'll have to review this one.

ehildenb:
Sorry @radumereuta  that this got so big. I tried to make sure that large parts of new functionality are tested in the integration tests (just by calling `kpyk prove | kpyk print`), and it's working with my downstream summarizer.

radumereuta:
The changes to my original comments look good.
Let me know when tests pass so I can have another look at the entire thing.

k-distribution/src/main/scripts/lib/pyk/__main__.py
```diff
@@ -81,33 +93,12 @@ def main(commandLineArgs, extraMain = None):
...
     elif extraMain is not None:
         extraMain(args, kompiled_dir)
 
-    args['output'].flush()
-
     if returncode != 0:
         _fatal('Non-zero exit code (' + str(returncode) + '): ' + str(kCommand), code = returncode)
 
 if __name__ == '__main__':
+    sys.setrecursionlimit(15000000)
```
radumereuta: How did you get this number?

k-distribution/tests/pyk/imp-import-graph.expected
```diff
@@ -0,0 +1,178 @@
...
+STRINGCOMMONINT
+STRINGCOMMONKEQUAL
+STRINGCOMMONSTRINGSYNTAX
+STRINGKORE
+STRINGKOREKEQUAL
+STRINGKORESTRINGCOMMON
+STRINGSTRINGCOMMON
+STRINGSTRINGKORE
+STRINGSYNTAX
+UNSIGNEDINTSYNTAX
```
radumereuta, line 178: If we add a new module to domains.md, is this going to fail?
Does `make update-results` work? I see it has no target in the Makefile.

radumereuta (approved):
Just the two questions.
lgtm otherwise, but it's a big diff so most likely I missed stuff.
Maybe wait for a second reviewer?

k-distribution/src/main/scripts/lib/pyk/__main__.py
ehildenb: It's a really big number that is smaller than the max limit on my machine. It's worked OK so far.

k-distribution/tests/pyk/imp-import-graph.expected
ehildenb: Yes, it will fail. `make CHECK=cp` will update the output.

I think we should switch `make update-results` to the `make CHECK=cp` system we use in the various semantics, it's cleaner to me. But that is a big project.

k-distribution/src/main/scripts/lib/pyk/__main__.py
radumereuta: Worth adding a comment then.

k-distribution/tests/pyk/imp-import-graph.expected
radumereuta: What do you mean by `make CHECK=cp` is cleaner?
I find `make update-results` way more intuitive from a naming convention. It's a make target as well.
If you want us to change it, then you should make an issue. Until then I think it's better if we stick to the same standard we use everywhere in our tests.

I removed automerge until we clarify this.

k-distribution/src/main/scripts/lib/pyk/__main__.py
ehildenb: Added.

k-distribution/tests/pyk/imp-import-graph.expected
ehildenb: NVM, I added `make update-results`.

`CHECK=cp` makes more sense to me, but either way `update-results` can just use `CHECK=cp` under the hood.

PR #2340, opened 2021-11-22
Add options to modify llvm-kompile's parameters in the Kompile frontend

### LINKED ISSUE #2088: Allow calling `llvm-kompile` in `library` mode through `kompile` frontend
Firefly currently uses a separate invocation to `kompile --no-llvm-kompile` and `llvm-kompile`, because it needs to call `llvm-kompile` in `library` mode.

I suggest making `library` and `main` a boolean toggle, and `search` and `execution` a boolean toggle.

Also, if the `-o SOME_FILE` directory doesn't exist, `llvm-kompile` does not automatically create it. Can it?

gtrepta:
fixes: https://github.com/kframework/k/issues/2088

Usage: `kompile --backend llvm ... --llvm-kompile-type <main|search|library|static> --llvm-kompile-output <filename> ...`

This gives better access to llvm-kompile's parameters instead of going through `--no-llvm-kompile`.


dwightguth (dismissed):
This needs a bit of work.

dwightguth (changes requested):
This needs a bit of work:

1. First, we should call the flag something like `llvm-kompile-output-type` and it should take an enumeration with the following values: main, search, static, library.
2. You should call the second flag `llvm-kompile-output` and it should probably not have a short option version. It should override the default output path in all cases and it should be an error to use together with `--enable-search`.
3. We need better documentation of exactly what these flags do and what they are used for. That includes describing the intended use case of each of the 4 enumeration values I named above, and more detailed descriptions of both flags that don't assume knowledge of the internals of the backend.

In case you're unaware, here's a bit about each of the 4 values I mentioned above:

1. main. Builds an interpreter and links in the main method. Generates an executable unless other flags are explicitly passed to llvm-kompile
2. search. Same as main, except the executable does search instead of single-path execution.
3. library. Same as main, except that it does not link in a main method. The user is expected to use it either by explicitly linking in their own main method.
4. static. Same as library, except it does not pass any `-l` flags to clang when doing linking. This is useful primarily if you are passing `-r -nostdlib` to llvm-kompile in order to perform partial linking and generate an object file that can be linked later on a machine that does not have K installed.

gtrepta:
@dwightguth
- What should the behavior be with `--llvm-kompile-output-type search`? Should that set the `--enable-search` flag?
- That would also mean `--llvm-kompile-output-type search --llvm-kompile-output xxxx` should throw an error?
- Are you expecting the documentation to be in the `--help` output, or somewhere else like the user manual?

dwightguth:
1. No, it shouldn't set --enable-search, it should just pass `search` as the type instead of `main` or `library`.
2. No, it should be fine for the user to pass whatever output path they want when they select search.
3. A short description should at the very least be in the --help message. We can put more elaborate details in the user manual if the description in the help message isn't complete. I just would prefer the help message be something that focuses on the use case rather than a simple set of factual technical details that are opaque to users not familiar with them.

gtrepta:
@dwightguth re-requested your review.

I want to mention that `library` and `static` will cause clang to error out if the appropriate additional arguments are not passed in. This isn't mentioned so far in the `--help` text since it's already kind of long and hard to read. But I'm wondering if this behavior is fine with you?

llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
```diff
@@ -36,4 +36,10 @@
 
     @Parameter(names="--enable-search", description="By default, to reduce compilation time, `krun --search` is disabled on the LLVM backend. Pass this flag to enable it.")
     public boolean enableSearch;
+
+    @Parameter(names="--llvm-kompile-type", description="Specifies the llvm backend's output type. Valid choices are main (build an interpreter), search (same as main, but the interpreter does search instead of single-path execution), library (same as main, but a main method isn't linked in. The user is expected to link their own), and static (same as library, but no '-l' flags are passed during linking. Used for making a portable object file).")
```
dwightguth: s/portable object file/partially linked object file/

llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
```diff
@@ -36,4 +36,10 @@
...
+    public String llvmKompileType = "main";
+
+    @Parameter(names="--llvm-kompile-output", description="Name the output binary from the llvm backend.")
```
dwightguth: s/Name/Name of/

dwightguth (approved):
looks good but please update the help text as suggested before merging.

PR #2345, opened 2021-11-25
Add simplification rules to spec.kore

### LINKED ISSUE #2287: Lift kprovex restriction to allow simplification rules in proofs
@ana-pantilie if we put simplification rules in spec.kore, does the haskell backend support that?
If not, how hard it would be? kprovex managed to avoid some bugs we had in kprove, but with added limitations which can be annoying.
definition.kore is expensive to generate for each proof so we want to avoid doing that.

radumereuta:
Fixes: #2287

Requires support from the haskell backend
@ana-pantilie I've made changes to the front end to include simplification rules in the spec.kore file.
I've also added a test that should pass when the backend handles the simplification rules correctly.


radumereuta:
@ehildenb this is ready for review. We have the front end changes and the backend. I added a simple test, but we could use more complex ones. If you have some from KEVM that would increase our confidence in these changes.
I'm expecting some potential issues since the compilation pipeline is different than kprove, but I can't anticipate more than that. I need examples.

k-distribution/tests/regression-new/issue-2287-simpl-rules-in-kprovex/a1-spec.k
```diff
@@ -0,0 +1,20 @@
...
+module VERIFICATION
+  import TEST
+
+  rule (X +Int Y) +Int Z => Z +Int (X +Int Y) [simplification, symbolic(X), concrete(Y, Z)]
+  rule (X +Int Y) +Int Z => Y +Int (X +Int Z) [simplification, concrete(X), symbolic(Y), concrete(Z)]
+  rule X +Int (Y +Int Z) => Z +Int (X +Int Y) [simplification, concrete(X, Y), symbolic(Z)]
+  rule X +Int (Y +Int Z) => Y +Int (X +Int Z) [simplification, concrete(X), symbolic(Y), concrete(Z)]
+  rule final(n +Int _M:Int) => 0 [simplification, concrete(_M)]
+endmodule
+
```
ehildenb: We should have associated error checks too, specifically for:

- non-`simplification` rules being added should be an error? (not sure if the haskell backend allows normal axiomatic rules here, but I think we should restrict this behavior anyway, since introducing new semantic rules changes the transition system, which is usually probably not what you want)
- Adding `syntax` ... declarations should be an error.
- Adding a `rule <k> ... </k> ` semantic rule (with a configuration) should be an error.

ehildenb: We should also have some `simplification` rules with side-conditions in this test. Maybe just have your last rule on line `11` have a side condition?

ehildenb:
We should also add a test for a non-functional `simplification` rule, like:

```
rule <k> foo => bar ... </k> [simplification]
```

added into the proof module. The current check I think will allow it, but I think we should dis-allow it for now.

ehildenb:
Specifically, we should just restrict to _only_ function/functional simplification rules, because that's all we need right now, and we don't want to think about whether it's safe to add non-functional simplification rules.

radumereuta:
> Specifically, we should just restrict to _only_ function/functional simplification rules,

I don't really get where this is coming from.

> because that's all we need right now, 

that's not really an argument

> and we don't want to think about whether it's safe to add non-functional simplification rules.

this is an argument, but let's do think about this now since we already have this issue in memory.
Had a quick look, and implementing this restriction might take more time than figuring out if we can allow simplification rules that are more complex.
@ana-pantilie do we need to enforce, in the front-end, some kind of restriction to simplification rules found in spec files?

ana-pantilie:
I think it should be safe to add any kind of simplification rule which would be correct to have inside the definition.

radumereuta:
Add the restrictions for now. When we figure out what it means to add more things, we can relax the checks.

radumereuta:
@ehildenb this is ready for re-review.
Pay special attention to the way I restricted the simplification rules to only allow function rules.
Also, I was thinking of adding an option to disable this check, in case it's too restrictive and you want to try something more complicated.

ehildenb (approved):

PR #2347, opened 2021-11-30
Fix missing package in dockerfile

dwightguth:
This appears to fix the broken k release which occurred as a result of the `python3` package not installing the `python` binary on Ubuntu Focal.


Baltoli (approved):
Looks good to me

PR #2349, opened 2021-11-30
add filename to yyerror

dwightguth:
Previously we were not adding the filename to the error message from yyerror. This is easy enough to add, so we do so.


Baltoli (approved):
Looks good.

radumereuta (approved):

PR #2352, opened 2021-11-30
make the "inner cast" be a synonym for a strict cast

dwightguth:
This would close the "type loophole" that lets you put a term of any sort anywhere. Since when you did this, a bunch of stuff broke, we don't really have more than the single instance in any of our semantics of this actually being used. Hence the reason this PR is so small.

The singular usage is on line 1649 of domains.k.


radumereuta:
Why not remove it completely?
If it's because we use it, can we give a warning for a transition period, and then remove it?

dwightguth:
As I have said elsewhere, it is not currently possible to strict-cast a term whose last production item is a non terminal. This feature is needed in order to disambiguate the particular rule that currently is using an inner cast as a workaround. Simply removing it entirely does not work because then that rule parses incorrectly.

radumereuta:
You mean this `rule {SB:String +String S:String}<:StringBuffer => (SB +String S)::String`
So strict cast doesn't go through parentheses? Is the RHS being checked?
Maybe we should change that too then.
Something feels weird here.

dwightguth:
Logically, a strict cast doesn't pass through a bracket because strict casts only affect the term they are applied to, which in this case would be a bracket. The inside of the bracket would still be non strict, and thus the bracket essentially negates the strictness of the cast. I guess maybe you could make an argument that since this makes strict casting a bracket completely useless, that we should instead change the behavior of the type inferencer so that a term inside a bracket inside a strict cast is also strict. That requires a much more involved change to the type inferencer, but it's doable and then you would be able to write the rule in question using a strict cast.

Is that what you would prefer to see instead? If we did that, we ought to be able to entirely delete the inner cast syntax. It might break some existing definitions in theory, because of the modified behavior of strict casts, though. This change is much more backwards compatible.

I'm willing to be flexible. Make your arguments for or against each idea.

dwightguth:
By the way, one argument against making strictness pass through casts is that someone could theoretically write a grammar where the bracket itself needed to be strict casted in order to disambiguate it, which would no longer be possible without unintended side effects if we made such a change to the inferencer. Now such a grammar is highly unlikely in practice, but it is a limitation on the expressivity of k in terms of the set of grammars it is able to faithfully represent, so if you're concerned about theoretical completeness in the design of the parser, it's probably worth keeping the inferencer code the way it is in this PR instead.

radumereuta (approved):
Now that you mention it, I don't have a better suggestion.
Let's keep it this way. If we run into issues, we can address them then.
Feel free to add automerge if you don't want to wait for Everett.

PR #2355, opened 2021-12-01
Fix a few minor bugs in bison table generation

dwightguth:
This should probably be reviewed one commit at a time, but making 3 separate PRs seemed overkill to me when all three fixes are so simple and pretty related. 

This fixes 3 issues:

1. A bug when you ran kparse-gen that reported an uninitialized variable being used
2. A bug where the resulting KORE would be incorrect if you requested location annotations on a sort with epsilon productions
3. A bug involving incorrect AST generation when using location annotations and cons lists.


ehildenb:
Is it easy to add a test of these changes? If so, please do so. Otherwise, LGTM.

radumereuta (commented):
Not that I'm complaining, but where are you getting these bugs from?
I haven't seen them on any issues.
If you can add the test that Everett mentioned, LGTM

dwightguth:
I had some spare time and the k project board was too messy to pick up a task from (we really need to clean it up tomorrow) so I decided to help out with a bug in the c semantics. Turns out it was good that I did because it uncovered several tricky bugs in the bison parser generation that would have seriously confused Maria.

ehildenb (approved):

radumereuta (approved):
lgtm

PR #2358, opened 2021-12-02
Improve checks for format attribute

### LINKED ISSUE #2281: [KIP] - Give a proper error message if the user forgets to use the `[token]`
K Version
---------

```
$ kompile --version
K version:    5.1.230
Build date:   Tue Oct 19 12:09:42 CEST 2021
```

Description
-----------

When I define a token, then I need to use the `[token]` attribute.
If I don't, I get a confusing `std::map` exception during runtime:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
/usr/bin/kore-print: line 31: 57203 Aborted                 (core dumped) "$@"
```

Input Files
-----------

```
// foo.k

module FOO-SYNTAX
  syntax Range ::= r"[a-z]" // missing [token] attribute
endmodule

module FOO
  imports FOO-SYNTAX
endmodule
```

Reproduction Steps
------------------

```
kompile foo.k
krun -cPGM='x'
```

Expected Behavior
-----------------

A clear and concise error message during kompile. Like *regular expressions are only allowed when followed by a `[token]` attribute or when using `syntax lexical`*

dwightguth:
Adds errors if:
* a numerical escape sequence in a format attribute is invalid
* a numerical escape sequence in a format attribute refers to a regular expression terminal
* a production with a regular expression terminal which corresponds to a symbol in the compiled KORE does not have a format attribute.
Fixes #2281


SchmErik (approved):
Looks good to me

kore/src/main/scala/org/kframework/builtin/Sorts.scala
```diff
@@ -7,6 +7,7 @@ import org.kframework.kore.KORE.Sort
 
 object Sorts {
   val Layout = Sort("#Layout")
+  val LineMarker = Sort("#LineMarker")
```
ehildenb, line 10: I don't see `#LineMarker` documented in https://kframework.org/USER_MANUAL/. Can you add a blurb there about it (maybe in the same place that `#Layout` is described).

PR #2370, opened 2021-12-08
Add --pattern support to the llvm backend via haskell backend tool

### LINKED ISSUE #1742: Concrete execution search
This is, properly speaking, an llvm backend task, but we are creating an issue here so it can live on the K project board. The goal here is to support krun --search in the llvm backend.

dwightguth:
Basically, we add support for --pattern passed to krun when you use the llvm backend, by means of delegating the search pattern matching to the haskell backend.

Fixes #1742


k-distribution/pl-tutorial/1_k/3_lambda++/lesson_4/tests/free-variable-capture.lambda.out
```diff
@@ -1,11 +1,23 @@
...
-  V ==K ( y ~> #freezer___LAMBDA1_ ( lambda x . lambda y . x ) ~> #freezer___LAMBDA0_ ( z ) ~> #freezer___LAMBDA1_ ( a ) )
-#Or
-  V ==K ( z ~> #freezer___LAMBDA1_ ( ( lambda x . lambda y . x ) y ) ~> #freezer___LAMBDA1_ ( a ) )
-#Or
-  V ==K ( z ~> #freezer___LAMBDA1_ ( closure ( .Map , x , lambda y . x ) y ) ~> #freezer___LAMBDA1_ ( a ) )
+  {
+    V:K
+  #Equals
+    z ~> #freezer___LAMBDA_Exp_Exp_Exp1_ ( ( lambda x . lambda y . x ) y ~> . ) ~> #freezer___LAMBDA_Exp_Exp_Exp1_ ( a ~> . ) ~> .
+  }
```
ehildenb, line 23: We have less search results in the post-state than the pre. Is this intentional? (just want to make sure it's double-checked)

k-distribution/src/main/scripts/bin/krun
```diff
@@ -540,20 +540,33 @@ if [ "$backendName" = "llvm" ]; then
...
     if [[ $searchType != "FINAL" ]]; then
       error "Backend llvm does not support --search-all, --search-one-or-more-steps, or --search-one-step yet"
     fi
+    disjunctFile="$(mktemp ${tempDir}/tmp.disjunction.XXXXXXXXXX)"
+    tempFiles+=("$disjunctFile")
+    if $search; then
+      llvmCommand="$kompiledDir/search"
+    else
+      llvmCommand="$kompiledDir/interpreter"
+    fi
```
ehildenb, line 564: I don't quite understand this.....

We already said `$search || [ -n $pattern ]]` above. So does this mean you can do a normal `krun --pattern SOME_PATTERN` _without_ it doing a search, and have it try to match the single final configuration with the given pattern and print out the substitution?

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -354,7 +354,7 @@ public static Module subsortKItem(Module module) {
     }
 
     public Rule parseAndCompileRule(CompiledDefinition compiledDef, String contents, Source source, Optional<Rule> parsedRule) {
-        Rule parsed = parsedRule.orElse(parseRule(compiledDef, contents, source));
+        Rule parsed = parsedRule.orElseGet(() -> parseRule(compiledDef, contents, source));
```
ehildenb, line 357: What is this change for?

k-distribution/pl-tutorial/1_k/3_lambda++/lesson_4/tests/free-variable-capture.lambda.out
dwightguth: This is intentional. The java backend implements some custom logic to ensure that heating and cooling rules can apply at any time while also preventing them from applying in an infinitely looping cycle. The Haskell and LLVM backends do not do this, and thus require that any heated term be fully evaluated before cooling. That is why you lose two behaviors in each instance of this particular program. I have previously spoken with @grosu and he has explicitly stated he is okay with making this change for the time being in order to be able to migrate the lessons in the pl-tutorial. We can revisit and readd these behaviors when we have an actual need to implement the full nondeterminism of heating/cooling in K.

k-distribution/src/main/scripts/bin/krun
dwightguth: Yes, if you pass `--pattern` but not `--search`, it will do single-path concrete execution and then try to match the result of such execution against the pattern. This is the same behavior krun has had in the past.

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth: orElse evaluates its argument unconditionally, which is pretty clearly not desired here. Without this change, parseRule will get called twice during k-compile-search-pattern even though one of those calls will be completely unused.

k-distribution/pl-tutorial/1_k/3_lambda++/lesson_4/tests/free-variable-capture.lambda.out
ehildenb: Ok, is this recorded in an issue somewhere? We don't want to forget.

k-distribution/src/main/scripts/bin/krun
ehildenb: Cool, thanks :)

kernel/src/main/java/org/kframework/kompile/Kompile.java
ehildenb: :+1:

ehildenb (approved):

k-distribution/pl-tutorial/1_k/3_lambda++/lesson_4/tests/free-variable-capture.lambda.out
dwightguth: I will create an issue.

dwightguth: https://github.com/kframework/llvm-backend/issues/473

PR #2375, opened 2021-12-09
Revert "`imports FOO.bar` with fixed OOM"

radumereuta:
Reverts kframework/k#2292
This needs more investigation. Tracking here: https://github.com/kframework/k/issues/2357


dwightguth (approved):
If tests pass, go ahead and mark it as automerge. If they don't, I would like to review again after you've fixed it.

PR #2390, opened 2021-12-16
Cache scanner after `kompile` and reuse during `k-compile-search-pattern`

dwightguth:
This PR will serialize the scanner during kompile and then reuse that scanner when compiling search patterns rather than having to recreate the scanner. This significantly brings down the time spent in `k-compile-search-pattern` when running `krun --pattern`.

The reason this diff is so large is that we have to ensure that the set of tokens in kompile and k-compile-search-pattern are *exactly identical* and *in the same order*, thus necessitating some work to ensure that tokens in the scanner get emitted in a completely deterministic ordering.

I will make kprovex also use this scanner in a followup PR. Note, however, that this will not be possible if the specification has macros in it... If we decide to support this case, we will have to fall back on recreating the scanner in the prover if the user does this. Thoughts?

I also fix a couple other minor unrelated bugs, including:
* krun not emitting the command line invocation for k-compile-search-pattern when invoking `krun -v`
* A false negative on the "invalid token precedence" error if the second production visited in the pair of inconsistent productions happens to not have the `prec` attribute.
* delete some really old dead code


kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -190,6 +190,9 @@ public Scanner getScanner() {
         }
         return scanner;
     }
+    public void setScanner(Scanner s) {
+      scanner = s;
```
radumereuta: Everything here is indented by 4. This line is indented only by 2.

radumereuta (commented):
Just one comment.
I'll let @SchmErik give the final approval

SchmErik (approved):
Looks good to me. I think this brings up a few questions about the code base but I'll ask them internally. @dwightguth I'll approve but please address @radumereuta before placing the automerge label

PR #2394, opened 2022-01-11
try to fix release by telling github actions to use correct java version

dwightguth:
I haven't tested this (it's kinda hard to test without merging), but here is the relevant documentation about this environment variable:

* https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#env
* https://github.com/actions/virtual-environments/blob/macOS-11/20211219.1/images/macos/provision/core/openjdk.sh
* https://github.com/actions/virtual-environments/blob/macOS-11/20211219.1/images/macos/macos-11-Readme.md


ehildenb (approved):

ehildenb:
Let's give it a try!

PR #2400, opened 2022-01-14
Tests for `[non-executable]` attribute

### LINKED ISSUE #2388: [KIP] - Add `non-executable` attribute on equations
Motivation
----------

@ehildenb requested this feature. A `non-executable` equation will not be tried during function evaluation.

I believe his request comes from needing a way to specify that an `smt-lemma` equation should only be sent to Z3, and not attempted in the prover.

**Very important**: the equation should not be treated like a function definition rule, meaning it should not be translated to Kore using the unification encoding, it should not accept priority attributes etc. It should basically be "outside of the semantics", similar to how `simplification` rules are. I imagine that an equation with just the `non-executable` attribute will serve as documentation.

The Haskell backend has a branch which parses the attribute and filters the `non-executable` equations out before constructing the function evaluators: https://github.com/kframework/kore/pull/2945

Link to the issue on the Kore repository: https://github.com/kframework/kore/issues/2933

I am not sure if the LLVM backend will have to make any changes or not. 

Example K Code
--------------

```
module EXAMPLE
  imports INT

  syntax Int ::= f(Int) [function]
  rule f(X) => 3 requires X >=Int 2 [smt-lemma, non-executable]

endmodule
```

Documentation
-------------

Equations can receive the `non-executable` attribute. This means that K will not attempt to apply it during function evaluation. One instance where this attribute is useful is when indicating that an equation is an `smt-lemma`, but shouldn't be used in another context other than sending it to Z3.

There aren't any to my knowledge.

Testing Approach
----------------

I think that if the frontend is merged first, I can then think of a test for the Haskell backend CI, and duplicate it here in a follow-up PR. I think it's important to test that removing the equation from the function evaluators doesn't affect what we send to Z3.

@ehildenb do you have any minimized example with a situation where you needed this?

Baltoli:
This PR adds tests to ensure that both the LLVM and Haskell backends correctly ignore `non-executable` function rules.

Additionally
* It ensures that in the frontend, `[non-executable]` rules are ignored when compiling the implicit side condition for `[owise]` rules. This follows the logic already in place for `[simplification]` rules.
* It adds a check to the frontend that forbids the attribute from being applied to rewrite (non-function) rules. Allowing the attribute on those rules would be harmless but useless, while complicating the backend implementations. We therefore choose to forbid it.

Note that this PR is ready for review, but is blocked until https://github.com/kframework/kore/pull/2945 is merged and the corresponding submodule update is merged to K. I will back out the submodule commits in this PR when that happens, and resubmit this for a final check.


Baltoli:
This PR pulls in the backend update PRs for both Haskell and LLVM so that the tests can be run properly before the submodules are merged to master.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1207,7 +1207,9 @@ private void convertRule(RuleOrClaim rule, int ruleIndex, boolean heatCoolEq, St
     }
 
     private boolean ignoreOwise(Rule notMatching) {
```
radumereuta: Probably this function needs to be renamed if we add more behaviors.

radumereuta (commented):
Just the one comment.
lgtm otherwise.
Ping me again when test are passing.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
Baltoli: Yeah that's fair - I guess `ignoreOwise` is a contraction of "ignore rules that don't contribute to `[owise]` side condition checks", but I'll give it a better name.

radumereuta (approved):
lgtm, but maybe wait for @ehildenb to have a look too?

Baltoli:
Thanks @radumereuta - it's blocked on the Haskell backend PR in any case, so there's no rush.

Baltoli:
@radumereuta @ehildenb The submodule updates have all been merged, so this is ready for a final review now.

radumereuta:
Feel free to add automerge if you don't want to wait for @ehildenb to review.

PR #2404, opened 2022-01-20
try to address issues with redirecting stdout to file

### LINKED ISSUE #2403: [Bug] [kparse] - Weird truncation behavior with stdout
K Version
---------

```
% kompile --version
K version:    v5.2.21-44-gd7d340bd4f
Build date:   Thu Jan 20 01:52:53 UTC 2022
```

Description
-----------

`kparse` is truncating the output of any file it's redirected to when using `--gen-glr-bisor-parser`. This results in weird behavior on CI (output looks strange).

Input Files
-----------

File `test.k`:

```
module TEAL-SYNTAX
  import INT-SYNTAX
endmodule

module TEAL-PARSER-SYNTAX
  imports TEAL-SYNTAX
endmodule

module TEAL-DRIVER
  imports TEAL-SYNTAX

  configuration
      <k> $PGM:Int </k>
endmodule
```

File `pgm1.teal`:

```
1
```

File `pgm2.teal`:

```
2
```

Reproduction Steps
------------------

Kompile the semantics with `--gen-glr-parser`, then call `kparse` on both programs (one after the other), redirecting the output to file `output`, and look at the contents of file `output` between each step. Note that we are using appending redirect for the second call to `kparse`. Also note that this reproduces on CI (an Ubuntu Focal image, and I'm also on Ubuntu Focal).

```
% kompile --backend llvm --verbose --gen-glr-bison-parser \
    --main-module TEAL-DRIVER                           \
    --syntax-module TEAL-PARSER-SYNTAX                  \
    test.k
Parse command line options                                   =  0.020s
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/prelude.md)
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/kast.md)
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/domains.md)
Importing: Source(/home/dev/src/k/test.k)
Outer parsing [64 modules]                                   =  1.056s
Parse configurations [0/3 declarations]                      =  0.411s
  New scanner: TEAL-DRIVER-RULE-CELLS                        =  0.188s
Parse rules [0/75 rules]                                     =  0.241s
Validate definition                                          =  0.183s
Apply compile pipeline                                       =  0.363s
  New scanner: TEAL-PARSER-SYNTAX-PROGRAM-GRAMMAR            =  0.087s
  New Bison parser: TEAL-PARSER-SYNTAX-PROGRAM-GRAMMAR       =  0.334s
Kompile to kore                                              =  0.343s
Save to disk                                                 =  0.045s
Backend                                                      =  1.204s
Total                                                        =  3.867s

% cat output


% kparse pgm1.teal >> output

% cat output
\dv{SortInt{}}("1")

% kparse pgm2.teal >> output

% cat output
\dv{SortInt{}}("2")
```

Note that I injected an extra blank line before each prompt for readability (the prompts are denoted with `%`).

Expected Behavior
-----------------

I expect `kparse` to not truncate whatever stream it's writing to when using the bison parser output. So the final `cat output` should result in:

```
% cat output
\dv{SortInt{}}("1")
\dv{SortInt{}}("2")
```

dwightguth:
Previously there was a bug where if the user did not specify --output-file to krun, kore-print, or kparse, but nonetheless redirected the output of the command to a file, the file would be truncated even if the user told the shell to append output. This happened because of a particular nuance of how we had been implementing the --output-file flag. This PR should solve this issue by means of adding some extra logic to those scripts. We also take the opportunity to refactor slightly and move some common code into a common script stored in the lib directory that gets sourced by multiple bash scripts.

Fixes #2403


k-distribution/src/main/scripts/lib/k.sh
```diff
@@ -0,0 +1,27 @@
...
+    "$@" 1>&2
+  else
+    "$@" > "$target"
+  fi
+}
+
+catNewline () {
+  cat "$@"
+  echo
+}
```
radumereuta: Please add some comments to describe why this file exists and what it contains.
Basically what you just wrote in the PR description, but add it here so it's easily accessible in the future.

radumereuta (dismissed):

k-distribution/src/main/scripts/lib/k.sh
ehildenb: Well it's also worth it to mention at the top of the file that this file is for utilities used across all our scripts, and so any common bash functionalities should be factored out to here.

radumereuta: Maybe rename this file to `utils` or `lib`?

ehildenb:
I've tested this against the TEAL semantics and confirmed that it fixes the initial problem. I'd still like to see tests for it (by having a custom `KPARSE=./kparseTwice` script as Dwight and I discussed on Slack, but if the extra documentation is enough for @radumereuta  then feel free to get merged.

ehildenb (approved):

Issue #2411, opened 2022-01-21
Test suite missing a segmentation fault


ehildenb:
Going back quite a ways, there has been a segmentation fault in the K test suite unnoticed.

This test two months ago: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/K/detail/release/175/pipeline/68

- Click "Build and Test K"
- Click "Show complete log"
- Search for "Segmentation"

Our testing infrastructure is missing this segfault somehow.

Pretty sure it's `issues-1602`, that one is segfaulting on my machine, discovered when the test started failing on: https://github.com/kframework/k/pull/2410

Baltoli:
Initial investigation:
* The underlying segmentation fault is an infinite recursion in the LLVM backend's Kore parser. I'll minimise the test case and raise an issue there.
* ~~The failure not being reported seems to be a subshell somewhere not picking up a return code.~~

Baltoli:
The reporting failure is down to `pipefail` not being set at the top level of the test suite Makefiles. This means that when we construct command pipelines like:
```
cat foo.test.in 2>/dev/null \
  | $k-dist/bin/krun foo.test --dry-run 1>/dev/null -v --no-exc-wrap  -d . \
  | diff - foo.test.out
```

we don't consider that the `krun` invocation itself could fail with a non-zero exit code; the pipeline is allowed to continue executing.

This normally wouldn't be a problem, because the `krun` invocation wouldn't produce the expected output, and the `diff` command would catch this as a failing test. However, in this test, the expected output is empty, and the `diff` invocation succeeds despite the segfault.

I think the best thing to do here is:
* Set `pipefail` globally on the test suite by overriding the `SHELL` Makefile variable. This seems to be the [consensus solution](https://stackoverflow.com/questions/33925523/how-to-set-pipefail-in-a-makefile), but it may break things unexpectedly.
* Disable this test and re-run the test suite with `pipefail` to make sure we're not missing anything else important.
* Fix the underlying problem in the backend's macro expander.
* Re-enable the test.

Baltoli:
There's a workaround for the segfault itself - because it's not an infinite recursion, and is just a really huge term, you can use `ulimit -s` to allocate more stack space. On my machine this fixes the test.

Baltoli:
[...]

PR #2413, opened 2022-01-25
Fix parser segfault in test suite

Baltoli:
The input to this test was deliberately constructed to exceed the stack size allocated to the bison parser by default. Unfortunately, it does the same to the Kore parser in the LLVM backend.

This PR fixes the issue by disabling macro expansion in the `krun` pipeline for this test. There are no macros in this test, so this is a safe workaround.

If this issue crops up again in the future, the workaround from a user perspective is to set a larger stack size using `ulimit -s`.


Baltoli:
Despite the lack of `pipefail` in the test suite, you can convince yourself that this fixes the underlying issue by opening the full test log and searching for "segmentation", as @ehildenb describes on #2411.

dwightguth (approved):

PR #2417, opened 2022-01-25
Kill the ocaml backend

dwightguth:
Pretty straightforward; deletes the ocaml backend and removes all references to it from the codebase except in the changelog.

There shouldn't be anything in the first commit that's not a straight deletion.


kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -200,18 +200,18 @@ public Module executionModule() {
...
     public K parseSingleTerm(Module module, Sort programStartSymbol, KExceptionManager kem, String s, Source source) {
-        try (ParseInModule parseInModule = RuleGrammarGenerator.getCombinedGrammar(module, kompileOptions.strict())) {
+        try (ParseInModule parseInModule = RuleGrammarGenerator.getCombinedGrammar(module, true)) {
             Tuple2<Either<Set<KEMException>, K>, Set<KEMException>> res = parseInModule.parseString(s, programStartSymbol, source);
             kem.addAllKException(res._2().stream().map(e -> e.getKException()).collect(Collectors.toSet()));
             if (res._1().isLeft()) {
                 throw res._1().left().get().iterator().next();
             }
-            return new TreeNodesToKORE(Outer::parseSort, kompileOptions.strict()).down(res._1().right().get());
+            return new TreeNodesToKORE(Outer::parseSort, true).down(res._1().right().get());
```
ehildenb, line 209: Is there ever a case where we don't pass `true` to this function anymore then? Same with `getCombinedGrammar(...)`. If so, should we just remove this argument altogether? Maybe you're planning a follow-up PR?

dwightguth: It turns out there is at least one case where false is passed to this function. We could probably change it since the case in question was just a toString method, but I chose to limit the size of the diff rather than trying to incorporate complex non-deletion changes into a very large diff. We can definitely perform further cleanups later if we want to, but I'd rather keep this diff as close to just deleting things as we easily can.

ehildenb (approved):

radumereuta (approved):
awesome

PR #2420, opened 2022-02-01
Fix version script when releases are missing

Baltoli:
This change walks back through the rev-list on the `release` branch to find the most recently tagged version, and uses this as the merge base from which version strings are constructed.

Doing so prevents the output from breaking if the most recent commits on `master` don't have releases branching from them (for example, if the release process has become broken).

Example output:
```console
$ kompile --version
K version:    v5.2.71-4-gc7b822b5d4
Build date:   Tue Feb 01 10:22:10 GMT 2022
```

### DIFF
--- a/scripts/get-version-string.sh
+++ b/scripts/get-version-string.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-MERGE_BASE=$(git merge-base master HEAD)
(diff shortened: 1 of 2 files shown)


radumereuta (approved):
Let's see if this way is more resilient.

PR #2427, opened 2022-02-07
Install make in empty test container

Baltoli:
I found that #2424 wasn't the right solution; it was installing Make in the container that _builds_ the K `.deb` file.

Instead, what we need to do is make sure that it's installed in the container that installs the `.deb` and runs the tests.

I'm not sure what's changed to cause us to no longer pull in Make on Ubuntu Focal, but I think this is the right solution, as it encodes the dependency on Make only for our own use case (rather than for all consumers of the package).


dwightguth:
The answer as to why this is happening now is that we no longer pull in opam, btw.

dwightguth (approved):

PR #2428, opened 2022-02-08
Add flag --emit-json-spec to kprovex

### LINKED ISSUE #2288: Support dumping json empty configuration at kompile time
Currently when using the pyk library, we have to get a "blank" configuration to use for manipulating K terms and stuff. To produce this, I run a garbage proof out to depth 0, and strip the configuration. For example, I may run this proof out to depth 0:

```
claim <k> #execute => . ... </k>
```

This will give me a configuration that looks like this:

```
<generatedTop>
  <kevm>
    <k> #execute ... </k>
    <mode> _0 </mode>
    <schedule> _1 </schedule>
    ...
</generatedTop>
```

Which I then strip to look like:

```
<generatedTop>
  <kevm>
    <k> K_CELL </k>
    <mode> MODE_CELL </mode>
    <schedule> SCHEDULE_CELL </schedule>
    ...
</generatedTop>
```

So it would be nice if when I call `kompile --emit-json`, it dumped the empty configuration to a file `empty-config.json` in the `*-kompiled` directory. Then I could just read that file off instead using this ad-hoc way of generating it.

ehildenb:
Fixes: #2288 

This PR does:

- Add flag to `kprovex` called `--emit-json-spec` which emits the spec module as JSON to the specified file.
- Fixes the JSON serialization of `KImport` to include the `node` header (all KAST JSON terms must have `node` header).
- Adds unit tests of the pyk library which do:
  - Serialize pretty proof to json using new flag -> read json proof in pyk and write out as pretty to file -> run kprovex on the newly serialized proof using pyks `KProve.prove()`
  - Serialize pretty proof to json using new flag -> read json proof in pyk and prove using `KProve.proveClaim()`.

### DIFF
--- a/k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
+++ b/k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
@@ -1,0 +1,83 @@
+import json
+import os
+import shutil
+import unittest
+from pathlib import Path
+
+from pyk.kast import KApply, KClaim, KDefinition, KRequire, KVariable
+from pyk.kastManip import rewriteAnywhereWith
+from pyk.ktool import KProve
+
+
+class EmitJsonSpecTest(unittest.TestCase):
+    TEST_DIR = 'emit-json-spec-tests'
+    MAIN_FILE = 'verification.k'
+    USE_DIR = f'{TEST_DIR}/.kprove'
+
+    def setUp(self):
+        shutil.rmtree(self.USE_DIR, ignore_errors=True)
+        os.makedirs(self.USE_DIR)
+
+        self.kprove = KProve(f'{self.TEST_DIR}/verification-kompiled', self.MAIN_FILE, self.USE_DIR)
+        self.kprove.proverArgs += ['-I', '.']
+        self._update_symbol_table(self.kprove)
+
+        with open(f'{self.TEST_DIR}/looping-spec.json', 'r') as spec_file:
+            module = json.load(spec_file)['term']
+
+        claim = extract_claims(module)[0]
+        self.claim = KClaim(
+            body=eliminate_generated_top(claim['body']),
+            requires=claim['requires'],
+            ensures=claim['ensures'],
+            att=None
+        )
+
+        module['localSentences'] = [self.claim]
+        self.module = module
+
+    def tearDown(self):
+        shutil.rmtree(self.USE_DIR)
+
+    @staticmethod
+    def _update_symbol_table(kprove):
+        def paren(f):
+            def unparse(*args):
+                return '(' + f(*args) + ')'
+            return unparse
+
+        kprove.symbolTable['_+Int_'] = paren(kprove.symbolTable['_+Int_'])
+
+    def test_prove_claim(self):
+        # When
+        result = self.kprove.proveClaim(self.claim, 'looping-1')
+
+        # Then
+        self.assertEqual(result['label'], '#Top')
+
+    def test_prove(self):
+        # Given
+        spec_name = 'looping-2-spec'
+        spec_path = Path(f'{self.USE_DIR}/{spec_name}.k')
+        spec_module_name = spec_name.upper()
+
+        self.module['name'] = spec_module_name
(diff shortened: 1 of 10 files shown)


k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(result['label'], '#Top')
+
+    def test_prove(self):
+        # Given
+        spec_name = 'looping-2-spec'
+        spec_path = Path(f'{self.USE_DIR}/{spec_name}.k')
+        spec_module_name = spec_name.upper()
+
+        self.module['name'] = spec_module_name
+        definition = KDefinition(self.module, [self.module], requires=[KRequire(self.MAIN_FILE)])
```
tothtamas28, line 65: Maybe `--emit-json-spec` could return the module in a `KDefinition` with requires (it is easier to extract the module if the definition is not needed than to reconstruct the definition when it is). What do you think, @ehildenb?

kernel/src/main/java/org/kframework/kprovex/KProve.java
```diff
@@ -85,6 +87,18 @@ public int run() {
...
+        if (kproveOptions.emitJsonSpec != null) {
+            try {
+                String out = new String(ToJson.apply(specModule));
+                PrintWriter fOut = new PrintWriter(kproveOptions.emitJsonSpec);
+                fOut.println(out);
+                fOut.close();
+            } catch (FileNotFoundException e) {
+                System.err.println("Could not open node output file: " + kproveOptions.emitJsonSpec);
+                e.printStackTrace();
+            }
```
radumereuta: This utility already incorporates all the typical K error handling.
```suggestion
            FileUtil.save(new File(kproveOptions.emitJsonSpec), ToJson.apply(specModule));
```

ehildenb: I should also fix the location in the source-code that I copy-pasted this pattern from.

ehildenb: In `StateLog.java` I believe.

ehildenb: Fixed.

k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
ehildenb: It's actually not clear to me that it makes sense to do so. The "Definition" associated with the `specModule` is not super well defined, and is also not directly computed by the current frontend. I would basically have 2 choices:

- A definition with _just_ the `specModule` in it, so I'd be filling in sensible dummy parameters for it like you do here in pyk, so that the requires is set up correctly basically. `specModule` becomes the `mainModule` of this definition, which isn't very sensible.
- Assume that the user wants the entire kompiled definition, and not just a dummy definition with the single module in it. Once again the `specModule` becomes the `mainModule` of this definition, which isn't very sensible.

So I'm inclined to say "emitting just the module is more flexible, and it's up to the user to determine how to use that info". Since it's enough for now to accomplish our goals, I think we should wait and see if more info/structure than that is needed.

k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(result['label'], '#Top')
+
+
+def extract_claims(module):
+    return [claim for claim in module['localSentences'] if claim['node'] == 'KClaim']
+
+
+def eliminate_generated_top(term):
+    rule = KApply('<generatedTop>', [KVariable('CONFIG'), KVariable('_')]), KVariable('CONFIG')
+    return rewriteAnywhereWith(rule, term)
```
ehildenb, line 83: @tothtamas28 I think these should be moved to the pyk library itself, in `kast.py` or `kastManip.py`. But we may want to wait to do that for when we begin upstreaming more things from summarizer, if you just want to mark it down in an issue.

ehildenb:
@radumereuta this is ready for review.

radumereuta (approved):

PR #2434, opened 2022-02-12
Allow specifying lemmas in pyks KProve.proveClaim

ehildenb:
Part of: https://github.com/runtimeverification/ksummarize/issues/39

~Blocked on: https://github.com/runtimeverification/k/pull/2438~

Thanks to https://github.com/runtimeverification/k/pull/2345 we are able to add lemmas to the specification module.

This PR:

- Allows threading through list of lemmas to pyk's `KProve.proveClaim()` method.
- Adds a test to demonstrate a failing proof that passes when a lemma is added.
- Prints out rules/claims without all the extra space between rule/claim body and attributes if the requires/ensures clauses are empty.


k-distribution/tests/pyk/emit-json-spec-tests/emit_json_spec_test.py
```diff
@@ -73,6 +73,17 @@ def test_prove(self):
         # Then
         self.assertEqual(result['label'], '#Top')
 
+    def test_prove_claim_with_lemmas(self):
```
tothtamas28: This test could go into its own module (including a separate `verification.k` file), what do you think?

(Perhaps we should also extract a superclass `PykTestCase` that takes care of the initialization of the KProve instance.)

ehildenb: I will make these updates after this one goes through: https://github.com/runtimeverification/k/pull/2438

k-distribution/tests/pyk/simple_proofs_test.py
```diff
@@ -0,0 +1,41 @@
+import json
+from pathlib import Path
+
+from kprove_test import KProveTest
+from pyk.kast import KApply, KAtt, KClaim, KRule, KToken, KVariable
+from pyk.kastManip import rewriteAnywhereWith
+
+
+class EmitJsonSpecTest(KProveTest):
```
tothtamas28:
```suggestion
class SimpleProofsTest(KProveTest):
```

k-distribution/tests/pyk/simple_proofs_test.py
```diff
@@ -0,0 +1,41 @@
...
+    @staticmethod
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_prove_claim_with_lemmas(self):
+        # When
+        new_lemma = KRule(KToken('pred1(3) => true', 'Bool'), requires=KToken('pred1(4)', 'Bool'), att=KAtt(atts={'simplification': ''}))
+        new_claim = KClaim(KToken('<k> foo => bar ... </k> <state> 3 |-> 3 </state>', 'TCellFragment'), requires=KToken('pred1(4)', 'Bool'))
+        result1 = self.kprove.proveClaim(new_claim, 'claim-without-lemma')
+        result2 = self.kprove.proveClaim(new_claim, 'claim-with-lemma', lemmas = [new_lemma])
```
tothtamas28:
```suggestion
        # Given
        new_lemma = KRule(KToken('pred1(3) => true', 'Bool'), requires=KToken('pred1(4)', 'Bool'), att=KAtt(atts={'simplification': ''}))
        new_claim = KClaim(KToken('<k> foo => bar ... </k> <state> 3 |-> 3 </state>', 'TCellFragment'), requires=KToken('pred1(4)', 'Bool'))
        
        # When
        result1 = self.kprove.proveClaim(new_claim, 'claim-without-lemma')
        result2 = self.kprove.proveClaim(new_claim, 'claim-with-lemma', lemmas = [new_lemma])
```

tothtamas28 (approved):
LGTM (except a few small issues, see below).

k-distribution/tests/pyk/simple_proofs_test.py
ehildenb: Done.

PR #2441, opened 2022-02-16
Add the `kit` command line tool

tothtamas28:
Fixes https://github.com/runtimeverification/erc20-verification/issues/329

### DIFF
--- a/pyk/Makefile
+++ b/pyk/Makefile
@@ -17,6 +17,7 @@ clean:
 	rm -rf $(VENV_DEV_DIR) $(VENV_PROD_DIR) definitions build src/pyk.egg-info .myp_cache
 	find -type d -name __pycache__ -prune -exec rm -rf {} \;
 	$(MAKE) -C pyk-tests clean
+	$(MAKE) -C kit-tests clean
 
 
 # Virtualenv
@@ -49,7 +50,7 @@ $(VENV_PROD_DIR): $(VENV_PROD)
 
 # Tests
 
-test: test-unit test-integration test-pyk
+test: test-unit test-integration test-pyk test-kit
 
 test-unit: $(VENV_DEV)
 	$(ACTIVATE_DEV) && python3 -m unittest discover -vt src pyk.tests
@@ -59,7 +60,9 @@ test-integration: $(VENV_DEV)
 
 test-pyk: $(VENV_PROD)
 	$(ACTIVATE_PROD) && $(MAKE) -C pyk-tests
-	rm -rf $(VENV_PROD_DIR) build src/pyk.egg-info
+
+test-kit: $(VENV_PROD)
+	$(ACTIVATE_PROD) && $(MAKE) -C kit-tests
 
 
 # Checks
--- a/pyk/kit-tests/Makefile
+++ b/pyk/kit-tests/Makefile
@@ -1,0 +1,23 @@
+K_ROOT := $(abspath ../..)
+K_BIN  := $(K_ROOT)/k-distribution/target/release/k/bin
+
+export PATH := $(K_BIN):$(PATH)
+
+KIT := kit
+
+
+.PHONY: default clean test
+
+default: test
+
+clean:
+	rm -rf .kit
+	rm -rf .kprove-*
+
+test:
+	$(KIT) init --force verification.k looping-spec.k
+	$(KIT) expand 3a8c..
+	$(KIT) expand 3442..
+	$(KIT) expand b528..
+	$(KIT) expand 0678..
+	$(KIT) cover  1846.. --by b528.. --semantic
--- a/pyk/kit-tests/control-flow.k
+++ b/pyk/kit-tests/control-flow.k
@@ -1,0 +1,90 @@
+module CONTROL-FLOW-SYNTAX
+  imports INT-SYNTAX
+  imports BOOL-SYNTAX
+  imports ID-SYNTAX
+
+  syntax Exp ::= IExp | BExp
+
+
+  syntax IExp ::= Int
+                | Id
+                | "(" IExp ")" [bracket]
+
+  syntax IExp ::= IExp "*" IExp [seqstrict]
+                | IExp "/" IExp [seqstrict]
+                | IExp "+" IExp [seqstrict]
+                | IExp "-" IExp [seqstrict]
+                | IExp "^" IExp [seqstrict]
+
+  syntax BExp ::= Bool
+                | "(" BExp ")" [bracket]
+
+  syntax BExp ::= IExp "==" IExp [seqstrict]
+                | IExp "!=" IExp [seqstrict]
+                | IExp ">"  IExp [seqstrict]
+                | IExp ">=" IExp [seqstrict]
+                | IExp "<"  IExp [seqstrict]
+                | IExp "<=" IExp [seqstrict]
+
+  syntax BExp ::= BExp "&&" BExp [seqstrict]
+                | BExp "||" BExp [seqstrict]
+
+
+  syntax Stmt ::= Id "=" IExp ";" [strict(2)]
+                | Stmt Stmt [left]
+                | "if" "(" BExp ")" Block "else" Block [strict(1)]
+                | "while" "(" BExp ")" Block
+
+  syntax Block ::= "{" "}"
+                 | "{" Stmt "}"
+endmodule
+
+
+module CONTROL-FLOW
+  imports CONTROL-FLOW-SYNTAX
+  imports INT
+  imports BOOL
+  imports MAP
+
+  configuration <T>
+                  <k> $PGM:Stmt </k>
+                  <mem> .Map </mem>
+                </T>
+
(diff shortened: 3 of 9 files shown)


k-distribution/src/main/scripts/lib/kit/state.py
```diff
@@ -0,0 +1,82 @@
...
+        return State(config, constraints)
+
+    def as_term(self) -> Dict:
+        return ml_and(list(self))
+
+    def as_rule(self, next_state: 'State', path_constraints: List[Dict] = [], *, claim=False) -> Dict:
+        pre = self.let(constraints=self.constraints + path_constraints)
+        post = next_state.let(constraints=[c for c in next_state.constraints if c not in self.constraints])
+
+        body = pushDownRewrites(KRewrite(pre.config, post.config))
```
ehildenb: This is basically the old `buildRule`?

k-distribution/src/main/scripts/lib/kit/utils.py
```diff
@@ -0,0 +1,153 @@
...
+    if not path.exists():
+        raise ValueError(f'File does not exist: {path}')
+    if not path.is_file():
+        raise ValueError(f'Path is not a file: {path}')
+
+
+def file_path(s: str) -> Path:
+    path = Path(s)
+    check_file_path(path)
+    return path
```
ehildenb: Maybe we should have a `pyk/file_util.py` for these ones. Or something similar.

k-distribution/src/main/scripts/lib/kit/utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+def merge_with(f, d1: Mapping, d2: Mapping) -> Dict:
+    res = dict(d1)
+    for k, v2 in d2.items():
+        if k in d1:
+            v1 = d1[k]
+            res[k] = f(v1, v2)
+        else:
+            res[k] = v2
+    return res
```
ehildenb: I think these ones can be moved to `pyk/util.py`

k-distribution/src/main/scripts/lib/kit/utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def eliminate_generated_top(term):
+    rule = KApply('<generatedTop>', [KVariable('CONFIG'), KVariable('_')]), KVariable('CONFIG')
+    return rewriteAnywhereWith(rule, term)
+
+
+def extract_lhs(term):
+    return traverseTopDown(term, if_rewrite(get_lhs))
+
```
ehildenb: We should make a PR which cleans up `kastManip.py` according to the new style guide, and organizes it a bit (with headers explaining the sections). Bits that shouldn't exist there should find new homes.

Then these (and most of the rest of this file) can be migrated there.

k-distribution/src/main/scripts/lib/kit/utils.py
ehildenb: So we can have a PR which cleans up `util.py` and organizes it, and styles it.

k-distribution/src/main/scripts/lib/kit/state.py
tothtamas28: Yes.

k-distribution/src/main/scripts/lib/kit/__main__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Dummy command for testing
+    command_subparsers.add_parser('prove', help='Load prover from config file.')
+
+    return argument_parser
+
+
+def init(args):
+    if not args.work_dir.exists():
+        raise ValueError(f'Working directory doe not exist: {args.work_dir}')
```
ehildenb:
```suggestion
        raise ValueError(f'Working directory does not exist: {args.work_dir}')
```

k-distribution/src/main/scripts/lib/kit/__main__.py
```diff
@@ -0,0 +1,115 @@
...
+    with open(cfg_file, 'w') as f:
+        f.write(cfg.to_json())
+        f.write('\n')
+
+
+def _extract_claim_from_spec_file(spec_file: Path) -> KClaim:
+    with open(spec_file, 'r') as f:
+        spec = json.load(f)
+
+    module = KFlatModule.from_dict(spec['term'])
```
ehildenb: Would it be possible to make it so we can say:

```suggestion
    module = KFlatModule(spec['term'])
```

instead?

ehildenb: Or maybe this isn't a good idea, not sure.

k-distribution/src/main/scripts/lib/kit/__main__.py
```diff
@@ -0,0 +1,115 @@
...
+    return claim
+
+
+def _create_cfg_from_claim(claim: KClaim) -> KCFG:
+    init, target = ConstrainedTerm.from_claim(claim)
+    cfg = KCFG()
+    _, init_id = cfg.insertNode(init.as_term())
+    _, target_id = cfg.insertNode(target.as_term())
+    cfg.init = init_id
+    cfg.target = target_id
```
ehildenb: `target` states needs to be a list or set:

```suggestion
    cfg.target = [target_id]
```

k-distribution/src/main/scripts/lib/kit/constrained_term.py
```diff
@@ -0,0 +1,82 @@
...
+        return ConstrainedTerm(config, [constraint])
+
+    @staticmethod
+    def from_claim(claim) -> Tuple['ConstrainedTerm', 'ConstrainedTerm']:
+        body = drop_var_prefixes(claim.body)
+        lhs = extract_lhs(body)
+        rhs = extract_rhs(body)
+        requires = drop_var_prefixes(claim.requires)
+        ensures = drop_var_prefixes(claim.ensures)
+        return ConstrainedTerm(lhs, [requires]), ConstrainedTerm(rhs, [ensures])
```
ehildenb: I think splitting into LHS and RHS should be handled somewhere else perhaps, maybe as a method of `KInner`.

So this should instead return something like `ConstrainedTerm(claim.body, claim.requires #And claim.ensures)`.

We could also lift functions like `.lhs()` and `.rhs()` from `KInner` to this class as well, by computing, `removeUselessConstraints(mlAnd([getLHS(self.config)] + self.constraints))`.

k-distribution/src/main/scripts/lib/kit/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    return traverseTopDown(term, if_ktype(KRewrite, lambda rw: rw.lhs))
+
+
+def extract_rhs(term: KInner) -> KInner:
+    return traverseTopDown(term, if_ktype(KRewrite, lambda rw: rw.rhs))
+
+
+def drop_var_prefixes(term: KInner) -> KInner:
+    term = traverseTopDown(term, if_ktype(KVariable, drop_ques))
+    term = traverseTopDown(term, if_ktype(KVariable, drop_unds))
```
ehildenb: Oh this is so easy to read, and compact!

k-distribution/src/main/scripts/lib/kit/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    rewriteAnywhereWith,
+    traverseTopDown,
+)
+
+KI = TypeVar('KI', bound=KInner)
+
+
+def eliminate_generated_top(term: KInner) -> KInner:
+    rule = KApply('<generatedTop>', [KVariable('CONFIG'), KVariable('_')]), KVariable('CONFIG')
+    return rewriteAnywhereWith(rule, term)
```
ehildenb: What do you think of being able to write instead:

```suggestion
def eliminate_generated_top(term: KInner) -> KInner:
    rule = KRewrite(KApply('<generatedTop>', [KVariable('CONFIG'), KVariable('_')]), KVariable('CONFIG'))
    return rewriteAnywhereWith(rule, term)
```

Not a huge difference, basically have the various `rewrite*` functions take a `KRewrite` as their first argument, instead of just a tuple of LHS/RHS. It seems appropriate, and more structured.

k-distribution/src/main/scripts/lib/kit/__main__.py
tothtamas28: It is doable, though I think it is best to have a single, straightforward `__init__` function (that basically just provides values for each attribute) and static factory methods for all the rest.

k-distribution/src/main/scripts/lib/kit/utils.py
tothtamas28: Yes, I think this is a good idea. Such rewrite / replace functions can even be methods of the `KRewrite` class.

k-distribution/src/main/scripts/lib/kit/__main__.py
ehildenb: Only if you think ti makes sense and is "pythonic"

PR #2448, opened 2022-02-18
Refactor `pyk` AST functions into classes

### LINKED ISSUE #2444: Separate out `kprelude.py` files
There are many places in the pyk library where we define helpers for building standard K terms. For example the `mlTop()`, `mlAnd()`, `intToken(_)`, `boolToken(_)`, etc...

These should be factored out as helpers in a file `k-prelude.py`. Ideally, we would have that somewhat match the structure of the actual builtins for K. Another option would be `k-builtins.py` or something.

Link to builtins: https://github.com/runtimeverification/k/tree/master/k-distribution/include/kframework/builtin

### LINKED ISSUE #2447: Architecture of pyk
Currently the architecture of pyk is:

- File `kast.py`:
  - Defines JSON data-structures of KAST.
  - Defines pretty-printer for KAST.
  - No error-checking anywhere.
- File `kastManip.py`:
  - Flat list of functions.
  - Mostly functions which do manipulations on KAST terms (specifically inner terms), eg `match`, `rewriteWith`, etc...

Proposed structure:

- Abstract class `KAST`:
  - `to_json` and `from_json` required to have implementations
- Subclass `KInner(KAST)`:
  - Inner syntax of K (for dataclasses like `KApply(KInner)`, `KToken(KInner)`, etc...)
  - Provides functions like `match`, so you can do things like `this.match(that)`, or `this.rewriteWith(KRewrite)`.
- Subclass `KOuter(KAST)`:
  - Outer syntax of K (for dataclasses like `KFlatModule(KOuter)`, `KImport(KOuter)`, `KProduction(KOuter)`, etc...)
  - Potential to have much better validation at construction time using types (catch errors sooner).
  - Subclass `KDefinition(KOuter)`:
    - Holds methods like `buildSymbolTable` and `prettyPrint`.

Anything that doesn't fit in this class structure can stay as flat list of functions in some file somewhere.

tothtamas28:
Fixes: https://github.com/runtimeverification/k/issues/2447
Fixes: https://github.com/runtimeverification/k/issues/2444

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,7 +126,10 @@ pipeline {
     }
     stage('Build and Package on Ubuntu Focal') {
       when {
-        branch 'release'
+        anyOf {
+          branch 'release'
+          changeRequest()
+        }
         beforeAgent true
       }
       stages {
@@ -247,7 +250,10 @@ pipeline {
     }
     stage('Build and Package on Arch Linux') {
       when {
-        branch 'release'
+        anyOf {
+          branch 'release'
+          changeRequest()
+        }
         beforeAgent true
       }
       stages {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,10 +126,7 @@ pipeline {
     }
     stage('Build and Package on Ubuntu Focal') {
       when {
-        anyOf {
-          branch 'release'
-          changeRequest()
-        }
+        branch 'release'
         beforeAgent true
       }
       stages {
@@ -250,10 +247,7 @@ pipeline {
     }
     stage('Build and Package on Arch Linux') {
       when {
-        anyOf {
-          branch 'release'
-          changeRequest()
-        }
+        branch 'release'
         beforeAgent true
       }
       stages {
--- a/k-distribution/src/main/scripts/lib/pyk/kast.py
+++ b/k-distribution/src/main/scripts/lib/pyk/kast.py
@@ -1,255 +1,1085 @@
 import json
 import sys
+from abc import ABC, abstractmethod
+from dataclasses import InitVar, dataclass
+from enum import Enum
+from itertools import chain
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Final,
+    FrozenSet,
+    Iterable,
+    Iterator,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    final,
+    overload,
+)
 
-from .cli_utils import fatal, notif, warning
-from .utils import combine_dicts
+from typing_extensions import TypeAlias
 
+from .cli_utils import fatal, warning
 
-def KApply(label, args):
-    return {'node': 'KApply', 'label': label, 'variable': False, 'arity': len(args), 'args': args}
+T = TypeVar('T', bound='KAst')
 
 
-def isKApply(k):
-    return k['node'] == 'KApply'
+class KAst(ABC):
 
+    @classmethod
+    @abstractmethod
+    def from_dict(cls: Type[T], d: Dict[str, Any]) -> 'KAst':
+        node = d['node']
+        return globals()[node].from_dict(d)
 
-def KAs(pattern, alias, att=None):
-    return {'node': 'KAs', 'pattern': pattern, 'alias': alias, 'att': att}
+    @abstractmethod
+    def to_dict(self) -> Dict[str, Any]:
+        ...
 
+    @staticmethod
+    def from_json(s: str) -> 'KAst':
+        return KAst.from_dict(json.loads(s))
 
-def isKAs(k):
-    return k['node'] == 'KAs'
+    @final
+    def to_json(self) -> str:
+        return json.dumps(self.to_dict(), sort_keys=True)
 
+    @classmethod
+    def _check_node(cls: Type[T], d: Dict[str, Any], expected: Optional[str] = None) -> None:
+        expected = expected if expected is not None else cls.__name__
+        actual = d['node']
+        if actual != expected:
+            raise ValueError(f"Expected '{expected}' as 'node' value, found: '{actual}'")
 
-def KConstant(label):
-    return KApply(label, [])
 
+@final
+@dataclass(frozen=True)
+class KAtt(KAst, Mapping[str, Any]):
+    _atts: FrozenSet[Tuple[str, Any]]
 
-def isKConstant(k):
-    return isKApply(k) and len(k['args']) == 0
+    def __init__(self, atts: Mapping[str, Any] = {}):
+        object.__setattr__(self, '_atts', frozenset(atts.items()))
 
+    def __getitem__(self, key: str) -> Any:
+        return self.atts[key]
 
-def KSequence(items):
-    return {'node': 'KSequence', 'arity': len(items), 'items': items}
+    def __iter__(self) -> Iterator[str]:
+        return (k for k, _ in self._atts)
 
+    def __len__(self) -> int:
+        return len(self._atts)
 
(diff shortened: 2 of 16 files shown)


k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -1,216 +1,327 @@
...
+class KToken(KInner):
+    token: str
+    sort: str
+
+    def __init__(self, token: str, sort: str):
+        object.__setattr__(self, 'token', token)
+        object.__setattr__(self, 'sort', sort)
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {'node': 'KToken', 'token': self.token, 'sort': self.sort}
```
ehildenb: Why don't `KVariable` and `KToken` have `copy` method and `KApply` does?

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -366,19 +363,19 @@ def _pushDownRewrites(_kast):
...
                 return lhs
-            if isKVariable(lhs) and isKVariable(rhs) and lhs['name'] == rhs['name']:
+            if type(lhs) is KVariable and type(rhs) is KVariable and lhs.name == rhs.name:
                 return lhs
-            if isKApply(lhs) and isKApply(rhs) and lhs['label'] == rhs['label'] and len(lhs['args']) == len(rhs['args']):
-                newArgs = [KRewrite(lArg, rArg) for (lArg, rArg) in zip(lhs['args'], rhs['args'])]
-                return KApply(lhs['label'], newArgs)
+            if type(lhs) is KApply and type(rhs) is KApply and lhs.label == rhs.label and lhs.arity == rhs.arity:
+                newArgs = [KRewrite(lArg, rArg) for (lArg, rArg) in zip(lhs.args, rhs.args)]
+                return lhs.copy(args=newArgs)
```
ehildenb: Ah, is it because sometimes we want to `copy` a `KApply` in this custom way?

k-distribution/src/main/scripts/lib/pyk/kast.py
tothtamas28: Good point, I'll make sure to add them (and some other utility functions as well).

k-distribution/src/main/scripts/lib/pyk/kastManip.py
tothtamas28: These classes are immutable (as far as Python makes such a thing possible), so copying their instances without any modification is not necessary. 

But perphaps `copy` is not the best name for this function after all. I chose this name as an analogous concept in Scala has that name. I'd call it `with`, but that's a reserved word in Python. @ehildenb, what do you think about `let`, e.g. `t2 = t1.let(token='T2')`?

ehildenb: `let` sounds fine, or `.updatedWith(...)` is perhaps a bit more descriptive?

k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -1,216 +1,360 @@
...
+def isKSequence(k):
+    return type(k) is dict and k['node'] == 'KSequence'
+
+
+def KRewrite(lhs, rhs):
+    return {'node': 'KRewrite', 'lhs': lhs, 'rhs': rhs}
+
+
+def isKRewrite(k):
+    return type(k) is dict and k['node'] == 'KRewrite'
```
ehildenb: Everything to this point is `KInner`, and everything after it is `KOuter`.

It may be good to pull out a separate `kast/kinner.py`, for example.

ehildenb: Let's get `KInner` fleshed out in this PR, and we can do `KOuter` in a follow-up PR.

Unless you already have work for both, and then we can do it all in 1 shot.

But just in terms of being able to merge incremental progress, so that I can make sure the submodule updates go through, may make sense to merge this with just `KInner` first.

k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -1,255 +1,1065 @@
...
+        *,
+        sort: Optional[KSort] = None,
+        items: Optional[Iterable[KProductionItem]] = None,
+        klabel: Optional[str] = None,
+        att: Optional[KAtt] = None,
+    ) -> 'KProduction':
+        sort = sort or self.sort
+        items = items if items is not None else self.items
+        klabel = klabel if klabel is not None else self.klabel
+        att = att or self.att
```
ehildenb: Why do some use the `sort = sort or self.sort`, and other the `sort = sort if sort is not None else self.sort` technique?

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -464,43 +442,26 @@ def labelsToDots(kast, labels):
...
+
+    if type(kast) is KFlatModule:
+        sentences = (sentence.map_att(f) for sentence in kast.sentences)
+        return kast.copy(sentences=sentences)  # type: ignore
+
+    if type(kast) is KDefinition:
+        modules = (module.map_att(f) for module in kast.modules)
+        return kast.copy(modules=modules)  # type: ignore
+
+    return kast
```
ehildenb: Wow, how much nicer this is :)

k-distribution/src/main/scripts/lib/pyk/kastManip.py
```diff
@@ -529,24 +490,19 @@ def minimizeRule(rule, keepVars=[]):
...
 
-    if ruleRequires is not None:
-        ruleRequires = buildAssoc(KToken('true', 'Bool'), '_andBool_', dedupe(flattenLabel('_andBool_', ruleRequires)))
-        ruleRequires = simplifyBool(ruleRequires)
-
-    if ruleEnsures is not None:
-        ruleEnsures = buildAssoc(KToken('true', 'Bool'), '_andBool_', dedupe(flattenLabel('_andBool_', ruleEnsures)))
-        ruleEnsures = simplifyBool(ruleEnsures)
+    ruleRequires = buildAssoc(KToken('true', 'Bool'), '_andBool_', dedupe(flattenLabel('_andBool_', ruleRequires)))
+    ruleRequires = simplifyBool(ruleRequires)
```
ehildenb: Love it. So much more readable if we don't have to check for `None`s everywhere.

k-distribution/src/main/scripts/lib/pyk/kast.py
tothtamas28: For `sort: Optional[KSort]` the only falsy value is `None`, but for `sort: Optional[str]` the falsy values are `''` and `None`. I still need to add precondition checks to make sure that `sort: str` (and all similar `str` fields) are nonempty, and then the expressions can be uniformly simplified to `sort or self.sort`.

Jenkinsfile
```diff
@@ -126,7 +126,10 @@ pipeline {
     }
     stage('Build and Package on Ubuntu Focal') {
       when {
-        branch 'release'
+        anyOf {
+          branch 'release'
+          changeRequest()
+        }
```
```suggestion
        branch 'release'
```

Jenkinsfile
```diff
@@ -247,7 +250,10 @@ pipeline {
     }
     stage('Build and Package on Arch Linux') {
       when {
-        branch 'release'
+        anyOf {
+          branch 'release'
+          changeRequest()
+        }
```
```suggestion
        branch 'release'
```

ehildenb: I'll commit these changes together once I finish the rest of the review.

ehildenb (approved):

PR #2454, opened 2022-02-24
Fix for `--version` when `.git` is missing

### LINKED ISSUE #2440: `.deb` packaging doesn't pick up version string
The `.deb` packaging process should pick up the version string written in the PACKAGE file; #2439 indicates that this isn't happening for some reason.

Seems to be happening for brew packaging as well.

Baltoli:
This makes sure that the version string script outputs an empty string if git isn't installed, or there is no git directory alongside the source code. Doing so means that the package version file gets used instead when building release packages (`dpkg` or `brew`).

### DIFF
--- a/scripts/get-version-string.sh
+++ b/scripts/get-version-string.sh
@@ -1,6 +1,17 @@
 #!/bin/bash
 
 SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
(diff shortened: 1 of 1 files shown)


radumereuta:
What is the output of `-v` in this case?

Baltoli:
The format isn't new - this PR just restores what used to happen.

```console
$ kompile --version
K version:    5.2.0
Build date:   Thu Feb 24 08:59:45 CST 2022
```

radumereuta (approved):

PR #2458, opened 2022-02-28
Upstream kcfg.py

tothtamas28:


ehildenb:
Hmmmm, we probably should have a testing strategy for this class, or at least write some unit tests for it.

ehildenb (approved):

PR #2459, opened 2022-02-28
Tests for `llvm-krun`

Baltoli:
We currently don't have any tests for `llvm-krun`; this means that changing the behaviour of that script is liable to break things.

I have a pending PR (https://github.com/runtimeverification/llvm-backend/pull/491) that changes how input configurations are loaded. To make sure that this and future changes are correct, this PR adds a test that exercises the important behaviours of `llvm-krun`.


k-distribution/tests/regression-new/llvm-krun/Makefile
```diff
@@ -0,0 +1,34 @@
+DEF=test
+EXT=
+TESTDIR=.
+KOMPILE_BACKEND=llvm
+KOMPILE_FLAGS+=--gen-glr-bison-parser
+
+CHECK=| diff - test.out
+LLVM_KRUN_FLAGS=-p -d ./test-kompiled
```
Baltoli: This could of course all be abstracted (cf. the existing `krun` tests), but I don't think it's worth doing for the sake of this single test case.

radumereuta:
lgtm but I'm going to let Dwight give the final approval.

k-distribution/tests/regression-new/llvm-krun/Makefile
```diff
@@ -0,0 +1,34 @@
+DEF=test
+EXT=
```
dwightguth: I'm not entirely sure why these tests are passing given that you didn't specify an extension. I'm not entirely sure you're actually running the test here, because ktest.mak chooses which files to run by looking for files with a certain extension. Please address.

dwightguth (changes requested):

k-distribution/tests/regression-new/llvm-krun/Makefile
Baltoli: Sorted

dwightguth (commented):
I still don't understand how it's supposed to be running these tests. You've created a make target named 1.test, but why would it be running that target if there is no file named 1.test in the directory? Have you run make manually locally to make sure it ran the tests?

Baltoli:
I don't think it needs the extension, does it? This test directory isn't using the `krun` test structure, so it only needs `DEF` to be set for the `kompile` target to work.

Then, the top-level ktest group runs the default target in the `llvm-krun` subdirectory:
```console
$ pwd
/home/bruce/code/k/k-distribution/tests/regression-new
$ make SUBDIRS=llvm-krun       
make -C llvm-krun
make[1]: Entering directory '/home/bruce/code/k/k-distribution/tests/regression-new/llvm-krun'
Makefile:32: warning: overriding recipe for target 'clean'
../../../include/kframework/ktest.mak:177: warning: ignoring old recipe for target 'clean'
/home/bruce/code/k/k-distribution/bin/kompile --gen-glr-bison-parser --no-exc-wrap --backend llvm  test.k -d .
./test-kompiled/parser_PGM test.in > test.kore
/home/bruce/code/k/k-distribution/bin/llvm-krun -p -d ./test-kompiled \
  -c PGM test.kore Foo korefile | diff - test.out
/home/bruce/code/k/k-distribution/bin/llvm-krun -p -d ./test-kompiled \
  -c PGM "$(cat test.kore)" Foo kore | diff - test.out
/home/bruce/code/k/k-distribution/bin/llvm-krun -p -d ./test-kompiled \
  -c PGM test.in Foo prettyfile | diff - test.out
/home/bruce/code/k/k-distribution/bin/llvm-krun -p -d ./test-kompiled \
  -c PGM "$(cat test.in)" Foo pretty | diff - test.out
make[1]: Leaving directory '/home/bruce/code/k/k-distribution/tests/regression-new/llvm-krun'
```

You can also see [the tests being run in the CI](https://office.runtimeverification.com/jenkins/blue/rest/organizations/jenkins/pipelines/K/branches/PR-2459/runs/5/nodes/52/log/?start=0#:~:text=%5Bexec%5D%20/home/jenkins%2Dslave/workspace/K_PR%2D2459%402/k%2Ddistribution/bin/llvm%2Dkrun%20%2Dp%20%2Dd%20./test%2Dkompiled%20%5C%0A%20%20%20%20%20%5Bexec%5D%20%20%20%2Dc%20PGM%20test.in%20Foo%20prettyfile%20%7C%20diff%20%2D%20test.out).[^1]

[^1]: Chrome highlight link so won't work in other browsers, but should drop you to the correct line in the CI logs automatically. You can also grep for `-c PGM` (matches 7-10).

dwightguth (approved):

PR #2467, opened 2022-03-05
Minor pyk fixes and tests

ehildenb:
Fixes related to making the newest version of pyk work with summarizer.

~**TODO**: Write tests.~


k-distribution/src/main/scripts/lib/pyk/kast.py
```diff
@@ -114,9 +114,13 @@ def let_att(self: W, att: KAtt) -> W:
         ...
 
     def map_att(self: W, f: Callable[[KAtt], KAtt]) -> W:
+        if self.att is None:
+            return self
         return self.let_att(att=f(self.att))
 
     def update_atts(self: W, atts: Mapping[str, Any]) -> W:
+        if self.att is None:
+            return self.let_att(att=KAtt.of(**atts))
```
tothtamas28: For a `WithAtt` instance `att` should never be `None` . If there are such failures, we should root-cause them.

k-distribution/tests/pyk/Makefile
```diff
@@ -33,11 +33,17 @@ kompile: $(llvm_imp_kompiled) $(haskell_imp_kompiled)
...
+PYTHON      := python3
+PIP         := pip
+PIP_VERSION := 22.0.4
 
 $(VENV):
-	virtualenv -p python3.8 venv && $(ACTIVATE) && pip install -e $(K_LIB)/pyk
+	   $(PYTHON) -m venv venv                 \
+	&& $(ACTIVATE)                            \
+	&& $(PIP) install --upgrade $(PIP)        \
+	&& $(PIP) install --editable $(K_LIB)/pyk
```
ehildenb: To get this working, I had to use my global pip to do `pip install --upgrade pip`, which made a user-local install of pip I think.

Then I had to remove my global install of pip, and remove my global install of virtualenv. Then do `pip install virtualenv` to make a user install. Then this change make it work on my system (because now I have pip22.0.4).

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -641,7 +640,8 @@ def buildRule(ruleId, initConstrainedTerm, finalConstrainedTerm, claim=False, pr
     ruleBody = pushDownRewrites(KRewrite(initConfig, finalConfig))
     ruleRequires = simplifyBool(unsafeMlPredToBool(initConstraint))
     ruleEnsures = simplifyBool(unsafeMlPredToBool(finalConstraint))
-    ruleAtt = None if claim or priority is None else KAtt(atts={'priority': str(priority)})
+    attDict = {} if claim or priority is None else {'priority': str(priority)}
+    ruleAtt = KAtt(atts=attDict)
```
ehildenb, line 644: This was where we were getting `None` values for `rule.att`. I've added a test to called `BuildRuleTest` which exercises this.

k-distribution/tests/pyk/Makefile
tothtamas28: So this fix didn't work: https://github.com/runtimeverification/k/pull/2467/commits/1fa69386abdf734414437c011817ff569110750d?

ehildenb: I've backed out these changes.

tothtamas28 (approved):

PR #2470, opened 2022-03-08
Refactor KCFG

tothtamas28:


ehildenb:
Updating branch to test new CI.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -1,59 +1,100 @@
...
+            return {'id': self.id, 'term': self.term.to_dict()}
+
+    class EdgeLike(ABC):
+        source: 'KCFG.Node'
+        target: 'KCFG.Node'
+
+    @dataclass(frozen=True)
+    class Edge(EdgeLike):
+        source: 'KCFG.Node'
+        target: 'KCFG.Node'
```
ehildenb: Why are we repeating these fields from the superclass `EdgeLike`? Is this some Python convention I'm not aware of, or actually a requirement?

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -65,407 +106,335 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            if node_key in nodes:
+                raise ValueError(f'Multiple declarations of node: {node_key}')
+            nodes[node_key] = node.id
+
+        for edge_dict in dct.get('edges') or []:
+            source_id = resolve(edge_dict['source'])
+            target_id = resolve(edge_dict['target'])
+            condition = KInner.from_dict(edge_dict['condition'])
+            depth = edge_dict['depth']
+            cfg.create_edge(source_id, target_id, condition, depth)
```
ehildenb: It seems that we should be able to do something that reads like this instead (not sure it's possible):

```suggestion
        for edge_dict in dct.get('edges') or []:
            cfg.create_edge(Edge.from_dict(edge_dict))
```

The arguments don't quite line up, but this seems cleaner.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -1,59 +1,100 @@
...
+
+    @dataclass(frozen=True)
+    class Cover(EdgeLike):
+        source: 'KCFG.Node'
+        target: 'KCFG.Node'
+
+        @property
+        def subst(self) -> Optional[Dict[str, KInner]]:
+            subst = matchWithConstraint(self.target.term, self.source.term)
+            return minimizeSubst(subst) if subst is not None else None
```
ehildenb: I wonder if we should be returning the unminimized substitution, instead of the minimized one. The unminimized one has strictly more information (namely, the complete list of variables that show up in the target term as keys).

The information can be reconstructed though, by just getting the free variables of the target term. Whichever way you think is better should be OK.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
tothtamas28: Good point, in general we could omit these type hints in the subclasses. However, in this case they are necessary for the `@dataclass` decorator to work properly.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
tothtamas28: We could do `cfg._create_edge_from_dict(node_mapping, edge_dict)` where `resolve` (currently a nested function) is a method of `node_mapping` (function `resolve` is necessary in the current implementation as node ids in the dict are not required to be valid hashes.)

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
```diff
@@ -0,0 +1,105 @@
...
+        constraints = tuple(flattenLabel('#And', constraint))
+        object.__setattr__(self, 'config', config)
+        object.__setattr__(self, 'constraints', constraints)
+
+    def __iter__(self):
+        return chain([self.config], self.constraints)
+
+    @cached_property
+    def term(self) -> KInner:
+        return mlAnd(self)
```
ehildenb: `term` usually refers to something structural without constraints.

Maybe this should be `cterm(...)` or `pattern(...)`

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
```diff
@@ -0,0 +1,105 @@
...
+        subst = Subst(subst)
+
+        antecedents = set(flattenLabel('_andBool_', simplifyBool(unsafeMlPredToBool(mlAnd(self.constraints)))))
+        consequents = {
+            conjunct
+            for conjunct in flattenLabel('_andBool_', simplifyBool(unsafeMlPredToBool(subst(mlAnd(pattern.constraints)))))
+            if conjunct not in antecedents
+        }
+
+        return subst, antecedents, consequents
```
ehildenb: This seems backward to me. I would do `pattern.match(term)` instead of `term.match(pattern)`. I guess I'm not sure actually though. I would normally say that "`pattern` _matches_ `term`", or "`term` _is matched by_ `pattern`".

I guess it probably doesn't matter too much.

But also, we should not be calling `simplifyBool` or `unsafeMlPredToBool` here I don't think.... it seems to assume too much about the structure. I think this same implementation can happen without calling those functions, the only thing of note that `simplifyBool` is probably doing is (i) removing any instances of `#Top`, and (ii) deduplicating clauses. Both of which should be maintained as invariants of the `CTerm` class anyway.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -28,13 +31,48 @@
...
+
+    def __call__(self, term: KInner) -> KInner:
+        return self.apply(term)
+
+    def compose(self, other: 'Subst') -> 'Subst':
+        return Subst(compose_substs(self, other))
+
+    def apply(self, term: KInner) -> KInner:
+        return substitute(term, self)
+
```
ehildenb: Is this a good place for this? I guess it should be with the definition of `KInner` I suppose? Should be fine for now.

Or perhaps in its own file.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -212,28 +266,59 @@ def addOccurance(k):
...
+    counter: Counter = Counter()
+
+    def count(term: KInner) -> None:
+        if type(term) is KVariable:
+            counter[term.name] += 1
 
-    collectBottomUp(kast, _getNumOccurances)
-    return numOccurances
+    collectBottomUp(term, count)
+    return counter
```
ehildenb: Please include a test for `count_vars`. `collectBottomUp` has given me weird problems in the past and I'd like to make sure that we are locking in the correct behavior.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -212,28 +266,59 @@ def addOccurance(k):
...
+        if type(term) is KVariable:
+            return Counter(term.name) if rhs else Counter()
+        if type(term) is KRewrite:
+            return recur(term.rhs, rhs=True)
+        if type(term) is KApply:
+            return sum((recur(t, rhs=rhs) for t in term.args), Counter())
+        if type(term) is KSequence:
+            return sum((recur(t, rhs=rhs) for t in term.items), Counter())
+        return Counter()
+    return recur(term)
```
ehildenb: Can't this be determined instead by `extract_rhs` and `extract_lhs` an dthen calling `count_vars` on each of them? I guess that's asymptotically less efficiencient, but probably easier to read.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -1,59 +1,105 @@
...
+    @dataclass(frozen=True)
+    class Cover(EdgeLike):
+        source: 'KCFG.Node'
+        target: 'KCFG.Node'
+
+        @cached_property
+        def subst(self) -> Optional[Subst]:
+            match_res = self.source.term.match(self.target.term)
+            if not match_res:
+                return None
```
ehildenb: Every cover should have a substitution. If `match_res is None`, it sohuld probably be a fatal error I think?

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -65,407 +111,379 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    @property
+    def uncovered(self) -> Set[Node]:
+        return {node for node in self.nodes if not self.is_covered(node.id)}
+
+    @property
+    def frontier(self) -> Set[Node]:
+        return {node for node in self._nodes.values() if self.is_frontier(node.id)}
+
+    @property
```
ehildenb: Does this annotation make it memoized or something?

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -65,407 +111,379 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        self._init.discard(node_id)
+        self._target.discard(node_id)
+        self._stuck.discard(node_id)
+
+    def create_edge(self, source_id: str, target_id: str, condition: KInner = TRUE, depth=1) -> Edge:
+        source = self.node(source_id)
+        target = self.node(target_id)
+
+        if target.id in self._edges.get(source.id, {}):
+            raise ValueError(f'Edge already exists: {source.id} -> {target.id}')
```
ehildenb: This might be too restrictive, but let's leave this in as an error for now and relax it later if needed.

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
```diff
@@ -65,407 +111,379 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            res = (edge for _, targets in self._edges.items() for _, edge in targets.items())
+
+        return {edge for edge in res if not target_id or target_id == edge.target.id}
+
+    def create_cover(self, source_id: str, target_id: str) -> Cover:
+        source = self.node(source_id)
+        target = self.node(target_id)
+
+        if source.id in self._covers:
+            raise ValueError(f'Cover already exists: {source.id} -> {target.id}')
```
ehildenb: Similarly, this may be too restrictive, we may have to relax this later.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -522,14 +600,14 @@ def _removeSourceMap(att):
...
+    """Inline <generatedTop> in a configuration.
 
-    -   Input: Constrained term which contains <generatedTop> and <generatedCounter>.
-    -   Output: Constrained term with those cells removed.
+    -   Input: Constrained term.
+    -   Output: Constrained term with <generatedTop> inlined.
     """
     rule = KApply('<generatedTop>', [KVariable('CONFIG'), KVariable('_')]), KVariable('CONFIG')
-    return rewriteAnywhereWith(rule, constrainedTerm)
+    return rewriteAnywhereWith(rule, term)
```
ehildenb: I don't think this is a very good name for this function, what's the thinking here? I usually think of inlining as "moving the definition of X into the use-sites of X". This almost looks more like `extract_config` to me.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
ehildenb: If we do `pattern.match(term)`, we can also have `pattern.substitute(term)`, and should have the property:

```
match = pattern.match(term)
if match:
    new_term = pattern.substitute(match)
    assertEqual(term, new_term)
```

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
tothtamas28: It makes function `covers` accessible as an attribute, i.e. `cfg.covers`.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
tothtamas28: Renamed.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
tothtamas28: I added this function so that variables can be prefixed with `_` and `?` accordingly based on the number of their occurrences in either side in claims.

The problem I see with a solution based on `extract_lhs` and `extract_rhs` is that it cannot differentiate between certain terms that have to handled differently. Consider e.g. `<cell> V </cell>` and `<cell> V => V +Int 1 </cell>`, where the first should be rewritten as `<cell> _V </cell>` while the second is fine as is.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
tothtamas28: Moved to its own module.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
tothtamas28: Based on these two Google searches over stackoverflow.com, both forms are used (I searched for "string" as that is the most common form of matching):

* ["pattern matches a string"](https://www.google.com/search?q=%22pattern+matches+a+string%22+site%3Astackoverflow.com)
* ["string matches a pattern"](https://www.google.com/search?q=%22string+matches+a+pattern%22+site%3Astackoverflow.com)

Sometimes even in the same sentence: https://stackoverflow.com/a/45632286 :slightly_smiling_face:

I prefer `term.match(pattern)` for two reasons:

* It aligns with the order formulas in an implication (or order of sets of models in a set inclusion), so something like this reads more naturally: `if term.match(pattern): ...`. This is also the direction we draw our abstraction arrows for `KCFG`.
* `def match(pattern)` makes it obvious that the term to be passed as an argument is the pattern.

The property you identified is nice though: `pattern.substitute(pattern.match(term)) == term`. (Right now we have: if `term.match(pattern) == subst` then `subst.apply(pattern) == term`.)

We can also consider which one of the two functions is more likely to be passed to a higher order function as argument, for example:

* `map(term.match, [pattern1, pattern2, pattern3])` - match a term over multiple patterns, which is analogous to FP pattern matching (e.g. `term match { case pattern1 => ... }` in Scala)
* `map(pattern.match, [term1, term2, term2])` - match a pattern over multiple terms, which is some sort of classification of terms w.r.t. pattern.

tothtamas28: > simplifyBool is probably doing is (i) removing any instances of #Top, and (ii) deduplicating clauses. Both of which should be maintained as invariants of the CTerm class anyway

Good point about maintaining the invariant. Though this function also does rewrites such as:
```k
(true ==K (N >=Int 0)) => (N >=Int 0)
```

W. r. t. `unsafeMlPredToBool`: is there an other way to ensure that no ML connectives leak into `requires`-`ensures`? Or can we assume that the only ML connectives that occur are `#Top` and `#And`?

k-distribution/src/main/scripts/lib/pyk/pyk/kcfg.py
tothtamas28: Fixed.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
tothtamas28: Tests added.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
tothtamas28: I wouldn't call it `pattern`, as that suggest pattern matching as context. I guess `cterm` is fine, though if `cterm: CTerm`, then `cterm.cterm` can be misleading, as `cterm.cterm: KInner`.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
ehildenb: Let's make sure we have tests of `pattern.substitute(term.match(pattern)) == term`.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
tothtamas28: Renamed to `cterm`.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
tothtamas28: Tests added.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
```diff
@@ -0,0 +1,105 @@
...
+            return None
+
+        subst = Subst(subst)
+
+        antecedents = set(flattenLabel('_andBool_', simplifyBool(unsafeMlPredToBool(mlAnd(self.constraints)))))
+        consequents = {
+            conjunct
+            for conjunct in flattenLabel('_andBool_', simplifyBool(unsafeMlPredToBool(subst(mlAnd(pattern.constraints)))))
+            if conjunct not in antecedents
+        }
```
ehildenb: We still should not be bringing the `Bool` theory into this module, it's too specific for these use cases.

A `CTerm` is a _Matching Logic Pattern_ with a specific form, not a _term_ with a set of _boolean constraints_.

So converting to bool just to apply simplifications is a bad idea. We should not have `Bool` leak into this module at all.

I think you can accomplish the same thing without converting it to `Bool` sort.

ehildenb: If later we find it's necessary for some reason, then we can do it. But we really should be _very_ careful about using `unsafeMlPredToBool` where it's not strictly needed, it is actually unsafe.

tothtamas28: Pushed a change, please check.

> We should not have Bool leak into this module at all.

Does this apply to `CTerm.to_rule` as well? It can be moved to `KCFG` if so.

tothtamas28: `CTerm.to_rule` and `CTerm.from_rule` removed for now.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
```diff
@@ -0,0 +1,43 @@
...
+    def match(self, pattern: 'CTerm') -> Optional[Tuple[Subst, List[KInner]]]:
+        subst = match(pattern=pattern.config, term=self.config)
+
+        if subst is None:
+            return None
+
+        assumptions = set(self.constraints)
+        obligations = [constraint for constraint in map(subst, pattern.constraints) if constraint not in assumptions]
+
+        return subst, obligations
```
ehildenb: The obligations here need to be `#Top`, correct? Can we document this?

tothtamas28: This method can be considered a generalization of `kastManip.matchWithConstraint` (though it's not as smart yet, so we can improve that later). If `obligations` is empty, it indicates a syntactic match. Otherwise, it contains the obligations that need to be shown (assuming `self.constraints`) for the implication to hold.

ehildenb: Well, specifically the obligations need to be shown _valid_, not just _satisfiable_. That is, `simplify(obligations)` must be `#Top`, it's not enough that there exists a model of the remaining obligations, there must be no models of the negation of the obligatios (`#Not obligations` must be `#Bottom`, unsatisfiable).

Otherwise we are doing some sort of hybrid between matching on the configuration and unification on the constraints. It seems dangerous to expose this method as it is to me, because it is some weird hybrid of matching and unification. We should eventually support some limited unification, but in that case we should be doing unification of both the configuration and the constraints.

I would rather fail with a fatal error if `len(obligations) > 0` for the moment, because (i) thats good enough for the hacky version of the tool as it is right now because we make sure to maintain syntactic constraint stability externally anyway, and (ii) we can relax this restriction later as we add more semantic information (better to start with limited safe version and expand functionality).

k-distribution/src/main/scripts/lib/pyk/pyk/tests/test_subst.py
```diff
@@ -0,0 +1,84 @@
...
+            (x, {'x': a}, a),
+            (f(x), {'x': f(x)}, f(f(x))),
+            (f(a, g(x, a)), {'x': b}, f(a, g(b, a))),
+            (f(g(h(x, y, z))), {'x': a, 'y': b, 'z': c}, f(g(h(a, b, c))))
+        )
+
+        for i, [pattern, subst, expected] in enumerate(test_data):
+            with self.subTest(i=i):
+                # When
+                actual = Subst(subst)(pattern)
```
ehildenb: Does this read better or does `pattern.substitute(Subst(subst))` read better? I'm not sure....

I do like the idea of "substitution application" being a thing, which this seems to express.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
ehildenb: So, for example, if `3 >Int X` was in `obligations`, then it is not valid because `X ==Int 5` doesn't work, but it is satisfiable because `X ==Int 2` does work. This means matching should not succeed.

tothtamas28: > Well, specifically the obligations need to be shown valid

Definitely. What the `kit` tool does for example is construct a claim for the validity of `(/\ self.constraints) -> (/\ obligations)`. This is necessary as otherwise even the simplest nontrivial example (sum-to-n) fails to converge.

> It seems dangerous to expose this method as it is to me, because it is some weird hybrid of matching and unification.

How about I add a docstring elaborating on how `obligations` is supposed to be interpreted? Then each caller can use it according to its needs, i.e. fail if `len(obligations) > 0`, or do something with it otherwise. I think it is a useful function is this context.

Alternatively, I can add both the syntactic and semantic versions of `match`, with the syntactic one being a special case. I can also remove it, but then I'll have tor reproduce it in `kit` (and will probably end up upstreaming it eventually).

k-distribution/src/main/scripts/lib/pyk/pyk/tests/test_subst.py
tothtamas28: > I do like the idea of "substitution application" being a thing, which this seems to express.

Indeed, `subst(pattern)` is syntactic sugar for `subst.apply(pattern)`.

I think handling a substitution as `KInner -> KInner` makes sense, as semantically it is already `KVariable -> KInner`. Function `apply` is just its homomorphic extension.

k-distribution/src/main/scripts/lib/pyk/pyk/cterm.py
tothtamas28: Changed to return a matching logic formula instead of a list.

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
```diff
@@ -208,32 +161,63 @@ def addOccurance(k):
...
+    def count(term: KInner) -> None:
+        if type(term) is KVariable:
+            counter[term.name] += 1
+
+    collect(count, term)
+    return counter
+
+
+def count_rhs_vars(term: KInner) -> Counter:
+    def recur(term: KInner, *, rhs=False) -> Counter:
```
ehildenb: What does the second argument `*,` mean here?

k-distribution/src/main/scripts/lib/pyk/pyk/subst.py
```diff
@@ -0,0 +1,113 @@
...
+    def __mul__(self, other: 'Subst') -> 'Subst':
+        return self.compose(other)
+
+    def __call__(self, term: KInner) -> KInner:
+        return self.apply(term)
+
+    def compose(self, other: 'Subst') -> 'Subst':
+        from_other = ((k, self(v)) for k, v in other.items())
+        from_self = ((k, v) for k, v in self.items() if k not in other)
+        return Subst(dict(chain(from_other, from_self)))
```
ehildenb, line 38: So in `subst1.compose(subst2)`, we are preferring the bindings from `subst2`? Seems more like an `update` than a `compose` to me.

ehildenb: So in `subst1.compose(subst2)`, we are preferring the bindings from `subst2`? The name `compose` seems ambiguous about which is preferred, but `update` would be pretty clear I think.

ehildenb (approved):

k-distribution/src/main/scripts/lib/pyk/pyk/kastManip.py
tothtamas28: Arguments that follow `*` are [keyword-only arguments](https://peps.python.org/pep-3102/). In this case it means for example that `recur(term, True)` is not syntactically correct, only `recur(term, rhs=True)` is.

k-distribution/src/main/scripts/lib/pyk/pyk/subst.py
tothtamas28: Semantically, `Subst` can be interpreted as a total function over variables:

* if `(x, v) in subst.items()` for some `v` then semantically `subst(x) = v`;
* otherwise, `subst(x) = x`.

Assuming this interpretation, this method aims to implement function composition:

* For all `(x, v) in from_other`, we have `v = subst1(subst2(x))`. Moreover, `(x, v) in from_other` for some `v` iff `x in susbt2`.
* For all `(x, v) in from_self`, we have `v = subst1(x)`. Moreover, `(x, v) in from_self` for some `v` iff `x in subst1` and `x not in subst2`. Thus we have `v = subst1(subst2(x))`.
* We have `(x, v) not in from_other` and `(x, v) not in from_self` for any `v` iff `x not in subst1` and `x not in subst2`. Thus `x = subst1(subst2(x))`.

ehildenb: Nice, good intuition here then :)

PR #2473, opened 2022-03-09
Add test class for pyk.kast.prettyPrintKast

tothtamas28:


k-distribution/src/main/scripts/lib/pyk/pyk/tests/test_pretty_print_kast.py
```diff
@@ -0,0 +1,21 @@
...
+
+    SYMBOL_TABLE: Final[Mapping[str, Callable]] = {}
+
+    def test(self):
+        for i, (kast, expected) in enumerate(self.TEST_DATA):
+            with self.subTest(i=i):
+                actual = prettyPrintKast(kast, self.SYMBOL_TABLE)
+                actual_tokens = actual.split()
+                expected_tokens = expected.split()
+                self.assertListEqual(actual_tokens, expected_tokens)
```
ehildenb, line 21: Why do we split them instead of just comparing directly? An dwhat is `with self.subTest(i=1):` doing?

ehildenb (approved):

k-distribution/src/main/scripts/lib/pyk/pyk/tests/test_pretty_print_kast.py
tothtamas28: For now I wanted to ignore whitespace characters. But later on we can strengthen the tests to check them as well.

tothtamas28: `with self.subTest(i=i)` assigns an index to the given iteration so that if there is a failure the output will show which iteration the failure occurred in:

https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

This looked like a good enough way to do parametric tests without introducing a new dependency.

PR #2476, opened 2022-03-10
Print source lines in error messages

### LINKED ISSUE #2064: [Feature] [kompile] - Improvements to error messages
Description
========
There were several improvements that have been brought up in the slack conversation. This issue is created to discuss two improvements to the kompile error messages. These requests aren't urgent and aim to help improve user experience.

Include filename at the top line of error messages
------------------------------------------------------------------
When compiling files that contain the `requires` directive, there could be a compiler error in the file being required rather than the file that provided as the input file on the command line. In a setting where there are two files each with a module, the main file is supplied as the parameter to the `kompile` command.

Here is one file primary.k
```
requires "secondary.k"

module PRIMARY

  syntax Foo ::= "bar"

endmodule

```
Here is secondary.k
```
module SECONDARY

  syntax Foo ::= | "baz"

endmodule

```

kompiling primary.k resulting in the following error message:
```
$ kompile primary.k 
18.

Was expecting one of:

"non-assoc" ...
    <STRING> ...
    <STRING_REGEX> ...
    "(" ...
    "left" ...
    "right" ...
    "List" ...
    "NeList" ...
    <UPPER_ID> ...
    <LOWER_ID> ...
    <NAT> ...
    <UPPER_ID> ...
    "List" ...
    "NeList" ...
    <LOWER_ID> ...
    )
        Source(/home/erik/k/k-distribution/k-tutorial/1_basic/09_unparsing/secondary.k)
        Location(3,18,3,18)

```

When reading this error, the user sees `ParseException: Encountered " "|" "|"" at line 3, column 18.` at the top. However, the information regarding the filename can only be found at the bottom of the error message.  This partial information forces the user to constantly look at the end of the error message to find the filename. A more helpful error message would contain the file name like so: `ParseException: Encountered " "|" "|"" in secondary.k at line 3, column 18.` By doing so, the user can get more complete error location information. We may want to consider only printing the file name rather than the full file path so that message is easy to read.

Colorize the exception summary
Colorizing exception summary can make it stand out from the rest of the messages.

SchmErik:
At the time of writing, error messages display location information like
so:
```
[Error] Compiler: Found existential variable not supported by concrete backend.
        Source(/home/erik/error/test.k)
        Location(11,39,11,41)
```

Sometimes these messages can be difficult to decipher due to the lack of
information from the actual source code. Add this information in the
error message so that the errors look like this:

```
[Error] Compiler: Found existential variable not supported by concrete backend.
        Source(/home/erik/error/test.k)
        Location(11,39,11,41)
        |    rule <k> a => b ... </k> requires ?_ ==Int _
        |                                      ^~~
```

For longer lines, error messages look like this:

```
        Source(/home/erik/error/test.k)
        Location(15,11,17,13)
           .          v~~~
        15 |    claim _
        16 |          =>
        17 |           .
           .         ~~^
```

If error messages span more than 3 lines, the lines between the first
and last line are replaced with a single line containing `...` to avoid
taking up too much space on the terminal.


SchmErik:
@radumereuta could you look over this PR? I need to add the option flag to enable this but other than that this implements the new error message format. my complaint is that I wish I could make this a little more OO but I don't think the format we chose lends itself very well for OO but I could be wrong...

radumereuta:
Not sure why you say it's not very OO, this looks like I would expect it to.
lgtm so far.

SchmErik:
This PR was based on a fork. I'm going to close this and open another based off of the runtimeverification/k repo

PR #2489, opened 2022-03-17
Allow macros on hooked sorts

### LINKED ISSUE #2488: [Bug] kompile should allow macro productions on hooked sort
Description
-----------
The current implementation of macros on productions does not allow macro productions on hooked sorts.
Code like this:
```
syntax Int ::= "LABEL" [macro]
rule LABEL => 4 // or some other number
```
would result in errors like this:
```
$ kompile test.k
[Error] Compiler: Cannot add new constructors to hooked sort Int
```
    
However, `"LABEL"` in this case is not a constructor but a term that would be translated to `4` during compilation so this error is not applicable to productions that are macros. `kompile` needs to be fixed to allow cases like this.

SchmErik:
The previous implementation of macros on productions forgot to account for
macros on hooked sorts. This led kompile to assume that macro productions on
hooked sorts implied that the productions were constructors. Code like this

```
syntax Int ::= "LABEL" [macro]
rule LABEL => 4 // or some other number
```
would result in errors like this:

```
$ kompile test.k
[Error] Compiler: Cannot add new constructors to hooked sort Int
```

However, `"LABEL"` in this case is not a constructor but a term that
would be translated to `4` during compilation so this error is not
applicable to productions that are macros. This change add changes to
the CheckAtt class to allow macros on hooked sorts.

Fixes: https://github.com/runtimeverification/k/issues/2488


kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -51,7 +51,7 @@ private void check(Production prod) {
             Att sortAtt =  m.sortAttributesFor().getOrElse(prod.sort().head(), () -> Att.empty());
             if (sortAtt.contains(Att.HOOK()) && !sortAtt.get(Att.HOOK()).equals("ARRAY.Array") && !(sortAtt.get(Att.HOOK()).equals("KVAR.KVar") && isSymbolicKast)) {
                 if (!prod.att().contains(Att.FUNCTION()) && !prod.att().contains(Att.BRACKET()) &&
-                    !prod.att().contains("token") && !(prod.klabel().isDefined() && macros.contains(prod.klabel().get()))) {
+                    !prod.att().contains("token") && !prod.att().contains("macro") && !(prod.klabel().isDefined() && macros.contains(prod.klabel().get()))) {
```
Baltoli, line 54: This isn't a change to this specific PR (which looks good to me), but it would be nice in general to have this kind of deeply nested / long chain code factored out into a helper method with a comment explaining the logic. Perhaps it would be worth someone spending some time on the API design of the frontend to make this kind of code easier to write

Baltoli (approved):
Looks good to me!

PR #2502, opened 2022-03-22
Fix for multiplicity star cells generated by a function

theo25:
This PR addresses https://github.com/runtimeverification/llvm-backend/issues/488.

When a cell with multiplicity star that belongs to a map cell collection is generated by a function instead of using the singleton cell constructor and then is passed to a function expecting a collection, the generated KORE would contain an injection from the cell type to the cell collection type instead of a call to the singleton cell collection constructor. The PR fixes this by making sure that the call to the singleton cell collection constructor is generated in this case.

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
+++ b/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
@@ -46,6 +46,7 @@ public class AddSortInjections {
 
     private final Module mod;
     private final Map<KLabel, KLabel> collectionFor;
+    private final ConfigurationInfoFromModule configurationInfo;
 
     private int freshSortParamCounter = 0;
     private Set<String> sortParams = new HashSet<>();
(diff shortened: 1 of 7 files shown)


kernel/src/main/java/org/kframework/compile/AddSortInjections.java
```diff
@@ -117,6 +117,11 @@ public Boolean merge(Boolean a, Boolean b) {
         return internalAddSortInjections(body, sort);
     }
 
+    private Sort getCellSort(KLabel klabel) {
+        ConfigurationInfoFromModule configInfo = new ConfigurationInfoFromModule(mod);
```
dwightguth: can we actually add this as a field to the class rather than constructing it on the fly like this? it would be much cheaper if we are calling this function many times, which we probably are.

dwightguth (commented):
functionally it looks okay, but please see my comment about performance.

kernel/src/main/java/org/kframework/compile/AddSortInjections.java
theo25: Fixed

dwightguth (approved):

PR #2513, opened 2022-03-25
Reorganize pyk distribution

### LINKED ISSUE #2508: Reorganize pyk distribution
* Move `k-distribution/src/main/scripts/lib/pyk` to `pyk`
* Install `pyk` in distributed Docker images
* Add `pyk/README.md` with installation instructions

tothtamas28:
Fixes #2508, #2515


ehildenb:
Test

ehildenb:
Noticing that GitHub action builds are acting weird, so I'm trying to see if it's because a codeowner needs to participate or not? But it seems that alone si not fixing it...

pyk/Makefile
```diff
@@ -0,0 +1,57 @@
...
+	&& $(ACTIVATE_DEV)                     \
+	&& pip install -r requirements/dev.txt
+
+
+VENV_PROD     := venv-prod/pyvenv.cfg
+ACTIVATE_PROD := . venv-prod/bin/activate
+
+$(VENV_PROD):
+	   virtualenv -p python3.8 venv-prod \
+	&& $(ACTIVATE_PROD)                  \
```
ehildenb: Let's add .PHONY targets `venv-prod` and `venv-dev` so that we can easily set these up. Also do `@echo $(ACTIVATE_PROD)` at the end of the recipe, to print out what the user needs to do to directly activate the virtual environment themselves after doing `make venv-prod`. See https://github.com/runtimeverification/erc20-verification/blob/master/Makefile#L154

ehildenb: Also, make `venv-dev` a variable, as `VENV_DEV_DIR := venv-dev`, and re-use that variable directly instead. Makes it easier to change that directory in the future.

tothtamas28: Done: 536a951188d26bf9c24f237500464740306287a1

ehildenb (approved):

PR #2517, opened 2022-03-25
Print source lines in error messages

### LINKED ISSUE #2064: [Feature] [kompile] - Improvements to error messages

SchmErik:
At the time of writing, error messages display location information like
so:
```
[Error] Compiler: Found existential variable not supported by concrete backend.
        Source(/home/erik/error/test.k)
        Location(11,39,11,41)
```

Sometimes these messages can be difficult to decipher due to the lack of
information from the actual source code. Add this information in the
error message so that the errors look like this:

```
[Error] Compiler: Found existential variable not supported by concrete backend.
        Source(/home/erik/error/test.k)
        Location(11,39,11,41)
        |    rule <k> a => b ... </k> requires ?_ ==Int _
        |                                      ^~~
```

For longer lines, error messages look like this:

```
        Source(/home/erik/error/test.k)
        Location(15,11,17,13)
           .          v~~~
        15 |    claim _
        16 |          =>
        17 |           .
           .         ~~^
```

If error messages span more than 3 lines, the lines between the first
and last line are replaced with a single line containing `...` to avoid
taking up too much space on the terminal.

### DIFF
--- a/k-distribution/tests/regression-new/checks/nestedFunctionContextInFun.k.out
+++ b/k-distribution/tests/regression-new/checks/nestedFunctionContextInFun.k.out
@@ -1,7 +1,11 @@
 [Error] Inner Parser: Parse error: unexpected token '[' following token '('.
 	Source(nestedFunctionContextInFun.k)
 	Location(8,16,8,17)
+	8 |	rule 0 => #fun([[0 => 1]] <bar> 0 </bar>)(0)
+	  .	               ^
 [Error] Inner Parser: Parse error: unexpected token '[' following token '('.
 	Source(nestedFunctionContextInFun.k)
 	Location(10,16,10,17)
+	10 |	rule 0 => #fun([[[[0 => 1]] <bar> 0 </bar>]] <bar> 0 </bar>)(0)
+	   .	               ^
 [Error] Compiler: Had 2 parsing errors.
--- a/k-distribution/tests/regression-new/cell-bag-sort-llvm/test.k.out
+++ b/k-distribution/tests/regression-new/cell-bag-sort-llvm/test.k.out
@@ -1,4 +1,9 @@
 [Error] Compiler: Cell bags are only supported on the Java backend. If you want this feature, comment on https://github.com/runtimeverification/k/issues/1419 . As a workaround, you can add the attribute type="Set" and add a unique identifier to each element in the set.
 	Source(test.k)
 	Location(18,17,32,21)
+	   .	                v~~~~~~~~~~~~~~~~~
+	18 |	  configuration <T color="yellow">
+	   |		...
+	32 |	                </T>
+	   .	                ~~~^
 [Error] Compiler: Had 1 structural errors.
--- a/k-distribution/tests/regression-new/checkClaimError/rule-spec.k.out
+++ b/k-distribution/tests/regression-new/checkClaimError/rule-spec.k.out
@@ -1,4 +1,6 @@
 [Error] Compiler: Only claims and simplification rules are allowed in proof modules.
 	Source(rule-spec.k)
 	Location(6,10,6,43)
+	6 |	    rule <k> doIt(foo) => doIt(0) ... </k>
+	  .	         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 [Error] Compiler: Had 1 structural errors.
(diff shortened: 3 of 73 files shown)


k-distribution/tests/regression-new/checks/nestedFunctionContextInFun.k.out
```diff
@@ -1,7 +1,11 @@
 [Error] Inner Parser: Parse error: unexpected token '[' following token '('.
 	Source(nestedFunctionContextInFun.k)
 	Location(8,16,8,17)
+	8 |	rule 0 => #fun([[0 => 1]] <bar> 0 </bar>)(0)
+	  .	               ^~
```
radumereuta: This doesn't look right.
The error message says only one character is of blame here, but the underline shows two.
The location information for K is an interval that is inclusive on the left, and exclusive on the right: `[0,1)`

radumereuta (changes requested):
I think there is a _off by one_ error here. The ending location in K points to the next character after a token, so that when you select a piece of code, the location matches with what your editor is telling you.

The correctness of this is debatable, but I think it makes it easier to find the location of the problematic code in your editor.

k-distribution/tests/regression-new/checks/nestedFunctionContextInFun.k.out
SchmErik: hey I looked at this closer and i think it's correct in that the error message underlines the location. If there's only one incorrect character, the location should be (8,16, 8, 16) but somehow, the lexer is picking up two characters.

See the following diff which is a part of this PR.

```
diff --git a/k-distribution/tests/regression-new/checks/tokenCheck.k.out b/k-distribution/tests/regression-new/checks/tokenCheck.k.out
index eb2f33f4b1..b3f85ac44c 100644
--- a/k-distribution/tests/regression-new/checks/tokenCheck.k.out
+++ b/k-distribution/tests/regression-new/checks/tokenCheck.k.out
@@ -1,7 +1,11 @@
 [Error] Compiler: Sort X was declared as a token. Productions of this sort can only contain [function] or [token] labels.
        Source(tokenCheck.k)
        Location(10,16,10,21)
+       10 |      syntax X ::= fail()
+          .                   ^~~~~~
 [Error] Compiler: Sort Y was declared as a token. Productions of this sort can only contain [function] or [token] labels.
        Source(tokenCheck.k)
        Location(13,16,13,16)
+       13 |      syntax Y ::= Z
+          .                   ^

```

radumereuta: Both of these show terms parsed by the outer parser.
It seems we may have some sort of inconsistency.

radumereuta (approved):
This is looking way better than I originally imagined.

PR #2529, opened 2022-03-30
Perform fixes in pyk for ksummarize

tothtamas28:


pyk/src/pyk/cterm.py
```diff
@@ -31,15 +31,28 @@ def cterm(self) -> KInner:
...
         subst = match(pattern=pattern.config, term=self.config)
 
         if subst is None:
             return None
 
-        obligation = self._ml_impl(self.constraints, map(subst, pattern.constraints))
+        condition = self._ml_impl(self.constraints, map(subst, pattern.constraints))
 
-        return subst, obligation
+        return subst, condition
```
ehildenb: I would call this `match_with_constraint` instead, to match the notion of `constraint` we're using elsewhere. Specifically, it's a condition which can be any number of clauses, but it's _intended_ to be attached to exactly 1 configuration clause.

pyk/src/pyk/kcfg.py
```diff
@@ -58,26 +58,29 @@ def to_dict(self) -> Dict[str, Any]:
 
         def to_rule(self, claim=False, priority=50) -> KRuleLike:
             sentence_id = f'BASIC-BLOCK-{self.source.id}-TO-{self.target.id}'
-            init_term = mlAnd([self.source.term, self.condition])
-            final_term = self.target.term
-            return buildRule(sentence_id, init_term, final_term, claim=claim, priority=priority)
+            init_term = mlAnd([self.source.cterm.term, self.condition])
+            final_term = self.target.cterm.term
+            rule, _ = buildRule(sentence_id, init_term, final_term, claim=claim, priority=priority)
+            return rule
```
ehildenb, line 64: `self.source.cterm.term` returns the term _and_ its constraints? I just want to confirm that we're calling `buildRule` with the appropriate arguments.

pyk/src/pyk/kcfg.py
```diff
@@ -110,40 +113,40 @@ def __exit__(self, exc_type, exc_value, traceback):
         return False
 
     @property
-    def nodes(self) -> Set[Node]:
-        return set(self._nodes.values())
+    def nodes(self) -> List[Node]:
+        return list(self._nodes.values())
```
ehildenb, line 117: Why do we return as `list` now instead of `set`? Is it so that we can do things like `cfg.nodes()[0]`?

pyk/src/pyk/utils.py
```diff
@@ -78,6 +78,19 @@ def find_common_items(l1: Iterable[T], l2: Iterable[T]) -> Tuple[List[T], List[T
     return (common, newL1, newL2)
 
 
+def intersperse(iterable: Iterable[T], delimiter: T) -> Iterator[T]:
```
ehildenb, line 81: Is this used downstream?

ehildenb (approved):

pyk/src/pyk/kcfg.py
tothtamas28: Primarily, but also because lists are cheaper to construct than sets.

pyk/src/pyk/kcfg.py
tothtamas28: Yes. `cterm.term` is the same as `mlAnd([cterm.config] + cterm.constraints)`.

pyk/src/pyk/utils.py
tothtamas28: Yes: https://github.com/runtimeverification/evm-semantics/blob/kevm-pyk-dist/kevm_pyk/src/kevm_pyk/solc_to_k.py#L253

PR #2538, opened 2022-04-13
Add kompile feature to pyk.ktool

### LINKED ISSUE #2536: Add kompile feature to pyk.ktool
This can be done on two levels:

1. A function `kompile` that does some simple validation on its parameters, then invokes `kompile` as a subprocess.
2. A class, perhaps `Kompiler`, that fixes some `kompile` parameters, and wraps a `KDefinition` to transparently manage its recompilation on modification. It can expose methods like `add_sentence` that mutate the wrapped `KDefinition`.

tothtamas28:
Related: #2536


pyk/src/pyk/ktool/kprove.py
```diff
@@ -3,41 +3,82 @@
...
+    return args
+
+
+def _kprovex(spec_file: str, *args: str) -> CompletedProcess:
+    run_args = ['kprovex', spec_file] + list(args)
+    notif(' '.join(run_args))
+    return run(run_args, capture_output=True)
 
 
 class KProve(KPrint):
```
ehildenb, line 84: We should probably have class `KProve` make use of function `kprovex` I think?

pyk/src/pyk/ktool/kprint.py
```diff
@@ -0,0 +1,23 @@
...
+        self.symbolTable = buildSymbolTable(self.definition, opinionated=True)
+        self.definitionHash = hash_str(self.definition)
+
+    def prettyPrint(self, kast, debug=False):
+        """Given a KAST term, pretty-print it using the current definition.
+
+        -   Input: KAST term in JSON.
+        -   Output: Best-effort pretty-printed representation of the KAST term.
+        """
+        return prettyPrintKast(kast, self.symbolTable, debug=debug)
```
ehildenb, line 56: Maybe we should move the definition of `prettyPrintKast` into this file?

ehildenb: And encourage people to go thorugh the class interface.

pyk/src/pyk/integration_tests/test_emit_json_spec.py
```diff
@@ -1,35 +1,38 @@
...
+    SPEC_JSON_FILE = 'definitions/imp-verification/haskell/looping-spec.json'
 
     def setUp(self):
         super().setUp()
 
-        with open(self.SPEC_FILE, 'r') as spec_file:
+        spec_file = Path(self.SPEC_FILE)
+        directory = Path(self.KOMPILE_OUTPUT_DIR)
+        emit_json_spec = Path(self.SPEC_JSON_FILE)
+        kprovex(spec_file, directory=directory, emit_json_spec=emit_json_spec, dry_run=True)
```
ehildenb, line 29: Can this use `self.kprove` instead? Or should it be calling the `kprovex` function directly? If `KProve` is made to call `kprovex` function, I guess this should go through the class interface.

pyk/src/pyk/integration_tests/kompiled_test.py
```diff
@@ -0,0 +1,38 @@
...
+        main_file = Path(self.KOMPILE_MAIN_FILE)
+        self.assertTrue(main_file.is_file())
+        self.assertEqual(main_file.suffix, '.k')
+
+        output_dir = Path(self.KOMPILE_OUTPUT_DIR) if self.KOMPILE_OUTPUT_DIR else None
+
+        include_dirs = [Path(include_dir) for include_dir in self.KOMPILE_INCLUDE_DIRS]
+        self.assertTrue(all(include_dir.is_dir() for include_dir in include_dirs))
+
+        self.kompiled_dir = kompile(
```
ehildenb, line 27: Do you think it's beneficial to just make a `Kompile` class now, so that downstream we're already creating `Kompile` objects and calling `_.kompile()` on them, and won't have to make that change later?

I guess down the road, we will have things like:

```
kevm = Kompile(...)
....
kevm.init('before-lemmas', 'MODULE-NAME', 'llvm)                  // Actually calls kompile
...
kevm.add_lemmas_to_module(lemma_list, 'INTERNAL-MODULE')
...
kevm.init('after-lemmas', 'ANOTHER-MODULE', 'haskell')
```

Maybe this isn't a good pattern to use for some reason?

What do you think?

pyk/src/pyk/ktool/kprint.py
tothtamas28: I wouldn't do that. Having the ability to quickly print a simple `KAst` based on a custom symbol table can be very useful. But `KPrint` requires a kompiled definition, so that makes this very cumbersome.

pyk/src/pyk/integration_tests/kompiled_test.py
tothtamas28: I agree that we need a smart kompiler that only invokes the `kompile` command when necessary.

But maybe we don't actually need smart methods like `add_lemmas_to_module`. After all, we're going to read the main definition from disk, so perhaps we only need the following features.

* Collect transitive dependencies for a `KDefinition`.
* Compare timestamps like `make`.

This method of course doesn't have much domain-specific knowledge built in, so for example adding a simplification rule to a dependency triggers build. But projects should be organized in a way that these artifacts do not mix anyways, so I do not see this as a big issue.

pyk/src/pyk/ktool/kprove.py
tothtamas28: That's tricky, because some clients rely on updating `KProve.prover`, e.g.: https://github.com/runtimeverification/erc20-verification/blob/master/ksummarize/__main__.py#L197.

pyk/src/pyk/ktool/kprint.py
ehildenb: Well, I wasn't thinking we would move the definition of prettyPrintKast into the class itself, just move the definition of that function here. But not sure if it's better. It just seems like a logical organization of the code, but no strong opinion on it.

ehildenb (approved):

pyk/src/pyk/ktool/kprint.py
tothtamas28: Oh yeah, that makes sense, sorry for the confusion.

pyk/src/pyk/integration_tests/test_emit_json_spec.py
tothtamas28: Tried with `self.kprove.prove`, but it passes spurious flags to `kprove` (e.g. `--spec-module`).

PR #2542, opened 2022-04-14
Tweaked join/spawn semantics for Tutorial Lesson 17

### LINKED ISSUE #2533: Fork/Join Example for Tutorial Lesson 1_basic/17_cell_multiplicity
It would be useful to have an example of a program to pass to ``krun`` based on ``module LESSON-17-B`` in the following tutorial lesson: [17_cell_multiplicity](https://github.com/runtimeverification/k/tree/master/k-distribution/k-tutorial/1_basic/17_cell_multiplicity). The proposed example would be a program with roughly this behaviour:

1. Program starts with a single thread i.e. a single ``<thread>``  - thread ``0``.
2. Thread ``0`` spawns another thread - thread ``1`` and delegates the computation of some expression to thread ``1`` e.g. ``24 + 6``
3. Thread ``1`` computes the expression and returns the value e.g. ``30``
4. Thread ``0`` joins with thread ``1``, so that the value in thread ``0`` is now ``30``. At this point, thread ``1`` is destroyed. So there is just one thread in the final state.

---

The reason I ask for such an example is:

1. I get the feeling that this was the intent of the semantics - to be able to write simple programs like the above.
2. I am unable to see how such a program could be written in the example language. Maybe the example language should be tweaked a little? Or maybe I just haven't figured out how to write such an example myself.

Below is what I've tried

``krun -cPGM='join spawn { x = 24; y = 6; return x + y;  };'``

However, no rules match this and the program stays stuck in the initial state:

```
<generatedTop>
  <threads>
    <thread>
      <id>
        0
      </id>
      <k>
        join spawn { x = 24 ;  y = 6 ;  return x + y ;  .Stmts } ; ~> .
      </k>
    </thread>
  </threads>
  <state>
    .Map
  </state>
  <next-id>
    1
  </next-id>
</generatedTop>
```

The only matching rule requires the join expression to be an ``Int``:

`<k> join ID:Int ; => I ...</k>`

Since there is no heat/cool evaluation semantics defined for ``join`` expressions, the above program never evaluates the spawn expression and so never gets to a state that can match the above rule.

Also, the rule matching ``spawn`` expressions seems to only allow the entire ``k`` cell to be replaced with the thread ID that was just spawned:

`<k> spawn { Ss } => NEXTID </k>`

Whereas if we wanted heat/cool semantics on ``join`` then we'd probably want to loosen this rule. 

---

I made some small tweaks to the semantics for this lesson [here](https://gist.github.com/ColmBhandal/c0fe1485b1ffa6e98ebb60de5f3f7c91). Notable changes are as follows:

1. Join now includes strict heat/cool evaluation: `| "join" Exp ";"[strict]`
2. The spawn rule now matches a ``k``-sequence, rather than only matching the full contents of the ``k``-cell, as specified by the three dots at the end: `<k> spawn { Ss } => NEXTID ...</k>`

As a result, if we run the same program as earlier:

``krun -cPGM='join spawn { x = 24; y = 6; return x + y;  };'``

We now get the below result, which is what I would have expected. Note that the forked thread has been destroyed and its value has been returned to the main thread. However, maybe I misinterpreted the intent of the example?

```
<generatedTop>
  <threads>
    <thread>
      <id>
        0
      </id>
      <k>
        30 ~> .
      </k>
    </thread>
  </threads>
  <state>
    x |-> 24
    y |-> 6
  </state>
  <next-id>
    2
  </next-id>
</generatedTop>
```

ColmBhandal:
Closes #2533

### DIFF
--- a/k-distribution/k-tutorial/1_basic/17_cell_multiplicity/README.md
+++ b/k-distribution/k-tutorial/1_basic/17_cell_multiplicity/README.md
@@ -121,7 +121,7 @@ module LESSON-17-B
   rule <k> I1:Int + I2:Int => I1 +Int I2 ...</k>
 
   rule <thread>...
-         <k> spawn { Ss } => NEXTID...</k>
+         <k> spawn { Ss } => NEXTID ...</k>
        ...</thread>
        <next-id> NEXTID => NEXTID +Int 1 </next-id>
        (.Bag => 
--- a/k-distribution/k-tutorial/1_basic/17_cell_multiplicity/README.md
+++ b/k-distribution/k-tutorial/1_basic/17_cell_multiplicity/README.md
@@ -102,7 +102,7 @@ module LESSON-17-B
                | Int 
                | Exp "+" Exp [seqstrict]
                | "spawn" "{" Stmts "}"
-               | "join" Exp ";"
+               | "join" Exp ";" [strict]
 
   configuration <threads>
                   <thread multiplicity="*" type="Map">
@@ -121,7 +121,7 @@ module LESSON-17-B
   rule <k> I1:Int + I2:Int => I1 +Int I2 ...</k>
 
   rule <thread>...
-         <k> spawn { Ss } => NEXTID </k>
+         <k> spawn { Ss } => NEXTID...</k>
        ...</thread>
        <next-id> NEXTID => NEXTID +Int 1 </next-id>
        (.Bag => 


k-distribution/k-tutorial/1_basic/17_cell_multiplicity/README.md
```diff
@@ -121,7 +121,7 @@ module LESSON-17-B
   rule <k> I1:Int + I2:Int => I1 +Int I2 ...</k>
 
   rule <thread>...
-         <k> spawn { Ss } => NEXTID </k>
+         <k> spawn { Ss } => NEXTID...</k>
```
Baltoli: A very minor style nitpick:
```suggestion
         <k> spawn { Ss } => NEXTID ...</k>
```

Baltoli (approved):
These changes make sense to me - thanks @ColmBhandal. One tiny stylistic change but otherwise happy to merge.

~~@dwightguth @ehildenb I don't know what needs to be done to get the CI to run properly on this given that Colm is working from a fork of the repo - is there a button one of you can press to do so?~~

Didn't notice that I had permission to click this button - ignore previous!

PR #2546, opened 2022-04-18
Update outer parser endColumn

radumereuta:
There was an inconsistency between the Outer parser and Inner regarding the position of the last character of a term which created some issues in #2517.
Inner: notice the `^~` extends past the single character pointed by the error message.
```
[Error] Inner Parser: Parse error: unexpected token '[' following token '('.
	Source(nestedFunctionContextInFun.k)
	Location(8,16,8,17)
	8 |	rule 0 => #fun([[0 => 1]] <bar> 0 </bar>)(0)
	  .	               ^~
``` 
Outer:
```
 [Error] Compiler: Sort X was declared as a token. Productions of this sort...
        Source(tokenCheck.k)
        Location(10,16,10,21)
+       10 |      syntax X ::= fail()
+          .                   ^~~~~~
 [Error] Compiler: Sort Y was declared as a token. Productions of this sort...
        Source(tokenCheck.k)
        Location(13,16,13,16)
+       13 |      syntax Y ::= Z
+          .                   ^
```
The reasoning behind pointing the end column past the last character is to match with what the editor tells you when you select a piece of text.


Baltoli (approved):

SchmErik (approved):

radumereuta:
Check to see if the instructions for installing stack are up to date. (see Docker file)
Bruce: stack was down at one point. Check that also.
Dwight: modify the docker file somewhere to rebuild stack.

PR #2549, opened 2022-04-19
Delete dead code

radumereuta:
Should be simple to review. I used IntelliJs automatic refactoring.


dwightguth:
Looks like Jenkins is out of disk space again. I will review though.

dwightguth (approved):
This is fine as far as it goes, so I'm approving, but I am 100% certain that a more aggressive dead code detector will find more than this paltry amount of dead code.

PR #2550, opened 2022-04-20
use strict equality for upper bounds on sort parameters

dwightguth:
This change makes it so that if a production whose sort is a sort variable exists under another term, the sort we infer for that parameter will always be exactly equal to the expected sort at that position. There is probably an even more aggressive change we can make which propagates this assumption through the Java code and entirely eliminates such sort variables from consideration, but I will leave that for a future PR.

Some analysis on the c semantics indicates that this tends to have a modest improvement to the time spent doing sort inference. The mean, median, Q1, Q3, and 99th percentile of time spent in sort inference all drop as a result of this change, which makes sense because we are getting down on the search space. That being said, there is some possibility that in some cases this change might increase the time spent parsing rules somewhat; I have not observed this in practice, but the maximum time spent inferring for an individual rule did increase in the c semantics, although the overall time for the entire parsing did drop.

I intend to profile a couple other semantics tomorrow and will report back before marking this as ready for review.


dwightguth:
Some more performance data:

EVM: drops from 9.5 seconds total to 7.6 seconds. Overall, fewer rules take longer than 500 milliseconds to do type inference. Not a huge difference, but still an improvement.
WASM: drops from 9.7 seconds total to 6.7 seconds. Similar story to EVM.

I want to find another semantics that takes longer, but I am waiting to hear back from coworkers with suggestions of semantics to try as I tried several other semantics and either had trouble building them, or else they also parsed rules rather quickly.

radumereuta (approved):
Worst proof from KEVM:
```
6580 evm-semantics/tests/specs/erc20/ds/transferFrom-failure-1-b-spec.k:8   parse: 287 typeInf:5686
1280 evm-semantics/tests/specs/erc20/ds/transferFrom-failure-1-b-spec.k:8   parse: 255 typeInf: 421
```

mkr-mcd-specs - some worst cases:
```
master:
31512 /home/radu/work/mkr-mcd-spec/clip.md:250   parse: 822 typeInf:30569
  579 /home/radu/work/mkr-mcd-spec/clip.md:292   parse:  41 typeInf: 418
23112 /home/radu/work/mkr-mcd-spec/clip.md:302   parse: 797 typeInf:22205
 2025 /home/radu/work/mkr-mcd-spec/clip.md:327   parse: 117 typeInf:1814
134996 /home/radu/work/mkr-mcd-spec/clip.md:347   parse:1542 typeInf:133191
22211 /home/radu/work/mkr-mcd-spec/clip.md:377   parse:1522 typeInf:20575
 4258 /home/radu/work/mkr-mcd-spec/clip.md:402   parse: 291 typeInf:3842                   

this-branch:
 3284 /home/radu/work/mkr-mcd-spec/clip.md:250   parse:1233 typeInf:1868
  655 /home/radu/work/mkr-mcd-spec/clip.md:292   parse: 103 typeInf: 378
12487 /home/radu/work/mkr-mcd-spec/clip.md:302   parse:1000 typeInf:11305
 4594 /home/radu/work/mkr-mcd-spec/clip.md:327   parse: 403 typeInf:4062
20978 /home/radu/work/mkr-mcd-spec/clip.md:347   parse:2042 typeInf:18627
 4917 /home/radu/work/mkr-mcd-spec/clip.md:377   parse:1686 typeInf:3064
 1148 /home/radu/work/mkr-mcd-spec/clip.md:402   parse: 414 typeInf: 640             
```
Total kompile time dropped from **3m30s** to **1m33s** !!!

PR #2551, opened 2022-04-20
modify frontend to use new Haskell backend serialization

dwightguth:


haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
```diff
@@ -41,9 +46,13 @@ public void accept(Backend.Holder h) {
...
         List<String> args = new ArrayList<>();
-        args.add("kore-parser");
-        args.add("--no-print-definition");
+        args.add("kore-exec");
         args.add("definition.kore");
+        args.add("--module");
+        args.add(kompileOptions.mainModule(files));
+        args.add("--output");
+        args.add("haskellDefinition.bin");
+        args.add("--serialize");
```
ehildenb, line 55: I assume this option implies "don't run the thing yet"?

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -111,7 +111,7 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
                 Module mod = getExecutionModule(module);
                 ModuleToKORE converter = new ModuleToKORE(mod, def.topCellInitializer, kompileOptions);
                 String koreOutput = getKoreString(k, mod, converter);
-                String defPath = files.resolveKompiled("definition.kore").getAbsolutePath();
+                String defPath = files.resolveKompiled("haskellDefinition.bin").getAbsolutePath();
```
ehildenb, line 114: Some of our makefile recipes depend on definition.kore.

Should it now _always_ be the case that we use haskellDefinition.bin instead? Never using definition.kore?

Also, I think that Amelei Leiden (working on Dedukti backend for K), relied on parsing definition.kore directly, so maybe we should emit both files (or maybe emit both with a flag).

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellBackend.java
dwightguth: Yes, --serialize means no actual execution takes place.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
dwightguth: It's certainly safe for a make rule targeting the haskell backend to use haskellDefinition.bin in a dependency to indicate that the definition is or is not built up to date. If you always build haskellDefinition.bin by invoking kompile, it is also safe to use definition.kore as you have previously, because kompile always updates both files. Of course, if you build the semantics, decide to try a different backend version but the same frontend version, and rebuild haskellDefinition.bin by invoking kore-exec --serialize directly, then any rule that depends on definition.kore to know when the compiled definition has changed will still think the target that depends on definition.kore is up to date. But it's possible people may generally just choose to rerun the frontend when they get the error saying their serialized definition is out of date.

All this being said, if your goal is just to know whether kompile needs to be rerun when k source files change, I would instead recommend using kdep and depending on the file `timestamp` instead of either of these.

dwightguth:
Roughly speaking, the performance improvement from this change on C is what I reported previously: taking 0 steps takes roughly 7 seconds (almost entirely consisting of simplifying the initial configuration) instead of roughly 40 seconds.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
ehildenb: What about the ability to emit the original `definition.kore` for Amelie? Is that still going to be supported somehow? If not, I'd ask that you add a flag to do so.

I don't think using `--backend kore` is good enough either, I want to know that the definition.kore is the exact one we're feeding to the haskell backend.

dwightguth: It's still emitted regardless. The kore-exec --serialize tool needs it as input.

ehildenb (approved):

PR #2553, opened 2022-04-21
Add check to detect duplicate initialization of cells

### LINKED ISSUE #2171: [Bug] [kompile] - does not report duplicate <k> cell declaration
K Version
---------

```
K version:    5.1.165
Build date:   Fri Aug 27 00:41:16 UTC 2021
```

Description
-----------

Importing `DEFAULT-CONFIGURATION` and declaring your own `<k>` kompiles successfully, and produces weird results when both cells use the `$PGM` variable.

Input Files
-----------

File `test.k`:

```
module TEST-PRE
    imports DEFAULT-CONFIGURATION
    imports INT

    configuration <test> 0 </test>
endmodule

module TEST
    imports TEST-PRE

    configuration <k> $PGM:Int </k> <test/> 
                                                      
endmodule     
```

Reproduction Steps
------------------

Call `kompile` on the definition:

```
% kompile --backend llvm test.k
```

No error output. Inspect `test-kompiled/configVars.sh`:

```
% cat test-kompiled/configVars.sh 
#!/usr/bin/env bash

declaredConfigVar_PGM='Int'
declaredConfigVar_PGM='KItem'
declaredConfigVars=(
    'PGM'
    'PGM'
)
```

Expected Behavior
-----------------

`kompile` should complain that I've redeclared the `<k>` cell. Or perhaps we should detect the double-declaration of the `$PGM` variable.

SchmErik:
This pull request adds a check to detect duplicate initialization of configuration cells.

Closes: #2171


radumereuta:
The error is reported from the llvm-backend at this location: https://github.com/runtimeverification/llvm-backend/blob/master/matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala#L797
Investigate if this is not an issue with the check, instead of modifying the solution chosen to fix this issue.

radumereuta:
@SchmErik is there any update on this?

SchmErik:
See #2827 closing

PR #2557, opened 2022-04-22
Report bad parameters to `kompile` etc.

### LINKED ISSUE #2548: Kompile/kprove doesn't complain about bad command line arguments
Talking to @tothtamas28 we got to the conclusion that one way to improve the usability of K is to have a better CLI
For example, if you give a bad argument, kompile/kprove is not going to complain, and it's going to send that argument to the backend.

Maybe we can find a better way to send arguments to the backend and have the front end sanitize its arguments.

Another solution would be to make sure we get a complaint from the backend if we send it bad arguments.

Ex: kompile doesn't complain
```
 kompile test.k -v -xxxxxxx
```

Baltoli:
This PR improves the reporting of errors when too few or too many _main parameters_ are passed to K tools that use the outer parsing options. Previously, we validated that at least one option was passed, but if too many were, we'd silently discard the extras.

Because of the way that JCommander is designed, anything passed at the command line that isn't registered as a named `@Parameter` will get added to the "main" parameter list. There is no way to distinguish between "misspelt option" and "additional file to parse" when using JCommander in this way.

An alternative design to this PR would be to set the main parameter to be a single `String`, rather than a `List<String>`. Doing so would provide a similar behaviour to this code, but without the ability to customise the message.

The new messages generated by this PR look like this:
```
outer parsing. The main files provided were: [test.k, extra.k]
```
```
outer parsing. No main file was provided.
```

The test for this feature doesn't play nicely with the existing Makefile infrastructure, so I've written it to just manually invoke `kompile` with the intended flags. If we want more tests like this in the future, we can revisit it. I'm happy with it as a one-off.

Fixes #2548


Baltoli:
Thanks @tothtamas28 !

kernel/src/main/java/org/kframework/utils/options/OuterParsingOptions.java
```diff
@@ -32,10 +33,33 @@ public OuterParsingOptions(File mainDefinitionFile) {
...
+        return base + " The main files provided were: " + mainFileErrorList();
+      }
+    }
+
     public synchronized File mainDefinitionFile(FileUtil files) {
         if (mainDefinitionFile == null) {
-            if (parameters == null || parameters.size() == 0) {
-                throw KEMException.criticalError("You have to provide exactly one main file in order to do outer parsing.");
+            if (parameters == null || parameters.size() != 1) {
+                throw KEMException.criticalError(mainFileErrorMessage());
```
radumereuta: This file has 2 and 4 spaces indentation.
Let's keep it consistent at 4.

radumereuta (changes requested):
> An alternative design to this PR would be to set the main parameter to be a single String, rather than a List<String>. Doing so would provide a similar behaviour to this code, but without the ability to customise the message.

I like this way better since it means we have to write and maintain less code.
The error message looks something like this:
```
[Error] Critical: Only one main parameter allowed but found several: "test.k"
and "-asf" (ParameterException: Only one main parameter allowed but found
several: "test.k" and "-asf")
```
I think this is readable enough (except for the duplication).

Baltoli:
Fine by me @radumereuta - will make the change

kernel/src/main/java/org/kframework/utils/options/OuterParsingOptions.java
```diff
@@ -24,20 +24,20 @@ public OuterParsingOptions() {}
...
     @Parameter(description="<file>")
-    private List<String> parameters;
+    private String mainParameter;
 
     private File mainDefinitionFile;
 
     public synchronized File mainDefinitionFile(FileUtil files) {
         if (mainDefinitionFile == null) {
-            if (parameters == null || parameters.size() == 0) {
+            if (mainParameter == null) {
```
radumereuta, line 37: Does this `if` ever trigger?
Judging by your test, it doesn't.

Baltoli: It does trigger; the test invocation is:
```console
$ make -C bad-flags 
make: Entering directory '/home/bruce/code/k/k-distribution/tests/regression-new/bad-flags'
/home/bruce/code/k/k-distribution/bin/kompile 2>&1 | diff - no-flags.out 
/home/bruce/code/k/k-distribution/bin/kompile --syntax-module TEST --no-exc-wrap --badflag --extra --backend llvm  test.k -d . 2>&1 | diff - extra-flags.out 
make: Leaving directory '/home/bruce/code/k/k-distribution/tests/regression-new/bad-flags'
```

The contents of the "no flags supplied" test are:
```console
$ cat bad-flags/no-flags.out 
outer parsing.
```

You can see that for this test to pass, the `if` has to trigger (i.e. `mainParameter == null`).

radumereuta: You are right, I wasn't paying attention.

radumereuta (approved):

PR #2560, opened 2022-04-25
Add `#trace` debugging hook

Baltoli:
This hook exposes the implementation in https://github.com/runtimeverification/llvm-backend/pull/513, along with a note explaining that calling it is expensive and should only be done as a debugging aid.

Additionally, an integration test for the hook is added to ensure that it's printing the expected syntax.

Requires https://github.com/runtimeverification/llvm-backend/pull/513 to be merged, and the corresponding submodule update to go through.

### DIFF
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -2427,7 +2427,8 @@ logged, which requires re-parsing the underlying K definition. Subsequent calls
 do not incur this overhead again; the definition is cached.
 
 ```k
-  syntax {Sort} K ::= #trace(value: Sort) [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol]
+  syntax K ::= #trace(value: KItem) [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol]
+             | #traceK(value: K)    [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol]
 ```
 
 ### Implementation of high-level I/O streams in K
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -2419,6 +2419,17 @@ finishes.
   syntax K ::= #logToFile(name: String, value: String) [function, functional, hook(IO.log), impure, returnsUnit, symbol]
 ```
 
+Terms can also be logged to standard error in _surface syntax_, rather than as
+KORE using `#trace`. This operator has similar semantics to `#logToFile` (i.e.
+it returns `.K`, but prints as an impure side effect). Note that calling
+`#trace` is equivalent to invoking the `kprint` tool for **each** term that is
+logged, which requires that the underlying K definition is re-parsed. Doing so
+is expensive, and so `#trace` should only be used as a debugging aid.
+
+```k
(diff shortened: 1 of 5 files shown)


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -2419,6 +2419,17 @@ finishes.
...
 
+Terms can also be logged to standard error in _surface syntax_, rather than as
+KORE using `#trace`. This operator has similar semantics to `#logToFile` (i.e.
+it returns `.K`, but prints as an impure side effect). Note that calling
+`#trace` is equivalent to invoking the `kprint` tool for the first term that is
+logged, which requires re-parsing the underlying K definition. Subsequent calls
+do not incur this overhead again; the definition is cached.
+
+```k
+  syntax {Sort} K ::= #trace(value: Sort) [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol]
```
dwightguth: I have to wonder if we might be better off changing the signature of this somewhat. People are probably going to end up pretty surprised if they try to trace a builtin sort, because it's not going to work. I would suggest we instead have two overloads, both of which have the same hook tied to them: one that takes a KItem and one that takes a K. The reason I suggest two is that if you just do K, then everything will get pretty-printed with a `~> .` at the end, but if you just do KItem, then you won't be able to pretty print k sequences. That being said, using the KItem one in most cases ought to be sufficient because it will automatically inject whatever sort you want into KItem and the injection will be pretty-printed invisibly, so you ought to get the same output as if you had pretty-printed a particular sort. Except that it will be able to handle builtin sorts.

dwightguth (changes requested):

k-distribution/include/kframework/builtin/domains.md
Baltoli: What's the best way to write that overloaded version? If I have:
```k
  syntax K ::= #trace(value: K)     [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol, klabel(#traceK)]
             | #trace(value: KItem) [function, functional, hook(IO.traceTerm), impure, returnsUnit, symbol, klabel(#traceKItem)]
```

then I get parsing ambiguities on the argument:
```
[Error] Inner Parser: Parsing ambiguity.
1: syntax K ::= "#trace" "(" K ")" [function, functional, hook(IO.traceTerm), impure, klabel(#traceK), returnsUnit, symbol]
    #traceK(`bar_*__TEST_Bar_Foo_Foo`(`foo(_)_TEST_Foo_Int`(#token("2","Int")),`foo(_)_TEST_Foo_Int`(#token("0","Int"))))
2: syntax K ::= "#trace" "(" KItem ")" [function, functional, hook(IO.traceTerm), impure, klabel(#traceKItem), returnsUnit, symbol]
    #traceKItem(`bar_*__TEST_Bar_Foo_Foo`(`foo(_)_TEST_Foo_Int`(#token("2","Int")),`foo(_)_TEST_Foo_Int`(#token("0","Int"))))
	Source(/home/bruce/code/k/k-distribution/tests/regression-new/trace/test.k)
	Location(13,20,13,47)
	13 |	    (#trace(F2) ~> #trace(bar foo(2) * foo(0)) ~> .K)
	   .	                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dwightguth (approved):

PR #2566, opened 2022-04-28
Replacement parser kernel

dwightguth:
This PR essentially deletes the "parser" part of the parser kernel and replaces it with a new parser algorithm which is easier to understand and maintain. It also includes several smaller refactorings that arose during this process. Functionally, the old and new parser kernels behave identically. In terms of performance, on its own, the new parser kernel may be slower than the old one in a few cases, but in those cases which we have observed, there is usually a very massive nest of ambiguities, which we are able to postprocess faster due to the changes in the commit titled "move all priority filters into a single pass for performance". Overall the time spent parsing in large semantics is on par or faster with this branch on all the semantics I have tested.

I tried very hard to document the new algorithm so if anything is hard to follow, please let me know and I can add additional documentation. Bear in mind it's not likely to be completely understandable without reading the two papers I linked to in the javadoc of the EarleyParser class, though.

This has been tested against EVM, WASM, Michelson, and C in addition to the regression test suite.


kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+    public String toString() {
+      return "syntax " + sorts.get(sort) + " ::= " +
+              items.stream().map(EarleyProductionItem::toString).reduce((s1, s2) -> s1 + " " + s2).orElse("");
+    }
+
+    /**
+     * @return true if prod has the token attribute, false otherwise.
+     */
+    public boolean isToken() {
+      Boolean isToken = this.isToken;
```
radumereuta, line 175: Why is the extra variable needed?
Same for `needsLabel` and `isMInt`

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+    }
+
+    /**
+     * Obtain the parse tree associated with a tuple (S, i, j) of sort, start-index, and end-index.
+     * @param sort The sort that was just parsed.
+     * @param start The start-index that the production was parsed from.
+     * @return The unique parse tree node associated with that sort, start-index, and end-index.
+     */
+    private Set<Term> completedParses(int sort, int start) {
+      CompleteParseTreeNode[][] nodes = completedParseTrees;
```
radumereuta, line 435: Again, `nodes` and `node` doesn't seem to be needed.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+      }
+      // add state to set
+      states.add(state);
+      // compute metadata about new state
+      membership.get(state.start).put(state, state);
+      if (state.item != state.prod.items.size() && state.prod.items.get(state.item).isNonTerminal()) {
+        EarleyNonTerminal nt = (EarleyNonTerminal)state.prod.items.get(state.item);
+        completor.get(nt.sort).add(state);
+      } else if (state.item == state.prod.items.size()) {
+        // if the state is coimplete, add the proper deriations to the CompleteParseTreeNode
```
radumereuta:
```suggestion
        // if the state is complete, add the proper derivations to the CompleteParseTreeNode
```

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+
+    }
+    return productions;
+  }
+
+  private EarleyProductionItem toEarley(ProductionItem item, Scanner scanner) {
+    if (item instanceof NonTerminal) {
+      Integer sort = sorts.get(((NonTerminal)item).sort());
+      if (sort == null) {
+        sort = 1; // <NULL>, a sort with no productions
```
radumereuta, line 677: Shouldn't this be an error?
When can you find a sort in a module that doesn't exist already?

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+      startLine = data.lines[t.startLoc];
+      startColumn = data.columns[t.startLoc];
+      endLine = data.lines[t.endLoc];
+      endColumn = data.columns[t.endLoc];
+    }
+    String msg = data.words.length-1 == k ?
+        "Parse error: unexpected end of file" :
+        "Parse error: unexpected token '" + data.words[k].value + "'";
+    if (k != 0) {
+      msg = msg + " following token '" + data.words[k - 1].value + "'.";
```
radumereuta, line 880: Since Erik added the underline to error messages do we still need this message?

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -0,0 +1,1023 @@
...
+
+  /*
+   * Construct the new parse tree for a state during the "complete" step.
+   *
+   * @param state The completed state corresponding to the non-terminal that was just parsed
+   * @param nextState The state to compute the parse tree for
+   * @param completed The state that was just prior to the non-terminal that was completed
+   * @param k The end-index of the parse tree
+   * @param data The {@link ParserMetadata} about the sentence being parsed.
+   */
```
radumereuta: ^?

radumereuta (changes requested):
This file can also be deleted: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/utils/algorithms/SCCTarjan.java

Overall pretty good. Indeed seems more straightforward. The previous implementation had other things in mind which didn't get implemented. The code seemed obfuscated because of that.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
dwightguth: This is a pretty standard lazy initialization procedure in java. By using a local variable and performing exactly one read and one write to the field, we are able to use a benign data race in the threading model of Java to lazily compute these fields from multiple threads without any synchronization at all. The reason we initialize these fields lazily is that it turns out that the hashmap lookup of looking every productions attributes up is pretty expensive even though we usually do not use the attributes of anwhere close to all the productions in the grammar in practice. So it's important as an optimization.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
dwightguth: Same deal here.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
dwightguth: this happens due to private productions, if you import a module privately, it is possible to have modules that rely on the existence of certain sorts despite those sort declarations not existing in that module. If you look, the old parser does something very similar in Grammar.java.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
dwightguth: Maybe not... I would suggest we make this change in a future PR though.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
dwightguth: Dead documentation I forgot to delete.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
radumereuta: Fair point, minimize the diff.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
radumereuta: Please add a comment with this explanation. This is not obvious from the code.

radumereuta (approved):

PR #2567, opened 2022-04-28
Specify kompiled dir directly

### LINKED ISSUE #1302: Allow specifying build output directory explicitly
Currently, all the K commands take `--directory` argument, which specifies where the main definition file `main.k` is located and then builds `main-kompiled` directory there.

This can be somewhat annoying, because we may have multiple ways of building the same semantics, and so need multiple subdirectories where we make copies of all the definition files. The implicit name `main-kompiled` is also annoying, because our `Makefile` then has to explicitly mention the word `main-` in the recipes for building the backend files.

It would be nicer if we could instead say exactly what we want the `kompiled` directory to be called, and if `krun` and `kprove` could take exactly that directory as output. For example:

```
kompile --build-directory .build/defn/web3 -I src/web3 --main-module WEB3 --syntax-module WEB3 src/web3/web3.k
krun --build-directory .build/defn/web3
```

Where the `interpreter` that is generated by `kompile` would now live at `.build/defn/web3/interpreter`.

radumereuta:
Fixes: #1302


tothtamas28 (commented):
I tried out this feature, seems to be working as expected. :rocket:

kernel/src/test/java/org/kframework/utils/inject/DefinitionLoadingModuleTest.java
```diff
@@ -19,11 +22,38 @@ public void testReadDefinition() throws IOException {
...
+    @Test
+    public void testSaveDefinition() throws IOException, KEMException {
+        OutputDirectoryOptions options = new OutputDirectoryOptions();
+        new JCommander(options, "-o", "src/test/resources/test-kompiled");
+
+        FileUtil fu = new FileUtil(null, new File("."), new File("src/test/resources/test-kompiled"), null, null);
+        File abcd = fu.resolveDefinitionDirectory("abacd");
+        System.out.println(abcd);
+    }
 }
```
Baltoli: Could we have a test for a non `*-kompiled` directory as well?

Baltoli (changes requested):

Baltoli (approved):
Looks good!

PR #2570, opened 2022-04-29
Start to deprecate `kprove`

Baltoli:
This PR is a first step towards deprecating the legacy `kprove` tool as `kprove-legacy`.

It moves the `kprove` code path to a new tool `kprove-legacy`, and adds a *temporary*, untested shim script to retain a `kprove` command. This will allow users to migrate from `kprove` to `kprove-legacy` without painful submodule updates.

Once that's done, we can go ahead and remove the `kprove` shim to allow us to continue with the second part of the renaming.


radumereuta (approved):
lgtm

Issue #2558, opened 2022-04-24
[kompile] Allow removing ACU collection patterns from definition


ehildenb:
Currently we have limited support for ACU collection patterns in K. A backend for K is being built that targets Dedukti, which is a language that does not support ACU collection patterns (or at least, the languages that Dedukti targets do not support them).

So a simple rule like these can be problematic:

```
    rule [r1]: <k> X:Id => I ...</k> <env>... X |-> I ...</env>

    rule [r2]: <k> X = I:Int; => . ... </k>
         <env>... X |-> (_ => I) ...</env>

    rule [r3]: <k> var (X,Xs => Xs);_ </k>
         <env> Rho:Map (.Map => X |-> 0) </env> requires notBool (X in_keys(Rho))
```

We would want the frontend to translate them to this:

```
    // _|->_ turned into _in_keys(_) constraint + lookup
    rule [r1.translated]: <k> X:Id => ENV[X] ...</k> <env> ENV </env> requires X in_keys(ENV)

    // _|->_ turned into _in_keys(_) constraint + write
    rule [r2.translated]: <k> X = I:Int; => . ... </k>
         <env> ENV => ENV [ X <- I ] </env>
      requires X in_keys(ENV)

    // .Map is removed from LHS
    rule [r3.translated]: <k> var (X,Xs => Xs);_ </k>
         <env> ENV => ENV ( X |-> 0 ) </env>
      requires notBool (X in_keys(ENV)) 
```

This could be done in two ways.

- Add option `--no-acu-maps` to the `kompile` command, which does this transformation.
- Add an option `--pyk-transfom <pyk-definition-transformer>`, which will call the given definition transformer at kompile time after the existing kompile pipeline. This makes it easier for people to write their own transformations of the K definition. We could ship `no-acu-maps.py` as a built-in transformation they can turn on.

radumereuta:
@tothtamas28 can you come up with a skeleton script that does some sort of transformation to the rules described above.
This will be called by `kompile --post-process <script.py>` and it expects the JSON file at stdout.

tothtamas28:
> a skeleton script that does some sort of transformation to the rules described above

Created a PR for it: #2729

radumereuta:
@tothtamas28 can we close this now?

tothtamas28:
> can we close this now?

If we decide not to implement this, then I think so:

> We could ship `no-acu-maps.py` as a built-in transformation they can turn on.

radumereuta:
[...]

Issue #2522, opened 2022-03-28
[Bug] [kast] - `ArrayIndexOutOfBoundsException` on term with top-level `#And`


tothtamas28:
K Version
---------

```
$ kompile --version
K version:    v5.3.2-4-g3884c6e34
Build date:   Fri Mar 25 13:20:52 CET 2022
```

Description
-----------

`kast` fails with `ArrayIndexOutOfBoundsException` on term with top-level `#And`.

Input Files
-----------

`example.k`:

```k
module EXAMPLE-SYNTAX
    imports INT-SYNTAX
endmodule


module EXAMPLE
    imports EXAMPLE-SYNTAX
    imports INT

    configuration <k> $PGM:Int </k>
endmodule
```

`term.json`:

```json
{
  "format": "KAST",
  "version": 1,
  "term": {
    "node": "KApply",
    "label": "#And",
    "variable": false,
    "arity": 2,
    "args": [
      {
        "node": "KApply",
        "label": "<k>",
        "variable": false,
        "arity": 1,
        "args": [
          {
            "node": "KVariable",
            "name": "N",
            "originalName": "N"
          }
        ]
      },
      {
        "node": "KApply",
        "label": "_>=Int_",
        "variable": false,
        "arity": 2,
        "args": [
          {
            "node": "KVariable",
            "name": "N",
            "originalName": "N"
          },
          {
            "node": "KToken",
            "sort": "Int",
            "token": "0"
          }
        ]
      }
    ]
  }
}
```

Reproduction Steps
------------------

```bash
$ kompile example.k --backend haskell
$ kast term.json -i json -o pretty -m EXAMPLE --debug
```

Expected Behavior
-----------------

Expected output:

```k
<k>
  N
</k>
#And N >=Int 0
```

Actual output:

```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
        at scala.collection.mutable.WrappedArray$ofRef.apply(WrappedArray.scala:193)
        at org.kframework.unparser.KPrint.filterConjunction(KPrint.java:252)
        at org.kframework.unparser.KPrint.filterSubst(KPrint.java:229)
        at org.kframework.unparser.KPrint.abstractTerm(KPrint.java:197)
        at org.kframework.unparser.KPrint.prettyPrint(KPrint.java:122)
        at org.kframework.unparser.KPrint.prettyPrint(KPrint.java:114)
[...]

PR #2572, opened 2022-04-29
JSON v2 - serialize label params

radumereuta:
Fixes: #2522
Fixes: #1509

PR #2581, opened 2022-05-05
Allow calling JSON post-processing step

### LINKED ISSUE #2453: Ensure conformance of pyk AST to K Frontend AST

ehildenb:
As part of: https://github.com/runtimeverification/k/issues/2558

Fixes: https://github.com/runtimeverification/k/issues/2453

Blocked on https://github.com/runtimeverification/k/issues/2522

This pull request does:

- Adds parameter `--post-process <command>`, which allows users to run the definition through some post-processing step they like via the JSON format.
- Adds a test of this parameter on `imp-haskell`, which exercises that `jq .` (the identity function on the JSON).
- Factors out `buildBisonParser` function for some code that I noticed was duplicated in `run` method of `Kompile`.

The onus is on the user here to make sure the produced definition is valid if they use `--post-process`.


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -215,6 +225,20 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
         return def;
     }
 
+    private Definition postProcessJSON(Definition defn, String postProcess) {
+    //public K externalParse(String parser, String value, Sort startSymbol, Source source, CompiledDefinition compiledDef) {
+        List<String> command = new ArrayList<>(Arrays.asList(postProcess.split(" ")));
+        Map<String, String> environment = new HashMap<>();
+        RunProcess.ProcessOutput output = RunProcess.execute(environment, files.getProcessBuilder(), command.toArray(new String[command.size()]));
+
```
ehildenb: @radumereuta I don't know the proper way to do this. What I want is to pipe the JSON of the defn into this process, and extract the stdout. I don't know how to pipe the JSON of the defn into stdin here.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -215,6 +225,25 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
...
+    private Definition postProcessJSON(Definition defn, String postProcess) {
+    //public K externalParse(String parser, String value, Sort startSymbol, Source source, CompiledDefinition compiledDef) {
+        List<String> command = new ArrayList<>(Arrays.asList(postProcess.split(" ")));
+        Map<String, String> environment = new HashMap<>();
+        try {
+            String inputDefinition = new String(ToJson.apply(defn), "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw KEMException.criticalError("Could not encode definition to JSON!");
+        }
+        RunProcess.ProcessOutput output = RunProcess.execute(environment, files.getProcessBuilder(), command.toArray(new String[command.size()]));
```
ehildenb: @radumereuta here I want to feed the string `inputDefinition` into the process that's running here on stdin. But I don't know how to do that.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -132,4 +132,7 @@ public boolean isKore() {
 
     @Parameter(names="--top-cell", description="Choose the top configuration cell when more than one is provided. Does nothing if only one top cell exists.")
     public String topCell;
+
+    @Parameter(names={"--post-process"}, description="JSON KAST => JSON KAST converter to run on definition after kompile pipeline.")
+    public String postProcess;
```
radumereuta, line 140: KAST is a lossy format meant for readability.
To make sure we can output the definition and load it back we need to use KORE.
The JSON format is not ready right now.
This PR should try and fix one of the issues: https://github.com/runtimeverification/k/pull/2572
But there might be more inconsistencies.
So I would mark this PR as blocked for now.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -215,6 +207,35 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
...
+        if (output.exitCode != 0) {
+            throw KEMException.criticalError("Post-processing returned a non-zero exit code: "
+                    + output.exitCode + "\nStdout:\n" + new String(output.stdout) + "\nStderr:\n" + new String(output.stderr));
+        }
+        return JsonParser.parseDefinition(output.stdout);
+    }
```
radumereuta, line 244: I see two ways to send the JSON content to the external processor:
- save it to a temp file and give it as input to your processor
- modify `RunProcess.execute` to take as input a OutputStream. This link seems helpful: https://stackoverflow.com/questions/18903549/writing-to-inputstream-of-a-java-process

I would prefer the first option though since it leaves room for debugging. I could inspect the JSON file and look for errors when they will happen.

ehildenb: I went with the temp file solution, and indeed having it present to inspect was nice.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
ehildenb: Done, I've marked it as blocked.

ehildenb:
Latest run of this in the test-suite produces this output:

```
/home/dev/src/k/k-distribution/bin/kompile --post-process 'jq .' --verbose --debug --no-exc-wrap --backend haskell  imp.k -d .
Parse command line options                                   =  0.030s
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/prelude.md)
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/kast.md)
Importing: Source(/home/dev/src/k/k-distribution/target/release/k/include/kframework/builtin/domains.md)
Importing: Source(/home/dev/src/k/k-distribution/tests/regression-new/imp-haskell/imp.k)
Outer parsing [65 modules]                                   =  1.084s
Parse configurations [0/3 declarations]                      =  0.365s
  New scanner: IMP-RULE-CELLS                                =  0.285s
Parse rules [0/117 rules]                                    =  0.345s
Validate definition                                          =  0.162s
Apply compile pipeline                                       =  0.492s
Post process step: jq . /tmp/tmp-compiled8562961800281154208.json =  0.287s
java.lang.NullPointerException
        at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getBoolean(JsonObjectBuilderImpl.java:229)
        at org.kframework.parser.json.JsonParser.lambda$toFlatModule$2(JsonParser.java:151)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
        at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1085)
        at org.kframework.parser.json.JsonParser.toFlatModule(JsonParser.java:151)
        at org.kframework.parser.json.JsonParser.toDefinition(JsonParser.java:129)
        at org.kframework.parser.json.JsonParser.parseJsonDef(JsonParser.java:115)
        at org.kframework.parser.json.JsonParser.parseDefinition(JsonParser.java:101)
        at org.kframework.kompile.Kompile.postProcessJSON(Kompile.java:244)
        at org.kframework.kompile.Kompile.run(Kompile.java:155)
        at org.kframework.kompile.KompileFrontEnd.run(KompileFrontEnd.java:78)
        at org.kframework.main.FrontEnd.main(FrontEnd.java:57)
        at org.kframework.main.Main.runApplication(Main.java:118)
        at org.kframework.main.Main.runApplication(Main.java:108)
        at org.kframework.main.Main.main(Main.java:56)
[Error] Internal: Uncaught exception thrown of type NullPointerException (NullPointerException: null)
make: *** [../../../include/kframework/ktest.mak:80: imp-kompiled/timestamp] Error 1
```

So it's failing sometime reading in the JSON it looks like.

radumereuta:
@ehildenb this is unblocked now.
I fixed a few more issues related to the translation to JSON, but there are A LOT more related to attributes.
Because we store them as actual classes, we have to treat them individually. (see `toAtt`)
I really don't like how this is turning out. We should discuss the problem, maybe we can find a better solution.

radumereuta:
I tested this change with KEVM and it worked.
@ehildenb please review as well.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -135,4 +135,7 @@ public boolean isKore() {
 
     @Parameter(names="--debug-type-inference", description="Filename and source line of rule to debug type inference for. This is generally an option used only by maintaners.")
     public String debugTypeInference;
+
+    @Parameter(names={"--post-process"}, description="JSON KAST => JSON KAST converter to run on definition after kompile pipeline.")
```
Baltoli, line 139: I can see that this is documented in the `--help` output, but would something like `--post-process-json` be a clearer name for the option?

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
```diff
@@ -296,7 +298,27 @@ public static Att toAtt(JsonObject data) {
...
+            } else if (key.equals(Att.PREDICATE())) {
+                newAtt = newAtt.add(Att.PREDICATE(), Sort.class, toSort(attMap.getJsonObject(key)));
+            } else if (key.equals("cellOptAbsent")) {
+                newAtt = newAtt.add("cellOptAbsent", Sort.class, toSort(attMap.getJsonObject(key)));
+            } else if (key.equals("cellFragment")) {
+                newAtt = newAtt.add("cellFragment", Sort.class, toSort(attMap.getJsonObject(key)));
+            } else if (key.equals("sortParams")) {
+                newAtt = newAtt.add("sortParams", Sort.class, toSort(attMap.getJsonObject(key)));
+            } else
+                newAtt = newAtt.add(key, attMap.getString(key));
```
Baltoli, line 321: What's the change that's meant we need to special-case these now? It might be worth pulling this if -else chain out into its own method.

kernel/src/main/java/org/kframework/unparser/ToJson.java
```diff
@@ -137,7 +139,32 @@ public static JsonStructure toJson(Att att) {
...
+            } else if (key._1().equals(Att.PREDICATE())) {
+                jattKeys.add(key._1(), toJson(att.get(Att.PREDICATE(), Sort.class)));
+            } else if (key._1().equals("cellOptAbsent")) {
+                jattKeys.add(key._1(), toJson(att.get("cellOptAbsent", Sort.class)));
+            } else if (key._1().equals("cellFragment")) {
+                jattKeys.add(key._1(), toJson(att.get("cellFragment", Sort.class)));
+            } else if (key._1().equals("sortParams")) {
+                jattKeys.add(key._1(), toJson(att.get("sortParams", Sort.class)));
+            } else
+                jattKeys.add(key._1(), att.att().get(key).get().toString());
```
Baltoli, line 167: Likewise here, if the logic is similar to the other if-else chain, how do we ensure they stay in sync?

pyk/src/pyk/kast.py
```diff
@@ -812,7 +812,7 @@ def let(
     ) -> 'KProduction':
         sort = sort if sort is not None else self.sort
         items = items if items is not None else self.items
-        klabel = klabel if klabel is not None else self.klabel
+        klabel = klabel if klabel is not None else self.klabel  # TODO figure out a way to set klabel to None
```
Baltoli, line 815: Should this TODO be resolved?

radumereuta: @tothtamas28 ^

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
radumereuta: IMO it would make it unnecessarily long.
JSON is just an implementation detail.

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
radumereuta: I think this location is fine, but you are correct in noticing this isn't pretty.
Some of our attributes can take entire classes: [here](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/attributes/Att.scala#L65-L66)
which makes the java code very convenient.
Since now we want to serialize and deserialize the entire definition we ran into the issue that we couldn't deserialize some attributes back into the structure they came from. This code special cases those classes (the ones I could find).

kernel/src/main/java/org/kframework/unparser/ToJson.java
radumereuta: This is the serialization logic. The pair of the above^.

pyk/src/pyk/kast.py
tothtamas28: This can be resolved in a later PR. We will need to perform similar changes for some other classes as well.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
Baltoli: Fair enough - seems fine to me

Baltoli (approved):

PR #2582, opened 2022-05-05
Nix improvements

### LINKED ISSUE #2577: Nix improvements
There are a few issues with the nix derivations in this repo, which cause issues when trying to use a K shell in a different project like the [evm-semantics](https://github.com/runtimeverification/evm-semantics). Namely:

* When compiling the EVM semantics repo within a K nix shell, the nix derivation for K should re-export/link the `include` folder from the `llvm-backend` derivation as well as the `cmake` folder from `llvm-backend` source.
* the release flag is not being passed down to `haskell-backend` in `default.nix`
* cannot use `import ./deps/k {}` when K is a submodule inside `evm-semantics` (fails with `getting attributes of path '/.git/modules/deps/k': No such file or directory`)

Other suggestions:
* the version of Z3 is currently unpinned in the nix derivation. Given how sensitive K seems to be to the right version of Z3, Niv + overlay could be used to select a specific version from github.
* it seems more nix-y for `nix/k.nix` to read in the version number from `package/version` rather than the current way of replacing the version via `package/version.sh`.

goodlyrottenapple:
This PR introduces imporvements mentioned in #2577, namely:

* pins Z3 to version 4.8.15
* links missing `cmake/` and `include/` folders from the `llvm-backend` derivation to the `k` derivation

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -8,7 +8,6 @@ in
 # Build an optimized release package.
 # Currently requires dependents to use LTO. Use sparingly.
 , release ? false
-, haskell-backend ? null
 }:
 
 let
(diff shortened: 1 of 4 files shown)


radumereuta:
@calin1304 you are the last person we asked to review nix changes.
Could you look at this one as well?

calin1304:
Ok, I will take a look at it.

calin1304:
LGTM

radumereuta (approved):

PR #2584, opened 2022-05-06
Force manual Z3 installation in published docker images

### LINKED ISSUE #2583: Incorrect Z3 version in K docker image
@gtrepta reports on Slack:
> The version of z3 in our k dockerhub containers should be at `4.8.15`, right? I've pulled `runtimeverificationinc/kframework-k:ubuntu-focal-5.3.34` and `z3 --version` in the container returns `Z3 version 4.8.7 - 64 bit`. Can anyone replicate this?

I can reproduce this on my machine; we should instead install Z3 4.8.15.

Baltoli:
We specify that users should install Z3 version 4.8.15 when using K. However, in our published Docker images, we supply 4.8.7 because that's the version supplied by Ubuntu's package repositories.

#2583 identifies that the intent of our Dockerfiles is not reflected by the end result - we install Z3 from source, then immediately trample on it by installing the dependency from apt.

This PR addresses the issue by making Z3 a `Recommended` package in our debian packaging scripts. This means that _some_ version of Z3 (4.8.7) will be installed by default when users install K from `apt`, but in our Docker images we can override this and install the correct version from source.

### DIFF
--- a/package/debian/control.bionic
+++ b/package/debian/control.bionic
@@ -10,7 +10,8 @@ Package: kframework
 Architecture: any
 Section: devel
 Priority: optional
-Depends: bison , clang-10 , default-jre-headless , flex , gcc , libgmp-dev , libjemalloc-dev , libmpfr-dev , libyaml-0-2 , libz3-4 , lld-10 , pkg-config , z3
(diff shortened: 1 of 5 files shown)


gtrepta (approved):
Looks good to me!

dwightguth (approved):

Issue #2559, opened 2022-04-24
match returns empty substitution for identity substitution


ehildenb:
Currently, if you do `match(KVariable('X'), KVariable('X'))`, you will get back `FrozenDict({})`, instead of `FrozenDict({'X': KVariable('X')})`.

It's important to get back the later, because if you have some match like `match(KApply('f', [KVariable('X'), KVariable('X')]), KApply('f', [KVariable('X'), intToken(3)]))` (in psuedocode: `match(f(X, X), f(X, 3))`), you want the match to fail when it tries to combine the submatches for the arguments of `f`. Currently combining the submatches does not fail, beacuse the match for the first argument gives back the identity substitution as the empty substitution.

This is affecting the alias undo-er in updating the KEVM version: https://github.com/runtimeverification/erc20-verification/pull/376/files

For now, I'll hack around it there.

PR #2586, opened 2022-05-09
Perform fixes in pyk for ksummarize

tothtamas28:
Fixes https://github.com/runtimeverification/erc20-verification/issues/431


pyk/src/pyk/subst.py
```diff
@@ -103,3 +157,36 @@ def combine_dicts(dict1: Optional[Mapping[K, V]], dict2: Optional[Mapping[K, V]]
...
+
+def replaceWith(rule, pattern):
+    (ruleLHS, ruleRHS) = rule
+    if ruleLHS == pattern:
+        return ruleRHS
+    return pattern
+
+
+def replaceAnywhereWith(rule, pattern):
+    return bottom_up(lambda p: replaceWith(rule, p), pattern)
```
ehildenb: If we're going to move these, I think we might as well rename them too (for the new naming convention).

I also think perhaps they belong in a different file, called `rewrite.py` or something.

ehildenb: But not sure if it makes sense to make a separate file yet. I just would not expect to look in `subst.py` for tools do do rewriting. Maybe `transform.py`?

ehildenb: Though `transform.py` to me sounds like definition transformations, so maybe don't use that name. Here is fine, maybe we just do the rename for now.

pyk/src/pyk/subst.py
```diff
@@ -46,6 +55,51 @@ def replace(term):
 
         return bottom_up(replace, term)
 
+    def unapply(self, term: KInner) -> KInner:
+        new_term = term
+        for var_name in self:
+            new_term = replaceAnywhereWith((self[var_name], KVariable(var_name)), new_term)
+        return new_term
```
ehildenb: We need a way to take into account that not all substitutions should be reversed on the input term.

For example `CALLVALUE_CELL ==Int 0` should not be reversed, because then a bunch of other `0` in the configuration, which are not related to `CALLVALUE_CELL`, get reversed.

For example, here: https://github.com/runtimeverification/erc20-verification/blob/master/ksummarize/src/ksummarize/util.py#L111 we filter to only consider unapplying substitutions where the value being replaced with a variable is not a KToken.

pyk/src/pyk/subst.py
```diff
@@ -46,6 +55,51 @@ def replace(term):
...
+
+
+def extract_subst(term: KInner) -> Tuple[Subst, KInner]:
+
+    def _subst_for_terms(term1: KInner, term2: KInner) -> Optional[Subst]:
+        if type(term1) is KVariable and not type(term2) is KVariable:
+            return Subst({term1.name: term2})
+        if type(term2) is KVariable and not type(term1) is KVariable:
+            return Subst({term2.name: term1})
```
ehildenb: And here we are filtering for only substitutions that are not of the form `KVariable == KVariable`.

So it could be that the extra check for whether one side is a token (mentioned here https://github.com/runtimeverification/k/pull/2586/files#r869556888) actually happens here, so that when you have `KVariable == KToken`, it doesn't get counted as a substitution at all by `extract_subst`.

pyk/src/pyk/cterm.py
```diff
@@ -16,9 +17,15 @@ class CTerm:
 
     def __init__(self, term: KInner) -> None:
         config, constraint = splitConfigAndConstraints(term)
-        constraints = tuple(flattenLabel('#And', constraint))
+        constraints = dedupe(flattenLabel('#And', constraint))
```
ehildenb: Is `dedupe` faster or something like `list(set(...))`?

pyk/src/pyk/subst.py
tothtamas28: I like the name `rewrite.py` for the whole module. What do you think?

pyk/src/pyk/cterm.py
tothtamas28: Good point, `list(set(...))` is faster.

At this point, do we use `dedupe` over non-hashable types? If not, we can change the default implementation there so that it is as fast as `list(set(...))`, but keeps the order of elements.

pyk/src/pyk/subst.py
tothtamas28: I think this can be achieved by masking the subterms that should not be reveresed by fresh variables before unapplying. Would that work? If so, we can add a helper function for that.

pyk/src/pyk/subst.py
ehildenb: Yep, sounds good.

tothtamas28: Done.

pyk/src/pyk/kcfg.py
```diff
@@ -347,6 +356,25 @@ def edges(self, *, source_id: Optional[str] = None, target_id: Optional[str] = N
 
         return [edge for edge in res if not target_id or target_id == edge.target.id]
 
+    def contains_edge(self, edge: Edge) -> bool:
+        if other := self.edge(source_id=edge.source.id, target_id=edge.target.id):
+            return edge == other
+        return False
+
+    def create_edge(self, source_id: str, target_id: str, condition: KInner = TRUE, depth=1) -> Edge:
```
ehildenb: I'm not sure it makes sense to have a default `depth` for this one actually. Maybe we should force the user to supply the `depth`?

pyk/src/pyk/rewrite.py
```diff
@@ -46,6 +55,54 @@ def replace(term):
 
         return bottom_up(replace, term)
 
+    def unapply(self, term: KInner) -> KInner:
+        new_term = term
+        for var_name in self:
+            lhs = self[var_name]
+            if type(lhs) is not KToken:
+                rhs = KVariable(var_name)
+                new_term = replace_anywhere((lhs, rhs), new_term)
```
ehildenb: We don't need this `KToken` check anymore here, because `extract_subst` does not consider substitutions for `KToken` anymore.

pyk/src/pyk/rewrite.py
```diff
@@ -103,3 +160,40 @@ def combine_dicts(dict1: Optional[Mapping[K, V]], dict2: Optional[Mapping[K, V]]
...
+
+
+def rewrite(rule: Tuple[KInner, KInner], term: KInner) -> KInner:
+    """
+    Rewrite a given term at the top with the supplied rule.
+
+    :param rule: A rewrite rule of the form (lhs, rhs).
+    :param term: Term to rewrite.
+    :return: The term with the rewrite applied once at the top.
+    """
```
ehildenb: Are these docstrings helpful to keep anymore? Will they be used eventually maybe?

pyk/src/pyk/tests/test_rewrite.py
```diff
@@ -82,3 +89,46 @@ def test_apply(self):
...
+
+    def test(self):
+        for i, [term, expected_subst, expected_term] in enumerate(self.TEST_DATA):
+            with self.subTest(i=i):
+                # When
+                actual_subst, actual_term = extract_subst(term)
+
+                # Then
+                self.assertDictEqual(dict(actual_subst), expected_subst)
+                self.assertEqual(actual_term, expected_term)
```
ehildenb: Can we also add a test of `extract_subst` followed by `unapply_subst`, we have such an example here: https://github.com/runtimeverification/erc20-verification/blob/master/ksummarize/src/ksummarize/tests/test_subst.py#L11

Also, let's add a test of just `extract_subst` when a substitution with a `KToken` is involved (it shouldn't treat it as a substitution basically, so the extracted subst should be empty).

pyk/src/pyk/subst.py
ehildenb: I don't understand this, what you're doing by filtering for `KToken` is already enough.

pyk/src/pyk/cterm.py
ehildenb: So we would continue to use `dedupe`, but it would be faster and preserve order? I would like it to preserve order, that seems like a nice property...

pyk/src/pyk/kcfg.py
tothtamas28: Done. (I also removed the default value for `condition`, otherwise I would have had to change the order of arguments.)

pyk/src/pyk/rewrite.py
tothtamas28: Done.

pyk/src/pyk/rewrite.py
tothtamas28: > Will they be used eventually maybe?

I think they should be (though I have to admit, I really have to force myself to write them).

pyk/src/pyk/tests/test_rewrite.py
tothtamas28: > Also, let's add a test of just extract_subst when a substitution with a KToken is involved

That one is already covered: c57df5db2e92a52ce7d45458fddeda3ca2e9b730.

tothtamas28: > Can we also add a test of extract_subst followed by unapply_subst

Done.

pyk/src/pyk/cterm.py
tothtamas28: Added a function `unique` for this purpose: `dedupe(xs) == list(unique(xs))`, `tuple(dedupe(xs)) == tuple(unique(xs))`, etc. We can remove / reimplement `dedupe` once we're sure we do not try to use it on collections of unhashable elements.

pyk/src/pyk/tests/test_rewrite.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class SubstTest(TestCase):
+
+    def test_eq(self):
+        # Given
+        test_data = (
+            ({}, {}),
+            ({'x': x}, {}),
+            ({}, {'x': x}),
```
ehildenb: Are these substitutions supposed to be considered equal?

Under composition then yes, I agree they should behave identically, but they should not be considered equal I don't think. Remember this issue where it becomes important that they are treated differently: https://github.com/runtimeverification/k/issues/2559

tothtamas28: Right now `Subst` is assumed to encode a total function, and `Subst({}) == Subst({'x': x})`.

But in cases where substitutions are supposed to behave like partial functions they shouldn't be considered equal, as for example `Subst({'x': x}).apply(x) == x` whereas `Subst({}).apply(x)` should fail / return `None`.

Let's think about a way to address both cases without creating too much confusion.

ehildenb: I don't think that substitutions should necessarily be thought of as functions, but they can be interpreted as functions. But there are multiple acceptable interpretations of them as functions.

What I really want is a `_.union(_)` operator that fails when a substitution does not compose correctly, I think it's a better name than `_.compose(_)` for this operation.

`match(KVariable('X'), KVariable('X)).union(match(KVariable('X'), KApply('f', [KVariable['X']))` should fail. If it doesn't fail, then there is no functional interpretation of `Subst` at all, because here we wolud be giving multiple values for the same input variable `X`.

So I think we should store the substitution as `{'X', KVariable('X')}`, not as `{}`, and leave it up to the individual operations no how to interpret that. `_.union(_)` will treat it as defining the value of the substitution on `X` (and so it can't be redefined to be something else), and `_.apply(_)` will treat it as the identity mapping.

The other argument for storing the identity substitution is that it's strictly more information. It's always possible to take a substitution and remove those identities. But taking one without the identity substitutions and calculating which identities to add requires having access to the original terms that we are creating substitutions for (to find out the list of their free variables).

pyk/src/pyk/rewrite.py
```diff
@@ -103,3 +159,40 @@ def combine_dicts(dict1: Optional[Mapping[K, V]], dict2: Optional[Mapping[K, V]]
...
+    lhs, rhs = rule
+    subst = match(lhs, term)
+    if subst is not None:
+        return subst(rhs)
+    return term
+
+
+def rewrite_anywhere(rule: Tuple[KInner, KInner], term: KInner) -> KInner:
```
ehildenb: Can we also make all these functions take as argument a `KRewrite` instead of a `Tuple[KInner,KInner]`?

pyk/src/pyk/cterm.py
```diff
@@ -16,10 +16,15 @@ class CTerm:
...
-        constraints = tuple(flattenLabel('#And', constraint))
+        constraints = tuple(sorted(set(flattenLabel('#And', constraint)), key=self._constraint_sort_key))
         object.__setattr__(self, 'config', config)
         object.__setattr__(self, 'constraints', constraints)
 
+    @staticmethod
+    def _constraint_sort_key(term: KInner) -> Tuple[int, str]:
+        term_str = str(term)
+        return (len(term_str), term_str)
+
```
ehildenb: I think let's add a `_normalize_constraints` function which both does this sorting and other trivial simplifications like this one: https://github.com/runtimeverification/erc20-verification/pull/428/files#r872984309

ehildenb: And call it whenever the term or constraint changes (and at initialization of course).

tothtamas28: Good idea.

> call it whenever the term or constraint changes

Fortunately there's no need for that, this class is (designed to be) immutable.

pyk/src/pyk/rewrite.py
tothtamas28: Maybe `Union[Tuple[KInner, KInner], KRewrite]`?

pyk/src/pyk/tests/test_rewrite.py
tothtamas28: > What I really want is a _.union(_) operator that fails when a substitution does not compose correctly

You mean do a union of the dictionaries but return `None` if there is a key `k` such that `s1[k] != s2[k]`? I agree that only makes sense if we allow `{'x': x}`.

> it's a better name than _.compose(_) for this operation

Compose is a different operation: given two substitutions whose interpretations as total functions are `f` and `g`, it returns a substitution whose interpretation as a total function is the function composition of `f` and `g`. The two operations coincide on substitutions such that if `f(x) != g(x)`, then `f(x) = x` or `g(x) = x`.

> So I think we should store the substitution as `{'X', KVariable('X')}`, not as `{}`,

That is a straightforward solution for enabling `union`. What we lose by doing that is a normal form w.r.t. the function interpretation, i.e. `s1 == s2` iff `s1(x) == s2(x)` for any `x`. But maybe that's not an important property after all.

Let me try to propose a solution that fits all the use cases considered here. If we can't come up with such a solution, let's fall back to losing the normal form.

pyk/src/pyk/rewrite.py
tothtamas28: Or let's just add them as convenience methods to `KRewrite`.

pyk/src/pyk/tests/test_rewrite.py
ehildenb: I mean, the _only_ use cases we have right now are `_.union(_)` and `_.apply(_)` (and the second is probably not needed, but `_.union(_)` as I've described it is definitely needed for any sufficiently advanced rewriting engine). I don't think we have any uses of `_.compose(_)` as it is.

Like I said, I don't give a function interpretation to substitutions, they are substitutions, not functions. There are several _natural_ interpretations of them as functions, both partial and total, and we can expose conveniences for that, but we 100% do need `_.union(_)` to operate as I described.

Not having this `_.union(_)` operation makes `Subst` unusable for anyone trying to build their own pattern matching algorithms, they'll have to do one of:

- maintain the `Dict` underneath directly, and then just wrap it in a `Subst` at the end, which loses them all the benefits of having a `Subst` class at all,
- Manually track the free variables in the `pattern` term, and make sure that they fail `_.compose(_)` correctly if the returned substitution for the first argument is missing entries for free variables in that `pattern`, then why do we have the `Subst` convenience class anyway? or
- Make sure that we always have "clean-slate" variables, which is possible _sometimes_ while making sure that we maintain user-supplied variable names, but not easy in general. Maintaining user-supplied variable names is very difficult to do in general, but so far we have been able to do so. Also, once you start variable renaming (using a routine like `vars_apart`) those variables "leak" into the rest of your code-base and you have to start calling `vars_apart` in more places. `vars_apart` is not a cheap routine.

I have chosen to sacrifice user-defined variable names here to work around this current limitation of `Subst` class (third option above, and the easiest one to implement): https://github.com/runtimeverification/erc20-verification/blob/master/ksummarize/src/ksummarize/util.py#L251. Fortunately here, I'm pretty sure that those fresh variables are not going to leak into the rest of the term (they are only used in that routine there), which made me OK making that choice there, but I don't want to make fresh variables unless we absolutely need to.

pyk/src/pyk/rewrite.py
ehildenb: Why `Union`? We're already updating all the downstream code-bases, and it's easy enough for people to pass a `KRewrite(...)` instead of a `Tuple`, since they're isomorphic.

It's just a matter of stricter typing.

pyk/src/pyk/cterm.py
ehildenb: Nice :)

pyk/src/pyk/tests/test_rewrite.py
tothtamas28: I think I see your point now.

By the way, is this operation called `union` in the literature? If not, I'd call it `merge`.

There is a function `combine_dicts` in `pyk.rewrite`. Maybe after adding this new method, it'll be possible to no longer expose it. What do you think?

> I don't think we have any uses of _.compose(_) as it is.

Let's remove it then,

pyk/src/pyk/rewrite.py
tothtamas28: Alright, I'll add them to `KRewrite` as methods then.

pyk/src/pyk/tests/test_rewrite.py
ehildenb: Yeah, that's exactly what `combine_dicts` was for, we can remove it.

I think of it as the `Map` in K (the operation `_Map_`, which is only defined if the keys are distinct and is otherwise bottom). [I guess we actually call that `Map.concat`](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md#map-concatenation).

`merge`, `concat`, and `union` all sound fine to me

pyk/src/pyk/rewrite.py
ehildenb: Hmmmm, but that requires `Subst` to be defined when `KRewrite` is defined, is that possible?

ehildenb: But that would be cool! Just not worth mangling dependencies to achieve.

tothtamas28: > Hmmmm, but that requires Subst to be defined when KRewrite is defined

Yeah, good point. But we can merge `pyk.rewrite` into `pyk.kast`. Then we can maybe rename that module to `pyk.core`. Or is that too much?

pyk/src/pyk/tests/test_rewrite.py
ehildenb: Another simpler case of why we want to be able to know quickly "was the variable present in the original pattern" is the following `terminal` heuristic:

```
def kevm_is_terminal(cterm: CTerm) -> bool:
    config, _ = cterm
    k_subst = match(KSequence([kevm_halt(), KVariable('###CONTINUATION')]), getCell(config, 'K_CELL'))
    if k_subst and type(k_subst['###CONTINUATION']) is KVariable:
        statuscode_cell = getCell(config, 'STATUSCODE_CELL')
        if type(statuscode_cell) is KApply and statuscode_cell.arity == 0:
            return True
    return False
```

Here, we get the substitution for the `<k>` cell, and want to test not only that it worked (matching), but also that the variable `###CONTINUATION` actually had a binding in the mapping. I chose variable `###CONTINUATION` to practice defensive programming (avoid name clashes when I can), but I would also want this code to work if I happened to choose the same variable name for the continuation in the `<k>` cell as the user, without having to do `vars_apart`.

Notice that with the current implementation of `Subst`, this code would not work because if the user also chose variable name `###CONTINUATION` (or I chose a variable here that ended up clashing with a user), then the check `type(k_subst['###CONTINUATION'])` would give a "no such element" error.

pyk/src/pyk/cterm.py
tothtamas28: Done.

pyk/src/pyk/tests/test_rewrite.py
tothtamas28: Done.

pyk/src/pyk/rewrite.py
tothtamas28: Done.

pyk/src/pyk/kastManip.py
```diff
@@ -179,6 +160,45 @@ def extract_rhs(term: KInner) -> KInner:
...
+    conjuncts = flattenLabel('#And', term)
+    subst = Subst()
+    rem_conjuncts: List[KInner] = []
+
+    for conjunct in conjuncts:
+        new_subst = _extract_subst(conjunct)
+        if new_subst is None:
+            rem_conjuncts.append(conjunct)
+        else:
+            subst = subst.compose(new_subst)
```
tothtamas28: I guess this should be `union`. How should we handle the case when `subst.union(new_subst) is None`?

ehildenb: I think it should be a fatal error basically. We shouldn't have two different bindings for the same variable in the constraints.

It could very well be sound logically, but for now let's just make it a fatal error and then we'll be forced to deal with that case correctly when it comes up.

ehildenb: So you can have a check at the end of the function `if not subst: fatal('Producing subst failed!')`

ehildenb: Also, maybe this can be changed to say `if not new_subst:` instead of `if new_subst is None:`? I'm not sure when to use one or the other.

pyk/src/pyk/kast.py
```diff
@@ -1105,6 +1137,118 @@ def let_att(self, att: KAtt) -> 'KDefinition':
...
+
+    def __mul__(self, other: 'Subst') -> 'Subst':
+        return self.compose(other)
+
+    def __call__(self, term: KInner) -> KInner:
+        return self.apply(term)
+
+    def minimize(self) -> 'Subst':
+        return Subst({k: v for k, v in self.items() if v != KVariable(k)})
+
```
ehildenb, line 1164: `minimize()` is maybe not a good name, because it implies that it's giving a different representation of the same substitution.

This is fundamentally changing the substitution, and returning a different substitution. So maybe it should be `remove_identities(...)` instead.

I really want to hammer home the point that _substitutions are not functions_, they have interpretations as functions. They are bindings for variables. The substitutions `Subst({'X': KVariable('X')})` and `Subst({})` are not the same substitution.

pyk/src/pyk/kast.py
```diff
@@ -326,6 +329,35 @@ def let(self, *, lhs: Optional[KInner] = None, rhs: Optional[KInner] = None) ->
...
+        return term
+
+    def apply(self, term: KInner) -> KInner:
+        """
+        Attempt rewriting once at every position in a term bottom-up.
+
+        :param term: Term to rewrite.
+        :return: The term with rewrites applied at every node once starting from the bottom.
+        """
+        return bottom_up(self.apply_top, term)
```
ehildenb, line 351: Nice, this looks really good.

pyk/src/pyk/kast.py
```diff
@@ -1105,6 +1137,118 @@ def let_att(self, att: KAtt) -> 'KDefinition':
...
+    """Perform syntactic pattern matching and return the substitution.
+
+    -   Input: a pattern and a term.
+    -   Output: substitution instantiating the pattern to the term.
+    """
+
+    # TODO simplify
+    def _match(pattern: KInner, term: KInner) -> Optional[Dict[str, KInner]]:
+
+        def combine_dicts(dict1: Optional[Mapping[K, V]], dict2: Optional[Mapping[K, V]]) -> Optional[Dict[K, V]]:
```
ehildenb: Now that we have `Subst` supporting a proper `_.union(_)` operator, maybe `_match(...)` can be made to return an `Optional[Subst]` instead of an `Optional[Dict[str, KInner]]`, and we can get rid of `combine_dicts` altogether?

pyk/src/pyk/utils.py
```diff
@@ -91,6 +93,16 @@ def intersperse(iterable: Iterable[T], delimiter: T) -> Iterator[T]:
...
+    for elem in iterable:
+        if elem in elems:
+            continue
+        else:
+            elems.add(elem)
+            yield elem
+
+
 def dedupe(xs: Iterable[T]) -> List[T]:
     res = []
```
ehildenb: Is `dedupe` used anywehre in this library at hte moment? If not, let's remove it (and force downstream projects to switch to `unique`).

tothtamas28: Done.

pyk/src/pyk/kast.py
tothtamas28: Done.

pyk/src/pyk/kast.py
tothtamas28: The name comes from here:

https://github.com/runtimeverification/k/blob/fefe02c92cd16a9440275cc0cba9598e5337ba31/pyk/src/pyk/kastManip.py#L781

We can call it another name, but I'd prefer something shorter than `remove_identities`.

pyk/src/pyk/kastManip.py
tothtamas28: In this particular case it is safe to do `not new_subst`, as `extract_subst` never returns `Subst()`. But I'd like to keep it like this so that the reader doesn't have to think through this cornercase.

tothtamas28: > So you can have a check at the end of the function if not subst: fatal('Producing subst failed!')

Done.

ehildenb (approved):

PR #2592, opened 2022-05-12
Use Release again for packaging

gtrepta:
The build for the K package was changed to FastBuild in #1185 because of an llvm bug with LTO and tail call optimization. We require a minimum version of llvm now that no longer has this bug, so we should be able to change it back without issue.

Relevant issues:
https://github.com/runtimeverification/firefly/issues/374
https://github.com/runtimeverification/evm-semantics/pull/696


dwightguth:
I'm actually going to close this. A large part of why we don't use Release by default is that enabling lto makes compilation a lot slower, which is a poor choice for day to day development. So I think the default choice here is actually correct. If you think we should try to bundle a version of k capable of doing lto into a release, we can do that, but it's more complicated than this because this will make it impossible to build without lto.

PR #2593, opened 2022-05-13
Jammy (22.04) CI

Baltoli:
This PR makes several changes to the K CI scripts in preparation for the Bionic / Jammy switch:
* Ubuntu Focal is made the primary CI distribution on which the full K exercises test suite is run, replacing Bionic.
* Jammy is added as an extra variant to the set of distributions on which we build release packages; [recent changes to `dpkg`](https://wiki.debian.org/ToolChain/LTO) mean that jammy-specific options need to be passed when building `.deb` packages.

Important notes for this PR:
* Bionic is not yet removed from the CI scripts; this needs to follow a separate audit of downstream projects.
* As written, the Jenkins rules in this PR run more work on each new PR. This is to make sure that Bionic, Focal and Jammy all work _without_ having to attempt a merge to master. This PR should be followed up immediately by a [separate PR](https://github.com/runtimeverification/k/pull/2594) that undoes this change, and only runs all three versions on master.


package/debian/control.jammy
```diff
@@ -0,0 +1,18 @@
+Source: kframework
+Section: devel
+Priority: optional
+Maintainer: Dwight Guth <dwight.guth@runtimeverification.com>
+Build-Depends: clang-14 , cmake , debhelper (>=9) , flex , libboost-test-dev , libgmp-dev , libjemalloc-dev , libmpfr-dev , libyaml-dev , maven , openjdk-11-jdk , zlib1g-dev
```
dwightguth: this should be debhelper >= 10 since compat is set to 10.

package/debian/control.jammy
```diff
@@ -0,0 +1,18 @@
...
+Standards-Version: 3.9.6
+Homepage: https://github.com/runtimeverification/k
+
+Package: kframework
+Architecture: any
+Section: devel
+Priority: optional
+Depends: bison , clang-14 , default-jre-headless , flex , gcc , libgmp-dev , libjemalloc-dev , libmpfr-dev , libyaml-0-2 , libz3-4 , lld-14 , pkg-config
```
dwightguth: as I discovered in my other PR, this needs to include llvm-14.

package/debian/rules.jammy
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/make -f
```
dwightguth, line 1: this file appears identical to the existing rules file, so let's delete it and keep using that one.

dwightguth (changes requested):
looks okay, needs some minor improvements.

package/debian/rules.jammy
Baltoli: It differs in the line `export DEB_BUILD_MAINT_OPTIONS=hardening=-stackprotector optimize=-lto`; the link in the PR body gives some context that I'll add here in a comment.

radumereuta:
I'm going to let Dwight review this since he seems more knowledgeable than I am.

dwightguth (approved):

PR #2599, opened 2022-05-17
Reverse pyk match operator

ehildenb:
I was trying to explain the algorithm behind summarization to @xc93 and realized that on this line:

https://github.com/runtimeverification/erc20-verification/blob/master/ksummarize/src/ksummarize/kevm_summarize.py#L211

we are saying `term.match(pattern)`. I would much prefer for us to say `pattern.match(term)`, it matches the same direction as `match(pattern, term)` that I'm used to for a long time and makes more sense in my head. I can't provide a better justification for it than this, but I did confirm that this is the only use that needs to change.

### DIFF
--- a/pyk/src/pyk/cterm.py
+++ b/pyk/src/pyk/cterm.py
@@ -47,8 +47,8 @@ class CTerm:
     def hash(self) -> str:
         return self.term.hash
 
-    def match(self, term: 'CTerm') -> Optional[Subst]:
-        match_res = self.match_with_constraint(term)
+    def match(self, cterm: 'CTerm') -> Optional[Subst]:
+        match_res = self.match_with_constraint(cterm)
 
         if not match_res:
             return None
@@ -60,13 +60,13 @@ class CTerm:
 
         return subst
 
-    def match_with_constraint(self, term: 'CTerm') -> Optional[Tuple[Subst, KInner]]:
-        subst = match(pattern=self.config, term=term.config)
+    def match_with_constraint(self, cterm: 'CTerm') -> Optional[Tuple[Subst, KInner]]:
+        subst = match(pattern=self.config, term=cterm.config)
 
         if subst is None:
             return None
 
-        constraint = self._ml_impl(term.constraints, map(subst, self.constraints))
+        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))
 
         return subst, constraint
 
--- a/pyk/src/pyk/cterm.py
+++ b/pyk/src/pyk/cterm.py
@@ -47,8 +47,8 @@ class CTerm:
     def hash(self) -> str:
         return self.term.hash
 
-    def match(self, pattern: 'CTerm') -> Optional[Subst]:
-        match_res = self.match_with_constraint(pattern)
+    def match(self, term: 'CTerm') -> Optional[Subst]:
+        match_res = self.match_with_constraint(term)
 
         if not match_res:
             return None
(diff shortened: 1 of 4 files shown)


ehildenb:
Also, `pattern.match(term)` matches better the python way of defining `match` as:

```
class CTerm:
    def match(self, other):
        return match(self.config, other.config)
```

tothtamas28 (commented):
A previous discussion we had on this topic: https://github.com/runtimeverification/k/pull/2470#discussion_r831342486.

To sum up. my take on the issue is the following.

1. Both "pattern matches term" and "term matches pattern" are in use.
2. Both `map(patterm.match, iterable)` and `map(term.match, iterable)` have natural use cases.
3. `term.match(pattern) is not None` corresponds to "the set of models of `term` is a subset of the set of models of `pattern`". Notice that the order of operands is the same in both the code and its logical interpretation. This is also the direction we draw our abstraction edges in `KCFG`.

But I do not have a strong opinion on this one, so I'm fine with `pattern.match(term)`.

pyk/src/pyk/tests/test_match.py
```diff
@@ -30,10 +35,13 @@ def test_match_and_subst(self):
             with self.subTest(i=i):
                 # When
                 subst = match(pattern, term)
+                subst_cterm = _as_cterm(pattern).match(_as_cterm(term))
 
                 # Then
                 self.assertIsNotNone(subst)
                 self.assertEqual(subst(pattern), term)
+                self.assertIsNotNone(subst_cterm)
+                self.assertEqual(subst_cterm(pattern), term)
```
tothtamas28: I think `CTerm` should have its own set of tests.

pyk/src/pyk/cterm.py
```diff
@@ -47,8 +47,8 @@ def term(self) -> KInner:
     def hash(self) -> str:
         return self.term.hash
 
-    def match(self, pattern: 'CTerm') -> Optional[Subst]:
-        match_res = self.match_with_constraint(pattern)
+    def match(self, term: 'CTerm') -> Optional[Subst]:
+        match_res = self.match_with_constraint(term)
```
tothtamas28:
```suggestion
    def match(self, cterm: 'CTerm') -> Optional[Subst]:
        match_res = self.match_with_constraint(cterm)
```

pyk/src/pyk/cterm.py
```diff
@@ -60,13 +60,13 @@ def match(self, pattern: 'CTerm') -> Optional[Subst]:
 
         return subst
 
-    def match_with_constraint(self, pattern: 'CTerm') -> Optional[Tuple[Subst, KInner]]:
-        subst = match(pattern=pattern.config, term=self.config)
+    def match_with_constraint(self, term: 'CTerm') -> Optional[Tuple[Subst, KInner]]:
+        subst = match(pattern=self.config, term=term.config)
```
tothtamas28:
```suggestion
    def match_with_constraint(self, cterm: 'CTerm') -> Optional[Tuple[Subst, KInner]]:
        subst = match(pattern=self.config, term=cterm.config)
```

tothtamas28 (changes requested):

ehildenb:
Yeah, I remember being unsure what the correct direction was before, but I think the fact that I spent an hour being baffled at how the summarization algorithm was working was a strong enough indicator of what I naturally expected while working with the library.

pyk/src/pyk/cterm.py
ehildenb: Done.

pyk/src/pyk/cterm.py
ehildenb: Done.

pyk/src/pyk/tests/test_match.py
ehildenb: Done.

tothtamas28 (approved):

PR #2605, opened 2022-05-18
Pyk tools upgrades

ehildenb:
Some changes to pyk library to facilitate downstream progress:

- `KProve.prove` now respects the `dry_run` option without crashing. This is to facilitate inlining the call to `kprovex --dry-run --emit-json-spec ...` into the `ksummarize` python script, instead of in the `ktoken` bash script.
- `pushDownRewrites` is renamed to `push_down_rewrites`, and handles `KSequence` a bit better now (a test is added). Variables of this method are renamed as well to follow new conventions.
- `minimizeTerm` is renamed to `minimize_term`, and the parameters are renamed as well.
- The classes `KPrint` and `KProve` have their fields and methods renamed with the new naming convention.
- `splitConfigAndConstraint` will not fail silently when the passed in `KInner` is not actually a configuration and a constraint.
- Fixes `KProduction` to store a `KLabel` instead of a `str`.
- Renames `buildSymbolTable` to `build_symbol_table`, and removes several unused functions from `ktool.kprint.py`.
- Factors out `unparser_for_production` method, and adds test that was previously failing with `build_symbol_table`.
- Add `verified` attribute back to KCFG for edges.
- Several cases where we forgot to change a `kapply.label` to a `kapply.label.name` are fixed, and tests are added for those cases.
- The unused function `replaceKLabels` is removed (I've checked that it's unused in KEVM and ERC20).
- `KCFG.remove_node` now correctly adjusts the `expanded` set, a test that was failing is added.
- A bug in `reachable_nodes` where it would not traverse covers in forward-search mode is fixed, and a test added.


pyk/src/pyk/kastManip.py
```diff
@@ -265,6 +257,8 @@ def splitConfigAndConstraints(kast):
         else:
             constraints.append(c)
     constraint = buildAssoc(mlTop(), '#And', constraints)
+    if not term:
+        fatal(f'Could not find configuration for: {kast}')
```
tothtamas28:
```suggestion
        raise ValueError(f'Could not find configuration for: {kast}')
```

pyk/src/pyk/kcfg.py
```diff
@@ -166,6 +172,7 @@ def to_dict(self) -> Dict[str, Any]:
         init = sorted(self._init)
         target = sorted(self._target)
         expanded = sorted(self._expanded)
+        verified = [{"source": source, "target": target} for source, target in sorted(self._verified)]
```
```suggestion
        verified = [{"source": source_id, "target": target_id} for source_id, target_id in sorted(self._verified)]
```

pyk/src/pyk/kcfg.py
```diff
@@ -337,6 +355,7 @@ def remove_node(self, node_id: str) -> None:
         self._init.discard(node_id)
         self._target.discard(node_id)
         self._expanded.discard(node_id)
+        self._verified = set([(source_id, target_id) for source_id, target_id in self._verified if source_id != node_id and target_id != node_id])
```
```suggestion
        self._verified = set((source_id, target_id) for source_id, target_id in self._verified if source_id != node_id and target_id != node_id)
```

pyk/src/pyk/kcfg.py
```diff
@@ -320,6 +335,9 @@ def get_or_create_node(self, cterm: CTerm) -> Node:
     def remove_node(self, node_id: str) -> None:
         node_id = self._resolve(node_id)
 
+        for pred in [edge.source for edge in self.edges(target_id=node_id)]:
+            self.discard_expanded(pred.id)
```
```suggestion
        for in_edge in self.edges(target_id=node_id):
            self.discard_expanded(in_edge.source.id)
```

pyk/src/pyk/kcfg.py
```diff
@@ -564,11 +605,10 @@ def paths_between(self, source_id: str, target_id: str, *, traverse_covers=False
 
         return paths
 
-    def reachable_nodes(self, node_id: str, *, reverse=False, traverse_covers=False) -> Set[Node]:
-        node = self.node(node_id)
+    def reachable_nodes(self, start_node_id: str, *, reverse=False, traverse_covers=False) -> Set[Node]:
```
tothtamas28: Good catch. :+1:

tothtamas28: What do you think?

```suggestion
    def reachable_nodes(self, source_id: str, *, reverse=False, traverse_covers=False) -> Set[Node]:
```

tothtamas28 (approved):
Great set of improvements. :sparkles:

pyk/src/pyk/kastManip.py
ehildenb: Done.

pyk/src/pyk/kcfg.py
ehildenb: Done.

pyk/src/pyk/kcfg.py
ehildenb: Done.

pyk/src/pyk/kcfg.py
ehildenb: Done.

pyk/src/pyk/kcfg.py
ehildenb: Done.

PR #2606, opened 2022-05-18
Replace direct wirtes to stderr by standard logging in pyk

tothtamas28:


pyk/src/pyk/__main__.py
```diff
@@ -56,8 +57,6 @@ def main(extraMain=None):
         kprover = KProve(kompiled_dir, args['main-file'])
         finalState = kprover.prove(Path(args['spec-file']), args['spec-module'], args=args['kArgs'])
         args['output_file'].write(finalState.to_json())
-        if finalState != mlTop():
-            warning('Proof failed!')
 
     elif args['command'] == 'graph-imports':
```
ehildenb, line 66: How are we handling the case where `prove` returns non-zero exit code?

pyk/src/pyk/ktool/kprove.py
```diff
@@ -24,6 +18,8 @@
 from ..prelude import mlTop
 from .kprint import KPrint
 
+_LOGGER: Final = logging.getLogger(__name__)
+
```
ehildenb, line 26: This will get the same `_LOGGER` across different files?

ehildenb:
It looks good, can you make a branch of ERC20 repo (after update is merged: https://github.com/runtimeverification/erc20-verification/pull/428), which demonstrates how it's used in that codebase (and make sure it covers all our needs).

pyk/src/pyk/ktool/kprove.py
tothtamas28: This will create a logger instance for the `pyk.ktool.kprove` module.

From the [docs](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial):

> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
>
> ```python3
> logger = logging.getLogger(__name__)
> ```
>
>This means that logger names track the package/module hierarchy, and it’s intuitively obvious where events are logged just from the logger name.

pyk/src/pyk/__main__.py
tothtamas28: In such cases `prove` used to write an error message to `stderr` and raised `SystemExit`. Now it raises `RuntimeError` with the same error message as argument.

tothtamas28:
> can you make a branch of ERC20 repo (...) which demonstrates how it's used in that codebase

Created an issue: https://github.com/runtimeverification/erc20-verification/issues/453

pyk/src/pyk/ktool/kprove.py
```diff
@@ -115,17 +111,15 @@ def prove(self, specFile, specModuleName, args=[], haskellArgs=[], logAxiomsFile
...
+        _LOGGER.info(' '.join(command))
         process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, env=commandEnv)
         stdout, stderr = process.communicate(input=None)
         try:
             finalState = KAst.from_dict(json.loads(stdout)['term'])
         except Exception:
-            sys.stderr.write(stdout + '\n')
-            sys.stderr.write(stderr + '\n')
-            fatal(f'Exiting: process returned {process.returncode}')
+            raise RuntimeError(f'Process returned {process.returncode}')
```
ehildenb: We should include the stdout/stderr in the output somehow, via logging perhaps, at `logging.error(...)` level.

Getting that output is necessary for understanding how to fix the proof claim we're feeding to the backend.

ehildenb:
Does this logging infrastructure support the same timing information that `notif(...)` and other commands were printing out? That has been a key feature I've been using for identifying cost centers.

pyk/src/pyk/ktool/kprove.py
tothtamas28: We should probably pass this information along with the exception. Then the caller can decide what to do with that data.

Eventually, we should introduce a class `KProveError` (or a hierarcy of such classes) that stores the information extracted from the output of `kprove` in a structured form.

tothtamas28:
> Does this logging infrastructure support the same timing information that `notif(...)` and other commands were printing out? That has been a key feature I've been using for identifying cost centers.

It seems possible using `logging.Filter`: https://stackoverflow.com/a/43052949.

Alternatively, we can:

1. Implement a post-processing script that extracts performance information from the log.
2. Instrument critical parts of the code to collect and expose performance metrics.

pyk/src/pyk/ktool/kprove.py
ehildenb: That sounds good, let's do `f'Process returned {process.returncode}\nStdout:\n{stdout}\nStderr:\n{stderr}'` for now I think. I made a note on the project board about introducing an error class heirarchy.

ehildenb:
Let's make sure that we can pass some option which will turn on logging events around each subprocess call we can make (eg. calls into frontend, or other processes).  And make sure that we can turn on timestamping the log entries. That way we can still get course process-level profiling easily.

That's what the `--profile` option for both `kevm` and `krun` and `kalgo` means, is "course level profiling information", so let's use the same option.

tothtamas28:
> Let's make sure that we can pass some option which will turn on logging events around each subprocess call we can make (eg. calls into frontend, or other processes). And make sure that we can turn on timestamping the log entries. That way we can still get course process-level profiling easily.

As a proof of concept I added the `--verbose` flag to the `pyk` command line tool.

```bash
(venv-prod) .../pyk/pyk-tests $ kompile ../k-files/imp-verification.k --output-definition definitions/imp-verification --backend haskell -I k-files --emit-json

(venv-prod) .../pyk/pyk-tests $ pyk definitions/imp-verification prove ../k-files/imp-verification.k ../k-files/imp-unproveable-spec.k IMP-UNPROVEABLE-SPEC
# json output

(venv-prod) .../pyk/pyk-tests $ pyk --verbose definitions/imp-verification prove ../k-files/imp-verification.k ../k-files/imp-unproveable-spec.k IMP-UNPROVEABLE-SPEC
INFO 2022-05-19 14:01:27,713 pyk.ktool.kprove - kprovex ../k-files/imp-unproveable-spec.k --definition definitions/imp-verification -I definitions --spec-module IMP-UNPROVEABLE-SPEC --output json
# json output

```

pyk/src/pyk/ktool/kprove.py
```diff
@@ -115,17 +111,15 @@ def prove(self, specFile, specModuleName, args=[], haskellArgs=[], logAxiomsFile
         command += args
         commandEnv = os.environ.copy()
         commandEnv['KORE_EXEC_OPTS'] = ' '.join(haskellArgs + haskellLogArgs)
-        notif(' '.join(command))
+        _LOGGER.info(' '.join(command))
         process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, env=commandEnv)
         stdout, stderr = process.communicate(input=None)
```
ehildenb: So I think we need a log entry _after_ the `subprocess` call as well, so that if we have `-v` turned on, we can see the time delta for each of these commands.

Something like:

```
LOG: TIMESTAMP: Running: COMMAND
LOG: TIMESTAMP: Finished with returncode {returncode}: COMMAND
```

ehildenb: Maybe we factor out a `call_process` helper that has this specific logging instrumentation around the subprocess call and only exposes the subprocess uses we have used before.

pyk/src/pyk/ktool/kprove.py
tothtamas28: Done.

pyk/src/pyk/ktool/kprove.py
tothtamas28: Done.

pyk/src/pyk/ktool/kompile.py
```diff
@@ -28,10 +31,10 @@ def kompile(
...
-    proc_res = _kompile(str(main_file), *args)
-
-    if proc_res.returncode:
-        raise RuntimeError(f'Kompilation failed for: {main_file}')
+    try:
+        _kompile(str(main_file), *args)
+    except CalledProcessError as err:
+        raise RuntimeError(f'Command kompile exited with code {err.returncode} for: {main_file}', err.stdout, err.stderr)
 
     kompiled_dir = _kompiled_dir(main_file, output_dir)
```
ehildenb, line 39: So I don't quite understand this, isn't the exception that `run_process` is throwing already enough? Or is this so we can have a more specfiic error message in this case?

ehildenb (approved):

pyk/src/pyk/ktool/kompile.py
tothtamas28: Good point. There are two reasons for doing it like this.

* Although `CalledProcessError` contains `stderr` and `stdout`, when the exception is uncaught, they will not be shown in the stack trace.
* We will eventually create some custom error class that we can raise instead of `RuntimeError`.

PR #2609, opened 2022-05-18
Change type of KToken.sort from str to KSort in pyk.kast

### LINKED ISSUE #2602: KToken in pyk library should store a KSort, not a str
Currently KToken class in pyk https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kast.py#L195

It stores a `sort: str`, and should be storing a `sort: KSort` instead, for proper serialization/deserialization. This is reflected here: https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/kore/ADT.scala#L88

tothtamas28:
Fixes #2602


ehildenb (approved):

pyk/src/pyk/kast.py
```diff
@@ -276,23 +276,23 @@ def match(self, term: KInner) -> Optional[Subst]:
 @dataclass(frozen=True)
 class KToken(KInner):
     token: str
-    sort: str
+    sort: KSort
 
-    def __init__(self, token: str, sort: str):
+    def __init__(self, token: str, sort: KSort):
```
radumereuta: Shouldn't we have an extra constructor that takes `str` and wraps it into a `KSort` in order to minimize the changes and make it easier to instantiate terms?
I like: `KToken('pred1(3) => true', BOOL)`
I don't like: `KToken('...', KSort('K'))`

radumereuta (commented):
Just one question.
I removed the `automerge` label. Feel free to add it back if you don't want to add the extra constructor.

pyk/src/pyk/kast.py
ehildenb: I would like to have this constructor too as well, actually. Especially since we'll be adding parametric sorts later, and we want to make it easier for users to specify non-parametric sorts/productions without having to do this.

ehildenb: Maybe also `KProduction` AST tree should have smart constructors for this too. Like `KNonTerminal` should take in either a `KSort` or a `str`.

tothtamas28: Done.

PR #2610, opened 2022-05-18
Small optimization in OuterParsing markdown

radumereuta:
Probably 100ms faster on some definitions.
```
Outer parsing [65 modules]                                   =  0.592s
```
I tried to measure hot spots in the outer parsing step, but it seems time is being spent a little bit everywhere.


Baltoli (approved):

Baltoli:
Looks good to me

PR #2611, opened 2022-05-19
Add KLabel class to pyk

radumereuta:
to bring it more in line with the java kore classes

### DIFF
--- a/pyk/src/pyk/kast.py
+++ b/pyk/src/pyk/kast.py
@@ -121,7 +121,7 @@ class WithKAtt(KAst, ABC):
 
 
 class KInner(KAst, ABC):
-    _INNER_NODES: Final = {'KVariable', 'KSort', 'KToken', 'KApply', 'KAs', 'KRewrite', 'KSequence'}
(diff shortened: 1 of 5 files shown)


pyk/src/pyk/kast.py
```diff
@@ -311,13 +311,53 @@ def match(self, term: KInner) -> Optional[Subst]:
...
 
+@final
+@dataclass(frozen=True)
+class KLabel(KInner):
+    name: str
+    params: Tuple[KSort, ...]
+
+    def __init__(self, name: str, params: Iterable[KSort] = ()):
+        object.__setattr__(self, 'name', name)
+        object.__setattr__(self, 'params', tuple(params))
```
ehildenb: Can we make this take `params: Iterable[Union[KSort, str]]` instead, so that we can pass in sort parameters as strings?

ehildenb (approved):

pyk/src/pyk/kast.py
tothtamas28: Good idea.

PR #2621, opened 2022-05-25
Fix label of matching logic connectives in pyk

tothtamas28:


pyk/src/pyk/prelude.py
```diff
@@ -60,29 +71,34 @@ def leInt(i1, i2):
...
-def mlEqualsTrue(b):
-    return mlEquals(TRUE, b)
+def mlTop(sort: Union[str, KSort] = K) -> KApply:
+    return KApply(KLabel('#Top', (sort,)))
 
 
-def mlTop():
-    return TOP
+def mlBottom(sort: Union[str, KSort] = K) -> KApply:
+    return KApply(KLabel('#Top', (sort,)))
```
ehildenb: Should we name hardcoded sorts like `Sort_K` or `Sort_GENERATED_TOP_CELL`? To make them more distinct, and to allow also later `Label_GENERATED_TOP_CELL`.

tothtamas28: Yeah, that makes sense. We can also put them in a module `sorts` or in a class `Sorts`, and then have `sorts.K` or `Sorts.K`, repsectively.

ehildenb: Oh that wolud be great. Could we have things like `Sorts.K` and `Labels.EmptyK` defined in `prelude.py`? It seems the most appropriate place for them.

tothtamas28: I agree, will move them there.

tothtamas28: Done.

pyk/src/pyk/kast.py
```diff
@@ -307,8 +300,9 @@ def match(self, term: KInner) -> Optional[Subst]:
         return None
 
 
-TRUE = KToken('true', BOOL)
-FALSE = KToken('false', BOOL)
+TRUE = KToken('true', 'Bool')
+FALSE = KToken('false', 'Bool')
+ktokenDots = KToken('...', 'K')
```
ehildenb, line 306: Are these here instead of `prelude.py` because of technical reasons? They seem a better fit there.

ehildenb: I see that `TRUE` is used in various places here. But the others are not. I wonder if we can avoid using `TRUE` in various places in this file...

pyk/src/pyk/kast.py
```diff
@@ -349,23 +371,50 @@ def map_inner(self: 'KLabel', f: Callable[[KInner], KInner]) -> 'KLabel':
...
+            if bad_arg:
+                raise TypeError(f"KApply() got an unexpected keyword argument '{bad_arg}'")
+            if args:
+                raise TypeError("KApply() got multiple values for argument 'args'")
+            args = kwargs['args']
+
+        elif len(args) == 1 and isinstance(args[0], Iterable) and not isinstance(args[0], KInner):
+            args = args[0]
+
         object.__setattr__(self, 'label', label)
```
ehildenb, line 415: I'm really having trouble trying to quickly understand what functionality this gives us. Can you add a small comment explaining the use case of this constructor?

ehildenb: Also, how does constructor overloading work here? How will python know which `__init__` to call?

pyk/src/pyk/tests/test_subst.py
```diff
@@ -108,15 +99,18 @@ def test_unapply(self):
...
         (mlEquals(a, b), {}, mlEquals(a, b)),
         (mlEquals(x, a), {'x': a}, mlTop()),
-        (mlEquals(x, t), {}, mlEquals(x, t)),
+        (mlEquals(x, _0), {}, mlEquals(x, _0)),
         (mlEquals(x, y), {}, mlEquals(x, y)),
         (mlAnd([mlEquals(a, b), mlEquals(x, a)]), {'x': a}, mlEquals(a, b)),
-        (mlEqualsTrue(int_eq(a, b)), {}, mlEqualsTrue(int_eq(a, b))),
-        (mlEqualsTrue(int_eq(x, a)), {'x': a}, mlTop()),
+        (mlEqualsTrue(_EQ(a, b)), {}, mlEqualsTrue(_EQ(a, b))),
+        (mlEqualsTrue(_EQ(x, a)), {'x': a}, mlTop()),
```
ehildenb, line 113: Oh this is really nice (the `_EQ(a, b)`).

pyk/src/pyk/kast.py
ehildenb: Made an issue: https://github.com/runtimeverification/k/issues/2630

ehildenb (approved):
Just a couple questions, feel free to add automerge.

pyk/src/pyk/kast.py
tothtamas28: There is a single method with actual implementation, that's the one that's going to be invoked during execution. Declarations with the `@overload` decorator only provide type hints for the type checker.

In this case, the `@overload` declarations express that `KApply(f, x, y, z)` and `KApply(f, [x, y, z])` should type check, and the implementation makes sure that these invocations are handled correctly (I added tests to check this).

PR #2624, opened 2022-05-25
[pyk] Proof rule-profiling, several renames and simplifications

### LINKED ISSUE #2536: Add kompile feature to pyk.ktool

ehildenb:
This PR:

- Adds enough functionality to the python `KProve` class that KEVM can use it directly instead of bash script.
- Adds rule profiling to the `KProve` class in pyk, so that `kevm ...` can easily call it to generate rule profiles.
- Renames `unsafeMlPredToBool` to `ml_pred_to_bool` and makes it safe by default (can get unsafe behavior as needed with `unsafe=True`). Refactors and simplifies this function to handle allowable cases directly, and fail horribly if something unsafe is done. Adds several tests of cases that came up when trying the summarizer with this branch.
- Renames `boolToMlPred` to `bool_to_ml_pred`.
- Renames `buildAssoc` to `build_assoc`, and makes it handle the `KLabel/str` more gracefully.
- Remove many unused functions from `kastManip.py` (checked here, KEVM, and ERC20 repo): `whereMatchingBottomUp`, `getOccurances`, `count_rhs_vars`, `drop_var_prefixes`, `drop_ques`, `drop_unds`, `structurallyFrameKCell`, `applyCellSubst`, `hasExistentials`, `onCells`, and `abstractCell`. `structurallyFrameKCell` had a test for it, but the use it had downstream has since been removed. `abstractCell` has been moved downstream, because I don't like it and don't want to encourage it.
- Specializes all uses of `build_assoc` that can be to `mlAnd` or `mlOr`, and adds needed sorts.
- Renames `propagateUpConstraints => propagate_up_constraints`

These changes make it possible to:

- Use `pyk` for `kevm kompile ...` and `kevm prove ...` across the haskell backend test-suite: https://github.com/runtimeverification/evm-semantics/pull/1258
- Move all the logic of `ktoken summary-init ...` into `pyk` instead of bash (PR to come, work is local).

### DIFF
--- a/pyk/requirements/base.txt
+++ b/pyk/requirements/base.txt
@@ -1,1 +1,3 @@
 graphviz==0.19.1
+tabulate==0.8.6
+types-tabulate==0.8.6
--- a/pyk/setup.cfg
+++ b/pyk/setup.cfg
@@ -9,6 +9,8 @@ package_dir =
     =src
 install_requires =
     graphviz == 0.19.1
+    tabulate == 0.8.6
+    types-tabulate == 0.8.6
 
 [options.package_data]
 * = py.typed
--- a/pyk/src/pyk/__main__.py
+++ b/pyk/src/pyk/__main__.py
@@ -11,7 +11,7 @@ from .kast import KApply, KAst, KLabel, flattenLabel, readKastTerm
 from .kastManip import (
     minimize_term,
     minimizeRule,
-    propagateUpConstraints,
+    propagate_up_constraints,
     removeSourceMap,
     splitConfigAndConstraints,
 )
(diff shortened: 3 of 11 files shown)


pyk/requirements/base.txt
```diff
@@ -1 +1,3 @@
 graphviz==0.19.1
+tabulate==0.8.6
+types-tabulate==0.8.6
```
tothtamas28, line 3: This is probably a dev dependency.

tothtamas28 (approved):

pyk/requirements/base.txt
ehildenb: I'm not sure, I need it downstream in KEVM.

PR #2642, opened 2022-06-06
JSON post process - re added

radumereuta:
Adding back the changes from #2581, w/o the commit that broke C-semantics


radumereuta:
The C semantics doesn't crash anymore, but it still fails with the post-processing step because of #2671

dwightguth (commented):
I would still like someone else to review this who is more familiar with the json and python code, but this seems fine from my perspective.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -150,6 +156,11 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
         // final check for sort correctness
         for (Module m : mutable(kompiledDefinition.modules()))
             m.checkSorts();
+        if (kompileOptions.postProcess != null) {
+            kompiledDefinition = postProcessJSON(kompiledDefinition, kompileOptions.postProcess);
+            files.saveToKompiled("jsoned.txt", kompiledDefinition.toString());
```
ehildenb: I would call this `post-processed.txt` instead.

ehildenb (approved):

PR #2646, opened 2022-06-07
Call `lldb` instead of `gdb` in `krun --debugger` on macOS

goodlyrottenapple:
This replicates the functionality of `krun-legacy` which calls `lldb` on macOS instead of `gdb`

### DIFF
--- a/k-distribution/k-tutorial/1_basic/19_debugging/README.md
+++ b/k-distribution/k-tutorial/1_basic/19_debugging/README.md
@@ -69,6 +69,8 @@ command listed in this output message and add the corresponding
 Please note that the path will be different on your machine than the one
 listed above. 
 
+> *Note* The `krun --debugger` command will launch the `lldb` debugger on macOS instead of `gdb`. See https://lldb.llvm.org/use/map.html for a mapping between `gdb` and `lldb` commands. 
+
 ## Basic commands
 
 The most basic commands you can execute in the K GDB session are to run your
--- a/k-distribution/src/main/scripts/bin/krun
+++ b/k-distribution/src/main/scripts/bin/krun
@@ -31,6 +31,12 @@ dryRun=false
 expandMacros=true
 filterSubst=
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  DBG_CMD="lldb -- "
+else
+  DBG_CMD="gdb --args "
+fi
+
 # setup temp files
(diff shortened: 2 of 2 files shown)


radumereuta (changes requested):
I think we should add a warning here https://github.com/runtimeverification/k/tree/master/k-distribution/k-tutorial/1_basic/19_debugging
saying that OSX uses lldb and the commands might be different.
No need to go into more details than this.

Baltoli:
`lldb` and `gdb` have different python scripting interfaces, so our [existing integration with the LLVM backend](https://github.com/runtimeverification/llvm-backend/blob/master/gdb/kllvm.py) isn't going to work at all on LLDB.

If we want to improve the state of support for K on macOS in general, it might well be worth porting those scripts over to LLDB's API. This is particularly relevant given that GDB support is basically non-existent on macOS / arm64.

radumereuta (approved):

Baltoli:
I think the text in the README should highlight explicitly that the K support _won't work_, rather than just that the commands will be different

k-distribution/k-tutorial/1_basic/19_debugging/README.md
```diff
@@ -69,6 +69,8 @@ command listed in this output message and add the corresponding
 Please note that the path will be different on your machine than the one
 listed above. 
 
+> *Note* The `krun --debugger` command will launch the `lldb` debugger on macOS instead of `gdb`. See https://lldb.llvm.org/use/map.html for a mapping between `gdb` and `lldb` commands.
```
Baltoli, line 72: I'd mention it here

Baltoli (changes requested):

k-distribution/k-tutorial/1_basic/19_debugging/README.md
goodlyrottenapple: do you want to add the lldb support and leave this open until then?

Baltoli: Yeah I think it's very much worth doing, but I don't know how straightforward the port will be. I'll open an issue

Baltoli: https://github.com/runtimeverification/k/issues/2650

Baltoli:
Will be superseded by https://github.com/runtimeverification/k/pull/2802, and can be closed once that PR is merged.

PR #2647, opened 2022-06-07
Pyk: KCFG stores aliases for nodes

nishantjr:
* Keep a dict: alias -> node_id
* Create a `short_id` function that returns the alias if present, or else the shortened hash.
* `pretty_print` and `to_dot` use `short_id`s instead of `shorten_hash`.


nishantjr:
@ehildenb I went for `short_id` instead of `alias` cause I geel its more descriptive.

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,37 @@ def resolve(node_id: str) -> str:
...
+        for alias, id in (dct.get('aliases') or {}).items():
+            cfg.add_alias(name=alias, id=id)
+
         return cfg
 
+    def short_id(self, node: Node) -> str:
+        for alias, hash in self._aliases.items():
+            if node.id == hash:
+                return alias
+        return shorten_hash(node.id)
```
ehildenb: This just returns the first alias it finds? What if the user defines an alias, and we have a short-id for the node too? We should probably allow it to keep shortening the aliases?

Also, what should it do if there are two aliases for the same node?

pyk/src/pyk/tests/test_kcfg.py
```diff
@@ -276,46 +335,47 @@ def test_pretty_print(self):
...
+            'covers': cover_dicts((8, 11)),                  # Loops back
         }
         cfg = KCFG.from_dict(d)
 
-        print(set(map(lambda node: node.id, cfg.reachable_nodes(nid(5), reverse=True, traverse_covers=True))))
+        def _short_id(i) -> str:
+            return cfg.short_id(node(i))
 
         # TODO: Why are all nodes (besides the target) frontiers?
         # TODO: Add a cover
```
ehildenb: If a node is interior, but not `expanded` or `covered`, it will be `frontier`.

You need to add `expanded` to nodes that have already had their successors computed.

pyk/src/pyk/kcfg.py
nishantjr: @ehildenb 

> What if the user defines an alias, and we have a short-id for the node too? We should probably allow it to keep shortening the aliases?

A short-id is an alias if it has one, otherwise it is the shorten hash. Do you mean we should look for a shorter alias?

> Also, what should it do if there are two aliases for the same node?

I'm not sure. I was thinking we could just disallow that for now.

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,37 @@ def resolve(node_id: str) -> str:
         for verified_ids in dct.get('verified') or []:
             cfg.add_verified(resolve(verified_ids['source']), resolve(verified_ids['target']))
 
+        for alias, id in (dct.get('aliases') or {}).items():
```
```suggestion
        for alias, id in dct.get('aliases', {}).items():
```

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,37 @@ def resolve(node_id: str) -> str:
...
+            cfg.add_alias(name=alias, id=id)
```
```suggestion
            cfg.add_alias(name=alias, id=resolve(id))
```

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,37 @@ def resolve(node_id: str) -> str:
...
+
         return cfg
 
+    def short_id(self, node: Node) -> str:
```
```suggestion
    def alias(self, node_id: str) -> str:
```

pyk/src/pyk/tests/test_kcfg.py
```diff
@@ -263,11 +258,75 @@ def test_paths_between(self):
...
+        self.assertEqual(node(1), cfg.node('d33..d8'))
+        self.assertEqual(node(1), cfg.node(node(1).id))
+
+        # Matches no nodes
+        with self.assertRaisesRegex(ValueError, 'Unknown node: deadbeef\\.\\.\\.d8'):
+            self.assertEqual(node(1), cfg.node('deadbeef...d8'))
+
+        # Bad short hash: Has digits between dots
+        with self.assertRaisesRegex(ValueError, 'Bad short hash: 3\\.c62e73544\\.\\.\\.'):
+            cfg.node('3.c62e73544...')
```
tothtamas28: These are not really `KCFG` tests, so they should be moved to their own module (e.g. `test_hash`).

pyk/src/pyk/tests/test_kcfg.py
```diff
@@ -263,11 +258,75 @@ def test_paths_between(self):
...
+            'edges': edge_dicts((0, 1), (0, 2), (1, 2), (1, 3), (2, 3), (3, 0)),
+        }
+        cfg = KCFG.from_dict(d)
+
+        self.assertEqual(node(1), cfg.node('d33...d8'))
+        self.assertEqual(node(1), cfg.node('d33..d8'))
+        self.assertEqual(node(1), cfg.node(node(1).id))
+
+        # Matches no nodes
+        with self.assertRaisesRegex(ValueError, 'Unknown node: deadbeef\\.\\.\\.d8'):
```
tothtamas28: Since you are checking for equality of the string representation, consider using `assertRaises` for readability.

Also, consider

```suggestion
        with self.assertRaisesRegex(ValueError, r'Unknown node: deadbeef\.\.\.d8'):
```

pyk/src/pyk/kcfg.py
tothtamas28: What do we want to use aliases for?

If they are just for pretty printing, then we should store them as `node_id -> alias`, which ensures fast lookup and that there is at most one alias per node.

But probably we also want to use them like `cfg.node(alias)`. In this case it makes sense to admit multiple aliases, so we also need `alias -> node_id`. The other dict used for pretty printing can then be used to store the default alias, with methods that maintain it in a sensible way.

pyk/src/pyk/utils.py
```diff
@@ -159,7 +163,20 @@ def shorten_hashes(value: Any, leftChars=6, rightChars=6) -> Any:
     return result
 
 
+def deconstruct_short_hash(h: str) -> Tuple[str, str]:
+    x = h.lower()
+    if is_hash(x):
+        return (x, x)
```
tothtamas28, line 169: Shouldn't it be:

```suggestion
        return (x, '')
```

Also, I think we should support short hashes without the separator (e.g. `'abcd'`), interpreted as the hash prefix (e.g. `'abcd..'`).

pyk/src/pyk/utils.py
```diff
@@ -159,7 +163,20 @@ def shorten_hashes(value: Any, leftChars=6, rightChars=6) -> Any:
     return result
 
 
+def deconstruct_short_hash(h: str) -> Tuple[str, str]:
```
tothtamas28, line 166: Let's choose between `..` and `...` as separator.

I think the shorter the better.

tothtamas28 (changes requested):

pyk/src/pyk/utils.py
nishantjr: > Shouldn't it be: `(x, '')`

No, the first argument is what we want the prefix, and the second is what we want the suffix to match.
In the case when the full hash is supplied, but the left and the right must match. I guess it doesn't make a difference unless we're comparing with a shorthash thought.

> Also, I think we should support short hashes without the separator (e.g. 'abcd'), interpreted as the hash prefix (e.g. 'abcd..').

I'm worried about aliases also parsing as hashes though

pyk/src/pyk/utils.py
nishantjr: I agree. I wasn't sure if we wanted to stay compatible.

pyk/src/pyk/kcfg.py
nishantjr: @tothtamas28 this may also return the shortened hash, not just an alias.

pyk/src/pyk/utils.py
tothtamas28: I think we do. What is the advantage of choosing `...` over `..`?

pyk/src/pyk/utils.py
tothtamas28: Good point. But this problem is already present, consider the case when someone provides an alias that is also a long hash. So maybe we should syntactically differentiate hashes from aliases by having a prefix character for one of them (e.g. `@somealias`, `#abcd`, etc).

pyk/src/pyk/kcfg.py
nishantjr: Aliases may also passed to any command that takes an id, so I think the second is better.

I'm going to allow multiple aliases, and display all aliases when pretty printing.

nishantjr: Regarding maintaining a default alias, I'm not sure its worth the engineering effort. We would need an interface for setting the default alias as well.

pyk/src/pyk/utils.py
nishantjr: Thought I saw `...` being used in the web interface, but think I was mistaken.

pyk/src/pyk/kcfg.py
```diff
@@ -537,6 +589,10 @@ def add_verified(self, source_id: str, target_id: str) -> None:
         target_id = self._resolve(target_id)
         self._verified.add((source_id, target_id))
 
+    def add_alias(self, name: str, id: str) -> None:
+        id = self._resolve(id)
+        self._aliases[name] = id
```
```suggestion
    def add_alias(self, node_id: str, alias: str) -> None:
        node_id = self._resolve(node_id)
        self._aliases[alias] = node_id
```

tothtamas28: This method should probably raise an exception if the alias is already present.

pyk/src/pyk/kcfg.py
```diff
@@ -562,6 +618,9 @@ def remove_verified(self, source_id: str, target_id: str) -> None:
             raise ValueError(f'Edge is not verified: {(source_id, target_id)}')
         self._verified.remove((source_id, target_id))
 
+    def remove_alias(self, name: str) -> None:
+        self._aliases.pop(name)
```
```suggestion
    def remove_alias(self, alias: str) -> None:
        self._aliases.pop(alias)
```

tothtamas28: This method should probably raise an exception if the alias is not present.

pyk/src/pyk/kcfg.py
```diff
@@ -419,6 +467,10 @@ def remove_node(self, node_id: str) -> None:
         self._expanded.discard(node_id)
         self._verified = set((source_id, target_id) for source_id, target_id in self._verified if source_id != node_id and target_id != node_id)
 
+        del_aliases = [alias for alias, id in self._aliases.items() if id == node_id]
+        for a in del_aliases:
+            del self._aliases[a]
```
```suggestion
        del_aliases = [alias for alias, id in self._aliases.items() if id == node_id]
        for alias in del_aliases:
            self.remove_alias(alias)
```

pyk/src/pyk/kcfg.py
```diff
@@ -322,7 +366,7 @@ def _short_label(label):
         for node in self.nodes:
             nodeAttrs = self.node_attrs(node.id)
             classAttrs = ' '.join(nodeAttrs)
-            label = shorten_hashes(node.id) + (classAttrs and ' ' + classAttrs)
+            label = self.short_id(node) + (classAttrs and ' ' + classAttrs)
```
tothtamas28: The label should be based on `node_short_info` if we want to display all aliases.

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,42 @@ def resolve(node_id: str) -> str:
...
+            result = {}
+            for (k, v) in value.items():
+                result[self.short_ids(k)] = self.short_ids(v)
+        elif type(value) is set:
+            result = set()
+            for item in value:
+                result.add(self.short_ids(item))
+        else:
+            assert(False)
+        return result
```
tothtamas28: I think this function can be removed.

pyk/src/pyk/kcfg.py
```diff
@@ -253,8 +257,42 @@ def resolve(node_id: str) -> str:
...
+    def aliases(self, node_id: str) -> List[str]:
+        node_id = self._resolve(node_id)
+        return [alias for alias, value in self._aliases.items() if node_id == value]
+
+    def short_id(self, node: Node) -> str:
+        """ Return a user friendly name for a node. This may be either an alias or a shortened hash. """
+        for alias, hash in self._aliases.items():
+            if node.id == hash:
+                return '@' + alias
+        return shorten_hash(node.id)
```
tothtamas28: Do we need this function? I think displaying one alias without having an explicit default alias might be confusing.

pyk/src/pyk/kcfg.py
```diff
@@ -197,6 +199,7 @@ def to_dict(self) -> Dict[str, Any]:
         target = sorted(self._target)
         expanded = sorted(self._expanded)
         verified = [{"source": source_id, "target": target_id} for source_id, target_id in sorted(self._verified)]
+        aliases = self._aliases
```
```suggestion
        aliases = dict(self._aliases)
```

pyk/src/pyk/kcfg.py
```diff
@@ -351,21 +395,25 @@ def _short_label(label):
 
         return graph.source
 
-    def _resolve_all(self, short_id: str) -> List[str]:
-        return [node_id for node_id in self._nodes if compare_short_hashes(short_id, node_id)]
+    def _resolve_all(self, id_like: str) -> List[str]:
```
tothtamas28: I'd call this `_resolve_hash`, and keep its old behavior.

Then `_resolve_or_none` can be implemented as follows.
* If `id_like` is a hash, then delegate to `_resolve_hash`, and raise an exception if there are multiple results.
* Otherwise, we have an alias. If it starts with `@`, drop the first character.
* Try to resolve the alias.

As a result, `@` is optional, and can be used to enforce strings to be resolved as aliases (as something starting with `@` is never a hash). So for example
* `abcd` is resolved as a hash
* `@abcd` is resolved as alias `abcd`
* `hello` is resolved as alias `hello`
* `@hello` is resolved as alias `hello`

Optionally, we can also refine `_resolve` to give a more specific error message if resolution fails:
* If `id_like` is a hash: `Unknown node hash: ...`
* Otherwise: `Unknown alias: ...`

What do you think?

pyk/src/pyk/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
+from unittest import TestCase
+
+from ..utils import deconstruct_short_hash
+
+
+class KCFGTestCase(TestCase):
```
tothtamas28:
```suggestion
class ShortHashTest(TestCase):
```

tothtamas28: Please add a few tests for successful deconstruction (prefix with / without the dots, suffix, both prefix and suffix, full hash, etc).

tothtamas28 (changes requested):

pyk/src/pyk/kcfg.py
tothtamas28: > I'm going to allow multiple aliases, and display all aliases when pretty printing.

In this case a default alias is not necessary.

pyk/src/pyk/kcfg.py
nishantjr: `remove_alias` doesn't work because the node has already been removed from the CFG. `_resolve` throws.

pyk/src/pyk/kcfg.py
nishantjr: Any particular reason for this? `_aliases` is already a dict.

pyk/src/pyk/kcfg.py
nishantjr: I like the refactoring. But wasn't the whole point of adding the `@` to avoid confusion between the two?

pyk/src/pyk/kcfg.py
tothtamas28: `_aliases` is a mutable collection, so in order protect class invariants we should return a copy and not the original reference.

pyk/src/pyk/kcfg.py
tothtamas28: > But wasn't the whole point of adding the @ to avoid confusion between the two?

My point is that if `id_like` is not a valid hash, we can omit the `@`, because we know we have an alias. But this is just a small usability improvement.

In any case, the input should be validated to be either a hash or an alias (if `@` is optional, the latter does not have to be checked).

pyk/src/pyk/kcfg.py
tothtamas28: Are you sure? `remove_alias` doesn't seem to invoke `_resolve`.

tothtamas28 (approved):

pyk/src/pyk/kcfg.py
nishantjr: Makes sense, thanks.

PR #2649, opened 2022-06-08
A packaging error is occurring during builds due to write permissions 

F-WRunTime:
[Tempfile ](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) allows for creating a tempfile with mkdtemp()

### DIFF
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -3,7 +3,7 @@ import logging
 import os
 from pathlib import Path
 from subprocess import CalledProcessError, CompletedProcess
-from tempfile import mkdtemp
+from tempfile import TemporaryDirectory
 from typing import Dict, Final, Iterable, List, Mapping, Optional, Tuple
 
 from tabulate import tabulate
@@ -82,7 +82,11 @@ class KProve(KPrint):
     def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
-        self.use_directory = (mkdtemp()) if use_directory is None else Path(use_directory)
+        if not use_directory:
+            self._temp_dir = TemporaryDirectory()
+            self.use_directory = self._temp_dir.name
+        else: 
+            self.use_directory = Path(use_directory)
         # TODO: we should not have to supply main_file_name, it should be read
         self.main_file_name = main_file_name
         self.prover = ['kprove']
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -82,7 +82,7 @@ class KProve(KPrint):
     def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
-        self.use_directory = ( mkdtemp() ) if use_directory is None else Path(use_directory)
+        self.use_directory = (mkdtemp()) if use_directory is None else Path(use_directory)
         # TODO: we should not have to supply main_file_name, it should be read
         self.main_file_name = main_file_name
         self.prover = ['kprove']
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -82,7 +82,7 @@ class KProve(KPrint):
     def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
-        self.use_directory = (mkdtemp() / 'kprove') if use_directory is None else Path(use_directory)
+        self.use_directory = ( mkdtemp() ) if use_directory is None else Path(use_directory)
         # TODO: we should not have to supply main_file_name, it should be read
         self.main_file_name = main_file_name
         self.prover = ['kprove']
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 from pathlib import Path
+from tempfile import mkdtemp
 from subprocess import CalledProcessError, CompletedProcess
 from typing import Dict, Final, Iterable, List, Mapping, Optional, Tuple
 
@@ -78,7 +79,7 @@ def _kprove(spec_file: str, *args: str) -> CompletedProcess:
 
 class KProve(KPrint):
 
-    def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
+    def __init__(self, kompiled_directory, main_file_name=None, use_directory=mkdtemp()):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
         self.use_directory = (self.directory / 'kprove') if use_directory is None else Path(use_directory)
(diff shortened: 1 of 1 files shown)


F-WRunTime:
This issue presented itself downstream in kevm-semantics during the packaging phase. Issue can be seen here: 
https://ci.runtimeverification.com/jenkins/job/KEVM/job/master/171/ 
Builds 167 also presented this issue as well.

pyk/src/pyk/ktool/kprove.py
```diff
@@ -78,7 +79,7 @@ def _kprove(spec_file: str, *args: str) -> CompletedProcess:
 
 class KProve(KPrint):
 
-    def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
+    def __init__(self, kompiled_directory, main_file_name=None, use_directory=mkdtemp()):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
         self.use_directory = (self.directory / 'kprove') if use_directory is None else Path(use_directory)
```
nishantjr: I think there's a problem with using `mktemp()` as an default argument in this way.
It would reuse the same `mktemp()` directory for every instance of `KProve` we create. See [1].
This wouldn't be a big deal in thsi case because in most cases we would only create a single KProve instance in any program.

[1]: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

nishantjr (changes requested):
I'd like to postpone creating the directory until we know we need it.

1. Revert the signature to `use_directory=None`.
2. In the `__init__` function only store the `use_directory` as a private parameter
3. create a function/property `use_directory()` that creates a temp dir, if `_use_directory` is `None`.
4. I think if a user passes in a non-default parameter it should be the user's responsibility to create the directory. So lets drop the call to `mkdir` on the next line.

pyk/src/pyk/ktool/kprove.py
F-WRunTime: I'm not too familiar with all the options but specifically this problems had occurred from this call out of [evm-semantics ](https://github.com/runtimeverification/evm-semantics/blob/master/kevm_pyk/src/kevm_pyk/kevm.py#L94) and would test the default scenario. If there is a way to specifically run this change within that repository we can try that?

F-WRunTime: Curious would following still work without the need of a private var and method for handling if default was None? 
If I understand init() should still get executed on each object instantiation of KProve. 
`        self.use_directory = (mkdtemp() / 'kprove') if use_directory is None else Path(use_directory)
        self.use_directory.mkdir(parents=True, exist_ok=True)
`
mkdtemp() will handle the folder creation process for us and pass back the full path. 
And to your point in 4. we can drop mkdir and require the user to create the directory. Could also provide a 'try/except' around line 84/85 to provide an error that folder must be provided ?

nishantjr: > I'm not too familiar with all the options but specifically this problems had occurred from this call out of [evm-semantics ](https://github.com/runtimeverification/evm-semantics/blob/master/kevm_pyk/src/kevm_pyk/kevm.py#L94) and would test the default scenario. If there is a way to specifically run this change within that repository we can try that?

You can clone EVM semantics, and follow the instructions here to [build its dependencies.](https://github.com/runtimeverification/evm-semantics/blob/master/README.md#build-dependencies)
Make sure to update the KSubmodule before runing `make deps`.
The test that's failing is on [this line of the Jenkinsfile](https://github.com/runtimeverification/evm-semantics/blob/master/Jenkinsfile#L27).
That test would pass even without your PR right now, because the failure is caused by KEVM being packaged readonly.
This doesn't happen during the normal testing unfortunately, so lets manually make the EVM semantics kompiled directory (`find -name '*-kompiled'`) readonly before running the test to force the failure.

nishantjr: > Curious would following still work without the need of a private var and method for handling if default was None?

`mkdtemp()` itself creates the temp directory. Its created whether or not we use `mkdir`. I'd rather not have that happen if we're not using it. We don't need the `kprove` subdir under it either.

pyk/src/pyk/ktool/kprove.py
```diff
@@ -81,8 +82,7 @@ class KProve(KPrint):
     def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
         super(KProve, self).__init__(kompiled_directory)
         self.directory = Path(self.kompiled_directory).parent
-        self.use_directory = (self.directory / 'kprove') if use_directory is None else Path(use_directory)
-        self.use_directory.mkdir(parents=True, exist_ok=True)
+        self.use_directory = (mkdtemp()) if use_directory is None else Path(use_directory)
```
nishantjr: @F-WRunTime Lets run `mkdtemp` only when we need the directory because it creates a temp directory.

nishantjr (changes requested):

pyk/src/pyk/ktool/kprove.py
nishantjr: Something like this:>

```
     def __init__(self, kompiled_directory, main_file_name=None, use_directory=None): 
           ...
        self._use_directory = use_directory

    def use_directory(self) -> str:
        if not self._use_directory:
            self._use_directory = mkdtemp(...)
        return self._use_directory
```

I don't like the idea of IO happening in the initializer because that's not what users expect and it may cause issues like this one. It would also leave the system littered with unused temp directories otherwise.

ehildenb: I don't think it shuold be a problem? @tothtamas28 is there a standard Python programming convention we can lean on here? Is it usually considered bad to have IO in the initializer?

I think `tempfile` can be set to also make sure it delete the directory on cleanup time? So we can just use that.

We should also be logging the temp file name though: `_LOGGER.info('Using temporary directory: {self.use_directory}')` or something, when we create it.

ehildenb: At this point, let's just get this merged though.

ehildenb: At least with some logging.

nishantjr: Unfortunately, `mkdtemp` doesn't do the deletion. 

> The user of [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) is responsible for deleting the temporary directory and its contents when done with it.

https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

But, anyway, lets get the packaging working again.

F-WRunTime: Thank you.  I see what you mean. There's an option as well to use tempfile.TemporaryDirectory(). 
and tempfile.TemporaryDirectory().name will give us the directory. 
Auto cleanup of the directory occurs on garbage collection. 

Does data need to persist post execution?  Would that be of use in the over arching use? Proposed final solution. Please note I'm still working on setting an environment, without a machine to run the tests. I'm going to use a remote box for now. Here's the proposed change I'll push in a bit and I will test changes before merge. This will be my goal for the day.

nishantjr: @F-WRunTime Acually Everett realised that [`TempoararyDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) works better than `mkdtemp` and automatically deletes the directory for us.
If we use that instead of `mkdtemp`, its OK to create the directory in `__init__`.

Lets keep the type of `KProve` to `Path`, so something like:

```
def __init__(...)
    ...
    if not use_directory:
        self._temp_dir = TemporaryDirectory()
        self.use_directory = self._temp_dir.name
```

F-WRunTime: Alright sounds good. Updated and test are running on CI. Working up env setup for evm-semantics tests unless someone else has it setup we can skip waiting on me while I'm working through this exercise.

nishantjr (approved):
LGTM!

F-WRunTime:
Thanks you.  I see what you mean. There's an option as well to use tempfile
in a way to auto clean up folders it creates when the object is destroyed
or the process exits.

 Would that be of use in the over arching use? Or does data need to persist
post execution?

PR #2652, opened 2022-06-09
Add extra options for kompiling to haskell

radumereuta:
--haskell-backend-command
--no-haskell-binary


ana-pantilie (approved):
I tested it and it seems to be working. Thank you!

PR #2655, opened 2022-06-10
Relax simplification rules restriction in kprove

### LINKED ISSUE #2654: [Bug] [kprovex] - Not allowing simplification, anywhere rule in spec module
K Version
---------

```
$ kompile --version
K version:    5.3.0
Build date:   Thu Jun 09 21:31:49 UTC 2022
```

Description
-----------

A `simplification, anywhere` rule is not being allowed in the `--spec-module`. It should be allowed so that the user does not have to call `kompile` again to try out this rule.

Input Files
-----------

File `test.k`:

```
module TEST
    imports MAP
    imports INT
    imports BOOL

    syntax Pgm ::= "a" | "b"

    configuration <k> $PGM:Pgm </k> <mem> .Map </mem>

    rule <k> a => b </k> <mem> MEM => MEM [ 3 <- 5 ] </mem>

endmodule

module TEST-SPEC
    imports TEST

    rule [unif-map]: { MAP [ K <- V ] #Equals MAP [ K <- V' ] } => { V #Equals V' } [simplification, anywhere]

    claim <k> a => b </k> <mem> MEM => MEM [ 3 <- ?C ] </mem>
      ensures ?C <=Int 5 andBool 5 <=Int ?C

endmodule
```

Reproduction Steps
------------------

```
+ kompile --backend haskell test.k --main-module TEST --syntax-module TEST
+ kprove test.k --spec-module TEST-SPEC
[Error] Compiler: Simplification rules need to be function/functional like.
        Location(17,22,17,84)
        17 |        rule [unif-map]: { MAP [ K <- V ] #Equals MAP [ K <- V' ] } => { V
#Equals V' } [simplification, anywhere]
           .                        
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Expected Behavior
-----------------

The rule should be allowed, and the spec should pass. If the rule `unif-map` is moved to module `TEST` instead, then we get `#Top` as output.

radumereuta:
Fixes #2654


PetarMax (approved):
This also fixes the bug in the 0-to-K tutorial with `simple-spec.k` that originated the issue.

PR #2659, opened 2022-06-15
New simplification rules

dwightguth:
This PR adds explicitly to the K prelude a number of simplification rules that were implicitly part of the existing implementation of unification in the Haskell backend. It also adds the `injective` attribute to the map item and set item constructors so that they will be decomposed correctly by the new unifier.


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -440,6 +440,11 @@ module MAP-KORE-SYMBOLIC [kore,symbolic]
   rule K1 in_keys(M [ K2 <- _ ]) => true requires K1 ==K K2 orBool K1 in_keys(M) [simplification]
   rule K1 in_keys(M [ K2 <- _ ]) => K1 in_keys(M) requires K1 =/=K K2 [simplification]
 
+  rule {false #Equals Key in_keys(.Map)} => #Ceil(Key) [simplification]
+  rule {Key in_keys(.Map) #Equals false} => #Ceil(Key) [simplification]
+  rule {false #Equals Key in_keys(Key' |-> Val @M)} => #Ceil(Key) #And #Ceil(Key' |-> Val @M) #And #Not({Key #Equals Key'}) #And {false #Equals Key in_keys(@M)} [simplification]
+  rule {Key in_keys(Key' |-> Val @M) #Equals false} => #Ceil(Key) #And #Ceil(Key' |-> Val @M) #And #Not({Key #Equals Key'}) #And {Key in_keys(@M) #Equals false} [simplification]
```
ana-pantilie: In order for that `#Ceil` to not be trivial, `Key` should be a set var.

```suggestion
  rule {false #Equals @Key in_keys(.Map)} => #Ceil(@Key) [simplification]
  rule {@Key in_keys(.Map) #Equals false} => #Ceil(@Key) [simplification]
  rule {false #Equals @Key in_keys(Key' |-> Val @M)} => #Ceil(@Key) #And #Ceil(Key' |-> Val @M) #And #Not({@Key #Equals Key'}) #And {false #Equals @Key in_keys(@M)} [simplification]
  rule {@Key in_keys(Key' |-> Val @M) #Equals false} => #Ceil(@Key) #And #Ceil(Key' |-> Val @M) #And #Not({@Key #Equals Key'}) #And {@Key in_keys(@M) #Equals false} [simplification]
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -860,6 +865,25 @@ operations listed above.
...
+  rule {true #Equals notBool @B} => {false #Equals @B} [simplification]
+  rule {notBool @B #Equals true} => {@B #Equals false} [simplification]
+  rule {false #Equals notBool @B} => {true #Equals @B} [simplification]
+  rule {notBool @B #Equals false} => {@B #Equals true} [simplification]
+
+  rule {true #Equals @B1 andBool @B2} => {true #Equals @B1} #And {true #Equals @B2} [simplification]
+  rule {@B1 andBool @B2 #Equals true} => {@B1 #Equals true} #And {@B2 #Equals true} [simplification]
+  rule {false #Equals @B1 orBool @B2} => {false #Equals @B1} #And {false #Equals @B2} [simplification]
+  rule {@B1 orBool @B2 #Equals false} => {@B1 #Equals false} #And {@B2 #Equals false} [simplification]
+endmodule
```
ana-pantilie, line 881: Why use set variables here?

dwightguth: I don't remember the exact details here except that an element variable wasn't matching some term that ought to be simplifiable.

ana-pantilie (approved):

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -254,7 +254,7 @@ An element of a `Map` is constructed via the `|->` operator. The key is on the
 left and the value is on the right.
 
 ```k
-  syntax Map ::= KItem "|->" KItem                      [function, functional, hook(MAP.element), klabel(_|->_), symbol, latex({#1}\mapsto{#2})]
+  syntax Map ::= KItem "|->" KItem                      [function, functional, hook(MAP.element), klabel(_|->_), symbol, latex({#1}\mapsto{#2}), injective]
```
ehildenb, line 257: Huh, why is this a `function` instead of a constructor?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -533,7 +538,7 @@ The set with zero elements is represented by `.Set`.
 An element of a `Set` is constructed via the `SetItem` operator.
 
 ```k
-  syntax Set ::= SetItem(KItem)               [function, functional, hook(SET.element), klabel(SetItem), symbol]
+  syntax Set ::= SetItem(KItem)               [function, functional, hook(SET.element), klabel(SetItem), symbol, injective]
```
ehildenb, line 541: Why is this a `function` instead of a constructor? We would get injectivity for free if it was a constructor.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1089,11 +1113,20 @@ module INT-KAST [kast]
...
-  rule I1:Int ==K I2:Int => I1 ==Int I2
+  rule I1:Int ==K I2:Int => I1 ==Int I2 [simplification]
+  rule {K1 ==Int K2 #Equals true} => {K1 #Equals K2} [simplification]
+  rule {true #Equals K1 ==Int K2} => {K1 #Equals K2} [simplification]
+  rule {K1 ==Int K2 #Equals false} => #Not({K1 #Equals K2}) [simplification]
+  rule {false #Equals K1 ==Int K2} => #Not({K1 #Equals K2}) [simplification]
+  rule {K1 =/=Int K2 #Equals true} => #Not({K1 #Equals K2}) [simplification]
+  rule {true #Equals K1 =/=Int K2} => #Not({K1 #Equals K2}) [simplification]
+  rule {K1 =/=Int K2 #Equals false} => {K1 #Equals K2} [simplification]
+  rule {false #Equals K1 =/=Int K2} => {K1 #Equals K2} [simplification]
```
ehildenb, line 1129: It seems to me we could get more mileage out of these rules by lifting all of them to `==K` instead, and then reversing the specialization of `==K => ==Int` to instead generalize `==Int => ==K`. Then we could have one set of equality reasoning rules that works for all of `==Int`, `==Bool`, and `==String`.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -2060,11 +2093,19 @@ module K-EQUAL-SYNTAX
...
-  rule K1:Bool ==K K2:Bool => K1 ==Bool K2
+  rule K1:Bool ==K K2:Bool => K1 ==Bool K2 [simplification]
+  rule {K1 ==K K2 #Equals true} => {K1 #Equals K2} [simplification]
+  rule {true #Equals K1 ==K K2} => {K1 #Equals K2} [simplification]
+  rule {K1 ==K K2 #Equals false} => #Not({K1 #Equals K2}) [simplification]
+  rule {false #Equals K1 ==K K2} => #Not({K1 #Equals K2}) [simplification]
+  rule {K1 =/=K K2 #Equals true} => #Not({K1 #Equals K2}) [simplification]
+  rule {true #Equals K1 =/=K K2} => #Not({K1 #Equals K2}) [simplification]
+  rule {K1 =/=K K2 #Equals false} => {K1 #Equals K2} [simplification]
+  rule {false #Equals K1 =/=K K2} => {K1 #Equals K2} [simplification]
```
ehildenb, line 2108: Like these rules become duplicate if we instead generalize `==Bool => ==K` as the above comment suggests.

ehildenb (changes requested):
We need tests. There must have been simplification problems that did not work before that prompted you to add these lemmas? Please add them as `runLemma => doneLemma` style tests of this behavior. You can look also at `map-symbolic-tests` for how it tests _negative_ simplification (simplifications that shouldn't happen).

k-distribution/include/kframework/builtin/domains.md
ehildenb: Using set variables here seems kinda uneccary to me here, but also the lemmas seem to still be true with use of set variables. So I suppose it's strictly more general anyway. But I would like to see some tests.

ehildenb: Especially if you could add tests which trigger use of the set variables that normal variables can't handel. The only way I can think to do that is to try simplifying something like `{ true #Equals (true #Or false) }` or `{ #Bottom #Equals false }`.

k-distribution/include/kframework/builtin/domains.md
dwightguth: I mean, it's not really a regular constructor symbol in that the LLVM backend evaluates it to a built-in map with one element. However, the haskell backend, with the new unification algorithm, does need to understand that unifying `K1 |-> V1 ?= K2 |-> V2` is the same as unifying `K1 ?= K2 /\ V1 ?= V2`

k-distribution/include/kframework/builtin/domains.md
dwightguth: The disadvantage of this is we would no longer be able to do SMT reasoning about equalities because we can't generate SMT for `==K` due to its dependence on sort injections to represent equalities.

ehildenb: Hmmm, you mean for a term where `==Int` is embedded somewhere inside the term (like in teh condition of an `#ite`)?

dwightguth:
Okay, this PR now includes the tests that inspired it. There are basically 5 groups of changes here:

1. notBool simplification
2. andBool/orBool simplification
3. ==K simplification
4. ==Int simplification
5. in_keys simplification

`unification_lemmas` tests `==Int` and `andBool`. `unification_lemmas2` tests `==K` and `notBool`. The test for `in_keys` is actually already part of the regression test suite: it is `map-symbolic-tests-haskell/inkeys-9-spec.k`

dwightguth:
`unification_lemmas` also fails if you don't use set variables for `andBool` and `in_keys`.

k-distribution/include/kframework/builtin/domains.md
ehildenb: Well, the LLVM backend can choose to implement it however it likes, it's a hook afterall.

But semantically, this isn't a function is it? And if it is a constructor, we get `injective` for free.

I've made an issue: https://github.com/runtimeverification/k/issues/2683

ehildenb (approved):

k-distribution/include/kframework/builtin/domains.md
dwightguth: I mean, we can encode `I1 ==Int I2` as just `(= I1 I2)` in smt. If we use ==K instead, we cannot encode this in smt because we would have to encode something like `(= (kseq (inj I1) dotk) (kseq (inj I2) dotk))` and smt does not support parametric symbols like `inj`

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -440,6 +440,11 @@ module MAP-KORE-SYMBOLIC [kore,symbolic]
   rule K1 in_keys(M [ K2 <- _ ]) => true requires K1 ==K K2 orBool K1 in_keys(M) [simplification]
   rule K1 in_keys(M [ K2 <- _ ]) => K1 in_keys(M) requires K1 =/=K K2 [simplification]
 
+  rule {false #Equals @Key in_keys(.Map)} => #Ceil(@Key) [simplification]
+  rule {@Key in_keys(.Map) #Equals false} => #Ceil(@Key) [simplification]
+  rule {false #Equals @Key in_keys(Key' |-> Val @M)} => #Ceil(@Key) #And #Ceil(Key' |-> Val @M) #And #Not({@Key #Equals Key'}) #And {false #Equals @Key in_keys(@M)} [simplification]
+  rule {@Key in_keys(Key' |-> Val @M) #Equals false} => #Ceil(@Key) #And #Ceil(Key' |-> Val @M) #And #Not({@Key #Equals Key'}) #And {@Key in_keys(@M) #Equals false} [simplification]
```
ehildenb, line 446: @dwightguth these rules seem a little weird to me. On the LHS, we have `@Key`, and on the RHS we have `#Ceil(@Key)`, which seems to imply that it's not `#Bottom` at the very least. So then why not just use `KEY` instead of `@KEY` I guess....

Also, `#Not({@KEY #Equals KEY'})` seems strange. This allows for valuations where `@KEY` and `KEY'` don't intersect I guess....

Anyway, I think it's still sound. Just wanted to make a note here for posterity.

dwightguth: Originally I did have Key and Ceil both and Ana pointed out that that was redundant and suggested changing it to a set variable. I don't think anything would break if we changed it back to on an element variable and just deleted the ceil though. Do you want me to try it?

PR #2665, opened 2022-06-17
update lesson 15 to discuss start symbol parsing

### LINKED ISSUE #2661: Update k-tutorial to include start symbol defaults
You always get an error when you try to parse a user list because it's not considered as a start symbol.
In lesson 12, if you try to parse `1` you will always have an ambiguity between the single element, and the list with one element.
The start symbol is the sort of $PGM. The default [configuration](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/kast.md#default-configuration) specifies K, which is the top sort (includes everything), except users lists.

nwatson22:
Addressed a small potential issue with the k tutorial. Some of the example programs in lesson 12 weren't parsing inputs like I expected because syntactic lists don't inherit from the K sort during parsing, meaning you can't parse a syntactic list at the top level when the start symbol is left as the default. Alluded to this edge case in lesson 15 where configurations and the start symbol are introduced so other users can find an explanation somewhere in the tutorial if they are also confused about the behavior of those programs.


k-distribution/k-tutorial/1_basic/15_configurations/README.md
```diff
@@ -26,8 +26,12 @@ initialized with the value of the `$PGM` **configuration variable**.
 Configuration variables function as inputs to `krun`. These terms are supplied
 to `krun` in the form of ASTs parsed using a particular module. By default, the
 `$PGM` configuration variable uses the main syntax module of the definition.
+
 The cast on the configuration variable also specifies the sort that is used as
-the entry point to the parser, in this case the `K` sort.
+the entry point to the parser, in this case the `K` sort. In many situations it
+is useful to cast to other sorts there as well. The sort used for the `$PGM`
+variable is is referred to as the start symbol. During parsing, The default
+start symbol `K` subsumes all user-defined sorts except for syntactic lists.
```
radumereuta: You have some typos:
`is is`, `parsing, The default`

Here is an alternative warding. Feel free to mix and match.

Normally, you would choose a sort from the language grammar for better control over the accepted language.
But in this case, `K` is used to specify all user-defined sorts except for syntactic lists.
Syntactic lists are excluded because they will always produce an ambiguity error when parsing a single element
and it would defeat the original purpose of being convenient.

radumereuta (changes requested):

radumereuta (approved):
lgtm

PR #2644, opened 2022-06-07
Add Nix flake derivation

goodlyrottenapple:
This PR adds an alternative way of building K using the (currently) experimental flakes feature, as per https://github.com/runtimeverification/k/issues/2607. The flake is largely independent of the current nix derivation, however since it uses the current `nix/k.nix` file, some tweaks had to be made to make sure this derivation builds both with the current `default.nix` file and the new `flake.nix`. The changes to `nix/k.nix` mostly involve pulling out the `ttuegel` functions for cleaning the git source out of `nix/k.nix` into `default.nix` and passing in the cleaned source as a function argument to `nix/k.nix` directly.

The changes to `shell.nix` and `test.nix` are mainly cosmetic.

Before the new flake derivation can replace the current one in the CI, `tests.nix` has to be ported into the flake.

PR #2666, opened 2022-06-17
Fixing a change causing failure to merge master and release in automation 

F-WRunTime:
### Background:    
Fixing pipeline failure, further info can be seen in [Jenkins build failure](https://ci.runtimeverification.com/jenkins/job/K/job/master/362/)   

TLDR from logs:   

There is an issue in merging master on to release as line 7 of nix/k.nix is modified on each release update by the patch version.sh script.  From what I can tell master needs to provide just the Major/Minor version number i.e. 5.3.0 and the  version.sh script updates the following files with the patch version:    
```bash
    sed --in-place 's/^K_VERSION=.*$/K_VERSION='${version}'/'                                                         install-k
    sed --in-place 's/^    name = "k-.*";$/    name = "k-'${version}'";/'                                             nix/k.nix
    sed --in-place 's/^pkgver=.*$/pkgver='${version}'/'                                                               package/arch/PKGBUILD
    sed --in-place 's/^kframework (.*) unstable; urgency=medium$/kframework ('${version}') unstable; urgency=medium/' package/debian/changelog
    sed --in-place 's/^K_VERSION=.*$/K_VERSION='${version}'/'                                                         src/main/scripts/test-in-container-debian

```
This may also be fixable by manually resolving the merge conflict on release.   

### Testing Merge
Tested using this branch as the latest "master" commit and ran through the Jenkinsfile steps to confirm a successful merge same as is done [here](https://github.com/runtimeverification/k/blob/master/Jenkinsfile#L508)   

Results of a test merge on my local are as follows: 
```bash
$ git branch
  hotfix/Dockerfile-update
  hotfix/nix-release-packaging
  master
* release
$ git merge --no-edit hotfix/nox-release-packaging 
merge: hotfix/nox-release-packaging - not something we can merge
$ git merge --no-edit hotfix/nix-release-packaging 
Auto-merging nix/k.nix
Merge made by the 'ort' strategy.
 default.nix         |  52 ++++++++++----
 flake.lock          | 634 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 flake.nix           | 111 +++++++++++++++++++++++++++++
 nix/k.nix           |  89 +++++++++++-------------
 nix/shell.maven.nix |  12 ++--
 nix/sources.nix     | 227 +++++++++++++++++++++++++++++++++--------------------------
 shell.nix           |  18 ++---
 test.nix            |  70 ++++++++++---------
 8 files changed, 1004 insertions(+), 209 deletions(-)
 create mode 100644 flake.lock
 create mode 100644 flake.nix
```
And looks good. Should avoid a failure to publish the release
No other changes from the previous master commit should have been removed. 
@radumereuta  @goodlyrottenapple  can you please confirm no unintended effects of this change regress your intended changes from #2644 
@ehildenb  Can you please confirm my logic on the release process.

### DIFF
--- a/nix/k.nix
+++ b/nix/k.nix
@@ -4,7 +4,7 @@
 
 let
   unwrapped = mavenix.buildMaven {
-    name = "k-${version}";
+    name = "k-5.3.0";
     infoFile = ./mavenix.lock;
     inherit src;
 


goodlyrottenapple:
I would like to keep the current change,
https://github.com/runtimeverification/k/blob/b1c8430118b5f758b5a2e4804c110d0edaec01ca/nix/k.nix#L7
namely, have nix read out the current version of `package/version`, since I would like to be able to append the current git commit hash to the version name here
https://github.com/runtimeverification/k/blob/b1c8430118b5f758b5a2e4804c110d0edaec01ca/flake.nix#L55-L58
this is of course not strictly necessary, but makes things a bit easier to find in the nix store for debugging purposes. Either way, it seems more nix-y to read the version from the file than to use sed to edit `nix/k.nix`

If we do end up keeping this change, this line will need to be removed
https://github.com/runtimeverification/k/blob/b1c8430118b5f758b5a2e4804c110d0edaec01ca/package/version.sh#L25

nix/k.nix
```diff
@@ -4,7 +4,7 @@
 
 let
   unwrapped = mavenix.buildMaven {
-    name = "k-${version}";
+    name = "k-5.3.0";
```
ehildenb, line 7: @goodlyrottenapple where was this `${version}` identifier being read from in the first place? I see on line 1 that `version` is imported.

I'm skeptical about reading this from the file `package/version`, because that file is populated/touched in the build process I think, or maybe it only stores the major/minor versions but not the patch.

So maybe we should just go back to the previous way of doing it for now? Is there anything wrong with hardcoding the version number here and having `version/version.sh` update it?

F-WRunTime:
@goodlyrottenapple @radumereuta 
Since the root of the issue is ultimately a merge conflict between master and release. We have an alternative solution to support the nix change. 
As you mentioned an update to version.sh -- remove the sed line), and manually resolving the merge conflict between release & master. 

This option is fine with me so long as there are no issues with continuing to support the format of the version as MAJOR.MINOR.PATCH -- Current PATCH is revised by version.sh on the release branch directly and pushed as part of the packaging process. 

Does anyone have concerns with this secondary solution?

goodlyrottenapple:
@F-WRunTime the nix version will be whatever the contents of `package/version` are, which seems to be the same as what the `package/version.sh` script does:

https://github.com/runtimeverification/k/blob/b1c8430118b5f758b5a2e4804c110d0edaec01ca/package/version.sh#L23

ehildenb:
But the fail `package/version` is not correct, it only contains `MAJOR.MINOR.0`, not `MAJOR.MINOR.PATCH`, and `PATCH` is filled in by the release job.

So will this nix derivation see the version before or after the `PATCH` is filled in? It must see the one after.

goodlyrottenapple:
@ehildenb I'm not sure what you mean, if I download the release source here https://github.com/runtimeverification/k/releases/download/v5.3.80/kframework-5.3.80-src.tar.gz both `package/version` and `nix/k.nix` contain the string `5.3.80`.

That means when evaluating, nix would replace `"k-${version}";` with `"k-5.3.80";`

PR #2667, opened 2022-06-20
Check existence of `use_directory` in `KProve`

tothtamas28:
As of https://github.com/runtimeverification/k/pull/2649 the `KProve` constructor argument `use_directory`, if provided, should point to an existing directory. This PR adds a precondition check.


nishantjr (approved):
Looks good

PR #2670, opened 2022-06-20
Test for #Ceil(Set1 Set2)

denis-bogdanas:


Baltoli:
Closing as Denis and Ana are no longer at RV and the issue here appears specific to the old Haskell backend; if we need to reopen this in future we can.

PR #2673, opened 2022-06-21
[nix] Add new flake workflow

goodlyrottenapple:
This PR adds more nix infrastructure improvements including:

* A new github workflow using the new nix flakes infrastructure; we will be phasing out the old nix expressions eventually
* Nix flake now includes running tests via `nix build .#test`
* You can now re-generate `nix/mavenix.lock` via `nix run .#update-maven`
* The build artifacts of the flake expression are significantly smaller now (~5mb vs ~50mb); whilst not essential, this does help with nix's enormous general hunger for storage + reduces amount of stuff we push to the cachix cache (we have a rolling 5gb storage limit, so older stuff gets deleted, which should in general be ok)
* `pyk` is now being packaged and can be used from e.g. `pkgs.python3Packages.pyk` when the `k` overlay gets added to nixpkgs.


k-distribution/tests/regression-new/imp-haskell/Makefile
```diff
@@ -2,7 +2,7 @@ DEF=imp
 EXT=imp
 TESTDIR=.
 KOMPILE_BACKEND=haskell
-KOMPILE_FLAGS=--post-process 'cat' --no-haskell-binary
+KOMPILE_FLAGS=--post-process 'cat'
```
goodlyrottenapple, line 5: This should not be set ad-hoc per test but instead added to the global `KOMPILE_FLAGS+=...` directive in the .mak files

k-distribution/include/kframework/ktest-fail.mak
```diff
@@ -1,9 +1,10 @@
 # path to the current makefile
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 # path to the kompile binary of this distribuition
-KOMPILE=$(abspath $(MAKEFILE_PATH)/../../bin/kompile)
+K_BIN=$(abspath $(MAKEFILE_PATH)/../../bin)
+KOMPILE=${K_BIN}/kompile
 # and kdep
-KDEP=$(abspath $(MAKEFILE_PATH)/../../bin/kdep)
+KDEP=${K_BIN}/kdep
```
goodlyrottenapple, line 7: using `K_BIN` here makes it much easier to run these tests in nix, since we only need to override the `K_BIN` env variable with the nix store path to the K binaries.

flake.nix
```diff
@@ -102,6 +119,47 @@
...
+                patchShebangs tests/regression-new/*
+                substituteInPlace tests/regression-new/append/kparse-twice \
+                  --replace '"$(dirname "$0")/../../../bin/kparse"' '"${k}/bin/kparse"'
+                ${lib.optionalString stdenv.isDarwin ''
+                  for mak in include/kframework/*.mak
+                  do
+                    substituteInPlace $mak \
+                      --replace 'KOMPILE_FLAGS+=--no-exc-wrap' 'KOMPILE_FLAGS+=--no-exc-wrap --no-haskell-binary'
+                  done
+                ''}
```
goodlyrottenapple: we add the `--no-haskell-binary` option to `KOMPILE_FLAGS` for all tests if running on macOS

flake.nix
```diff
@@ -19,25 +19,43 @@
...
                 cp -rv $src/* $out
                 chmod -R u+w $out
                 mkdir -p $out/llvm-backend/src/main/native/llvm-backend/matching
                 cp -rv ${llvm-backend}/matching/* $out/llvm-backend/src/main/native/llvm-backend/matching
+                ${prev.lib.optionalString prev.stdenv.isDarwin ''
+                  substituteInPlace $out/kernel/src/main/java/org/kframework/parser/KRead.java \
+                    --replace 'gcc' 'clang'
+                  substituteInPlace $out/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java \
+                    --replace 'gcc' 'clang'
+                ''}
```
goodlyrottenapple: this is a hack whilst waiting on #2681

flake.nix
```diff
@@ -19,25 +19,43 @@
...
+                }";
               paths = prev.lib.attrValues
                 prev.haskell-backend-stackProject.hsPkgs.kore.components.exes;
             };
-
+            k-version =
+              prev.lib.removeSuffix "\n" (builtins.readFile ./package/version);
             src = prev.stdenv.mkDerivation {
-              name = "llvm-source";
+              name = "k-${k-version}-${self.rev or "dirty"}-src";
```
radumereuta, line 38: Can you give me some examples of what this would look like in different situations?
Why do we have an `or` there?

flake.nix
radumereuta: https://github.com/runtimeverification/k/pull/2686

k-distribution/tests/regression-new/imp-haskell/Makefile
radumereuta: Is this going to fail on the M1 every time?
We still want to test the default behavior.

nix/k.nix
```diff
@@ -4,18 +4,23 @@
...
   unwrapped = mavenix.buildMaven {
-    name = "k-${version}";
+    name = "k-${version}-maven";
     infoFile = ./mavenix.lock;
     inherit src;
 
+    # By default, mavenix copies the jars defined in `submodules` of mavenix.lock to `$out/share/java`.
+    # The following flag disables this, since we copy the jars ourselves.
+    # Otherwise, the jars are needlessly duplicated and bloat the cachix cache.
+    copySubmodules = false;
```
radumereuta, line 14: Nice

nix/k.nix
```diff
@@ -4,18 +4,23 @@
...
+
     # Cannot enable unit tests until a bug is fixed upstream (in Mavenix).
     doCheck = false;
 
     # Add build dependencies
     #
     nativeBuildInputs = [ makeWrapper ];
 
-    buildInputs = [ clang flex gcc git gmp jdk mpfr pkgconfig python3 z3 ];
+    buildInputs = [ flex git gmp jdk mpfr pkgconfig python3 z3 ];
```
radumereuta: Why?

nix/k.nix
```diff
@@ -32,12 +37,26 @@ let
       patchShebangs k-distribution/src/main/scripts/lib
     '';
 
-    # Make sure to link the cmake/ and include/ folders from the llvm-backend source repo and the llvm-backend derivation, 
-    # as these may be expected/required when compiling other projects, e.g. the evm-semantics repo
+    # We first copy the bin, include and lib folders from the build and then replace all copied jars which aready exist
```
radumereuta:
```suggestion
    # We first copy the bin, include and lib folders from the build, and then replace all copied jars which already exist
```

radumereuta (commented):
Mostly questions and typos.

flake.nix
goodlyrottenapple: when nix builds this derivation, it checks if the current folder, which is a git repo has been modified. if not, `self.rev` will point to the current git commit hash, otherwise it is empty so the `or "dirty"` will substitute in "dirty" instead...

nix/k.nix
goodlyrottenapple: these are not needed during the compile step, but rather are runtime dependencies, so its cleaner to remove them.

nix/k.nix
goodlyrottenapple: > Is this going to fail on the M1 every time?
yes I'm afraid so

nix/k.nix
goodlyrottenapple: in fact i suspect most of those except for `jdk` are probably runtime rather than build time dependencies

k-distribution/tests/regression-new/imp-haskell/Makefile
goodlyrottenapple: yes I'm afraid so

nix/k.nix
goodlyrottenapple: yup, removed all the buildInputs, turns out they don't do anything

flake.nix
radumereuta: Please add a comment on why we added the `--no-haskell-binary` option.
We should try to add a fix eventually.
Maybe have an issue on the board.

k-distribution/tests/regression-new/imp-haskell/Makefile
radumereuta, line 5: This means we're not going to test this option on Linux builds, only on the M1.
I think this is fine for now.

radumereuta:
Overall lgtm
When it's ready I still like it if someone else has a look over it. Maybe @F-WRunTime and @calin1304

README.md
```diff
@@ -204,12 +204,74 @@ addition to the usual Maven build setup:
...
+
+If you change any `pom.xml`, you must run 
+
+```
+nix run .#update-maven
+```
+
+and commit the updated `nix/mavenix.lock` file.
+
+## Building with Nix
```
ehildenb: This is the old way to build with Nix? Can we say "not recommended" here to point them at the other one which is recommended?

goodlyrottenapple:
@F-WRunTime any idea why jenkins keeps failing? I can't see any obvious error message on quick glance.

F-WRunTime:
> @F-WRunTime any idea why jenkins keeps failing? I can't see any obvious error message on quick glance.

I'm looking now.

radumereuta:
Should pass after this: https://github.com/runtimeverification/k/pull/2719
I'm waiting on this to see if it passes: https://github.com/runtimeverification/k/pull/2713
If yes, then we need to just merge with master and tests should run fine again.

goodlyrottenapple:
still broken :/

README.md
```diff
@@ -204,12 +204,74 @@ addition to the usual Maven build setup:
...
+trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
+```
+
+To build the K Framework itself, run:
+
+```bash
+nix build .
+```
+
+This will build all of K and put a link to the resultingbinaries in the `result/` folder.
```
calin1304:
```suggestion
This will build all of K and put a link to the resulting binaries in the `result/` folder.
```

flake.nix
```diff
@@ -6,31 +6,47 @@
     nixpkgs.follows = "haskell-backend/nixpkgs";
     llvm-backend.inputs.nixpkgs.follows = "haskell-backend/nixpkgs";
     flake-utils.url = "github:numtide/flake-utils";
-    mavenix.url = "github:nix-community/mavenix";
+    mavenix.url = "github:goodlyrottenapple/mavenix";
```
calin1304, line 9: Why use a fork of `nix-community/mavenix` ?

goodlyrottenapple: See the diff here https://github.com/nix-community/mavenix/compare/master...goodlyrottenapple:mavenix:master
Basically I added some tweaks to `buildMaven` to stop the resulting derivation copying some jars we depend on. Instead we soft link to them from the mavenix generated repo derivation. See here:

https://github.com/runtimeverification/k/blob/1dfda423272bd1ddf9067382e153ace741ce7dcf/nix/k.nix#L39-L46

.github/workflows/test-flake.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          name: runtimeverification
+          extraPullNames: kore
+          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+
+      - name: 'Build K Framework'
+        run: GC_DONT_GC=1 nix build .
+
+      # - name: 'Test K Framework'
+      #   run: GC_DONT_GC=1 nix build .#test
```
Baltoli: Why is this commented? Is it just to reduce CI load while we test the changes?

README.md
```diff
@@ -204,12 +204,74 @@ addition to the usual Maven build setup:
...
+```
+substituters = https://cache.nixos.org https://cache.iog.io
+trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
+```
+
+To build the K Framework itself, run:
+
+```bash
+nix build .
+```
```
Baltoli, line 250: Is it worth mentioning the `GC_DONT_GC` issue here?

flake.nix
Baltoli: If we're going to depend on this change, even if it's a small one we probably want the code to live in the RV org - @F-WRunTime @ehildenb?

flake.nix
```diff
@@ -68,14 +97,32 @@
...
-        inherit overlays;
+
         packages = {
           inherit (pkgs) k;
+          pyk = pkgs.python38Packages.pyk;
+
+          update-maven = pkgs.writeShellScriptBin "update-maven" ''
+            #!/bin/sh
+            ${pkgs.nix}/bin/nix-build --no-out-link -E 'import ./nix/flake-compat-k-unwrapped.nix' \
+              || echo "^~~~ expected error"
```
Baltoli, line 118: What's the error here, and why do we expect it to happen?

Baltoli (changes requested):

.github/workflows/test-flake.yml
goodlyrottenapple: yeah those tests take over an hour on the public runner I think. We either only run them on release or cut down on the tests ran?

README.md
goodlyrottenapple: yeah good idea

flake.nix
goodlyrottenapple: this is hackery I inherited :P here is the comment from the original code, perhaps it should be copied over to the flake
https://github.com/runtimeverification/k/blob/aeecef0c3de3f13d3c33fb42e77cef5ef4e6f940/nix/update-maven.sh#L7-L10

.github/workflows/test-flake.yml
Baltoli: Seems reasonable to me - we could maybe open an issue to define a "core" set of tests that don't take ages to run, and comment / delete this code.

flake.nix
Baltoli: ouch - could probably use a further comment at the call site just to keep it clear. I do something similar in the C semantics (invoke a make target I know will fail, because it creates a directory at some point that I can later `cd` into and run successful commands)

.github/workflows/test-flake.yml
goodlyrottenapple: https://github.com/runtimeverification/k/issues/2726

flake.nix
goodlyrottenapple: added a comment in place

flake.nix
goodlyrottenapple: Added a comment and a link to the issue in haskell-backend

Baltoli (approved):

PR #2674, opened 2022-06-21
Hotfix/release merge conflict

F-WRunTime:
This should resolve merge commit issues with release and update the appropriate script to not modify the nix/k.nix file when building the release packages on the release branch.


F-WRunTime:
There are a lot of commits but I was trying to intertwine the changes between master and release in order to avoid any additional merge issues between the two branches with the current process. I tried testing this locally by taking the branch in the current state and merging it was master locally and did not get any issues. 

The only file change will be to remove the sed command to avoid a failure updating the file. 

Best test to do here would like test locally as well the merge with release. Let me know if you have any questions or suggestions.

radumereuta:
LGTM but @ehildenb was the original developer for the build system so I'll let him give the final approval.

F-WRunTime:
@ehildenb If you're comfortable with the history sequencing I did I think we're good to merge this into master. Let me know if you have any questions.

ehildenb (approved):

PR #2678, opened 2022-06-22
Add sort correctness at the end of the kompile pipeline

### LINKED ISSUE #2671: [kompile] Could not find sorts: [GeneratedTopCell]
```
module TEST-SYNTAX
  imports INT-SYNTAX
  syntax S ::= func(Int) [function]
  syntax S ::= "a"
endmodule

module TEST
  imports TEST-SYNTAX
  imports INT
  configuration <k> $PGM:Int </k> <env> .Map </env>
  rule [[ func(1) => a ]]
       <env> .Map </env>
endmodule

$ kompile test.k --no-exc-wrap --post-process cat

[Error] Compiler: Could not find sorts: [GeneratedTopCell]
	Location(3,16,3,36)
	3 |	  syntax S ::= func(Int) [function]
	  .	               ^~~~~~~~~~~~~~~~~~~~
```
The kompiler generates at runtime a production `syntax S ::= func(Int) GeneratedTopCell [function]` in order to parse
the rule, but that sort isn't being defined anywhere.
This fails only now because I've used the post processing step. Sort checks are done only when a new definition is being created, not when it's being modified.

radumereuta:
Fixes #2671

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkFuncWithContext.k
+++ b/k-distribution/tests/regression-new/checks/checkFuncWithContext.k
@@ -1,0 +1,13 @@
+module CHECKFUNCWITHCONTEXT-SYNTAX
+  imports INT-SYNTAX
+  syntax S ::= func(Int) [function]
+  syntax S ::= "a"
+endmodule
+
+module CHECKFUNCWITHCONTEXT
+  imports CHECKFUNCWITHCONTEXT-SYNTAX
+  imports INT
+  configuration <k> $PGM:Int </k> <env> .Map </env>
+  rule [[ func(1) => a ]]
+       <env> .Map </env>
+endmodule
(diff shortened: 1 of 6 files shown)


dwightguth (changes requested):
Something is very suspicious about this change. GeneratedTopCell is a sort, like Bool, that is always defined in a k definition. I'm not entirely clear on why we would need to add an explicit declaration for it here. It suggests to me that we may have a different problem somewhere else.

Baltoli:
This seems to be fixed if you pull the configuration out into a separate module that the syntax module imports:
```k
module TEST-CONFIGURATION
  imports INT
  imports MAP

  configuration <k> $PGM:Int </k> <env> .Map </env>
endmodule

module TEST-SYNTAX
  imports TEST-CONFIGURATION
  imports INT-SYNTAX

  syntax S ::= func(Int) [function]
  syntax S ::= "a"
endmodule

module TEST
  imports TEST-SYNTAX
  imports INT

  rule [[ func(1) => a ]]
       <env> .Map </env>
endmodule
```

radumereuta:
This only happens when the function is in the syntax module.
We generate the production to parse functions with context in place, but the syntax module doesn't include the configuration yet.
So it's a corner case, but it also means the C-semantics fails this check.

dwightguth (approved):

PR #2682, opened 2022-06-27
Stable ordering when pretty printing, deserializing aliases.

nishantjr:
This fixes non-deterministic tests.


pyk/src/pyk/kcfg.py
```diff
@@ -50,6 +51,11 @@ def id(self) -> str:
         def to_dict(self) -> Dict[str, Any]:
             return {'id': self.id, 'term': self.cterm.term.to_dict()}
 
+        def __lt__(self, other):
+            if not isinstance(other, KCFG.Node):
+                return NotImplemented
+            return self.id < other.id
+
```
ehildenb, line 57: Can't this be enforced with our type-checker?

```suggestion
        def __lt__(self, other: KCFG.Node) -> bool:
            return self.id < other.id

```

pyk/src/pyk/kcfg.py
```diff
@@ -58,6 +64,11 @@ class EdgeLike(ABC):
         def pretty_print(self, kprint: KPrint) -> List[str]:
             assert False, 'Must be overridden'
 
+        def __lt__(self, other):
+            if not isinstance(other, KCFG.EdgeLike):
+                return NotImplemented
+            return (self.source, self.target) < (other.source, other.target)
+
```
ehildenb, line 70: Same here.

pyk/src/pyk/kcfg.py
nishantjr: That's a `return`, not `raises`. I think python will call the other classes implementation to see if that works.

pyk/src/pyk/kcfg.py
```diff
@@ -1,5 +1,6 @@
 import json
 from abc import ABC, abstractmethod
+from collections import OrderedDict
```
tothtamas28: `dict` is insertion ordered as of Python 3.7, so this is probably not necessary.

pyk/src/pyk/kcfg.py
```diff
@@ -285,8 +296,9 @@ def pretty_print(self, kprint: KPrint) -> List[str]:
         def _print_subgraph(indent: str, curr_node: KCFG.Node, prior_on_trace: List[KCFG.Node]) -> List[str]:
             ret: List[str] = []
 
+            edges_from = sorted(self.edge_likes(source_id=curr_node.id))
             if curr_node in processed_nodes:
-                if len(self.edge_likes(source_id=curr_node.id)) == 0:
+                if len(edges_from) == 0:
```
```suggestion
                if not edges_from:
```

pyk/src/pyk/kcfg.py
```diff
@@ -323,11 +335,11 @@ def _print_subgraph(indent: str, curr_node: KCFG.Node, prior_on_trace: List[KCFG
             return ret
 
         ret = []
-        init = self.init
+        init = sorted(self.init)
         while init:
             ret.append(self.node_short_info(init[0]))
             ret.extend(_print_subgraph('', init[0], [init[0]]))
-            init = [node for node in self.nodes if node not in processed_nodes]
+            init = [node for node in sorted(self.nodes) if node not in processed_nodes]
```
```suggestion
            init = sorted(node for node in self.nodes if node not in processed_nodes)
```

tothtamas28 (approved):

PR #2684, opened 2022-06-28
Support converting Ceil and Exists conditions to Bool unsafely

ehildenb:
If the user passes `unsafe=True`, we now support converting `Ceil` conditions to `Bool`, and `#Exists` conditions.

I'm pretty sure the transformation is actually safe, but seeing as how this has only been a problem in the summarizer we can mark it as unsafe for now.

These cases were discovered when @jberthold was trying the summarizer on a new proof, and the conditions given back by the prover contained `#Ceil` and `#Exists` conditions. This will make it easier to track down what part of the specification needed to change.

### DIFF
--- a/pyk/src/pyk/kastManip.py
+++ b/pyk/src/pyk/kastManip.py
@@ -1,7 +1,9 @@
+import logging
 from collections import Counter
 from typing import (
     Callable,
     Dict,
+    Final,
     List,
     Mapping,
     Optional,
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kastManip.py
```diff
@@ -604,9 +617,9 @@ def buildRule(ruleId, initConstrainedTerm, finalConstrainedTerm, claim=False, pr
     return (minimizeRule(rule, keepVars=newKeepVars), vremapSubst)
 
 
-def abstractTermSafely(kast, baseName='V'):
+def abstract_term_safely(kast, base_name='V'):
```
tothtamas28: Please add type hints.

tothtamas28 (approved):

pyk/src/pyk/kastManip.py
ehildenb: Done.

Issue #2612, opened 2022-05-19
[KIP] - Support for symbolic sets


ana-pantilie:
We need to support symbolic sets in the frontend, similar to the way we support symbolic maps. The frontend should add similar axioms to the ones added here: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L179.

This needs to be implemented in the frontend first and only then can the changes to the backend be made.

I don't know the frontend details, but this comment from `domains.md`, module `MAP-KORE-SYMBOLIC` provides some documentation on what is generated for K maps and should be generated as well for K sets:
```
/*
// The rule below is automatically generated by the frontend for every sort
// hooked to MAP.Map. It is left here to serve as documentation.

  rule #Ceil(@M:Map (@K:KItem |-> @V:KItem)) => {(@K in_keys(@M)) #Equals false} #And #Ceil(@M) #And #Ceil(@K) #And #Ceil(@V)
    [anywhere, simplification]
*/
```

So for the sorts hooked to `SET.Set`, I think the rules should look like:
```
rule #Ceil(@S:Set SetItem(@K:KItem)) => { (@K in @S) #Equals false } #And #Ceil(@S) #And #Ceil(@K) [anywhere, simplification]
```

Motivation
----------

This reasoning is currently implemented in the Haskell backend internally which is a problematic approach. Currently, it is blocking https://github.com/runtimeverification/haskell-backend/issues/2928. Also, because of it being implemented in the backend users can't provide their own simplification rules over set functions like `-Set`.

ehildenb:
@ana-pantilie can you write down the K rules that correspond to these set-definedness conditions? You can likely infer them by inspecting the same ones generated for Map, and specializing them to sets.

This will help the implementor to know what Kore axioms to generate here.

radumereuta:
We should update domains.md to include some simplification rules for `Set` functions similar to the ones in `MAP-KORE-SYMBOLIC`

radumereuta:
We should also add tests for Sets similar to maps: https://github.com/runtimeverification/k/tree/master/k-distribution/tests/regression-new/map-symbolic-tests

radumereuta:
1. ModuleToKore.java needs to be updated
2. domains.md needs to be updated to clone `module MAP-KORE-SYMBOLIC [kore,symbolic]`
3. port the test suite `regression-new/map-symbolic-tests` to work for sets.

[...]

PR #2689, opened 2022-06-28
Symbolic implementation for "in Set", intersectSet, -Set, |Set

denis-bogdanas:
## Scope

Symbolic implementation and basic lemmas for the following operations: `E in Set`, `intersectSet`, `-Set`, `|Set`. Also interaction between `in Set` and the other 3. The test suite is somewhat an extension of the tests for symbolic map.

Due to haskell backend limitations (https://github.com/runtimeverification/haskell-backend/issues/3107), it is not possible to write rules that match `Set` constructor in its general form `Set Set`, only in the particular form `Set SetItem`. For this reason this PR mostly tests and implements expressions of the form `Set <operation> (Set SetItem)` and the opposite. And a few extra cases.

## Tests

There's some redundancy in tests, I didn't really minimize them. Instead they are reasonably complete for the expression types mentioned above.

For some of the tests output could be further minimized, but the rules for this are too narrow or I'm unsure whether their place is in domains, so I moved those tests to failing.

Also, tests 37 and 38 for all operations gets stuck, probably due to Haskell.

## Limitations

Master version of haskell backend fails this PR, it only works with https://github.com/runtimeverification/haskell-backend/pull/3062 . So a fix to haskell will have to be implemented first.

Also the `#Ceil` rule has to be implemented in the front-end, to work for cell collections of type set. This will be done in another PR.

## Related Haskell Issues
https://github.com/runtimeverification/haskell-backend/issues/3107
https://github.com/runtimeverification/haskell-backend/issues/3124

Haskell PR used here: https://github.com/runtimeverification/haskell-backend/pull/3062


denis-bogdanas:
Fixes #2612 .

denis-bogdanas:
@ehildenb why Nix tasks didn't start? Could be because this PR uses a non-master version of Haskell?

denis-bogdanas:
@ana-pantilie Some failures with k-master:
`diff-6-spec.k` is getting stuck.

`diff-11-spec.k` returns incorrect result.
Test:
```
claim <k> diff ( SetItem(Y:MyId)      -Set (S SetItem(X:MyId)) )  => . </k>
```
Expected:
```
  <k> diffResult ( SetItem ( Y:MyId ) -Set ( S SetItem ( X:MyId ) ) ) ~> . </k>
#And
  { false #Equals X:MyId in S }
```
Essentially no rules should apply except `#Ceil`.

Actual:
```
  <k>
    diffResult ( SetItem ( Y:MyId ) -Set SetItem ( X:MyId ) ) ~> .
  </k>
#And { false #Equals X:MyId in S }
```

denis-bogdanas:
@ana-pantilie Please look at the cyclic dependency mentioned above. This is what blocks progress on this PR.

denis-bogdanas:
I still need to add a fix to this branch but no longer can push. Related to https://github.com/runtimeverification/haskell-backend/issues/3124

Here it is, domains.md lines 626-630 :
```
  //Last side conditions are workarounds for: https://github.com/runtimeverification/haskell-backend/issues/3124
  rule intersectSet(S1 SetItem(E), S2) => intersectSet(S1, S2) SetItem(E) requires         E in S2 andBool notBool E in S1 [simplification]
  rule intersectSet(S1 SetItem(E), S2) => intersectSet(S1, S2)            requires notBool E in S2 andBool notBool E in S1 [simplification]     
  rule intersectSet(S1, S2 SetItem(E)) => intersectSet(S1, S2) SetItem(E) requires         E in S1 andBool notBool E in S2 [simplification]
  rule intersectSet(S1, S2 SetItem(E)) => intersectSet(S1, S2)            requires notBool E in S1 andBool notBool E in S2 [simplification]
```

ana-pantilie:
@denis-bogdanas Thanks. Does this PR depend on https://github.com/runtimeverification/k/pull/2692 being merged first?

denis-bogdanas:
As they stand now, the 2 PRs are independent. But it is preferable to merge https://github.com/runtimeverification/k/pull/2692 first because then we can remove `#Ceil(Set SetItem)` rule implemented here in domains.md. Otherwise it will have to be removed in a separate, 3rd PR.

ehildenb:
Move whatever tests you need to the `failing` subdirectory to get this merged. Then we can update later and move them back into main directory.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -522,7 +522,7 @@ number of unbound keys being mached. In other words, one unbound variable is
 linear, two is quadratic, three is cubic, etc.
 
 ```k
-  syntax Set ::= Set Set                  [left, function, functional, hook(SET.concat), klabel(_Set_), symbol, assoc, comm, unit(.Set), idem, element(SetItem), format(%1%n%2)]
+  syntax Set ::= Set Set                  [left, function, hook(SET.concat), klabel(_Set_), symbol, assoc, comm, unit(.Set), idem, element(SetItem), format(%1%n%2)]
```
ehildenb, line 525: @dwightguth this seems sound to me (ok to remove the extra axiom), but it also seems incorrect. `_Set_` should be functional too, right?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -610,6 +610,72 @@ that two different sets will return the same element, even if they are similar.
...
+  imports private K-EQUAL
+  imports private BOOL
+
+  // Symbolic in
+  //Unsupported: "Set1 Set2" matching. Using "Set SetItem(E) instead."
+  //rule E  in ( S1 S2)                  => E in S2 requires notBool E in S1       [simplification]
+  //rule E  in (S1 _S2)                  => true    requires         E in S1       [simplification]
+  rule E1 in (S SetItem(E2))           => true requires E1 ==K E2 orBool E1 in S     [simplification]
+  rule E1 in (S SetItem(E2))           => E1 in S           requires E1 =/=K E2      [simplification]                  
+  rule E1 in (S SetItem(E2))           => E1 in SetItem(E2) requires notBool E1 in S [simplification]
```
ehildenb: This seems like it will result in a non-deterministic search though the `Set` on the RHS of the `_in_` operator, which will be poor performance.

How will this actually behave operationally?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -610,6 +610,72 @@ that two different sets will return the same element, even if they are similar.
...
+
+  // Symbolic intersectSet
+  rule intersectSet(_S, .Set) => .Set                                                                                      [simplification]                                                                                         
+  rule intersectSet(.Set, _S) => .Set                                                                                      [simplification]
+  rule intersectSet(S, S) => S                                                                                             [simplification]
+  rule intersectSet(S1 SetItem(E), S2) => intersectSet(S1, S2) SetItem(E) requires         E in S2                         [simplification]
+  rule intersectSet(S1 SetItem(E), S2) => intersectSet(S1, S2)            requires notBool E in S2 ensures notBool E in S1 [simplification]     
+  rule intersectSet(S1, S2 SetItem(E)) => intersectSet(S1, S2) SetItem(E) requires         E in S1                         [simplification]
+  rule intersectSet(S1, S2 SetItem(E)) => intersectSet(S1, S2)            requires notBool E in S1 ensures notBool E in S2 [simplification]
```
ehildenb, line 648: Same here, do we have an idea of the performance characteristics here? What if you are intersecting 2 100-element concrete sets that are exactly duplicates of each other? What about 2 100-element sets that have no elements in common?

ehildenb: One way around the non-determinism is to use `Set2List`, and then do list intersection, and then convert back to a `Set`. This may be the way to go.

Can we add tests of the above two scenarios? (the 100-element sets).

k-distribution/include/kframework/builtin/domains.md
dwightguth: No, this is a correct change. I marked it functional originally because the assumption was that sets are idempotent, but we've been treating sets as nilpotent instead, so `SetItem(X) SetItem(X)` is bottom, which means this can't be functional. `unionSet` ought to be functional though.

jberthold:
This PR has become quite stale, and has performance issues (see [related test](https://github.com/runtimeverification/k/pull/2689/commits/bf2402b16418f39f3995ff17cff0642372163a38)). A new branch and PR will be opened to deal with #2612

jberthold:
Superseded by #2920

PR #2690, opened 2022-06-28
Requires in haskell is LHS

### LINKED ISSUE #2687: [Bug] [kompile] - Reporting a `requires` clause variable as RHS
K Version
---------

```
+ kompile --version
K version:    5.3.0
Build date:   Tue Jun 28 15:38:19 UTC 2022
```

Actual version: 6f12ca393c

Description
-----------

A variable on the LHS of the rule is being seen as on teh RHS of the rule for the haskell backend.

Input Files
-----------

```
module TEST
    imports MAP
    imports INT
    imports BOOL

    syntax Pgm ::= "a" | "b"

    rule <k> a => b </k>

    syntax Int ::= f ( Int ) [function, functional, no-evaluators]

    rule <k> a => b </k> requires f(_X) ==Int 3

endmodule
```

Reproduction Steps
------------------

```
+ kompile --backend haskell test.k --main-module TEST --syntax-module TEST
[Error] Compiler: Found variable _X on right hand side of rule, not bound on
left hand side. Did you mean "?_X"?
        Location(14,37,14,39)
        14 |        rule <k> a => b </k> requires f(_X) ==Int 3
           .                                        ^~
```

Expected Behavior
-----------------

This should kompile just fine for the haskell backend, variables in the `requires` clause should be considered as part of LHS for haskell backend.

radumereuta:
Fixes: #2687

### DIFF
--- a/k-distribution/tests/regression-new/checkWarns/checkRHSHaskell.k
+++ b/k-distribution/tests/regression-new/checkWarns/checkRHSHaskell.k
@@ -1,0 +1,25 @@
+// This should kompile just fine for the haskell backend, variables
+// in the requires clause should be considered as part of LHS for haskell backend.
+// https://github.com/runtimeverification/k/issues/2687
+
+module CHECKRHSHASKELL-SYNTAX
+
+endmodule
+
+module CHECKRHSHASKELL
+    imports CHECKRHSHASKELL-SYNTAX
+    imports MAP
+    imports INT
+    imports BOOL
+
+    syntax Pgm ::= "a" | "b"
+
+    rule <k> a => b </k>
+
+    syntax Int ::= f ( Int ) [function, functional, no-evaluators]
+
+    configuration <k> $PGM:Pgm </k>
+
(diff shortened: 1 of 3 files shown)


dwightguth (approved):
The change itself seems fine. I'm not sure the code that was there before was correct, but I guess if it's passing the tests it needs to pass, I could be mistaken. I'm just a little confused about whether it's better to treat the requires clause as "left-hand-side and right-hand side", which is what it is currently doing, or "left-hand-side only", which seems more intuitive.

I'm approving though as it seems to pass the test you wrote.

ehildenb:
I think the ML encoding is pretty much that requires clause is fully LHS, and ensures is RHS. It's encoded as `LHS #And REQ #Implies \next RHS #And ENS` if I understand correctly.

That's also how the pyk library treats it when building larger transition system structures like CFGs.

ehildenb:
And thank you for the fast turnaround @radumereuta

PR #2692, opened 2022-06-29
ModuleToKORE: generating #Ceil(Set) rule for generated sets

denis-bogdanas:
Prerequisite for https://github.com/runtimeverification/k/pull/2689
Tested with that PR, replacing domain.md rule with the generated rule, and works.


ehildenb:
We should have some tests.

denis-bogdanas:
@ehildenb I wrote 2 quite exhaustive tests. One for domains set another for cell set. The `.out` files are what the current out is, for the sake of review. Not necesarily correct. The domains set output is exactly what expected.

But the cell set test indicates that respective `#Ceil` rule is not generated. I confirmed this in `definition.kore`. This is because the algorithm detects generated sets by looking at rules hooked to `SET.in`, similarly to how `#Ceil(Map)` is looking for `MAP.in_keys`. Looks like that hook is not generated for set cells. And it is needed for building `#Ceil` rule.

Generating the hooked `SET.in` rule for cells is beyond my competence. I guess someone else should continue from this point.

jberthold:
This PR was superseded by #2961 . Which we unfortunately had to revert because of its performance impact (we have to get back to this at some point but after other ceil simplification optimisations)

Issue #2683, opened 2022-06-27
_|->_ and SetItem (and likely others) maybe should be constructors


ehildenb:
Here: https://github.com/runtimeverification/k/pull/2659/files#r907667239

I've noticed that we are marking things like `_|->_` and `SetItem` as `function, functional`, which doesn't make any sense. These shoudl be constructors. Nice thing about making them constructors is that we get (for free) that (i) they are total, and (ii) they are injective.

dwightguth:
Hooks are a property of functions. There is no such thing as a hooked constructor in K, so I'm not entirely sure what you're envisioning here. Of course as in all things this is a decision that can be revisited, but I don't think you fully understand how much is involved in changing this, nor how little, seemingly, we gain from it.

ehildenb:
@grosu @xc93 or @virgil-serbanuta we need to know what the difference between a constructor in ML is and an injective total function.

If they have the same axiomatization in ML, then we do not need to make any changes. If they have different axiomatization, then we need to inspect the differences.

My interpretation of these symbols is that they are (free) constructors. But it's unclear if it's necessary, of if having them be total injective functions is OK.

ehildenb:
According to this: https://github.com/runtimeverification/haskell-backend/blob/master/docs/introduction-to-matching-logic.pdf

The symbol defined here has `function, functional, injective`, which means:

- It's defined on all inputs.
- And it satisfies same-constructor no-confusion axiom.

These axioms seem to be satisfied by constructors, so this should be sound.

It still feels strange to me to declare these as function and not constructor, but I don't think it leads ot any soundness issues.

virgil-serbanuta:
They are not constructors because they don't have the "no junk" property. In order to get that, you need `_Map_`, which does not have the "no confusion" property since `_Map_(A, B) == _Map_(B, A)`.

ehildenb:
They should have no-junk property I think? But they do not have no-confusion. But there is a distinction in initial algebra semantics betwene having them be injective functions or constructors.

Here is an example where the distinction matters: https://github.com/runtimeverification/k/pull/2693#issue-1289523457. When "defunctionalizing" the LHS of rules, we do not want to consider these fnuctions to be constructors, because we want to do pattern-matching on them if possible. Especially the symbols like `_AccountCellMap_` and `AccountCellMapItem`.

[...]

PR #2693, opened 2022-06-30
pyk getters for KOuter classes, updates to unparsing

ehildenb:
This adds several methods to the KAST outer classes `KFlatModule` and `KDefinition`:

- `syntax_productions`: return productions that have a klabel.
- `functions`: return things marked as function, but _not_ non-free-constructors (@dwightguth this distinction becomes important in a routine in pyk I've written to remove function symbols from the LHS of rules, here we want things marked in this PR as "NON_FREE_CONSTRUCTORS" to remain on the LHS of rules so you can do normal pattern matching on them).
- `constructors`: return things that are not functions.
- The unparser only builds symbols in the symbol table for `syntax_productions`, not all `productions`.

### DIFF
--- a/pyk/src/pyk/kast.py
+++ b/pyk/src/pyk/kast.py
@@ -1235,7 +1235,9 @@ class KFlatModule(KOuter, WithKAtt):
 
     @property
     def functions(self) -> List[KProduction]:
-        return [prod for prod in self.syntax_productions if 'function' in prod.att.atts or 'functional' in prod.att.atts and not (prod.klabel and KFlatModule._is_non_free_constructor(prod.klabel.name))]
+        _functions = [prod for prod in self.syntax_productions if 'function' in prod.att.atts or 'functional' in prod.att.atts]
+        _functions = [f for f in _functions if not (f.klabel and KFlatModule._is_non_free_constructor(f.klabel.name))]
+        return _functions
 
     @property
     def constructors(self) -> List[KProduction]:
--- a/pyk/src/pyk/kast.py
+++ b/pyk/src/pyk/kast.py
@@ -1224,6 +1224,10 @@ class KFlatModule(KOuter, WithKAtt):
         return [sentence for sentence in self.sentences if type(sentence) is KProduction]
 
     @property
+    def syntax_productions(self) -> List[KProduction]:
+        return [prod for prod in self.productions if prod.klabel]
+
+    @property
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kast.py
```diff
@@ -1223,6 +1223,26 @@ def __iter__(self) -> Iterator[KSentence]:
...
         return [sentence for sentence in self.sentences if type(sentence) is KProduction]
 
+    @property
+    def syntax_productions(self) -> List[KProduction]:
+        return [prod for prod in self.productions if prod.klabel]
+
+    @staticmethod
+    def _is_non_free_constructor(label: str) -> bool:
+        is_cell_map_constructor = label.endswith('CellMapItem') or label.endswith('CellMap_')
+        is_builtin_data_constructor = label in ['_Set_', '_List_', '_Map_', 'SetItem', 'ListItem', '_|->_']
```
tothtamas28:
```suggestion
        is_builtin_data_constructor = label in ['_Set_', '_List_', '_Map_', 'SetItem', 'ListItem', '_|->_']
```
```suggestion
        is_builtin_data_constructor = label in {'_Set_', '_List_', '_Map_', 'SetItem', 'ListItem', '_|->_'}
```

pyk/src/pyk/kast.py
```diff
@@ -1223,6 +1223,26 @@ def __iter__(self) -> Iterator[KSentence]:
     def productions(self) -> List[KProduction]:
         return [sentence for sentence in self.sentences if type(sentence) is KProduction]
 
+    @property
+    def syntax_productions(self) -> List[KProduction]:
+        return [prod for prod in self.productions if prod.klabel]
+
+    @staticmethod
+    def _is_non_free_constructor(label: str) -> bool:
```
tothtamas28: Consider moving this method under the `functions` property.

tothtamas28 (approved):

pyk/src/pyk/kast.py
tothtamas28: Or alternatively, this could even be a public method of `KLabel`.

ehildenb: I moved it into the `functions` property, because I'm hoping that there are not many places where this distintion between injective functions and constructors is important. If it does become important, we should revisit this issue: https://github.com/runtimeverification/k/issues/2683

PR #2696, opened 2022-07-01
Add module pyk.kore

### LINKED ISSUE #2626: Kore data-structures and parser in pyk
We eventually want the pyk library to be able to read/write Kore terms directly, and do some small manipulations of them without having to go through frontend Kast. For example, if we are just adding a constraint to a state, or renaming some variables, then there should be no reason to go through frontend Kast.

Additionally, pyk should be able to directly translate Kore => Kast without going through the frontend unparsing routines (the other direction is harder, and we will have to rely on the frontend for that).

To facilitate this long-term, we should add Kore data-structures in `kore.py` in the pyk library, and a parser/unparser for it.

The Kore language should be LL1 or LL2, se we should be able to role a fast parser for it. Several examples exist, here is the most recent rewrite of the Haskell backends kore-parser, I guess this can just be translated 1-for-1: https://github.com/runtimeverification/haskell-backend/pull/2967. Here is some documentation on the Kore language itself: https://github.com/runtimeverification/haskell-backend/tree/master/docs (look at kore-syntax.md)

tothtamas28:
Closes #2626

This PR introduces the `pyk.kore` module. It contains:
* Class hierarchy for the Kore abstract syntax
* Parsing and unparsing in textual Kore format
* Parsing and unparsing in JSON format

Currently, only syntactic validation is performed on the AST (e.g. identifiers are lexically correct), semantic validation rules described in https://github.com/runtimeverification/haskell-backend/blob/master/docs/kore-syntax.md#validity are not implemented.


radumereuta (commented):
These parser files are very large. I don't feel confident reviewing them. At least not by myself.
Are they generated by any chance?

tothtamas28:
> Are they generated by any chance?

No. Since the Kore grammar is relatively simple, I wrote the parser manually.

ehildenb (approved):

ehildenb:
Looks fine to me on first glance (no deep review of the Kore). I'm sure we'll find problems as we test, but no need to hang up the first version of it too long.

ehildenb:
We likely will want to use some faster parser eventually, but we can cross that bridge when we come to it, and by then we'll have more test-cases we can use.

PR #2698, opened 2022-07-03
Calculate used variable names better in buildRule

ehildenb:
This PR does several things:

- Moves `splitConfigAndConstraint` to inside `CTerm`, so now there are no dependencies of `cterm.py` on `kastManip.py`.
- Renames variables in `buildRule`, and renames method to `build_rule`.
- Lifts `build_rule` to operate over `CTerm`, instead of over `KInner`.
- Adds a test where `build_rule` is producing variable names in the rule which cause the frontend to emit warnings.
- Fixes bug where `build_rule` is naming variables poorly.
- Adds a method `with_constraint` to `CTerm` for easily making new `CTerm` with an added constraint.
- Removes an old test of `build_rule` that becomes redundant with the added tests here.
- Moves the `test_configuration.py` tests into unit tests under `test_kastManip.py` so they don't build definitions every time (they don't need to).

The main change to `buildRule` is that if you have a rule:

```
    rule <k> K_CELL ... </k> <mem> MEM => MEM' </mem>
```

It will now realize that `K_CELL` should be anonymous, and rename it to `_K_CELL` before feeding it to the frontend. This is to avoid a bunch of frontend warnings.


pyk/src/pyk/cterm.py
```diff
@@ -79,3 +92,6 @@ def _ml_impl(antecedents: Iterable[KInner], consequents: Iterable[KInner]) -> KI
             return consequent
 
         return mlImplies(antecedent, consequent, Sorts.GENERATED_TOP_CELL)
+
+    def with_constraint(self, new_constraint: KInner) -> 'CTerm':
+        return CTerm(mlAnd([self.config, new_constraint] + list(self.constraints), Sorts.GENERATED_TOP_CELL))
```
tothtamas28: I'd call it `add_constraint` instead. To me, `with_constraint` suggests "take the configuration, and couple it with the constraint provided."

```suggestion
    def add_constraint(self, constraint: KInner) -> 'CTerm':
        return CTerm(mlAnd((self.config, constraint) + self.constraints, Sorts.GENERATED_TOP_CELL))
```

pyk/src/pyk/cterm.py
```diff
@@ -15,11 +14,25 @@ class CTerm:
...
 
     def __init__(self, term: KInner) -> None:
-        config, constraint = splitConfigAndConstraints(term, Sorts.GENERATED_TOP_CELL)
+        config, constraint = CTerm._split_config_and_constraints(term)
         constraints = CTerm._normalize_constraints(flattenLabel('#And', constraint))
         object.__setattr__(self, 'config', config)
         object.__setattr__(self, 'constraints', constraints)
 
+    @staticmethod
+    def _split_config_and_constraints(kast: KInner) -> Tuple[KInner, KInner]:
```
tothtamas28: As a first step, I'd just move this to `cterm.py`. Later, when we are ready to eliminate calls to the function, e.g. by replacing

```python
config, constraint = split_config_and_constraints(term)
```

by

```python
config, *constraints = Cterm(term)
constraint = mlAnd(constraints)
```

we can make it a private static method.

tothtamas28 (approved):

nishantjr (approved):
Looks good.

pyk/src/pyk/cterm.py
ehildenb: Done.

pyk/src/pyk/cterm.py
ehildenb: Done.

PR #2699, opened 2022-07-04
Adding lesson on basic proofs

PetarMax:
Introducing Lesson 22 on basic proofs, using examples from the 0-to-K tutorial


charala1:
@PetarMax should we name the examples according to the `lesson-22-*` scheme as in the rest of the tutorial?

PetarMax:
Yes, please, that naming convention should probably be respected.

charala1:
> Yes, please, that naming convention should probably be respected.

Done. However, I think we should put the whole listings in as opposed to asking the reader to add definitions into the modules (and use `lesson-22-{a, b, c...}` prefixes for the various files). Maybe this is overkill, I don't know.

PetarMax:
I'd prefer not to to present the entire file at once, because then I'd have to refer the reader back. Like this, it's all inlined in the text. Alternatively, the lesson itself could be the file, given literate programming. Perhaps that's the way to go?

Baltoli:
See [commands.sh](https://github.com/runtimeverification/k/blob/master/k-distribution/k-tutorial/1_basic/commands.sh) for examples of how the tutorial's literate definitions are compiled in CI. You want each `A`, `B`, `C` module to be individually compilable with the appropriate setting for the main module.

k-distribution/k-tutorial/1_basic/22_proofs/README.md
```diff
@@ -1 +1,444 @@
...
+claim <k> $a = A:Int ; $b = B:Int ;
+          if (A < B) {
+            $c = $b ;
+          } else {
+            $c = $a ;
+          }
+        => . ... </k>
+        <store> STORE => STORE [ $a <- A ] [ $b <- B ] [ $c <- ?C:Int ] </store>
+    ensures (?C ==Int A) orBool (?C ==Int B)
+```
```
nwatson22: Would it make more sense to write `if ($a < $b) {` instead of `if (A < B) {`? My understanding is that this would make it a claim about programs that declare two variables and then compare them, etc. while currently it is a claim about programs which declare two variables `$a` and `$b` and then compares two `Int` values which happened to evaluate to the same values as `$a` and `$b` were initially assigned to, and the former behavior intuitively makes more sense to me as a property we would want to prove.

k-distribution/k-tutorial/1_basic/22_proofs/README.md
```diff
@@ -1 +1,444 @@
...
+      _Gen5
+    </stack>
+  </generatedTop>
+#And
+  {
+    true
+  #Equals
+    A <Int B
+  }
+```
```
nwatson22, line 290: On Friday when I tested this I was only getting this output after including `MAP-SYMBOLIC`, otherwise it was getting stuck earlier on some reasoning about maps. Maybe this is just caused by a quirk in my setup.

k-distribution/k-tutorial/1_basic/22_proofs/README.md
```diff
@@ -1 +1,444 @@
...
+STORE [ $a <- A:Int ] [ $b <- B:Int ] [ $c <- B:Int ]
+```
+
+we see that we should be within the constraints of the `ensures`, since the
+value of `$c` in the store equals `B` in this branch. We next examine the
+negative and positive constraints of the output and, more often than not, the
+goal is to instruct K how to use the information from the final configuration
+and the positive constraints to falsify one of the negative constraints. This
+is done through *simplifications*.
```
nwatson22, line 324: When I went through this I thought the parts about negative constraints could use more explanation, like how the negative constraints are generated or something, or more on their meaning. I just found it confusing that at least in this example the negative constraint is just the negation of something that is actually true but that the prover wasn't able to prove, whereas other constraints are generated by taking branches where it is assumed that constraint holds.

k-distribution/k-tutorial/1_basic/22_proofs/README.md
PetarMax: Yes. program variables should be used throughout the program except in those initial assignments, we've discussed that on Thursday, I think?

k-distribution/k-tutorial/1_basic/22_proofs/README.md
PetarMax: Hmmm, let's take a look at that today.

PetarMax: Ahhh, it might be that @charala1 removing `domains.md` from being imported [here](https://github.com/runtimeverification/k/pull/2699/commits/9628cfa4b69ed317a5908014c9eb537fa8b114ee) caused this. I think we need that back.

k-distribution/k-tutorial/1_basic/22_proofs/README.md
```diff
@@ -1 +1,448 @@
...
+```
+
+and we observe, from the first equality, that the existential `?C` should be
+instantiated with `B`. This would make both branches of the `#And` true,
+falsifying the outside `#Not`. We just need to show `K` how to conclude that
+`?C ==Int B`. We do so by introducing the following simplification into the
+`VERIFICATION` module:
+
+```
+rule { M:Map [ K <- V ] #Equals M [ K <- V' ] } => { V #Equals V' } [simplification, anywhere]
```
bbyalcinkaya: The `anywhere` attribute should be removed since `[simplification, anywhere]` is now forbidden.

```suggestion
rule { M:Map [ K <- V ] #Equals M [ K <- V' ] } => { V #Equals V' } [simplification]
```

k-distribution/k-tutorial/1_basic/22_proofs/README.md
```diff
@@ -1 +1,448 @@
...
+  }
+```
+
+This output starts with a message telling us at which point the proof failed,
+followed by the final state, which consists of three parts: some negative
+Matching Logic (ML) constraints, the final configuration (`<generatedTop> ...
+</generatedTop>`), and some positive ML constraints. Generally speaking,
+these positive and the negative constraints could arise from various sources,
+such as (but not limited to) branches taken by the execution
+(e.g. `{ true #Equals A <Int B }` or `#Not ( { true #Equals A <Int B>} ) `),
```
radumereuta:
```suggestion
(e.g. `{ true #Equals A <Int B }` or `#Not ( { true #Equals A <Int B } ) `),
```

ehildenb (approved):

PR #2706, opened 2022-07-05
Pyk: Additional args for kompile, KProve has kompile named constructor.

nishantjr:
Besides the args needed for KEVM, I've allowed the `kompile` to take additional unstructured args. This is so we don't get blocked on adding new args to this function if we need an argument we don't explicitly support.

### DIFF
--- a/pyk/src/pyk/ktool/kompile.py
+++ b/pyk/src/pyk/ktool/kompile.py
@@ -28,7 +28,7 @@ def kompile(
     hook_namespaces: Iterable[str] = (),
     emit_json=False,
     concrete_rules: Iterable[str] = (),
-    additional_args: Iterable[str] = ()
+    additional_args: Iterable[str] = (),
 ) -> Path:
     check_file_path(main_file)
 
--- a/pyk/src/pyk/ktool/kompile.py
+++ b/pyk/src/pyk/ktool/kompile.py
@@ -19,17 +19,32 @@ class KompileBackend(Enum):
 def kompile(
     main_file: Path,
     *,
+    main_module: Optional[str] = None,
+    syntax_module: Optional[str] = None,
     backend: Optional[KompileBackend],
     output_dir: Optional[Path] = None,
     include_dirs: Iterable[Path] = (),
+    md_selector: Optional[str] = None,
+    hook_namespaces: Iterable[str] = (),
     emit_json=False,
+    concrete_rules: Iterable[str] = (),
 ) -> Path:
     check_file_path(main_file)
 
     for include_dir in include_dirs:
         check_dir_path(include_dir)
 
-    args = _build_arg_list(backend=backend, output_dir=output_dir, include_dirs=include_dirs, emit_json=emit_json)
+    args = _build_arg_list(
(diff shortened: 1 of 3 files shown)


pyk/src/pyk/ktool/kprove.py
```diff
@@ -98,6 +99,10 @@ def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
         with open(self.kompiled_directory / 'mainModule.txt', 'r') as mm:
             self.main_module = mm.read()
 
+    @staticmethod
+    def kompile(main_file: Path, *args, **kwargs) -> 'KProve':
+        return KProve(kompile(main_file, *args, **kwargs))
+
```
ehildenb: Do we have a use-case for this? This adds a dependency between kprove/kompile which may not make sense down the road, and it's not very much work for the user to just call `kompile` themselves.

pyk/src/pyk/ktool/kompile.py
```diff
@@ -43,13 +60,25 @@ def kompile(
...
+    additional_args: Iterable[str]
 ) -> List[str]:
     args = []
 
+    if main_module:
+        args.extend(['--main-module', main_module])
+
+    if syntax_module:
+        args.extend(['--syntax-module', syntax_module])
+
```
ehildenb, line 81: Do we want to default this to the `main_module` if it's not provided? That way we get less warnings from the frontend, but get the same default behavior?

pyk/src/pyk/ktool/kprove.py
nishantjr: OK. I thought you wanted the same interface as with `KEVM`.

pyk/src/pyk/ktool/kompile.py
nishantjr: I'd prefer letting the frontend deal with defaults, rather than duplicating that logic. If we feel that the warning is an overkill we should change the behaviour in the frontend rather than having an opinion here.

nishantjr: It's also easy enough for the client to suppress the warning by supplying a syntax module.

ehildenb (approved):

pyk/src/pyk/ktool/kompile.py
```diff
@@ -28,6 +28,7 @@ def kompile(
     hook_namespaces: Iterable[str] = (),
     emit_json=False,
     concrete_rules: Iterable[str] = (),
+    additional_args: Iterable[str] = ()
```
tothtamas28: This is a matter of personal preference, but I like to add the trailing comma to such lists. It is easier to extend, rearrange, etc. this way.

```suggestion
    additional_args: Iterable[str] = (),
```

tothtamas28 (approved):

pyk/src/pyk/ktool/kompile.py
nishantjr: My thinking here was that this should always be the last argument irrespective.  But it makes sense for consistency.

PR #2710, opened 2022-07-07
Add warning for syntax priority / associativity mismatch

### LINKED ISSUE #2460: [KIP] - Add a Warning when defining 'syntax left' for operators with different priority
Motivation
----------
Tried to define associativity for all operators (different priorities) together using ```syntax``` left. Then encountered runtime error.
It is conflicting the sdf-inspired parser K is using. Follow [this thread](https://runtimeverification.slack.com/archives/C7E701MFG/p1646181067237979?thread_ts=1646180449.465819&cid=C7E701MFG) for more information:

It will be good to emit a warning than a runtime error.

Example K Code
--------------
```
module CALC-ARITH
    syntax Exp ::= "-" Exp      [neg, color(blue), seqstrict]
                 > Exp "*" Exp   [times, color(green), seqstrict] 
                 | Exp "/" Exp   [div, color(green), seqstrict]
                 | Exp "%" Exp   [mod, color(green), seqstrict]
                 > Exp "+" Exp   [plus, color(green), seqstrict]
                 | Exp "-" Exp   [minus, color(green), seqstrict]

    syntax left times div mod plus minus
endmodule

module CALC-COMP
    syntax Exp ::= Exp "<"  Exp [lt,  color(red)]
                 | Exp "<=" Exp [leq, color(red)]
                 | Exp ">"  Exp [gt,  color(red)]
                 | Exp ">=" Exp [geq, color(red)]
                 | Exp "==" Exp [eq,  color(red)]
                 | Exp "!=" Exp [neq, color(red)]
                //  | isEven(Exp)  [function, color(red)]
endmodule

module CALC-LOGIK
    syntax Exp ::= "!" Exp [not, color(blue)]
                 > Exp "&&" Exp [and,   color(blue)]
                 > Exp "||" Exp [or,    color(blue)]
                 > Exp "->" Exp [imply, color(blue)]
    syntax left and or imply
endmodule

module CALC-SYNTAX
    imports CALC-ARITH
    imports CALC-COMP
    imports CALC-LOGIK
    imports INT
    imports BOOL

    syntax Exp::= Int | Bool | "(" Exp ")"   [bracket]

    syntax priorities plus > lt > not
endmodule

module CALC
    imports CALC-SYNTAX

    rule <k> - I:Int => 0 -Int I ...</k>
    rule <k> I1:Int * I2:Int => I1 *Int I2 ...</k>
    rule <k> I1:Int / I2:Int => I1 /Int I2 ...</k> requires I2 =/=Int 0
    rule <k> I1:Int % I2:Int => I1 %Int I2 ...</k> requires I2 =/=Int 0
    rule <k> I1:Int + I2:Int => I1 +Int I2 ...</k>
    rule <k> I2:Int - I2:Int => I2 -Int I2 ...</k>

    syntax Bool ::= isKResult(K) [function, symbol]
    
    rule isKResult(_:Int) => true
    rule isKResult(_:Bool) => true
    rule isKResult(_) => false [owise]
endmodule

```
A waring should be prompt during compilation on line
```    
syntax left times div mod plus minus
``` 
and line 
```    
 syntax left and or imply 
```.

Documentation
-------------

Warning: ``` syntax left times div mod plus minus ``` defines associativity for operators with different priorities which could break the priorities intended. Recommend to break the definition into 
```    
syntax left times div mod
syntax left plus, minus
```
if it is not an intended behaviour.


Testing Approach
----------------

### PR DESCRIPTION
**Pull request by Baltoli**
Fixes #2460 

This PR adds a compiler check that the labels in a syntax associativity sentence are not in different priority groups, and updates the relevant sections of the K tutorial and user manual to reflect this.

Note that the error message doesn't have location information, because the `SyntaxAssociativity` sentence that I use in the check doesn't seem to include any - is there a better way to phrase this implementation?

### CODE REVIEW

**Comment by ehildenb** [comment]
Grab location information here and make sure it's preserved in the frontend KORE (not the same as backend Kore) that is produced: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java#L141

KIL are the flat parsed data-structures produced by the frontend outer parser.

File: k-distribution/tests/regression-new/syntax-groups/syntaxGroups.k [hunk 1]
```diff
@@ -0,0 +1,11 @@
...
+  syntax Exp  ::= foo()
+                > Exp "*" Exp [times]
+                > Exp "+" Exp [plus]
+
+  syntax left times plus
+endmodule
+
+module SYNTAXGROUPS
+  imports SYNTAXGROUPS-SYNTAX
+endmodule
```
radumereuta: We can move this test to `regression-new/checkWarns` to keep things organized.

kernel/src/main/java/org/kframework/compile/checks/CheckSyntaxGroups.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public CheckSyntaxGroups(Set<KEMException> errors, Module module, KExceptionManager kem) {
+        this.errors = errors;
+        this.kem = kem;
+        this.module = module;
+    }
+
+    public void check(Sentence s) {
+        if(s instanceof SyntaxAssociativity) {
+            Set<Tag> tags = JavaConverters.setAsJavaSet(((SyntaxAssociativity) s).tags());
```
radumereuta: Usually we use `mutable` from `collections.scala`
You will need to add
```
import static org.kframework.Collections.*;
```
```suggestion
            Set<Tag> tags = mutable(((SyntaxAssociativity) s).tags());
```

radumereuta (changes requested):
Two minor issues.

radumereuta (approved):

dwightguth (approved):

PR #2711, opened 2022-07-08
Forbid `[anywhere, simplification]` on rules

### LINKED ISSUE #2688: Forbid `anywhere` attribute on `simplification` rules
Now that we don't need to put `anywhere` on simplification rules targeting ML predicates, we should forbid this behavior so that lemmas written in this way need to be updated.

Later, if it makes sense, we can un-forbid it.

Baltoli:
Fixes #2688.

The rationale for this change is linked in the issue above; we already forbid other attributes in combination with `simplification` so the actual code change is very small.


dwightguth:
You are going to need to update the places in the test suite where this is used, as well as delete the check that asserts the oppose wrt simplification rules in spec modules.

radumereuta:
lgtm, but I'm not sure what is wrong with CI
I'm going to wait and have another look when tests are passing.

radumereuta (approved):

radumereuta:
@Baltoli this has a failing test. Probably fine to automerge if you address that.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -452,11 +452,6 @@ public void proverChecksX(Module specModule, Module mainDefModule) {
                 if (s instanceof Rule && !s.att().contains(Att.SIMPLIFICATION()))
                     errors.add(KEMException.compilerError("Only claims and simplification rules are allowed in proof modules.", s));
             }
-            if (s instanceof Rule && (s.att().contains(Att.SIMPLIFICATION()) && !s.att().contains(Att.ANYWHERE()))) {
-                KLabel kl = m.matchKLabel((Rule) s);
-                if (!m.functions().contains(kl))
-                    errors.add(KEMException.compilerError("Simplification rules need to be function/functional like.", s));
-            }
```
radumereuta, line 459: Why did you remove this entire check?

Baltoli: Dwight suggested above that I do so. If we kept the check as is, it would fire for every single simplification rule in a spec because they can no longer have the anywhere attribute.

dwightguth (approved):

PR #2712, opened 2022-07-08
Update docker agent to focal

Baltoli:
The CI build was breaking because the latest version of Calibre (which we use to build the tutorial from markdown) requires a newer version of `libc` than is provided in Ubuntu Bionic.

This PR upgrades the Jenkins agent from bionic to focal.


radumereuta:
Already merged: https://github.com/runtimeverification/k/pull/2719
This PR would have deserved more noise in the #k channel :)

PR #2725, opened 2022-07-12
KCFG: fix printing substitutions, Sort by nodes not node content

nishantjr:
The previous PR was passing because I had mocked `KPrint`. Unfortunately I had marked the PR as automerge.
This PR fixes printing `Subst`s, and sorts nodes by their content rather than hash, making for a more stable output, when hashes change.


pyk/src/pyk/ktool/kprint.py
```diff
@@ -108,6 +109,8 @@ def prettyPrintKast(kast, symbol_table, debug=False):
         sys.stderr.flush()
     if kast is None or kast == {}:
         return ""
+    if type(kast) is Subst:
+        return '\n'.join(key + ' |-> ' + prettyPrintKast(value, symbol_table) for key, value in kast.items())
```
ehildenb: I don't think this is a good idea, beacuse it complicates the type of `prettyPrintKast`. Please add type annotations to `prettyPrintKast`.

I'd rather it be a method of `Subst` called `pretty_print`, which takes a `KDefinition` I suppose.

nishantjr:
@tothtamas28 I add the pretty printing to `KPrint` rather than as a method to avoid a circular dependency.

pyk/src/pyk/tests/test_pretty_print_kast.py
```diff
@@ -10,17 +10,19 @@
...
 
 
 class PrettyPrintKastTest(TestCase):
-    TEST_DATA: Final[Tuple[Tuple[KAst, str], ...]] = (
+    TEST_DATA: Final[Tuple[Tuple[Union[KAst, Subst], str], ...]] = (
         (KRule(TRUE), 'rule  true\n  '),
         (KRule(TRUE, ensures=TRUE), 'rule  true\n  '),
-        (KRule(TRUE, ensures=KApply('_andBool_', [TRUE, TRUE])), 'rule  true\n   ensures ( true\n   andBool ( true\n           ))\n  ')
+        (KRule(TRUE, ensures=KApply('_andBool_', [TRUE, TRUE])), 'rule  true\n   ensures ( true\n   andBool ( true\n           ))\n  '),
+        (Subst({'X': TRUE, 'Y': KApply('_andBool_', [TRUE, TRUE])}), 'X |-> true\nY |-> _andBool_ ( true , true )'),
```
ehildenb: Seems like a test that belongs in `test_subst.py` instead.

pyk/src/pyk/ktool/kprint.py
nishantjr: I would too, but that introduces a circular dependency between `kprint` and `kast`.  It would need to take a `KPrint` to make the recursive calls to print the values. How do you feel about moving `Subst` to its own file?

nishantjr: Hmm, I've got an idea to get around this. Give me a bit.

nishantjr: @ehildenb For now, I've got around the circular dependency by not typing `Subst.pretty_print`.

pyk/src/pyk/kast.py
```diff
@@ -63,6 +63,17 @@ def _check_node(cls: Type[T], d: Dict[str, Any], expected: Optional[str] = None)
...
+        # shallow copy version of dataclass.astuple.
+        return tuple([self.__dict__[field.name] for field in fields(type(self))])
+
+    def __lt__(t1, t2):
+        if not isinstance(t2, KAst):
+            return NotImplemented
+        if type(t1) == type(t2):
+            return t1._as_shallow_tuple() < t2._as_shallow_tuple()
+        return type(t1).__name__ < type(t2).__name__
+
```
nishantjr: @tothtamas28 I don't use `@dataclass(order=True)` because that doesn't work with ABCs/inheritance.

ehildenb (dismissed):
Pull out the haskell backend submodule update please.

nishantjr:
> Pull out the haskell backend submodule update please.

Oops. fixed.

pyk/src/pyk/kast.py
```diff
@@ -63,6 +63,17 @@ def _check_node(cls: Type[T], d: Dict[str, Any], expected: Optional[str] = None)
         if actual != expected:
             raise ValueError(f"Expected '{expected}' as 'node' value, found: '{actual}'")
 
+    def _as_shallow_tuple(self) -> Tuple[Any, ...]:
+        # shallow copy version of dataclass.astuple.
+        return tuple([self.__dict__[field.name] for field in fields(type(self))])
```
tothtamas28:
```suggestion
        return tuple(self.__dict__[field.name] for field in fields(type(self)))
```

pyk/src/pyk/kast.py
```diff
@@ -210,6 +221,9 @@ def unapply(self, term: KInner) -> KInner:
             new_term = KRewrite(lhs, rhs).replace(new_term)
         return new_term
 
+    def pretty_print(self, kprint) -> str:
```
tothtamas28: What do you think about `pretty` or `pretty_str`? For me, `print` suggests side effects.

pyk/src/pyk/tests/mock_kprint.py
```diff
@@ -0,0 +1,13 @@
+from typing import cast
+
+from ..kast import KAst
+from ..ktool.kprint import KPrint, prettyPrintKast
+
+
+class MockKPrint:
```
tothtamas28: Did you consider implementing this with inheritance?

Also, if this class is used in only a single test, let's make it local to that module.

pyk/src/pyk/tests/test_pretty_print_kast.py
```diff
@@ -10,20 +10,25 @@
...
+    unparser_for_production,
 )
-from pyk.ktool import prettyPrintKast, unparser_for_production
 
 success_production = KProduction(KSort('EndStatusCode'), [KTerminal('EVMC_SUCCESS')], klabel=KLabel('EVMC_SUCCESS_NETWORK_EndStatusCode'))
 
 
 class PrettyPrintKastTest(TestCase):
-    TEST_DATA: Final[Tuple[Tuple[KAst, str], ...]] = (
+    TEST_DATA: Final[Tuple[Tuple[Union[KAst, Subst], str], ...]] = (
```
tothtamas28: Is this weakening required?

pyk/src/pyk/kast.py
nishantjr: I like `pretty`.

pyk/src/pyk/tests/mock_kprint.py
nishantjr: Its used by both the `KCFG` pretty print and `Subst` now.
Oh! Didn't realize we could just not call the constructor. I'm more familiar with mocking in C++.

pyk/src/pyk/kast.py
nishantjr: I've changed this to return an `Iterator[str]` of lines. This makes things like indentation much easier, and matches what we're doing in `KCFG`.

pyk/src/pyk/tests/test_pretty_print_kast.py
nishantjr: Fixed.

pyk/src/pyk/kcfg.py
```diff
@@ -44,17 +44,12 @@ def id(self) -> str:
...
-            return self.id < other.id
-
     class EdgeLike(ABC):
         source: 'KCFG.Node'
         target: 'KCFG.Node'
 
         @abstractmethod
-        def pretty_print(self, kprint: KPrint) -> List[str]:
+        def pretty(self, kprint: KPrint) -> Iterable[str]:
             assert False, 'Must be overridden'
```
```suggestion
            ...
```

tothtamas28 (approved):

pyk/src/pyk/kast.py
tothtamas28: I think the return type should remain `str`. If the string needs further processing, splitting it at newlines before doing all the other stuff should be simple, but e.g. for debugging in an interpreter joining the strings all the time is cumbersome.

nishantjr: There's another functional advantage of `Iterable[str]`. When pretty printing large CFGs, if we use `str` (or `List[str]`) we need to render the entire subtree at each branch before adding indentation. This creates a noticeable lag (I've seen as much as 5 seconds). If we use `Iterable[str]` then we may use generators/yield make `pretty` lazy and allow a better user experience.

PR #2728, opened 2022-07-15
Fix method to_dict for some KAst subclasses

tothtamas28:


pyk/src/pyk/kast.py
```diff
@@ -491,13 +500,14 @@ def match(self, term: KInner) -> Optional[Subst]:
...
-class KRewrite(KInner):
+class KRewrite(KInner, WithKAtt):
     lhs: KInner
     rhs: KInner
 
-    def __init__(self, lhs: KInner, rhs: KInner):
+    def __init__(self, lhs: KInner, rhs: KInner, att=EMPTY_ATT):
         object.__setattr__(self, 'lhs', lhs)
         object.__setattr__(self, 'rhs', rhs)
+        object.__setattr__(self, 'att', att)
```
ehildenb, line 511: Do `KRewrite` have attributes in the frontend?

pyk/src/pyk/kast.py
```diff
@@ -738,14 +767,13 @@ def from_dict(cls: Type['KRegexTerminal'], d: Dict[str, Any]) -> 'KRegexTerminal
...
             regex=d['regex'],
-            precede_regex=d.get('precede_regex', ''),
-            follow_regex=d.get('follow_regex', ''),
+            precede_regex=d['precedeRegex'],
+            follow_regex=d['followRegex'],
         )
 
     def to_dict(self) -> Dict[str, Any]:
-        return {'node': 'KRegexTerminal', 'regex': self.regex, 'precedeRegex': self.precede_regex or None, 'follow_regex': self.follow_regex or None}
+        return {'node': 'KRegexTerminal', 'regex': self.regex, 'precedeRegex': self.precede_regex, 'followRegex': self.follow_regex}
```
ehildenb, line 775: This seems like something we can add a unit test for in `test_kast.py`.

pyk/src/pyk/kast.py
tothtamas28: Yes:

https://github.com/runtimeverification/k/blob/fa7fb0ef91a201896cea6fafc684e39ed1d9ad06/kore/src/main/scala/org/kframework/kore/ADT.scala#L97

And the JSON parser expects it to be present:

https://github.com/runtimeverification/k/blob/fa7fb0ef91a201896cea6fafc684e39ed1d9ad06/kernel/src/main/java/org/kframework/parser/json/JsonParser.java#L381

pyk/src/pyk/kast.py
tothtamas28: You mean to test whether K writes in / is able to parse this format?

ehildenb: Yes, some unit tests of serializing/deserializing things to JSON basically.

If we don't have a unit test harness for that, we should take this as a chance to make one, and add this particular test there. Then in the future as we change the JSON format we can add more tests.

ehildenb (approved):

pyk/src/pyk/kast.py
tothtamas28: `KompiledTest.setUp` now parses `compiled.json` if `KOMPILE_EMIT_JSON` is set to `True`.

PR #2729, opened 2022-07-15
Example project for `kompile --post-process` using `pyk`

tothtamas28:
Related: #2558


radumereuta:
This looks great. I think we should have it committed somewhere in the K repo for future reference. Not in the root directory though. We should add it to the tests as well.
Can you also add a simple tree transformation? For example, rename anonymous variables into named anonymous variables.

You should let Amelie know about this.

k-distribution/tests/regression-new/post-process/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	   --backend haskell            \
+	   --output-definition kompiled \
+	   --emit-json                  \
+	   --post-process 'python3 transform.py'
+
+
+# virtualenv
+
+$(VENV):
+	   virtualenv -p python3.8 $(VENV_DIR) \
```
radumereuta: This line specifies the python version which is not stable on all machines.
For example, I have 3.10
It's going to be really annoying to change it to test on my local machine.
Can we do something about it?
This holds true for the test in the main pyk folder as well.

radumereuta (changes requested):

k-distribution/tests/regression-new/post-process/Makefile
tothtamas28: I agree that moving towards supporting 3.10 as well would make sense (considering it's the Python version Ubuntu 22.04 uses). But right now, Python 3.8 is the only supported version for `pyk`. 

Fortunately, multiple Python versions can be installed on a given system from source without messing with the system's Python paths (`make altinstall`): https://docs.python.org/3/using/unix.html#building-python

Virtualenv then can pick up the specified version from its install location.

Do you consider this a reasonable workaround for now?

radumereuta: Not for regression-new.
This test suite should be for quick testing and it should work on as many machines as possible.
If you move this test to the pyk folder then it should be fine. We already have this issue there.
This way, we keep regression-new worry free. You don't have to know yet another detail.

tothtamas28: Done.

radumereuta (approved):
I can't make this work on my machine.
Ubuntu 22.04 can only build python 3.8.13.
I'm approving this for now, but we should really strive to support a wider range of versions. This is too frustrating.

The code lgtm

PR #2734, opened 2022-07-18
single-query z3 type inference algorithm

dwightguth:


dwightguth:
This avenue of approach hasn't really been fruitful, so I'm closing this PR.

PR #2736, opened 2022-07-19
CODEOWNERS: remove USER_MANUAL from codeowner restrictions

ehildenb:
This means that people reviewing USER_MANUAL file need to pay extra special attention @radumereuta

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,3 @@
 /k-distribution/include/kframework/builtin/* @dwightguth
-/Jenkinsfile @kframework/admin
-/.github/workflows/* @kframework/admin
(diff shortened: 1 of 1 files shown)


radumereuta (approved):

radumereuta:
> This means that people reviewing USER_MANUAL file need to pay extra special attention @radumereuta

Can we add more people as code owners instead of removing them?

dwightguth (approved):
Looks fine, but I wonder if Radu is right. Who all is currently in runtimeverification/admin?

ehildenb:
Seems like something we can have a follow up PR for after discussion. The issue is still on the board.

PR #2738, opened 2022-07-20
Multipe modules in --emit-json-spec

### LINKED ISSUE #2574: [kprove] intermediate claim modules not included in output of --emit-json-spec
Given the following K definition:

```
module TEST
    imports BOOL
    imports INT
    rule <k> 1 => 2 ... </k>
endmodule

module SOME-OTHER
    imports TEST
endmodule

module TEST-INTERMEDIATE
    imports TEST
    claim <k> 3 => 4 ... </k>
endmodule

module TEST-SPEC
    imports TEST-INTERMEDIATE 
    claim <k> 4 => 5 ... </k>
endmodule
```

You can run the following commands:

```
$ kompile --backend haskell test.k --main-module TEST --syntax-module TEST

$ kprovex test.k --spec-module TEST-SPEC --dry-run --emit-json-spec spec.json
kore-exec /home/dev/src/k/./test-kompiled/definition.kore --module TEST --prove /home/dev/src/k/./.kprove-2022-04-30-04-50-34-081-d9e7acef-013b-41e2-ad32-6a0bb24d4dad/spec.kore --spec-module TEST-SPEC --output /home/dev/src/k/./.kprove-2022-04-30-04-50-34-081-d9e7acef-013b-41e2-ad32-6a0bb24d4dad/result.kore

$ cat spec.json | jq .
{
  "format": "KAST",
  "version": 1,
  "term": {
    "node": "KFlatModule",
    "name": "TEST-SPEC",
    "imports": [
      {
        "node": "KImport",
        "name": "TEST-INTERMEDIATE",
        "isPublic": true
      }
    ],
    "localSentences": [
      {
        "node": "KClaim",
        "body": {
          "node": "KApply",
          "label": "<generatedTop>",
          "variable": false,
          "arity": 2,
          "args": [
            {
              "node": "KApply",
              "label": "<k>",
              "variable": false,
              "arity": 1,
              "args": [
                {
                  "node": "KSequence",
                  "arity": 2,
                  "items": [
                    {
                      "node": "KRewrite",
                      "lhs": {
                        "node": "KToken",
                        "sort": "Int",
                        "token": "4"
                      },
                      "rhs": {
                        "node": "KToken",
                        "sort": "Int",
                        "token": "5"
                      },
                      "att": ""
                    },
                    {
                      "node": "KVariable",
                      "name": "_DotVar1",
                      "originalName": "_DotVar1"
                    }
                  ]
                }
              ]
            },
            {
              "node": "KVariable",
              "name": "_DotVar0",
              "originalName": "_DotVar0"
            }
          ]
        },
        "requires": {
          "node": "KToken",
          "sort": "Bool",
          "token": "true"
        },
        "ensures": {
          "node": "KToken",
          "sort": "Bool",
          "token": "true"
        },
        "att": {
          "node": "KAtt",
          "att": {
            "org.kframework.attributes.Location": "Location(24,11,24,30)",
            "org.kframework.attributes.Source": "Source(/home/dev/src/k/test.k)",
            "org.kframework.definition.Production": "syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]"
          }
        }
      }
    ],
    "att": {
      "node": "KAtt",
      "att": {
        "org.kframework.attributes.Location": "Location(21,1,26,10)",
        "org.kframework.attributes.Source": "Source(/home/dev/src/k/test.k)"
      }
    }
  }
}
```

The key here is that we see only one claim (the one from module `TEST-SPEC`), but not the claim from the included module `TEST-INTERMEDIATE`. This is because we are only outputting a single module in JSON. I propose we instead output every module imported by the `--spec-module` but not already included in the `--main-module`, that we there is enough information to reconstruct an entire definition.

This is annoying for the summarizer, because it means that we can only have one module full of claims at a time.

radumereuta:
Emit all spec modules not included in the main module but included by the spec module


radumereuta:
The JSON file looks something like this now.
```
{
  "format": "KAST",
  "version": 2,
  "mainSpecModule": "TEST-SPEC",
  "term": [
    {
      "node": "KFlatModule",
      "name": "TEST-INTERMEDIATE",
      "imports": [
        {
          "node": "KImport",
          "name": "TEST",
          "isPublic": true
        }
      ],
      ....
    },
    {
      "node": "KFlatModule",
      "name": "TEST-SPEC",
      "imports": [
        {
          "node": "KImport",
          "name": "TEST-INTERMEDIATE",
          "isPublic": true
        }
      ],
      ....
    }
  ]
}
```

ehildenb:
Let's add a test please!

radumereuta:
Updated the format a bit:
```
{
  "format": "KAST",
  "version": 2,
  "term": {
    "node": "KFlatModuleList",
    "mainModule": "TEST-SPEC",
    "term": [
      {
        "node": "KFlatModule",
        "name": "TEST-INTERMEDIATE",
        "imports": [
          {
            "node": "KImport",
            "name": "TEST",
            "isPublic": true
          }
        ],
      },
      {
        "node": "KFlatModule",
        "name": "TEST-SPEC",
        "imports": [
          {
            "node": "KImport",
            "name": "TEST-INTERMEDIATE",
            "isPublic": true
          }
        ],
      }
    ]
  }
}
```

kernel/src/main/java/org/kframework/parser/json/JsonParser.java
```diff
@@ -65,6 +65,7 @@ public class JsonParser {
                              , KNONTERMINAL         = "KNonTerminal"
                              , KMODULE              = "KModule"
                              , KFLATMODULE          = "KFlatModule"
+                             , KFLATMODULELIST      = "KFlatModuleList"
                              , KIMPORT              = "KImport"
                              , KPRODUCTION          = "KProduction"
                              , KREGEXTERMINAL       = "KRegexTerminal"
```
ehildenb, line 71: I don't think this is necessary is it? Can't we just use a `KDefinition` here instead?

ehildenb: Or, if we do it this way, then both the Java code and the Python code should be modified so that a `KDefinition` contains a `KFlatModuleList` instead of a `List[KFlatModule]`.

But I still think we should just use a `KDefinition` here, because then we get all the facilities that `KDefinition` has (like `syntax_productions` and `functions` methods).

radumereuta: The problem with using a `KDefinition` is that it recursively serializes all included modules, along with ensuring other soundness checks.
For this particular case, we want to serialize only the modules included by the main-spec-module but not by the main-def-module. So `KFlatModuleList` is just that, a list of modules.
We do a similar thing in Java in kprove: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kprovex/ProofDefinitionBuilder.java#L51
I had to introduce the new name because of how the serializer and deserializer is built. At each node in the AST I need to make a decision about how to deserialize the rest of the term.

ehildenb (approved):

PR #2741, opened 2022-07-21
Add Everett as a codeowner to /builtin

### LINKED ISSUE #2727: Update code owner files
We should add more code owners to the K repo so we don't depend on one person.

radumereuta:
@ehildenb or @dwightguth please review.
Hopefully, this makes it so at least one approval is required, and not both of you.


radumereuta:
After further reading, this will require both approvals. (haven't tested this)
The proper fix would be to create a team with Everett and Dwight in it and add that team as a codeowner.

radumereuta:
I'm reading conflicting info. Let's try it this way.

dwightguth (approved):

PR #2742, opened 2022-07-22
pyk: KPrint's symbol_table should type as a Dict so we can update it.

nishantjr:


ehildenb:
Subsumed by: https://github.com/runtimeverification/k/pull/2756

PR #2743, opened 2022-07-22
Drop the tabulate dependency

### LINKED ISSUE #2739: PYK library drop the tabulate dependency
Right now the latest LTS version of Ubuntu is 22.04 which comes by default with python3.10.
The pyk library only works with python3.8.
Even `python3.8.13+` fails because it requires `tabulate==0.8.10`
Ideally, the makefiles should be more flexible and not call for a specific version.

tothtamas28:
Fixes #2739


pyk/src/pyk/ktool/kprove.py
```diff
@@ -1,3 +1,4 @@
+import csv
```
tothtamas28, line 1: Replaced `tabulate` with `csv`. The profile csv file can then be easily turned into a table by external tools. (Maybe we could even leave data processing / aggregation to external tools, and just write the raw profile data.)

ehildenb (approved):

PR #2744, opened 2022-07-22
Add kore-rpc client

tothtamas28:
Blocked on: https://github.com/runtimeverification/haskell-backend/pull/3146

TODOs:
* [ ] Unit tests
* [ ] Integration tests
* [ ] Consider implementing non-blocking behaior with `asyncio`
* [ ] Implement the `cancel` endpoint (only makes sense for non-blocking behavior)

### DIFF
--- a/pyk/src/pyk/kore/client.py
+++ b/pyk/src/pyk/kore/client.py
@@ -110,10 +110,12 @@ class JsonRpcClient(ContextManager['JsonRpcClient']):
 @dataclass(frozen=True)
 class KoreClientError(Exception):
     message: str
+    code: int
     context: Tuple[str, ...]
 
-    def __init__(self, message: str, context: Sequence[str] = ()):
+    def __init__(self, message: str, code: int, context: Sequence[str] = ()):
         object.__setattr__(self, 'message', message)
+        object.__setattr__(self, 'code', code)
         object.__setattr__(self, 'context', tuple(context))
         super().__init__(message)
 
@@ -249,7 +251,7 @@ class KoreClient(ContextManager['KoreClient']):
             return self._client.request(method, **params)
         except JsonRpcError as e:
             assert e.code != -32602, 'Malformed request'
-            raise KoreClientError(message=e.data['error'], context=e.data['context'])
+            raise KoreClientError(message=e.data['error'], code=e.code, context=e.data['context'])
 
     @staticmethod
     def _state(pattern: Pattern) -> Dict[str, Any]:
--- a/pyk/src/pyk/kore/client.py
+++ b/pyk/src/pyk/kore/client.py
@@ -103,6 +103,7 @@ class JsonRpcClient(ContextManager['JsonRpcClient']):
         if 'error' not in response:
             return
 
+        assert response['error']['code'] not in {-32700, -32600}, 'Malformed JSON-RPC request'
         raise JsonRpcError(**response['error'])
 
 
@@ -250,7 +251,7 @@ class KoreClient(ContextManager['KoreClient']):
         try:
             return self._client.request(method, **params)
         except JsonRpcError as e:
-            assert e.code != -32602, 'Malformed request'
+            assert e.code not in {-32601, -32602}, 'Malformed Kore-RPC request'
             raise KoreClientError(message=e.data['error'], code=e.code, context=e.data['context'])
 
     @staticmethod
--- a/pyk/src/pyk/kore/client.py
+++ b/pyk/src/pyk/kore/client.py
@@ -6,6 +6,7 @@ from dataclasses import InitVar, dataclass
 from enum import Enum
 from typing import (
     Any,
+    ContextManager,
     Dict,
     Final,
     Iterable,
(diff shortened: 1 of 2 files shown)


ehildenb:
One thing that I see missing is the ability to handle some sort of "error" return state, where the Haskell backend may need to return that the query failed for some reason (z3 died, out of memory, out of disk, whatever....).

Another issue is that when checking satisfiabilities of implications (or anything), there really should be three responses: sat, unsat, unknown. If we can't have the unknown value, we need to have the above error ^^^ so that there is some way for the backend to signal unknown to the user.

For example, you make have `P1 #Implies P2`, and the haskell backend can compute "if substitution `S` is applied to `P2`, then it _could_ be satisfiable, but I still don't know if the result is satisfiable." In this case, wewant the substitution and predicate, but also the knowledge that the result is potentially unsat still. Falling back to the error case above would work here at least to signal the outside application that we don't know sat/unsat, but maybe a more direct representation of that is appropriate.

@goodlyrottenapple thoughts?

tothtamas28:
> One thing that I see missing is the ability to handle some sort of "error" return state, where the Haskell backend may need to return that the query failed for some reason (z3 died, out of memory, out of disk, whatever....).

In such a case `KoreClient` raises `KoreClientError`. What we could do is - given the server specifies what sort of error responses it may return - refine this class to raise more specific exceptions.

pyk/src/pyk/kore/client.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def close(self) -> None:
+        self._client.close()
+
+    def _request(self, method: str, **params: Any) -> Dict[str, Any]:
+        try:
+            return self._client.request(method, **params)
+        except JsonRpcError as e:
+            assert e.code != -32602, 'Malformed request'
+            raise KoreClientError(message=e.data['error'], code=e.code, context=e.data['context'])
```
ehildenb: I see here the `KoreClientError` but I guess this is only for actually bad input data. Can it be extended to also handle cases like "BackendCrashed" or "SatisfiabilityUnknown"?

tothtamas28: Whenever the server sends an error response, this exception will be raised. If both server and client implementations are correct, the error code will be either `-32603 Internal error` or `-32000..-32099 Server error`, The client has assertions to check this, as it should never send a malformed request, which all the other error codes correspond to.

Both `BackendCrashed` and `SatisfiabilityUnknown` would fall into the `Server error` category. I'd say however that `SatisfiabilityUnkonwn` should be a `result`, and not an `error` response.

tothtamas28: Here's an actual stack trace.

```
>>> from pyk.kore.syntax import *
>>> from pyk.kore.client import *
>>> import logging
>>> logging.basicConfig(level=logging.INFO)
>>> client = KoreClient('localhost', 3000)
INFO:pyk.kore.client:Connected to host: localhost:3000
>>> client.execute(Top(SortApp('IntSort')))
INFO:pyk.kore.client:Sending request to localhost:3000: {"jsonrpc": "2.0", "id": 1, "method": "execute", "params": {"state": {"format": "KORE", "version": 1, "term": {"tag": "Top", "sort": {"tag": "SortApp", "name": "IntSort", "args": []}}}}}
INFO:pyk.kore.client:Waiting for response from localhost:3000...
INFO:pyk.kore.client:Received response from localhost:3000: {"error":{"data":{"context":["\\top (<unknown location>)","sort 'IntSort' (<unknown location>)","(<unknown location>)"],"error":"Sort 'IntSort' not defined."},"code":-32002,"message":"Could not verify KORE pattern"},"jsonrpc":"2.0","id":1}

Traceback (most recent call last):
  File "/home/ttoth/git/k/pyk/src/pyk/kore/client.py", line 252, in _request
    return self._client.request(method, **params)
  File "/home/ttoth/git/k/pyk/src/pyk/kore/client.py", line 96, in request
    self._check(data)
  File "/home/ttoth/git/k/pyk/src/pyk/kore/client.py", line 107, in _check
    raise JsonRpcError(**response['error'])
pyk.kore.client.JsonRpcError: Could not verify KORE pattern

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ttoth/git/k/pyk/src/pyk/kore/client.py", line 280, in execute
    result = self._request('execute', **params)
  File "/home/ttoth/git/k/pyk/src/pyk/kore/client.py", line 255, in _request
    raise KoreClientError(message=e.data['error'], code=e.code, context=e.data['context'])
pyk.kore.client.KoreClientError: Sort 'IntSort' not defined.
```

goodlyrottenapple: I agree with @tothtamas28 on `SatisfiabilityUnknown` being a result. As to the user errors, currently we have two user error codes, namely:
- -32001 Cancel request unsupported in batch mode
- -32002 Could not verify KORE pattern

I will start adding them to the API doc in a separate section to keep a track of them.

ehildenb: Excellent, thank you @goodlyrottenapple . I'll approve this PR, @tothtamas28 feel free to merge when you're ready!

ehildenb (approved):

PR #2751, opened 2022-07-27
Add back error for function simplification rules

### LINKED ISSUE #2750: [Bug] [kompile] - simplification rules for constructors shouldn't be allowed
K Version
---------

```
$ kompile --version
K version:    v5.3.125-0-gdf4dc1a50a
Build date:   Tue Jul 26 13:41:04 EEST 2022
```

Description
-----------

`kompile` succesfully kompiles the definition.

Input Files
-----------

File `test.k`:
```
module TEST

    syntax MySort ::= constr1() | constr2()

    rule constr1() => constr2() [simplification]

endmodule
```

Reproduction Steps
------------------

```

```

Expected Behavior
-----------------

The frontend should error when encountering simplification rules with constructor symbols at the top.

radumereuta:
Fixes: #2750
This was removed [here](https://github.com/runtimeverification/k/pull/2711/files#diff-487106080746c9047c18b3addae0c79a7a377a2e1c517ac087bfd17eecf38b7eL455)
But after a discussion with @ana-pantilie we think it should still be checked.

### DIFF
--- a/k-distribution/tests/regression-new/issue-2287-simpl-rules-in-kprovex/a2-spec.k.out
+++ b/k-distribution/tests/regression-new/issue-2287-simpl-rules-in-kprovex/a2-spec.k.out
@@ -1,9 +1,9 @@
-[Error] Compiler: Simplification rules need to be function/functional/mlOp like.
+[Error] Compiler: Simplification rules expect function/functional/mlOp symbols at the top of the left hand side term.
 	Source(a2-spec.k)
 	Location(7,8,7,55)
 	7 |	  rule <k> (X +Int Y) +Int Z => Z +Int (X +Int Y) </k> [simplification, symbolic(X), concrete(Y, Z)]
 	  .	       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-[Error] Compiler: Simplification rules need to be function/functional/mlOp like.
+[Error] Compiler: Simplification rules expect function/functional/mlOp symbols at the top of the left hand side term.
 	Source(a2-spec.k)
 	Location(8,8,8,25)
 	8 |	  rule notAFunction => 1 [simplification]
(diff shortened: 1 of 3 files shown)


k-distribution/tests/regression-new/issue-2287-simpl-rules-in-kprovex/a2-spec.k.out
```diff
@@ -0,0 +1,11 @@
+[Error] Compiler: Simplification rules need to be function/functional/mlOp like.
```
ana-pantilie: I'd make the error message more exact:
```suggestion
[Error] Compiler: Simplification rules expect function/functional/mlOp symbols at the top of the left hand side term.
```

ana-pantilie (approved):
I have a suggestion about the error message, otherwise LGTM. Note that I don't know the details of the Java code, so probably someone else would be better at reviewing that 😁

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -447,6 +447,15 @@ public void proverChecksX(Module specModule, Module mainDefModule) {
                 errors.add(KEMException.compilerError("Found syntax declaration in proof module. Only tokens for existing sorts are allowed.", s));
 
         ModuleTransformer mt = ModuleTransformer.fromSentenceTransformer((m, s) -> {
+            if (s instanceof Rule && (s.att().contains(Att.SIMPLIFICATION()))) {
+                KLabel kl = m.matchKLabel((Rule) s);
+                scala.collection.Set<Production> prods = m.productionsFor().get(kl).getOrElse(() -> {
+                    throw KEMException.criticalError("Could not find productions for label " + kl.name(), s);
+                });
```
ehildenb: This is probably a somewhat common code pattern? Could we write a function called `getUniqueProductionFor(kl)`, which encapsulates all this logic? Basically it calls `productionsFor().get(kl).getOrElse(() -> { error })` and then asserts that the length of the returned productions in exactly 1, and then returns that unique production? I'm guessing that there are many places in the code using this pattern.

ehildenb (approved):

ehildenb:
Looks good. One suggestion (can be a separate issue).

kernel/src/main/java/org/kframework/kompile/Kompile.java
radumereuta: Found a way that's even more flexible.

Issue #2627, opened 2022-05-26
Include pyk in debian package


ehildenb:
All uses of pyk currently rely on the git submodule being present. `This is suitable for development, but for packaging we should be deploying actual packages to PyPi.`

- ~[ ] Add an RV account to PyPi.~
- ~[ ] Deploy pyk to PyPi on CI deployments.~
- [ ] Add a line in [here](https://github.com/runtimeverification/k/blob/master/package/package) to build the pyk virtualenv and copy it into `DESTDIR` at the appropriate location. Downstream CI doesn't need to build it.
- [ ] Adjust downstream CI to use this deployment instead of the submodule.
- [x] https://github.com/runtimeverification/devops/issues/51

ehildenb:
We can call it `kframework` on PyPi, so people can do `pip install kframework`. Or `lib-kframework` or `kframework-pyk` or something.

No need to commit to `pyk` name.

radumereuta:
We might not need virtualenv for deployments. We can discuss this after the first iteration.

PR #2753, opened 2022-07-27
Include pyk in Focal package

ehildenb:
This includes the pyk library in the Focal package. I do not think this is the most widely accepted way to package python dependencies, but I have verified that the package installs and uninstalls cleanly, and added tests on CI that python is picking it up correctly. So this should at least enable downstream to start using pyk like a proper package, and then we can clean up the packaging process later.

This only enables it for Ubuntu Focal for the moment, just so we can begin testing this with downstream repos KEVM and ksummarize.

Part of: https://github.com/runtimeverification/k/issues/2627

### DIFF
--- a/package/debian/rules
+++ b/package/debian/rules
@@ -28,7 +28,7 @@ override_dh_auto_build:
 
 override_dh_auto_install:
 	package/package
-	@mkdir -p $(DESTDIR)$(PREFIX)/lib/python3/dist-packages
+	@mkdir -p $(DESTDIR)$(PREFIX)/lib/$(PYTHON_DEB_VERSION)/dist-packages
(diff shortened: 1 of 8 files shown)


dwightguth (approved):

package/debian/rules
```diff
@@ -13,15 +13,23 @@ export DEB_BUILD_MAINT_OPTIONS=hardening=-stackprotector
...
 	dh $@ 
 
 override_dh_auto_build:
-	mvn --batch-mode package -DskipTests -Dllvm.backend.prefix=/usr -Dllvm.backend.destdir=$(shell pwd)/debian/kframework
+	mvn --batch-mode package -DskipTests -Dllvm.backend.prefix=$(PREFIX) -Dllvm.backend.destdir=$(DESTDIR)
 
 override_dh_auto_install:
-	DESTDIR=$(shell pwd)/debian/kframework PREFIX=/usr package/package
+	package/package
+	@mkdir -p $(DESTDIR)$(PREFIX)/lib/python3/dist-packages
```
ehildenb:
```suggestion
	@mkdir -p $(DESTDIR)$(PREFIX)/lib/$(PYTHON_DEB_VERSION)/dist-packages
```

PR #2754, opened 2022-07-28
Duplicate comm simplification rules

### LINKED ISSUE #2587: [KIP] - `comm` attribute on `simplification` rules
Motivation
----------

Now that the user can write simplification rules over the `\equals` KORE construct, we frequently find ourselves in situations where we write rules like the following:

```
rule { A #Equals B } => C [simplification]
rule { B #Equals A } => C [simplification]
```

While it might be nice for the backend to recognize these are in fact the same axiom (since equals is commutative), it would be difficult in practice to make the equation matcher in the backend handle this, and would also likely be slower than simply using both forms of the equation.

We would like instead to be able to automatically generate the second form of the rule from the first. This makes it easier to specify both forms of the rule, which means we are less likely to run into reasoning errors where we forgot to specify the second form, which in turn will lead to faster verification.

Example K Code
--------------

Instead of writing:

```
module LEMMAS
  rule { I +Int J #Equals I} => #Bottom
    requires J =/=Int 0 [simplification]
  rule { I #Equals I +Int J} => #Bottom
    requires J =/=Int 0 [simplification]
endmodule
```

One would write:

```
module LEMMAS
  rule { I +Int J #Equals I} => #Bottom
    requires J =/=Int 0 [simplification, comm]
endmodule
```

Documentation
-------------

If the `comm` attribute is added to a rule, then it means that the top symbol on the left hand side, being commutative, should match in either order of operands. In order to implement this, a second rule is generated by the compiler which swaps the order of operands but otherwise leaves the axiom unchanged. It is an error if the rule does not have the `simplification` attribute or if the top symbol on the LHS does not have the `comm` attribute.

Testing Approach
----------------

A simple test to ensure both forms of the lemma simplify is all that is required.

radumereuta:
Fixes: #2587


radumereuta:
Failing with this message. Coming from the java backend: https://github.com/runtimeverification/k/blob/master/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java#L176
```
     [exec] [Error] Critical: Unexpected combination of assoc, comm, idem attributes found.
     [exec] Currently only sets, maps, and lists are supported: syntax {Sort1, Sort2} Sort2
     [exec] ::= "{" Sort1 "#Equals" Sort1 "}" [comm, format(%1%i%n%2%d%n%3%i%n%4%d%n%5),
     [exec] klabel(#Equals), mlEquals, mlOp, symbol]
     [exec] 	Source(/home/jenkins-runner/workspace/K_PR-2754/k-distribution/target/release/k/include/kframework/builtin/kast.md)
     [exec] 	Location(148,35,148,148)
     [exec] 	148 |	                                | "{" Sort1 "#Equals" Sort1 "}" [klabel(#Equals), symbol, mlOp, mlEquals, comm, format(%1%i%n%2%d%n%3%i%n%4%d%n%5)]
```

radumereuta:
Hack the Java backend to remove the `comm` attribute in Definition.java.
Only if there is no assoc or idem present.
If we find issues with the LLVM backend we should talk again.
We should test it with some larger semantics as well.

kernel/src/main/java/org/kframework/compile/ResolveComm.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public K genCommRule(K body, Module m) {
+        return new RewriteAwareTransformer(true) {
+            @Override
+            public K apply(KApply k) {
+                if (k.klabel().name().equals("#withConfig")) {
+                    super.apply(k);
+                    return k;
+                }
+                if ((isRHS() && !isLHS()) || k.klabel() instanceof KVariable || !m.attributesFor().contains(k.klabel())) {
```
ehildenb, line 65: Is this only swapping at the top-most position of the LHS, or at every position with a `comm` symbol? We should only swap at the top-most position.

For example, if we have:

```
    rule { 1 +Int 2 #Equals 3 } => #Top [comm, simplification]
```

It should generate rule:

```
    rule { 3 #Equals 1 +Int 2 } => #Top
```

But not rule:

```
    rule { 3 #Equals 2 +Int 1 } => #Top
```

ehildenb: Maybe this could be added as a test somehow.

ehildenb (approved):

dwightguth (changes requested):
I'd definitely like to see tests for this.

radumereuta:
Try to remove the `comm` attribute from the rules generated by this step.

radumereuta:
I need to test it to see what happens if I add `comm` to more productions since the backend generates extra axioms for each.

dwightguth (approved):
looks fine. Adding the commutativity axioms to the kore definition shouldn't affect anything though since both backends appear to be ignoring them.

ehildenb:
Please add an issue to identify what's going on with the backend, so we can get that fixed too. The backends should not have to have the `comm` attribute stripped.

PR #2756, opened 2022-07-28
Upstreaming pyk functionality

ehildenb:
Pulling issue from this list: https://github.com/runtimeverification/ksummarize/issues/82

- Move a test that doesn't actually need a kompiled definition from integration tests to unit tests.
- Adds methods `module_names`, `production_for_klabel`, `production_for_cell_sort`, and `empty_config` to `KDefinition`.
- The type `SymbolTable` is now an alias for `Dict` instead of `Mapping`, which makes it behave nicer when the symbol table needs to be patched.
- Factor out method `build_claim` and `KCFG.Edge.to_claim`, instead of passing a boolean into `build_rule`.
- Adds class `Bool` to the `pyk.prelude`, which has helpers for constructing common `Bool` patterns.
- Rename method `simplifyBool => simplify_bool`.

### DIFF
--- a/pyk/src/pyk/prelude.py
+++ b/pyk/src/pyk/prelude.py
@@ -1,6 +1,7 @@
 from typing import Final, Iterable, Optional, Union, final
 
-from .kast import TRUE, KApply, KInner, KLabel, KSort, KToken
+from .kast import FALSE, TRUE, KApply, KInner, KLabel, KSort, KToken
+from .utils import unique
 
 
 @final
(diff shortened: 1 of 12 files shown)


pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -150,3 +150,19 @@ def test(self):
 
         # Then
         self.assertEqual(config_actual, config_expected)
+
+
+class BooleanTest(TestCase):
+
+    def test_bool_simplify(self):
```
ehildenb:
```suggestion
    def test_simplify_bool(self):
```

pyk/src/pyk/prelude.py
```diff
@@ -24,6 +25,36 @@ def __init__(self):
...
+    def orBool(items: Iterable[KInner]) -> KInner:
+        return build_assoc(Bool.false, KLabel('_orBool_'), unique(items))
+
+    @staticmethod
+    def notBool(item: KInner) -> KApply:
+        return KApply(KLabel('notBool_'), [item])
+
+    @staticmethod
+    def impliesBool(antecedent: KInner, consequent: KInner) -> KApply:
+        return KApply(KLabel('_impliesBool_'), [antecedent, consequent])
```
ehildenb, line 55: How should we stylize these? Should we just do `Bool.and`? Will that work in python or be confused with `and` operator.

tothtamas28: Unfortunately, `Bool.and` wouldn't work. I think this current naming is fine.

tothtamas28: But I think we should consider implementing this as a module. That way, depending on context, we can both

```python3
from prelude import bool

bool.andBool(...)
```

and

```python3
from prelude.bool import andBool

andBool(...)
```

tothtamas28 (approved):

pyk/src/pyk/prelude.py
ehildenb: By a module you mean a new file like `pyk/prelude/bool.py`?

tothtamas28: Yes. exactly.

PR #2758, opened 2022-07-29
[nix] Refactored submodule check via rv-nix-tools

goodlyrottenapple:
Added a new CI workflow `Update / Nix flake submodule sync` which checks if flake inputs match the submodules. If not, the CI updates the lock file and pushes a new commit to the current branch with the update. Not sure if this is the best way to do this, but it's the only way to keep the nix flake in sync with the submodules unless we add nix to the Jenkins runner which updates the submodule dependencies.


.github/workflows/update.yml
```diff
@@ -52,3 +52,66 @@ jobs:
...
+      - id: config
+        run: |
+          ref=${{ github.ref }}
+          if [ "${{ github.event_name }}" == 'pull_request' ]; then
+            ref="${{ github.event.pull_request.head.sha }}"
+          fi
+          ref="${ref#refs/heads/}"
+          echo "::set-output name=ref::$ref"
+          git config --global user.name github-actions
+          git config --global user.email github-actions@github.com
```
F-WRunTime, line 69: Might be better if the commits come from a runtimeverification email. I'd recommend devops@runtimeverification.com

F-WRunTime (approved):

Baltoli (approved):

radumereuta (approved):

PR #2761, opened 2022-07-29
Fix release, add pyk to Jammy package

ehildenb:


ehildenb:
Evidence this works: https://ci.runtimeverification.com/jenkins/blue/organizations/jenkins/K/detail/PR-2761/7/pipeline/

ehildenb:
Evidence this works https://ci.runtimeverification.com/jenkins/blue/organizations/jenkins/K/detail/PR-2761/8/pipeline/

dwightguth (approved):
This looks fine, but you are aware that the build-depends of an Ubuntu package don't get installed on the user's machine, right?

ehildenb:
Yes, I'm aware. As far as I know, these packages are only needed for building the tool (specifically for setuptools, which is what handles building the python project).

I may be incorrect, but we won't know until we downstream test. Just want to get the release working.

Issue #2762, opened 2022-07-30
[Bug] pyk and kast disagree on kore terms


ehildenb:
K Version
---------

```
$ kompile --version
K version:    5.3.0
Build date:   Sat Jul 30 00:45:20 UTC 2022
```

Actually version e72bb80786 with some modifications to pyk.

Description
-----------

`pyk` is writing a Kore term to text that the Frontend does not want to parse.

Input Files
-----------

File `test.k`:

```
requires "domains.md"

module TEST
    imports INT
    imports MAP
    imports BOOL

    configuration <k> $PGM:KItem </k>
                  <state> .Map </state>

    syntax Bool ::= pred1 ( Int ) [function, functional, klabel(pred1), symbol, no-evaluators]
    syntax KItem ::= "foo" | "bar"

    rule <k> foo => bar ... </k>
         <state> ... 3 |-> N ... </state>
      requires pred1(N)

endmodule
```

File `test1.kore`:

```
inj { SortBool { }, SortKItem { } } ( Lblpred1 { } ( \dv { SortInt { } } ( "3" ) ) ) 
```

File `test2.kore`:

```
inj { SortBool , SortKItem } ( Lblpred1 { } ( \dv { SortInt } ( "3" ) ) )
```

Reproduction Steps
------------------

Kompile the definition:

```
+ kompile --backend haskell test.k --output-definition test-kompiled
```

Kast the first file:

```
+ kast --input kore test1.kore --definition test-kompiled
[Error] Critical: Parse error
 (ParseError: ERROR: Line 1: Column 16: Expected '}' or ',', but '{'
inj { SortBool { }, SortKItem { } } ( Lblpred1 { } ( \dv { SortInt { } } ( "3"
) ) )
               ^)
```

Kast the second file:

```
+ kast --input kore test2.kore --definition test-kompiled
pred1(#token("3","K"))
```

Expected Behavior
-----------------

The file `test1.kore` was generated by `pyk` on the KAST term `pred1(#token('3', 'Int'))`. When removing the `{}` on `SortBool`, `SortKItem`, and `SortInt{}`, the frontend the allows it (that's `test2.kore`).

So either `pyk` should not be generating the kore term with the extra curlys, or the frontend should be able to parse the first file.

Additionally, the frontend is giving back `pred1(#token('3', 'K'))`, when it should be giving back `pred1(#token('3', 'Int'))`. So somehow it's losing sort information here.

PR #2763, opened 2022-07-30
Initial conversion facilities between Kast and Kore in pyk

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/issues/2762~

This PR does several things:

- Adjusts `run_process` in `cli_utils` to always propogate stderr.
- Updates the argument parser for `pyk` main functionality to have a factored out shared arguments.
- Allows multiple `--verbose` options like `-vvv`.
- Adds methods `kast_to_kore` and `kore_to_kast` which use the frontend `kast` binary for doing the conversion to `KPrint`.
- Adds a test of `kast_to_kore` and `kore_to_kast`, which identified the issue runtimeverification/k#2762.

Currently, the test is adjusted to work around issue runtimeverification/k#2762, so this should not be merged until that one is resolved. Ideally we would have one set of pairs of `kore, kast`, and test conversion both directions. Right now we have to maintain two separate pairs.


pyk/src/pyk/kore/syntax.py
```diff
@@ -255,7 +255,7 @@ def dict(self) -> Dict[str, Any]:
 
     @property
     def text(self) -> str:
-        return self.name + ' ' + _braced(sort.text for sort in self.sorts)
+        return self.name
```
ehildenb: This change in how `SortApp.text` works is needed in order to get these tests passing, because the frontend does not want to parse sorts with `{}` behind them: https://github.com/runtimeverification/k/issues/2762

ehildenb: Thankfully this breaks some tests in the kore unparser test-suite, so that's good :)

pyk/src/pyk/integration_tests/test_kore_to_kast.py
```diff
@@ -0,0 +1,30 @@
...
+    KOMPILE_EMIT_JSON = True
+
+    KPROVE_USE_DIR = '.simple-proof-test'
+
+    @staticmethod
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_kast_to_kore(self):
+        # TODO: We want to not have to store 2 lists, and want to test (kast_to_kore(kore_to_kast(kore)) == kore)
```
ehildenb:
```suggestion
```

pyk/src/pyk/__main__.py
```diff
@@ -85,39 +89,44 @@ def main():
...
-    pykArgs = argparse.ArgumentParser()
-    pykArgs.add_argument('kompiled-dir', type=str, help='Kompiled directory for definition.')
-    pykArgs.add_argument('--verbose', '-v', default=False, action='store_true', help='Set log level to INFO.')
 
-    pykCommandParsers = pykArgs.add_subparsers(dest='command')
+    shared_options = argparse.ArgumentParser(add_help=False)
+    shared_options.add_argument('-v', '--verbose', action='count', help='Verbosity level, repeat for more verbosity.')
+
+    kompiled_definition_options = argparse.ArgumentParser(add_help=False)
+    kompiled_definition_options.add_argument('kompiled-dir', type=str, help='Kompiled directory for definition.')
```
nishantjr: Lets use `--definition` or `'definition` here for consistency with `krun` and friends.

```suggestion
    kompiled_definition_args.add_argument('definition', type=str, help='Kompiled directory for definition.')
```

pyk/src/pyk/__main__.py
```diff
@@ -85,39 +89,44 @@ def main():
...
 
 
 def create_argument_parser():
-    pykArgs = argparse.ArgumentParser()
-    pykArgs.add_argument('kompiled-dir', type=str, help='Kompiled directory for definition.')
-    pykArgs.add_argument('--verbose', '-v', default=False, action='store_true', help='Set log level to INFO.')
 
-    pykCommandParsers = pykArgs.add_subparsers(dest='command')
+    shared_options = argparse.ArgumentParser(add_help=False)
+    shared_options.add_argument('-v', '--verbose', action='count', help='Verbosity level, repeat for more verbosity.')
```
nishantjr: Lets use `_args` or `_options` consistently. 

```suggestion
    logging_args = argparse.ArgumentParser(add_help=False)
    logging_args.add_argument('-v', '--verbose', action='count', help='Verbosity level, repeat up to two times for more verbosity.')
```

pyk/src/pyk/__main__.py
```diff
@@ -27,24 +28,25 @@ def main():
...
     if args['command'] == 'print':
         printer = KPrint(kompiled_dir)
+        _LOGGER.info(f'Reading Kast from file: {args["term"].name}')
         term = KAst.from_json(args['term'].read())
         if type(term) is dict and 'term' in term:
             term = term['term']
-        if term == mlTop():
-            args['output_file'].write(printer.pretty_print(term))
-        else:
+        if term != mlTop():
```
nishantjr: Is this correct when the sort of `term` is not `K`?

pyk/src/pyk/ktool/kprint.py
```diff
@@ -49,6 +58,31 @@ def __init__(self, kompiled_directory: str):
         self.symbol_table = build_symbol_table(self.definition, opinionated=True)
         self.definition_hash = hash_str(self.definition)
 
+    def kore_to_kast(self, kore: Kore) -> KAst:
+        with NamedTemporaryFile('w') as ntf:
+            ntf.write(kore.text)
+            ntf.flush()
+            _LOGGER.info(f'Wrote file: {ntf.name}')
+            kast_command = ['kast', '--input', 'kore', '--output', 'json', ntf.name, '--definition', str(self.kompiled_directory)]
+            command_env = os.environ.copy()
```
nishantjr, line 67: What does this do? Would it be OK to not specify the `env` and let it inherit from the current process?

pyk/src/pyk/integration_tests/test_kore_to_kast.py
```diff
@@ -0,0 +1,29 @@
...
+    @staticmethod
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_kast_to_kore(self):
+        kore_kast_pairs = (
+            (App('inj', [SortApp('SortBool'), SortApp('SortKItem')], [App('Lblpred1', [], [DV(SortApp('SortInt'), String('3'))])]), KApply('pred1', [intToken(3)])),
+        )
+        for i, (kore, kast) in enumerate(kore_kast_pairs):
+            with self.subTest(i=i):
```
nishantjr: Consider naming each `subTest`. This is important for documenting the intention of each test, and what aspect of the functionality we're testing.

```suggestion
        tests = (
            ('injections',
             App('inj', [SortApp('SortBool'), SortApp('SortKItem')], [App('Lblpred1', [], [DV(SortApp('SortInt'), String('3'))])]),
             KApply('pred1', [intToken(3)])
            ),
        )
        for i, (name, kore, kast) in enumerate(kore_kast_pairs):
            with self.subTest(name, i=i):
```

nishantjr (changes requested):

pyk/src/pyk/ktool/kprint.py
ehildenb: You get an empty env if you don't specifically do this inheritance. This is how you pass the env along.

nishantjr: Are you sure? If `PATH` wasn't passed along, that would make the default pretty useless. By not specify, I mean pass `None`, or don't provide and argument, not pass in the empty dict.

From `subprocess.run`'s documentation, it seems that the default is to inherit:
> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment. It is passed directly to Popen.

pyk/src/pyk/__main__.py
ehildenb: I don't know, but that also seems an unrelated question ot this refactoring. This refactoring preserves old behavior, but makes it so we only have one write statement to the `args['output']`. Is there a better way to check for `mlTop()`? I can use that instead, but anything beyond that should go in it's own issue (unrelated to this PR).

pyk/src/pyk/integration_tests/test_kore_to_kast.py
ehildenb: We don't name subtests anywhere else, and I am not convinced it's a good idea. I haven't had any problems finding the problematic subtests in other examples.

I also am just trying to set up the testing harness here, so that as we discover bugs we have a fast way to make tests for them which does not involve invoking the entire codebase or making a new harness. So I'm not trying to make it perfect this time around.

This test, for example, has no sensible name. It's just the first breakage I ran into trying to use `kast_to_kore` and `kore_to_kast`, so I added it as a test.

pyk/src/pyk/ktool/kprint.py
ehildenb: Huh, I thought we were discarding it because we had exactly the `PATH` problem before, so we started manually inheriting. But it appears that maybe this is a breaking in `run_process`, where it should default to inheriting instead of overriding: https://github.com/runtimeverification/k/blob/a3b3f34bb6d08179ee5bc7d3489d8f98b0f46bd6/pyk/src/pyk/cli_utils.py#L52.

Issue made: https://github.com/runtimeverification/k/issues/2769

This change (that the issue describes) will end up making a lot more changes than just here, so I'd rather it be a separate PR.

pyk/src/pyk/__main__.py
ehildenb: Done.

pyk/src/pyk/__main__.py
ehildenb: Done.

pyk/src/pyk/cli_utils.py
```diff
@@ -51,8 +51,12 @@ def run_process(
     try:
         res = subprocess.run(args, input=input, env=env, capture_output=True, check=True, text=True)
     except CalledProcessError as err:
+        sys.stderr.write(err.stderr)
+        sys.stderr.flush()
         logger.log(log_level, f'Completed with status {err.returncode}: {command}')
         raise
+    sys.stderr.write(res.stderr)
+    sys.stderr.flush()
```
tothtamas28: I don't think this is a good idea in general. In certain cases I guess it is, but then the caller can do the printing.

But we can also add a boolean argument with a default value to the function to enable this feature when needed.

ehildenb: How could this be a bad idea though? If a process we are calling produces stderr, we want to be able to see that. I can put it behind a flag, but I think it should be turned on by default, it's harder to track down where an error is coming from if this isn't output.

ehildenb: I added a flag for disabling it if the process succeeds. If we notice it becomes really annoying for something, we can change it again later.

The motivation for this is that the ksummarizer was generating rules for a long time with incorrect variable naming conventions, but I didn't know at all because stderr of the prover was being dropped silently (I had no hint things were wrong). These were warnings I wanted to fix directly, not ignore.

pyk/src/pyk/__main__.py
nishantjr: OK, we should be checking the label of the KApply.

pyk/src/pyk/integration_tests/test_kore_to_kast.py
nishantjr: I think `issue-1234` (or whatever the issue number is) is a good name. Not having names makes refactoring the test suite difficult later, because we don't know the intention of the tests.

pyk/src/pyk/cli_utils.py
nishantjr: I think using `subprocess.check_output(stderr=None, ...)` would get you this behaviour, with the subprocess printing immediately rather than upon completion. We can add `stderr` as an optional parameter, to allow other behaviours.

pyk/src/pyk/ktool/kprint.py
nishantjr: Hmm, looking at the implementation of `run_process`, it seems to be doing the right thing. I'm surprised its getting an empty environment.

ehildenb: No the default env is `None`: https://github.com/runtimeverification/k/blob/a3b3f34bb6d08179ee5bc7d3489d8f98b0f46bd6/pyk/src/pyk/cli_utils.py#L43

pyk/src/pyk/__main__.py
ehildenb: Once again, not related to this issue at all. The question to ask yourself when reviewing code is: "is this an improvement over the current situation?" If so, then there is no need to block merging, you can make an issue with checklists or an immediate follow-up PR with things that aren't breakages. I will fix it here, but keep in mind that every time we bounce a PR back and forth, we are wasting at least an entire day, especially when we're in different timezones across the world. That only has to happen 3 times for a feature to be blocked from merging for _an entire week_, which is insane when I am working with deadlines for conferences which are on the order of weeks out. I need to deliver useful results in terms of web interface of prover by end of august when I have a demo, and would rather do it on merged code than telling people to check out specific branches.

pyk/src/pyk/integration_tests/test_kore_to_kast.py
ehildenb: Not worth blocking a merge over.

pyk/src/pyk/cli_utils.py
ehildenb: I've backed out this change completely, can deal with it another time.

pyk/src/pyk/__main__.py
ehildenb: Also, see Chesterton's Fence: https://fs.blog/chestertons-fence/

ehildenb: https://github.com/runtimeverification/k/issues/2773

pyk/src/pyk/integration_tests/test_kore_to_kast.py
ehildenb: Done.

pyk/src/pyk/__main__.py
ehildenb: I've backed out this change as well. Hopefully we can agree in the future that when changes are strictly improvements, we don't need to block merge because they could be _a little bit better_. "perfect is the enemy of the good".

tothtamas28 (approved):

pyk/src/pyk/__main__.py
nishantjr: Right, right. Sorry I didn't mean to imply we should change it immediately.

pyk/src/pyk/ktool/kprint.py
nishantjr: Right, and  that invokes the default behaviour for  `subprocess.popen`.

nishantjr (approved):

pyk/src/pyk/__main__.py
ehildenb: Then don't request changes! Hit Approve instead!

PR #2768, opened 2022-08-02
add flag to disable simplification rules for \ceil

dwightguth:
This is definitely not complete as we are basically forgetting axioms, but it ought to be sound (please confirm), and it decreases the time symbolically executing the c program `int main() { return 0; }` from like 8 minutes to 1 minute 20 seconds.


ehildenb:
This causes breakages in KEVM that seem non-trivial to fix. I'd rather not merge this, since it seems to be unusable for any non-trivial symbolic execution, but we can have it here for studying performance of the backend until we come with a more general fix.

radumereuta:
`--post-process` would be great for this case, wouldn't it?

ehildenb:
`--post-process` happens before `ModuleToKORE` though, right? So we need to also avoid generating those `#Ceil` rules that ModuleToKORE generates too. But yes! It could be used for things like this.

I think also, we can have `--post-process-kore` now, because pyk library supports Kore datastructures directly. So we could pretty easily extract this as a Kore->Kore pass in python that downstream users can incorporate in their semantics (so C semantics could just store the Python script needed to strip out these definitons).

I'll make an issue for `--post-process-kore`.

radumereuta:
But do we need a `-post-process-kore` option?
You just call a script after kompile finishes.

ehildenb:
True, it just integrates it better into `kompile` I guess, made an issue: https://github.com/runtimeverification/k/issues/2770

dwightguth:
Can we merge this? I'm aware it's not generally very useful, but this PR has no effect on anything if the flag isn't specified, so the fact that it causes problems for other semantics doesn't really matter; they can just not pass the flag. Meanwhile, it dramatically improves performance for the C examples I'm working on currently. I'm aware that as the examples get more complex, we may reach a point where this flag is more detrimental than helpful, but nevertheless, right now it would allow us to make much-needed progress on the NASA grant, so I would really rather we just go with it. As @ehildenb has said in the past, if code is strictly better than what came before, we shouldn't delay merging it just because it could be even better.

I have no objection to deleting this flag later once it is no longer needed.

ehildenb (approved):

radumereuta (approved):
Please add a `TODO:` comment saying we should remove this code when it's no longer needed.

PR #2771, opened 2022-08-03
Upstream `get_basic_block` from ksummarize repo

ehildenb:
This adds function `KProve.get_claim_basic_block`.

- Method `flattenLabel` is renamed to `_flatten_label`.
- Method `KProve._write_claim_definition` is updated with variable naming scheme, and changed to return both the Path that it wrote the definition to and the main module name.
- First a test is added in the IMP language of getting a basic block of a proof.
- Method `KProve.get_claim_basic_block` is implemented and exercised by the test.

This PR takes the implementation wholesale from ksummarize repo, with the intention of eventually simplifying it via JSON RPC.

### DIFF
--- a/pyk/src/pyk/integration_tests/test_proofs.py
+++ b/pyk/src/pyk/integration_tests/test_proofs.py
@@ -58,17 +58,29 @@ class ImpProofTest(KProveTest):
 
     def test_get_basic_block(self):
         # Given
-        new_claim = KClaim(KToken('<k> s = 0 ; while ( n > 0 ) { s = s + n ; n = n - 1 ; } => . ... </k> <state> n |-> (N => 0) s |-> (_ => (N *Int (N +Int 1)) /Int 2) </state>', 'KCell'))
+        new_claim = KClaim(KToken('<k> $s = 0 ; while ( 0 <= $n ) { $s = $s + $n ; $n = $n + -1 ; } => . ... </k> <state> $n |-> (N => 0) $s |-> (_ => (N *Int (N +Int 1)) /Int 2) </state>', 'KCell'))
 
         # When
-        post_state = self.kprove.get_basic_block(new_claim, 'imp-spec')
+        post_depth_actual, post_branching_actual, post_state = self.kprove.get_claim_basic_block('imp-basic-block', new_claim)
         post_state_pretty_actual = self.kprove.pretty_print(post_state)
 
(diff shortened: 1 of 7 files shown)


pyk/src/pyk/integration_tests/test_proofs.py
```diff
@@ -42,3 +42,45 @@ def test_prove_claim_rule_profile(self):
...
+                                     '      $n |-> N $s |-> 0\n' +                                                                                                   \
+                                     '    </state>\n' +                                                                                                              \
+                                     '  </T>\n' +                                                                                                                    \
+                                     '  <generatedCounter>\n' +                                                                                                      \
+                                     '    _DotVar3\n' +                                                                                                              \
+                                     '  </generatedCounter>\n' +                                                                                                     \
+                                     '</generatedTop>'
```
ehildenb: @nishantjr is this how you want it formatted?

nishantjr: Yeah, but we don't need the `+`s or the line continuations I think.

nishantjr: see here: https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/tests/test_kcfg.py#L338

nishantjr:
```suggestion
        post_state_pretty_expected = ('<generatedTop>\n'
                                      '  <T>\n'
                                      '    <k>\n'
                                      '      if ( 0 <=Int N ) { { $s = $s + $n ; $n = $n + -1 ; } while ( 0 <= $n ) { $s = $s + $n ; $n = $n + -1 ; } } else { }\n'
                                      '      ~> _DotVar2\n'
                                      '    </k>\n'
                                      '    <state>\n'
                                      '      $n |-> N $s |-> 0\n'
                                      '    </state>\n'
                                      '  </T>\n'
                                      '  <generatedCounter>\n'
                                      '    _DotVar3\n'
                                      '  </generatedCounter>\n'
                                      '</generatedTop>')
```

I'm not sure if the parens are needed.

nishantjr (approved):
Looks good, besides the formatting.

pyk/src/pyk/integration_tests/test_proofs.py
ehildenb: The parens are needed, at least I couldn't fine a way without. Looks much nicer.

PR #2778, opened 2022-08-06
pyk.kastManip: fix bug in simplify_bool

ehildenb:
There are several places where `Bool.notBool` is fed a list of arguments instead of a singleton. These are all fixed.

### DIFF
--- a/pyk/src/pyk/kastManip.py
+++ b/pyk/src/pyk/kastManip.py
@@ -117,14 +117,14 @@ def simplify_bool(k):
         return None
     simplify_rules = [ (KApply('_==K_', [KVariable('#LHS'), Bool.true]), KVariable('#LHS'))                                                                     # noqa
                      , (KApply('_==K_', [Bool.true, KVariable('#RHS')]), KVariable('#RHS'))                                                                     # noqa
-                     , (KApply('_==K_', [KVariable('#LHS'), Bool.false]), Bool.notBool([KVariable('#LHS')]))                                                    # noqa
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kastManip.py
```diff
@@ -117,14 +117,14 @@ def simplify_bool(k):
         return None
```
nishantjr, line 117: Please add a type signature for `simplify_bool`

pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -152,17 +152,15 @@ def test(self):
         self.assertEqual(config_actual, config_expected)
 
 
-class BooleanTest(TestCase):
+class SimplifyBoolTest(TestCase):
 
     def test_simplify_bool(self):
```
nishantjr, line 157:
```suggestion
    def test_simplify_bool(self) -> None:
```

Enable type checking here by adding a type signature.

nishantjr (approved):
Beyond adding type signatures, looks good.

pyk/src/pyk/kastManip.py
ehildenb: I will get this in follow-on PR.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: I will get this in follow-on PR.

PR #2784, opened 2022-08-08
Fix Ubuntu Jammy Release

ehildenb:
- Specify `python3` instead of `python3.8` in the `Makefile`, so it works on both focal and jammy (this doesn't affect the packaging, just development).
- Implement the `_tag` property of the `Kore` datatype, to avoid triggering unimplemented behavior at build time.


ehildenb:
Evidence this works: https://ci.runtimeverification.com/jenkins/blue/organizations/jenkins/K/detail/PR-2784/3/pipeline/110

ehildenb:
I upgraded my machine to Jammy and the build started breaking in the same way. It appears that something is different about Google's ABC class now that it actually tries to call the class properties `_tag` at build time, so we had to implement them.

nishantjr (approved):
Looks good!

PR #2785, opened 2022-08-08
pyk/kprove: avoid division by zero error

### LINKED ISSUE #2765: [Bug] [pyk prove] - Crashes on empty rule-log
K Version
---------

```
5.3.113
```

Description
-----------

`pyk` rule productivity measurements trigger a potential division by 0 when no rules are applied. Here (https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/ktool/kprove.py#L230) we do a division that could be zero on the denominator (for a proof that takes zero steps, for example). So we should handle this case more gracefully.

ehildenb:
Fixes: #2765


nishantjr (approved):
LGTM

PR #2786, opened 2022-08-08
pyk: updates to cli_utils, better process management

ehildenb:
This PR makes several improvements to `run_process` and associated functions:

- `stderr` is defaulted to being printed on `run_process`, but can be turned off with option `suppress_stderr`.
- Default to `check` being turned on for `run_process`, but allow it to be turned off (needed for kprove, for example).
- `build_rule/build_claim` return a `Subst` instead of a `Dict`.
- Add `Subst.ml_pred` property and removes `kastManip.substToMlPred`, adds test of `Subst.ml_pred`.
- Rename `python3.8` in `Makefile` to `python3`.
- Add standard named targets `venv`, `venv-dev`, and `venv-clean` to the `pyk/Makefile` for consistent development across our repos.

### DIFF
--- a/pyk/Makefile
+++ b/pyk/Makefile
@@ -70,7 +70,7 @@ test-unit: venv-dev
 test-integration: venv-dev
 	$(ACTIVATE_DEV) && python3 -m unittest discover --verbose --top src pyk.integration_tests $(PYK_TEST_ARGS)
 
-test-pyk: $(VENV_PROD)
+test-pyk: venv-prod
 	$(ACTIVATE_PROD) && $(MAKE) -C pyk-tests
 	rm -rf $(VENV_PROD_DIR) build src/pyk.egg-info
 
--- a/pyk/Makefile
+++ b/pyk/Makefile
@@ -5,7 +5,7 @@ export PATH := $(K_BIN):$(PATH)
 
 
 .PHONY: default all clean install                                  \
-        $(VENV_DEV_DIR) $(VENV_PROD_DIR)                           \
+        venv venv-dev venv-clean                                   \
         test test-unit test-integration test-pyk                   \
         check-code-style isort check-isort check-flake8 check-mypy
 
@@ -13,17 +13,17 @@ default: check-code-style test-unit
 
 all: check-code-style test
 
-clean:
-	rm -rf $(VENV_DEV_DIR) $(VENV_PROD_DIR) definitions build src/pyk.egg-info .myp_cache
+clean: venv-clean
+	rm -rf definitions build src/pyk.egg-info .myp_cache
 	find -type d -name __pycache__ -prune -exec rm -rf {} \;
 	$(MAKE) -C pyk-tests clean
 
 
 # Virtualenv
 
-VENV_DEV_DIR := venv-dev
+VENV_DEV_DIR := build
 VENV_DEV     := $(VENV_DEV_DIR)/pyvenv.cfg
-ACTIVATE_DEV := . $(VENV_DEV_DIR)/bin/activate
+ACTIVATE_DEV := . $(CURDIR)/$(VENV_DEV_DIR)/bin/activate
 
 $(VENV_DEV):
 	   virtualenv -p python3.8 $(VENV_DEV_DIR) \
@@ -36,7 +36,7 @@ $(VENV_DEV_DIR): $(VENV_DEV)
 
 VENV_PROD_DIR := venv-prod
 VENV_PROD     := $(VENV_PROD_DIR)/pyvenv.cfg
-ACTIVATE_PROD := . $(VENV_PROD_DIR)/bin/activate
+ACTIVATE_PROD := . $(CURDIR)/$(VENV_PROD_DIR)/bin/activate
 
 $(VENV_PROD):
 	   virtualenv -p python3.8 $(VENV_PROD_DIR) \
@@ -45,6 +45,14 @@ $(VENV_PROD):
 
 $(VENV_PROD_DIR): $(VENV_PROD)
 	@echo $(ACTIVATE_PROD)
+
+venv: $(VENV_PROD_DIR)
+
+venv-dev: $(VENV_DEV_DIR)
+
+venv-clean:
+	rm -rf $(VENV_PROD_DIR)
+	rm -rf $(VENV_DEV_DIR)
 
 install:
 	pip3 install . --root=$(DESTDIR) --prefix=$(PREFIX)
@@ -56,10 +64,10 @@ PYK_TEST_ARGS :=
 
 test: test-unit test-integration test-pyk
 
-test-unit: $(VENV_DEV)
+test-unit: venv-dev
 	$(ACTIVATE_DEV) && python3 -m unittest discover --verbose --top src pyk.tests $(PYK_TEST_ARGS)
 
-test-integration: $(VENV_DEV)
+test-integration: venv-dev
 	$(ACTIVATE_DEV) && python3 -m unittest discover --verbose --top src pyk.integration_tests $(PYK_TEST_ARGS)
 
 test-pyk: $(VENV_PROD)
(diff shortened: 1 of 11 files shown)


pyk/src/pyk/ktool/kprove.py
```diff
@@ -138,19 +140,13 @@ def prove(
...
         command_env['KORE_EXEC_OPTS'] = kore_exec_opts
 
-        proc_output: str
-        try:
-            proc_output = run_process(command, _LOGGER, env=command_env).stdout
-        except CalledProcessError as err:
-            if err.returncode != 1:
-                raise RuntimeError(f'Command kprove exited with code {err.returncode} for: {spec_file}', err.stdout, err.stderr)
-            proc_output = err.stdout
+        proc_result = run_process(command, logger=_LOGGER, env=command_env, check=False, profile=self._profile)
```
nishantjr, line 143: I don't like this. We're ignoring all errors now, and will make tracking down errors a lot harder.

pyk/Makefile
```diff
@@ -56,10 +64,10 @@ PYK_TEST_ARGS :=
...
 
-test-unit: $(VENV_DEV)
+test-unit: venv-dev
 	$(ACTIVATE_DEV) && python3 -m unittest discover --verbose --top src pyk.tests $(PYK_TEST_ARGS)
 
-test-integration: $(VENV_DEV)
+test-integration: venv-dev
 	$(ACTIVATE_DEV) && python3 -m unittest discover --verbose --top src pyk.integration_tests $(PYK_TEST_ARGS)
 
 test-pyk: $(VENV_PROD)
```
nishantjr:
```suggestion
test-pyk: venv
```

pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -164,3 +174,38 @@ def test_simplify_bool(self):
...
+            (init_1, target_1, claim_1),
+            (init_2, target_2, claim_2),
+            (init_3, target_3, claim_3),
+        )
+
+        for i, (init, target, claim) in enumerate(test_data):
+            init_cterm = CTerm(init)
+            target_cterm = CTerm(target)
+            kclaim, _ = build_claim(f'claim{i}', init_cterm, target_cterm)
+            self.assertEqual(kclaim, claim)
```
nishantjr: Lets used `subTest`, or parametrized tests here. Otherwise it would be difficult to identify which tests failed.

pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -164,3 +174,38 @@ def test_simplify_bool(self):
...
+        init_2 = k(v1)
+        init_3 = k(v1)
+
+        target_1 = k(v2)
+        target_2 = k(v2)
+        target_3 = mlAnd([k(v2), mlEqualsTrue(constraint_1)])
+
+        claim_1 = KClaim(k(KRewrite(unds_v1, v2)), requires=constraint_1, att=KAtt({'label': 'claim0'}))
+        claim_2 = KClaim(k(KRewrite(unds_v1, ques_unds_v2)), att=KAtt({'label': 'claim1'}))
+        claim_3 = KClaim(k(KRewrite(unds_v1, ques_v2)), ensures=constraint_2, att=KAtt({'label': 'claim2'}))
```
nishantjr: For readablity, I'd prefer grouping this the other way, or even just putting it directly in the table:

(Please check that I've copy-pasted correctly)
```suggestion
        constraint_1 = KApply('_<=Int_', [intToken(0), v2])
        constraint_2 = KApply('_<=Int_', [intToken(0), ques_v2])

       test_data = [ # Name, init, final, expected 
           ('foo',
            mlAnd([k(v1), mlEqualsTrue(constraint_1)]),
            k(v2),
            KClaim(k(KRewrite(unds_v1, v2)), requires=constraint_1, att=KAtt({'label': 'claim0'}))
           ),
           ('bar',
             k(v1),
             k(v2),
             KClaim(k(KRewrite(unds_v1, ques_unds_v2)), att=KAtt({'label': 'claim1'}))),
           ('foobar',
             k(v1),
             k(v2),
             KClaim(k(KRewrite(unds_v1, ques_v2)), ensures=constraint_2, att=KAtt({'label': 'claim2'})))
```

Again, I feel if you can't find a good name for a test, then what you are testing is not well-defined and it creates problems for those dealing with the code later. Its of course fine to have a couple of test called `complex-test-1` if you need to of course.

nishantjr (changes requested):
Requested some changes.

pyk/src/pyk/cli_utils.py
```diff
@@ -36,25 +39,38 @@ def file_path(s: str) -> Path:
...
+    if not logger:
+        logger = _LOGGER
+
+    logger.info(f'Running: {command}')
     try:
-        res = subprocess.run(args, input=input, env=env, capture_output=True, check=True, text=True)
+        stderr = subprocess.PIPE if suppress_stderr else None
+        if profile:
+            start_time = time.time()
+        res = subprocess.run(args, input=input, env=env, stdout=subprocess.PIPE, stderr=stderr, check=check, text=True)
```
nishantjr, line 63: Great! I like that we're inheriting stderr.

pyk/src/pyk/cli_utils.py
```diff
@@ -36,25 +39,38 @@ def file_path(s: str) -> Path:
 
 def run_process(
     args: Union[str, Sequence[str]],
-    logger: Logger,
     *,
-    log_level: int = logging.DEBUG,
+    check: bool = True,
+    profile: bool = False,
+    suppress_stderr: bool = False,
```
nishantjr, line 45: How do you feel about doing this instead:
```suggestion
    stderr = subprocess.PIPE,
```

This would give us the flexability to capture both stdout and stderr, or whatever else the user wants.

nishantjr (changes requested):

pyk/src/pyk/ktool/kprove.py
ehildenb: We're not ignoring all errors. We're always propogating stderr, and it will immediately fail on json decoding if it was any error code except `1` (which means "prover exited normally and produced output"). Any other error causes json decoding to fail (several lines down).

We have to not do json decoding here because we need to handle `dry_run` case.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: You still have the function name for the test.

ehildenb: And it displays in terminal.

pyk/src/pyk/ktool/kprove.py
nishantjr: Ah! OK, great.

pyk/src/pyk/tests/test_kastManip.py
nishantjr: This is three tests really.

nishantjr (approved):
Approved. Though I still think those subtests need names.

pyk/Makefile
ehildenb: Done.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: Done.

pyk/src/pyk/cli_utils.py
ehildenb: No, I want users to have to turn off stderr intentionally, _not_ to have it off by default. We should never be silently ignoring warnings from any software we're calling. If a piece of our software is being too annoying with warnings, we should fix either the source of the warnings or change our software to be quieter. If some other piece of software is giving annoying warnings, we can still turn it off.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: Done.

PR #2787, opened 2022-08-09
KIT upstream

nishantjr:
This is best reviewed commit by commit, rather than as one patch.

This PR:
1. Gets the summarizer working by copying over the code from the other repo, changing only paths and divergences since the last PR merge.
2. Move and refactors functions to their appropriate places commit by commit.


pyk/src/pyk/cfg_manager.py
```diff
@@ -142,7 +142,7 @@ def applyMacros(definition: KDefinition, term: KInner) -> KInner:
 
 def getAliases(definition: KDefinition) -> List[KRule]:
 
-    def _applyMacrosToAlias(sent):
+    def _applyMacrosToAlias(sent: KRule) -> KRule: # Typing fails
         lhs = sent.body.lhs
         rhs = sent.body.rhs
         new_rhs = applyMacros(definition, rhs)
```
nishantjr: Since typing fails here (`type(KRule.body) == KInner`, and has no `lhs`, `rhs`), it means that there no relevant aliases in both the IMP and KEVM definitions. All code the depends on aliases is redundant.

pyk/src/pyk/cfg_manager.py
```diff
@@ -123,145 +119,6 @@ def _cellVarsToLabels(_kast):
...
-    if missing_clause:
-        return None
-
-    else:
-        full_matches = reduce(_combineClauseMatches, matches)
-        if len(full_matches) == 0:
-            return None
-        return full_matches
-
-
```
nishantjr: All this code can now be removed without consequence. If we'd like to bring in AC rewriting at a later point, I think its important to have unit tests for it. I am not comfortable adding it in its current state.

pyk/src/pyk/ktool/kprint.py
```diff
@@ -126,11 +126,13 @@ def build_symbol_table(definition: KDefinition, opinionated=False) -> SymbolTabl
...
             assert(prod.klabel)
+
             label = prod.klabel.name
-            if 'symbol' in prod.att and 'klabel' in prod.att:
-                label = prod.att['klabel']
             unparser = unparser_for_production(prod)
+
             symbol_table[label] = unparser
+            if 'symbol' in prod.att and 'klabel' in prod.att:
+                symbol_table[prod.att['klabel']] = unparser
```
nishantjr: This was needed because of some weirdness in unparsing List units:

The production had `prod.klabel.name = List{"_,_"}_Id` and  klabel attribute as `List{"_,_"}`.
The AST nodes had `KApply.label = 'List{"_,_"}'`.

I guess we need to locate the java code for unparsing and duplicate it's logic here.

pyk/Makefile
```diff
@@ -75,6 +75,9 @@ check-code-style: check-isort check-flake8 check-mypy
 isort: $(VENV_DEV)
 	$(ACTIVATE_DEV) && isort src
 
+autoflake: $(VENV_DEV)
+	$(ACTIVATE_DEV) && autoflake -r --in-place --remove-all-unused-imports --ignore-init-module-imports src
```
nishantjr, line 89: We need `--ignore-init-module-imports`  so that it doesn't empty out `ktool.__init__`.

pyk/src/pyk/integration_tests/test_kprove.py
```diff
@@ -103,4 +103,4 @@ def _config(k: str, state: str) -> CTerm:
         init = _config('if (B:Bool) { $n = 2;  } else { $n = 3; }', '$n |-> 1')
         final = _config('.K', '$n |-> 1')
         result = self.kprove.prove_cterm('prove-cterm-multiple-results', init, final)
-        self.assertEqual(len(result), 2)
+        self.assertEqual(len(result), 1)  # Should be 2
```
nishantjr: I expected `_config('.K', '$n |-> 2') #Or _config('.K', '$n |-> 3')` but only got `n |-> 2` (even when running kprove from the command line directly).

pyk/src/pyk/kast.py
```diff
@@ -81,7 +81,16 @@ class KAtt(KAst, Mapping[str, Any]):
...
+
+        def _freeze(m: Any) -> Any:
+            # TODO: This should properly recurse through the data structure.
+            if isinstance(m, dict):
+                return FrozenDict(((k, _freeze(v)) for (k, v) in m.items()))
+            return m
+
+        frozen = _freeze(atts)
+        assert isinstance(frozen, FrozenDict)
+        object.__setattr__(self, 'atts', frozen)
```
nishantjr: I didn't want to pull in the full complexity for `sanitize_config` into `KProve.parse_token`. This meant that some Attrs are left in the Ast. In the case of variables, the `org.kframework.kore.Sort` attribute is a dictionary. This causes a crash as dicts are not hashable (and worse, are mutable). This code converts those to FrozenDicts.

I think it's probably a good idea to keep these attributes around too, since they encode the sort information for variables.

pyk/setup.cfg
```diff
@@ -18,7 +18,8 @@ where = src
 
 [options.entry_points]
 console_scripts =
-    pyk = pyk.__main__:main
+    pyk    = pyk.__main__:main
+    kit.py = pyk.ktool.kit:main
```
ehildenb:
```suggestion
    pyk    = pyk.__main__:main
    kit    = pyk.ktool.kit:main
```

pyk/kit-tests/test-kit.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+summary_script="$1" ; shift
+source ../kit-shell
+cd $(dirname ${summary_script})
+source $(basename ${summary_script})
```
ehildenb, line 8: Are we `cd ...` so that the kit-script can just have shorter relative ptahs in `kit-load ...`, or for some other reason?

pyk/src/pyk/cfg_manager.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # TODO: This is somewhat hacky. We shouldn't have to touch the config this much.
+    # Likely, the frontend should just be giving us initial states with these already in place.
+    def _remove_cell_map_definedness(_kast):
+        if type(_kast) is KApply:
+            if _kast.label.name.endswith('CellMap:in_keys'):
+                return Bool.false
+            elif _kast.label.name.endswith('CellMapItem'):
+                return _kast.args[1]
+        return _kast
```
ehildenb: Remove this block entirely, we can re-implement downstream and find a better solution.

pyk/src/pyk/cfg_manager.py
```diff
@@ -0,0 +1,247 @@
...
+
+    new_term = substitute(init_term, free_vars_subst)
+    new_term = remove_generated_cells(new_term)
+    new_term = bottom_up(_remove_cell_map_definedness, new_term)
```
ehildenb: Remove these two as well.

pyk/src/pyk/cfg_manager.py
```diff
@@ -0,0 +1,247 @@
...
+def sanitize_config(defn: KDefinition, init_term: KInner) -> KInner:
+
+    def _var_name(vname):
+        new_vname = vname
+        while new_vname.startswith('_') or new_vname.startswith('?'):
+            new_vname = new_vname[1:]
+        return new_vname
+
+    free_vars_subst = {vname: KVariable(_var_name(vname)) for vname in collectFreeVars(init_term)}
+
```
ehildenb: I think actually, we should try removing `sanitize_config` altogether, and see if we can get a simple working version for IMP with it removed. Can do that in a follow-up PR, but if you try removing it here, and the test output looks reasonable, I'd rather start from the smaller base of functionality than have to strip it out later once we begin to rely on it.

pyk/src/pyk/cfg_manager.py
```diff
@@ -0,0 +1,247 @@
...
+
+class CFGManager:
+    summary_name: Final[str]
+    strategy_name: Final[str]
+    spec_file: Final[Path]
+    main_file: Final[Optional[Path]]
+    summary_dir: Final[Path]
+    config_file: Final[Path]
+    kompiled_dir: Final[Path]
```
ehildenb: We're using `definition_dir` now instead of `kompiled_dir` for consistency with the new `--definition` flag to the frontend.

pyk/src/pyk/cfg_manager.py
```diff
@@ -0,0 +1,247 @@
...
+        constraints = [bool_to_ml_pred(ml_pred_to_bool(c, unsafe=True)) for c in flatten_label('#And', constraint)]
+        new_term = mlAnd([config] + constraints)
+
+    return new_term
+
+
+class CFGManager:
+    summary_name: Final[str]
+    strategy_name: Final[str]
+    spec_file: Final[Path]
```
ehildenb, line 65: This one is only relevant for proofs, but we want to support KRun as well eventually. Can it be removed?

pyk/src/pyk/ktool/kprint.py
ehildenb: Hmmm, or we need to identify the bug which is giving the wrong labels to productions and ast nodes.

pyk/src/pyk/ktool/kit.py
```diff
@@ -0,0 +1,387 @@
...
+        abstract_constraints = [config_match.apply(abstract_constraint) for abstract_constraint in abstract_constraints]
+        abstract_constraints = list(filter(lambda x: not CTerm._is_spurious_constraint(x), [config_match.apply(abstract_constraint) for abstract_constraint in abstract_constraints]))
+        impl = CTerm._ml_impl(concrete_constraints, abstract_constraints)
+        if impl != mlTop(Sorts.GENERATED_TOP_CELL):
+            fail_str = kprint.pretty_print(impl)
+            negative_cell_constraints = list(filter(_is_negative_cell_subst, concrete_constraints))
+            if len(negative_cell_constraints) > 0:
+                fail_str = f'{fail_str}\n\nNegated cell substitutions found (consider using _ => ?_):\n' + '\n'.join([kprint.pretty_print(cc) for cc in negative_cell_constraints])
+            return (False, f'Implication check failed, the following is the remaining implication:\n{fail_str}')
+    return (True, '')
```
ehildenb: Look through the tests you've added and see if this is actually used anywhere. If there are _any_ functionalities not used in the tests currently added, rip them out entirely.

Only add in the ufncitonality that is exercised by your IMP test. Much of this code is old and completely unused for a long itme, and so we'd rather upstream the core and a framework for extending, the bring in what's needed, then get stuck with that old code.

Every piece of code we take with us is a commitment ot "how we used to be doing things", and we should not constrain ourselves too much here.

ehildenb:
Good that you have it working with downstream summarizer repo, but let's drop that commitment now (will re-visit it when it's time to integrate downstream).

Focus on minimizing the added code, just the basic framework, and what's needed for `show-cfg`, `build-edge`, and `verify-edge`.

ehildenb:
Also, I was thinking diretory structure more like `src/pyk/src/...` and `src/kit/src/...` so that they each have their own package (and kit package depends on pyk package), but I'm not sure. This also seems like it should work OK, especially since you found the way to have `kit` be a binary produced by the build system here.

pyk/setup.cfg
nishantjr: That creates a conflict with the bash `kit` function. We can resolve this in a future PR.

pyk/kit-tests/test-kit.sh
nishantjr: Yeah, just so that we do not need to write out the path several times in the test script.

pyk/src/pyk/cfg_manager.py
nishantjr: Yes, but thats a whole new PR. We'd need to make the CFG.target optional I think.

pyk/src/pyk/cfg_manager.py
nishantjr: I've removed `sanitize_config`, but kept `remove_generated_cells`. Otherwise the generated claims do not parse.

pyk/src/pyk/ktool/kit.py
nishantjr: This functionality was only exercised for pretty printing errors in verification. I've removed `check_implication`, and added a test for the unhappy path.

pyk/src/pyk/cfg_manager.py
nishantjr: Removed

pyk/src/pyk/cfg_manager.py
nishantjr: see comment re `sanitize_config`.

nishantjr:
> Good that you have it working with downstream summarizer repo, but let's drop that commitment now (will re-visit it when it's time to integrate downstream).
> 
> Focus on minimizing the added code, just the basic framework, and what's needed for show-cfg, build-edge, and verify-edge.

This changes the entire mandate of the PR completely. Lets get this merged first and iterate from there.
I've "quarantined" code we aren't too sure about into `cfg_manager.py` and `kit.py`.
I feel that `case-split` is an essential part of the core too. Perhaps, I can "quarantine" a version of `parse_token` that allows uses K variables. Otherwise I fear that it won't be finished before my internship is done (August 15th).


> Also, I was thinking diretory structure more like src/pyk/src/... and src/kit/src/... so that they each have their own package (and kit package depends on pyk package), but I'm not sure. This also seems like it should work OK, especially since you found the way to have kit be a binary produced by the build system here.

Lets do that if it becomes necessary. I think `kit` will be basically a cli wrapper around KCFG, and everything else it uses will be generic pyk tools.

pyk/src/pyk/ktool/kprove.py
```diff
@@ -143,7 +143,7 @@ def prove(
             proc_output = run_process(command, _LOGGER, env=command_env).stdout
         except CalledProcessError as err:
             if err.returncode != 1:
-                raise RuntimeError(f'Command kprove exited with code {err.returncode} for: {spec_file}', err.stdout, err.stderr)
+                raise RuntimeError(f'Command kprove exited with code {err.returncode} for: {spec_file}', err.stdout, err.stderr) from err
```
tothtamas28: Note that exceptions will be implicitly chained here even if `from` is not provided. (I'm fine with making this explicit though.)

pyk/src/pyk/ktool/kompile.py
```diff
@@ -34,6 +34,7 @@ def kompile(
 ) -> Path:
     check_file_path(main_file)
 
+    include_dirs = list(include_dirs)
```
tothtamas28, line 39: Good catch.

pyk/src/pyk/kastManip.py
```diff
@@ -1,7 +1,8 @@
 import logging
 from collections import Counter
+from typing import Callable
+from typing import Counter as CounterType
```
tothtamas28: What do you think?

```python3
import typing

counter: typing.Counter[str] = Counter()
```

pyk/src/pyk/kcfg.py
```diff
@@ -535,6 +535,21 @@ def create_edge(self, source_id: str, target_id: str, condition: KInner, depth:
         self._edges[source.id][target.id] = edge
         return edge
 
+    def create_case_split(self, source_id: str, constraints: Iterable[KInner]) -> List[str]:
```
tothtamas28: How about `split_node` or `case_split_node`?

pyk/src/pyk/kast.py
```diff
@@ -81,7 +81,16 @@ class KAtt(KAst, Mapping[str, Any]):
     atts: FrozenDict[str, Any]
 
     def __init__(self, atts: Mapping[str, Any] = {}):
-        object.__setattr__(self, 'atts', FrozenDict(atts))
+
+        def _freeze(m: Any) -> Any:
```
tothtamas28, line 84: I do not think think enforcing this here is a good idea. As an analogue, consider `tuple` and `list` -- `FrozenDict` and `dict` are supposed to have the same relationship.

You could introduce a function `freeze` in `pyk.util` though that does this transformation on its argument.

nishantjr: I'm not sure I follow. You mean I should move the function to `pyk.util`, right? The actual implementation looks OK?

tothtamas28: Sorry, my mistake. I thought it was an invariant enforced on `FrozenDict` itself. By the way, can a `dict` be a value in an attribute map?

pyk/src/pyk/ktool/kprove.py
nishantjr: Oh, I've been seeing very different error messages since I've started chaining explicitly. The previous messages made me mildly panic everytime I saw them, thinking something went wrong when trying to handle the exception.

pyk/src/pyk/kastManip.py
nishantjr: I like it.

pyk/src/pyk/kast.py
nishantjr: Unfortunately it is, :-( for example to store the sort of variables, as seen in this test https://github.com/runtimeverification/k/blob/96f8ea1818875ef64a7ceaaf320c161e0befe3b6/pyk/src/pyk/integration_tests/test_parse_token.py#L23

pyk/setup.cfg
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/cfg_manager.py
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/kastManip.py
```diff
@@ -509,23 +514,21 @@ def setCell(constrainedTerm, cellVariable, cellValue):
     return mlAnd([substitute(config, subst), constraint])
 
 
-def removeUselessConstraints(constrainedTerm, keepVars=None):
-    (state, constraint) = split_config_and_constraints(constrainedTerm)
-    constraints = flatten_label('#And', constraint)
-    usedVars = collectFreeVars(state)
+def remove_useless_constraints(cterm: CTerm, keepVars=None) -> CTerm:
```
ehildenb: Also rename all the variables in the method with the new scheme while you're at it.

pyk/src/pyk/kcfg.py
```diff
@@ -535,6 +543,21 @@ def create_edge(self, source_id: str, target_id: str, condition: KInner, depth:
...
+            _cterm = CTerm(mlAnd([source.cterm.term, _constraint]))
+            _node = self.get_or_create_node(_cterm)
+            self.create_edge(source.id, _node.id, _constraint, 0)
+            self.add_verified(source.id, _node.id)
+            return _node.id
+
+        branch_node_ids = [_add_case_edge(constraint) for constraint in constraints]
+        self.add_expanded(source.id)
+        return branch_node_ids
+
```
ehildenb: Should we have another marker than `expanded` to show a node split as complete or not?

ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/ktool/kprint.py
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/cfg_manager.py
ehildenb: Makes sense.

ehildenb:
Looks pretty good, just the one variable renaming requests. Once it's passing tests it LGTM.

pyk/src/pyk/integration_tests/test_parse_token.py
```diff
@@ -0,0 +1,39 @@
...
+            return KAtt(FrozenDict({'org.kframework.kore.Sort': FrozenDict({'node': 'KSort', 'name': sort})}))
+
+        tests: Iterable[Tuple[str, KToken, KInner]] = (
+            ('variable', KToken('N', 'Int'), KVariable('N', vattr('K'))),  # TODO: This should parse as an int.   # noqa
+            (
+                '==Int',
+                KToken('N ==Int 1', 'Bool'),
+                KApply('_==Int_', KVariable('N', vattr('Int')), intToken(1)),
+            ),  # noqa
+        )
```
tothtamas28: `# noqa` can be removed here.

pyk/src/pyk/integration_tests/test_parse_token.py
```diff
@@ -0,0 +1,39 @@
...
+    @staticmethod
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_parse_token(self) -> None:
+        def vattr(sort: str) -> KAtt:
+            return KAtt(FrozenDict({'org.kframework.kore.Sort': FrozenDict({'node': 'KSort', 'name': sort})}))
+
+        tests: Iterable[Tuple[str, KToken, KInner]] = (
+            ('variable', KToken('N', 'Int'), KVariable('N', vattr('K'))),  # TODO: This should parse as an int.   # noqa
```
tothtamas28:
```suggestion
            ('variable', KToken('N', 'Int'), KVariable('N', vattr('K'))),  # TODO: This should parse as an int.
```

tothtamas28 (approved):

ehildenb:
Please back out the submodule update.

ehildenb (approved):

PR #2788, opened 2022-08-09
pyk/kprove: Remove rule profiling entirely

ehildenb:
The rule profiling code is a big ugly beast that only was used to find out that KEVM is being very unproductive when applying rules. We should remove it, so we don't have to commit to supporting it.


radumereuta (approved):
Is there a new alternative available?

PR #2789, opened 2022-08-09
[pyk] Upstream KProve.prove_cterm and KPrint.parse_token

ehildenb:
See https://github.com/runtimeverification/k/pull/2788. Somehow adding the types to `KProve.prove` triggers a typing error in that code, and it's not worth keeping that code anyway.

This brings in two new functionalities from downstream:

- `KProve.prove_cterm`, which invokes the prover given a pair of `CTerm` and source/target. Adds some tests.
- `KPrint.parse_token`, which invokes `_kast` to parse a given token. Currently only works on program grammar (not on rule grammar), which limits its applicability (something we need to ask the frontend for more support on).
- Adds some TODO currently failing tests for `parse_token` and `kast_to_kore`.
- Refactors `kprove.py` module a bit. Inlines `_kprove` into `kprove`, then renames the whole thing `_kprove` to be similar to the other ktool modules.
- Adds typing to `KProve.prove`, which triggers some typing error in the rule profiling code, which is not worth keeping anyway.

I tried to do this one as well, but failed: https://github.com/runtimeverification/pyk/issues/8. It can happen in a follow-up PR.

### DIFF
--- a/pyk/src/pyk/integration_tests/test_emit_json_spec.py
+++ b/pyk/src/pyk/integration_tests/test_emit_json_spec.py
@@ -3,7 +3,8 @@ from pathlib import Path
 
 from ..kast import EMPTY_ATT, KAst, KDefinition, KFlatModuleList, KRequire
 from ..kastManip import remove_generated_cells
-from ..ktool import KompileBackend, _kprove
+from ..ktool import KompileBackend
+from ..ktool.kprove import _kprove
 from .kprove_test import KProveTest
 
 
--- a/pyk/src/pyk/integration_tests/test_emit_json_spec.py
+++ b/pyk/src/pyk/integration_tests/test_emit_json_spec.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 from ..kast import EMPTY_ATT, KAst, KDefinition, KFlatModuleList, KRequire
 from ..kastManip import remove_generated_cells
-from ..ktool import KompileBackend, kprove
+from ..ktool import KompileBackend, _kprove
 from .kprove_test import KProveTest
 
 
@@ -25,7 +25,7 @@ class EmitJsonSpecTest(KProveTest):
 
         spec_file = Path(self.SPEC_FILE)
         emit_json_spec = Path(self.SPEC_JSON_FILE)
-        kprove(spec_file, kompiled_dir=self.KOMPILE_OUTPUT_DIR, emit_json_spec=emit_json_spec, dry_run=True)
+        _kprove(spec_file, kompiled_dir=self.KOMPILE_OUTPUT_DIR, emit_json_spec=emit_json_spec, dry_run=True)
 
         with open(self.SPEC_JSON_FILE, 'r') as spec_file:
             kfml: KFlatModuleList = KAst.from_dict(json.load(spec_file)['term'])
--- a/pyk/src/pyk/integration_tests/test_kprint.py
+++ b/pyk/src/pyk/integration_tests/test_kprint.py
@@ -1,4 +1,4 @@
-from ..kast import KApply, KToken
+from ..kast import KApply, KToken, KVariable
 from ..ktool import KompileBackend
 from ..prelude import intToken
 from .kprove_test import KProveTest
@@ -21,6 +21,7 @@ class ImpParseTest(KProveTest):
             ('int-token', KToken('3', 'Int'), intToken(3)),
             ('id-token', KToken('abc', 'Id'), KToken('abc', 'Id')),
             ('add-aexp', KToken('3 + 4', 'AExp'), KApply('_+_', [intToken(3), intToken(4)])),
+            # TODO: ('add-int', KToken('3 +Int V', 'Int'), KApply('_+Int_', [intToken(3), KVariable('V')])),
             # TODO: ('k-cell', KToken('<k> . </k>', 'KCell'), KApply('<k>', [KApply(Labels.EMPTY_K)])),
         )
 
--- a/pyk/src/pyk/integration_tests/test_kprint.py
+++ b/pyk/src/pyk/integration_tests/test_kprint.py
@@ -1,0 +1,29 @@
+from ..kast import KApply, KToken
+from ..ktool import KompileBackend
+from ..prelude import intToken
+from .kprove_test import KProveTest
+
+
+class ImpParseTest(KProveTest):
+    KOMPILE_MAIN_FILE = 'k-files/imp.k'
+    KOMPILE_BACKEND = KompileBackend.HASKELL
+    KOMPILE_OUTPUT_DIR = 'definitions/imp'
+    KOMPILE_EMIT_JSON = True
+
+    KPROVE_USE_DIR = '.imp'
+
+    @staticmethod
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_parse_token(self):
+        test_parses = (
+            ('int-token', KToken('3', 'Int'), intToken(3)),
+            ('id-token', KToken('abc', 'Id'), KToken('abc', 'Id')),
+            ('add-aexp', KToken('3 + 4', 'AExp'), KApply('_+_', [intToken(3), intToken(4)])),
+        )
+
+        for name, token, kast in test_parses:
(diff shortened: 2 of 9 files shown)


ehildenb:
The last commit merges this branch in: https://github.com/runtimeverification/k/pull/2788

Merging this will include those changes.

pyk/src/pyk/integration_tests/test_kprint.py
```diff
@@ -0,0 +1,30 @@
...
+    def _update_symbol_table(symbol_table):
+        pass
+
+    def test_parse_token(self):
+        test_parses = (
+            ('int-token', KToken('3', 'Int'), intToken(3)),
+            ('id-token', KToken('abc', 'Id'), KToken('abc', 'Id')),
+            ('add-aexp', KToken('3 + 4', 'AExp'), KApply('_+_', [intToken(3), intToken(4)])),
+            # TODO: ('k-cell', KToken('<k> . </k>', 'KCell'), KApply('<k>', [KApply(Labels.EMPTY_K)])),
+        )
```
nishantjr, line 26: While I love that we can use `kast` here this completely blocks the KIT upstream PR. The `case-split` command requires that we are able to parse k variables.

pyk/src/pyk/integration_tests/test_emit_json_spec.py
```diff
@@ -3,7 +3,7 @@
 
 from ..kast import EMPTY_ATT, KAst, KDefinition, KFlatModuleList, KRequire
 from ..kastManip import remove_generated_cells
-from ..ktool import KompileBackend, kprove
+from ..ktool import KompileBackend, _kprove
```
nishantjr:
```suggestion
from ..ktool.kompile import KompileBackend, _kprove
```

pyk/src/pyk/ktool/__init__.py
```diff
@@ -9,5 +9,5 @@
     pretty_print_kast,
     unparser_for_production,
 )
-from .kprove import KProve, kprove
+from .kprove import KProve, _kprove
```
nishantjr: We really shouldn't be encouraging the use of private members.

```suggestion
from .kprove import KProve
```

nishantjr (dismissed):

pyk/src/pyk/ktool/kprove.py
```diff
@@ -110,23 +107,22 @@ def __init__(self, definition_dir: Path, main_file: Optional[Path] = None, use_d
...
         if log_file.exists():
             log_file.unlink()
-        haskell_log_entries += ['DebugTransition']
-        haskell_log_entries += ['DebugAttemptedRewriteRules'] if rule_profile else []
-        haskell_log_entries = unique(haskell_log_entries)
-        haskell_log_args = ['--log', str(log_file), '--log-format', 'oneline', '--log-entries', ','.join(haskell_log_entries)]
+        _haskell_log_entries = list(haskell_log_entries)
+        _haskell_log_entries += ['DebugTransition']
+        _haskell_log_entries = list(unique(_haskell_log_entries))
+        _haskell_log_args = ['--log', str(log_file), '--log-format', 'oneline', '--log-entries', ','.join(_haskell_log_entries)]
```
nishantjr: Whats the reason for adding the underscore here?

pyk/src/pyk/integration_tests/test_kprint.py
nishantjr: Let me know if https://github.com/runtimeverification/k/pull/2787#issuecomment-1210258327 and https://github.com/runtimeverification/k/pull/2787/commits/7c68dfd46d4490d78ab20aa161cec3a411b6ec51 is a reasonable solution to this situation.

pyk/src/pyk/ktool/kprove.py
ehildenb: typing

pyk/src/pyk/ktool/__init__.py
ehildenb: We must.

pyk/src/pyk/integration_tests/test_emit_json_spec.py
ehildenb: I'm not sure this works, I tried it, but will try again.

pyk/src/pyk/integration_tests/test_kprint.py
ehildenb: I mean, yeah, that's fine, but I wanted to get a cleaner version merged here, and then open the relevant issues on the frontend, and then get a hacky version merged.

I'll incorporate a hacky rule-parser version here as well.

pyk/src/pyk/ktool/kprove.py
ehildenb: The type of `haskell_log_entries` is an immutable `Iterable` as parameters, can't add to it I don't think. So just make a new mutable list `_haskell_log_entries` and use that.

pyk/src/pyk/ktool/__init__.py
ehildenb: The `_` is to discourage that use, but there will likely always be cases where we want this escape hatch downstream.

pyk/src/pyk/ktool/kprove.py
nishantjr: The variable can be changed, we just don't know about the thing it points to. It may be mutable or not. Changing the first line to:

```
haskell_log_entries = list(haskell_log_entries) 
```

should make it mutable (and the mypy should be able to infer that it is mutable).

pyk/src/pyk/integration_tests/test_emit_json_spec.py
nishantjr: Oh, why wouldn't it work?

pyk/src/pyk/ktool/__init__.py
nishantjr: True, but placing it in `__init__` makes it look as if is part of the public interface.

pyk/src/pyk/ktool/kprove.py
nishantjr: E.g. This types fine https://github.com/runtimeverification/k/blob/kit-upstream-2/pyk/src/pyk/ktool/kompile.py#L39

radumereuta (approved):

pyk/src/pyk/ktool/kprove.py
ehildenb: Ah you're right. I adjusted it.

pyk/src/pyk/ktool/__init__.py
ehildenb: I'd rather make it a separate issue/PR. We're the only users at the moment, we can make changes as we go.

pyk/src/pyk/integration_tests/test_emit_json_spec.py
ehildenb: I don't know, probably cause I'm bad at computers. I've updated it with these `_` imports hidden and made more explicit where needed.

pyk/src/pyk/ktool/__init__.py
ehildenb: I've updated the PR with this change, hiding those imports by default.

ehildenb:
@nishantjr I'd rather not make `parse_token` the hacky version. I know it doesn't quite do what we need yet for `case-split`, but I think we can just not include case-split on the first merge of KIT.

I've made an issue to extend `kast` to allow rule-grammar parsing here: https://github.com/runtimeverification/k/issues/2796

pyk/src/pyk/integration_tests/test_kprint.py
ehildenb: I didn't have time to make the hacky version. I would prefer to just get this merged for now, with the tests marking what we need.

tothtamas28 (approved):

pyk/src/pyk/integration_tests/test_kprint.py
ehildenb: Marking this as resolved to get this merged.

PR #2792, opened 2022-08-10
Add the `black` code formatter

tothtamas28:
Related: https://github.com/runtimeverification/devops/issues/96

### DIFF
--- a/pyk/src/pyk/integration_tests/test_defn.py
+++ b/pyk/src/pyk/integration_tests/test_defn.py
@@ -1,13 +1,4 @@
-from ..kast import (
-    KApply,
-    KClaim,
-    KRewrite,
-    KSort,
-    KToken,
-    KVariable,
-    assocWithUnit,
-    constLabel,
-)
+from ..kast import KApply, KClaim, KRewrite, KSort, KToken, KVariable, assocWithUnit, constLabel
 from ..kastManip import push_down_rewrites
 from ..ktool import KompileBackend
 from ..prelude import Sorts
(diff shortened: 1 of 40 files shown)


pyk/src/pyk/integration_tests/test_defn.py
```diff
@@ -29,16 +20,37 @@ def _update_symbol_table(symbol_table):
...
+                                    [KToken('x', 'Id'), KApply('_,_', [KToken('y', 'Id'), KApply('.List{"_,_"}')])],
+                                )
+                            ],
+                        )
+                    ],
+                ),
+                KApply('<state>', [KApply('.Map')]),
+            ],
+        )
+        # fmt: off
```
ehildenb: Oh no. We can't have this as the format. Can we turn off the line-breaking features of black? Formatted code should be beautiful and compact, not taking a book to read.

tothtamas28: I agree with you w.r.t. the highlighted instance. In such cases formatting can be turned off locally:

```python3
# fmt: off
...
# fmt: on
```

In general I wouldn't disable formatting related to line breaking though.

Baltoli:
On the C semantics, I found `black` produced pretty ugly formatting for our code, so I switched over to using [autopep8](https://github.com/runtimeverification/c-semantics/blob/dc8d99a12ccc05adf403b98dd1ee98eed319039f/.github/workflows/clang-format-check.yml#L26-L37), which is a bit more configurable. It would be nice to have an org-wide python style, though, so I might revisit using `black` with local disabling pragmas

tothtamas28:
> On the C semantics, I found `black` produced pretty ugly formatting for our code, so I switched over to using autopep8, which is a bit more configurable

Did you experience formatting issues for code other than large nested data structures? If not, I'd say `fmt: off` or `fmt: skip` is an acceptable workaround (for me at least).

Otherwise, we can also consider [YAPF](https://github.com/google/yapf), which seems to be quite configurable.

Baltoli:
I think it was primarily the nested data structures so I'll take another look and see if `# fmt: off` would solve it.

ehildenb (approved):

ehildenb:
I somewhat agree that this may be uglier than the original code. I approved so you aren't blocked, but if you find another formatter does something prettier (more compact), then lets use that instead.

In particular, I really don't like enforced line wrapping. I have a wide screen, I want to use it. Figure people can turn line wrapping on in their editors.

tothtamas28:
> if you find another formatter does something prettier (more compact), then lets use that instead

I played with `yapf` a bit, but couldn't come up with a configuration that results in a nicer format than what `black` provides out of the box. The closest I got to what I find acceptable is this:

```toml
[tool.yapf]
based_on_style = "pep8"
column_limit = 120
coalesce_brackets = true
dedent_closing_brackets = true  # false works as well
space_between_ending_comma_and_closing_bracket = false
split_arguments_when_comma_terminated = true
each_dict_entry_on_separate_line = false
```

Here, the only option I miss from `black` is  `coalesce_brackets` ([docs](https://github.com/google/yapf#knobs)), which turns this:

```python3
config_expected = '\n'.join(
    [
        '<T>',
        '  <k>',
        '    int x , y ;',
        '  </k>',
        '  <state>',
        '    .Map',
        '  </state>',
        '</T>',
    ]
)
```

into this:

```python3
config_expected = '\n'.join([
    '<T>',
    '  <k>',
    '    int x , y ;',
    '  </k>',
    '  <state>',
    '    .Map',
    '  </state>',
    '</T>',
])
```

On the other hand,  I find `black` is more consistent when it comes to alignment and trailing commas, and breaks up complex expressions in a more comprehensible way.

PR #2795, opened 2022-08-10
always make progress on each z3 query

dwightguth:
This PR makes a slight modification to the way we search for type inference solutions. Intead of computing a solution then looking for more solutions that are not less than that solution, then discarding non-maximal solutions, we instead compute a solution, look for solutions greater than it until we find one which is maximal, then look for solutions not less than the maximal solution. This slightly cuts down the search space we need to explore during type inference, which yields moderate improvements in performance when I measured, on the order of 5-30%.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/inner/disambiguation/TypeInferenceVisitor.java
+++ b/kernel/src/main/java/org/kframework/parser/inner/disambiguation/TypeInferenceVisitor.java
@@ -119,6 +119,25 @@ public class TypeInferenceVisitor extends SetsTransformerWithErrors<KEMException
           inferencer.computeModel();
         }
         once = false;
+        boolean isMaximal = false;
+        do {
+          inferencer.pushGreater();
+          switch(inferencer.status()) {
+            case SATISFIABLE:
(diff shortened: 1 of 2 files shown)


radumereuta (approved):
A few details about where the C semantics is faster:
```
24389 /home/radu/work/c-semantics/semantics/implementation/implementation.k:150   parse: 302 typeInf:23898
  888 /home/radu/work/c-semantics/semantics/implementation/implementation.k:150   parse: 246 typeInf: 463

17166 /home/radu/work/c-semantics/semantics/language/stmt/compound.k:30   parse: 236 typeInf:16734
  697 /home/radu/work/c-semantics/semantics/language/stmt/compound.k:30   parse: 205 typeInf: 310

8955 /home/radu/work/c-semantics/semantics/checker/language/concepts/conversions.k:71   parse:1011 typeInf:6139
4298 /home/radu/work/c-semantics/semantics/checker/language/concepts/conversions.k:71   parse: 453 typeInf:3122
```
I can't say much about correctness, but tests seem to pass.

```
$ z3 --version
Z3 version 4.8.15 - 64 bit
```

PR #2800, opened 2022-08-11
[pyk] Code cleanups, faster initialization

ehildenb:
This PR introduces some accumulated cleanup changes, fixes a typo bug, and reduces KEVM initialization time:

- This one https://github.com/runtimeverification/k/pull/2786 accidentally renamed `bool_to_ml_pred` to be `bool_ml_pred`. That is fixed, and a test is added for `bool_to_ml_pred`.
- Types are added to `simplify_bool` method.
- `CTerm.term` is renamed `CTerm.kast`. Not 100% on this one, but eventually I want `CTerm.kore` as well, so I was thinking it made sense.
- `KPrint` only actually loads the definition from disk if it's used by the program, by changing `KPrint.definition`, `KPrint.definition_hash`, and `KPrint.symbol_table` into `@property` instead of direct fields. On KEVM, the overhead of running each conformance test with pyk is ~4s originally. With this change, it drops to ~2s (because to run the KEVM conformance tests the kevm pyk runner never actually needs to use the definition or unparser). The KEVM tests are normally running with the bash script in around ~0.05s.
- Renames `readKastTerm` to `read_kast`, and adds a new one `read_kast_definition` which is actually what is used throughout the rest of the codebase. Adds types to both `read_kast` and `read_kast_definition`.

### DIFF
--- a/pyk/src/pyk/integration_tests/kprove_test.py
+++ b/pyk/src/pyk/integration_tests/kprove_test.py
@@ -28,7 +28,7 @@ class KProveTest(KompiledTest, ABC):
 
         self.kprove = KProve(self.kompiled_dir, kprove_main_file, Path(self.KPROVE_USE_DIR))
         self.kprove.prover_args += list(chain.from_iterable(['-I', include_dir] for include_dir in kprove_include_dirs))
-        self._update_symbol_table(self.kprove.symbol_table)
+        self._update_symbol_table(self.kprove._symbol_table)
 
     def tearDown(self):
         shutil.rmtree(self.KPROVE_USE_DIR, ignore_errors=True)
--- a/pyk/src/pyk/integration_tests/kprove_test.py
+++ b/pyk/src/pyk/integration_tests/kprove_test.py
@@ -28,9 +28,7 @@ class KProveTest(KompiledTest, ABC):
 
         self.kprove = KProve(self.kompiled_dir, kprove_main_file, Path(self.KPROVE_USE_DIR))
         self.kprove.prover_args += list(chain.from_iterable(['-I', include_dir] for include_dir in kprove_include_dirs))
-        # force computation of the symbol_table before updating it
-        if self.kprove.symbol_table:
-            self._update_symbol_table(self.kprove._symbol_table)
+        self._update_symbol_table(self.kprove.symbol_table)
 
     def tearDown(self):
         shutil.rmtree(self.KPROVE_USE_DIR, ignore_errors=True)
--- a/pyk/src/pyk/integration_tests/kprove_test.py
+++ b/pyk/src/pyk/integration_tests/kprove_test.py
@@ -28,7 +28,9 @@ class KProveTest(KompiledTest, ABC):
 
         self.kprove = KProve(self.kompiled_dir, kprove_main_file, Path(self.KPROVE_USE_DIR))
         self.kprove.prover_args += list(chain.from_iterable(['-I', include_dir] for include_dir in kprove_include_dirs))
-        self._update_symbol_table(self.kprove._symbol_table)
+        # force computation of the symbol_table before updating it
+        if self.kprove.symbol_table:
+            self._update_symbol_table(self.kprove._symbol_table)
 
     def tearDown(self):
         shutil.rmtree(self.KPROVE_USE_DIR, ignore_errors=True)
--- a/pyk/src/pyk/ktool/kprint.py
+++ b/pyk/src/pyk/ktool/kprint.py
@@ -74,10 +74,11 @@ class KPrint:
 
     definition_dir: Path
     use_directory: Path
-    _definition: Optional[KDefinition]
-    symbol_table: SymbolTable
-    definition_hash: str
     _profile: bool
+
+    _definition: Optional[KDefinition]
+    _definition_hash: Optional[str]
+    _symbol_table: Optional[SymbolTable]
 
     def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, profile: bool = False) -> None:
         self.definition_dir = Path(definition_dir)
@@ -87,8 +88,9 @@ class KPrint:
             td = TemporaryDirectory()
             self.use_directory = Path(td.name)
         check_dir_path(self.use_directory)
-        self.symbol_table = build_symbol_table(self.definition, opinionated=True)
-        self.definition_hash = hash_str(self.definition)
+        self._definition = None
+        self._definition_hash = None
+        self._symbol_table = None
         self._profile = profile
 
     @property
@@ -97,6 +99,20 @@ class KPrint:
             self._definition = readKastTerm(self.definition_dir / 'compiled.json')
             assert self._definition is not None
         return self._definition
+
+    @property
+    def definition_hash(self) -> str:
+        if not self._definition_hash:
+            self._definition_hash = hash_str(self.definition)
+            assert self._definition_hash is not None
+        return self._definition_hash
+
+    @property
+    def symbol_table(self) -> SymbolTable:
+        if not self._symbol_table:
+            self._symbol_table = build_symbol_table(self.definition, opinionated=True)
+            assert self._symbol_table is not None
+        return self._symbol_table
 
     def parse_token(self, ktoken: KToken) -> KInner:
         output = _kast(self.definition_dir, ktoken.token, sort=ktoken.sort, profile=self._profile)
--- a/pyk/src/pyk/ktool/kprint.py
+++ b/pyk/src/pyk/ktool/kprint.py
@@ -42,7 +42,6 @@ from ..kast import (
 from ..kore.parser import KoreParser
 from ..kore.syntax import Kore
 from ..prelude import Bool, Labels, Sorts
-from ..utils import hash_str
 
 _LOGGER: Final = logging.getLogger(__name__)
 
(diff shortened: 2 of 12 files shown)


pyk/src/pyk/integration_tests/kprove_test.py
```diff
@@ -28,7 +28,9 @@ def setUp(self):
 
         self.kprove = KProve(self.kompiled_dir, kprove_main_file, Path(self.KPROVE_USE_DIR))
         self.kprove.prover_args += list(chain.from_iterable(['-I', include_dir] for include_dir in kprove_include_dirs))
-        self._update_symbol_table(self.kprove.symbol_table)
+        # force computation of the symbol_table before updating it
+        if self.kprove.symbol_table:
+            self._update_symbol_table(self.kprove._symbol_table)
```
tothtamas28: To me it looks like the original version achieves the same thing. Did you run into issues with that?

pyk/src/pyk/ktool/kprint.py
```diff
@@ -87,11 +88,32 @@ def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, p
...
+        if not self._definition_hash:
+            self._definition_hash = hash_str(self.definition)
+            assert self._definition_hash is not None
+        return self._definition_hash
+
+    @property
+    def symbol_table(self) -> SymbolTable:
+        if not self._symbol_table:
+            self._symbol_table = build_symbol_table(self.definition, opinionated=True)
+            assert self._symbol_table is not None
```
nishantjr:
```suggestion
```

The type system checks this already.

pyk/src/pyk/ktool/kprint.py
```diff
@@ -87,11 +88,32 @@ def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, p
...
+        if not self._definition:
+            self._definition = readKastTerm(self.definition_dir / 'compiled.json')
+            assert self._definition is not None
+        return self._definition
+
+    @property
+    def definition_hash(self) -> str:
+        if not self._definition_hash:
+            self._definition_hash = hash_str(self.definition)
+            assert self._definition_hash is not None
```
nishantjr:
```suggestion
```

Checked by type system.

pyk/src/pyk/ktool/kprint.py
```diff
@@ -87,11 +88,32 @@ def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, p
...
+        return self._definition_hash
```
tothtamas28: Did you consider  `self.definition.hash` instead?

tothtamas28 (approved):

pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -103,33 +104,39 @@ def test_minimize_term(self):
                 self.assertEqual(actual, expected)
 
 
-class MlPredToBoolTest(TestCase):
+class BoolMlPredConversionsTest(TestCase):
+
+    # TODO: We'd like for bool_to_ml_pred and ml_pred_to_bool to be somewhat invertible.
+    test_data_1 = (
```
nishantjr:
```suggestion
    test_data_ml_pred_to_bool : Iterator[Tuple[str, bool, ...]] = (
```

pyk/src/pyk/tests/test_kastManip.py
```diff
@@ -103,33 +104,39 @@ def test_minimize_term(self):
...
+        ('not-top', False, KApply(KLabel('#Not', [Sorts.GENERATED_TOP_CELL]), [mlTop()]), Bool.notBool(Bool.true)),
+        ('equals-term', True, KApply(KLabel('#Equals'), [f(a), f(x)]), KApply('_==K_', [f(a), f(x)])),
+        ('simplify-and-true', False, KApply(KLabel('#And', [Sorts.GENERATED_TOP_CELL]), [mlEqualsTrue(Bool.true), mlEqualsTrue(Bool.true)]), Bool.true),
+        ('ceil-set-concat-no-sort', True, KApply(KLabel('#Ceil', [KSort('Set'), Sorts.GENERATED_TOP_CELL]), [KApply(KLabel('_Set_'), [KVariable('_'), KVariable('_')])]), KVariable('Ceil_0f9c9997')),
+        ('ceil-set-concat-sort', True, KApply(KLabel('#Not', [Sorts.GENERATED_TOP_CELL]), [KApply(KLabel('#Ceil', [KSort('Set'), Sorts.GENERATED_TOP_CELL]), [KApply(KLabel('_Set_'), [KVariable('_'), KVariable('_')])])]), Bool.notBool(KVariable('Ceil_0f9c9997'))),
+        ('exists-equal-int', True, KApply(KLabel('#Exists', [Sorts.INT, Sorts.BOOL]), [KVariable('X'), KApply('_==Int_', [KVariable('X'), KVariable('Y')])]), KVariable('Exists_6acf2557')),
+    )
+    test_data_2 = (
```
nishantjr: Rename, add types, and move to just before the test that uses it.

nishantjr (approved):
Looks good! Changes requested.

pyk/src/pyk/integration_tests/kprove_test.py
ehildenb: Yeah, calling `self.kprove.symbol_table` forces it to actually load the definition, so that `_symbol_table` is not `None` anymore. Then we can patch it.

pyk/src/pyk/ktool/kprint.py
ehildenb: Good idea, I'll see if it works.

pyk/src/pyk/ktool/kprint.py
ehildenb: I'll try it.

pyk/src/pyk/integration_tests/kprove_test.py
tothtamas28: But doesn't `self._update_symbol_table(self.kprove.symbol_table)` also evaluate the property before passing the value to the function? (Sorry, I might be missing some detail.)

ehildenb: Hmmm, that's a good point, but for some reason it wasn't working. I'll play with it a bit more and see if I can get it to work.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: Seems unnecssary, but I'll add.

ehildenb: I actually can't get this typing to work.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: I organized them, but still cannot get the typing to work.

pyk/src/pyk/ktool/kprint.py
ehildenb: Done.

pyk/src/pyk/ktool/kprint.py
ehildenb: Done.

pyk/src/pyk/ktool/kprint.py
ehildenb: Done.

pyk/src/pyk/ktool/kprint.py
ehildenb: Done.

pyk/src/pyk/integration_tests/kprove_test.py
ehildenb: Ah no, the original code doesn't work actually. Because it passes `self.kprove.symbol_table`, which is now a `@property` instead of a member field, and then `_patch_symbol_table` tries to update that property which is not allowed.

Instead here, we call the property to make sure it's oladed the first time, then we're patching on the `_symbol_table` underlying field.

pyk/src/pyk/tests/test_kastManip.py
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/tests/test_kastManip.py
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

pyk/src/pyk/integration_tests/kprove_test.py
tothtamas28: That's strange. I tried it now to see what's going on, but it didn't break integration tests. Maybe there's some discrepancy between Python 3.8 and 3.10.

ehildenb: I tried the original version again, and it worked! I am not sure what happened.

Though, it does seem strange to me that it did work honestly. Why would be be able to write to a `@property` directly?

tothtamas28: Good question. This is a read-only property, which indeed means it only has a getter, but no setter. But in this case, the setter is not invoked - what is called is `kprove.symbol_table['key'] = value`, which conceptually translates to `kprove.get_symbol_table().__setitem__('key', value)`, so it's fine.

ehildenb: Nice, makes sense. Thank you.

PR #2802, opened 2022-08-12
Use new LLDB script to debug on mac

Baltoli:
This is the corresponding frontend PR for https://github.com/runtimeverification/llvm-backend/pull/546. It launches the new `klldb.py` script when `--debugger` is passed to `krun` on macOS.

It also makes some small updates to the tutorial and user manual sections that deal with GDB, to explain that it won't currently work on macOS / LLDB yet.

Supersedes https://github.com/runtimeverification/k/pull/2646


dwightguth (approved):
Looks fine, but please double check that the install process for k installs python 3.

Baltoli:
macOS ships python3 by default for every version we support.

PR #2805, opened 2022-08-12
Update `execute` response types in `KoreClient`

### LINKED ISSUE #2804: Follow modifications to HS-Backend JSON-RPC protocol
https://github.com/runtimeverification/haskell-backend/pull/3192 modifies the HS backend's JSON-RPC endpoint for `execute`. Specifically:
* The reason value `final-state` was removed;
* The `states` field was renamed to `next-states`;
* A new field `state` was added;
* The `condition` field within the list elements of `next-states` was flattened into the parent. States in the `execute` response are now always triplets of `term` (previously another `state`), `predicate` and `substitution`, each containing a kore json term;
* The `depth` fields in the state list were replaced by a single `depth` field at the top level;
* A new `rule` field was added, which is filled if `reason` is `cut-point-rule` or `terminal-rule`.
More details on these changes can be found [in the related issue](https://github.com/runtimeverification/haskell-backend/issues/3143#issuecomment-1210068504) and [in the JSON RPC server API documentation](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md) as soon as the pull request is merged

tothtamas28:
Fixes #2804

### DIFF
--- a/pyk/Makefile
+++ b/pyk/Makefile
@@ -37,10 +37,10 @@ PYK_TEST_ARGS :=
 test: test-unit test-integration test-pyk test-kit
 
 test-unit: poetry-install
-	$(POETRY_RUN) python -m unittest discover tests --verbose $(PYK_TEST_ARGS)
+	$(POETRY_RUN) python -m unittest discover tests --failfast --verbose $(PYK_TEST_ARGS)
 
 test-integration: poetry-install
-	$(POETRY_RUN) python -m unittest discover integration_tests --verbose $(PYK_TEST_ARGS)
+	$(POETRY_RUN) python -m unittest discover integration_tests --failfast --verbose $(PYK_TEST_ARGS)
 
 test-pyk: poetry-install
 	$(POETRY_RUN) $(MAKE) -C pyk-tests
--- a/pyk/k-files/kore-rpc-test.k
+++ b/pyk/k-files/kore-rpc-test.k
@@ -1,0 +1,16 @@
+module KORE-RPC-TEST-SYNTAX
+    imports INT-SYNTAX
+endmodule
+
+
+module KORE-RPC-TEST
+    imports KORE-RPC-TEST-SYNTAX
+
+    rule [r01]: 0 => 1
+    rule [r12]: 1 => 2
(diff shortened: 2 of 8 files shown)


pyk/src/pyk/integration_tests/kore/test_kore_client.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class SimpleKoreClientTest(KoreClientTest):
+    KOMPILE_MAIN_FILE = 'k-files/kore-rpc-test.k'
+    KORE_MODULE_NAME = 'KORE-RPC-TEST'
+
+    TEST_DATA: Final[Tuple[Tuple[int, Mapping[str, Any], ExecuteResult], ...]] = (
+        (0, {}, BranchingResult(state=state(2), depth=2, next_states=(state(4), state(3)))),
+        (0, {'max_depth': 2}, DepthBoundResult(state=state(2), depth=2)),
+        (4, {}, StuckResult(state=state(6), depth=3)),
```
ehildenb: Why is this reporting back depth 3? Sholudn't it be depth 2? 4 => 5 => 6 is two rewrite steps.

tothtamas28: @jberthold, is this on purpose or a bug in the backed?

jberthold: I think it is probably a bug, resulting from the logic that implements the steps.
When we are in state `6` here, the execution does perform a "rewrite step", but which results in state `Remaining 6` (could not be rewritten). However, the "remaining state" would still be simplified so it is not necessarily the same as before (it is here, of course). 
Will analyse this example to confirm that, and fix it. I think calling the simplifier again is redundant here anyway but have to check.

ehildenb (approved):

pyk/src/pyk/integration_tests/kore/test_kore_client.py
ehildenb: Issue made: https://github.com/runtimeverification/haskell-backend/issues/3210

PR #2807, opened 2022-08-12
Kast - parse with the rule grammar

### LINKED ISSUE #2796: kast: allow parsing using the Rule grammar
Currently `kast` will not allow direct parsing using the rule grammar.

I'd propose a new flag `kast --rule-grammar`, which allows using the full rule-grammar parser from the command-line.

This is needed for the pyk library, because we need to parse expressions like `N:Int ==Int 3`, or `<k> foo => bar </k>`.

We still want to specify the sorts for these parses, we just want the grammar from rule language added as well.

We do not _need_ configuration completion here, we will only ever pass in single cells for the moment. But if configuration completion is easy to add, then that would be good too.

radumereuta:
Fixes: #2796

```console
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule
A=>#token("2","Int")
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps ""
#ruleRequiresEnsures(#SemanticCastToInt(A)=>#token("2","Int"),#token("true","Bool"),#token("true","Bool"))
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts"
#ruleRequiresEnsures(A=>#token("2","Int"),isInt(A),#token("true","Bool"))
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,body"
A=>#token("2","Int")
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,concretizeCells,body"
[Error] Compiler: AC matching of multiple cell variables not yet supported.
encountered variables [A, _DotVar0] in cell <generatedTop>
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,addImplicitComputationCell,concretizeCells,body"
`<generatedTop>`(`<k>`(``#SemanticCastToInt(A)=>#token("2","Int")``~>V0),V1,V2)
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,addImplicitComputationCell,concretizeCells,body"
`<generatedTop>`(`<k>`(``A=>#token("2","Int")``~>V0),V1,V2)
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,addImplicitComputationCell,concretizeCells,body" --output kore
kseq{}(inj{SortGeneratedTopCell{}, SortKItem{}}(Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortInt{}, SortKItem{}}(\rewrites{SortInt{}}(VarA:SortInt{},\dv{SortInt{}}("2"))),VarV0:SortK{})),VarV1:SortGeneratedCounterCell{},VarV2:SortStateCell{})),dotk{}())
radu@radu:~/work/test$ 
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,addImplicitComputationCell,concretizeCells,body"
`<generatedTop>`(`<k>`(``A=>#token("2","Int")``~>V0),V1,V2)
radu@radu:~/work/test$ kast -e 'A:Int => 2' --input rule --steps "resolveCasts,addImplicitComputationCell,concretizeCells,body" --output pretty
<generatedTop>
  <k>
    A => 2 ~> V0 ~> .
  </k>
  V2
</generatedTop>
radu@radu:~/work/test$ kast --steps help
For --input rule, apply these steps, in this order, separated by comma:
   macros - apply macro transformations
   closeCells - transform #Dots and #NoDots, into appropriate collection elements
   resolveCasts - transform #SemanticCastToSort nodes to side conditions
   addImplicitComputationCell - add the <generatedTop> cell
   concretizeCells - configuration concretization
   body - return only the body of the rule
```
Here are a few more quick examples. Please try it out.
I think this gives us the most flexibility and future upgradeability.
I chose only the steps that @ehildenb requested.
I added `macro` because I could. It was supposed to be there in the first place.
The order is fixed. Some of the steps rely on each other to be applied first.

I like this approach because it groups together all the kompilation steps, instead of adding more and more options that can get lost in the list of `kast --help`.


kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -92,14 +105,51 @@ public int run() {
...
+                        parsed = ExpandMacros.forNonSentences(unparsingMod, files.get(), def.kompileOptions, false).expand(parsed);
+                    }
+                    ConfigurationInfoFromModule configInfo = new ConfigurationInfoFromModule(mod);
+                    LabelInfo labelInfo = new LabelInfoFromModule(mod);
+                    SortInfo sortInfo = SortInfo.fromModule(mod);
+
+                    Rule r = (Rule) new AddImplicitComputationCell(configInfo, labelInfo).apply(mod, Rule.apply(parsed, BooleanUtils.TRUE, BooleanUtils.TRUE, Att.empty()));
+                    r = (Rule) new ConcretizeCells(configInfo, labelInfo, sortInfo, mod, true).concretize(mod, r);
+
+                    kprint.get().prettyPrint(def.kompiledDefinition, unparsingMod, s -> kprint.get().outputFile(s), r.body(), sort);
```
ehildenb:
```suggestion
                    Rule r = (Rule) new AddImplicitComputationCell(configInfo, labelInfo).apply(mod, Rule.apply(parsed, BooleanUtils.TRUE, BooleanUtils.TRUE, Att.empty()));
                    r = (Rule) new ConcretizeCells(configInfo, labelInfo, sortInfo, mod, true).concretize(mod, r);

                    kprint.get().prettyPrint(def.kompiledDefinition, unparsingMod, s -> kprint.get().outputFile(s), r.body(), sort);
```

Can we make it a flag whether we do configuration completion or not? Something like `--configuration-completion`? We can default to having it turned off.

We should add a test of both uses of the flag.

ehildenb:
With this definition as `imp.k`:

```
// Copyright (c) 2014-2019 K Team. All Rights Reserved.

module IMP-SYNTAX
  imports BOOL-SYNTAX
  imports INT-SYNTAX
  imports ID

  syntax AExp  ::= Int | Id
                 | "-" Int                    [klabae(-_), symbol, format(%1%2)]
                 | AExp "/" AExp              [klabel(_/_), symbol, left, strict, color(pink)]
                 > AExp "+" AExp              [klabel(_+_), symbol, left, strict, color(pink)]
                 | "(" AExp ")"               [bracket]
  syntax BExp  ::= Bool
                 | AExp "<=" AExp             [klabel(_<=_), symbol, seqstrict, latex({#1}\leq{#2}), color(pink)]
                 | "!" BExp                   [klabel(!_), symbol, strict, color(pink)]
                 > BExp "&&" BExp             [klabel(_&&_), symbol, left, strict(1), color(pink)]
                 | "(" BExp ")"               [bracket]
  syntax Block ::= "{" "}"                    [klabel({}), symbol]
                 | "{" Stmt "}"               [klabel({_}), symbol, format(%1%i%n%2%d%n%3)]
  syntax Stmt  ::= Block
                 | Id "=" AExp ";"            [klabel(_=_;), symbol, strict(2), color(pink), format(%1 %2 %3%4)]
                 | "if" "(" BExp ")"
                   Block "else" Block         [klabel(if(_)_else_), symbol, strict(1), colors(yellow, white, white, yellow), format(%1 %2%3%4 %5 %6 %7)]
                 | "while" "(" BExp ")" Block [klabel(while(_)_), symbol, colors(yellow,white,white), format(%1 %2%3%4 %5)]
                 > Stmt Stmt                  [klabel(__), symbol, left, format(%1%n%2)]

  syntax Pgm ::= "int" Ids ";" Stmt           [klabel(int_;_), symbol, format(%1 %2%3%n%4), colors(yellow,pink)]
  syntax Ids ::= List{Id,","}                 [klabel(_,_), symbol, format(%1%2 %3)]
endmodule

module IMP
  imports IMP-SYNTAX
  imports BOOL
  imports INT
  imports MAP
  imports SET
  imports BASIC-K

  syntax KResult ::= Int | Bool
  syntax KItem   ::= Pgm

  configuration <T color="yellow">
                  <k color="green"> $PGM:Pgm </k>
                  <state color="red"> .Map </state>
                </T>

  rule <k> X:Id => I ...</k> <state>... X |-> I ...</state>

  rule I1 / I2 => I1 /Int I2  requires I2 =/=Int 0
  rule I1 + I2 => I1 +Int I2
  rule - I1 => 0 -Int I1

  rule I1 <= I2 => I1 <=Int I2
  rule ! T => notBool T
  rule true && B => B
  rule false && _ => false

  rule {} => .   [structural]
  rule {S} => S  [structural]

  rule <k> X = I:Int; => . ...</k> <state>... X |-> (_ => I) ...</state>

  rule S1:Stmt S2:Stmt => S1 ~> S2  [structural]

  rule if (true)  S else _ => S
  rule if (false) _ else S => S

  rule while (B) S => if (B) {S while (B) S} else {}  [structural]

  rule <k> int (X,Xs => Xs);_ </k> <state> Rho:Map (.Map => X|->0) </state>
    requires notBool (X in keys(Rho))
  rule int .Ids; S => S  [structural]
endmodule
```

And `kompile imp.k --main-module IMP --syntax-module IMP`.

Then here are some input-output pairs (and what I would like them to be):

```
+ kast --input rule --sort Int --expression 3
`<generatedTop>`(`<T>`(`<k>`(#token("3","Int")~>V0),V1),V2)
# Should be: #token("3","Int")

+ kast --input rule --sort AExp --expression '3 + 4'
`<generatedTop>`(`<T>`(`<k>`(`_+_`(#token("3","Int"),#token("4","Int"))~>V0),V1),V2)
# Should be: `_+_`(#token("3","Int"),#token("4","Int")

+ kast --input rule --sort Int --expression '3 +Int 4'
`_+Int_`(#token("3","Int"),#token("4","Int"))
# Looks good.

+ kast --input rule --sort Int --expression '3 +Int V'
`_+Int_`(#token("3","Int"),#SemanticCastToInt(V))
# Looks OK. An option to remove semantic casts may be nice.

+ kast --input rule --sort KCell --expression '<k> . </k>'
`<generatedTop>`(`<T>`(`<k>`(.K),V0),V1)
# Should be: `<k>`(.K)
```

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
radumereuta: So you are ok with `<generatedTop>` cell?
What about `--sort-annotations` default to false to remove the `#SemanticCastToXXX`?

ehildenb: It would be better if we could handle the user supplying `<generatedTop>` cell as well. I guess we need to make sure to include the configuration module into the grammar we use?

So you are saying that default behavior would be to remove the semantic casts, and with `--sort-annotations` you can keep them? That sounds OK, but I think it's less descriptive of what it's actually doing, which is removing semantic casts, `--no-remove-semantic-casts`. Up to you on, either works for me.

k-distribution/tests/regression-new/kast-rule/kast-in2.kast
```diff
@@ -0,0 +1 @@
+<k> init(0) => 2 </k>
```
nishantjr, line 1: This looks like it is evaluating the `init()` function. ... or is that a macro now?

dwightguth:
I'm not going to review the code in depth yet, but `closeCells` and `resolveCasts` should probably not be optional. They should always be applied.

radumereuta:
> I'm not going to review the code in depth yet, but `closeCells` and `resolveCasts` should probably not be optional. They should always be applied.

Yes, they will always apply by default, but if you don't want them, then they won't. Your call. I can see situations where you may want that.
```
    --steps
      Apply specified kompilation steps to the parsed term. Only for --input 
      rule. Use --steps help for a detailed description of available steps.
      Default: [closeCells, resolveCasts, body]
```

dwightguth:
Additionally, I definitely see compilation passes that aren't on this list that, if not applied, mean that a term passed to this parser is not necessarily identical in meaning once parsed to the same term when put in a rule. For example, `_ +Int _`

dwightguth:
We should not be exposing the internal details of the compiler to users in the way that we would be if we completely disable all configuration completion. This will only lead to confusion and things not working as expected.

dwightguth:
As a general rule of thumb, if a compiler pass desugars some construct that does not exist in the compiled definition to some construct that does exist in the compiled definition, that desugaring process is probably not optional.

radumereuta:
> Additionally, I definitely see compilation passes that aren't on this list that, if not applied, mean that a term passed to this parser is not necessarily identical in meaning once parsed to the same term when put in a rule. For example, `_ +Int _`

> We should not be exposing the internal details of the compiler to users in the way that we would be if we completely disable all configuration completion. This will only lead to confusion and things not working as expected.

Not an exact rule parser. Take this as an advanced/experimental feature. This allows Everett to work on his experiments with more flexibility. Yes, some of the steps will fail if not applied properly, as you may see in the examples from the first comment. But this allows me to say: _do this_ instead of adding a new PR to the K repo, which can take way longer.

> As a general rule of thumb, if a compiler pass desugars some construct that does not exist in the compiled definition to some construct that does exist in the compiled definition, that desugaring process is probably not optional.

Agree, see my previous response. Let's address these in the future. I've added this system so people can experiment with it. If you would like, I can add a big WARNING to the help message.

ehildenb:
I think we should get it merged if it's passing the tests I supplied, and we can iterate. We won't know how it needs to be used until people start trying to use it. Two immediate cases are:

- Verbatim tokens in pyk, where you need to parse to get the actual AST.
- LaTeX rendering of things in code blocks.

ehildenb:
People will let us know when it does unexpected things.

radumereuta:
This still needs a proper code review.
@ehildenb I think I am providing by default what you requested. Check the committed tests and the first comment in this thread for more examples of how to use the `--steps` option.

ehildenb (approved):

ehildenb:
@radumereuta the new feature looks self-contained, so I approved. Could you also add a check that the user doesn't supply `--steps` unless it's also `--input rule`?

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -80,26 +94,89 @@ public static List<com.google.inject.Module> getModules() {
...
+                        r = (Rule) new ConcretizeCells(configInfo, labelInfo, sortInfo, mod, true).concretize(mod, r);
+                    K result = options.steps.contains(KompileSteps.body) ? r.body() : KApply(KLabel("#ruleRequiresEnsures"), KList(r.body(), r.requires(), r.ensures()));
+                    kprint.get().prettyPrint(def.kompiledDefinition, unparsingMod, s -> kprint.get().outputFile(s), result, sort);
                 }
+
+                sw.printTotal("Total");
+                return 0;
+            } else if (!options.steps.equals(Lists.newArrayList(KastFrontEnd.KompileSteps.closeCells, KastFrontEnd.KompileSteps.resolveCasts, KastFrontEnd.KompileSteps.body))) {
+                throw KEMException.innerParserError("Option --steps is available only with --input rule.");
             }
```
ehildenb: This seems like a less good way to tell that the option wasn't set. Can we just store a Null value if it's unset, and then if it's Null set the default here? It seems like a bad idea to have it be that we can't tell between the option not being set, and the option being set to the default (the user could say `--steps closeCells,resolveCasts,body` on the CLI, without `--input rule`, and this check wouldn't trigger).

radumereuta: I did it this way because I would like to have `kast -h` say what the default steps are:
```
--steps
      Apply specified kompilation steps to the parsed term. Only for --input 
      rule. Use --steps help for a detailed description of available steps.
      Default: [closeCells, resolveCasts, body]
```
Setting `null` is going to say `Default: []`
I'm not sure if there is a way to actually check if the option has actually been provided.
In any case, it's ignored if `--input` is not `rule`

dwightguth (changes requested):
As I said, we absolutely should not be exposing to users the internal behavior of the compiler. Compilation steps that, if not run, would lead to output parses which contain syntactic constructs which exist solely to implement a particular feature in the compiler and do not exist in a compiled definition must be run unconditionally. The same is true for any passes which, when not run, would lead to the behavior of the resulting parse being different from if that term were part of the rule. A good example of this latter is the resolution of anonymous variables.

radumereuta:
@dwightguth I'm not sure what you are on about here, but this PR is trying to add some new features that Everett requested:
- parse expressions like `N:Int ==Int 3`, or `<k> foo => bar </k>` (need to use the rule grammar)
- optionally apply configuration completion
- optionally remove sort annotations

This can be used for:
- Verbatim tokens in pyk, where you need to parse to get the actual AST.
- LaTeX rendering of things in code blocks.

The rule parser outputs some labels which are desugared in some kompile steps:
- `...` is desugared by `closeCells`. Applied by default, but if you want to output LaTeX, you may want to know about it. You have the option to do so.
- `concretizeCells` - optional if you choose to apply it.
- `resolveCasts` - move the sort annotations as predicates in the side conditions - requested by Everett.

This is no more experimental than adding `--disable-ceil-simplification-rules`. It only adds features. It doesn't break any backward compatibility. It only allows for more flexibility in our tools.
If you are worried this can fail in some obscure way, then we will deal with it when that time comes. It's a learning process.

In the meantime, I would appreciate it if your PR comments are more on point:
```
Example -> Output = Wrong because ...
```

dwightguth:
I'm aware that nothing is forcing anyone to use this functionality. My concern is that you are exposing to the user internal implementation details of how rules are compiled. That means if users come to use those details, they become part of the api of the compiler that we will have to continue to support if we intend to be able to use all the features of k. This significantly limits the ability of the compiler to refactor how we choose to compile rules. That means that that aspect of this change is to the detriment of the maintainability of the system as a whole, even if only one particular project happens to make use of it. So far as I'm aware, Everett has no interest in actually depending on the existence of constructs like `#dots`, `#SemanticCastToSort`, or `#ruleRequiresEnsures` so let's just remove the ability to turn off the passes that desugar constructs like this and call it good.

radumereuta:
@ehildenb will that work for you for now?

nishantjr:
@radumereuta At least for ksummarizer, that will be good enough.
If not hard, a `--complete-configurations` option may make sense, (off for use for latex generation, on for most other applications).

radumereuta:
Try to hide every label that is considered internal.
So look over the kompile steps and see which ones can be applied unconditionally.

Anonymous variables resolution should be on by default but we should be able to disable it.
For variables we can use injections. kore should be able to show the correct output. We need to call AddSortInjections step.

radumereuta:
A few examples after the update:
```console

$ kast -e 'A:Int => .K' --input rule
A=>.K
$ kast -e 'A:Int => .K' --input rule --output kore
\rewrites{SortK{}}(kseq{}(inj{SortInt{}, SortKItem{}}(VarA:SortInt{}),dotk{}()),dotk{}())
$ kast -e 'A:Int => .K' --input rule --steps configConcretization
[Error] Critical: Invalid value for --steps parameter. Allowed values:[help,
concretizeCells, anonVars]
$ kast -e 'A:Int => .K' --input rule --steps concretizeCells,anonVars
`<generatedTop>`(`<k>`(``A=>.K``~>V0),V1)
$ kast -e '_:Int => .K' --input rule --steps concretizeCells,anonVars
`<generatedTop>`(`<k>`(``V0=>.K``~>V1),V2)
$ kast -e '_:Int => .K' --input rule --steps ,
_=>.K
$ kast -e '_:Int => .K' --input rule
V0=>.K
$ kast -h
    --steps
      Apply specified kompilation steps to the parsed term. Only for --input 
      rule. Use --steps help for a detailed description of available steps.
      Default: [anonVars]
$ kast --steps help
For --input rule, apply these steps, in this order, separated by comma:
   concretizeCells - configuration concretization
   anonVars - rename anonymous variables to be unique
   , - empty list of steps
```
@nishantjr please make sure this is enough for your requirements
@dwightguth please have another look if this hides enough of the kompilation internals to not cause problems in the future.

nishantjr:
Looks good @radumereuta , thanks.

ehildenb:
Looks good to me as well!

kernel/src/main/java/org/kframework/parser/KRead.java
```diff
@@ -64,6 +64,8 @@ public K prettyRead(Module mod, Sort sort, CompiledDefinition def, Source source
                 return new KoreParser(mod.sortAttributesFor()).parseString(stringToParse);
             case PROGRAM:
                 return def.parseSingleTerm(mod, sort, kem, files, stringToParse, source);
+            case RULE:
+                throw KEMException.criticalError("Should have been handled directly by the kast front end: " + inputMode);
```
dwightguth: this should probably be an internal error.

dwightguth (approved):
This definitely looks better. I have one comment I suggest you fix before you merge, but I'm approving. There's still maybe a few things we could argue might want to be included in rule passes, but as long as we're fine with the understanding that the output of the rule parser here shouldn't be used to depend on any syntax from `kast.md`, it's probably fine. The most problematic cases have been addressed; at least what remains ought to have a clear and obvious syntax explicitly declared in kast.md, which makes it less of an internal detail by a certain amount.

PR #2808, opened 2022-08-15
[pyk] Add KDefinition.init_config, for initializing KRun

ehildenb:
Similar to `KDefinition.empty_config`, this is providing a way to construct initial configurations automatically. This is in the same way that `KRun` does this, by calling the `initGenerateTop(...)` K function, and rewriting it to the entire configuration.

Note that this is different than how the current frontend works. The frontend builds a `Map` where the keys are `KConfigVar` and the values are the k terms, and then builds a map lookup `CONFIG_VAR_MAP [ CONFIG_VAR_NAM ]` in each position where configuration variables are used. This is probably because the concrete backends never bothered to support `KVariable`, and so don't have a way to build the term with holes in it and the configuration variables treated as actual variables.

This approach instead drops the configuration variables directly as `KVariable` in the configuration, so you then can use `substitute` to populate their values. This is more flexible for pyk library purposes. So for example, you would first call `KDefinition.init_config`, and then apply the substitution `Subst({ '$PGM': PGM_CONFIG_VAR_VALUE ... })` to instantiate the configuration.


geo2a (approved):
That looks good to me, thanks @ehildenb! I'll try leveraging these changes early next week. I think we can merge now though.

pyk/src/pyk/integration_tests/test_defn.py
```diff
@@ -56,65 +51,108 @@ def test_print_prove_rewrite(self):
...
-            ]
+        minimized_claim_rewrite_expected = (
+            f'<T>\n'
+            f'  <k>\n'
+            f'    ( X + Y => X +Int Y )\n'
+            f'  </k>\n'
+            f'  <state>\n'
+            f'    STATE\n'
+            f'  </state>\n'
+            f'</T>'
```
tothtamas28:
```suggestion
            '<T>\n'
            '  <k>\n'
            '    ( X + Y => X +Int Y )\n'
            '  </k>\n'
            '  <state>\n'
            '    STATE\n'
            '  </state>\n'
            '</T>'
```

Will include `flake8` again to catch these issues: #2814.

pyk/src/pyk/kast.py
```diff
@@ -1633,6 +1637,48 @@ def _kdefinition_empty_config(_sort):
...
+                    _token_var = _term.args[1]
+                    if type(_token_var) is KToken and _token_var.sort == KSort('KConfigVar'):
+                        return KVariable(_token_var.token)
+            return _kast
+
+        init_prods = [prod for prod in self.syntax_productions if 'initializer' in prod.att]
+        _init_prod = [prod for prod in init_prods if prod.sort == sort]
+        if len(_init_prod) != 1:
+            raise ValueError(f'Did not find unique initializer for sort: {sort}')
+        init_prod = _init_prod[0]
```
tothtamas28, line 1669: We use this idiom quite a lot, let's extract a function to `pyk.util`. Maybe something like this:

```python3
from typing import Iterable, TypeVar, cast

T = TypeVar('T')


def single(iterable: Iterable[T]) -> T:
    it = iter(iterable)
    sentinel = object()
    
    fst = next(it, sentinel)
    if fst is sentinel:
        raise ValueError('Expected a single element, found: none')
    fst = cast(T, fst)

    snd = next(it, sentinel)
    if snd is not sentinel:
        raise ValueError(f'Expected a single element, found: {fst}, {snd}, ...')

    return fst
```

tothtamas28 (approved):

pyk/src/pyk/kast.py
nishantjr: Nice! Could we have this take an error message too please.

nishantjr (approved):
Looks good!

pyk/src/pyk/kast.py
tothtamas28: Good idea. Or perhaps even

```python3
error: Union[str, Callable[[Tuple[T, ...]], str], None] = None
```

where the callable receives either `()` or `(fst, snd)`.

pyk/src/pyk/integration_tests/test_defn.py
ehildenb: Done.

pyk/src/pyk/kast.py
ehildenb: This sounds like a more extensive change, made an issue: https://github.com/runtimeverification/k/issues/2820

Issue #2801, opened 2022-08-11
Cleanups from KIT upstream


ehildenb:
- [ ] `kit.py` should be renamed `kit.py`: https://github.com/runtimeverification/k/pull/2787/files#r942027065
- [ ]

PR #2815, opened 2022-08-18
Add `flake8` to `pyk`

### LINKED ISSUE #2814: Include `flake8` in `pyk` code style checks
There are some issues that `flake8` flag as formatting errors (e.g. `F541 f-string is missing placeholders`) that `black` does not take care of. Thus `flake8` has to be included again with a configuration that does not conflict with `black`: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8.

We can also consider adding some useful plugins: https://flake8.pycqa.org/en/latest/user/using-plugins.html.

tothtamas28:
Fixes runtimeverification/k#2814

### DIFF
--- a/pyk/src/tests/test_kastManip.py
+++ b/pyk/src/tests/test_kastManip.py
@@ -172,11 +172,11 @@ class BoolMlPredConversionsTest(TestCase):
                 self.assertEqual(bool_actual, bool_expected)
 
     test_data_bool_to_ml_pred = (
-        ('equals-true', False, KApply(KLabel('#Equals', [Sorts.BOOL, Sorts.K]), [Bool.true, f(a)]), f(a)),
+        ('equals-true', KApply(KLabel('#Equals', [Sorts.BOOL, Sorts.K]), [Bool.true, f(a)]), f(a)),
     )
 
     def test_bool_to_ml_pred(self):
-        for name, _, ml_pred_expected, bool_in in self.test_data_bool_to_ml_pred:
+        for name, ml_pred_expected, bool_in in self.test_data_bool_to_ml_pred:
             with self.subTest(name):
                 ml_pred_actual = bool_to_ml_pred(bool_in)
                 self.assertEqual(ml_pred_actual, ml_pred_expected)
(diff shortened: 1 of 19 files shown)


pyk/src/tests/test_kastManip.py
```diff
@@ -176,7 +176,7 @@ def test_ml_pred_to_bool(self):
     )
 
     def test_bool_to_ml_pred(self):
-        for name, unsafe, ml_pred_expected, bool_in in self.test_data_bool_to_ml_pred:
+        for name, _, ml_pred_expected, bool_in in self.test_data_bool_to_ml_pred:
             with self.subTest(name):
                 ml_pred_actual = bool_to_ml_pred(bool_in)
                 self.assertEqual(ml_pred_actual, ml_pred_expected)
```
ehildenb: We should just remove the `unsafe` parameter from this test completely. I thought before it would be possible for these two to be invertible, but I don't think so, so we should probably remove the comment saying so, I've struck out the line-item on this issue: https://github.com/runtimeverification/k/issues/2801

pyk/src/tests/test_kastManip.py
```diff
@@ -226,8 +227,12 @@ def test_simplify_bool(self) -> None:
...
-            self.assertEqual(bool_out_actual, bool_out)
+            with self.subTest(i=test_name):
+                # When
+                bool_out_actual = simplify_bool(bool_in)
+
+                # Then
+                self.assertEqual(bool_out_actual, bool_out)
 
 
 class BuildClaimtest(TestCase):
```
ehildenb, line 238: Are the `When/Then` enforced by `flake8`, or your own style choice?

ehildenb: And I think you can just do `self.subTest(test_name)`, no need for `i=`?

ehildenb (approved):

pyk/src/tests/test_kastManip.py
tothtamas28: > Are the `When/Then` enforced by flake8, or your own style choice?

Personal preference. I like this [idiom](https://en.wikipedia.org/wiki/Given-When-Then) as it enforces a clean test structure.

tothtamas28: > And I think you can just do `self.subTest(test_name)`, no need for `i=`?

Indeed, good catch.

PR #2819, opened 2022-08-19
Switch to Poetry

tothtamas28:
This PR makes `pyk` use [Poetry](https://python-poetry.org/) for dependency management.

* Virtual environments are now managed by Poetry.
* All project configuration is moved to `pyproject.toml`


ehildenb:
It looks like a nice refactoring, some questions:

- Quick list of benefits you've noticed over what we've been doing so far?
- Have you tried this with a downstream repo? If you switch KEVM over to Poetry as well, does it play nicely with this poetry install?
- Before, we had a way to step into the virtual environment directly, so we could run pyk commands on the shell. Is it still possible here? Can we make sure that that use case is straightforward and documented?

tothtamas28:
> Quick list of benefits you've noticed over what we've been doing so far?

The main benefit for us right now is that we do not have to worry about virtual environments: `poetry install` will install the package and its dependencies in a virtual environment that it manages, and then `poetry run` can be used to run a command in that environment.

> Have you tried this with a downstream repo? If you switch KEVM over to Poetry as well, does it play nicely with this poetry install?

There shouldn't be a any differences there, `pip install` still works on the project (though the build backend changed). Will check to make sure. **Edit**: it works as expected.

> Before, we had a way to step into the virtual environment directly, so we could run pyk commands on the shell. Is it still possible here? Can we make sure that that use case is straightforward and documented?

`poetry shell` spawns a new subshell within the virtual environment (I think this is more convenient than sourcing). I will update the docs.

pyk/src/tests/test_split_config.py
```diff
@@ -1,10 +1,10 @@
...
-from ..kastManip import splitConfigFrom
+from pyk.kast import KApply, KSequence, KVariable
+from pyk.kastManip import splitConfigFrom
 
 
-class TestPyk(TestCase):
+class SplitConfigTest(TestCase):
     def test_splitConfigFrom(self):
         k_cell = KSequence([KApply('foo'), KApply('bar')])
         term = KApply('<k>', [k_cell])
```
ehildenb: Can this test be moved into `test_kastManip.py`?

tothtamas28: Done.

ehildenb (approved):
One final question to look at: the `package/package` script and `package/debian/rules.{focal.jammy}` files handle the installation of pyk for our packaging. Anything you can see that needs to change there? We're using `pip install --target ... --root ...` to get the correct installation location. Should poetry handle this directly instead? Maybe that can be another PR if so.

tothtamas28:
> One final question to look at: the `package/package` script and `package/debian/rules.{focal.jammy}` files handle the installation of pyk for our packaging. Anything you can see that needs to change there?

Well, not really. Maybe if we wanted to be super prudent we could `poetry build` the package in some earlier step of the pipeline first and then install the generated wheel here, but for such a straightforward build I do not see the benefit. Though Poetry will be very useful if we decide to publish the package in PyPI or in a GitHub repo (Poetry supports publishing to private repos too).

PR #2822, opened 2022-08-19
Fix `pyk` packaging on Jammy

tothtamas28:


package/debian/rules.jammy
```diff
@@ -33,6 +33,7 @@ override_dh_auto_build:
 override_dh_auto_install:
 	package/package
 	mv $(DESTDIR)$(PREFIX)/local/bin/pyk $(DESTDIR)$(PREFIX)/bin/pyk
+	mv $(DESTDIR)$(PREFIX)/local/bin/kit.py $(DESTDIR)$(PREFIX)/bin/kit.py
```
tothtamas28, line 36: This is the key modification. I'm unsure why the build succeeded earlier, maybe the Python build didn't generate `kit.py`?

package/debian/Dockerfile
```diff
@@ -42,15 +42,14 @@ RUN    apt-get update            \
         pkg-config               \
         python3                  \
         python3-all              \
+        python3-distutils        \
```
ehildenb, line 45: This probably means that we need to put this in `package/debian/control` as a `Build-Depend`?

ehildenb (approved):

package/debian/rules.jammy
ehildenb: Weird that it didn't break before, but happy it's working now!

PR #2823, opened 2022-08-22
Move and rename some functions in pyk

### LINKED ISSUE #2816: Clean up naming in `pyk`
#2815 includes `pep8-naming` in code checks, which right now flags ca. 150 naming issues in `pyk`.

First, enable these errors by updating the `flake8` configuration in `setup.cfg`:

```ini
[flake8]
...
extend-ignore = ...  # remove N from the list
```

Then ensure that `make check-flake8` succeeds.

tothtamas28:
Fixes #2816

### DIFF
--- a/pyk/src/pyk/prelude.py
+++ b/pyk/src/pyk/prelude.py
@@ -68,15 +68,13 @@ def build_assoc(unit: KInner, label: Union[str, KLabel], terms: Iterable[KInner]
     return res or unit
 
 
-def buildCons(unit, cons, ls):
-    """Build a cons operator term given the cons and unit ops.
-
-    -   Input: unit, cons, and list of elements to join.
-    -   Output: cons-list style construction of the joined term.
-    """
-    if len(ls) == 0:
+def build_cons(unit: KInner, label: Union[str, KLabel], terms: Iterable[KInner]) -> KInner:
+    it = iter(terms)
+    try:
+        fst = next(it)
+        return KApply(label, (fst, build_cons(unit, label, it)))
+    except StopIteration:
         return unit
-    return KApply(cons, [ls[0], buildCons(unit, cons, ls[1:])])
 
 
 def token(x: Union[bool, int, str]) -> KToken:
--- a/pyk/src/pyk/prelude.py
+++ b/pyk/src/pyk/prelude.py
@@ -107,10 +107,10 @@ def leInt(i1, i2):  # noqa: N802
 def mlEquals(  # noqa: N802
     term1: KInner,
     term2: KInner,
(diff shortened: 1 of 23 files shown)


pyk/src/pyk/prelude.py
```diff
@@ -68,15 +68,14 @@ def build_assoc(unit: KInner, label: Union[str, KLabel], terms: Iterable[KInner]
...
-def buildCons(unit, cons, ls):
-    """Build a cons operator term given the cons and unit ops.
-
-    -   Input: unit, cons, and list of elements to join.
-    -   Output: cons-list style construction of the joined term.
-    """
-    if len(ls) == 0:
+def build_cons(unit: KInner, label: Union[str, KLabel], terms: Iterable[KInner]) -> KInner:
+    label = label if type(label) is KLabel else KLabel(label)
+    it = iter(terms)
```
ehildenb: Is it necassary to process `label` like this? It's being used directly in `KApply`, which already does the processing.

maybe we should make a new type for the type `Union[str, KLabel]`, or can we somehow automatically lift type `str` to type `KLabel`?

ehildenb: The better typing can be dealt with in another PR.

tothtamas28: > Is it necassary to process `label` like this?

Good point, fixed.

> can we somehow automatically lift type str to type KLabel?

We can write a helper function for `label if type(label) is KLabel else KLabel(label)`, but I think that's it.

pyk/src/pyk/prelude.py
```diff
@@ -89,52 +87,55 @@ def token(x: Union[bool, int, str]) -> KToken:
...
 
 
 # TODO default sort K can be tightened using basic type inference
-def mlEquals(
-    term1: KInner, term2: KInner, sort1: Union[str, KSort] = Sorts.K, sort2: Union[str, KSort] = Sorts.K
+def mlEquals(  # noqa: N802
+    term1: KInner,
+    term2: KInner,
+    sort1: Union[str, KSort] = Sorts.K,
+    sort2: Union[str, KSort] = Sorts.K,
```
ehildenb: I think we should consider renaming these to `sort_args` and `sort_return` or something, instead of `sort1/2`, because people might think it means the sorts of `term1` and `term2`, when it doesn't. Sort1 should be the sort of both arguments, sort2 is the sort that the result is projected into.

Doesn't have to be this PR

tothtamas28: Done.

ehildenb (approved):

PR #2826, opened 2022-08-22
 Fix `pyk` packaging on Arch

tothtamas28:


ehildenb (approved):

package/arch/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN    pacman -Sy   --noconfirm archlinux-keyring \
                        openssh                    \
                        python                     \
                        python-pip                 \
+                       python-poetry              \
```
ehildenb, line 25: The `makedepends` in `package/arch/PKGBUILD` should probably be updated with this build dependency.

ehildenb: Feel free to add back automerge.

tothtamas28: Done.

PR #2827, opened 2022-08-22
Add a check to detect duplicate k cell declarations

### LINKED ISSUE #2171: [Bug] [kompile] - does not report duplicate <k> cell declaration

SchmErik:
This pull request adds a check to detect when a `<k>` cell is being declared multiple times.

fixes: https://github.com/runtimeverification/k/issues/2171


kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -285,6 +285,17 @@ protected Definition resolveConfigBubbles(Definition definition, Module defaultC
...
+        // Note: this is done before the configuration bubbles are resolved because it allows us to use location
+        // information from the sentences that declare the configuration cells.
+        List<Bubble> KCellDeclarations = stream(definition.getAllConfigDecls())
+                .filter(s -> s.contents().contains("<k>") && s.contents().contains("</k>"))
+                .collect(Collectors.toList());
+        if (KCellDeclarations.size() > 1) {
+            KCellDeclarations.stream().forEach(k ->
+                errors.add(KEMException.compilerError("Multiple K cell declarations detected. Only one declaration is allowed.", k)));
+        }
+
```
radumereuta:
```
configuration <k> "<k>" </k>
```
If you do this at the text level, you will fail for such an example.

But you got the right idea. Below this function you have 
```java
Definition defWithParsedConfigs = ...

// replace config bubbles with the generated syntax and rules
```
You have to write a [K visitor](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/CloseCells.java#L102) over `defWithParsedConfigs` that looks for `#token("k","#CellName")`

Here is an example configuration
```
configuration <k> $PGM:KItem </k>
              <state> .Map </state>
Configuration(#cells(#configCell(#token("k","#CellName"),#cellPropertyListTerminator(.KList),#SemanticCastToKItem(#token("$PGM","KConfigVar")),#token("k","#CellName")),#configCell(#token("state","#CellName"),#cellPropertyListTerminator(.KList),`.Map`(.KList),#token("state","#CellName"))),#token("true","Bool"),[org.kframework.attributes.Location(Location(11,19,12,40)), org.kframework.attributes.Source(Source(/home/radu/work/test/test.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol])])
```

radumereuta (changes requested):

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
SchmErik: Ah, Thanks! I was kind of looking for something like this. In hindsight, I should have asked you!

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -339,6 +359,13 @@ private Definition resolveConfigBubbles(Definition def) {
                 Module(m.name(), m.imports(), parsed.get(m.name()).localSentences(), m.att()),
                 "replace configs").apply(defWithCaches);
 
+        List<Configuration> KCellDeclarations = stream(defWithParsedConfigs.Configs())
+                .filter(c -> configNameMatches.process(c.body(), "k")).collect(Collectors.toList());
+        if (KCellDeclarations.size() > 1) {
+            KCellDeclarations.stream().forEach(k ->
+                    errors.add(KEMException.compilerError("Multiple K cell declarations detected. Only one declaration is allowed.", k)));
+        }
+
```
radumereuta, line 371: Somehow this seems a bit convoluted.
Try this instead:
```java
    private static void checkDuplicateKCells(Definition defWithParsedConfigs) {
        // check for duplicate <k> cell declarations
        List<K> kcells = new ArrayList<>();
        stream(defWithParsedConfigs.mainModule().sentences())
                .filter(s -> s instanceof Configuration)
                .forEach(s -> new VisitK() {
                    @Override
                    public void apply(KApply k) {
                        if (k.klabel().equals(KLabel("#configCell"))) {
                            KToken kt = (KToken) k.klist().items().get(0);
                            assert kt.sort().equals(Sorts.CellName());
                            if (kt.s().equals("k"))
                                kcells.add(k);
                        }
                        super.apply(k);
                    }
                }.apply(((Configuration) s).body()));
        if (kcells.size() > 1)
            throw KEMException.compilerError("Only one <k> cell declaration is allowed. Previously declared at "
                    + kcells.get(0).source().orElse(null) + " and " + kcells.get(0).location().orElse(null), kcells.get(1));
    }
```

k-distribution/tests/regression-new/checks/duplicateKCells.k
```diff
@@ -0,0 +1,12 @@
+module DUPLICATERULES-SYNTAX
```
radumereuta:
```suggestion
module DUPLICATEKCELLS-SYNTAX
```

k-distribution/tests/regression-new/checks/duplicateKCells.k
```diff
@@ -0,0 +1,12 @@
...
+
+    configuration <k> $PGM:K </k>
+
+endmodule
+
+module DUPLICATEKCELLS
+    imports DUPLICATERULES-SYNTAX
```
radumereuta:
```suggestion
    imports DUPLICATEKCELLS-SYNTAX
```

radumereuta (changes requested):
I tried to test this but your solution didn't work when the two configurations are in the same module:

```
module TEST
    configuration <env>
                    <k> $PGM:Int </k>
                    <state> .Map </state>
                  </env>
    configuration <k> .K </k>
endmodule
```

or when you have nested <k> cells:
```
    configuration <k>
                    <k> $PGM:Int </k>
                    <state> .Map </state>
                  </k>
```


I added a code snippet that should work.
Please double check.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -115,6 +119,8 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
 
   lazy val productions: Set[Production] = sentences collect { case p: Production => p }
 
+  lazy val localConfigurations: Set[Configuration] = localSentences collect { case c: Configuration => c }
+
```
radumereuta: Changes to this file are no longer needed.

radumereuta (changes requested):
In my version of the code, I throw an exception because of this example:
```
    configuration <k>
                    <k> $PGM:Int </k>
                    <state> .Map </state>
                  </k>
```
The compilation continues for a bit, then fails badly with a cryptic error.

SchmErik:
> In my version of the code, I throw an exception because of this example:
> 
> ```
>     configuration <k>
>                     <k> $PGM:Int </k>
>                     <state> .Map </state>
>                   </k>
> ```
> 
> The compilation continues for a bit, then fails badly with a cryptic error.

I tried to kompile this code with the current head and it's also resulting in an exception. We should file a separate issue for this...

radumereuta:
> > In my version of the code, I throw an exception because of this example:
> > ```
> >     configuration <k>
> >                     <k> $PGM:Int </k>
> >                     <state> .Map </state>
> >                   </k>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The compilation continues for a bit, then fails badly with a cryptic error.
> 
> I tried to kompile this code with the current head and it's also resulting in an exception. We should file a separate issue for this...

What I was trying to say is that in my code I have
```java
if (kcells.size() > 1)
            throw KEMException.compilerError("Only one <k> cell declaration is allowed. Previously declared at "
                    + kcells.get(0).source().orElse(null) + " and " + kcells.get(0).location().orElse(null), kcells.get(1));
```
This means kompilation is stopped in this spot.
If you do it your way, kompile will run too far and get stuck in some asserts.
We have the opportunity to completely bypass that error with this message.

SchmErik:
ah, I see what you mean. I'm fine with taking either solution but the advantage to recording all declarations is that the user can understand where all of the declarations are located and gives them the ability to choose which ones to keep. I wonder if there's a way to do both: throw an exception that contains all of the exceptions...

radumereuta:
You could call `private void throwExceptionIfThereAreErrors()`. See usages.

radumereuta (approved):

PR #2829, opened 2022-08-24
Remove spurious `ABC` superclasses

tothtamas28:
Plus some other small fixes.


pyk/src/pyk/kast.py
```diff
@@ -132,7 +132,7 @@ def update(self, atts: Mapping[str, Any]) -> 'KAtt':
 EMPTY_ATT: Final = KAtt()
 
 
-class WithKAtt(KAst, ABC):
+class WithKAtt(ABC):
```
geo2a, line 135: I don't yet know the purpose of the `WithKAtt` class, but this looks suspiciously like a mistake. Should `WithKAttr` inherit  from `KAst` instead?

geo2a (approved):
Looks good to me! I only have one question about the `WithKAtt` class.

pyk/src/pyk/kast.py
tothtamas28: `WithKAtt` is a way to add an attribute `att: KAtt` to a class. The subclass implements `let_att` (i.e. defines how to "set" the `att`), and gets `map_att` and `update_atts` for free. E.g.:

```python3
class KVariable(KInner, WithKAtt)
    name: str
    att: KAtt

    ...

    def let_att(self, att: KAtt) -> 'KVariable':
        return self.let(att=att)
```

> Should `WithKAttr` inherit from `KAst` instead?

It could (with the intended meaning that whatever has an attribute `att` is also a `KAst`), but it is not necessary for `WithKAtt` to work.  Also, implementing `WithKAtt` would then make it necessary to implement `KAst` as well.

PR #2830, opened 2022-08-24
Apply python-project-template to `pyk`

tothtamas28:
https://github.com/runtimeverification/python-project-template/

### DIFF
--- a/pyk/.cruft.json
+++ b/pyk/.cruft.json
@@ -1,0 +1,18 @@
+{
+  "template": "https://github.com/runtimeverification/python-project-template.git",
+  "commit": "46685b928f6e7765d42b98c59cc0cdc5217e12f9",
+  "checkout": null,
+  "context": {
+    "cookiecutter": {
+      "project_name": "pyk",
+      "project_slug": "pyk",
+      "package_name": "pyk",
+      "version": "0.1.0",
+      "description": "",
+      "author_name": "Runtime Verification, Inc.",
+      "author_email": "contact@runtimeverification.com",
+      "_template": "https://github.com/runtimeverification/python-project-template.git"
+    }
+  },
+  "directory": null
+}
--- a/pyk/Makefile
+++ b/pyk/Makefile
@@ -30,23 +30,25 @@ poetry-install:
 
 POETRY_RUN := poetry run
 
+
 # Tests
 
-PYK_TEST_ARGS :=
+TEST_ARGS :=
 
 test: test-unit test-integration test-pyk test-kit
 
 test-unit: poetry-install
-	$(POETRY_RUN) python -m unittest discover tests --failfast --verbose $(PYK_TEST_ARGS)
+	$(POETRY_RUN) python -m unittest discover tests --failfast --verbose $(TEST_ARGS)
 
 test-integration: poetry-install
-	$(POETRY_RUN) python -m unittest discover integration_tests --failfast --verbose $(PYK_TEST_ARGS)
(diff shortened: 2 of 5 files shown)


pyk/.cruft.json
```diff
@@ -0,0 +1,18 @@
+{
+  "template": "https://github.com/runtimeverification/python-project-template.git",
+  "commit": "46685b928f6e7765d42b98c59cc0cdc5217e12f9",
+  "checkout": null,
+  "context": {
+    "cookiecutter": {
+      "project_name": "pyk",
+      "project_slug": "pyk",
+      "package_name": "pyk",
+      "version": "0.1.0",
```
ehildenb, line 10: This is pyk version, or template version?

ehildenb: If it's pyk version, we should probably try to make sure this stays in sync with actual version (like how KEVM makes sure on CI that they stay in sync)

pyk/Makefile
```diff
@@ -30,24 +30,24 @@ poetry-install:
...
-# Tests
 
-PYK_TEST_ARGS :=
+# Tests
 
 test: test-unit test-integration test-pyk test-kit
 
 test-unit: poetry-install
-	$(POETRY_RUN) python -m unittest discover tests --failfast --verbose $(PYK_TEST_ARGS)
+	$(POETRY_RUN) python -m unittest discover tests --failfast --verbose
```
ehildenb: We need a way to run an individual test, otherwise the turnaround time is too long. I was using this as to do `make test-intergration PYK_TEST_ARGS='-k defn_test'` for example, to just run the tests in defn_tests.

If you add a different way to run an individual test with a single short command (not having to manually activate virtual env or poetry), then that's fine, otherwise let's add this back.

pyk/pyproject.toml
```diff
@@ -6,21 +6,23 @@ build-backend = "poetry.core.masonry.api"
...
-mypy = "^0.971"
-pep8-naming = "^0.13.1"
+autoflake = "*"
+black = "*"
+flake8 = "*"
+flake8-bugbear = "*"
+flake8-comprehensions = "*"
+isort = "*"
+mypy = "*"
+pep8-naming = "*"
```
ehildenb, line 25: This will pull in the latest version of each?

pyk/.cruft.json
tothtamas28: This is the `version` parameter value passed to `cookiecutter` when first generating the project; `cruft` uses this information to regenerate the template for `cruft update`, i.e. when we want to apply a new version of the template to the project. So it is technically `pyk` version, but it is not supposed to be modifed.

pyk/pyproject.toml
tothtamas28: This expresses that any version of the respective dependency is accepted. So when we run `poetry install` for the first time, Poetry will not have any additional constraints other than what these dependencies may put on each other's version. In practice, Poetry will probably favor the most recent versions unless there is a conflict.

Note that we still have full control on the version of these dependencies, as `poetry.lock`, committed to the repo, freezes them to a given version, and we can decide when to bump a given tool to a newer version (like for example in e6b08046f5709ab8b66fc950943f677324861383). It is just not managed on the template level.

pyk/Makefile
tothtamas28: I see, this is a nice trick. I'll add it back.

pyk/pyproject.toml
ehildenb: Sounds good.

ehildenb (approved):

PR #2831, opened 2022-08-26
Fix comm simplification rule

radumereuta:
Fixes: https://github.com/runtimeverification/k/pull/2754
If the body is not changed, the rule remains but w/o the `comm` attribute.

I've tested it locally by making `+Int` comm.
I can't add another test unless I make changes to domains.md.
But if I'm making changes to builtin operations, shouldn't we add comm to more productions?


ehildenb:
We should add tests for bugs.

radumereuta:
Test added
@dwightguth @ehildenb please re-review.

dwightguth (approved):

PR #2834, opened 2022-08-29
Add kprove options to filter claims

### LINKED ISSUE #2812: [kprove] Control which claims the backend discharges
On the CLI, we'd like to be able to dynamically mark claims as trusted, or to exclude them. This could come in 3 forms (I think we could have uses for all 3):

- `--trusted CLAIM_ID_LIST`: Mark this set of claims as trusted.
- `--exclude CLAIM_ID_LIST`: Remove these claims from the definition before sending to backend.
- `--claims CLAIM_ID_LIST`: Only keep the named claims in the definition before sending to the backend (remove the others).

This allows us to select subsets of claims to discharge with different calls to the prover, instead of having to discharge them all at the same time. Currently, for the Foundry project, we are automatically generating 1 K-claim per foundry test, which means there will be at least hundreds of claims. For testing, it's a lot easier if we can say which claims we actually care about running for that run. For this, only the third flag `--claims` is needed. The other two are variants I thought others may find useful.

radumereuta:
Fixes: #2812


dwightguth (approved):

kernel/src/main/java/org/kframework/kprovex/ProofDefinitionBuilder.java
```diff
@@ -53,6 +60,7 @@ public Tuple2<Definition, Module> build(File specFile, String specModuleName, bo
         Map<String, Module> modulesMap = modules.stream().collect(Collectors.toMap(Module::name, m -> m));
         Definition parsedDefinition = compiledDefinition.getParsedDefinition();
         Module specModule = getModule(specModuleNameUpdated, modulesMap, parsedDefinition);
+        specModule = filter(specModule);
```
ehildenb, line 63: Do we need to apply it to _every_ module that is not in the `--main-module`? Will this only apply it to the `--spec-module`?

If it's not the case (will apply to all modules), feel free to automerge. The tests make me think that it should work on all the modules with claims in them.

ehildenb:
Also, can you add a test where there are 3 claims, 2 that are failing, and 1 that are passing. If you say `kprove --exclude faliing1 --exclude failing2`, the overall command should pass and give back `#Top`.

Baltoli (changes requested):
It would be good to have explicit tests for the `--trusted` and `--exclude` flags as well as `--claims`. At the moment I don't think those code paths will be exercised.

kernel/src/main/java/org/kframework/kprovex/ProofDefinitionBuilder.java
radumereuta: This step applies recursively to all modules imported by the spec module so nothing is left behind.

radumereuta:
@Baltoli  @ehildenb please check if these tests are enough.

Baltoli (approved):

ehildenb (approved):

PR #2838, opened 2022-08-30
Switched to poetry2nix for pyk packaging in nix

goodlyrottenapple:
Due to a transitive dependency on the python [hatchling](https://search.nixos.org/packages?channel=22.05&from=0&size=50&sort=relevance&type=packages&query=hatchling) package, pyk in nix now uses python 3.9. This change also forces us to switch to a newer branch of nixpkgs (version 22.05) which conflicts with the haskell backend. To circumvent the issue, haskell backend is being built using the original version of nixpkgs, whereas pyk and the rest of the k framework are being built with nixpkgs 22.05. This doesn't cause any issues in terms of env clashes, but may result in higher local disk usage by the nix store, as some basic build tools may be duplicated several times for the different versions of nixpkgs. However, this is a well know limitation of nix and this change doesn't make things particularly worse.


tothtamas28:
Is this build failure related: https://github.com/runtimeverification/evm-semantics/runs/8083420640?

goodlyrottenapple:
> Is this build failure related: https://github.com/runtimeverification/evm-semantics/runs/8083420640?

yes

tothtamas28 (approved):

PR #2840, opened 2022-08-30
Add int-simplification builtin

ehildenb:
This is a trimmed down version of KEVM's `INT-SIMPLIFICATION` module. It does not include the Java module that KEVM has, just the Haskell one. Opened this for discussion.

- [ ] Documentation.
- [ ] Tests.


radumereuta:
We can make use of the `comm` attribute we just added.
This can be in `domains.md` in a separate module that is not imported by the INT module. Users can import it when they need it.

Baltoli:
@ehildenb do you still want this upstreamed? It shouldn't be a huge effort to do so based on this old change; if not we can close the PR.

ehildenb:
It would be nice to upstream eventually, but we should pull in latest changes from KEVM, and the tests from there as well.

- INT-SIMPLIFICATION: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
- Tests: https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/functional/int-simplifications-spec.k

PR #2841, opened 2022-08-30
Update simple-untyped to work with the modern backends.

nishantjr:
All tests pass with the k-exercises [updateSimpleUntyped](https://github.com/kframework/k-exercises/pull/31) branch.


k-distribution/pl-tutorial/2_languages/1_simple/1_untyped/tests/threads/threads_04.simple.out
```diff
@@ -1,3 +1,23 @@
...
+    "2 3 "
+  }
 #Or
-  S ==K "3 2 "
+  {
+    S:String
+  #Equals
+    "3 2 "
+  }
+#Or
```
dwightguth, line 12: I expected us to lose some behaviors in some cases, but why have we gained behaviors?

dwightguth (commented):
btw --bound works now

k-distribution/pl-tutorial/2_languages/1_simple/1_untyped/tests/threads/threads_04.simple.out
nishantjr: I'm not sure why the original output had only two behaviours (maybe something to do with needing to label transitions in the Java backend?). Looking at both the semantics of the language, and the comment in the program, it seems that 4 behaviours are to be expected.

nishantjr:
> btw --bound works now

OK, thanks. Looks like there are some additional `==K` vs `#Equals` changes that I somehow missed.

.github/workflows/test-pr.yml
```diff
@@ -94,6 +94,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: kframework/k-exercises
+          ref: updateSimpleUntyped
```
radumereuta: For testing only.
Do not merge with this line!!!

jberthold:
Looked at the macos build failure  - maybe we should configure the extra nix cache for Haskell infrastructure? See [the setup for the haskell-backend](https://github.com/runtimeverification/haskell-backend/blob/cb521b62e9c9993e6bdb7c10a0ca32f99b460619/.github/workflows/release.yml#L21-L24)

.github/workflows/test-pr.yml
```diff
@@ -155,9 +155,8 @@ jobs:
         if: ${{ !startsWith(matrix.os, 'self') }}
         uses: cachix/cachix-action@v10
         with:
-          name: runtimeverification
-          extraPullNames: kore
-          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+          name: k-framework
+          authToken: '${{ secrets.CACHIX_PUBLIC_TOKEN }}'
```
ehildenb, line 160: @jberthold we do set up the `extra_nix_config` here, but it doesn't run on our own CI machines, so I guess it's not actually using taht. We could add back `extraPullNames: kore`, but I'd like to consolidate caches.

dwightguth (approved):

.github/workflows/test-pr.yml
jberthold: What I meant to suggest was to add the `https://cache.iog.io` substituter (IOHK Haskell infra cache), in `"Install Nix"`.
We do this about 8 lines above here but the `test-nix` job in  https://github.com/runtimeverification/k/blob/91f11338d4ccf3626f4e8874e10071e09d5131c6/.github/workflows/test-pr.yml#L192..L198 does not add that cache.
Thought it would be worth a try

PR #2846, opened 2022-08-31
pass --bound flag to llvm backend interpreter

dwightguth:


k-distribution/tests/regression-new/search-bound/1.test.out
```diff
@@ -0,0 +1 @@
+#Bottom
```
ehildenb, line 1: I don't understand why it's returning `#Bottom` here, I would think `<k> a </k>`.

There's no search pattern that is not being reached? I don't see `--pattern` in the additional arguments in the Makefile.

ehildenb: Turns out `--bound` is how many results to return, not how deep to go!

ehildenb (approved):

PR #2849, opened 2022-09-01
Use LLVM backend Python package

Baltoli:
This supersedes the draft changes in #2844 with a more robust solution as suggested by @tothtamas28.

The changes in this PR add the kllvm bindings as a "path" dependency to pyk, and updates the nix flake build accordingly. It also adds a trivial test case to make sure that pyk can call into the bindings correctly.

Because we point to a relative path dependency for the bindings library, we need to perform a few more steps for `pyk`'s `make install` target. [This issue](https://github.com/python-poetry/poetry/issues/5273#issuecomment-1193138069) gives some more context.

* [x] Requires https://github.com/runtimeverification/llvm-backend/pull/553 to go through and the submodule to be updated before this will pass CI.
* [x] Also requires https://github.com/runtimeverification/llvm-backend/pull/554 to have passed the submodule update as well.
* [x] Additionally, we'll need the commit hash for the LLVM backend to point to the correct branch in the flake's lock file.


Baltoli:
https://github.com/python-poetry/poetry/issues/5273#issuecomment-1193138069 when debian packaging

CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 /k-distribution/include/kframework/builtin/* @dwightguth @ehildenb
-/Jenkinsfile @runtimeverification/admin
-/.github/workflows/* @runtimeverification/admin
+/Jenkinsfile @runtimeverification/admin @F-WRunTime
```
dwightguth, line 2: I think it's probably more appropriate to just add Freeman to the admin team than to explicitly hardcode his username here.

dwightguth (commented):
it's not clear to me how these changes would integrate with #2858 . You probably should coordinate in that regard and one of these two should be merged first. Probably that one, not this one.

Baltoli:
Yes, I started working on this PR before I was aware of the work ongoing to pull Pyk out into its own repo. I will be coordinating with @tothtamas28 on the best way to get these changes working under that setup..

PR #2857, opened 2022-09-06
Include parser tests from haskell-backend

tothtamas28:


ehildenb (approved):
Is there a mechanical way to update the golden files? With the makefiles, we could do `make test CHECK=cp`, to copy over the golden/expected files. Is there something similar?

If it's easy to implement, let's add it. Otherwise let's just think about it.

PR #2858, opened 2022-09-07
Drop `pyk`

tothtamas28:
Moved to https://github.com/runtimeverification/pyk.


Jenkinsfile
```diff
@@ -141,7 +137,10 @@ pipeline {
     }
     stage('Build and Package on Ubuntu Jammy') {
       when {
-        branch 'release'
+        anyOf {
+          branch 'release'
+          changeRequest()
+        }
```
radumereuta: Why do we need to test this on every commit after removing pyk?
Same for bellow.

tothtamas28: You don't, I enabled these tests for testing purposes. The corresponding commit (https://github.com/runtimeverification/k/pull/2858/commits/7a8e3c4a2ae83315b9083ced152e30e05e2100c2) will be dropped before merge.

radumereuta:
lgtm, but I will give final approval when this is ready for review.

dwightguth:
Sorry, if I understand this correctly, pyk isn't going to be packaged with kframework anymore? But then how are you going to make sure that the version of pyk that gets installed is compatible with the version of kframework that gets installed?

tothtamas28:
> how are you going to make sure that the version of pyk that gets installed is compatible with the version of kframework that gets installed

`pyk` is tested on CI against a given version of K. When `pyk` is loaded, it could check the K version and emit a warning if there's a mismatch.

To have something more flexible though, my suggestion is to introduce semantic versioning for K so that any version with the same major version that is at least as recent is known to be compatible. And when there is a new feature in K we'd like to use, we simply bump the version we test and check against.

In practice, `pyk` will probably not break that often though: it would only break if K introduced a backward incompatible change in 1) the KAST JSON format 2) the KORE JSON format 3) its CLI. But those changes are relatively rare, so we decided to postpone the decision for adapting a versioning scheme until we see how things work out in practice.

radumereuta (approved):
This should have at least 2 approvals.

ehildenb (approved):

PR #2862, opened 2022-09-07
Emit error from k frontend when kore-exec fails

### LINKED ISSUE #2576: Better distinction between stuck backend and crashed backend
Whenever the backend crashes, the first message the user sees is the final configuration, and only after scrolling up they can see the particular error log. This can be confusing because the user might not be aware that the backend actually crashed, and the backend couldn't continue execution due to the error and not because of some problem the user can fix in the semantics/spec.

SchmErik:
kore-exec output can take up a lot of screen space. This change adds a
error message to claify that the call to kore-exec either resulted in a
stuck state or a crash when kore-exec returns a non-successful status.
closes: #2576


radumereuta:
Do not use the word _stuck_ since it has a different meaning in ML.
stuck = no rules from the semantics can apply
Could say - the backend stopped and can't continue. (see output for more info).

radumereuta (approved):

PR #2863, opened 2022-09-09
Add mention of `-no-haskell-binary` to backends tutorial

Baltoli:
We had a question in the public K channel about this; we should mention the issue in the tutorial for the time being given that it's long-term enough to warrant its own flag.


k-distribution/k-tutorial/1_basic/20_backends/README.md
```diff
@@ -58,6 +58,10 @@ programming language and then compile and execute it. Instead, it is an
 interpreter which reads the generated IR from `kompile` and implements in
 Haskell an interpreter that is capable of interpreting any K definition.
 
+Note that on arm64 macOS (Apple Silicon), there is a known issue that causes
+crashes in the Haskell backend. Pass the additional flag `--no-haskell-binary`
+to `kompile` to resolve this.
+
```
radumereuta:
```suggestion
Note that on arm64 macOS (Apple Silicon), there is a known issue with the `Compact`
library that causes crashes in the Haskell backend. Pass the additional flag
`--no-haskell-binary` to `kompile` to resolve this.

```

radumereuta (approved):
I made a suggestion to mention the `Compact` library so it's easier for users to link the issue to the solution.

Also, should we make the default for the M1 macs be `--no-haskell-binary`? We might need a way to turn it off though which would complicate things a bit.

Baltoli:
I think given that this is (hopefully) a transient bug we can leave the flag optional for now.

Issue #2869, opened 2022-09-10
Downstream Testing and Versioning


ehildenb:
We should be bumping the K minor version when there are API changes, if we are following some semblance of [semantic versioning](https://semver.org/). It's hard for us to measure API changes though, and the tool is changing fast. We also don't really make a distinction between a branch with "only bug-fixes" and a branch with "new features". This is a proposal to change how we do our CI for K repo.

### Current CI approach

- Jenkinsfile: https://github.com/runtimeverification/k/blob/master/Jenkinsfile
- Every PR:
  - Build K on Ubuntu Jammy
  - Install Jammy package
  - Run tests
- `master` job:
  - Bump patch version.
  - Merge master into `release`, push up new `release`.
- `release` job:
  - Build K on _every_ platform.
  - Run smoke tests on every platform.
  - Put together package release [on GitHub](https://github.com/runtimeverification/k/releases).
  - Put together [DockerHub image](https://hub.docker.com/r/runtimeverificationinc/kframework-k) (needed for downstream CI).

### Proposed changes

- [x] Implement a new job (as GitHub runner) which runs on each PR, and makes sure that the patch version is bumped (by checking `git diff origin/master HEAD -- package/version`). This job should automatically bump the patch version if not, and push it up to the PR branch. Similar to how [ksummarize keeps KEVM version in sync](https://github.com/runtimeverification/ksummarize/blob/master/.github/workflows/update-version.yml). https://github.com/runtimeverification/k/pull/2870
- [x] Now that patch versions will be checked and bumped on every PR automatically, we don't need separate `master` and `release` jobs. Move the current K `release` job to `master` phase, and remove `release` branch. https://github.com/runtimeverification/k/pull/2870
- [x] Migrate Pull Request workflows over to GitHub runners. https://github.com/runtimeverification/k/pull/2896
- [x] Migrate master workflows over to github runners.
  - [x] https://github.com/runtimeverification/k/pull/2940
  - [x] https://github.com/runtimeverification/k/pull/2922
  - [x] Migrate Debian Buster package to github workflow.
  - [x] Migrate Arch package to github workflow.
  - [x] https://github.com/runtimeverification/k/pull/2941
- [ ] https://github.com/runtimeverification/k/pull/2960
- [ ] https://github.com/runtimeverification/k/pull/3010
[...]

PR #2870, opened 2022-09-10
Bump patch versions automatically on master

### LINKED ISSUE #1928: In-submodule builds of K report incorrect patch version number
In-submodule builds of K do not by default report the version number correctly.

We should modify `mvn ...` to also call the `./package/version.sh sub` script, to update version numbers throughout the repo.

ehildenb:
Fixes: https://github.com/runtimeverification/k/issues/1928

- Bump patch versions automatically on PRs. It's set up to have the following logic (to allow manually setting version, and automatically bumping it):
  - If the major and minor version identifiers are different than the merge base (eg. bump from 5.3.X => 5.4.0), do not bump the patch number.
  - If the major and minor version identifiers are the same as the merge base, then automatically bump the patch number to be one more than the merge base patch number.
  - Do version string substitution through the repository.
  - If there are any changes in the repo, push them to the current branch.
- Set up concurrency groups on all github workflows, so that old jobs are appropriately cancelled (otherwise we run into GitHub public runner limit of 15 concurrent at a time pretty quickly).
- Removes the `push: ...` specifier for `update.yml` job. Should this remain @goodlyrottenapple? It seems unnecessary.

**TODO**:

- [ ] Enable this new status check as required on CI.


ehildenb:
It's working here 
image

ehildenb:
Note that I had to make this change to our organization secret: 
image

I think it should be OK, because secrets are still not shared with PRs run from forks.

ehildenb:
Appears to be working properly: 
image

ehildenb:
I've realized this won't play very nice in the multi-PR scenario, will push up a fix.

ehildenb:
I also had to bump the minor version number to get the job to trigger correctly this time around (since it compares against `origin/master` major and minor version numbers to decide if it needs to bump the patch number).

radumereuta:
If I understand this correctly, on each PR, I will submit some changes, then:
- the CI is going to change the version in a few locations in the repo
- add another commit to origin/branch
- origin/branch will now be ahead of local/branch, and I have to pull the changes (annoying)
- my working branch will contain extra modified files (which I have to know about)
- every single new PR from now on is going to contain the version bump (that's extra work for reviewers)

ehildenb:
Your understanding is not quite correct. CI will do this only the minimum number of times needed (not every time you push). That will be:

- If `master` branch is updated, and then this branch is pushed to, CI will now bump the version here to the next version after master branch. You will have to `git pull`.
- When you first open the PR, CI will automatically push the new version to your branch (within 30s).

It's _always_ safe to resolve the merge conflict with `git checkout --ours -- package/version` or `git checkout --theirs -- package/version`, CI will always make sure the version identifier is set correctly afterwards either way. Also, it _should_ always be possible to do `git fetch origin ; git rebase origin/MYBRANCH` with _no_ merge conflicts, which is a pretty painless process.

The wins are that we have an accurate version number no matter what, and a simplified release process, and better ability to do semantic versioning.

radumereuta:
Are you sure you still want to update to `5.4.0`?
It will make the changelog a bit weird.

ehildenb:
It's the easiest way to make sure we have version number continuity. I could also choose 5.3.202, if we know this one will be merged next.

ehildenb:
But already last night when I was trying, it would have been 5.3.199, so who knows?

radumereuta:
You're right. Leave it as `v5.4`. I will add a note in the changelog.

F-WRunTime (approved):

radumereuta:
Are you sure we're still running all workflows we intend?
I'm counting 10 GH workflows + 1 Jenkins before
and 4 GH workflows + 1 Jenkins on this branch.

ehildenb:
Where did you get 10 GH workflows? You can see in the PR that I didn't change any of the other workflows other than adding the concurrency throttle.

F-WRunTime:
I'm just counting 4 GH workflows and 1 Jenkins

radumereuta:
Look at other PRs. #2873 which doesn't have these changes has 8+1 (sorry, not 10)
image

ehildenb:
I see 2 there triggered by `push` events, which this PR took care of.

ehildenb:
I wonder why these checks are not required. I think I've identified the problems though:

Tests run on the other PR:

```
Test / Nix (ubuntu-latest) (pull_request) Successful in 8m
Update / Nix: Maven (pull_request) Successful in 7m
Test / Nix (macos-11) (pull_request) Successful in 20m
Update / Nix flake sumbodule sync (pull_request) Successful in 52s
continuous-integration/jenkins/pr-head — This commit looks good
 
Update / Nix: Maven (push) Successful in 9m
Update / Nix flake sumbodule sync (push) Successful in 43s


Test / Nix flake (ubuntu-latest) (pull_request) Successful in 5m
Test / Nix flake (macos-11) (pull_request) Successful in 11m
```

Tests run on this PR (before last commit):

```
Test / Nix (ubuntu-latest) (pull_request) Successful in 14m
Update / Nix: Maven (pull_request) Successful in 7m
Test / Nix (macos-11) (pull_request) Successful in 20m
Update / Nix flake sumbodule sync (pull_request) Successful in 42s
continuous-integration/jenkins/pr-head — This commit looks good
```

The tests in the first block on the other PR are run here. The tests in the second block are because I disabled `push` event handling on this PR for `Nix: Maven` and `Nix: submodule sync`. For some reason we were running this job doubly for pushes and pull-request events. The third block I suspect is because there was a stage name clash, and github's interface is not handling it correctly. Indeed, the output now contains all 7 entries (6 github runners, and 1 Jenkins):

```
Test / Nix (ubuntu-latest) (pull_request) In progress — This check has started...
Test Nix Flake / Nix flake (ubuntu-latest) (pull_request) In progress — This check has started...
Update / Nix: Maven (pull_request) In progress — This check has started...
Test / Nix (macos-11) (pull_request) In progress — This check has started...
Test Nix Flake / Nix flake (macos-11) (pull_request) In progress — This check has started...
Update / Nix flake sumbodule sync (pull_request) Successful in 42s
continuous-integration/jenkins/pr-head Pending — This commit is being built
```

ehildenb:
We generally expect every test to pass here right? If so, we should enable this list of tests as required.

radumereuta:
> We generally expect every test to pass here right? If so, we should enable this list of tests as required.

We might need to doublecheck with @goodlyrottenapple on this.

.github/workflows/update.yml
```diff
@@ -1,9 +1,10 @@
 name: 'Update'
 on:
   pull_request:
-  push:
-    branches-ignore:
```
dwightguth, line 5: Why is this changed?

dwightguth (approved):

.github/workflows/update.yml
ehildenb: I'm not sure why it was like this in the first place, but basically it resulted in us running CI double on each pull request, once for the branch trigger and once for the pull_request trigger. @goodlyrottenapple may know why we were doing this before, I suspect it was so that someone could debug this workflow in the past without opening a PR, but I don't know.

Either way, should not need to run it twice for every branch.

ehildenb:
I'm going to merge and then require these extra status checks.

PR #2873, opened 2022-09-12
Install kllvm when K is installed

Baltoli:
This PR properly installs the KLLVM bindings library generated by the LLVM backend when K is packaged and installed.


src/main/scripts/test-in-container
```diff
@@ -12,3 +12,6 @@ cd ~
 echo "module TEST imports BOOL endmodule" > test.k
 kompile test.k --backend haskell
 rm -rf test-kompiled
+
+# Make sure that the KLLVM bindings have been installed properly.
+python3 -c 'import kllvm'
```
tothtamas28: Consider including some very simple logic that exercises the native code through the bindings.

tothtamas28 (approved):

PR #2876, opened 2022-09-13
Bump patch version automatically on master branch

### LINKED ISSUE #1928: In-submodule builds of K report incorrect patch version number

ehildenb:
Fixes: https://github.com/runtimeverification/k/issues/1928

This is re-opening of #2870 but with an issue with GitHub runner cloning fixed.

The key issue was that by default the github runner does a shallow clone of depth 1, but this job tries to do a comparison at depth 2 to see if the version has been changed. With the shallow clone at depth 1, git was reporting back that the version had always been changed. Adding `fetch-depth: 0` sets it to arbitrary depth.

Also made it a bit easier to switch which branch this job is running on, by using `GITHUB_REF_NAME` in various places instead of `master` directly. I used branch `foo` to test these changes, and confirm that it works.


ehildenb:
@dwightguth I've tested this (on a branch named `foo`), and it seems to behave correctly now.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,32 @@
...
+  version-bump:
+    name: 'Version Bump'
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Check out code'
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.JENKINS_GITHUB_PAT }}
+          ref: ${{ github.ref_name }}
+          fetch-depth: 0
```
ehildenb, line 21:
```suggestion
          # fetch-depth 0 means deep clone the repo
          fetch-depth: 0
```

dwightguth (approved):
Approved. I don't entirely understand the details of why, but somehow shallow clones are more expensive for the server than a deep clone, so it's not at all a bad thing that we're doing a deep clone here.

PR #2879, opened 2022-09-13
Hotfix for `/usr/local` packaging on Jammy

Baltoli:
We've experienced this problem before; the solution is to move things out of `/usr/local' into `/usr`

### DIFF
--- a/package/debian/rules.jammy
+++ b/package/debian/rules.jammy
@@ -32,6 +32,7 @@ override_dh_auto_build:
 
 override_dh_auto_install:
 	package/package
+	mv $(DESTDIR)$(PREFIX)/local/lib/$(PYTHON_VERSION) $(DESTDIR)$(PREFIX)/lib/$(PYTHON_VERSION)
 	@mkdir -p $(DESTDIR)$(PREFIX)/lib/$(PYTHON_DEB_VERSION)/dist-packages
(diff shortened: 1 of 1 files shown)


radumereuta (approved):

package/debian/rules.jammy
```diff
@@ -32,6 +32,7 @@ override_dh_auto_build:
 
 override_dh_auto_install:
 	package/package
+	mv $(DESTDIR)$(PREFIX)/local/lib/$(PYTHON_VERSION) $(DESTDIR)$(PREFIX)/lib/$(PYTHON_VERSION)
 	@mkdir -p $(DESTDIR)$(PREFIX)/lib/$(PYTHON_DEB_VERSION)/dist-packages
 	ln --symbolic --relative $(DESTDIR)$(PREFIX)/lib/$(PYTHON_VERSION)/site-packages/kllvm $(DESTDIR)$(PREFIX)/lib/$(PYTHON_DEB_VERSION)/dist-packages/kllvm
```
ehildenb, line 37: Should we be following more what the 3rd line here does? For debian packages, the python versioning scheme is different. We could also just change the 3rd line to linke to `/local/lib` instead of `/lib`.

Baltoli: I believe the issue is that you can't have anything in `/usr/local`, so you can't link the debian python directory to the first location.

Issue #2882, opened 2022-09-14
test suite emits warnings on macro labels placed on rules


SchmErik:
### Which command

kompile

### What K Version?

v5.3.195-5-gf2be328531

### Operating System

Linux

### K Definitions (If Possible)

all tests run by CI under k-distribution/tests and k-distribution/pl-tutorial

### Steps to Reproduce

run `make -j8 --output-sync=recurse` from `k/k-distribution/tests/regression-new/`

### Expected Results

No test should result in warnings that indicate macros are used on rules.

PR #2883, opened 2022-09-15
regression-new: move [macro] attribute from rules to productions

SchmErik:
K now supports macros on productions and using the `[macro]` label on rules is highly discouraged. This PR moves all [macro] attributes from rules to productions in the regression-new folder as well as changes that rewrite <k> cells more explicitly. Subsequent commits will address the macro labels in the PL tutorial


radumereuta (approved):
Tested and the output looks a lot better.

radumereuta:
Partial fix for: https://github.com/runtimeverification/k/issues/2882

PR #2893, opened 2022-09-16
Bump patch version on PRs

ehildenb:

This switches to bumping the patch version on PRs. I propose we try this approach, to see if it's too disruptive to people's workflows, given the failures to bump `master` version directly.


ehildenb:
Here is an example where a PR was opened, and then another PR was merged, and this one had to bump the version number again. Because there was only 1 PR merged in the interim, it actually results in no merge conflicts, and deals with it automatically: https://github.com/runtimeverification/pyk/pull/19

So if there are 0 or 1 PRs merged between when you open your PR, and when you press "Update branch", GitHub will be able to merge automatically. If there are more, then maybe not, but you can still always do either `git checkout --[ours|theirs]` to resolve the conflict, or you can do a rebase, or you can use github's merge editor directly, which is also not bad to use on this 1-line conflict.

radumereuta:
Question: from now on, every time I have to do a code review on K, is it going to have these extra 5 files with the version bump?
That sounds like a cognitive burden we could deal without.

ehildenb:
Please try it for a bit. Currently we don't have a working K release, and you may find it's less burden than you expect @radumereuta

radumereuta (approved):
Ok, let's try this until we have something better.

.github/workflows/bump-version.yml
```diff
@@ -1,6 +1,6 @@
 name: 'Bump Version'
 on:
-  push:
+  pull_request:
     branches:
```
dwightguth, line 4: are we sure that this is the target branch and not the topic branch?

ehildenb: Yes: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request

dwightguth (approved):

PR #2896, opened 2022-09-20
Add GitHub workflow for PR job

ehildenb:
Blocked on: https://github.com/runtimeverification/k/pull/2893

This adds a GitHub workflow for PR jobs, which focuses on:

- [x] Runs the K Test CI.
- [x] Test building Ubuntu package, and run smoke test with it.
- [x] Disable Jenkinsfile tests.
- [x] Adds dependencies between test stages, so we're not using runners before we know we need them.
- [x] Factors out and minimizes the docker image used for testing PRs.
- [x] Factors out action for setting up docker image for CI.
- [x] Enable new required checks.


ehildenb:
@goodlyrottenapple here is what the new workflow looks like:

image

What's the relationship between the various "Flake" jobs an the non-Flake nix jobs? Are they somehow redundant? We could cut down a lot on overall time if we can remove all the non-Flake jobs.

.github/workflows/test-pr.yml
```diff
@@ -0,0 +1,258 @@
...
+        uses: cachix/cachix-action@v10
+        with:
+          name: runtimeverification
+          extraPullNames: kore
+          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+
+      - name: 'Build K Framework'
+        run: GC_DONT_GC=1 nix build .
+
+  test-nix:
```
goodlyrottenapple: We could probably try to phase this out in another PR. Most of the steps in `test-nix` (should) have an analogue in `build-nix-flake`

ehildenb:
@Baltoli I'm getting this error:

image

Which is saying "Optimization flag '-ffat-lto-objects' is not supported". Any quick ideas about how I could investigate it? It's for the new `test-ubuntu-package` workflow on this PR, when `dpkg-buildpackage` is being called to build the Jammy package inside the docker container at `.github/workflows/Dockerfile`

I get the same behavior with and without the last commit 5903a21, which removes the `-lto` flag from debian rules file.

dwightguth:
It would be helpful when reviewing this if it were clearer how to break the diff into components. For example, there seems to be some components that appear to consist of simply moving files around and renaming things, and other components that actually edit the files that result from these changes. I will be much better equipped to review if you can provide me with commit ranges that broadly encapsulate these two steps, because it is hard to review both types of changes in a single diff.

ehildenb:
What it looks like: 
image

ehildenb:
The first step is compacting the PR workflows into one commonly triggered file: `git diff b309ade958 6bbacc52d3`.

Then it's adjusting the workflow, setting up dependencies between jobs, and adding the rest of the PR testing functionality: `git diff 6bbacc52d3 e7a6859bcf`.

.github/workflows/test-pr.yml
ehildenb: Yes, I'd like to optimize the workflows to reduce overall CI time. There seems to be a lot of redundancy in the Nix steps, we can deal with that in another PR though.

dwightguth:
This is not true; I looked at those commits and it was clear that some changes were made in the commit range that you label "compacting" which are not simply moving code around.

I'm very serious about this; refactoring history so that changes like this are easier to review is an important step because without it, I very likely will miss things as part of the code review.

Please go ahead and refactor the history into something that respects the concept that moves and modifications are in separate commit ranges, and then I will review in more depth. As it currently stands, it is simply not practical to assume that I can review this PR without potentially missing important details.

ehildenb:
How it looks: 
image

ehildenb:
A more complete view:

image

ehildenb:
Now it looks like this:

image

.github/actions/with-docker/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      docker build . --file .github/workflows/Dockerfile.z3          --tag z3:${Z3_VERSION}
+      docker build . --file .github/workflows/Dockerfile.stack-deps  --tag stack:${K_VERSION}
+      docker build . --file .github/workflows/Dockerfile.maven-cache --tag maven:${K_VERSION}
+
+      docker build . --file .github/workflows/Dockerfile \
+          --tag runtimeverification/${TAG_NAME}          \
+          --build-arg Z3_VERSION=${Z3_VERSION}           \
+          --build-arg K_VERSION=${K_VERSION}             \
+          --build-arg LLVM_VERSION=14                    \
+          --build-arg USER_ID=$(id -u)
```
ehildenb:
```suggestion
          --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g)
```

ehildenb:
How it looks:

image

.github/workflows/test-pr.yml
```diff
@@ -140,12 +139,95 @@ jobs:
...
+        with:
+          submodules: recursive
+      - name: 'Set up Docker'
+        uses: ./.github/actions/with-docker
+        with:
+          tag: k-ci
+      - name: 'Build K'
+        run: docker exec -t k-ci /bin/bash -c 'mvn package -DskipTests'
+      - name: 'Unit Tests'
+        run: docker exec -t k-ci /bin/bash -c 'mvn --batch-mode verify -U'
```
ehildenb:
```suggestion
      - name: 'Build and Test K'
        run: docker exec -t k-ci /bin/bash -c 'mvn verify --batch-mode -U'
```

dwightguth (approved):
Looks good

PR #2898, opened 2022-09-20
Use 20 threads when testing

radumereuta:
similar to the available Jenkins executors


radumereuta:
Doesn't appear to run faster.
We also need to check how many threads the Java thread pool is using to make sure we're not creating extra contention there.

ehildenb:
`nproc` is correctly reporting thread counts on GitHub runners, which we [already are switching to](https://github.com/runtimeverification/k/pull/2896).

image

PR #2903, opened 2022-09-23
Compact workflows

ehildenb:
Moves all workflows into one file, part of https://github.com/runtimeverification/k/pull/2896 (the first three commits are the same).

This makes sense, because these are all workflows to be run with the same trigger (pull requests into master).

As a general rule of thumb for workflows, it seems that anything which should run with the same trigger should be grouped into the same workflow. In addition, you cannot add dependencies _between_ jobs in different workflows files, so now we can make a dependency chain between all these workflows to make most efficient use of our CI machines (no need to kick off jobs running tests if the version bumper is going to push another commit anyway). That's what https://github.com/runtimeverification/k/pull/2896 does.


ehildenb:
This is what we get: 
image

dwightguth (dismissed):

dwightguth (approved):

PR #2916, opened 2022-09-27
`kup` install tool + nix bash script

### LINKED ISSUE #2894: `kup` command for managing Nix K versions
Many of our users complain about our installation process, and we historically have trouble tools with different versions of K and such.

One installation process that everyone seems to like is the `curl SOME_URL | bash -` approach, especially if it does version management for the user. The Foundry tool, for example, has you [install `foundryup`](https://github.com/foundry-rs/foundry#installation), then you just run `foundryup` when you need to update, and automatically get the newest version, all installed in the users home directory. Similarly, the Rust compiler `rustc` has [`rustup`](https://rust-lang.github.io/rustup/) for managing compiler toolchains. [dapp.tools](https://github.com/dapphub/dapptools#installation) does similarly, they have a user run `curl https://dapp.tools/install | sh`, which does (i) install Nix, (ii) install the nix expression for that package at that version. Then we just have to rely on having Nix derivations for everything, which may be easier long-term than having platform specific binaries produced on CI (though we probably _do_ still want to populate some public Nix caches so that users can have super fast installation).

Here is `rustup help`:

```
rustup 1.24.3 (ce5817a94 2021-05-31)
The Rust toolchain installer

USAGE:
    rustup [FLAGS] [+toolchain] <SUBCOMMAND>

FLAGS:
    -v, --verbose    Enable verbose output
    -q, --quiet      Disable progress output
    -h, --help       Prints help information
    -V, --version    Prints version information

ARGS:
    <+toolchain>    release channel (e.g. +stable) or custom toolchain to set override

SUBCOMMANDS:
    show           Show the active and installed toolchains or profiles
    update         Update Rust toolchains and rustup
    check          Check for updates to Rust toolchains and rustup
    default        Set the default toolchain
    toolchain      Modify or query the installed toolchains
    target         Modify a toolchain's supported targets
    component      Modify a toolchain's installed components
    override       Modify directory toolchain overrides
    run            Run a command with an environment configured for a given toolchain
    which          Display which binary will be run for a given command
    doc            Open the documentation for the current toolchain
    man            View the man page for a given command
    self           Modify the rustup installation
    set            Alter rustup settings
    completions    Generate tab-completion scripts for your shell
    help           Prints this message or the help of the given subcommand(s)
```

### `kup` command

Here is what I would propose as the help message of this command:

```
kup 5.4.3 (ce5817a94 2021-05-31)
The K Framework installer

USAGE:
    kup [FLAGS] <SUBCOMMAND>

FLAGS:
    -v, --verbose    Enable verbose output
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    setup          Install the kup command locally
    list           Show the active and installed K versions
    list --all     Show all versions available (including remote ones)
    install        Download and install the stated version (do not switch default)
    update         Update K Framework default version and kup
    default        Set the default K version
    run            Run a command with an environment configured for a given K version
    shell          Give a shell with an environment configured for a given K version
    help           Prints this message or the help of the given subcommand(s)
```

And we should make it available at https://kframework.org/kup, via `curl https://kframework.org/kup | bash -s setup`.

We should support one special version name called `latest`, which saying `kup default latest` makes the installation track that version. If the installation is tracking the `latest` version, then `kup update` should also affect the default version (by updating to the latest available).

We should aim to not have to do any global installs or `sudo` commands other than what's required to get Nix set up on their system.

goodlyrottenapple:
Fixes: https://github.com/runtimeverification/k/issues/2894

This PR adds the initial `kup` tool (#2894) plus an installation script which provide the functionality to:

* install, remove and update all k framework semantics which have a nix flake expression (currently k itself and evm-semantics)
* drop into a temporary shell for the above

See below for a sample interaction with the tool:
image


package/kup/Makefile
```diff
@@ -0,0 +1,57 @@
+.PHONY: default all clean build install          \
```
tothtamas28: Please regenerate this project with [cruft](https://cruft.github.io/cruft/):

```
cruft create https://github.com/runtimeverification/python-project-template.git
```

This way, we can easily apply template changes.

goodlyrottenapple:
@ehildenb before this tool is useful for our users, we need to set up the caching for all the listed tools, so kup doesnt end up compiling stuff on the user's machine. I will open a separate PR for this in the K repo.

package/kup/src/kup/__main__.py
```diff
@@ -0,0 +1,185 @@
+import os, sys
```
tothtamas28: Please make sure `make` succeeds.

tothtamas28: Formatting issues can be fixed with `make format`.

package/kup/Makefile
tothtamas28: Consider moving the project to the repo root.

goodlyrottenapple: i used cookiecutter. is that not used any more? if not, please update the repo.

tothtamas28: `cruft` uses `cookiecutter`, it just adds some versioning capabilities.

goodlyrottenapple: i see, still, the readme needs updating in the `python-project-template` repo

ehildenb:
Closing and re-opening to trigger new workflows.

ehildenb:
It looks good to me, and seems to work as intended. @shd101wyy can you advise us on how to get this published on the website at https;//kframework.org/install? Specifically, the `package/nix/install` script?

tothtamas28 (approved):

PR #2920, opened 2022-09-28
Add rules for symbolic set simplification

jberthold:
First step to add simplification rules for `Set` simplification at `k` level.

This is a new version of #2689 which tries to avoid performance penalties from checking non-membership or non-equality while matching (non-deterministically) against set elements.

Fixes https://github.com/runtimeverification/haskell-backend/issues/3124

As a next step, the internal built-in set simplification will be removed from the backend to address https://github.com/runtimeverification/haskell-backend/issues/2928 , enabling the tests which are currently in directory `non-termination`. Some other tests are considered as currently failing since the simplifications are not powerful enough.


k-distribution/tests/regression-new/set-symbolic-tests/non-terminating/diff-21-spec.k
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 K Team. All Rights Reserved.
+// generated by gen-tests.sh
+
+requires "set-tests.k"
+
+module DIFF-21-SPEC
+    imports SET-TESTS
+
+    claim <k> diff ( Z:MyId in ( (S SetItem(X:MyId)) -Set SetItem(Y:MyId) ) )     => . </k> requires Y  ==K X
```
jberthold, line 9: All non-terminating tests are also looping when running against a version without any new set simplification rules.
The problematic terms that make the tests loop are `S SetItem(X) -Set SetItem(Y)` and `SetItem(Y) |Set (S SetItem(X))`. I added this information to the [relevant ticket  in `haskell-backend`](https://github.com/runtimeverification/haskell-backend/issues/2928)

k-distribution/tests/regression-new/set-symbolic-tests/failing/diff-11-spec.k.out.actual
```diff
@@ -0,0 +1,9 @@
+  <k>
+    diffResult ( SetItem ( Y:MyId ) -Set SetItem ( X:MyId ) ) ~> .
```
jberthold: This is very bizarre - somehow we are simplifying `SetItem(Y) -Set (S SetItem(X))` to `(SetItem(Y) -Set SetItem(X)` (removing the `S`), but that is *wrong*, `Y` could be present in `S`. 
_This happens regardless of whether the new simplification rules are imported or not!_
Same for `diff-12` and `diff-13` tests (also `diff-15`, not included here since not relevant with the reduced rules)

k-distribution/tests/regression-new/set-symbolic-tests/failing/intersect-37-spec.k.out.actual
```diff
@@ -0,0 +1,10 @@
+  <k>
+    intersectResult ( false ) ~> .
+  </k>
+#And
+  {
+    false
+  #Equals
+    X:MyId in SetItem ( x )
```
jberthold: Absolutely no idea what happens here! The prover appears to be inventing the `X /= x` constraint, maybe by way of confusing `x` and `y`?
Without the new rules, the result is as expected:
```
  #Not ( {
    X
  #Equals
    y
  } )
#And
  <k>
    intersectResult ( X:MyId in intersectSet ( SetItem ( X:MyId )
    SetItem ( y ) , SetItem ( x ) ) ) ~> .
  </k>
```

k-distribution/tests/regression-new/set-symbolic-tests/failing/union-5-spec.k.out.actual
```diff
@@ -0,0 +1,22 @@
+  #Ceil ( S
+  ( SetItem ( Y:MyId ) -Set SetItem ( X:MyId ) ) )
+#And
+  <k>
+    unionResult ( S
+    SetItem ( X:MyId )
+    ( SetItem ( Y:MyId ) -Set SetItem ( X:MyId ) ) ) ~> .
+  </k>
```
jberthold: This result is partly due to rewriting `|Set` into a `-Set` , but it is not correct.
We started with `(S SetItem(X)) |Set SetItem(Y)` and  `Y` could be in `S`, making the resulting term undefined.
Of course the `#Ceil` condition will (cum grano salis) prevent that but... it is a valid case that is dropped by the simplification.
Again, this happens regardless of whether the new rules are imported or not.

Many (removed) `union` tests have this same problem.

jberthold:
https://github.com/runtimeverification/k/compare/2612-no-union-rewriting has an alternative where I removed the rule `S1 |Set S2 => S1 (S2 -Set S1)`. It produces a few confusing results so keeping it on an extra branch.

ana-pantilie:
@jberthold are these results on master Haskell backend or on the feature branch which removes builtin simplification?

If these rules behave well on the feature branch, then the failing/looping tests should be moved to the `failing` directory. Make sure that the failing directory contains the correct, expected, result of the test not the broken one. We can then review the PR and merge it, and afterwards merge the backend PR and re-enable the failing tests.

jberthold:
> are these results on master Haskell backend or on the feature branch

This is all using the Haskell backend from master at the moment. Will check the behaviour with removed internal set simplifications.

k-distribution/tests/regression-new/set-symbolic-tests/failing/intersect-37-spec.k.out.actual
jberthold: As it turned out, the rule `rule (S SetItem(X)) |Set S  => S SetItem(X)` is causing this confusion (or at least removing it makes this test behave again). I don't quite understand why.

ana-pantilie: Please open an issue for this.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -550,7 +550,7 @@ elements present in either set.
 
 ```k
   syntax Set ::= Set "|Set" Set            [left, function, functional, hook(SET.union)]
-  rule S1:Set |Set S2:Set => S1 (S2 -Set S1)
+  // FIXME rule S1:Set |Set S2:Set => S1 (S2 -Set S1)
```
jberthold: I removed this rule to make the added `|Set` simplification rules kick in. A few `union` tests would not produce the desired output otherwise. OK to delete it?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,93 @@ that two different sets will return the same element, even if they are similar.
...
+                             ensures notBool (X in S) [simplification]
+
+  // intersectSet simplifications
+  rule intersectSet(.Set, _   ) => .Set    [simplification]
+  rule intersectSet( _  , .Set) => .Set    [simplification]
+  rule intersectSet( S  , S   ) =>  S      [simplification]
+
+  rule intersectSet( S SetItem(X), SetItem(X))     => SetItem(X)
+                                                        ensures notBool (X in S)      [simplification]
+  // rule intersectSet( S SetItem(X) , S)             => S ensures notBool (X in S)      [simplification]
```
jberthold: Adding this rule produces wrong results (test `intersect-37-spec.k` goes wrong). This is tracked [in a separate ticket](https://github.com/runtimeverification/haskell-backend/issues/3294) with a simple example.

k-distribution/tests/regression-new/set-symbolic-tests/failing/union-18-spec.k
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 K Team. All Rights Reserved.
+// generated by gen-tests.sh
+
+requires "set-tests.k"
+
+module UNION-18-SPEC
+    imports SET-TESTS
+
+    claim <k> union ( S                    |Set (S  SetItem(X:MyId)) ) => . </k>
```
jberthold, line 9: There is an explicit simplification to rewrite `S |Set (S {X})` to `S {X}` but it somehow isn't applied.

k-distribution/tests/regression-new/set-symbolic-tests/non-terminating/diff-21-spec.k
jberthold: Confirmed that these tests in `non-terminating` work when using a HS backend without internal set simplification (from PR https://github.com/runtimeverification/haskell-backend/pull/3062)

k-distribution/tests/regression-new/set-symbolic-tests/unsimplified/diff-12-spec.k
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022 K Team. All Rights Reserved.
+// generated by gen-tests.sh
+
+requires "set-tests.k"
+
+module DIFF-12-SPEC
+    imports SET-TESTS
+
+    claim <k> diff ( SetItem(Y:MyId)      -Set (S SetItem(X:MyId)) )  => . </k> requires Y =/=K X
```
jberthold, line 9: The tests in `unsimplified` were kept  from the prior PR #2689 but are not expected to work because this PR does not add the relevant simplification rules (because of their performance/complexity impact).

k-distribution/tests/regression-new/set-symbolic-tests/failing/intersect-37-spec.k.out.actual
jberthold: https://github.com/runtimeverification/haskell-backend/issues/3294

k-distribution/include/kframework/builtin/domains.md
dwightguth: If you're removing this you need to add an implementation of this hook to the Haskell backend.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,93 @@ that two different sets will return the same element, even if they are similar.
...
+  rule intersectSet( S1 SetItem(X), S2 SetItem(X)) => intersectSet(S1, S2) SetItem(X)
+                                                        ensures notBool (X in S1)
+                                                        andBool notBool (X in S2)     [simplification]
+
+  // membership simplifications
+  rule E in .Set           => false requires #Ceil(E)       [simplification]
+  rule E in SetItem(E)     => true  requires #Ceil(E)       [simplification]
```
dwightguth: You do not need the side condition here because E is an element variable and will not match \bottom regardless.

jberthold: > If you're removing this you need to add an implementation of this hook to the Haskell backend.

Good point, thank you. Looking back at #2689 I find that I can instead add `[concrete]` to avoid the rule in symbolic simplifications.

Removing the redundant `#Ceil`.

dwightguth (approved):
Looks good to me but I would like Ana to approve also.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,94 @@ that two different sets will return the same element, even if they are similar.
...
+  rule .Set |Set S    => S    [simplification]
+  rule S    |Set S    => S    [simplification]
+
+  rule (S SetItem(X)) |Set SetItem(X) => S SetItem(X)
+                             ensures notBool (X in S) [simplification]
+//  rule (S SetItem(X)) |Set S          => S SetItem(X)
+//                             ensures notBool (X in S) [simplification]
+  rule SetItem(X) |Set (S SetItem(X)) => S SetItem(X)
+                             ensures notBool (X in S) [simplification]
+  rule S          |Set (S SetItem(X)) => S SetItem(X)
```
ana-pantilie: Why was this rule kept, while its commutative equivalent was removed?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,94 @@ that two different sets will return the same element, even if they are similar.
...
+  // |Set simplifications
+  rule S    |Set .Set => S    [simplification]
+  rule .Set |Set S    => S    [simplification]
+  rule S    |Set S    => S    [simplification]
+
+  rule (S SetItem(X)) |Set SetItem(X) => S SetItem(X)
+                             ensures notBool (X in S) [simplification]
+//  rule (S SetItem(X)) |Set S          => S SetItem(X)
+//                             ensures notBool (X in S) [simplification]
+  rule SetItem(X) |Set (S SetItem(X)) => S SetItem(X)
```
ana-pantilie: Maybe we can avoid adding this rule explicitly.

@radumereuta was the commutative attribute added? Does it work for simplification rules as well?

Since I want to minimize any extra work needed to be done on this PR, this change, if possible, should be done in a subsequent PR.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -549,8 +549,8 @@ number and thus the time is effectively linear. The union consists of all the
 elements present in either set.
 
 ```k
-  syntax Set ::= Set "|Set" Set            [left, function, functional, hook(SET.union)]
-  rule S1:Set |Set S2:Set => S1 (S2 -Set S1)
+  syntax Set ::= Set "|Set" Set              [left, function, functional, hook(SET.union)]
+  rule S1:Set |Set S2:Set => S1 (S2 -Set S1) [concrete]
```
ana-pantilie, line 559: Sorry if I forgot any details you specified when we discussed this.

Since this rule depends on set difference, I would assume it needs to be concrete because having both the Haskell backend implementation and the newly added simplification rules for `-Set` somehow break this rule.

After modifying the backend, maybe we should take another look at this one.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,94 @@ that two different sets will return the same element, even if they are similar.
...
+  // rule intersectSet( S SetItem(X) , S)             => S ensures notBool (X in S)      [simplification]
+  rule intersectSet( S1 SetItem(X), S2 SetItem(X)) => intersectSet(S1, S2) SetItem(X)
+                                                        ensures notBool (X in S1)
+                                                        andBool notBool (X in S2)     [simplification]
+
+  // membership simplifications
+  rule _E in .Set           => false   [simplification]
+  rule E  in SetItem(E)     => true    [simplification]
+  rule E  in (S SetItem(E)) => true
+              ensures notBool (E in S) [simplification]
```
ana-pantilie: The first case is subsumed in the second one.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -612,6 +612,94 @@ that two different sets will return the same element, even if they are similar.
...
+
+// These two rules would be sound but impose a giant overhead on `in` evaluation:
+  // rule E1 in (S SetItem(E2)) => true requires E1 in S
+  //                                 ensures notBool (E2 in S) [simplification]
+  // rule E1 in (S SetItem(E2)) => E1 in S requires E1 =/=K E2
+  //                                 ensures notBool (E2 in S) [simplification]
```
ana-pantilie, line 694: Again, apologies if this has been discussed before.

`rule E1 in (S SetItem(E2)) => E1 in S requires E1 =/= E2 ensures notBool (E2 in S)` (the second rule) should be enough here.

Why is this an overhead on `in` evaluation? There's a specific matching case in the backend for rules like this, and most of the simplification rules on collections are written in this style.

k-distribution/tests/regression-new/set-symbolic-tests/diff-17-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    diffResult ( ( S1
```
ana-pantilie, line 2: What should this test be testing?

k-distribution/tests/regression-new/set-symbolic-tests/diff-19-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    diffResult ( S1 -Set ( S2
+    SetItem ( X:MyId ) ) ) ~> .
```
ana-pantilie, line 3: Same question as above.

k-distribution/tests/regression-new/set-symbolic-tests/inset-11-spec.k.out
```diff
@@ -0,0 +1,19 @@
+  <k>
+    inSetResult ( Y:MyId in SET
```
ana-pantilie, line 2: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/inset-4-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    inSetResult ( Y:MyId in SET
```
ana-pantilie, line 2: Same q as above

k-distribution/tests/regression-new/set-symbolic-tests/intersect-11-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    intersectResult ( intersectSet ( SetItem ( Y:MyId ) , S
```
ana-pantilie, line 2: Same q

k-distribution/tests/regression-new/set-symbolic-tests/intersect-17-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    intersectResult ( intersectSet ( S1
```
ana-pantilie, line 2: Again :)

k-distribution/tests/regression-new/set-symbolic-tests/intersect-19-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    intersectResult ( intersectSet ( S1 , S2
```
ana-pantilie, line 2: Again

k-distribution/tests/regression-new/set-symbolic-tests/intersect-23-spec.k.out
```diff
@@ -0,0 +1,16 @@
+  #Not ( {
+    Y
+  #Equals
+    Z
+  } )
+#And
+  <k>
+    intersectResult ( Z:MyId in SetItem ( Y:MyId ) ) ~> .
```
ana-pantilie, line 8: Shouldn't this get simplified?

k-distribution/tests/regression-new/set-symbolic-tests/intersect-31-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
```
ana-pantilie, line 1: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/intersect-37-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    intersectResult ( X:MyId in intersectSet ( SetItem ( X:MyId )
```
ana-pantilie, line 2: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/intersect-5-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    intersectResult ( intersectSet ( S
```
ana-pantilie, line 2: Same q

k-distribution/tests/regression-new/set-symbolic-tests/union-11-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( SetItem ( Y:MyId ) |Set ( S
```
ana-pantilie, line 2: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/union-17-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( ( S1
```
ana-pantilie, line 2: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/union-19-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( S1 |Set ( S2
```
ana-pantilie, line 2: Same

k-distribution/tests/regression-new/set-symbolic-tests/union-31-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( Z:MyId in ( S
+    SetItem ( X:MyId ) ) |Set SetItem ( Y:MyId ) ) ~> .
```
ana-pantilie, line 3: What is this testing?

k-distribution/tests/regression-new/set-symbolic-tests/union-37-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( X:MyId in ( SetItem ( X:MyId )
+    SetItem ( y ) ) |Set SetItem ( x ) ) ~> .
```
ana-pantilie: Same q

k-distribution/tests/regression-new/set-symbolic-tests/union-38-spec.k.out
```diff
@@ -0,0 +1,17 @@
+  #Not ( {
+    X
+  #Equals
+    x
+  } )
+#And
+  <k>
+    unionResult ( X:MyId in ( SetItem ( X:MyId )
+    SetItem ( y ) ) |Set SetItem ( x ) ) ~> .
```
ana-pantilie: Shouldn't this get simplified?

k-distribution/tests/regression-new/set-symbolic-tests/union-5-spec.k.out
```diff
@@ -0,0 +1,11 @@
+  <k>
+    unionResult ( ( S
+    SetItem ( X:MyId ) ) |Set SetItem ( Y:MyId ) ) ~> .
```
ana-pantilie, line 3: What is this testing?

k-distribution/include/kframework/builtin/domains.md
radumereuta: Since this [PR](https://github.com/runtimeverification/k/pull/2831), you should be able to use the `comm` attribute on simplification rules.
But this requires that the syntax and the rule are annotated for safety reasons.
So in this case, you may need to modify _domains.md_ as well.

k-distribution/tests/regression-new/set-symbolic-tests/diff-17-spec.k.out
jberthold: This test does not test any specific simplification. If anything, it tests that simplifications do not over-simplify by matching what they should not match.
As mentioned, I kept all the (comprehensive) tests from the older PR here, and the ones that have correct expected behaviour were kept as tests.

k-distribution/tests/regression-new/set-symbolic-tests/inset-11-spec.k.out
jberthold: Same as above, this test was kept from the previous set of tests because it produces a correct answer, not because it tests any particular simplification.
If you want I can clean up the tests that do not apply any more. However, we already found [a bug](https://github.com/runtimeverification/haskell-backend/issues/3294) with one of those "useless" tests.

k-distribution/tests/regression-new/set-symbolic-tests/intersect-23-spec.k.out
jberthold: No, unfortunately we do not have rule `E1 in (S SetItem(E2)) => E1 in S requlres E1 =/= E2` because it has severe performance impact.

k-distribution/tests/regression-new/set-symbolic-tests/union-38-spec.k.out
jberthold: Would be nice if it was, yes. We would need a rule like `rule X in ((SetItem(X) S) |Set _) => true ensures notBool (X in S)` to "see through the set union and discover item `X` - I might try that quickly

k-distribution/include/kframework/builtin/domains.md
jberthold: I cannot say for sure _why_ it happens, but :stopwatch: I wrote this test of `X in S requires X ==K X0` with S containing `X0` and 100 other variables.

### Version without this rule
```
jost@freshcode-1:~/work/RV/code/k/k-distribution$ time make -C tests/regression-new/set-symbolic-tests/ inset-100-elements-spec.k
make: Entering directory '/home/jost/work/RV/code/k/k-distribution/tests/regression-new/set-symbolic-tests'
/home/jost/work/RV/code/k/k-distribution/bin/kprove inset-100-elements-spec.k --warnings none --no-exc-wrap  -d . 2>&1 | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' | diff - inset-100-elements-spec.k.out
make: Leaving directory '/home/jost/work/RV/code/k/k-distribution/tests/regression-new/set-symbolic-tests'

real	0m11.867s
user	0m16.674s
sys	0m0.585s
```
### Version with this rule added
```
jost@freshcode-1:~/work/RV/code/k/k-distribution$ git diff
diff --git a/k-distribution/include/kframework/builtin/domains.md b/k-distribution/include/kframework/builtin/domains.md
index fe5f137f33..7c393a22ab 100644
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -695,8 +695,8 @@ module SET-KORE-SYMBOLIC [kore,symbolic]
 // These two rules would be sound but impose a giant overhead on `in` evaluation:
   // rule E1 in (S SetItem(E2)) => true requires E1 in S
   //                                 ensures notBool (E2 in S) [simplification]
-  // rule E1 in (S SetItem(E2)) => E1 in S requires E1 =/=K E2
-  //                                 ensures notBool (E2 in S) [simplification]
+  rule E1 in (S SetItem(E2)) => E1 in S requires E1 =/=K E2
+                                  ensures notBool (E2 in S) [simplification]
 
 endmodule
 
jost@freshcode-1:~/work/RV/code/k/k-distribution$ # (cd .. && mvn package)
jost@freshcode-1:~/work/RV/code/k/k-distribution$ time make -C tests/regression-new/set-symbolic-tests/ inset-100-elements-spec.k
make: Entering directory '/home/jost/work/RV/code/k/k-distribution/tests/regression-new/set-symbolic-tests'
/home/jost/work/RV/code/k/k-distribution/bin/kompile --no-exc-wrap --backend haskell  set-tests.k -d .
/home/jost/work/RV/code/k/k-distribution/bin/kprove inset-100-elements-spec.k --warnings none --no-exc-wrap  -d . 2>&1 | sed 's!'`pwd`'/\(\./\)\{0,2\}!!g' | diff - inset-100-elements-spec.k.out
^Cmake: *** [../../../include/kframework/ktest.mak:121: inset-100-elements-spec.k] Interrupt


real	2m21.244s
user	2m35.343s
sys	0m2.767s
```
Could it be that the simplifier ends up with 100 simplified results -- for all elements  which are not `X` --  and then continues to simplify each of these, getting 99 results each, etc. ... :boom: ?

k-distribution/include/kframework/builtin/domains.md
jberthold: Good question, actually... there was a problem with the other rule (which I now analysed and opened https://github.com/runtimeverification/haskell-backend/issues/3301 for). Both rules introduce wrong behaviour (see the new ticket) so I am removing this rule, too. The bug did not surface in the tests here, all tests are still passing without this rule (on master, using HS backend with internal set simplifications)

k-distribution/include/kframework/builtin/domains.md
jberthold: Thanks, I will use `comm` where appropriate then.

k-distribution/tests/regression-new/set-symbolic-tests/failing/union-18-spec.k
jberthold: This is related to the bug https://github.com/runtimeverification/haskell-backend/issues/3301, rule was removed.

k-distribution/tests/regression-new/set-symbolic-tests/union-38-spec.k.out
jberthold: :+1: Rules added, changing union-37 and union-38 (I also added `union-39` for the right-hand side of `|Set`)

PR #2921, opened 2022-09-28
Nix flake release - caching K and kup

goodlyrottenapple:
This CI run should keep master builds in the k-framkework cachix cache.


.github/workflows/release-nix.yml
```diff
@@ -0,0 +1,58 @@
...
+          - runner: ubuntu-latest
+            os: ubuntu-latest
+          - runner: macos-12
+            os: macos-12
+          - runner: MacM1
+            os: self-macos-12
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - name: 'Check out code'
+        uses: actions/checkout@v2.3.4
```
F-WRunTime: The latest version for checkout is @v3.0.2

F-WRunTime (dismissed):

dwightguth (dismissed):
We ought to already have a workflow that runs on pushes to master; this should probably be part of that workflow.

ehildenb:
There are no workflows on push to master currently @dwightguth , they are handled by Jenkins.

I'm working on migrating us over to GH Workflows for everything, but I think this job is fine.

We should rename the file and the workflow to "Master Push" instead of "Nix Release", and just have the particular job called Nix Release. This is in anticipation of me switching more of the Jenkins workflows over to GH runners.

.github/workflows/release-nix.yml
```diff
@@ -0,0 +1,58 @@
+name: "Release nix"
```
ehildenb:
```suggestion
name: "Master Push"
```

Also rename the file to `master-push.yml`. We'll be adding more jobs to it soon.

.github/workflows/release-nix.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: 'Build and cache K Framework'
+        run: |
+          export GC_DONT_GC=1
+          export CACHIX_AUTH_TOKEN=$RV_DEVOPS_CACHIX_TOKEN
+          nix shell nixpkgs#jq
+          nix-store --query --requisites --include-outputs \
+            $(nix-store --query --deriver \
+              $(nix build .#k --json | jq -r '.[].outputs | to_entries[].value')) \
+            | cachix push k-framework
```
ehildenb: I'd rather format this like this (for readabliity, and to make sure that it's more clear which subcommand fails if one of them fails):

```suggestion
          deriver=$(nix build .k --json | jq -r '.[].outputs | to_entries[].value')
          outputs=$(nix-store --query --deriver ${deriver})
          nix-store --query --requisites --include-outputs ${outputs} | cachix push k-framework
```

Same with the other steps in this job which use this pattern.

ehildenb (approved):

PR #2924, opened 2022-09-29
Add M1 mac to CI

Baltoli:
This PR enables CI on Apple Silicon for K, via the Nix flake derivation. Some minor changes to the Nix code are required to work around long-standing issues in an upstream Nix package; otherwise, the changes are formulaic following the PRs made to the LLVM and Haskell backends (https://github.com/runtimeverification/llvm-backend/pull/538, https://github.com/runtimeverification/haskell-backend/pull/3278)


.github/workflows/test-pr.yml
```diff
@@ -256,9 +269,16 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, macos-11]
-    runs-on: ${{ matrix.os }}
+        include:
+          - runner: ubuntu-latest
+            os: ubuntu-latest
+          - runner: macos-12
+            os: macos-12
+          - runner: MacM1
+            os: self-macos-12
```
goodlyrottenapple: I'm not sure the old-style nix expression will work on the M1 mac. Seems its failing on a seg-fault atm, so could try adding `GC_DONT_GC=1` to the `nix-build` commands. Otherwise best to just remove, as we will be migrating away from these old nix derivations at some point anyway.

Baltoli: If the long-term intent is to deprecate those derivations, I'll probably just remove it from this matrix. For this platform we're already being pretty aggressive about where you can build things due to tool compatibility, so having the flake build be the one true build shouldn't be an issue.

F-WRunTime (approved):

PR #2935, opened 2022-09-30
Move 'Nix flake submodule sync' to a separate workflow

### LINKED ISSUE #2718: [nix] Keep haskell-backend and llvm-backend flake inputs in sync with submodules
We need to add the following to the script which updates the haskell/llvm submodules and pushes to master:

```
nix flake lock \
  --override-input haskell-backend github:runtimeverification/haskell-backend/<hash> \
  --override-input llvm-backend github:runtimeverification/llvm-backend/<hash> \
  --recreate-lock-file
```

@F-WRunTime not sure where exactly we update and push the submodules?

goodlyrottenapple:
fixes #2718


F-WRunTime (approved):

goodlyrottenapple:
@ehildenb Can you approve? Also, It seems when I open a new branch/PR against `develop` directly, none of the CI runs get triggered.

dwightguth:
Will this fix the issue where submodule bump PRs fail tests until flake.lock is updated?

goodlyrottenapple:
yes

dwightguth (approved):

dwightguth:
This just needs to have the expected checks adjusted before it can be merged.

dwightguth:
@F-WRunTime I will let you make that change.

.github/workflows/udapte-deps.yml
```diff
@@ -0,0 +1,56 @@
...
+  nix-flake-submodule-sync:
+    name: 'Nix flake submodule sync'
+    runs-on: ubuntu-latest
+    steps:
+      - id: config
+        run: |
+          ref=${{ github.ref }}
+          if [ "${{ github.event_name }}" == 'pull_request' ]; then
+            ref="${{ github.event.pull_request.head.sha }}"
+          fi
```
ehildenb: This block should never fire now, and can be removed.

Also, we are using `ref=${{ github.ref }}` in this bash block, when we could probably just use `${GITHUB_HEAD_REF}` variable directly with is set by the github workflow?

Either way, this `if` statement doeeesn't make sense any more in a branch push, but it can be removed in follow-up.

.github/workflows/udapte-deps.yml
```diff
@@ -0,0 +1,56 @@
...
+          ref="${ref#refs/heads/}"
+          echo "::set-output name=ref::$ref"
```
ehildenb: This is also basically fully unnecessary now that I read it. I'll make a separate PR cleaning up some of this stuff and the maven steps probably.

ehildenb (approved):

ehildenb:
I've made the adjustments to the required status checks.

.github/workflows/udapte-deps.yml
ehildenb: I'll make these changes in my local branch and make a PR with updates.

.github/workflows/udapte-deps.yml
ehildenb: Same, I'll make these changes in my local branch then open a PR with updates.

PR #2936, opened 2022-09-30
Improve error handling in `k-util.sh`

### LINKED ISSUE #2817: Process failure in `k-util.sh` should be more informative
I had a definition that was failing at macro expansion stage.

It only became apparent to me that this was the failure stage after adding `-v` to `krun`.

`k-util.sh` https://github.com/runtimeverification/k/blob/master/k-distribution/src/main/scripts/lib/k-util.sh should print out failing commands, so that the user knows what phase failed. But then it should still fail overall.

Baltoli:
This PR reorganises the `execute` function in `k-util.sh`:
* The amount of noise produced in verbose mode is reduced by _only_ executing the actual command inside `-x`; all conditional tests are performed outside.
* The return code of the command is stored, and the existing error reporting function is used to report any failures to the user (even outside of verbose mode).

Tested locally to make sure the output is good. For example, by patching `kore-expand-macros` to just return 35:
```
+ kparse --definition ./test-kompiled --module TEST .krun-2022-09-30-12-37-54-1ys0guZ5YC/tmp.PGM.pretty.10LcVUXY1O
+ ret=0
+ llvm-krun -c PGM tmp.in.PGM.5P6NhEbdgW KItem korefile --directory /home/bruce/code/scratch/test-kompiled -v -save-temps --dry-run -nm -o tmp.in.npwbCwNMfZ
+ ret=0
+ kore-expand-macros ./test-kompiled .krun-2022-09-30-12-37-54-1ys0guZ5YC/tmp.in.npwbCwNMfZ
+ ret=35
[Error] krun: kore-expand-macros ./test-kompiled 
.krun-2022-09-30-12-37-54-1ys0guZ5YC/tmp.in.npwbCwNMfZ
```

Fixes #2817


k-distribution/src/main/scripts/lib/k-util.sh
```diff
@@ -73,14 +73,27 @@ fi
...
-  if ${verbose}; then
-    set -x
-  fi
+  set +e
+
+  declare -a cmd
   if ${profile}; then
     TIMEFORMAT="%lR %lU %lS $*"
-    time "$@"
+    cmd=(time "$@")
```
dwightguth: I would really prefer we continue to use the shell's time command rather than the one located at /usr/bin/time.

dwightguth (changes requested):

k-distribution/src/main/scripts/lib/k-util.sh
Baltoli: Sure, no problem. I switched to make the code a bit cleaner, but I'm happy to revert for the sake of minimising the dependency surface

dwightguth (approved):

k-distribution/src/main/scripts/lib/k-util.sh
```diff
@@ -73,14 +73,29 @@ fi
...
-  if ${verbose}; then
-    set -x
-  fi
+  set +e
+
   if ${profile}; then
+    if ${verbose}; then
+      set -x
+    fi
+
```
ehildenb, line 82: Why was this moved inside the `if ${profile}` block? It seems we have the same code on both branches of the `if ${profile}`, and it's in the same subshell anyway (outer parens).

ehildenb: Ah, it's to keep the `if ...` statement from showing up in the debug output?

ehildenb (approved):

k-distribution/src/main/scripts/lib/k-util.sh
Baltoli: Yes, that's exactly it. Lifting the `if` outside makes the verbose output much noisier, so I thought it was worth a small amount of duplicated code.

PR #2922, opened 2022-09-28
Migrate Ubuntu packaging to GitHub workflows

ehildenb:
Part of: https://github.com/runtimeverification/k/issues/2869

~Blocked on: https://github.com/runtimeverification/k/pull/2896~
~Blocked on: https://github.com/runtimeverification/k/pull/2921~
~Blocked on: https://github.com/runtimeverification/k/pull/2940~
~Blocked on: https://github.com/runtimeverification/k/pull/2950~

- [x] Parameterize the dockerfiles in the distro of Ubuntu to use.
- [x] Factor out GH action `test-package` for the packaging process for the Ubuntu.
- [x] Use the factored out test-package to test Ubuntu Jammy on PRs.
- [x] Use the factored out test-package to build and test Ubuntu Jammy+Focal on pre-releases, and attach to the prerelease.
- [x] Remove Jammy/Focal packaging from Jenkinsfile.
- [x] Migrate the Dockerhub image preparation to prerelease stage.

PR #2940, opened 2022-09-30
GitHub action `test-package` for testing the Ubuntu packages

ehildenb:
This PR is part of https://github.com/runtimeverification/k/pull/2922:

- Parameterizes the dockerfiles in `.github/workflows` to accept the Ubuntu base image, and LLVM version, as parameters.
- Adds a new action `.github/acitons/test-package/action.yml`, which is parameterized in the Ubuntu base image and the llvm version.
- Switches the PR job over to using this action, to demonstrate that it works.


.github/actions/with-docker/action.yml
```diff
@@ -16,21 +22,24 @@ runs:
...
       docker build . --file ${SUBDIR}.github/workflows/Dockerfile    \
           --tag runtimeverification/${TAG_NAME}                      \
+          --build-arg BASE_IMAGE=${BASE_IMAGE}                       \
           --build-arg Z3_VERSION=${Z3_VERSION}                       \
           --build-arg K_VERSION=${K_VERSION}                         \
-          --build-arg LLVM_VERSION=14                                \
+          --build-arg LLVM_VERSION=${LLVM_VERSION}                   \
           --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g)
 
       docker run                             \
```
F-WRunTime, line 45: The --user User:user  flag here will create all artifacts in the container as 'user' which doesn't exist within the host nodes shadow file. 

Checking some of the build agents (3/4) I did not see any files being created in the /opt/workspace/${SUBDIR} folder. This may not be an issue but I foresee we'll have upload/cleanup issues if anything eventually gets left behind by any workflows using the with-docker action.

F-WRunTime (commented):
The inputs are nice. 
If you want to add a workflow_dispatch option to test-pr it would enable running manually tests against any branch without having to open a PR.

.github/actions/with-docker/action.yml
ehildenb: Well, not really. It specifically creates them as the same uid and gid as the lxc containers `actions-runner` user, not as any specific user name.

The dockerfile itself sets up the `user:user` settings, so that it matches uid/gid of the external lxc container, that way the files that are created inside the container have premissions to be moved/deleted/modified outside the container.

F-WRunTime: Gotcha, I see on 43 it's seeing those ID's as parameters to the build arguments

F-WRunTime (approved):

dwightguth (approved):

PR #2949, opened 2022-10-04
Fix JQ error on M1 release CI

Baltoli:
Some combination of `nix shell`, Github actions, and a subshell invocation was breaking when run in CI. This PR uses the official nix shell action to avoid this problem.

I've tested these changes by removing the branch predicate from the workflow, and having it back out before pushing to cachix. The problem with `jq` is not present in this fixed version.

### DIFF
--- a/.github/workflows/master-push.yml
+++ b/.github/workflows/master-push.yml
@@ -41,16 +41,37 @@ jobs:
           signingKey: '${{ secrets.RV_DEVOPS_CACHIX_TOKEN }}'
 
       - name: 'Build and cache K Framework'
-        run: |
-          export GC_DONT_GC=1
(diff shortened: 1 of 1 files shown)


goodlyrottenapple (approved):

dwightguth (dismissed):
Please fix the commit history

Baltoli:
@dwightguth Cleaned up!

dwightguth (commented):
The change to the workflow looks okay but note that we aren't really trying to maintain correct version numbers on the `develop` branch. Normally I would say, it's fine, we can clean up the version number present on that branch in a future PR, but I'm not entirely sure that changing it on this branch won't interfere with the scripts that bump the version number on the master branch. So I would suggest rebasing that commit out.

dwightguth (approved):

PR #2950, opened 2022-10-04
Release workflow cleanups

ehildenb:
Part of: https://github.com/runtimeverification/k/pull/2922

This PR does several small cleanups on the `release.yml` file:

- Double => single quotes in places where possible for consistency.
- Make sure not more than one release job is running at a time (cancel existing ones in favor of new ones).
- Gives job names to each job in the release.yml file.
- Uses builtin GitHub event data in workflow for setting the `release_url` and `release_id`, instead of having a step to calculate it.
- Minor formatting.


.github/workflows/release.yml
```diff
@@ -97,16 +98,15 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const { owner, repo } = context.repo
-            await github.repos.deleteRelease({ owner, repo, release_id: ${{ steps.release.outputs.release_id }} })
+            await github.repos.deleteRelease({ owner, repo, release_id: ${{ github.events.release.release_id }} })
```
dwightguth: It looks like GitHub might allow both forms, but you're using "github.event" in some places and "github.events" other places. The documentation lists "github.event" so we probably should use this form everywhere just to be safe.

.github/workflows/release.yml
```diff
@@ -1,11 +1,16 @@
-name: "Release"
+name: 'Release'
 on:
   release:
     types:
       - prereleased
+concurrency:
```
dwightguth, line 6: This is a bad idea. We will end up with partial releases if we do this, which isn't desirable.

dwightguth (changes requested):

.github/workflows/release.yml
ehildenb: How about limiting to one release job ata  time? So we don't get the build servers croweded out by this job?

.github/workflows/release.yml
ehildenb: Fixed, all should be `event` instead.

.github/workflows/release.yml
ehildenb: I've removed the `cancel-in-progress`, so that we are limited to one release build going at a time, but they all will get scheduled.

dwightguth: Sounds good.

dwightguth (approved):

.github/workflows/release.yml
F-WRunTime: We could schedule a run per day? To capture everything within a 24 hour period. Or some cadence that seems reasonable.

F-WRunTime: To stick with the concurrency route, the ID used could be something along the lines of 
```
  group: ${{ github.workflow }}-<releasenumber> 
```
This would not however limit multiple releases running all at once.

PR #2952, opened 2022-10-05
Skip nix-maven step when possible

ehildenb:
Looking through the history, it appears that the `nix/mavenix.lock` file is onyl changing when one of four other files are changing:

- LLVM backend.
- Haskell backend.
- Nix source files (nix/*.nix)
- pom.xml files (**/pom.xml)

So this only runs this nix-maven step when one of those  files has changed, because it's an expensive step and the rest of the workflow doesn't run until it completes.

It also rarely has an effect, maybe once every 6 months. So we should not block CI on it.

In total, this PR:

- Only runs the nix-maven step if needed.
- Does not make the erroneous checks from before in the step for what type of github event this is (we know it's always a pull_request event).


ehildenb:
This reduces us too 43s for the nix-maven step (**this PR**): 
image

And without the if-statement guard (8m) (**develop branch**): 
image

dwightguth (approved):
The code looks fine, I have a question before we merge though. In the event that your pattern list of files is incomplete or somehow becomes incomplete in the future, what will happen? Will the out of date mavenix files be caught elsewhere in the CI?

ehildenb:
Yes, if it's something that fails the build (version must be bumped), then the Nix step later in CI will fail.

I believe once we have Nix flakes (and kup), that we can do away with this Mavenix caching, and the pure nix testing steps. Not sure though.

F-WRunTime (approved):

Issue #2938, opened 2022-09-30
`kup install k --version ...` does not work


ehildenb:
I have `kup` installed, and I tried to install a specific version of K:

```
% kup install k --version 5.4.11
❗ The package 'k' is already installed. Use 'kup update k' to update to the latest version.

% kompile --version
K version:    5.4.9
Build date:   Fri Sep 30 03:35:52 UTC 2022
```

So it's not installing the specific version I want, but staying on the prior version.

goodlyrottenapple:
`kup` expected the following command for updating already installed packages: 

```kup update k --version ...```

Also note that the version argument has to either be a git commit hash or the tag, so in this instance,

```
kup update k --version v5.4.11
```

**Comment by ehildenb** [comment]
How can I see the acceptable version tags? Can kup script pull them down? `kup list k --versions` or `kup list k --all` or something?

Or `kup version k --all`?

**Comment by goodlyrottenapple** [comment]
I have been thinking about this. github does have some APIs which do not require auth tokens, so we could potentially add a call to list the versions. however, we will probably have to manually hard-code the commit past which no nix flake exists and therefore cannot be installed via `kup`

**Comment by goodlyrottenapple** [comment]
I've started working on a `kup list <package>` command. The current output is:
```
kup list k
┌─────────────────────────┬──────────────────────────────────────────┐
│ Version                 │ Commit                                   │
├─────────────────────────┼──────────────────────────────────────────┤
│ vsc                     │ b1e46e5f80a5a289ec3520fc79d545eb9ce50cac │
│ v5.4.13                 │ ce15cf3dd86057a8755b0a562adaec6e85f0cc53 │
│ v5.4.12                 │ 4ba5f845950b74a84376e3639796c6a0b3dc6c2b │
│ v5.4.11                 │ 155d7aaa7fe68b51b290d45d2ea4bf21ead34f77 │
│ v5.4.9_6009803          │ 5f5e2fe94e1182dbae5a49466f10dc5bf9e6cc41 │
│ v5.4.9_6664c36          │ 5f5e2fe94e1182dbae5a49466f10dc5bf9e6cc41 │
│ v5.4.9_9f46a0a          │ 5f5e2fe94e1182dbae5a49466f10dc5bf9e6cc41 │
│ v5.4.9_6f2eea1          │ 5f5e2fe94e1182dbae5a49466f10dc5bf9e6cc41 │
[...]

PR #2957, opened 2022-10-05
`kup` improvements

goodlyrottenapple:
Fixes https://github.com/runtimeverification/k/issues/2938

* `kup update <package> --version ...` and `kup install <package> --version ...` behave the same now
* `kup list <package>` fetches the git history and shows commits/tags

image


ehildenb:
@goodlyrottenapple can you skip the Merge commits in the output? If that's easy, it would be nice to clean it up a bit.


Also, we should be ready to handle the pagination, but we probably don't need to yet.

ehildenb:
Any commit with title `Merge remote-tracking branch 'origin/develop'` can be skipped basically. It would also be nice to just display one commit per version, but that can be later.

goodlyrottenapple:
I can hard code it so that certain commits with a given prefix don't show up in the list. Bear in mind that different repositories, e.g. haskell-backend have different setup and e.g. have no version tags

goodlyrottenapple:
> Any commit with title `Merge remote-tracking branch 'origin/develop'` can be skipped basically

done

ehildenb:
Long term, we should probably be storing our own database with `kup` accessible versions, instead of relying on GitHub API, so we can get faster retrieval and storage of the information how we like. But for now this looks good :)

ehildenb (approved):

PR #2959, opened 2022-10-06
Simplify several workflows procedure

ehildenb:
This PR does several small cleanups:

- A double-quote to single-quote.
- Use checkout@v3 action uniformly throughout the workflows.
- Do not redundantly specify the checkout ref in several places.
- Do not compute the checkout ref as a job outptu anymore (not needed).
- Simplify logic for adding changes: run the update command then see if `git add && git commit` succeeds, and if so then do `git push`. No need to check ahead of time if the update command _will_ make changes.
- Uses JENKINS_GITHUB_PAT for update-deps checkout, so that the pussh retriggers workflows.


dwightguth (approved):
Seems fine to me assuming you've tested it.

F-WRunTime (approved):

PR #2961, opened 2022-10-06
2612 automatically add a ceil simplification rule for sets

### LINKED ISSUE #2612: [KIP] - Support for symbolic sets

jberthold:
Fixes #2612 
Fresh version of #2692, adding a simplification rule
`#Ceil(@S:Set SetItem(@E:KItem)) => {(@E in @S) #Equals false} #And #Ceil(@S) #And #Ceil(#E) [simplification]`
automatically.
Companion PR to #2920 (which added the rule explicitly in `domains.md` , now removing it here).


ana-pantilie (commented):
I can't review the Java changes, but the tests LGTM.

dwightguth (approved):
Looks alright although I haven't read the code in super fine detail. I assume you've inspected the resulting axiom to make sure it looks how you want?

PR #2963, opened 2022-10-06
Add preserves-definedness attribute for semantic rules

radumereuta:
First item of runtimeverification/haskell-backend#3373
Needs more testing.
Waiting on the haskell-backend to provide a way to check if these changes are correct.

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
@@ -5,13 +5,17 @@ import com.google.inject.Inject;
 import org.apache.commons.io.FilenameUtils;
 import org.kframework.Strategy;
 import org.kframework.attributes.Att;
+import org.kframework.attributes.Location;
+import org.kframework.attributes.Source;
 import org.kframework.compile.*;
 import org.kframework.definition.Module;
 import org.kframework.definition.*;
 import org.kframework.kompile.CompiledDefinition;
 import org.kframework.kompile.Kompile;
 import org.kframework.kompile.KompileOptions;
+import org.kframework.kore.KApply;
 import org.kframework.kore.KLabel;
+import org.kframework.kore.KVariable;
 import org.kframework.main.Tool;
 import org.kframework.utils.errorsystem.KEMException;
 import org.kframework.utils.errorsystem.KExceptionManager;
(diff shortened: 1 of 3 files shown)


radumereuta:
Did some quick checks with @ana-pantilie and we think the front-end part is correct for most of the tests we ran.
I tried to kompile the c-semantics and it passed.
KEVM printed some warnings during kompilation though:
```
Parse rules [1507/1507 rules]                                =  4.800s
Validate definition                                          =  0.496s
Apply compile pipeline                                       =  4.467s
Kompile to kore                                              =  7.096s
Save to disk                                                 =  0.211s
  Print definition.kore                                      =  1.413s
kore-exec: [3348821] Warning (LogMessage):
    The following rule was incorrectly marked preserves-definedness: /home/radu/work/evm-semantics/.build/usr/lib/kevm/include/kframework/evm.md:960:10-961:35 []
kore-exec: [3595568] Warning (LogMessage):
    The following rule was incorrectly marked preserves-definedness: /home/radu/work/evm-semantics/.build/usr/lib/kevm/include/kframework/evm.md:311:10-320:64 []
kore-exec: [3888174] Warning (LogMessage):
    The following rule was incorrectly marked preserves-definedness: /home/radu/work/evm-semantics/.build/usr/lib/kevm/include/kframework/evm.md:1869:10-1870:75 []
  Validate def                                               =  4.762s
Backend                                                      =  6.175s
```
I'm not sure why. Need to check tomorrow.

ehildenb:
I think the backend is handling it directly themselves at this point. @jberthold is that correct? Does the booster backend add the attribute at definition internalization time, or are we still relying on the frontend to put the attribute in?

jberthold:
> I think the backend is handling it directly themselves at this point. @jberthold is that correct? Does the booster backend add the attribute at definition internalization time, or are we still relying on the frontend to put the attribute in?

Yes and no. 
What the new backend does is:
* if a rule or equation has a `preserves-definedness` attribute, mark it as preserving definedness
* otherwise:
  - mark a _rule_ as preserving definedness if its RHS term only contain total function symbols or constructors;
  - mark an _equation_ as preserving definedness if its LHS- _and_ RHS terms both only contain total funciton symbols or constructors.
I don't fully understand the java code but it appears to do exactly the same.
What the OS backend does I would have to check first - I am not aware of any work on the definedness flag.

This analysis is conservative, as many rules guarantee to _use_ partial functions only on arguments where the functions are indeed defined - therefore we allow the user (or the front-end) to mark a rule as definedness-preserving manually.

radumereuta:
@jberthold I was actually interested in the usefulness of this PR.
Is it still needed? Can it be deleted if not?
This was started before the haskell booster to try and speed up the backend but I haven't heard anything about it since.

jberthold:
> @jberthold I was actually interested in the usefulness of this PR. Is it still needed? Can it be deleted if not? This was started before the haskell booster to try and speed up the backend but I haven't heard anything about it since.

I guess we could still keep the PR around but the sibling PR in `haskell-backend` https://github.com/runtimeverification/haskell-backend/pull/3304 was closed (unmerged) since our focus was the new backend.

PR #2966, opened 2022-10-07
Quick fix to use actions checkout for push and pull w/ generated

F-WRunTime:
Checkout tokens


dwightguth:
Looks like the commit history of this is messed up.

PR #2967, opened 2022-10-07
Rework update deps

ehildenb:
This PR does a few cleanups:

- Makes sure that all "Install Nix" steps are authenticated, to avoid rate limits.
- Makes sure that we use `JENKINS_GITHUB_PAT` for the clone on nix-flake sync step,   so that it retriggers workflows on push.
- Simplifies the logic of `git add/commit/push` in the flake submodule sync.


ehildenb:
Subsumed: https://github.com/runtimeverification/k/pull/2959

PR #2968, opened 2022-10-10
Fix quoting in IDs

### LINKED ISSUE #2315: [Bug] Id created by String2Id contains quotation marks
K Version
---------

```
K version:    v5.1.250-0-g07918e142
Build date:   Sat Nov 06 22:18:15 CET 2021
```

Description
-----------

The Id created by `String2Id("x")` is `"x"`, i.e., the Id includes the quotation marks of the string, which seems wrong. This behavior is demonstrated by the following example:

```
module STRING2ID
  imports ID
  imports BOOL
  imports K-EQUAL
  imports STRING

  syntax Bool ::= "test" [function]
  
  // Evaluates to true
  rule test => String2Id("x") ==K #token("\"x\"", "Id")
endmodule
```

Baltoli:
The problem here is that the re-wrapping of constant-folded values (i.e. after calling the compile-time Java implementation of the hook) into their argument positions will enquote a string result, even if it's of sort `Id`. For `Id` results, we want to just construct the token directly..

Fixes #2315


radumereuta (approved):

dwightguth (changes requested):
Id is not a special sort in k; the compiler should not be providing special behavior for it. This functionality needs to be common to any token sort that isn't marked with the STRING.String or BYTES.Bytes hook

dwightguth (approved):

dwightguth:
I approved but you might want to add a test for a non Id sort to the test suite also.

PR #2971, opened 2022-10-10
Possible fix for colliding container names on start of new job

F-WRunTime:
Running on a hunch --rm flag in docker run and running docker rm could be causing issues. It's worth trying to remove this simple flag from the action and see if this could resolve the conflicts.

As all workflows explicitly call 'docker stop' / 'docker rm' this is the least number of line changes to try this fix.
The alternative approach would be to flip this and add back the automatic clean-up and let docker handle this when we call 'docker stop' (remove any calls to 'docker rm')

Caveat:
This may not be the final fix if the docker service still has resources named for the container in the background when a rm or stop call is made and clean-up is supposed to begin/finish. If this is the case we may need a short delay or impose using container names with a UID (e.g. git commit) to fix the problem.


F-WRunTime:
I have pushed a change to also force remove. This should ignore any stop commands potentially still in progress and clean remove the container. 

There may still be an issue with concurrency when a workflow gets canceled. I have observed container collision occurs when a previous workflow was in the middle of building the jammy package and is then canceled and the next job spins up. ONLY when the next job starts and fails does the container actually get removed. I don't think this is a delay of removal but never being called at all.

dwightguth:
This seems reasonable but I'd like to get Everett's review also in case I missed anything.

F-WRunTime:
> This seems reasonable but I'd like to get Everett's review also in case I missed anything.
Originally going on a hunch before I could investigate further since it didn't hurt to remove some redundancy. 

BUT! I did some further testing [here](https://github.com/runtimeverification/devops-actions/actions/workflows/BuildInterruption.yaml) 
Using two different instances of the same workflow one with a functional change fixing the issue and on with what is currently running in the k repository workflow and causing containers to remain running. 

The only difference is the usage of the -t flag when stopping a container `docker stop -t 0 -f <container> ...` 
- This means to set the timeout for when to stop a container to 0 (default=10 seconds) 

You can play around with the two different workflows I created by starting a workflow and then immediately interrupting it and  starting another. This simulates and reproduces our current state in the k repository causing false failures.

[Successful Workflow](https://github.com/runtimeverification/devops-actions/actions/workflows/BuildInterruption.yaml)
[Failure Workflow](https://github.com/runtimeverification/devops-actions/actions/workflows/BuildInterruptionFAIL.yaml)

It does depend on the single github-runner-4 as we must be able to quickly restart the run on the same runner otherwise we get a random new runner which is why intermittent failures were occurring for k. Some pass some fail.. unpredictable.

F-WRunTime:
Merge conflicts fixed. Updated to remove extra 'rm' calls. Docker will clean-up on calling stop. We will also make sure if a stop call fails we know about it.

.github/actions/test-package/action.yml
```diff
@@ -39,8 +39,8 @@ runs:
       DISTRO: ${{ inputs.distro }}
     if: always()
     run: |
-      docker stop k-package-build-${DISTRO} || true
-      docker container rm k-package-build-${DISTRO} || true
+      docker stop -t 0 k-package-build-${DISTRO} || true
+      docker container rm -f k-package-build-${DISTRO} || true
```
ehildenb: If options have long and short variants, use the long variants in scripts please (I guess `-t` is `--timeout`?)

Same for `-f`. Make that change throughout the file.

Also, I guess we should remove the ` || true`? If that command doesn't work, something is wrong with our assumptinos about how Docker is working.

F-WRunTime: It would be best to know something fails. Yes removing ` || true` would help us identify an issue sooner rather than later.

F-WRunTime: @ehildenb one thing is with the rm is we can remove this and rely on the automatic clean-up. 

If we wish tokeep in the extra remove call it will require the ||true as it will likely fail more often than not

.github/workflows/release.yml
```diff
@@ -67,8 +67,7 @@ jobs:
       - name: 'Clean up Docker Container'
         if: always()
         run: |
-          docker stop -t 0 k-package-docker-test-jammy || true
-          docker rm -f k-package-docker-test-jammy || true 
+          docker stop -t 0 k-package-docker-test-jammy
```
ehildenb:
```suggestion
          docker stop --time=0 k-package-docker-test-jammy
```

ehildenb (approved):

Issue #1890, opened 2021-03-31
[Bug] [kprove] - Unable to find klabels in syntax priorities statement


ehildenb:
K Version
---------

Tell us what version of K you are using.

Example:

```
% kompile --version
K version:    5.0.0
Build date:   Tue Mar 30 23:29:40 UTC 2021
```

Description
-----------

A definition that includes a custom tag for some productions, and then uses them in a `syntax priorities` statement, is able to `kompile` fine, but at `kprove` time it fails to find the labels mentioned in the `syntax priorities` statement.

Input Files
-----------

File `test.k`:

```
module TEST
    imports INT

    configuration <k> .K </k>
    syntax Int ::= Int "seconds" [timeUnit]
 // ---------------------------------------
    rule N seconds => N [macro]

    syntax priorities timeUnit > _+Int_
endmodule
```

File `spec-2.k`:

```
requires "test.k"

module VERIFICATION
    imports TEST

    syntax KItem ::= runLemma ( Int ) | doneLemma ( Int )

    rule <k> runLemma(I) => doneLemma(I) ... </k>
endmodule

module SPEC-2
    imports VERIFICATION

    claim <k> runLemma(3 +Int 7) => doneLemma(10) ... </k>
endmodule
```

Reproduction Steps
------------------

First kompile the definition with the haskell backend, then run the proof:

```
% kompile test.k --backend haskell
[Warning] Compiler: Could not find main syntax module with name TEST-SYNTAX in
definition.  Use --syntax-module to specify one. Using TEST as default.
% kprove spec-2.k
[Error] Outer Parser: Could not find any productions for tag: _+Int_
        Source(/home/dev/src/mkr-mcd-spec-pr/test.k)
        Location(8,5,8,39)
```

Expected Behavior
-----------------

If this should be an error, then it should happen at `kompile` time. If not, then the proof should go through.

radumereuta:
This happens because we're building the spec definition from two halves.
The first half is parsed and cached during kompile, and the second half is parsed during kprove.
The simplest workaround is to `require "domains.md"`. This way you tell the front end where to find the missing tag.

A proper fix would require a bit more digging, and probably some significant redesign on how we build the proof definition.

ehildenb:
It seems that even adding the explicit `requires "domains.md"` to the definition itself (not to the proof) is enough.

radumereuta:
[...]

PR #2972, opened 2022-10-10
Relax module duplication check

### LINKED ISSUE #2910: Compiled binaries are not portable (problem with Nix builds)
When running the kevm docker with tag [ubuntu-focal-0e96c8d](https://hub.docker.com/layers/runtimeverificationinc/runtimeverification-evm-semantics/ubuntu-focal-0e96c8d/images/sha256-7ae7b8fece78a2ce4fd40f389a61a0814d40f770155c70b8cdcf25cdbc6a0cd4?context=explore) with `kevm prove` it gives me the error

```bash
[Error] Outer Parser: Module WORD previously declared at
Source(/home/jenkins-runner/workspace/KEVM_master/focal-build/.build/usr/lib/kevm/include/kframework/word.md)
and Location(7,1,336,10)
	Source(/usr/lib/kevm/include/kframework/word.md)
	Location(7,1,336,10)
	    .	v~~~~~~~~~~
	  7 |	module WORD
	    |		...
	336 |	endmodule
	    .	~~~~~~~~^
``` 

while `find / -name 'word.md'` results in only

```bash
/usr/lib/kevm/include/kframework/word.md
```

Does anyone know to to resolve this?

radumereuta:
Allows for file copy by checking: 1. filename 2. location and 3. module digest
Fixes: #2910


kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -202,15 +202,15 @@ public Set<Module> loadModules(
 
         new CollectProductionsVisitor(kore, context).visit(def);
 
-        // Tuple3 of moduleName, Source, Location
-        Map<String, List<Tuple3<String, Source, Location>>> groupedModules =
+        // Tuple3 of moduleName, Source, Location, digest
```
Baltoli: **Extremely** minor but the comment doesn't match the added text or new type

Baltoli (approved):

Baltoli (changes requested):
I don't think the error message in the test here is particularly clear - the modules here **are** in fact identical, but happen to be declared in the same file. We probably want to distinguish (if possible) this case from "the exact same module, but moved on disk".

kernel/src/main/java/org/kframework/kil/Module.java
```diff
@@ -57,5 +57,19 @@ public void toString(StringBuilder sb) {
...
+     * Used to determine if a module was modified.
+     * <a href="https://github.com/runtimeverification/k/issues/2910"> GH #2910</a>
+     * This allows for scripts to move files around after compilation but still give error messages
+     * if the contents of modules differ.
+     */
+    public long digest() {
+        StringBuilder mod = new StringBuilder();
+        toString(mod);
+        long hash = 7;
+        for (int i = 0; i < mod.length(); i++) {
```
dwightguth: this is not at all a collision resistant hash function and is thus not suitable as a proxy for equality.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -202,15 +202,15 @@ public Set<Module> loadModules(
...
+        Map<String, List<Tuple4<String, Source, Location, Long>>> groupedModules =
```
dwightguth: this is probably not what you want. The tuple of String, Source, Location was being used to identify modules that appeared in two places with different filenames or locations. You probably just instead want to identify two modules that have the same name but different contents, in which case you really only need module name and digest in the value of the map.

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -220,12 +220,20 @@ public Set<Module> loadModules(
...
-          Tuple3<String, Source, Location> firstMod = groupedModules.get(moduleName).get(0);
-          Tuple3<String, Source, Location> secondMod = groupedModules.get(moduleName).get(1);
-          KEMException ex = KEMException.outerParserError("Module " + moduleName + " previously declared at "
+          Tuple4<String, Source, Location, Long> firstMod = groupedModules.get(moduleName).get(0);
+          Tuple4<String, Source, Location, Long> secondMod = groupedModules.get(moduleName).get(1);
+          KEMException ex = KEMException.outerParserError("Module " + moduleName + " differs from previous declaration at "
                   + firstMod._2() + " and " + firstMod._3(), secondMod._2(), secondMod._3());
-          errors++;
-          kem.addKException(ex.getKException());
+          // give an error message only if we have
```
dwightguth, line 225: if you do what I suggested above, we don't really need this check.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -315,7 +315,7 @@ public void convert(boolean heatCoolEq, String prelude, StringBuilder semantics,
         semantics.append("\n// priority groups\n");
         genPriorityGroups(priorityList, priorityToPreviousGroup, priorityToAlias, topCellSortStr, semantics);
         semantics.append("endmodule ");
-        convert(attributes, module.att(), semantics, null, null);
+        convert(attributes, module.att().remove(Att.DIGEST()), semantics, null, null);
```
dwightguth, line 308: instead of doing this, let's just remove the digest attribute when converting from a FlatModule back to a Module.

dwightguth (changes requested):
some concerns; you probably also want to add `digest` to the reserved attribute list in the documentation.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
radumereuta: That's not going to work for kprove.
We store `CompileDefinition.parsedDefinition` which has `Module`.
If we strip that attribute we won't have it when we do `slurp` in kprove. Which is where the failure happens.

dwightguth:
Okay I am now very confused about exactly what this change is supposed to be fixing. Let's discuss in the meeting tomorrow. I don't understand why kprove would be worried about duplicate modules having a different location unless it's reparsing the definition during proof time, which it should not be doing anymore.

radumereuta:
Guessing by the comments on this PR, all 3 of us have different opinions on what should happen when we encounter duplicate modules.
This PR tries to solve a problem in kprove that appears very often in KEVM because of how the running scripts work there.
If you kompile a definition, and then move the files somewhere else, and run kprove, then you will run into this issue.
This can happen when you `require "domains.md"`. Outer parsing will recursively go through all the files and only later will compare if those modules do exist in the already compiled definition.

My implementation is the most strict I can think of. It only allows for moving files around.
@dwightguth you proposed to only check for module-name and digest. That would not work for duplicate modules in the same file. Like the test I modified in this PR. Maybe Bruce is right. We should have a somewhat different error message for that case.

Let's discuss this in the KDev meeting tomorrow.

dwightguth:
We probably should just keep all modules in the definition in lexical scope when parsing kprove and then make it an error to redeclare any of them in your spec module, including via `requires`. Then you're not reparsing modules that have already been parsed, and you won't have this issue even without needing to keep track of a cryptographic hash of the module.

radumereuta:
We sometimes `require "domains.md"` in the spec file is because of this issue: https://github.com/runtimeverification/k/issues/1890

kernel/src/main/java/org/kframework/parser/ParserUtils.java
```diff
@@ -220,12 +220,20 @@ public Set<Module> loadModules(
...
+          // the location is different or
+          // the hashes of the pretty printed contents are different.
+          if (!Paths.get(firstMod._2().source()).getFileName().equals(Paths.get(secondMod._2().source()).getFileName())
+            || !firstMod._3().equals(secondMod._3())
+            || !firstMod._4().equals(secondMod._4())) {
+              KEMException ex = KEMException.outerParserError("Module " + moduleName + " differs from previous declaration at "
+                      + firstMod._2() + " and " + firstMod._3(), secondMod._2(), secondMod._3());
+              errors++;
+              kem.addKException(ex.getKException());
+          }
```
radumereuta, line 236: > I don't think the error message in the test here is particularly clear - the modules here **are** in fact identical, but happen to be declared in the same file. We probably want to distinguish (if possible) this case from "the exact same module, but moved on disk".

@Baltoli do you have a suggestion?
I can't really think of a better message with what I have in hand:
- file name (trying to exclude the path)
- location in the file
- hashed contents

Baltoli: Could you keep the full path around and only use it for this check, perhaps? i.e. same full path, different location, any hashed contents => error message for a redeclared module name?

radumereuta: If you look at the original issue, someone tried to redefine the LIST module. So I'm not sure how keeping the full path would work for the general problem.
I could change the error message back to the original. Might be more intuitive?

Baltoli: Ah, good point. In that case I'd leave it as you have it in this change, and if we come up to a particular case that confuses someone we can revisit it.

dwightguth (approved):

Baltoli (approved):

PR #2982, opened 2022-10-13
Update README for LLVM version on macOS

Baltoli:
Homebrew is shipping a newer version of LLVM by default (15). We only support up to version 14, so the README has been updated to ensure users get the right version.

These instructions should be updated when https://github.com/runtimeverification/llvm-backend/issues/576 is addressed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ If you install this list of dependencies, continue directly to the [Build and In
 On macOS using [Homebrew](https://brew.sh/):
 ```shell
 git submodule update --init --recursive
-brew install bison boost cmake flex gcc gmp openjdk jemalloc libyaml llvm make maven mpfr pkg-config python stack zlib z3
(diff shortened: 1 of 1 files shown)


radumereuta:
How hard it would be to support llvm 15?
We will have to do that eventually, and I find it awkward to ask users to downgrade a package.
@theo25

Baltoli:
> How hard it would be to support llvm 15?

It's probably not a huge amount of work, but I'm aware of at least one API we use that needs a non-trivial refactoring to fix (https://github.com/runtimeverification/llvm-backend/issues/518), as well as whatever the problem identified in the initial issue is. It definitely needs fixing, and I agree with you in principle, but we should update the documentation in the meantime so that people don't get confused, even if it means a downgrade.

radumereuta (approved):

Issue #2983, opened 2022-10-13
New Ubuntu Jammy packaging fails on release job


Baltoli:
https://github.com/runtimeverification/k/actions/runs/3240999414/jobs/5312347235

The release job fails because it's unable to find the asset `kframework_ubuntu_jammy.deb`. Looking at the action that's responsible for building the package, it seems like the name we use in the release job is wrong (and should possibly be `kframework_amd64_jammy.deb`): https://github.com/runtimeverification/k/blob/4d2c5f18c90c6ad537ff2fcb8c99cfb5a29838b9/.github/actions/test-package/action.yml#L77-L81

F-WRunTime:
@ehildenb checkout the merge @Baltoli and I worked on last week. This may have fixed the issue, and this didn't get closed out yet. There may be an additional problem we didn't identify yet

PR #2986, opened 2022-10-13
Speculative fix for release workflow

Baltoli:
Potential fix for #2983 by renaming the path that the release job expects to find the built `.deb` files at.


F-WRunTime (approved):
These will clean up build containers and collect the appropriate deb files for building the debs.

PR #2988, opened 2022-10-13
Fix version calculation

### LINKED ISSUE #2915: Small inconsistency between git and package install in format of reported version
There is a small discrepancy in the format of the reported version number between K installed from git repository and that installed from package.

Git:

```
% kompile --version
K version:    v5.4.7-0-g0b0189cc60
Build date:   Tue Sep 27 13:19:45 CEST 2022
```

Package:

```docker
% docker run --rm runtimeverificationinc/kframework-k:ubuntu-focal-5.4.7 kompile --version
K version:    5.4.7
Build date:   Mon Sep 26 17:46:17 CDT 2022
```

It would be more consistent if the packaged version also reported its version number with a prefix `v`. Then
- parsing / regex matching would be simpler;
- the version number would still indicate whether it's a git or a package install (due to the git version using `--long` format of `git describe`).

radumereuta:
Fixes #2915

### DIFF
--- a/k-distribution/tests/regression-new/help/check-git-revision-format
+++ b/k-distribution/tests/regression-new/help/check-git-revision-format
@@ -7,5 +7,5 @@ if [ -z "$GIT_REVISION" ]; then
   exit 0
 fi
 
-VERSION_REGEX='v\d+\.\d+\.\d+-\d+-g[0-9a-z]+(-dirty)?$'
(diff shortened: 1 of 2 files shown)


Baltoli (approved):

Baltoli (changes requested):
Looking back at #2915, I actually don't think this is the change Tamas requested. His preference is to prefix the short, package, version number with `v`, but to leave the long format in place even at distance 0 so that the parser can work out where the installation came from. See #2989

radumereuta:
Leave it as is. We like to have the hash at the end to know if someone built K from git or got it from a package.

PR #2989, opened 2022-10-13
Add -v to packaged version string

### LINKED ISSUE #2915: Small inconsistency between git and package install in format of reported version

Baltoli:
This fixes the second complaint in #2915 (namely, that the packaged version string reported by the K tools doesn't have `v` in front of it).

### DIFF
--- a/kernel/src/main/java/org/kframework/utils/file/JarInfo.java
+++ b/kernel/src/main/java/org/kframework/utils/file/JarInfo.java
@@ -85,7 +85,7 @@ public class JarInfo {
             // the release version if we're not (e.g. from a release tarball).
             String version = mf.getMainAttributes().getValue("Implementation-Git-Describe");
             if (version.isEmpty()) {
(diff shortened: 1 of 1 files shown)


radumereuta (approved):
Right. Didn't notice that.

PR #2997, opened 2022-10-14
Add test case for priority bug

### LINKED ISSUE #539: [Bug] [krun] - Rule application doesn't respect priority attribute when applying on a cell with copies
K Version
---------

```
$ kompile --version

K version:    5.3.0
Build date:   Mon Jul 04 14:48:51 CDT 2022

```

Description
-----------

Rule application doesn't seem to respect the `priority` attribute when applying on a configuration with some form of multiplicity.

Input Files
-----------

Definition file (named `priorities.md`):
```
module PRIORITIES
  imports DOMAINS

  syntax Pgm ::= "addCell"
               | "foo"
               | "bar"
               | "buzz"
               | Pgm Pgm [right]

  configuration <cells>
                  <cell multiplicity="*" type="Map">
                    <id> 0 </id>
                    <k> $PGM:Pgm </k>
                  </cell>
                </cells>
                <order> .K </order>

  rule <k> addCell => .  ... </k>
       <id> Id </id>
       (.Bag => <cell>
                  <id> Id +Int 1 </id>
                  <k> foo </k> ...
                </cell> )

  rule P1:Pgm P2:Pgm => P1 ~> P2

  rule <k> foo => buzz ... </k>
       <order> . => ListItem(foo) ... </order>

  rule <k> bar => buzz ... </k>
       <order> . => ListItem(bar) ... </order> [priority(250)]

endmodule
```
Program file (named `test.priorities`):
```
addCell bar
```

Reproduction Steps
------------------

 `kompile priorities.md` followed by  `krun test.priorities`.

Expected Behavior
-----------------
When run, the `order` cell contains `ListItem ( foo ) ~> ListItem ( bar ) ~> .`, indicating rule that rewrites `bar` to `buzz`, with a lower priority (indicated by attribute `priority(250)`) fired before the rule that rewrite `foo` to `buzz` with higher (default) priority. 

Expect rule with lower priority to fire last, i.e., rewrite `bar => buzz` to occur after rewrite `foo => buzz`.

Baltoli:
Won't pass until https://github.com/runtimeverification/llvm-backend/pull/587 is merged and propagated; this PR adds the test case from #539 verbatim to the regression test suite.

Closes https://github.com/runtimeverification/llvm-backend/issues/539


Baltoli:
@dwightguth Given that https://github.com/runtimeverification/llvm-backend/pull/587 adds it to the backend, is it worth duplicating this test case in the frontend for clarity, do you think? If not I'll close this PR

PR #3010, opened 2022-10-20
Have with-docker step run CI script directly

ehildenb:
This avoids the user having to do the docker setup/teardown themselves by passing the CI script directly into the `with-docker` step.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -81,14 +81,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-      - name: 'Set up Docker'
-        uses: ./.github/actions/with-docker
-        with:
-          tag: k-ci
-          distro: jammy
-          llvm: 14
-      - name: 'Build and Test K'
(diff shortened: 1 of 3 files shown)


.github/workflows/test-pr.yml
```diff
@@ -81,31 +81,25 @@ jobs:
...
-          docker stop --time=0 k-ci
-          docker container rm --force k-ci || true
+      - name: 'Run K Tests'
+        uses: ./.github/actions/with-docker
+        with:
+          tag: k-ci
+          distro: jammy
+          llvm: 14
+          script: |
+            |
```
F-WRunTime: What does the double pipe for this do?

ehildenb: It's to avoid YAML parsing errors I guess: https://github.com/octokit/request-action/issues/6

Baltoli:
@ehildenb @F-WRunTime this also seems like a nice-to-have; what's the status here?

ehildenb:
I don't think I ever got this actually working, something about passing in multi-line YAML blocks to actions not working properly for bash scripts. Closing for now, maybe we can re-initiate investigation later.

PR #3012, opened 2022-10-21
Add location info for KToken

radumereuta:
This occurred when trying to move a rule like this:
```
rule 1 => #token("asdf", "Exp")
```
We didn't properly add the location information when we did the transformation from TreeNodes to KORE and it was failing with a `None.get`.

@sskeirik can you please confirm this solves your issue?


sskeirik:
@radumereuta I tested this commit on my code and it appears to solve the issue for me! Thanks.

sskeirik (approved):
I don't know if the code is right, but it works for my test case, so I'm approving it. You may want someone else with more familiarity with the code to chime in though.

radumereuta:
Thank you @sskeirik 
It's a very small change, so don't worry about it.
Just glad this fixed the problem.

ehildenb (approved):

PR #3013, opened 2022-10-21
Fix Ubuntu Focal Release

ehildenb:


ehildenb:
Evidence this works: https://github.com/runtimeverification/k/actions/runs/3318304378/jobs/5483758855

ehildenb:
@F-WRunTime note that I reverted your changes to the job triggers, because they seemed to be making it so that new commits were not triggering runs (I had to close/re-open this PR to get new commits triggered).

.github/workflows/test-pr.yml
```diff
@@ -1,7 +1,6 @@
 name: 'Test PR'
 on:
   pull_request:
-    types: [opened, edited, reopened]
```
F-WRunTime, line 4: Instead of removing add, this was a miss on my part in the last push. 
types: [opened, edited, reopened, **synchronize**]

.github/workflows/Dockerfile.stack-deps
```diff
@@ -11,13 +11,17 @@ RUN    apt-get update        \
 
 RUN curl -sSL https://get.haskellstack.org/ | sh
 
+# The image is built specifically for an environment with this user/group
+ARG USER
+RUN echo "User: ${USER?'USER argument must be provided during docker build'}"
+ARG GROUP=$USER
 ARG USER_ID=1000
```
F-WRunTime, line 17: Why not set this to USER_ID? from the build arg line?

ehildenb: This is where USER_ID is defined as an argument. Saying `USER_ID=1000` is providing a default, but if you pass `--build-arg USER_ID=...` it overrides that default.

.github/workflows/test-pr.yml
ehildenb: Done.

F-WRunTime (approved):

Issue #2316, opened 2021-11-09
[CI] Eliminate all warnings and compare stdout


radumereuta:
Currently, in regression tests, we have many warnings printed that are not being tested for.

- [ ] scala and java code warnings
- [ ] kprove syntax in spec module
- [ ] macro on rules instead of syntax (loads)
- [ ] could not find main syntax module

ehildenb:
Some notes:

- Leave any warnings due by the Java backend alone (instead of fixing them) unless the fix is easy/trivial.
- Get incremental progress merged.

radumereuta:
First stage is done. regression-new is the first line of defense and now it's free of errors.
Moving this issue to backlog and we can come back to it when we're going to remove the java backend.

Baltoli:
Java backend is gone now.

PR #3014, opened 2022-10-21
Fix2316 warnings in ci

radumereuta:
Eliminate some of the warnings from CI
Partial fix for #2316
I've mainly focused on regression-new. I still have some unwanted warnings that I need to figure out how to fix.
pl-tutorial still has many warnings, but most come from the java backend.
Not sure how to deal with those.


kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -145,7 +148,7 @@ private void check(Production prod) {
     }
 
     private void check(Rule rule) {
-        if (rule.isMacro()) {
+        if (rule.isMacro() && !backend.equals(Backends.JAVA)) {
```
ehildenb: This isn't really a good solution here, the better solution is to migrate the parts of the tutorial that use the Java backend over to the Haskell backend. Would rather leave these warnings on, because they need to be fixed for those definitions to work on the haskell backend properly.

ehildenb: And this is for macros too, so we shoudl probably just fix the definitions.

radumereuta: Reverted that change.

radumereuta:
@ehildenb please have another look.
This PR focuses mostly on regression-new.
There is only one warning left coming from the java backend:
```
/regression-new/bisimulation/examples/program-equivalence/imp-lambda$ kprove-legacy sum-symbolic-spec.k -m VERIFICATION --no-exc-wrap  -d .
#Top
[Warning] Internal: execution phase: missing SMTLib translation for #KSequence . Please re-run with the --debug-z3 flag. Search the logs starting with 'Z3 warning' to see the Z3 implication that generated the warning. (SMTTranslationFailure: missing SMTLib translation for #KSequence)
```

k-distribution/pl-tutorial/2_languages/2_kool/1_untyped/kool-untyped.md
```diff
@@ -238,7 +238,7 @@ syntax includes:
 
 Old desugaring rules, from SIMPLE
 ```k
-  rule if (E) S => if (E) S else {}                                 [macro]
+  rule if (E) S => if (E) S else {}
   rule for(Start Cond; Step) {S} => {Start while (Cond) {S Step;}}  [macro]
   rule var E1::Exp, E2::Exp, Es::Exps; => var E1; var E2, Es;       [macro-rec]
```
ehildenb: It seems we should remove `macro` from the `for` rule as well?

k-distribution/tests/regression-new/kprove-branchingAllowed/Makefile
```diff
@@ -3,5 +3,6 @@ EXT=def
 KOMPILE_BACKEND=haskell
 TESTDIR=.
 KPROVE_FLAGS=--branching-allowed 1 --spec-module DEF-SPEC
+export KORE_EXEC_OPTS=--log-level error
```
ehildenb, line 6: So here are we basically just telling it to be quiet about warnings for this test?

k-distribution/pl-tutorial/2_languages/2_kool/1_untyped/kool-untyped.md
radumereuta: Removed it.
For this PR I focused more on `regression-new` since the changes to the pl-tutorial would get pretty involved.
For example in kool-untyped `var` becomes a bit trickier since macros should disappear after they are applied, but here we're using macros for unwrapping multiple declarations.
That would require a bit of a rewrite, checking comments, and I'm not sure I should be doing these changes since I'm not the owner of the definition.

k-distribution/tests/regression-new/kprove-branchingAllowed/Makefile
radumereuta: Yes, we used this before in several other tests.
I like tests that are focused on a specific problem.
Since the haskell-backend can give warnings with location information and all sorts of runtime information this option is very useful.

ehildenb (approved):

PR #3016, opened 2022-10-24
replace 'functional' attribute with 'function, total'

### LINKED ISSUE #3008: Rename `functional` into `total-function` for better clarity
The name `functional` is non-descriptive of what it actually means.
The description in the user manual is good enough, but during training people always had to ask what this means.

We could add the `total-function` attribute as an alias for a while and slowly phase out `functional`.
Also, we shouldn't require `function` when we add `total-function`. This will make some conditions in the code a bit more complex.

dwightguth:
Fixes #3008


docs/user_manual.md
```diff
@@ -435,10 +435,11 @@ appears anywhere in the configuration. Semantically, it means that evaluation
...
 
-The `functional` attribute indicates to the symbolic reasoning engine that a
-given symbol is a *total function*, that is it has *exactly* one return value
-for every possible input. Note that it does not make sense to have only the
-`functional` attribute without the `function` attribute.
+The `total` attribute can be added to a production with the `function`
+attribute to indicate to the symbolic reasoning engine that a given symbol is a
+*total function*, that is it has *exactly* one return value for every possible
+input. While the `total` attribute can in theory be added to something which
+is not a function, this behavior is not currently implemented.
```
ehildenb: We should instead say that `total` means *at least* one return value for every possible input. With the additional restriction currently in place that it must be on a `function` production, we still get the same behavior (exactly one value for each input), and we can say as much. But it leaves us open to defining `total` symbols later that have _more_ than one value possible (think disjuncts on the RHS of rules).

ehildenb: This also gives a nice symmetry with `function`, which means *at most* one return value for every possible input. And then `function, total` means *exactly* one return value per input.

ehildenb: I guess what you have written doesn't exactly contradict this, but I think it could be made clearer.

ehildenb:
Not sure if we already do this, but we should restrict that `total` can only be used on `function` productions for the moment as a check in the frontend if we're not doing this. Our user manual also says we only handle this case.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1451,7 +1451,7 @@ private boolean isGeneratedInKeysOp(Production prod) {
     }
 
     private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functionRules, Set<Production> overloads) {
-        boolean isFunctional = !isFunction(prod);
+        boolean isFunctional = !isFunction(prod) || prod.att().contains(Att.TOTAL());
```
radumereuta, line 1454: Maybe this variable should be called `isTotalFunction` now.

dwightguth: This is still an accurate name because that's what the attribubte is called at the kore level.

ehildenb (approved):

PR #3019, opened 2022-10-27
Fix the release process

ehildenb:
This partially reverts a change in #2950 that I think is causing the release to break. I don't have a good awy to test this other than just merging it.

Example breakage: https://github.com/runtimeverification/k/actions/runs/3325085433

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,16 @@ concurrency:
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


ehildenb:
It seems that we may not be able to compute release ID how we were trying to, so I reverted back to the old way we were doing it.

F-WRunTime (approved):

PR #3029, opened 2022-11-03
LHS of #Exists binds variables

### LINKED ISSUE #2709: [Bug] [kompile] - Compilation fails for simplifications with #Exists on right-hand side
K Version
---------

```
$ kompile --version
K version:    v5.3.98-5-g993674655
Build date:   Thu Jul 07 14:35:59 BST 2022
```

Description
-----------

Compilation fails for either of the following two simplification, which have the Matching-Logic `#Exists` on right-hand side:

```
rule { C #Equals { C }:>Int } => #Exists I1. { C #Equals I1:Int } [simplification]
rule { C #Equals { C }:>Int } => #Exists ?I1. { C #Equals ?I1:Int } [simplification]
```

Input Files
-----------

File `exists-simpl.k`, version 1:
```
module EXISTS-SIMPL-SYNTAX 
endmodule

module EXISTS-SIMPL

  rule { C #Equals { C }:>Int } => #Exists I1. { C #Equals I1:Int } [simplification, anywhere]

endmodule
```

File `exists-simpl.k`, version 2:
```
module EXISTS-SIMPL-SYNTAX 
endmodule

module EXISTS-SIMPL

  rule { C #Equals { C }:>Int } => #Exists ?I1. { C #Equals ?I1:Int } [simplification, anywhere]

endmodule
```

Reproduction Steps
------------------

Run: `kompile exists-simpl.k --backend haskell`

For version 1: 
```
[Error] Compiler: Cannot encode equations with existential variables to KORE.
 If this is desired, please use #Exists with regular variables.
 Offending variables: [?I1]
```

For version 2:
```
[Error] Compiler: Found variable I1 on right hand side of rule, not bound on
left hand side. Did you mean "?I1"?
        Source(/Users/petarmax/Projects/RV/exists-simpl.k)
        Location(6,44,6,46)
        6 |       rule { C #Equals { C }:>Int } => #Exists I1. { C #Equals I1:Int }
[simplification]
```          .                                

Expected Behavior
-----------------

At least one of the two simplifications being allowed by the compiler.

### PR DESCRIPTION
**Pull request by radumereuta**
Fixes: #2709
Just for reference, this is how the passing rule looks like at different stages:
```
orig:     rule { C #Equals { C }:>Int } => #Exists I1. { C #Equals I1:Int } [simplification]
parsed:   rule #Equals(#SemanticCastToK(C),`project:Int`(#SemanticCastToK(C)))=>#Exists(#SemanticCastToInt(I1),#Equals(#SemanticCastToK(C),#SemanticCastToInt(I1))) requires #token("true","Bool") ensures #token("true","Bool") [org.kframework.attributes.Location(Location(9,8,9,68)), org.kframework.attributes.Source(Source(/home/radu/work/k/k-distribution/tests/regression-new/checks/existsLHSBoundPass.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]), simplification]
compiled: rule #Equals(C,`project:Int`(C))=>#Exists(I1,#Equals(C,I1)) requires #token("true","Bool") ensures #token("true","Bool") [UNIQUE_ID(bf3584916f98d9712a09c2f26f0a1bdd55ff9c3f406b4d0c5c83ee09104a5dc0), org.kframework.attributes.Location(Location(9,8,9,68)), org.kframework.attributes.Source(Source(/home/radu/work/k/k-distribution/tests/regression-new/checks/existsLHSBoundPass.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]), simplification]
kore:
// rule #Equals{K,#SortParam}(C,inj{Int,KItem}(`project:Int`(C)))=>#Exists{Int,#SortParam}(I1,#Equals{K,#SortParam}(C,inj{Int,KItem}(I1))) requires #token("true","Bool") ensures #token("true","Bool") [UNIQUE_ID(bf3584916f98d9712a09c2f26f0a1bdd55ff9c3f406b4d0c5c83ee09104a5dc0), org.kframework.attributes.Location(Location(9,8,9,68)), org.kframework.attributes.Source(Source(/home/radu/work/k/k-distribution/tests/regression-new/checks/existsLHSBoundPass.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]), simplification, sortParams({Q0})]
  axiom{R,Q0} \implies{R} (
    \top{R}(),
    \equals{Q0,R} (
      \equals{SortK{}, Q0}(VarC:SortK{},kseq{}(inj{SortInt{}, SortKItem{}}(Lblproject'Coln'Int{}(VarC:SortK{})),dotk{}())),
     \and{Q0} (
       \exists{Q0}(VarI1:SortInt{},\equals{SortK{}, Q0}(VarC:SortK{},kseq{}(inj{SortInt{}, SortKItem{}}(VarI1:SortInt{}),dotk{}()))),
        \top{Q0}())))
  [sortParams{}("{Q0}"), org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/radu/work/k/k-distribution/tests/regression-new/checks/existsLHSBoundPass.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(9,8,9,68)"), simplification{}(), UNIQUE'Unds'ID{}("bf3584916f98d9712a09c2f26f0a1bdd55ff9c3f406b4d0c5c83ee09104a5dc0")]
```

@ehildenb please review
@PetarMax does this properly fix the issue you were having?

### CODE REVIEW

**Comment by PetarMax** [comment]
Yes, this fixes the issue.

**Comment by ehildenb** [comment]
This solutino looks like it won't work to me. We are just tracking if we are _inside_ an exists, not the actual variable the exists binds. Does it work correctly on the following code?

```
rule foo(X) => #Exists Y . { I #Equals Y }
```

In particular, this one should:

- Warn about unused variable `X`
- No error about variable `Y`.
- Error about variable `I`, it's a universal variable only on the RHS of a rule, so should be `?I` instead.

**Comment by ehildenb** [comment]
In particular, you need to track the `boundVars` set, instead of just tracking "am I inside an Exists/Forall".

so your code should check:

- Am I under an `Exists X . ...`, if so, then add `X` to the bound vars set for when recursing into the `...`. (Same for `Forall`).
- When checking variables (the `KVariable` case), if a variable is unbound, the same checks as before happen. But if a variable is bound, then it does not have the same restrictinos as before.

**Comment by radumereuta** [comment]
```
rule foo(X) => #Exists Y . { I #Equals Y }
[Warning] Compiler: Variable 'X' defined but not used. Prefix variable name with underscore if this is intentional.
	Location(18,12,18,13)
	18 |	  rule foo(X) => #Exists Y . { I #Equals Y }
	   .	           ^
[Error] Compiler: Found variable I on right hand side of rule, not bound on left hand side. Did you mean "?I"?
	Location(18,32,18,33)
	18 |	  rule foo(X) => #Exists Y . { I #Equals Y }
	   .	                               ^
[Warning] Compiler: Variable 'I' defined but not used. Prefix variable name with underscore if this is intentional.
	Location(18,32,18,33)
	18 |	  rule foo(X) => #Exists Y . { I #Equals Y }
	   .	                               ^
```
Added `?`
```
rule foo(X) => #Exists Y . { ?I #Equals Y }
[Warning] Compiler: Variable 'X' defined but not used. Prefix variable name with underscore if this is intentional.
	Location(18,12,18,13)
	18 |	  rule foo(X) => #Exists Y . { ?I #Equals Y }
	   .	           ^
[Warning] Compiler: Variable '?I' defined but not used. Prefix variable name with underscore if this is intentional.
	Location(18,32,18,34)
	18 |	  rule foo(X) => #Exists Y . { ?I #Equals Y }
	   .	                               ^~
```
If you look at the code, I modified `GatherVarsVisitor`. This tracks if we are in the LHS of a `#Forall` or `#Exists`, and if we find a variable, we add it to the bound set.
The original error message was thrown from `CheckRHSVariables` which relies on the visitor I just changed.

**Comment by ehildenb** [comment]
Add these as tests please. No reason not to.

**Comment by radumereuta** [comment]
Done.
First two were testing the llvm backend, the second two tests the haskell backend. One fail, one passing.

File: k-distribution/tests/regression-new/checkWarns/existsLHSBoundFail.k.out [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+[Error] Compiler: Found variable I on right hand side of rule, not bound on left hand side. Did you mean "?I"?
+	Source(existsLHSBoundFail.k)
+	Location(10,32,10,33)
+	10 |	  rule foo(X) => #Exists Y . { I #Equals Y }
+	   .	                               ^
+[Error] Compiler: Variable 'I' defined but not used. Prefix variable name with underscore if this is intentional.
+	Source(existsLHSBoundFail.k)
+	Location(10,32,10,33)
+	10 |	  rule foo(X) => #Exists Y . { I #Equals Y }
+	   .	                               ^
```
ehildenb, line 15: Why do we get two errors about `I`? The second one seems spurious to me.

ehildenb: I guess it's one error for the missing `?`, and one for the missing `_`.

kernel/src/main/java/org/kframework/compile/GatherVarsVisitor.java
```diff
@@ -14,6 +15,7 @@ public class GatherVarsVisitor extends RewriteAwareVisitor {
     private final Set<KVariable> vars;
     private final Set<KEMException> errors;
     private final boolean errorExistential;
+    private boolean isInExistsLhs = false;
```
ehildenb: Why is this called `isInExistsLhs` instead of `isInMLBinder`?

kernel/src/main/java/org/kframework/compile/GatherVarsVisitor.java
```diff
@@ -37,6 +41,14 @@ public void apply(KApply k) {
         if (k.klabel() instanceof KVariable) {
             apply((KVariable) k.klabel());
         }
+        if (k.klabel().equals(KLabels.ML_EXISTS) || k.klabel().equals(KLabels.ML_FORALL)) {
+            boolean tmp = isInExistsLhs;
+            isInExistsLhs = true;
+            apply(k.items().get(0));
```
ehildenb, line 47: Why are we triggering re-traversal of the variable part of the pattern? I would think that we could just get it, assert that it's a variable (indeed, it's not a well-formed pattern if it's not), and then extract it directly. No need to recurse on that subterm.

I think that recursing on th esubterm is also why you have to have the `tmp` variable for `isInExistsLhs`, which makes this more confusing.

ehildenb: It also makes it more confusing because then `KVariable` case needs to be adjusted. I think you can do without adjusting that case if you change the code to say something like this (psuedocode):

```suggestion
    binderVar = k.items().get(0);
    if type(binderVar) is not KVariable {
        errors.add(KEMException.compilerError("Found non-variable in binder position of {LABEL}");
    }
    vars.add(binderVar);
```

Then we don't need the variable `isInExistsLhs`, and don't need to modify `KVariable` case.

kernel/src/main/java/org/kframework/compile/GatherVarsVisitor.java
radumereuta: Fixed.

kernel/src/main/java/org/kframework/compile/GatherVarsVisitor.java
radumereuta: You can have casts and injections. And maybe brackets.
It's a similar pattern we use in [ComputeUnboundVariables](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/checks/ComputeUnboundVariables.java#L69)
If you write something like `  rule foo(X) => #Exists 1 . { ?I #Equals 1 }` then the haskell-backend complains.
```
kore-exec: [16411] Error (ErrorParse):
    definition.kore:725:133: unexpected token TokenDomainValue
```
Not the pretties error message, but kind of gets out of scope for this issue.

ehildenb: I don't think it's out of scope for this issue at all, this issue is about handling Exists and Forall binding correctly.

I also just think that the implementation I propose above is cleaner honestly, and easier to understand. It took me a larger chunk of time to understand your code, with the temp variable being stored and unset and such. I don't think the check is actually complicated, but the way you implemented it made it seem so.

ehildenb: Is it actually reasonable to have a cast/injection/bracket on the first argument here? To me it seems like not at first glance. But perhaps it makes sense?

ehildenb (approved):

kernel/src/main/java/org/kframework/compile/GatherVarsVisitor.java
radumereuta: > I don't think it's out of scope for this issue at all, this issue is about handling Exists and Forall binding correctly.

Fair point.
```
rule: rule 1 => #Exists _Y . 2
parsed: rule #token("1","Int")=>#Exists(#SemanticCastToK(_Y),#token("2","Int")) requires #token("true","Bool") ensures #token("true","Bool") [org.kframework.attributes.Location(Location(18,8,18,27)), org.kframework.attributes.Source(Source(/home/radu/work/test/test.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol])]
```
Casts will appear under there. They are added by the type inferencer even if you don't provide one.
I'm not sure about injections since that's a free parameter so the type inferencer can choose something convenient.
Brackets probably not since we remove them right after parsing.

But in any case, this type of traversal has proven to be the most flexible and reliable for the front-end. If we decide in the future that we want to support a list of variables `#Exists A, B, C . 2` then this will still work correctly.

Anyway, thanks for the approval.

Issue #2774, opened 2022-08-04
`comm` attribute is being ignored on llvm and Haskell backends


radumereuta:
- [x] change the front end - @radumereuta 
- [ ] change the Haskell Backend - @ana-pantilie 
- [ ] change the LLVM Backend - @dwightguth 

Simplification rules annotated with the `comm` attribute in the Haskell backend are being [ignored](https://github.com/runtimeverification/haskell-backend/blob/b06757e252ee01fdd5ab8f07de2910711997d845/kore/src/Kore/Equation/Registry.hs#L125-L142)
Likely the LLVM backend does the same.

This is an issue because of https://github.com/runtimeverification/k/pull/2754
As a workaround, we removed the `comm` attribute when we generated those rules.

This needs to be investigated by the Haskell and the LLVM teams. @ana-pantilie @theo25

dwightguth:
I mean, if we wanted to we could probably just not emit those axioms in the frontend since they are not, as I said, being used. I believe the original intent was for them to be used during proof object generation, but that's a good ways away from being at all relevant.

Alternately, we could just give the old use of the comm attribute a different name, like `commutativity` or something.

radumereuta:
Right now the front end generates only two of those axioms: for `_Set_` and for `_Map_`, but as you say, if they're not being used, we could use a different keyword and disable this code path. But let's see how this PR behaves further.

radumereuta:
We should try to use `commutativity` as Dwight suggested and change the front end and the backends together so we don't miss functionality. We should try to keep the `comm` attribute on the new productions and simplification rules.

@dwightguth please point out where in the LLVM backend we need to make changes.

dwightguth:
LLVM backend needs to be changed in two places:

* https://github.com/runtimeverification/llvm-backend/blob/master/matching/src/main/scala/org/kframework/backend/llvm/matching/Parser.scala#L141
* https://github.com/runtimeverification/llvm-backend/blob/master/lib/ast/AST.cpp#L1185

This is not a hard change; I would suggest whoever is working on the frontend also do the backends.

@ana-pantilie I suggest sharing where in the Haskell backend this happens also.

ana-pantilie:
[...]

PR #3032, opened 2022-11-07
disable comm axiom generation

dwightguth:
Part of #2774 

These axioms are, to my knowledge, not used by either backend, so we simply disable them. I left the code that generated them in case we decide we want to bring them back later.


radumereuta (approved):

dwightguth:
This failed with a ClassFormatError which appears to happen if java bytecode that was generated by the compiler is malformed or corrupted. I don't think it has anything to do with this change; it probably has something to do with the M1 environment it ran on.

PR #3033, opened 2022-11-07
Improve GH workflows

radumereuta:
- rename master-pr readable name to be more intuitive
- remove testing trigger on PR edit

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -1,7 +1,7 @@
 name: 'Test PR'
 on:
   pull_request:
-    types: [opened, edited, reopened, synchronize]
+    types: [opened, reopened, synchronize]
     branches:
       - 'develop'
 concurrency:
--- a/.github/workflows/master-pr.yml
+++ b/.github/workflows/master-pr.yml
@@ -1,4 +1,4 @@
-name: 'Test PR'
+name: 'Switch master PRs to develop'
 on:
   pull_request:
     branches:


.github/workflows/test-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: 'Test PR'
 on:
   pull_request:
-    types: [opened, edited, reopened, synchronize]
+    types: [opened, reopened, synchronize]
```
radumereuta, line 4: Will this still trigger when the switch from master to develop happens?
It could be interpreted as a PR edit.

ehildenb: It appears that we really just need `synchronize`, which is: "Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.".

Docs here: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-37

So I think your cahnge is correct.

ehildenb (approved):

.github/workflows/test-pr.yml
ehildenb: Ah, but it looks like it isn't actually triggering.....

Hmmmm.

F-WRunTime: TLDR no, I think changing this will break causing a trigger on branch change. If I recall correctly (few weeks ago now) I added this for the specific purpose of making sure triggers occurred post a change to the target merge branch as an edit to the PR 'issue'. Edit should also include changing the description, labels, title

Is there a driving reason for changing this line? Was it causing a problem?

F-WRunTime (changes requested):

.github/workflows/test-pr.yml
radumereuta: I changed the description of the first comment in a PR and it triggered a new build.
That shouldn't happen since I didn't include any new changes to the code.
CI is kind of slow right now since it's testing way too much, and we're not caching enough of the build.
This PR is low priority though.

PR #3036, opened 2022-11-08
Automatically import BOOL when strictness is applied

### LINKED ISSUE #2564: [Bug] [kompile] - Implicit import of the module BOOL
K Version
---------

```
$ kompile --version
```
K version:    5.2.110
Build date:   Mon Mar 21 00:10:55 CET 2022

Description
-----------
The command `kompile lazy.k` returns the error:
``` 
file a bug report at https://github.com/runtimeverification/k/issues
(NoSuchElementException: key not found: _andBool_)
```

Input Files
-----------

The file `lazy.k` is:
```
module LAZY-SYNTAX
    // imports BOOL
    syntax MyBool ::= "mytrue" | "myfalse"
    syntax BExp ::= MyBool | BExp "myand" BExp [strict(1)]

    syntax KResult ::= MyBool

endmodule

module LAZY
    imports LAZY-SYNTAX

    rule mytrue  myand B2 => B2
    rule myfalse myand _  => myfalse

endmodule
```

Expected Behavior
-----------------

The command `kompile lazy.k` should return nothing.

To fix the problem, you can uncomment the second line (`imports BOOL`).

radumereuta:
Fixes: #2564
This is done for the compiled definition so we don't taint program parsing.

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkStrictBOOLInclusion.k
+++ b/k-distribution/tests/regression-new/checks/checkStrictBOOLInclusion.k
@@ -1,0 +1,17 @@
+module CHECKSTRICTBOOLINCLUSION-SYNTAX
+// https://github.com/runtimeverification/k/issues/2564
+// imports BOOL
+    syntax MyBool ::= "mytrue" | "myfalse"
+    syntax BExp ::= MyBool | BExp "myand" BExp [strict]
+
+    syntax KResult ::= MyBool
+
+endmodule
+
+module CHECKSTRICTBOOLINCLUSION
+    imports CHECKSTRICTBOOLINCLUSION-SYNTAX
+
+    rule mytrue  myand B2 => B2
+    rule myfalse myand _  => myfalse
+
+endmodule
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -1623,7 +1623,7 @@ Evaluation Strategy
 ### `strict` and `seqstrict` attributes
 
 The strictness attributes allow defining evaluation strategies without having
-to explicitely make rules which implement them. This is done by injecting
+to explicitly make rules which implement them. This is done by injecting
 *heating* and *cooling* rules for the subterms. For this to work, you need to
 define what a *result* is for K, by extending the  `KResult` sort.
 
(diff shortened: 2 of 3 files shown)


Baltoli (approved):

dwightguth (dismissed):
We should generally be very careful about automatically importing things.

Before we merge this, please check whether importing BOOL into the main module fixes this issue; if it does, we should probably just fix the error message.

ehildenb:
@dwightguth you can't even write a `requires` clause without `BOOL` being imported. Do we implicitly import bool in other contexts? What about `Map` module for configurations?

I think we should likely follow whatever we're doing elsewhere. As long as it's documented well, we should be OK with it. But I guess it should be disabled if the user wants too (eg. `--no-implicit-imports` or extend `--no-prelude`).

ehildenb:
The 95% case is that users want to have BOOL implicitly imported, but we should have an escape hatch (credit to Bruce). I think I agree with this.

dwightguth:
Sounds like the consensus is that this should just work, so I'll rescind my review.

k-distribution/tests/regression-new/checks/checkStrictBOOLInclusion.k
```diff
@@ -0,0 +1,17 @@
+module CHECKSTRICTBOOLINCLUSION-SYNTAX
+// https://github.com/runtimeverification/k/issues/2564
+// imports BOOL
```
ehildenb, line 3:
```suggestion
   imports private BOOL
```

ehildenb: Would this work? Is this a workaround?

radumereuta:
A summary:
1. crashing with `key not found: _andBool_` when we add strict inside the syntax module is clearly not acceptable.
2. manually adding `imports private BOOL` won't work because it's going to taint the program syntax.
 ```console
$ kast -e 'true andBool false'
`_andBool_`(#token("true","Bool"),#token("false","Bool"))
```
3. `ResolveStrict` generates the rules in place. So importing `BOOL` only in the main module won't fix the problem.
4. we have precedence with implicit imports when parsing the configuration. We [import MAP](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java#L282).
5. right now it's failing in the _95% of cases_. Let's have a _5% case_ and see how that workaround should behave. I'm importing the module by name, so if you define your own `domains.md` with your own `BOOL`, then you can override the contents.

One thing I agree with is that documentation should be updated, so I pushed another commit. Please re-review.

k-distribution/tests/regression-new/checks/checkStrictBOOLInclusion.k
radumereuta: Tested, doesn't work, see my latest comment.

ehildenb (approved):

Baltoli (approved):

PR #3045, opened 2022-11-15
Setting up GH workflow for Debian packaging

F-WRunTime:


ehildenb:
You can also remove `package/debian/Dockerfile` after this change I believe.

.github/workflows/Dockerfile
```diff
@@ -1,17 +1,19 @@
+ARG BASE_OS
+ARG BASE_DISTRO
 ARG BASE_IMAGE
 ARG K_VERSION
 ARG Z3_VERSION
-FROM z3:${BASE_IMAGE}-${Z3_VERSION} as Z3
+FROM z3:${BASE_DISTRO}-${Z3_VERSION} as Z3
 
-ARG BASE_IMAGE
+ARG BASE_OS
```
ehildenb, line 7:
```suggestion
ARG BASE_OS
ARG BASE_DISTRO
```

You have to repeat all arguments before every `FROM`.

.github/workflows/Dockerfile
```diff
@@ -1,17 +1,19 @@
...
 ARG K_VERSION
-FROM stack:${BASE_IMAGE}-${K_VERSION} as STACK
+FROM stack:${BASE_DISTRO}-${K_VERSION} as STACK
 
-ARG BASE_IMAGE
+ARG BASE_OS
```
ehildenb, line 12: Same here, you need to redeclare `BASE_DISTRO` variable again.

F-WRunTime: Strange Nuance I have not run into before.

ehildenb:
Evidence this works: https://github.com/runtimeverification/k/actions/runs/3484146828/jobs/5828379649

ehildenb (approved):

PR #3047, opened 2022-11-17
Migrate Arch packaging to GH runners

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/pull/3045~

- Parameterize the `test-package` action in the OS and the DISTRO, so that it works for non-debian based distros.
- Allow overriding the dockerfiles to use for `test-package` action.
- Add necessary Dockerfiles in `package/arch` for the `test-package` action to work with it.
- Move the scripts `src/main/scripts/test-in-container*` to `package/test-package` and `package/debian/test-package` scripts.
- Pass in the `test-package` and `build-package` script as parameters to `test-package` github action, so that it can be different for different OSs.
- Remove the Arch packaging from the Jenkins job.


ehildenb:
It can't find where the packages are moved to, with teh `pkg-name` variable: 
image

Same problem on both Jammy and Arch, so it's something to do with the changes to `test-package` and `with-docker`. Teh `build-package` scripts are supposed to move the packages into place.

F-WRunTime:
> It can't find where the packages are moved to, with teh `pkg-name` variable: image
> 
> Same problem on both Jammy and Arch, so it's something to do with the changes to `test-package` and `with-docker`. Teh `build-package` scripts are supposed to move the packages into place.

Curious. If I recall the package we want is the one coming out of the test-pr job? It could be it's running on two different runners and not being shared as expected via the filesystem? 

Could we look to utilize the upload-artifact and download-artifact which are temporary storage and usable between jobs. 
Upload: https://github.com/marketplace/actions/upload-a-build-artifact
Download: https://github.com/marketplace/actions/download-a-build-artifact

ehildenb:
No they're not shared between workflows, it's all in the same workflow.

.github/actions/test-package/action.yml
```diff
@@ -10,43 +11,64 @@ inputs:
...
+      BUILD_PACKAGE: ${{ inputs.build-package }}
+      PKG_NAME: ${{ inputs.pkg-name }}
     run: |
       set -euxo pipefail
-      docker exec -t k-package-build-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c 'mv package/debian debian'
-      docker exec -t k-package-build-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "mv debian/compat.${BASE_DISTRO} debian/compat"
-      docker exec -t k-package-build-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "mv debian/control.${BASE_DISTRO} debian/control"
-      docker exec -t k-package-build-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "mv debian/rules.${BASE_DISTRO} debian/rules"
-      docker exec -t k-package-build-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c 'dpkg-buildpackage'
+      docker exec -t k-package-build-${BASE_OS}-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "${BUILD_PACKAGE} ${PKG_NAME}"
```
ehildenb, line 59: Here we call the `BUILD_PACKAGE` script, which we assume moves the created package to `PKG_NAME` location.

.github/actions/test-package/action.yml
```diff
@@ -55,32 +77,33 @@ runs:
...
     shell: bash {0}
     env:
+      BASE_OS: ${{ inputs.os }}
       BASE_DISTRO: ${{ inputs.distro }}
+      TEST_PACKAGE: ${{ inputs.test-package }}
+      PKG_NAME: ${{ inputs.pkg-name }}
     run: |
       set -euxo pipefail
-      docker exec -t k-package-test-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "k-${BASE_DISTRO}/src/main/scripts/test-in-container-debian"
+      docker exec -t k-package-test-${BASE_OS}-${BASE_DISTRO}-${GITHUB_SHA} /bin/bash -c "${TEST_PACKAGE} ./${PKG_NAME}"
```
ehildenb: Then here we call `TEST_PACKAGE`, which we tell where the package has been moved to.

.github/workflows/test-pr.yml
```diff
@@ -30,6 +30,30 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: 'Build and Test Package'
+        uses: ./.github/actions/test-package
+        with:
+          dockerfile: package/arch/Dockerfile
+          os: archlinux
+          distro: base
+          build-package: package/arch/build-package
+          test-package: package/arch/test-package
+          pkg-name: kframework_arch_x86_64.pkg.tar.zst
```
ehildenb: Here we tell it where the package should be moved to when calling the `test-package` action (where `BUILD_PACKAGE` should move it to, and `TEST_PACKAGE` should assume it's at).

.github/workflows/test-pr.yml
```diff
@@ -30,6 +30,30 @@ jobs:
...
+      - name: 'Upload Package to Release'
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.JENKINS_GITHUB_PAT }}
+        with:
+          asset_content_type: 'application/gzip'
+          asset_name: 'K Framework Arch Package'
+          asset_path: kframework_arch_x86_64.pkg.tar.zst
```
ehildenb: And here we   do the asset upload (so this needs to match the prior step).

F-WRunTime: Hmm interesting my first thought is to check the cwd during the execution and maybe that's not what we expected. 
My brain is a bit mush at the moment, I can come back tomorrow morning and give it a run through manually to see if something jumps out at me.

ehildenb:
Evidence this works: https://github.com/runtimeverification/k/actions/runs/3597881818/jobs/6060106730

Jenkinsfile
```diff
@@ -35,68 +35,6 @@ pipeline {
         stash name: 'src', includes: "kframework-${env.VERSION}-src.tar.gz"
```
F-WRunTime, line 35: Was the source tar piece of this build process being primarily used as input to the the following steps? Is the rest of the Jenkinsfile still have value to keep?

F-WRunTime (approved):

Jenkinsfile
ehildenb: No, this is mostly some legacy thing, I don't think it's actually used anymore. Maybe in building the brew package?

PR #3048, opened 2022-11-17
Smoke tests2

radumereuta:


radumereuta:
This gets mounted as a volume: https://github.com/runtimeverification/k/blob/master/.github/actions/test-package/action.yml#L63
and the user sees them as external files.
Try to use `docker cp`.
Care: if you copy from host to container as root you take ownership as root, not as user. So take care.

ehildenb:
Closing for now, out of date, re-open if work resumes.

PR #3058, opened 2022-11-23
Implement python bindings option in `kompile`

### LINKED ISSUE #3056: Expose Python bindings compilation in `kompile`
Currently, to generate the Python bindings for the LLVM backend, you need to call into `llvm-kompile` directly. Ideally, `kompile` would expose an option like `--llvm-kompile-mode` that would allow the higher-level tool to use this feature.

Baltoli:
Fixes #3056

This PR allows for Python bindings to be directly generated by `kompile`, rather than via `llvm-kompile`.

### DIFF
--- a/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMBackend.java
+++ b/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMBackend.java
@@ -98,7 +98,7 @@ public class LLVMBackend extends KoreBackend {
                 llvmType = options.llvmKompileType;
                 break;
             default:
-                throw KEMException.criticalError("Non-valid argument for --llvm-kompile-type: " + options.llvmKompileType + ". Expected [main|search|library|static|python]");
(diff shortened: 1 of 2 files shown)


llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMBackend.java
```diff
@@ -86,22 +86,30 @@ public void accept(Backend.Holder h) {
...
             case "main":
             case "search":
-            case "library":
             case "static":
+            case "python":
                 llvmType = options.llvmKompileType;
                 break;
             default:
-                throw KEMException.criticalError("Non-valid argument for --llvm-kompile-type: " + options.llvmKompileType + ". Expected [main|search|library|static]");
+                throw KEMException.criticalError("Non-valid argument for --llvm-kompile-type: " + options.llvmKompileType + ". Expected [main|search|library|static|python]");
```
tothtamas28, line 102:
```suggestion
                throw KEMException.criticalError("Non-valid argument for --llvm-kompile-type: " + options.llvmKompileType + ". Expected [main|search|static|python]");
```

tothtamas28 (approved):

PR #3063, opened 2022-11-24
Better labels for heat/cool rules of a context

h0nzZik:
Currently, when a context declaration has a label, both derived rules (heating and cooling) share this label.
This is unfortunate because one cannot refer to the two rules separately, e.g. in `kore-repl`.
I propose that we append `-heat` and `-cool` to the labels of the two rules, respectively.

This PR is currently missing automated tests; I will create them if this idea (not necessarily the code) gets informally accepted.


radumereuta:
@h0nzZik Please rebase your changes onto the latest `develop` branch.
For the K repo, the workflow is like this:
- PRs should be opened against `develop` and when preliminary tests pass they get merged.
- after all test pass, we automatically bump the K version and merge into `master`.

radumereuta:
@h0nzZik why is this sill a draft?
What needs to be done to merge this?

h0nzZik:
@radumereuta It is a draft because I wanted to add some tests to this. I will mark it ready for review now, though.

Robertorosmaninho:
Hey @h0nzZik, could you please add some tests to this PR? Also, change its name as it's already ready for review 😄

radumereuta:
> Hey @h0nzZik, could you please add some tests to this PR? Also, change its name as it's already ready for review smile

@Robertorosmaninho I think you'll have to add the tests yourself.
Jan is not a K developer, you have more knowledge about adding regression tests.
Please tag me as a reviewer when this is done.

Robertorosmaninho:
Unfortunately, this modification doesn't seem to affect the rule.label production at all. I've tried all examples from [lesson 1.14](https://kframework.org/k-distribution/k-tutorial/1_basic/14_evaluation_order/) with `klabel{}("and")` and `label{}("plus")`. 

Both as attribute after `seqstrict ` and `strict` or with `rule[label]:`

radumereuta:
I tried this and worked:
```
context [abcd]: HOLE + HOLE
rule `<generatedTop>`(`<k>`(``HOLE~>`#freezer_+__TEST_Exp_Exp_Exp0_`(.KList)=>`_+__TEST_Exp_Exp_Exp`(HOLE,HOLE)``~>_DotVar1),_DotVar0) requires `_andBool_`(#token("true","Bool"),isKResult(HOLE)) ensures #token("true","Bool") [UNIQUE_ID(43abab7d45de3053ea6994842dc937c4ce4db854fdb3ba610ef18bb25effd25a), cool, cool-like, label(TEST.abcd-cool), org.kframework.attributes.Location(Location(13,19,13,30)), org.kframework.attributes.Source(Source(/home/radu/work/test/test.k)), org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol])]
```
That's what this PR does. I'm not sure though, it's all we want.

Robertorosmaninho:
Thanks, @radumereuta! 
I think that we want something similar to what you posted [here](https://github.com/runtimeverification/k/issues/3180#issue-1594960773). The most valuable case would be with `strict` and `seqstrict`. This can probably be addressed with `label()` or `klabel()`, but I couldn't get these attributes from [ResolveContexts.java](https://github.com/runtimeverification/k/pull/3063/files#diff-0b824d4b1f2e6d4c0ec9745745b810357d0519d331853a2ff434f6c9fafe2a56)

radumereuta:
We talked about this but we should have it written.
We also need to generate the labels in this step: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/ResolveStrict.java

gtrepta:
@radumereuta @Robertorosmaninho I'm looking at this right now.

For strict/seqstrict attributes, we need to number the klabel (if it exists) of the production and add it to the generated Context?

Robertorosmaninho:
> For strict/seqstrict attributes, we need to number the klabel (if it exists) of the production and add it to the generated Context?

That's an idea, but I think that the right way to do that is to generate label rules within these suffixes.

radumereuta:
> > For strict/seqstrict attributes, we need to number the klabel (if it exists) of the production and add it to the generated Context?
> 
> That's an idea, but I think that the right way to do that is to generate label rules within these suffixes.

I think I agree with @Robertorosmaninho
For the first implementation, let's try it w/o numbers. Just overload the label.
If I want traceability during debugging, I think I'm fine with breakpoints on all contexts.

Robertorosmaninho:
The main idea to get it on rule-labels is to enable this rules to be find by `k-rule-find` and `k-rule-apply`

gtrepta:
> For the first implementation, let's try it w/o numbers. Just overload the label.

@dwightguth will this be ok? It sounded like rule labels need to be unique based on our last conversation.

> The main idea to get it on rule-labels is to enable this rules to be find by `k-rule-find` and `k-rule-apply`

Side-note: It looks like k-rule-find will only print one rule label per source line. Testing on a context line for example only gave me the heating label. (Looks like there's an issue for this https://github.com/runtimeverification/llvm-backend/issues/725 )

Robertorosmaninho:
> Side-note: It looks like `k-rule-find` will only print one rule label per source line. Testing on a context line for example only gave me the heating label. (Looks like there's an issue for this https://github.com/runtimeverification/llvm-backend/issues/725 )

You're right! I filled this issue with this in mind! As the current implementation guarantees a unique rule label, I didn't worry much then, but as soon as this PR gets merged and multiple rules with the same label become possible, the tool will need to be updated to return a list of rules!

dwightguth:
Rule labels are meant to be completely unique. Stuff is written with that assumption in mind and stuff will break, not just debugging, if that's no longer true. You need a number if a production is strict in more than one position. Period.

dwightguth:
This looks okay, but if you have multiple sentences that all trigger the error, the compiler will only report one of them at a time, which is less than ideal. Not sure how easy it would be to fix but if you do then CheckAtt is the place to do it.

k-distribution/tests/regression-new/context-labels/context-spec.k.out
```diff
@@ -0,0 +1,638 @@
...
+      \and{SortGeneratedTopCell{}}(
+          /* predicate: */
+          /* D Sfa */ \top{SortGeneratedTopCell{}}(),
+          /* substitution: */
+          \top{SortGeneratedTopCell{}}()
+      ))
+Context:
+    (InfoReachability) while applying axioms
+kore-repl:  Debug (DebugAppliedLabeledRewriteRule):
+    The rule with label [  ] at location: context.k:15:8-15:14 was applied
```
radumereuta, line 37: What rule is this? With no label and on line `context.k:15`. The file only has 14 lines in total.

k-distribution/tests/regression-new/context-labels/context-spec.k.out
```diff
@@ -0,0 +1,638 @@
...
```
radumereuta, line 422: Same question here.

radumereuta:
> This looks okay, but if you have multiple sentences that all trigger the error, the compiler will only report one of them at a time, which is less than ideal. Not sure how easy it would be to fix but if you do then CheckAtt is the place to do it.

I think I prefer it this way. The problem is with the syntax declaration.
You fix that, the errors will disappear from all the rules.
I don't want to receive an error message for every KApp.

But if you move it to CheckAtt for productions, then you get the error message much sooner, and only once per syntax, and all at once. Might be worth implementing that way then.

k-distribution/tests/regression-new/context-labels/context-spec.k.out
gtrepta: Line 15 is `rule a => 0`. I don't know why kore-repl's `DebugAppliedLabeledRewriteRule` tries to display this one.

radumereuta (approved):

radumereuta:
Please add a check to see if `k-rule-find` works after this change.

k-distribution/tests/regression-new/checks/duplicateContextLabels.k.out
```diff
@@ -0,0 +1,9 @@
+[Error] Compiler: The generated label for a context rule conflicts with a user-defined label. Please consider renaming. (KEMException: [Error] Compiler:
```
radumereuta: What happened here?
Why does the error message end like that?

k-distribution/tests/regression-new/checks/duplicateContextLabels.k.out
```diff
@@ -0,0 +1,9 @@
+[Error] Compiler: The generated label for a context rule conflicts with a user-defined label. Please consider renaming. (KEMException: [Error] Compiler: 
+	Source(duplicateContextLabels.k)
+	Location(9,24,9,41)
+	9 |	  context [foo-left]:  foo( HOLE,    _ )
+	  .	                       ^~~~~~~~~~~~~~~~~)
```
radumereuta: Why do we have a closed parenthesis at the end here?

k-distribution/tests/regression-new/context-labels/context-spec.k.out
radumereuta: Ah, I see, it's unrelated.

radumereuta (commented):
Restarting to see if tests pass now that CI should be working again.

k-distribution/tests/regression-new/checks/duplicateContextLabels.k.out
gtrepta: I wanted both of the source locations involved in the conflict, so I passed an exception with one location into another. Is there a better way to do this?

https://github.com/runtimeverification/k/blob/bf57cb07ae65c62b613c05709e64185c039a5ed1/kernel/src/main/java/org/kframework/compile/ResolveContexts.java#L218

k-distribution/tests/regression-new/checks/duplicateContextLabels.k.out
gtrepta: See above.

k-distribution/tests/regression-new/checks/duplicateContextLabels.k.out
radumereuta: We have such cases. Take a look at this error: https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/checks/checkDuplicateList.k.out
You can search the message through the code.

radumereuta (approved):

radumereuta:
The last run failed because the runner was out of space.
I triggered an action to clean it: https://github.com/runtimeverification/devops-actions/actions/runs/5788235263
Please restart the build when that finishes.

radumereuta:
@gtrepta this failed with "k-rule-find: command not found" this time.

PR #3074, opened 2022-12-05
Fix debian releases

ehildenb:
- Only publishes a Jammy docker image.
- Makes sure that package for debian distros is put back into place after testing.


ehildenb:
Evidence this works: https://github.com/runtimeverification/k/actions/runs/3625463662/jobs/6113542968

F-WRunTime (approved):
Bye Focal!

F-WRunTime:
Working here for some reason. Trying to sync the two workflows to find what's different root cause of failure. 
https://github.com/runtimeverification/haskell-backend/actions/runs/3641759586/jobs/6148081508

F-WRunTime:
The piece 
```bash 
          install_url: https://releases.nixos.org/nix/nix-2.7.0/install
```
May require to be pushed into master for it to take effect on this branch or we need to push workflow_dispatch for this workflow for the branch specific changes to the workflow to take effect. 

@ehildenb Thoughts?

Issue #3070, opened 2022-12-01
LSP for K


ehildenb:
Language Server Protocol (LSP) is defined here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/

The features we want to implement with LSP are:

- [x] Tab/token completion.
- [x] Goto definition.
- [x] Bad-parse underlining.

Here is an existing Python package for LSP: https://pygls.readthedocs.io/en/latest/

We should be able to instantiate this to tab/token completion directly, by reading in compiled.json and taking in all the tokens that should be considered for tab completion.

radumereuta:
First version should take as input the compiled.json and try to implement Tab token completion.
Scan the definition, collect all the tokens and send them through the LSP

radumereuta:
The features described here have been implemented and added to the K repo.
More features related to the Language Server have been added here: https://github.com/runtimeverification/k-editor-support/issues/39

PR #3104, opened 2023-01-10
Lsp

radumereuta:
Part of #3070

Adds LSP support for:
- code completion for the current file and domains.md and kast.md (doesn't recurse into required files yet)
- highlights error messages directly in the editor for outer parsing (one second delay)

This can be tested with VSCode with this branch: https://github.com/runtimeverification/k-editor-support/tree/kLanguageClient/k-vscode

General idea:
1. a language client (tested with VSCode) will have to call the script `klsp`
2. `klsp` starts the server in a similar way we do for kompile. Meaning we have direct access to all the parsing and kompile steps infrastructure.
3. the client sends requests for code completion or diagnostics, and the server responds in the appropriate manner.

To set things up, I followed this tutorial: https://medium.com/ballerina-techblog/practical-guide-for-the-language-server-protocol-3091a122b750

For now, I only implemented the minimum necessary for code completion and diagnostics for outer parsing errors.
More features can be added easily: workspace diagnostics and code completion that follows recursive inclusions.

### DIFF
--- a/kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
+++ b/kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
@@ -3,9 +3,19 @@ package org.kframework.lsp;
 import org.eclipse.lsp4j.*;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.services.TextDocumentService;
+import org.kframework.attributes.Source;
+import org.kframework.kil.DefinitionItem;
+import org.kframework.parser.outer.Outer;
+import org.kframework.utils.StringUtil;
+import org.kframework.utils.file.FileUtil;
 
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -16,6 +26,8 @@ public class KTextDocumentService implements TextDocumentService {
     private KLanguageServer languageServer;
     private LSClientLogger clientLogger;
 
+    private Map<String, List<DefinitionItem>> elems = new HashMap<>();
+
     public KTextDocumentService(KLanguageServer languageServer) {
         this.languageServer = languageServer;
         this.clientLogger = LSClientLogger.getInstance();
@@ -25,6 +37,7 @@ public class KTextDocumentService implements TextDocumentService {
     public void didOpen(DidOpenTextDocumentParams didOpenTextDocumentParams) {
         this.clientLogger.logMessage("Operation '" + "text/didOpen" +
                 "' {fileUri: '" + didOpenTextDocumentParams.getTextDocument().getUri() + "'} opened");
+        outerParse(didOpenTextDocumentParams.getTextDocument().getUri());
     }
 
     @Override
@@ -43,6 +56,7 @@ public class KTextDocumentService implements TextDocumentService {
     public void didSave(DidSaveTextDocumentParams didSaveTextDocumentParams) {
         this.clientLogger.logMessage("Operation '" + "text/didSave" +
                 "' {fileUri: '" + didSaveTextDocumentParams.getTextDocument().getUri() + "'} Saved");
+        outerParse(didSaveTextDocumentParams.getTextDocument().getUri());
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
```diff
@@ -0,0 +1,194 @@
...
+                                                        doc += s.getDeclaredSort() + " ::= ";
+                                                        doc += p.toString();
+                                                        completionItem.setDocumentation(doc);
+                                                        completionItem.setKind(CompletionItemKind.Snippet);
+                                                        lci.add(completionItem);
+                                                    }
+                                                }))))));
+
+        return lci;
+    }
```
Baltoli: Can we break this code down somewhat? It's a bit tricky to follow that many nested stream operations

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
```diff
@@ -0,0 +1,194 @@
...
+
+    // for quick testing
+    public static void main(String[] args) throws InterruptedException, ExecutionException, URISyntaxException {
+        String uri = "file:///home/radu/work/test/test.k";
+        List<DefinitionItem> dis = Outer.parse(Source.apply(uri), FileUtil.load(new File(new URI(uri))), null);
+        List<DefinitionItem> doma = Outer.parse(Source.apply(domains.toString()), FileUtil.load(new File(domains)), null);
+        List<DefinitionItem> kst = Outer.parse(Source.apply(kast.toString()), FileUtil.load(new File(kast)), null);
+        List<CompletionItem> x = getCompletionItems(dis);
+        System.out.println(x.size());
+    }
```
Baltoli: Does this need to be removed?

Baltoli (changes requested):
A couple of comments on the code itself; I'm going to fire this up with my own editor and test the actual functionality now.

Baltoli:
Seems to work in my vscode once I've jumped through the hoops to install it.

goodlyrottenapple:
are we adding a klsp-vscode extension? if so will that be a separate PR here or its own repo?

radumereuta:
> are we adding a klsp-vscode extension? if so will that be a separate PR here or its own repo?

Yes, we need to update the current k-vscode extension to tell it how to start the `klsp`.
We have a repo dedicated to all things editor support [here](https://github.com/runtimeverification/k-editor-support/tree/kLanguageClient/k-vscode). I've made a draft PR with the necessary changes [here](https://github.com/runtimeverification/k-editor-support/pull/29)

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
radumereuta: I would like to keep this for now.
It makes it really easy to quickly fire up a test instead of trying to debug code through the editor.
We already have quite a few of these in the repo.

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
radumereuta: Basically, I'm doing a tree traversal to search for all the Terminals in productions.
I found this easier than spawning a bunch of visitors since the structure is fixed.
Note that I'm using the [kil](https://github.com/runtimeverification/k/tree/master/kernel/src/main/java/org/kframework/kil) data structures.
These are briefly used in the beginning when we do outer parsing. After that, we rely on the scala classes.
But in this case, I prefer using them because it gives me more control over the contents of the original file:
- the exact position of the production in the file
- the `requires` items for recursing over imported files (not implemented yet).

I've extracted the if block into a function and added some comments; maybe this can make it a bit more readable.

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
Baltoli: Sure, if it helps to test then keep it - maybe take the path as a command line parameter rather than hard-coding a path from your own machine.

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
Baltoli: Extracting the core and annotating the tree structure definitely helps. If this is the best way to do this one-off traversal then I'm fine with it as is.

kernel/src/main/java/org/kframework/lsp/KTextDocumentService.java
radumereuta: Fair point.

Baltoli (approved):

PR #3107, opened 2023-01-11
Build docker images in `${SUBDIR}` context

Baltoli:
I *think* this is the solution to the flaky CI we've been experiencing.

The `with-docker` action is written such that the docker build context for each `Dockerfile` is `.`, but we use a subdir argument in places to point it to a different checkout of K. This means that the build context can end up being the _parent_ directory of the K checkout we actually want to use. If that parent wasn't a recursive checkout, and we don't have the layers cached, then the Haskell backend files needed by `Dockerfile.stack-deps` are missing and the build fails.

My solution is to use `${SUBDIR}` as the build context for the docker layers; this matches my intuition for what the rest of the code is trying to do.

### DIFF
--- a/.github/actions/with-docker/action.yml
+++ b/.github/actions/with-docker/action.yml
@@ -7,7 +7,7 @@ inputs:
   subdir:
     description: 'Subdirectory where code is cloned.'
     required: false
-    default: ""
+    default: "./"
   os: 
     description: 'OS to setup Docker for.'
     required: true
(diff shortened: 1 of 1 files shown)


Baltoli:
FWIW I've kicked this PR 3 times to see if the flakiness resumes, and it's run fine each time; previously I'd have expected to see one of the runs flake out. This suggests to me that it's _probably_ the correct fix.

F-WRunTime (approved):

F-WRunTime:
I believe this is the right fix. It hadn't occurred to me but this is a similar issue I've seen before actually. Approved

PR #3109, opened 2023-01-11
some minor tweaks to existing backends

dwightguth:
These are some minor tweaks to the K frontend which are required to support our closed-source Maude backend. The code that uses this will come in a separate PR in another repo.


kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
```diff
@@ -92,10 +92,12 @@ public static String getKompiledStringAndWriteSyntaxMacros(ModuleToKORE converte
...
     }
 
-    public static Module getKompiledModule(Module mainModule) {
+    public static Module getKompiledModule(Module mainModule, boolean hasAnd) {
         mainModule = new GenerateSortPredicateRules(true).gen(mainModule);
         mainModule = ModuleTransformer.fromSentenceTransformer(new AddSortInjections(mainModule)::addInjections, "Add sort injections").apply(mainModule);
-        mainModule = ModuleTransformer.fromSentenceTransformer(new MinimizeTermConstruction(mainModule)::resolve, "Minimize term construction").apply(mainModule);
+        if (hasAnd) {
+          mainModule = ModuleTransformer.fromSentenceTransformer(new MinimizeTermConstruction(mainModule)::resolve, "Minimize term construction").apply(mainModule);
+        }
```
radumereuta, line 104: Can you write some description of `hasAnd`?
This is not very descriptive.

dwightguth: Sure.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -234,6 +234,7 @@ public void convert(boolean heatCoolEq, String prelude, StringBuilder semantics,
                 continue;
             }
         }
+        genSubsortAxiom(Production(Seq(), Sorts.KItem(), Seq(NonTerminal(Sorts.KConfigVar()))), syntax);
```
ehildenb: Can this subsort just be added to `kast.md` instead of being injected by the kompile pipeline?

ehildenb: Maybe here? https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/kast.md#basic-k-sorts

dwightguth: Hmm. I thought this was already declared and was just mistakenly getting filtered out because the code thought it was builtin syntax, but I see it actually isn't. I'll give it a try. It is not entirely clear though that we won't end up with the same problem of it getting filtered out because it's not actually part of the definition, though.

radumereuta:
What happened to the documentation?
I remember you added it.
Did you remove it by mistake when you forced push?

radumereuta (approved):
lgtm

PR #3115, opened 2023-01-17
Expand on direnv instructions

Baltoli:
There's an additional installation step for `direnv` that's easy to miss out; this PR makes it explicit in the macOS installation instructions.

It also changes the version of `llvm` that the default `envrc` file uses to be `@14` by default; this was out of sync in the README.


reekamaharaj (commented):
I would also add that if you are using zsh (I am) that you would need to run 
`eval "$(direnv hook zsh)"` 

Also the expected response 
`➜  k git:(master) eval "$(direnv hook zsh)"
direnv: loading ~/k/.envrc
direnv: export ~PATH`

However this stuff could go into a troubleshooting section. Which might be helpful for those on M1/M2 Macs.

Baltoli:
> I would also add that if you are using zsh (I am) that you would need to run `eval "$(direnv hook zsh)"`

The shell-specific parts are in the linked page; I don't think we necessarily need to duplicate it all here.

> Also the expected response `➜ k git:(master) eval "$(direnv hook zsh)" direnv: loading ~/k/.envrc direnv: export ~PATH`

Yes, this would be helpful to add. Thanks!

> However this stuff could go into a troubleshooting section. Which might be helpful for those on M1/M2 Macs.

Agreed; that would probably be best done as a separate PR / issue to this one.

reekamaharaj (approved):

radumereuta (approved):

PR #3116, opened 2023-01-17
Address homebrew enthusiastically upgrading the runner environment

Baltoli:
When we build the brew bottle for K macos packaging, homebrew by default will try to upgrade every single package on the system pre-emptively. Because lots of packages are installed on the runner, this can be fragile. The fix is to set an environment variable when running the bottle steps.


F-WRunTime (approved):
Easy fix. Looks good.

PR #3118, opened 2023-01-18
Fix copyright check

dwightguth:
This PR performs the following changes:

1. Removes the years from all copyright headers in the repo. It seems this is the direction most open source projects are heading in these days, because it means they don't need to be updated every year and never get out of date.
2. Fixes multiple bugs in the way we were checking copyright notices so that all files that get checked get checked correctly.
3. Makes the copyright check, which runs as part of `mvn verify`, check every file on every invocation, fixing issues with the incremental check being flaky.
4. Fixes all issues in the repo where the copyright notices were not consistent with the new format.

The net effect of these changes is that `mvn verify` and `mvn install` ought to always succeed now when run locally, and not just on CI. It also has the effect that our copyright notices are no longer out of date, and no longer have to be kept up to date. Legally this is fine as all works are protected by copyright even if the notice is not made explicit in any particular format.

I strongly suggest reviewing the first three commits separately from the fourth, which will likely not be reviewable in the GitHub UI since it touches a massive number of files. The fourth is purely routine however and contains only changes to copyright notices themselves.


pom.xml
```diff
@@ -242,11 +210,11 @@
               <sourceDirectories>
                 <sourceDirectory>${project.basedir}</sourceDirectory>
               </sourceDirectories>
-              <includes>${git.modified.files}</includes>
+              <includes>**/*.xml,**/*.xsd,**/*.md,**/*.k,**/*.kore,**/*.java,**/*.scala,**/*.jj,**/*.str,**/*.css,**/*.sty,**/*.tex,**/*.cls</includes>
```
radumereuta: We also have `*.jjt`

src/main/config/checkstyle-copyright.xml
```diff
@@ -4,40 +4,30 @@
...
   <module name="RegexpHeader">
-    <property name="header" value="^&lt;!-- Copyright \(c\) (\d\d\d\d-)?2019 K Team. All Rights Reserved. --&gt;$" />
+    <property name="header" value="^---$\n^copyright: Copyright \(c\) (Runtime Verification, Inc|K Team). All Rights Reserved.$" />
     <property name="fileExtensions" value="md" />
   </module>
   <module name="RegexpHeader">
-    <property name="header" value="^// Copyright \(c\) (\d\d\d\d-)?2019 (Runtime Verification, Inc.( \(RV-Match team\))?|K Team). All Rights Reserved.$" />
-    <property name="fileExtensions" value="k,kore,java,jj,str" />
+    <property name="header" value="^// Copyright \(c\) (Runtime Verification, Inc|K Team). All Rights Reserved.$" />
+    <property name="fileExtensions" value="k,kore,java,scala,jj,str" />
```
radumereuta: `jjt` here as well.

radumereuta (approved):
lgtm, except for adding the jjt extension.
But I didn't fully check the third commit. It's making my PC run very slowly.

ehildenb (approved):

PR #3123, opened 2023-01-26
Klsp go to definition

radumereuta:
I recommend just looking at the code in the editor. I moved code around quite a bit, and the diff might not be that readable.
- Implemented go to definition for imported modules and required files.
- Added contextual completion: syntax, imports, and rule
- added unit tests
- refactoring

Part of: https://github.com/runtimeverification/k/issues/3070


kernel/src/main/java/org/kframework/lsp/KLanguageServer.java
```diff
@@ -43,6 +43,8 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams initializ
         /* Check if dynamic registration of completion capability is allowed by the client. If so we don't register the capability.
            Else, we register the completion capability.
          */
+        // TODO: check if this needs to be guarded
```
Baltoli, line 46: Does it need to be guarded?

kernel/src/main/java/org/kframework/lsp/KPos.java
```diff
@@ -0,0 +1,70 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        KPos kPos = (KPos) o;
+        return line == kPos.line && character == kPos.character;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(line, character);
+    }
+}
```
Baltoli, line 70: I'm not convinced by this class; couldn't this be implemented as a method `humanReadable` on `Position` that gives you back the 1-based position from the 0-based internal representation?

Baltoli (changes requested):

kernel/src/main/java/org/kframework/lsp/KLanguageServer.java
radumereuta: I'm not sure how the registration of new features works between the client and the server.
Some features may be turned on and off while the plugin is activated.
I got the code from another example. I will have to recheck this part, but for now, I want to concentrate on adding the features, at least until the 8th.

kernel/src/main/java/org/kframework/lsp/KPos.java
radumereuta: The `Position` class is part of the lsp library. So I can't modify it.
I made a new `KPos` class in order to make sure I always get the correct arguments to the klsp.
Basically, I'm relying on the type checker to make sure I always do the correct translations.

kernel/src/main/java/org/kframework/lsp/KLanguageServer.java
Baltoli: OK, if it's a case of prioritising getting features implemented quickly before going back and sorting things out 100% properly then that's fine.

kernel/src/main/java/org/kframework/lsp/KPos.java
Baltoli: That makes sense; perfectly fine if Position isn't our code.

Baltoli (approved):

PR #3128, opened 2023-01-27
add maude backend to KompileOptions

dwightguth:
Without this, the Maude backend will crash if multiple productions with different arities have the same klabel attribute.


radumereuta (approved):
We should talk about giving these more user friendly names.
llvm -> concrete
haskell -> symbolic
...

PR #3142, opened 2023-02-06
Klsp inner

radumereuta:
@Baltoli I would like your review on this, but do not merge yet.
I've cut some corners for the presentation on Wednesday.

Added support for:
- go to definition for terms inside rules
- find references for productions inside rules
- had to create custom data structures because the old ones were not serializable.


kore/src/main/scala/org/kframework/definition/transformers.scala
```diff
@@ -116,4 +115,24 @@ class DefinitionTransformer(moduleTransformer: Module => Module) extends (Defini
...
 
+class KViz(f: K => K, name: String) extends (K => K) {
+  override def apply(input: K): K = {
+    input match {
+      case c: KToken  => f(c)
+      case tc: KApply =>
+        f(tc)
+        tc.items.forEach(apply)
+        tc
+      case _ => throw new AssertionError("Not expected downed term in visitor " + name + " term: " + input)
```
Baltoli, line 135: what's a downed term?

kernel/src/main/java/org/kframework/lsp/KLanguageServer.java
```diff
@@ -68,6 +77,14 @@ public void initialized(InitializedParams params) {
                     "textDocument/diagnostic", diagnosticRegistrationOptions);
             languageClient.registerCapability(new RegistrationParams(List.of(diagnosticRegistration)));
         }
+        // Register file watchers
+        List<FileSystemWatcher> watchers = new ArrayList<>();
+        watchers.add(new FileSystemWatcher(Either.forLeft("/**/" + CACHE_FILE_NAME),
```
Baltoli, line 82: Is this glob definitely portable?

Baltoli:
Seems fine to me so far; let me know when it's ready for a proper review.

kernel/src/main/java/org/kframework/lsp/KLanguageServer.java
radumereuta: I got it from here: https://github.com/lsp-and-implementation/language-server/blob/master/server/server-core/src/main/java/com/lspandimpl/server/core/BalLanguageServer.java#L130
And according to the LSP [documentation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) it should work.
I only tested it on my machine though.

kore/src/main/scala/org/kframework/definition/transformers.scala
radumereuta: Good question.
After parsing, we transform the parser classes into KApps with references to productions.
We then apply `down` to replace `KApply(KLabel("#KRewrite"), L, R)` into `KRewrite(L, R)`.
[Here](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/kore/interface.scala#L31) are the classes. We do this because it's easier to write transformers in kompile. But for the LSP I need all my terms to have references to the original production.
So instead of 
`Cache.save(down(trm))` and `Cache.load()` I had to modify it to
`Cache.save(trm)` and `down(Cache.load())`.

radumereuta (commented):
This is ready for review.
I tried to clean up the code a bit and add more comments.
It ended up quite large, so please ask questions. There are quite a few things that could be improved.

Baltoli (approved):

PR #3149, opened 2023-02-09
Add proper error message

radumereuta:
AddSortInjections  would fail with a stack trace if it failed to find a label in the main module.

The message now reads:
```
[Error] Compiler: Could not find productions for KApply with label _=__TEST-SYNTAX_Stuff_LValue_RValue in module TEST
	Source(/home/radu/work/test/./2.test)
	Location(1,1,1,6)
	1 |	a = b
	  .	^~~~~
```

The change is pretty simple. I don't think it warrants a test. But if you think it's worth it, I can also add a test.

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
+++ b/kernel/src/main/java/org/kframework/compile/AddSortInjections.java
@@ -24,6 +24,7 @@ import org.kframework.kore.KVariable;
 import org.kframework.kore.Sort;
 import org.kframework.parser.outer.Outer;
 import org.kframework.utils.errorsystem.KEMException;
+import scala.Option;
 import scala.Tuple2;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


radumereuta:
This was found by @virgil-serbanuta.
He forgot to import the syntax module into the main module.
Alternatively, we may want to allow for that. I'm not sure how deep the changes to the `AddSortInjections` step would be.

dwightguth:
I mean, if the production isn't in the main module, then it's not part of the definition that the compiler sees at runtime, so of course the compiler won't be able to compute sort injections for it. The only way you might be able to do so would be if you added the sorts and subsorts from the syntax module into the sort graph used by the algorithm, but that wouldn't be sound.

Adding a friendly error message is definitely the correct solution here.

radumereuta:
Is the error message good enough, then?
Can you approve?

Baltoli (approved):

PR #3151, opened 2023-02-10
Add codeowners for version file

radumereuta:
This is to avoid accidentally manually updating the file.

I've seen this happen with the K web update, and by accident on other PRs.
Hopefully, we'll be able to catch this kind of error if we're alerted.

@ehildenb @dwightguth @Baltoli thoughts?


Baltoli:
What does the codeowners file do in the web UI? How will this interact with the automatic version bumping script?

(I think alerting us to PRs that mangle the version is a really good idea, I just want to check what the workflow will look like)

dwightguth:
I agree with Bruce. We need to double check that this doesn't interfere with the scripts during deployment that modify this file.

The change itself looks fine other than that.

ehildenb:
I think it's a good idea. But you likely should add eth @runtimeverification/jenkins  team to the codeowners there, to avoid the CI problem mentioned.

And I think it's worth a try either way, if it breaks something, we can force a revert.

ehildenb (approved):

ehildenb:
I've approved, but please make the above change, and also consider adding Bruce and Freeman to the list.

radumereuta:
@ehildenb I get an error message saying that the team needs to be publicly visible and it has write access to the repository.
For example, I don't see the `jenkins` team in the runtimeverification github org. I can only see External Contractors and RV Staff.

Funny, it works for `@runtimeverification/admin`

CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 /k-distribution/include/kframework/builtin/* @dwightguth @ehildenb
 /Jenkinsfile @runtimeverification/admin
 /.github/workflows/* @runtimeverification/admin
-/package/version @dwightguth @ehildenb @radumereuta
+/package/version @dwightguth @ehildenb @radumereuta @runtimeverification/jenkins @F-WRunTime @Baltoli
```
Baltoli: This is the correct RV Jenkins user, right? https://github.com/rv-jenkins

```suggestion
/package/version @dwightguth @ehildenb @radumereuta @rv-jenkins @F-WRunTime @Baltoli
```

Baltoli:
Looks good but I see that the web UI is complaining - I added a suggestion for what I think is the right invocation to get rv-jenkins tagged.

Baltoli (approved):

PR #3160, opened 2023-02-15
Fix some of the issues from Clarity on tutorial lessons

fiedlr:
I did the no-brainer changes from https://github.com/runtimeverification/k/issues/2033#issuecomment-1216555730 , I ticked them off already to keep track of what I did here.

Let me know if something else needs to be done for deployment. Once this is merged, I'll start working on the harder changes.

### DIFF
--- a/k-distribution/k-tutorial/1_basic/09_unparsing/README.md
+++ b/k-distribution/k-tutorial/1_basic/09_unparsing/README.md
@@ -17,7 +17,7 @@ after the program has been interpreted, and converting it from an AST back into
 text that (in theory, anyway) could be parsed back into the same AST that was
 the output of the execution phase.
 
-In practice, unparsing is not always precisely reversible. It turns out
+In practice, parsing is not always precisely reversible. It turns out
 (although we are not going to cover exactly why this is here), that
 constructing a sound algorithm that takes a grammar and an AST and emits text
 that could be parsed via that grammar to the original AST is an
@@ -286,7 +286,7 @@ formatter interacts with the list of colors specified by the `colors`
 attribute. You can essentially think of the `color` attribute as a way of
 specifying that you want all the colors in the list to be the same color.
 
-Note that the `%c` and `%r` format codes are relatively primitive in nature. 
+Note that the `%c` and `%r` format codes are relatively primitive in nature.
 The `color` and `colors` attributes merely maintain a list of colors, whereas
 the `%c` and `%r` format codes merely control how to advance through that list
 and how individual text is colored.
@@ -316,8 +316,8 @@ module LESSON-09-D
 endmodule
 ```
 
-For a complete list of allowed colors, see 
-[here](https://github.com/kframework/llvm-backend/blob/master/lib/ast/AST.cpp#L381).
+For a complete list of allowed colors, see
+[here](https://github.com/kframework/llvm-backend/blob/master/lib/ast/AST.cpp#L440).
 
 ## Exercises
 
@@ -326,7 +326,7 @@ For a complete list of allowed colors, see
 `(`, `)`, `{`, and `}` a third color.
 
 2. Use the `format`, `color`, and `colors` attributes to tell the unparser to
-style the expression grammar from lesson 1.8, problem 3 according to your own
+style the expression grammar from Lesson 1.8, Exercise 3 according to your own
 personal preferences for syntax highlighting and code formatting. You can
 view the result of the unparser on a function term without evaluating that
 function by means of the command `kparse <file> | kore-print -`.
--- a/k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
+++ b/k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
@@ -1,7 +1,7 @@
 # Lesson 1.14: Defining Evaluation Order
 
 The purpose of this lesson is to explain how to use the `heat` and `cool`
-attributes, `context` and `context alias` sentences, and the `strict` and 
+attributes, `context` and `context alias` sentences, and the `strict` and
 `seqstrict` attributes to more compactly express heating and cooling in K,
 and to express more advanced evaluation strategies in K.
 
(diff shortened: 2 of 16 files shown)


k-distribution/k-tutorial/1_basic/09_unparsing/README.md
```diff
@@ -316,7 +316,7 @@ module LESSON-09-D
 endmodule
 ```
 
-For a complete list of allowed colors, see 
+For a complete list of allowed colors, see
 [here](https://github.com/kframework/llvm-backend/blob/master/lib/ast/AST.cpp#L381).
```
Baltoli: Can you please point this link to line 440? It seems to have drifted.

k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
```diff
@@ -292,8 +292,7 @@ side condition that the denominator is non-zero. Run `krun --search` on the
 program `1 / 0 + 2 / 1` and observe all possible outputs of the program. How
 many are there total, and why?
 
-2. Modify your solution from lesson 1.11 problem 2 to remove the `eval`
-function and instead evaluate expressions from left to right using the
+2. Rework the evaluator function from Lesson 1.11, Exercise 2 to evaluate expressions from left to right using the
```
Baltoli: I don't think this change is correct. I presume that Lesson 1.11 has a *function* that evaluates expressions, but this exercise is looking to do so with *rewriting* instead.

k-distribution/k-tutorial/1_basic/19_debugging/README.md
```diff
@@ -169,7 +168,7 @@ line of code containing the function or rule.
 
 For example, consider the following K definition (`lesson-19-b.k`):
 
-```k
+```{.k .numberLines .line-anchors startFrom="1"}
```
Baltoli: These don't seem to get rendered on GitHub for me.

Baltoli (changes requested):

k-distribution/k-tutorial/1_basic/19_debugging/README.md
Baltoli: @fiedlr These changes also break the K Markdown parser, unfortunately

fiedlr: @Baltoli That's a shame, it works with Pandoc as a way to add additional classNames and attributes to the code block. Is there a way to do this on our website and not break the K Parser? My point is that it's not ideal to reference line numbers without actually displaying line numbers.

Baltoli: Yeah I totally agree with the sentiment; changing the Markdown parser to support this is likely to be a bigger job in the frontend (if it's even possible with the library we use). Is there a way to rewrite the surrounding text instead?

radumereuta: I wrote the [parser for the tags](https://github.com/runtimeverification/k/blob/master/kernel/src/main/jjtree/TagSelector.jjt), so we can do whatever we want there.
Can the website interpret that line? We might be able to update the parser in a way that is backward compatible.

k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
fiedlr: The problem is the sentence is confusing as is. Unless I did something additional by accident, my solution to Lesson 1.11 Problem 2 evaluates expressions both with and without `eval`. But what we can do is to reference the problem that is referenced by Lesson 1.11 Problem 2, where there is no eval and write it straight away like this:

2. Rework the solution to Lesson 1.9, Exercise 2 to evaluate expressions from left to right using the `seqstrict` attribute.

k-distribution/k-tutorial/1_basic/19_debugging/README.md
radumereuta: @Baltoli depending on what we want here, we might just need to update the regex [here](https://github.com/runtimeverification/k/blob/master/kernel/src/main/jjtree/TagSelector.jjt#L144)
It might also be a great first task for @Robertorosmaninho 
The changes are limited to a single file. We have unit tests, and integration tests to update.
I guess the question remains, is the website going to display what we want?

fiedlr: @Baltoli @radumereuta If I checked correctly, we use `markdown-it` for generating HTML from `md`. It doesn't seem to work on their live website https://markdown-it.github.io/

I'll investigate if it's possible to do it another way with this.

fiedlr: There's a `markdown-it` plugin https://www.npmjs.com/package/markdown-it-plugins that claims to have line-numbers, so it should not be a huge change. Styling can be addressed in the CSS.
@yiyi-wang-rv

k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
fiedlr: @Baltoli Let me know if the current change works, thanks!

k-distribution/k-tutorial/1_basic/19_debugging/README.md
Baltoli: @yiyi-wang-rv can this actually be done entirely through CSS, do you know?

k-distribution/k-tutorial/1_basic/14_evaluation_order/README.md
Baltoli: Yes, as written in the comment that's much clearer!

k-distribution/k-tutorial/1_basic/19_debugging/README.md
fiedlr: @radumereuta @Baltoli For the time being, I removed the tags as they break the parser anyway. Let's move this discussion to Slack so that the PR doesn't get stale. Could we merge this after it passes tests? Thanks.

Baltoli (approved):

PR #3162, opened 2023-02-15
Update Imp++ to use LLVM and Haskell backends

dwightguth:
This is a pretty straightforward PR that implements some of the preliminaries to deleting the Java backend from the K frontend; namely, we need to convert the pl-tutorial as much as we can over to the llvm and haskell backends. This PR contains the changes necessary to convert IMP++.

Essentially, we do the following:

* Delete the #parseKAST function since it interferes with proving on the Haskell backend and is about to be deleted with the deletion of the Java backend anyway. Also deletes the accompanying integration tests since they would no longer pass.
* Allow `<generatedCounter>` to be mentioned in rules, in order to allow the sum-to-n proof for IMP++ to mention this cell so it can prove things that use bang variables.
* tweak the makefiles for IMP++ to use the llvm backend and tweak flags as needed to support this.
* update definitions to work with the llvm and haskell backends
* update output files and specifications so that they still pass. They are largely equivalent to what existed before with the exception that some of the search tests no longer discover all the behaviors they did before. This is due to the known issue that strictness is incapable of heating one result, partially evaluating it, cooling it, then evaluating a different result before heating the first result and finishing evaluation. We have already made analogous changes to previous tutorial lessons, and anticipate making further in the future. I have checked to ensure that we have only lost behaviors and no behaviors have changed or been added.


k-distribution/pl-tutorial/1_k/4_imp++/lesson_8/imp.md
```diff
@@ -493,8 +493,8 @@ with the declared configuration cell multiplicity. The unique identifier
 of the new thread is generated using the `fresh` side condition.
 
 ```k
-  rule <k> spawn S => !T:Int ...</k> <env> Rho </env>
-       (.Bag => <thread>... <k> S </k> <env> Rho </env> <id> !T </id> ...</thread>)
+  rule <k> spawn S => !T:Int +Int 1 ...</k> <env> Rho </env>
+       (.Bag => <thread>... <k> S </k> <env> Rho </env> <id> !T +Int 1 </id> ...</thread>)
```
radumereuta, line 497: Why do you need to add `+Int 1`?

k-distribution/pl-tutorial/1_k/4_imp++/lesson_6/imp.k
```diff
@@ -98,7 +99,7 @@ module IMP
   rule <k> halt; ~> _ => . </k>
 
   rule <k> spawn S => . ...</k> <env> Rho </env>
-       (.Bag => <thread>... <k> S </k> <env> Rho </env> ...</thread>)
+       (.Bag => <thread>... <k> S </k> <env> Rho </env> <id> !T +Int 1 </id> ...</thread>)
```
radumereuta: This generates a warning you should fix so we don't clutter the output on CI.
```
[Warning] Compiler: Variable '!T' defined but not used. Prefix variable name with underscore if this is intentional.
	Source(/home/radu/work/k/k-distribution/pl-tutorial/1_k/4_imp++/lesson_6/imp.k)
	Location(102,62,102,64)
```

k-distribution/include/kframework/builtin/kast.md
```diff
@@ -282,6 +282,17 @@ module RULE-CELLS
 
   syntax #OptionalDots ::= "..." [klabel(#dots), symbol]
                          | ""    [klabel(#noDots), symbol]
+
+  syntax Int
```
radumereuta, line 286: Not happy about adding this sort here, but I understand the overall problem.
We might have issues if we want to restrict sorts to have a single point of origin.

.github/workflows/test-pr.yml
```diff
@@ -98,6 +98,7 @@ jobs:
           repository: kframework/k-exercises
           token: ${{ secrets.JENKINS_GITHUB_PAT }}
           submodules: recursive
+          ref: 'imp++'
```
radumereuta: Why is this an issue? And how is it helping?

radumereuta (approved):
lgtm overall.
Just a few minor things.

k-distribution/pl-tutorial/1_k/4_imp++/lesson_8/imp.md
dwightguth: This is because the default value for the initial thread is id 0, but if we try to create a new thread before the fresh counter is incremented, it would try to create a second thread 0 and crash.

k-distribution/include/kframework/builtin/kast.md
dwightguth: I mean, we're unlikely to do that because being able to declare sorts this way is kind of needed as a form of forward declaration to break certain types of cycles between modules.

.github/workflows/test-pr.yml
dwightguth: this is just for testing purposes because we need to test against the modified version of the exercises. I will revert this change before merging.

PR #3173, opened 2023-02-20
Add selection range in lsp

radumereuta:
To test open VSCode in workspace mode and press Ctrl+Shift+LRArrows
It should increase and decrease the selection of the text depending on the AST.
This is useful when you have a large expression and you want to select something valid.

Part of #3070


Baltoli (commented):
This is cool - nice work @radumereuta !

Baltoli (approved):

PR #3174, opened 2023-02-20
Fix another batch of suggestions from Clarity on tutorial lessons

fiedlr:
I fixed another few of suggestions from https://github.com/runtimeverification/k/issues/2033#issuecomment-1216555730, this time a bit more complicated changes.

### DIFF
--- a/k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
+++ b/k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
@@ -44,8 +44,7 @@ endmodule
 In this example, some of the `|` symbols separating productions in a single
 block have been replaced with `>`. This serves to describe the
 **priority groups** associated with this block of productions.
-
-In this example, the first priority group consists of the atoms of the
+The first priority group consists of the atoms of the
 language: `true`, `false`, and the bracket operator. In general, a priority
 group starts either at the `::=` or `>` operator and extends until either the
 next `>` operator or the end of the production block. Thus, we can see that the
@@ -316,11 +315,23 @@ modify the definition so that you will get the alternative parse.
 2. Modify your solution from Lesson 1.3, Exercise 2 so that unary negation should
 bind tighter than multiplication and division, which should bind tighter than
 addition and subtraction, and each binary operator should be left associative.
-Write these priority and associativity declarations both inline and explicitly.
+Write these priority and associativity declarations explicitly, and then
+try to write them inline.
 
 3. Write a simple grammar containing at least one ambiguity that cannot be
 resolved via priority or associativity, and then use the `prefer` attribute to
 resolve that ambiguity.
+
+4. Explain why the following grammar is not labeled ambiguous by the K parser when parsing `abb`, then make the parser realize the ambiguity.
+```k
+module EXERCISE4
+
+syntax Expr ::= "a" Expr "b"
+              | "abb"
+              | "b"
+
+endmodule
+```
 
 ## Next lesson
 
--- a/k-distribution/k-tutorial/1_basic/02_basics/README.md
+++ b/k-distribution/k-tutorial/1_basic/02_basics/README.md
@@ -84,11 +84,11 @@ The first thing to realize is that this K definition contains 5 productions.
 Productions are introduced with the **syntax** keyword, followed by a **sort**,
 followed by the operator `::=` followed by the definition of one or more
 productions themselves, separated by the `|` operator. There are different
-types of productions, but for now we only care about **constructors** and 
+types of productions, but for now we only care about **constructors** and
 **functions**. Each declaration separated by the `|` operator is individually
 a single production, and the `|` symbol simply groups together productions that
 have the same sort. For example, we could equally have written an identical K
-definition like so:
+definition `lesson-02-b.k` like so:
 
 ```k
 module LESSON-02-B
@@ -105,7 +105,15 @@ module LESSON-02-B
 endmodule
 ```
 
-Or even:
+You can try compiling and running `lesson-02-b.k` to see that it produces the same output as `lesson-02-a.k`:
+
+```
+kompile lesson-02-b.k
+krun -cPGM='colorOf(Banana())' --definition 'lesson-02-b-kompiled'
+```
+
+where the `--definition` attribute points to the *directory* containing a compiled version of `LESSON-02-B`.
+Even the following definition is equivalent:
 
 ```k
 module LESSON-02-C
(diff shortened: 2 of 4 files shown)


k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -316,12 +315,24 @@ modify the definition so that you will get the alternative parse.
...
+```k
+module EXERCISE4
+
+syntax Expr ::= "a" Expr "b"
+              | "abb"
+              | "b"
+
+endmodule
+```
+
```
radumereuta, line 335: It's not labeled ambiguous because we have a scanner that matches the longest token.
Not sure how you would fix this.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -316,12 +315,24 @@ modify the definition so that you will get the alternative parse.
 2. Modify your solution from Lesson 1.3, Exercise 2 so that unary negation should
 bind tighter than multiplication and division, which should bind tighter than
 addition and subtraction, and each binary operator should be left associative.
-Write these priority and associativity declarations both inline and explicitly.
+Write these priority and associativity declarations explicitly and then
+try to think whether they can be written inline.
```
radumereuta: To me, it sounds like you want to make the inline part optional.

k-distribution/k-tutorial/1_basic/15_configurations/README.md
```diff
@@ -242,9 +242,9 @@ of this term in later lessons.
 
 1. Modify the definition from the previous exercise in this lesson so that the
 Boolean cell you created is initialized to false. Then add a production
-`syntax Stmt ::= Bool ";" Exp`, and a rule that uses this `Stmt` to set the
+`syntax Stmt ::= Bool ";" Stmt | ""`, and a rule that uses this `Stmt` to set the
```
radumereuta: Productions that don't contain any syntax `""` are very tricky for the inner parser to handle. So they are forbidden.
You can get degenerate things like `rule =>` because you can match nothing, anywhere.
You can use them in your program language but using a separate module that is not included in the main semantics module. This is a tricky thing, and I don't think it's worth adding to the basic lessons.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
fiedlr: Right, that's the point... I thought it could be useful adding it so that the reader learns about this.

For me, the solution is to use a sequence of tokens "a" and "b", not a token "abb". This does not *fix* the grammar, it just makes the parser realize the ambiguity.
```k
module EXERCISE2

  syntax Expr ::= "a" Expr "b"
              | "a" "b" "b"
              | "b"

endmodule
```

```
[Error] Inner Parser: Parsing ambiguity.
1: syntax Expr ::= "a" Expr "b"
    `a_b_EXERCISE2_Expr_Expr`(`b_EXERCISE2_Expr`(.KList))
2: syntax Expr ::= "a" "b" "b"
    `abb_EXERCISE2_Expr`(.KList)
	Source(<command line: -e>)
	Location(1,1,1,4)
```

Maybe I should change the phrasing to "make the parser realize the ambiguity"? Or we could move it out of the exercises and add a note somewhere, but I thought the exercise is kind of fun.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
fiedlr: I'd like to keep it mandatory without telling the reader it's actually possible.
How about the following?

```
Write these priority and associativity declarations explicitly and then
answer whether they can be written inline.
``` 

Not that important a change though, we can drop it :-)

k-distribution/k-tutorial/1_basic/15_configurations/README.md
fiedlr: Good to know. I just wanted to make the exercise easier for testing the actual solution by enabling a `reset` after a `true`. We can change it to something like

```k
  syntax Stmt ::= Bool ";" Stmt | Exp
  syntax Stmt ::= "reset" ";" Stmt
```

So the sequence could be `-cPGM='true;reset;1'`, for example.

Or I might have misunderstood the exercise myself :-)

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
fiedlr: Actually I thought of the exercise while thinking about why my grammar in Exercise 3 isn't detected as ambiguous, so this shouldn't require any preliminary knowledge of K.

radumereuta: Yeah. You convinced me.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
radumereuta: How about
```
Write these priority and associativity declarations explicitly, and then
try to write them inline.
```

k-distribution/k-tutorial/1_basic/15_configurations/README.md
radumereuta: I think I understand it in the same way. But making it explicit may make it a bit too easy.
Leaving the `Stmt` sequencing out makes the user think about more than just copy/pasting something in.
So I think it is fine as it was.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
fiedlr: Works for me!

k-distribution/k-tutorial/1_basic/15_configurations/README.md
fiedlr: Hmm... Why not, I mean one can figure it out how to test it.... But personally I would stick to making exercises ask a single thing ~> in this case, do I understand how to write the semantics for reset by actually referencing the initial value instead of copy&pasting it/hardwiring it twice? But sure, we can keep it this way.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
Baltoli: Yeah, I'm happy to include this as a neat illustration of why the scanner-based parser works the way it does.

Baltoli (approved):

Baltoli:
This all looks fine to me; Radu's comments addressed the queries I had with the changes. Thanks for doing this @fiedlr !

radumereuta:
Hey @fiedlr. To merge changes into develop, we don't press the merge button manually. We add the `automerge` tag and let CI pick the order to merge PRs. Doing it this way can cause history problems.
Relying on the CI system also means we do minimal work, and we make sure we test every single PR being merged.

I don't think it matters in this case, but you should add the `automerge` in the future.

fiedlr:
@radumereuta Sorry, I checked the commit log and there are recent manual merges, so I thought it was fine. I'll do automerge in the future, thanks for letting me know.

Issue #3169, opened 2023-02-17
Add the option to add line numbers to code blocks on the K website


fiedlr:
See https://github.com/runtimeverification/k/pull/3160#discussion_r1108394995 for the reasons why.

PR #3176, opened 2023-02-21
[K MD Parser] Extend TagSelector to Support Line Numbers in Code Blocks

Robertorosmaninho:
Fixes: https://github.com/runtimeverification/k/issues/3169

This PR implements the infrastructure needed to enable this [feature](https://github.com/runtimeverification/k/issues/3169). It allows the K Markdown Parser to recognize attributes on tags like
```
{.k .numberLines .line-anchors startFrom="1"}
``` 
The idea is mainly to enable line numbers to support in K code blocks on tutorials and further documentation, as explained [here](https://github.com/runtimeverification/k-web-theme/pull/27).


radumereuta (changes requested):
@Robertorosmaninho if you look at the diff, you will see some changes to version files.
This happens because you based your changes on the master branch, but you were supposed to use the develop branch.
Please do a git rebase onto develop and force push.
The changes lgtm though.

Robertorosmaninho:
I've just rebased it, thanks for the explanation @radumereuta!

Baltoli (approved):

k-distribution/tests/regression-new/markdownSelectors/test.md
```diff
@@ -24,6 +24,11 @@ module TEST
   rule <k> 1 => 2 </k>
 ```
 
+
+```{.k .discard .numberLines startFrom="0"}
+  rule <k> 2 => 3 </k>
+```
+
```
Baltoli: This rule will not interact with the Makefile in the way I think you intend it to; the selector for `kompile` is:
```
k|(keep&!discard)
```
which means that because this rule has a `k` tag, it's selected and the `discard` tag doesn't need to be considered.

You can test this locally even if you can't build the Haskell backend by running `make -n` and running the `kompile` and `krun` commands it shows you.

Baltoli (changes requested):

Baltoli:
The test output `a-spec-md.out` needs to be updated as well, because you've added lines to the test source code. This is hard to tell without the Haskell backend working on your machine, but this diff will get you the correct code I think:
```diff
-[Error] Outer Parser: Module TEST differs from previous declaration at Source(test.md) and Location(7,1,32,10). This can happen if --md-selector differs for kompile and kprove
+[Error] Outer Parser: Module TEST differs from previous declaration at Source(test.md) and Location(7,1,37,10). This can happen if --md-selector differs for kompile and kprove
        Source(test.md)
-       Location(7,1,32,10)
+       Location(7,1,37,10)
           .    v~~~~~~~~~~
         7 |    module TEST
           |            ...
-       32 |    endmodule
+       37 |    endmodule
           .    ~~~~~~~~^
 [Error] Outer Parser: Had 1 outer parsing errors.
```

radumereuta:
> The test output `a-spec-md.out` needs to be updated as well, because you've added lines to the test source code. This is hard to tell without the Haskell backend working on your machine, but this diff will get you the correct code I think:
> 
> ```diff
> -[Error] Outer Parser: Module TEST differs from previous declaration at Source(test.md) and Location(7,1,32,10). This can happen if --md-selector differs for kompile and kprove
> +[Error] Outer Parser: Module TEST differs from previous declaration at Source(test.md) and Location(7,1,37,10). This can happen if --md-selector differs for kompile and kprove
>         Source(test.md)
> -       Location(7,1,32,10)
> +       Location(7,1,37,10)
>            .    v~~~~~~~~~~
>          7 |    module TEST
>            |            ...
> -       32 |    endmodule
> +       37 |    endmodule
>            .    ~~~~~~~~^
>  [Error] Outer Parser: Had 1 outer parsing errors.
> ```

@Robertorosmaninho normally you would write `make update-results` on your machine and it would take care of everything. But in this case you can just get the diff from Bruce.

k-distribution/tests/regression-new/markdownSelectors/test.md
Robertorosmaninho: What do you think about changing the MD selector to `'(k&!discard)|(keep&!discard)'` ?
This still will execute the correct cells while preserving the idea of the test.

radumereuta: @Robertorosmaninho, instead of modifying the Makefile, I would modify this list of attributes to
`{.discard .numberLines startFrom="0"}`
This way, you don't include this rule in the definition, so you don't have to modify the output of the test file.
You will have to modify the output of the error message, though, since it changes location information.

k-distribution/tests/regression-new/markdownSelectors/Makefile
```diff
@@ -2,7 +2,7 @@ DEF=test
 EXT=test
 TESTDIR=.
 KOMPILE_BACKEND=haskell
-KOMPILE_FLAGS=--md-selector 'k|(keep&!discard)'
+KOMPILE_FLAGS=--md-selector '(k&!discard)|(keep&!discard)'
```
Baltoli:
```suggestion
KOMPILE_FLAGS=--md-selector '(k|keep) & !discard'
```
saves a bit of repetition.

Baltoli (approved):
One minor change, but looks good to me if tests pass.

k-distribution/tests/regression-new/markdownSelectors/Makefile
radumereuta: Why do you need to modify this file?
Just change this
```
{.k .discard .numberLines startFrom="0"}
to this:
{.discard .numberLines startFrom="0"}
```

Robertorosmaninho: I think the idea is to reproduce precisely the behavior we expect to see in the production code. Although, I don't believe that it's a strong need.

Robertorosmaninho: But I like @Baltoli's idea! It's even clearer than before while preserving the behavior!

radumereuta: I usually prefer minimizing the diff and keeping the PR focused on one thing.
I added that expression to test that we can parse all the elements, not to make any particular sense.
But ok, I guess...

radumereuta (approved):
lgtm
You may want to confirm with @fiedlr that this is working for him.
After that, you can add `automerge`.

k-distribution/tests/regression-new/markdownSelectors/Makefile
```diff
@@ -2,7 +2,7 @@ DEF=test
 EXT=test
 TESTDIR=.
 KOMPILE_BACKEND=haskell
-KOMPILE_FLAGS=--md-selector 'k|(keep&!discard)'
+KOMPILE_FLAGS=--md-selector '(k|keep) & !discard'
```
fiedlr, line 5: Can you please explain what this does? It keeps only those for tests that are marked either k or keep and are not  discard? I.e., the discards pass through the parser but are not compiled?

kernel/src/main/jjtree/TagSelector.jjt
```diff
@@ -141,7 +141,7 @@ SKIP :
 
 TOKEN : /* TAG */
 {
-  <TAG: (["_","a"-"z","A"-"Z","0"-"9","#","$","%","^","*","+",".","-"])+>
+  <TAG: (["a"-"z", "A"-"Z", "0"-"9", "#", "$", "%", "^", "*", "+", ".", "-", "\"", "=", "_"])+>
```
fiedlr, line 144: So we are not adding `:` after all?

k-distribution/tests/regression-new/markdownSelectors/Makefile
Baltoli: The `--md-selector` flag consumes a boolean expression over tag names, and keeps only those code blocks that satisfy the expression. Non-satisfying blocks are discarded before being parsed. https://kframework.org/k-distribution/k-tutorial/1_basic/08_literate_programming/

kernel/src/main/jjtree/TagSelector.jjt
Robertorosmaninho: I'm still waiting for someone from the web-dev team to answer me if it may be helpful for them. Although, since it's blocking others' PR, we can move forward with this, and if necessary, I can open other PR and implement it.

fiedlr: Ok

fiedlr (approved):
Even though I do not understand why you didn't choose the solution that @radumereuta mentioned, if you insist that  `.k` has to be in the selector and it's worth making a change to the `Makefile` and @Baltoli @radumereuta are ok with it, then I don't have a reason to stall this after all tests pass :-) The `:` can be added later.

Baltoli:
@Robertorosmaninho Please go ahead and fix your merge conflict, then feel free to automerge.

Robertorosmaninho:
> Even though I do not understand why you didn't choose the solution that @radumereuta mentioned, if you insist that `.k` has to be in the selector and it's worth making a change to the `Makefile` and @Baltoli @radumereuta are ok with it, then I don't have a reason to stall this after all tests pass :-) The `:` can be added later.

As far as I got from @yiyi-wang-rv's answer, the [library](https://www.npmjs.com/package/markdown-it-attrs) we're currently using doesn't support `:` and `;`, so we can consider it as fully implemented, at least for now!

PR #3181, opened 2023-02-22
Update simple to llvm backend

dwightguth:
Updates simple-typed-static and simple-typed-dynamic to use the llvm backend. Requires corresponding updates to the exercises.

### DIFF
--- a/k-distribution/pl-tutorial/2_languages/1_simple/2_typed/2_dynamic/exercises/functions-with-throws/tests/div-nondet.simple.out
+++ b/k-distribution/pl-tutorial/2_languages/1_simple/2_typed/2_dynamic/exercises/functions-with-throws/tests/div-nondet.simple.out
@@ -1,7 +1,17 @@
-  S ==K "0\n"
+  {
+    S:String
+  #Equals
+    "1\n"
+  }
 #Or
-  S ==K "1\n"
+  {
+    S:String
+  #Equals
+    "2\n"
+  }
 #Or
-  S ==K "2\n"
-#Or
-  S ==K "3\n"
+  {
+    S:String
+  #Equals
+    "3\n"
+  }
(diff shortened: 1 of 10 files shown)


k-distribution/pl-tutorial/2_languages/1_simple/2_typed/2_dynamic/exercises/typed-exceptions/tests/div-nondet.simple.out
```diff
@@ -1,7 +1,17 @@
...
+  }
 #Or
-  S ==K "2\n"
-#Or
-  S ==K "3\n"
+  {
+    S:String
+  #Equals
+    "3\n"
+  }
```
ehildenb, line 17: This one appears to have one less result?

k-distribution/pl-tutorial/2_languages/1_simple/2_typed/2_dynamic/exercises/functions-with-throws/tests/div-nondet.simple.out
```diff
@@ -1,7 +1,17 @@
...
+  }
 #Or
-  S ==K "2\n"
-#Or
-  S ==K "3\n"
+  {
+    S:String
+  #Equals
+    "3\n"
+  }
```
ehildenb, line 17: Same here, one fewer results.

dwightguth:
@ehildenb The missing result is due to the fact that the llvm and Haskell backends are currently unable to heat one operand of a symbol, then partially evaluate it, then cool it, then heat another argument, evaluate it to a result, cool it, and then heat the first operand again and fully evaluate it. This is a known issue with the nondeterminism of strictness that we have decided to ignore for the time being. The same issue has led to missing results in several other tests where those definitions have already been modified and those search results removed. I've already spoken to Grigore about this issue and the consensus is that we should not block the migration of the tutorial off of the java backend on this particular issue.

ehildenb:
Sounds good, is this documented in an issue with a minimal example somewhere?

dwightguth:
It is not. I'll create an issue tomorrow since I'm done coding for the day. I'll link it here and then I'll let you re review

radumereuta (approved):
lgtm

dwightguth:
Here is the issue I created: https://github.com/runtimeverification/k/issues/3185 It has a minimal example.

PR #3194, opened 2023-02-27
Add K-Javalette

hjorthjort:


web/pages/projects.md
```diff
@@ -79,3 +79,4 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
 - [Boogie (2020)](https://github.com/kframework/boogie-semantics)
 - [Elrond (2020-2021)](https://github.com/runtimeverification/elrond-semantics)
 - [P4 (2016-2021)](https://github.com/kframework/p4-semantics)
+- [K-Javalette (2022)](https://github.com/bbyalcinkaya/k-javalette)
```
radumereuta, line 82: This looks like it's under active development.
If that's true, then it should go under `Featured`. 
And also add a short description of the project.

radumereuta (changes requested):
Please rebase your changes onto `develop` branch.
`master` is automatically update with the right version by CI.
We don't push directly to it. We do this so we make sure that the latest `master` branch always reports the correct version and it passes all the tests.

web/pages/projects.md
hjorthjort: It was a contained MSc project and it's no longer being actively developed. That might change soon though. It's really a toy/teaching language so I'm not sure it belongs in the featured section. But if you think it should go there I can write a short description and move it.

radumereuta: Ah, I see. It's fine in the archive section then.
Just make sure to rebase the changes onto `develop`.

hjorthjort: Done.

radumereuta (approved):

PR #3197, opened 2023-02-27
Update deps GH workflow 

F-WRunTime:


.github/workflows/master-push.yml
```diff
@@ -53,3 +53,17 @@ jobs:
...
+
+      - name: 'Update dependents'
+        env:
+          GITHUB_TOKEN: ${{ secrets.JENKINS_GITHUB_PAT }}
+        run: |
+          set -x
+          version="${GITHUB_SHA}"
+          curl --fail                                                          \
+            -X POST                                                            \
+            -H "Accept: application/vnd.github+json"                           \
```
ehildenb: This should run in the `release.yml` workflow, and replace this: https://github.com/runtimeverification/k/blob/master/.github/workflows/release.yml#L345

Note there that we also are not using `GITHUB_SHA` for the version, but the output from a previous stage.

ehildenb: This should be removed now.

.github/workflows/release.yml
```diff
@@ -355,17 +355,20 @@ jobs:
...
           VERSION: ${{ needs.macos-build.outputs.version }}
         run: |
-          curl --fail 'https://ci.runtimeverification.com/jenkins/buildByToken/buildWithParameters' \
-              --data job=Devops/master \
-              --data token=$JENKINS_DEVOPS_TOKEN \
-              --data UPDATE_DEPS=true \
-              --data UPDATE_DEPS_REPO=runtimeverification/k \
-              --data UPDATE_DEPS_VERSION=v$VERSION
+          set -x
+          version="${GITHUB_SHA}"
```
ehildenb:
```suggestion
          set -x
          version="${VERSION}"
```

ehildenb: You can also remove the `v` from eth line 371 in `"version":"'v${VERSION}'"`, and add it here instead, if you want.

.github/workflows/release.yml
```diff
@@ -355,17 +355,20 @@ jobs:
...
           VERSION: ${{ needs.macos-build.outputs.version }}
         run: |
-          curl --fail 'https://ci.runtimeverification.com/jenkins/buildByToken/buildWithParameters' \
-              --data job=Devops/master \
-              --data token=$JENKINS_DEVOPS_TOKEN \
-              --data UPDATE_DEPS=true \
-              --data UPDATE_DEPS_REPO=runtimeverification/k \
-              --data UPDATE_DEPS_VERSION=v$VERSION
+          set -x
+          version=v"${GITHUB_SHA}"
```
ehildenb:
```suggestion
          version=v"${VERSION}"
```

F-WRunTime:
Thanks copy / paste error there

ehildenb (approved):

PR #3198, opened 2023-02-28
Fix loc info for gen config cells

radumereuta:
Make it so generated productions from the configuration cells, have the exact location of the cell they are coming from instead of the entire configuration.

This made it look weird in the IDE.


Baltoli (approved):

radumereuta:
Turns out we had a test, but nobody noticed the location information was a bit rough.

PR #3230, opened 2023-03-15
Automatically use `--no-haskell-binary` in macOS regression tests

### LINKED ISSUE #3222: regression-new/array-haskell test segfaults on MacOS
Running `mvn verify` on MacOS reports a segfault in the array-haskell tests. Reproducing the test manually confirms this:
```
cd k-distribution/tests/regression-new/
make -C array-haskell
cd array-haskell
kompile --syntax-module TEST --no-exc-wrap --backend haskell test.k -d .
set -o pipefail; (cat 1.test.in 2>/dev/null || true) | krun 1.test --no-exc-wrap -d . | diff - 1.test.out
```

Note that if the test suit had additionally passed `--no-haskell-binary` to `kompile` above it would resolve the issue.

Baltoli:
Fixes https://github.com/runtimeverification/k/issues/3222 by patching the `--no-haskell-binary` flag in at the Makefile level, rather than in Nix.

Also fixes a minor issue where `grep -P` is not available on macOS; we use the portable `egrep` alternative instead in that test.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -150,17 +150,6 @@
                 patchShebangs tests/regression-new/*
                 substituteInPlace tests/regression-new/append/kparse-twice \
                   --replace '"$(dirname "$0")/../../../bin/kparse"' '"${k}/bin/kparse"'
-                ${
-                # we add the `--no-haskell-binary` flag due to the compact library
-                # (used to create the binary haskell files) being broken on Mac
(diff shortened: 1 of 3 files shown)


radumereuta (approved):

Scott-Guest (commented):
Confirmed that the flag is provided and the tests pass on my macOS machine. Thanks for fixing this!

PR #3232, opened 2023-03-15
Update test-pr.yml

F-WRunTime:
This is an approach to enable workflow checks to be run for non org community contributors


Baltoli:
LGTM per discussions on Slack, but will leave final approval to @dwightguth / @ehildenb

dwightguth (approved):
I didn't know about this feature but it looks like exactly what we need. As long as we only approve prs made in good faith, we should be fine.

Baltoli (approved):

PR #3242, opened 2023-03-16
Update pl-tutorial/1_k/5_types

dwightguth:
This PR updates the 5_types set of lessons in the PL tutorial to use the llvm and haskell backends. It also disables some lessons which require features not currently supported by either backend.

5_types does not have any exercises, so there is not a companion PR to this PR, and it can be merged as soon as it's approved and tests are passing. However, it is currently blocked on https://github.com/runtimeverification/haskell-backend/pull/3549


k-distribution/pl-tutorial/1_k/5_types/lesson_5.5/lambda.k
```diff
@@ -42,7 +42,7 @@ module LAMBDA
   rule int <= int => bool
   rule if bool then T:Type else T => T
   syntax KItem ::= env(Map)
-  syntax Exp ::= Exp "->" Exp                  [strict]
+  syntax Exp ::= Exp "->" Exp                  [seqstrict, klabel(functionType)]
```
radumereuta, line 45: Why does it need to be `seqstrict`?

radumereuta (approved):
lgtm with just one question

k-distribution/pl-tutorial/1_k/5_types/lesson_5.5/lambda.k
dwightguth: it doesn't really need to be, but making it sequential doesn't affect correctness at all, but it does dramatically decrease the amount of branching the Haskell backend does and thus makes it considerably faster.

PR #3244, opened 2023-03-16
Remove references to java backend from main regression test suite

### LINKED ISSUE #3168: Remove all tests using Java backend from repo
This issue blocks #3167.

We need to remove all tests that use the Java backend from the repository so the backend itself can be deleted. This means converting as many tests as possible to the LLVM and Haskell backends, and deleting tests that are no longer relevant or test features that do not exist on the new backends.

dwightguth:
Here we remove all calls to the Java backend from the main test suite in preparation for deletion of the backend itself.

Tests were converted to the llvm or haskell backend from the Java backend, run, and then gradually refined until it was either clear the test relied on behavior of the Java backend that couldn't easily be replicated, in which case we deleted the test, or until the test behaved in a sensible fashion, at which point test output files were updated as needed.

We also remove references to a couple different command line flags which will be deleted when the Java backend is removed.


k-distribution/tests/regression-new/concrete-function/2.a.out
```diff
@@ -1,3 +1,3 @@
 <k>
-  sym2 ( 222 ) +Num 1
+  foo ( sym2 ( 222 ) +Num 1 ) ~> .
```
radumereuta, line 2: Is this correct? I see there is a rule for `foo`.

radumereuta (changes requested):
Overall looks good.
But there are still a few warnings printed by the haskell backend for missing `no-evaluators`.
We should try to keep the output clean so it's easier to spot where something fails.

k-distribution/tests/regression-new/concrete-function/2.a.out
dwightguth: That rule doesn't apply because it can't prove that +Num is defined on those arguments.

dwightguth:
There was one remaining test left that still had warnings about functions without evaluators. I fixed it. Please re-review.

radumereuta (approved):
There are still a few warnings thrown from the haskell backend through kprove-legacy, but they are small. We may want to tackle those with another update to remove kprove-legacy.

dwightguth:
I agree that should be a separate pr.

PR #3251, opened 2023-03-17
Delete the java backend!

### LINKED ISSUE #3167: Delete Java Backend

dwightguth:
This is it! We delete the Java backend from the codebase, remove references to it, clean up the K prelude to remove modules that are no longer needed, delete krun-legacy, and clean up all the places in the code where we were branching on whether or not we were using the Java backend. we also delete a couple command line flags that were only being used by the Java backend.


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
 ---
 
+K Framework 6.0.0
+=================
+
+Removals
+--------
+
+- Deleted the 'java' backend of K.
+
```
radumereuta: I think we should have a separate PR with just the version bump. There are many other changes that should be included in this changelog.
We should also consider removing `kprove-legacy` before that.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -49,7 +49,7 @@ GlobalOptions getGlobalOptions_UseOnlyInGuiceProvider() {
     public OutputDirectoryOptions outputDirectory = new OutputDirectoryOptions();
 
     // Common options
-    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [llvm|haskell|kore|java]. Each creates the kompiled K definition.")
+    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [llvm|haskell|kore]. Each creates the kompiled K definition.")
```
radumereuta, line 52: The `kore` backend doesn't exist.

radumereuta (approved):
Tested the changes locally and all looks good.

I think we should have a separate PR for the version bump. Other than that, looking good.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
dwightguth: It does, in fact. It emits a kore definition and then immediately stops. It's not really used in practice, but it does exist.

CHANGELOG.md
dwightguth: The problem with doing what you suggest is that we are no longer able to say "if you want to use the Java backend, download K 5.5"

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
radumereuta: I was looking at `Backends.java` and it doesn't have an option there. I guess this is something else to fix.

CHANGELOG.md
radumereuta: Then should we add a minor version bump to cover all the changes until now and then immediately bump it to v6?

dwightguth: That could work, yes. Will you tackle the minor version bump?

radumereuta: I can do that. Either tonight or tomorrow on the train.
Should we remove `kprove-legacy` before v6 as well?

dwightguth: Let's ask about kprove-legacy at the meeting Thursday. I'm not sure how much it's still being used currently.

radumereuta: I made an issue and added it to the board:
https://github.com/runtimeverification/k/issues/3257

ehildenb:
Let's merge!

dwightguth:
It needs conflicts resolved. Feel free to assign someone to do that; I'm hoping to be back to work feeling better from my illness on Monday, but I'm not feeling well enough yet to do it today.

radumereuta:
> It needs conflicts resolved. Feel free to assign someone to do that; I'm hoping to be back to work feeling better from my illness on Monday, but I'm not feeling well enough yet to do it today.

We decided to make a blog post about it and release them at the same time: https://github.com/runtimeverification/k/issues/3290

PR #3259, opened 2023-03-20
Allow bison parsers to be generated as libraries

### LINKED ISSUE #3258: Generate Bison parser libraries
A weakness of the "K as a library" approach (https://github.com/runtimeverification/web-common/issues/388) - getting from surface language to KORE requires a parsing step which might be non-obvious

We don't want to rely on these executables when generating embeddable shared library bindings; instead, we should be able to turn the bison parsers into libraries that we can _link into the generated shared library_ (or at least pass along with the interpreter library).

Baltoli:
Fixes https://github.com/runtimeverification/k/issues/3258

This PR is related to my ongoing work towards documenting "K as a library" (i.e. using the LLVM backend to generate libraries users can use to build their own tools, rather than relying on the K scripts to run their programs. One missing link in the existing toolchain that I identified was generating KORE from a surface program; doing so currently would require the user to shell out to one of the bison-generated parsers. This PR makes a small improvement to that workflow by allowing the bison parsers to be generated as _shared libraries_ rather than as executables; the parser libraries can then be shipped alongside the C bindings libraries for user definitions.

After this PR, users will be able to run something like:
```console
$ kompile test.k -v             \
  --gen-glr-bison-parser        \
  --bison-parser-library        \
  --llvm-kompile-type c         \
  --llvm-kompile-output libfoo.so
```
to generate a pair of shared libraries (`libfoo.so`, `libparse_KItem_TEST.so`) that provide an LLVM backend runtime and GLR Bison parser, respectively. These can then be embedded into an application.

The changes in this PR are as follows:
* Update the Bison wrapper code to use a similar "file or buffer" abstraction to the LLVM backend's `writer` type; this lets us output to a file directly as before for best performance, but also to keep the parsed KORE in memory if required.
* Add some preprocessor definitions that create a function `char *parse_SORT(...)` as well as the existing bison `main` entrypoint.
* Add frontend flags that generate the library following the syntax above.
* Add a custom integration test that links a small application against a generated parser and calls it in library mode.


dwightguth:
I'd like to review this, but I might not get to it until next week. Is this blocking anything currently?

Baltoli:
> I'd like to review this, but I might not get to it until next week. Is this blocking anything currently?

No, I'm only using it in the context of my own work on a blog post; I'm using `kup` overrides to point the project at branches so there's absolutely no rush at all to review. Enjoy your week!

k-distribution/tests/regression-new/bison-parser-library/.gitignore
```diff
@@ -0,0 +1 @@
+test
```
radumereuta, line 1: Why is this necessary?
I'm trying to avoid having files being created which are ignored by `git`, and also not cleaned by `make clean`

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -125,6 +125,9 @@ public boolean isKore() {
     @Parameter(names="--bison-stack-max-depth", description="Maximum size of bison parsing stack (default: 10000).")
     public long bisonStackMaxDepth = 10000;
 
+    @Parameter(names="--bison-parser-library", description="Generate a shared library rather than an executable for Bison parsers")
+    public boolean genBisonParserLibrary;
```
radumereuta, line 125: Would it be possible to generate it by default as a shared library and have an executable that uses this library?
I'm only asking because it would eliminate the need for an extra option.
If it complicates the generation or other things, then this solution is fine.

k-distribution/include/kframework/cparser/main.c
```diff
@@ -1,37 +1,102 @@
...
-#include "node.h"
-#include "parser.tab.h"
-#include "scanner.h"
+
+struct string_buffer {
+  FILE *fp;
+  char *buf;
+  size_t capacity;
+  size_t idx;
+};
```
radumereuta, line 17: I'm not very familiar with C, but this seems like there should be something implemented already.

radumereuta (commented):
A few questions.
Java part looks fine.
The C part I'm not that familiar with.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
Baltoli: Yes, that's definitely possible and perhaps even desirable in the future. I'd need to benchmark it, and it would require a backend PR, so I'm inclined to leave it for now.

k-distribution/include/kframework/cparser/main.c
Baltoli: Yeah, the problem is managing dependencies in C - there's similar code in the LLVM backend, but reusing that or using a library would require far more code than just implementing the type here.

k-distribution/tests/regression-new/bison-parser-library/.gitignore
Baltoli: My mistake - should be added to the make clean stanza for the custom test.

k-distribution/include/kframework/cparser/main.c
```diff
@@ -1,37 +1,102 @@
...
+  } else {
+    va_start(args, format);
+    int required = vsnprintf(NULL, 0, format, args) + 1;
+    va_end(args);
+
+    while ((sb->capacity - sb->idx) < required) {
+      string_buffer_grow(sb);
+    }
+
+    va_start(args, format);
```
dwightguth: What is going on with this code? We should not be calling va_start or va_end multiple times in a single function call. There is way to copy va_lists if you need to use it more than once.

Baltoli: Addressed

theo25 (approved):
I went over the C code and the regression test and it all looks good to me.

dwightguth (approved):

PR #3270, opened 2023-03-23
Replacing llvm 14 with llvm 15 as a pre-requirement for K

Robertorosmaninho:
This PR aimed to update the LLVM pre-requirement for K. This new version fixes a bug on lldb that didn't allow macOS users to debug their K definitions.


Baltoli (approved):

Baltoli:
Approved, but can you also add a caveat to the GDB tutorial page that mentions that you may need to upgrade LLVM if you have an older version installed?

PR #3277, opened 2023-03-23
Correct documentation on `simplification` attribute

### LINKED ISSUE #3255: Confusion between `priority` and `simplification` attributes
Currently, we allow users to say something like `simplification(N)`, which is the priority of that simplification rule.

This is confusing in the backend, because we're parsing both the `simplification` and the `priority` attribute.

What may be cleaner is to change it so that to say `simplification(N)`, instead you say `priority(N), simplification`, so that `simplification` is just a simple boolean, and `priority` is what it currently is, an int that defaults to `50`.

ehildenb:
Fixes: https://github.com/runtimeverification/k/issues/3255

Following the discussion about `simplification` attribute being handled in the booster backend, we need to clarify the difference/interaction between the `simplification` and `priority` attribute.


docs/user_manual.md
```diff
@@ -1285,12 +1285,12 @@ rule (X +Int Y) +Int Z => X +Int (Y +Int Z) [simplification]
...
-integer argument which is the rule's _priority_; if the optional argument is not
-specified, it is equivalent to a priority of 50. Simplification rules are
-applied in order of their priority. `simplification` rules may not have the
-`priority` attribute.
+**Order**: The `simplification` attribute accepts an optional integer argument
+which is the rule's _priority_; if the optional argument is not specified, it
+is equivalent to a priority of 50. Simplification rules are applied in order of
+their priority. `simplification` rules may not have the `priority` attribute.
+Unlike the `priority` attribute, this `simplification` index is a _suggestion_
+to the backend, not a semantic requirement.
```
radumereuta: The _suggestion_ part is confusing to me. What does that even mean?
Note that I'm not aware of the discussion you had, so this is what I think of when reading this for the first time.

This text should answer:
 - what is the relationship between simplification rules and normal rules with priority?
 - what does the optional argument do in relation to the previous bullet point?

ehildenb: The first one is answered by "`simplification` rules may not have the `priority` attribute" I believe. That should be the only real relationship. Different backends are free to interpret them differently. The old haskell backend, for example, would apply definition rules first, then simplifications. We probably do the other way around for the new backend.

ehildenb: @radumereuta I've updated the documentation, is it more clear?

docs/user_manual.md
```diff
@@ -1285,12 +1285,16 @@ rule (X +Int Y) +Int Z => X +Int (Y +Int Z) [simplification]
...
-**Order**: Simplification rules are applied after definition rules, if no
-definition rule did apply. The `simplification` attribute accepts an optional
-integer argument which is the rule's _priority_; if the optional argument is not
-specified, it is equivalent to a priority of 50. Simplification rules are
-applied in order of their priority. `simplification` rules may not have the
-`priority` attribute.
+**Order**: The `simplification` attribute accepts an optional integer argument
+which is the rule's _simplification priority_; if the optional argument is not
+specified, it is equivalent to a simplification priority of 50. Backends
+_should_ attempt simplification rules in order of their _simlpification
```
radumereuta: Typo. simlpification -> simplification

docs/user_manual.md
```diff
@@ -1285,12 +1285,16 @@ rule (X +Int Y) +Int Z => X +Int (Y +Int Z) [simplification]
...
+priority_, but are not required to do so; in fact, the backend is free to apply
+`simplification` rules at _any time_. Because of this, users must ensure that
+simplification rules are sound regardless of their order of application. This
+differs from the `priority` attribute in that rules with the `priority`
+attribute _must_ be applied in their priority order by the backend. It is an
+error to have the `priority` attribute on a `simplification` rule.
```
radumereuta, line 1297: You say that simplification rules should be sound regardless of the order they apply.
Then why do I have the option to add a priority? What changes? Do I get a more deterministic execution when possible?
It's not clear from this explanation why should I bother adding those priorities.

@ana-pantilie maybe you have some suggestions here?

ehildenb: Because you want the backend to try one simplification path _before_ it tries another one. It's a heuristic you can supply to the backend.

docs/user_manual.md
ehildenb:
```suggestion
_should_ attempt simplification rules in order of their _simplification
```

docs/user_manual.md
ehildenb: Note that I'm trying to avoid specifying "how" to use it, and more the requirements of the attribute, as this is the manual and not a user guide.

radumereuta (approved):
Not 100% convinced, but I'm not going to block this.
Let's see if someone complains.

PR #3279, opened 2023-03-24
Drop `kprove-legacy`

### LINKED ISSUE #3257: Remove kprove-legacy
Since we are removing the java-backend, we should consider removing kprove-legacy as well.
There are several leftover examples in regression-new that would need to be converted, but w/o the java backend to worry about, it shouldn't be that hard.

Baltoli:
This PR removes all outstanding uses of `kprove-legacy`, and migrates them to `kprove`. Closes #3257 

The changes made are as follows:
* [x] Add `KPROVE_OR_LEGACY=$(KPROVE)` to any tests that immediately support it without any other changes.
* [x] Remove `checkWarnsClaim` - this warning is no longer relevant (rule -> claim deprecation)
* [x] Add the correct or_legacy stanza to tests that also require `--def-module` to be removed
* [x] Remove tests for #1471, which were only relevant to the legacy prover
* [x] Fix tests that needed to be pulled out into a separate verification module
* [x] Remove call to legacy prover in help tests
* [x] Refactor out `KPROVE_OR_LEGACY` uses
* [x] Cut the old script and entrypoints out

The `kprove` module needs to wait until https://github.com/runtimeverification/k/pull/3251 is merged, as there's a dependency between that module and the Java backend. Once the Java backend is gone, we can clean up the `kprove` / `kprovex` modules fully.


radumereuta (commented):
The entire package `kernel/org.framework.kprove` should be removed. Except for the `KProveOpts` file which should be moved to `kprovex` and clean up the outdated options.
Then rename the `kprovex` package to `kprove`.

This is going to cause large conflicts with the removal of the java backend.
So I would suggest we delay until we merge #3251.

Baltoli:
Yes, I tried applying that refactoring earlier and came to the same conclusion. It'll go out cleanly once the Java backend is gone.

radumereuta (approved):

PR #3288, opened 2023-03-29
Deleting axiom production when the it doesn't have any terms.

### LINKED ISSUE #3268: Invalid axioms generated (asserting \bottom{S} unconditionally)
The K-compiler currently generates invalid axioms like `axiom{} \bottom{S}() [constructor{}()] // no junk` (for `S` in `SortSet`, `SortMap`, `SortList`, and `SortK`).
These axioms state that `\bottom` is `\top` and could be used to prove anything we want (the Haskell back-end skips these invalid axioms at the moment).

To reproduce:
```shell
$ echo "module TEST endmodule" > test.k
$ kompile test.k
$ grep -e "axiom.. .bottom" test-kompiled/definition.kore
  axiom{} \bottom{SortList{}}() [constructor{}()] // no junk
  axiom{} \bottom{SortK{}}() [constructor{}()] // no junk
  axiom{} \bottom{SortMap{}}() [constructor{}()] // no junk
  axiom{} \bottom{SortSet{}}() [constructor{}()] // no junk
```

Robertorosmaninho:
Closes #3268

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -704,7 +704,8 @@ public class ModuleToKORE {
     }
 
     private void genNoJunkAxiom(Sort sort, StringBuilder sb) {
-        sb.append("  axiom{} ");
+        StringBuilder sbTemp = new StringBuilder();
+        sbTemp.append("  axiom{} ");
         boolean hasToken = false;
         int numTerms = 0;
         for (Production prod : iterable(mutable(module.productionsForSort()).getOrDefault(sort.head(), Set()).toSeq().sorted(Production.ord()))) {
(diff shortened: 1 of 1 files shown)


jberthold (approved):
LGTM

radumereuta (changes requested):
Can we add a comment to explain this? It's not at all clear what is happening and why this should work.

Robertorosmaninho:
Probably @jberthold or @dwightguth can talk more about why the issue is happening. However, I assume that it's just due to how the function was implemented without considering the case of an empty list of terms to be written.

This implementation should work (to delete this axiom completely) because if it reaches the point to print the terms of the axiom and no term was "discovered", then the only string already printed is the "  axiom{} " that we can safely delete and return the function.

It's important to note that this modification, for now, only cleans the generated `definition.kore` as the Haskell-backend already ignores this kind of axiom production, as mentioned by @jberthold on the original issue.

radumereuta:
I was referring to adding comments in the code.
I don't like that we are hardcoding a number there which works incidentally right now.
To solve this problem, I would:
1. create a new temp StringBuffer at the top of the function,
2. add to it throughout the function
3. at the end, have this test and decide whether to add the contents accumulated to `sb`.

This way, you can modify the code and not rely on the fact that the term has 10 characters long.

radumereuta (approved):

PR #3289, opened 2023-03-30
Add check that labels provided to --concrete-rules correspond to actual rules

### LINKED ISSUE #3285: Check that the label provided to --concrete-rules actually exists
As noted in runtimeverification/evm-semantics#1634, the labels provided to `--concrete-rules` are never checked to ensure they correspond to an actual rule.

For example, define `concrete.k`
```
module CONCRETE-SYNTAX
imports INT-SYNTAX
syntax Exp ::= Int
             | Exp "+" Exp

endmodule

module CONCRETE
imports CONCRETE-SYNTAX
imports INT

rule [assoc]: (X + Y) + Z => X + (Y + Z)

endmodule
```
and run
```
kompile --backend haskell --concrete-rules CONCRETE.fake concrete.k
```
No error or warning is reported.

Scott-Guest:
Closes #3285.

As noted in runtimeverification/evm-semantics#1634, K silently accepts non-existent labels in `--concrete-rules`. This PR adds an error check to ensure each provided label actually has a corresponding rule.

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
+++ b/kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
@@ -1,3 +1,4 @@
+// Copyright (c) K Team. All Rights Reserved.
 package org.kframework.compile;
 
 import org.kframework.attributes.Att;
--- a/kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
+++ b/kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
@@ -1,0 +1,39 @@
+package org.kframework.compile;
+
+import org.kframework.attributes.Att;
+import org.kframework.definition.Definition;
+import org.kframework.definition.DefinitionTransformer;
+import org.kframework.definition.Rule;
+import org.kframework.utils.errorsystem.KEMException;
+
+import javax.annotation.Nullable;
+import java.util.HashSet;
+import java.util.List;
+
+public class MarkExtraConcreteRules {
+
+    /**
+     * Mark with [concrete] rules with labels enumerated in `--concrete-rules`.
+     */
+    public static Definition mark(Definition def, @Nullable List<String> extraConcreteRuleLabels) {
+        if (extraConcreteRuleLabels == null) {
+            return def;
+        }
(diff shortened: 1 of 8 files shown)


kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
```diff
@@ -0,0 +1,40 @@
...
+        if (extraConcreteRuleLabels == null) {
+            return def;
+        }
+        HashSet<String> concreteLabelsSet = new HashSet<>(extraConcreteRuleLabels);
+        Definition result = DefinitionTransformer.fromSentenceTransformer((mod, s) -> {
+            if (s instanceof Rule) {
+                Rule r = (Rule) s;
+                String label = r.att().getOption(Att.LABEL()).getOrElse(() -> null);
+                if (label != null && extraConcreteRuleLabels.contains(label)) {
+                    concreteLabelsSet.remove(label);
```
radumereuta, line 30: Is it possible to have multiple rules with the same label?
I would change this pass to add the used labels to a new collection and compare `!(concreteLabelsSet - usedLabels).isEmpty()`

radumereuta (changes requested):
looks good overall, except for that small change.

kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
radumereuta: Ah, I just noticed `concreteLabelsSet` is a copy.
Then can you rename it to something like `unused`?
I think it makes more sense that way.

radumereuta (approved):

kernel/src/main/java/org/kframework/compile/MarkExtraConcreteRules.java
ehildenb: Rule labels are supposed to be unique. I believe the backends assume this too. It should be an error to have multiple rules with same label.

Scott-Guest: Ah good to know, I'll update the code accordingly. (I also double-checked, and an error is indeed reported if there are non-unique labels).

PR #3295, opened 2023-04-04
Updating to new action upload

F-WRunTime:
What do ya'll think of moving to this new release action? @radumereuta Not sure it will fix the socket issue. I don't see a clear answer on what caused this just yet, but while investigating thought this was a good time to also consider fixing the java warnings. The current action we use is archived and no longer supported as a reminder/note [GH](https://github.com/actions/upload-release-asset).. 

This is the docs for this action if we would like to utilize other features. 
https://github.com/marketplace/actions/upload-to-github-release


F-WRunTime:
Test Run in [Devops-Actions ](https://github.com/runtimeverification/devops-actions/actions/runs/4602885639/jobs/8132339778)

[Release Link ](https://github.com/runtimeverification/devops-actions/releases)

For example utilize 'tag' ?

ehildenb:
The GH action being deprecated: https://github.com/actions/upload-release-asset

Recommends switching to this action: https://github.com/softprops/action-gh-release

Any reason to not use that recommendation?

F-WRunTime:
> The GH action being deprecated: https://github.com/actions/upload-release-asset
> 
> Recommends switching to this action: https://github.com/softprops/action-gh-release
> 
> Any reason to not use that recommendation?

I saw this. I did not like it as it did not have an official 1.0.0 release. Though this is an easy enough transition to use.

ehildenb (approved):

ehildenb:
@F-WRunTime I think that you'll need to also change the source-tarball code to use the new action after you merge to master.

ehildenb:
Let's not merge until this is resolved: https://github.com/runtimeverification/k/pull/3315

ehildenb:
I believe this can be closed now, we have a successful release upload with just using `gh` script, no need for external tool: https://github.com/runtimeverification/k/actions/runs/4643698226

PR #3299, opened 2023-04-05
Switch K master job over to GH Runners

ehildenb:
Currently we're using the Jenkinsfile to trigger release creation on master pushes.

This switches us to GH Runners, and this should be the last thing in this repo using Jenkins.


F-WRunTime (approved):

ehildenb:
I'm merging this without testing, will revert if I don't have a fix for breakages by the end of the day.

PR #3310, opened 2023-04-07
Set release title with gh release creation

ehildenb:
Because the title field was missing, `gh` command was actually just creating a tag, not a release. So then the action which uploads the nix data to the release was not able to find the appropriate place to upload the results: https://github.com/runtimeverification/k/actions/runs/4634493287/jobs/8202728444


ehildenb:
Setting the title did indeed create a proper release: https://github.com/runtimeverification/k/releases/tag/v5.6.0-test

Which triggered this job to be able to upload teh Nix assets correctly: https://github.com/runtimeverification/k/actions/runs/4635712481

F-WRunTime (approved):

Issue #3035, opened 2022-11-08
Remove "antiLeft" priority predicates from rewrite rules


ana-pantilie:
Rewrites rules which get assigned priorities get parsed into Kore which looks like:
```
  axiom{} \rewrites{SortGeneratedTopCell{}} (
    \and{SortGeneratedTopCell{}}(
      \not{SortGeneratedTopCell{}}(priorityLE2{}()),
      rule38LHS{}(VarX:SortInt{},Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'DotVar1:SortK{})),
    \and{SortGeneratedTopCell{}} (
      \top{SortGeneratedTopCell{}}(), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortS{}, SortKItem{}}(Lblone'Unds'PRIORITY'Unds'S{}()),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})))
  [org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/Users/anapantilie/RV/haskell-backend/test/priority/priority.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(13,8,13,23)"), priority{}("3"), UNIQUE'Unds'ID{}("6cd5667589c7d76f6e1b18e23fa939ee02476bb24c4229d69331b460c555490c")]
```

That `\not{SortGeneratedTopCell{}}(priorityLE2{}())` encodes the priority of the rewrite rule into Kore, and it was deemed necessary a long time ago for constructing proof objects. This is no longer one of our goals, and the Haskell backends do not (and shouldn't) use this priority predicate in any way. I suggest we modify the frontend to avoid producing these predicates in the first place.

Can we also avoid keeping the actual LHS of rules in those `rule38LHS{}` aliases? Why do we need this?

- [x] Frontend flag to generate semantics axioms in new way. For a `rule LHS => RHS requires REQ ensures ENS`, we should generated `\rewrites(\and(LHS, REQ), \and(RHS, ENS))`. No aliases, no antileft (`#Not ...`) pattern.
- [x] runtimeverification/haskell-backend#3567
- [x] Same for LLVM backend.
[...]

PR #3314, opened 2023-04-07
Add flag to disable antiLeft priority predicates in KORE

Baltoli:
Related to: https://github.com/runtimeverification/k/issues/3035

This PR modifies ModuleToKORE such that a simplified KORE format can be generated for semantic rewrite rules when `--disable-kore-antileft` is passed. Rather than encoding rule priorities into explicit groups within the KORE theory, this new form just generates constrained terms for rewrites, meaning that a KORE backend must interpret the priority attribute to rewrite correctly.

As expected, the LLVM backend behaves correctly in the presence of this change (at least for a simple test case). A test case for the flag on the LLVM backend is added to the regression test suite, which gives some degree of assurance that the changes in this PR are at least well-formed with respect to the output KORE. The Haskell backend crashes when trying to rewrite the new rules; I have not investigated further.

For ease of reviewing, consider this input definition:
```k
module TEST
  imports BOOL
  imports INT

  syntax Foo ::= foo(Int) | bar() | baz()

  rule foo(I) => bar() requires I  >Int 0 [priority(40)]
  rule foo(I) => baz() requires I >=Int 0 [priority(60)]
endmodule
```

Previously, for the rule with priority 60, we would generate KORE like:
```
  alias rule3LHS{}(SortInt{},SortGeneratedCounterCell{},SortK{}) : SortGeneratedTopCell{}
  where rule3LHS{}(VarI:SortInt{},Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'DotVar1:SortK{}) :=
    \and{SortGeneratedTopCell{}} (
      \equals{SortBool{},SortGeneratedTopCell{}}(
        Lbl'Unds-GT-Eqls'Int'Unds'{}(VarI:SortInt{},\dv{SortInt{}}("0")),
        \dv{SortBool{}}("true")), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortFoo{}, SortKItem{}}(Lblfoo'LParUndsRParUnds'TEST'Unds'Foo'Unds'Int{}(VarI:SortInt{})),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})) []

  axiom{} \rewrites{SortGeneratedTopCell{}} (
    \and{SortGeneratedTopCell{}}(
      \not{SortGeneratedTopCell{}}(priorityLE40{}()),
      rule3LHS{}(VarI:SortInt{},Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'DotVar1:SortK{})),
    \and{SortGeneratedTopCell{}} (
      \top{SortGeneratedTopCell{}}(), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortFoo{}, SortKItem{}}(Lblbaz'LParRParUnds'TEST'Unds'Foo{}()),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})))
  [org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/bruce/code/scratch/test.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody \"requires\" Bool [klabel(#ruleRequires), symbol]"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(8,8,8,42)"), priority{}("60"), UNIQUE'Unds'ID{}("3086847842e848918ea69e0d9f1189d6e073b39807e55d7c08eb4abc3ad485e4")]
```

The new-format KORE generated for the same rule is:
```
  axiom{} \rewrites{SortGeneratedTopCell{}} (
      \and{SortGeneratedTopCell{}} (
        \equals{SortBool{},SortGeneratedTopCell{}}(
        Lbl'Unds-GT-Eqls'Int'Unds'{}(VarI:SortInt{},\dv{SortInt{}}("0")),
        \dv{SortBool{}}("true")),
        Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortFoo{}, SortKItem{}}(Lblfoo'LParUndsRParUnds'TEST'Unds'Foo'Unds'Int{}(VarI:SortInt{})),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})),
      \and{SortGeneratedTopCell{}} (
      \top{SortGeneratedTopCell{}}(), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortFoo{}, SortKItem{}}(Lblbaz'LParRParUnds'TEST'Unds'Foo{}()),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})))
  [org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/bruce/code/scratch/test.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody \"requires\" Bool [klabel(#ruleRequires), symbol]"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(8,8,8,42)"), priority{}("60"), UNIQUE'Unds'ID{}("3086847842e848918ea69e0d9f1189d6e073b39807e55d7c08eb4abc3ad485e4")]
```

The indentation isn't ideal, but fixing that would require far more invasive changes to the code generator than are warranted for this PR. The test case added verifies that this change generates valid KORE.


ehildenb:
@Baltoli one of the concerns was pattern matching out the config from the constraint. Currently for a rule:

```
rule LHS => RHS requires REQ ensures ENS
```

It seems to be outputting:

```
\rewrites(\and(REQ, LHS), \and(ENS, RHS))
```

Can we instead have the configuration coming first?

```
\rewrites(\and(LHS, REQ), \and(RHS, ENS))
```

Would the LLVM backend be OK with that? The Haskell backend will have to change either way, but I think this will allow them to have more strict matching of the components of the rule, rather than searching through the constraints for the correct one.

Baltoli:
Will implement that change next week - _should_ be OK but will have to double check!

Baltoli:
Update - the LLVM backend _isn't_ OK with the reordered rules.

Baltoli:
LLVM backend support PR is here: https://github.com/runtimeverification/llvm-backend/pull/729; this PR is blocked until that change is upstreamed to K.

ana-pantilie:
This PR also depends on https://github.com/runtimeverification/haskell-backend/pull/3568 because `kompile` calls `kore-exec` for definition serialization.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1343,6 +1364,16 @@ private void genSemanticsRuleLHSWithAlias(String ruleAliasName, List<KVariable>
         }
     }
 
+    private void genSemanticsRuleLHSNoAlias(K requires, K left, List<KVariable> freeVars, String topCellSortStr,
+                                            String previousGroupName, StringBuilder sb)
+    {
```
radumereuta: You wrote too much C :))
Open brace should be on the previous line.

radumereuta (approved):
The java part lgtm
Let's see if we can make it work on the backends.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
ana-pantilie:
```suggestion
    private void genSemanticsRuleLHSNoAlias(K requires, K left, List<KVariable> freeVars, String topCellSortStr,
                                            String previousGroupName, StringBuilder sb) {
```

PR #3319, opened 2023-04-07
Fix release tarball upload

ehildenb:
Fixes: https://github.com/runtimeverification/k/actions/runs/4640548932/jobs/8212601256


F-WRunTime (approved):
Interesting it requires repository, thought it would assume that with the 'checkout' action.

PR #3324, opened 2023-04-10
Provide better documentation for concrete and symbolic attributes

ana-pantilie:
Fixes https://github.com/runtimeverification/haskell-backend/issues/3564


docs/user_manual.md
```diff
@@ -1344,20 +1344,39 @@ warning if it encounters such a claim.
...
+arguments are to match.
+
+A concrete pattern is a pattern which does not contain variables or unevaluated
+functions, otherwise the pattern is symbolic.
+
+The semantics of the two attributes is defined as follows:
+- If a simplification rule is marked `concrete`, then _all_ arguments must be
+concrete for the rule to match.
+- If a simplification rule is marked `symbolic`, then the rule will match only
+if _at least one_ argument is symbolic.
```
radumereuta: Is that correct? From @virgil-serbanuta comment on the issue I get that all arguments need to be symbolic.

radumereuta:
@virgil-serbanuta I'm tagging you for review since you are the one that encountered the issue.

docs/user_manual.md
ana-pantilie: I understood the opposite from the discussion on the issue and on Slack. I'll take another look, thank you for pointing this out!

docs/user_manual.md
```diff
@@ -1344,20 +1344,39 @@ warning if it encounters such a claim.
...
+
+For example, the following will only match when all arguments
 are concrete:
 
 ```k
 rule X +Int (Y +Int Z) => (X +Int Y) +Int Z [simplification, concrete]
 ```
 
-These rules will re-associate and commute terms to combine concrete arguments:
+Conversely, the following will only match when _all_ arguments
```
ana-pantilie:
```suggestion
Conversely, the following will only match when all arguments
```

ana-pantilie:
Thanks @radumereuta , I should have taken a better look at Virgil's example and at the actual code! I can confirm it's as you understood, so this is ready for re-review. Also @virgil-serbanuta please review to make sure I didn't get it wrong again :).

radumereuta (commented):
What happens if you put both concrete and symbolic on the same rule?
I would expect it to be an error.
What if you give them arguments and only half of the variables are concrete and the other half are symbolic. Then I would expect that to work.
Does that make sense? Does the backend support such a behavior?

ana-pantilie:
@radumereuta Hmm, good questions.

If you add `symbolic` and `concrete` at the same time the frontend catches it and emits an error:
```
[Error] Compiler: Rule cannot be both concrete and symbolic in the same
variable.
        Source(/home/ana/RV/haskell-backend/test/imp/imp.k)
        Location(223,8,223,29)
        223 |     rule X:Int +Int Y:Int => Y [simplification, symbolic, concrete]
            .          ^~~~~~~~~~~~~~~~~~~~~
```

The following, however, isn't caught by the frontend and I think it should be:
```
rule X:Int +Int Y:Int => Y [simplification, symbolic(X), concrete(X)]
```

The following works fine, as it should:
```
rule X:Int +Int Y:Int => Y [simplification, symbolic(X), concrete(Y)]
```

ehildenb:
The rules for `concrete` and `symbolic` are different for axioms and for simplification rules.

- For axioms: if you have concrete/symbolic on _any_ of the rules for a given production, it must be the same on _all_ the rules for that production. No calling out of individual variables allowed. https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/checks/CheckKLabels.java#L152 (see discussion: https://github.com/runtimeverification/k/issues/1591)
- For simplification: Each rule must be individually sound anyway. So you can have independent `concrete(...), symbolic(...)` attributes with on cross-checking between rules.

radumereuta:
I've added a better error message for when a variable is used on both `symbolic` and `concrete` attributes.
@ana-pantilie can you please review that part? I'm going to approve it. Add `automerge` if everything looks good.

radumereuta (approved):

virgil-serbanuta:
Thank you!

Sorry for the late reply, I was on vacation, but this looks good!

PR #3328, opened 2023-04-10
add tutorial lesson 2.1

dwightguth:
This is ready for review and can be merged as soon as everyone has finished reviewing it and any comments are addressed.


k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+
+1. Top-level rewrite rules, which rewrite a configuration composed of cells to
+another configuration;
+2. Function rules, which define the behavior of a function written over
+arbitrary input and output types; and
+3. Simplification rules, which describe ways in which the symbolic execution
+engine ought to simplify terms containing symbolic values.
+
+This lesson introduces three more types of rules, of which the first are
+**macros**. A production is a macro if it has the `macro` attribute, and all
```
Baltoli:
```suggestion
This lesson introduces three more types of rules, the first of which are
**macros**. A production is a macro if it has the `macro` attribute, and all
```

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+endmodule
+```
+
+In this example, we see that an if statement without an `else` clause is
+defined in terms of one with an `else` clause. As a result, we would only
+need to give a single rule for how to rewrite `if` statements, rather than
+two separate rules for two types of `if` statements. This is a common pattern
+for dealing with program syntax that contains an optional component to it.
+
+It's worth noting, by default, macros are not applied recursively. To be more
```
Baltoli:
```suggestion
It is worth noting that by default, macros are not applied recursively. To be more
```

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+
+**NOTE**: future lessons will discuss KORE, an intermediate language for K,
+which is based on Matching Logic. KORE defines a concept called an "alias"
+which is similar, but not exactly the same as the concept we are about to
+introduce, which is a particular variant of a macro. It's worth remembering
+to keep these two concepts separate in your mind.
+
+Macros can be very useful in helping you define a programming language.
+However, they can be disruptive while pretty printing a configuration. For
+example, I might write a set of macros that transforms the code the user wrote
```
Baltoli:
```suggestion
example, you might write a set of macros that transforms the code the user wrote
```

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+
+K defines a relatively straightforward but novel solution to this problem,
+which is known as a K **alias**. An alias in K is very similar to a macro,
+with the exception that the rewrite rule will also be applied **backwards**
+during the pretty-printing process.
+
+It is very simple to make a production be an alias instead of a macro: simply
+use the `alias` or `alias-rec` attributes instead of the `macro` or `macro-rec`
+attributes. For example, if I were to declare the example involving `if`
+statements above using an alias instead of a macro, the `Stmt` term
```
Baltoli:
```suggestion
attributes. For example, if the example involving `if` statements
above was declared using an alias instead of a macro, the `Stmt` term
```

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+with how various terms are pretty-printed by invoking `krun` on them.
+
+## `anywhere` rules
+
+There is one other type of rule that this lesson intends to introduce, which
+is called the **anywhere rule**. An anywhere rule is specified by adding the
+`anywhere` attribute to a rule. Such a rule is similar to a function rule
+in that it does not participate in cell completion, and will apply anywhere
+that the left-hand-side matches in the configuration, but distinct in that the
+symbol in question is still possible to be matched on in the left-hand side
```
Baltoli:
```suggestion
symbol in question can still be matched in the left-hand side
```

Baltoli (commented):
Some very minor comments on phrasing; otherwise looks great!

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+important that rules be defined correctly to always terminate, otherwise the
+macro expansion phase will run forever. In practice, the types of macros
+usually written are very easy to ensure this property on, however.
+
+### Exercise
+
+Using a `Nat` sort containing the constructors `0` and `S` (i.e., a
+peano-style axiomatization of the natural numbers where `S(N) = N + 1`,
+`S(S(N)) = N + 2`, etc), write a macro that will compute the sum of two
+numbers.
```
ehildenb: I think it's worth linking to an article on Peano arithmetic, maybe here: https://wiki.haskell.org/Peano_numbers

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+
+It is very simple to make a production be an alias instead of a macro: simply
+use the `alias` or `alias-rec` attributes instead of the `macro` or `macro-rec`
+attributes. For example, if I were to declare the example involving `if`
+statements above using an alias instead of a macro, the `Stmt` term
+`if (E) {} else {}` would be pretty-printed as `if (E) {}`. This is because
+during pretty-printing, the term participates in another macro-expansion pass.
+However, this macro expansion step will only apply rules with the `alias`
+or `alias-rec` attribute, and, critically, it will treat the left-hand side
+of such rules as if they were the right-hand side, and vice versa.
```
ehildenb:
```suggestion
or `alias-rec` attribute, and, critically, it will _reverse_ the rule by treating the left-hand side
of such rules as if they were the right-hand side, and vice versa.
```

ehildenb (approved):

dwightguth:
@Scott-Guest we would like you to do this lesson and let us know if you encounter any issues or if any of the wording is unclear to you. I will trade off between you and Roberto so you aren't overloaded on top of your other tasks.

Baltoli (approved):

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+arbitrary input and output types; and
+3. Simplification rules, which describe ways in which the symbolic execution
+engine ought to simplify terms containing symbolic values.
+
+This lesson introduces three more types of rules, the first of which are
+**macros**. A production is a macro if it has the `macro` attribute, and all
+rules whose top symbol on the left hand side is a macro are **macro rules**
+which define the behavior of the macro. Like function rules and simplification
+rules, macro rules do not participate in cell completion. However, unlike
+function rules and simplification rules, macro rules are applied **statically**
```
Scott-Guest, line 28: This point (that function rules and simplification rules don't participate in cell completion) was never explicitly mentioned in the earlier lessons and could use some elaboration.

Roughly, we could
- In Lesson 1.13, more clearly emphasize that (non-contextual) function rules are independent of any particular cells. Instead, they are used to reduce a "func(args)" term regardless of where it appears in the overall configuration.
    - Surprisingly, we never really state this explicitly anywhere right now
- In Lesson 1.15, remark that function rules don't participate in configuration completion (because they are independent of the configuration).
- In Lesson 1.22, make a similar point about simplification rules.

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+natural numbers where `S(N) = N + 1`, `S(S(N)) = N + 2`, etc), write a macro
+that will compute the sum of two numbers.
+
+## Aliases
+
+**NOTE**: future lessons will discuss KORE, an intermediate language for K,
+which is based on Matching Logic. KORE defines a concept called an "alias"
+which is similar, but not exactly the same as the concept we are about to
+introduce, which is a particular variant of a macro. It's worth remembering
+to keep these two concepts separate in your mind.
```
Scott-Guest: Personal preference, but I think we could just leave this out, or say something terser like "**NOTE**: The use of "alias" here refers to a particular variant of macro. While similar, note that this is distinct from the concept of "alias" that we will later introduce in Lesson X.Y"

The presumption is that lessons are completed in order, so we're effectively saying "don't confuse this with something you haven't heard of yet" while half-heartedly introducing KORE.

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+
+### Exercise
+
+Modify `LESSON-01` above to use an alias instead of a macro and experiment
+with how various terms are pretty-printed by invoking `krun` on them.
+
+## `anywhere` rules
+
+There is one other type of rule that this lesson intends to introduce, which
+is called the **anywhere rule**. An anywhere rule is specified by adding the
```
Scott-Guest: Nitpick: This reads slightly awkward to me. Maybe rephrase as "The last type of rule we will introduce in this lesson is the **anywhere rule**."

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+`anywhere` attribute to a rule. Such a rule is similar to a function rule
+in that it does not participate in cell completion, and will apply anywhere
+that the left-hand-side matches in the configuration, but distinct in that the
+symbol in question can still be matched against in the left-hand side of other
+rules, even during concrete rewriting. The reasoning behind this is that
```
Scott-Guest: Looking back, I don't think the earlier lessons ever actually state this restriction on function symbols appearing in the LHS of other rules. We should add something about this to Lesson 1.13

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+endmodule
+```
+
+In this example, we see that an if statement without an `else` clause is
+defined in terms of one with an `else` clause. As a result, we would only
+need to give a single rule for how to rewrite `if` statements, rather than
+two separate rules for two types of `if` statements. This is a common pattern
+for dealing with program syntax that contains an optional component to it.
+
+It is worth noting that by default, macros are not applied recursively. To be
```
Scott-Guest, line 62: Missing a comma: "It is worth noting that, by default, ..."

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+more precise, by default a macro that arises as a result of the expansion of
+the same macro is not rewritten further. This is primarily to simplify the 
+macro expansion process and reduce the risk that improperly defined macros will
+lead to non-terminating behavior.
+
+It is possible, however, to tell K to expand a macro recursively. To do this,
+simply replace the `macro` attribute with the `macro-rec` attribute. Note that
+K does not do any kind of checking to ensure termination here, so it is
+important that rules be defined correctly to always terminate, otherwise the
+macro expansion phase will run forever. In practice, the types of macros
```
Scott-Guest: Nitpick: This sentence feels awkward. Maybe rephrase as "Fortunately, in practice, it's very easy to ensure this property for the types of macros that are typically written."

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+rules whose top symbol on the left hand side is a macro are **macro rules**
+which define the behavior of the macro. Like function rules and simplification
+rules, macro rules do not participate in cell completion. However, unlike
+function rules and simplification rules, macro rules are applied **statically**
+before rewriting begins, and the macro symbol is expected to no longer appear
+in the initial configuration for rewriting once all macros in that
+configuration are rewritten.
+
+The rationale behind macros is they allow you to define one piece of syntax
+in terms of another piece of syntax without any runtime overhead associated
```
Scott-Guest, line 34: Might be worth mentionng that this is a common practice in PLs called "desugaring"

k-distribution/k-tutorial/2_intermediate/01_macros/README.md
```diff
@@ -4,6 +4,151 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+with the cost of rewriting one to the other. For example, in a language with
```
Scott-Guest: Might be worth mentioning that this is a common practice in PLs called "desugaring". Would make it easier for the reader to google related examples where these macro rules could be useful

Scott-Guest (approved):
A few small suggestions, but LGTM overall.

I also commented on a few places where I think we actually need to add further explanation to previous lessons.

ehildenb (approved):

Baltoli (approved):

PR #3338, opened 2023-04-13
Pt.1 change to push location for mvn packages

F-WRunTime:
Step 1 to update distributionManagement in Pom for first release.


.github/actions/test-package/action.yml
```diff
@@ -25,6 +25,18 @@ inputs:
...
+    required: false
+    default: 'false'
+  mv-deploy-id:
+    description: 'Maven deploy id: runtime.verification or runtime.verification.snapshots'
+    required: false
+    default: 'runtime.verification.snapshots'
+  token:
+    description: 'GitHub token for publishing to GitHub Maven Packages.'
+    required: false
+    default: ''
```
ehildenb: I would still not include it in the action, rather just in the overall release workflow. We know it only has to run once, so just one call to the mvn-deploy action should be enough, no need to make this action handle more things.

F-WRunTime: Moving to release file.

F-WRunTime:
An error in using submodules; recursive in actions/checkout@v3 [fails here ](https://github.com/runtimeverification/elrond-semantics/actions/runs/4672723422/jobs/8275200835)

A simple workaround is to run the explicit git command.

dwightguth (commented):
Why are there two different calls to mvn deploy in this workflow? As far as I can tell, one is Linux and the other is MacOS, but I don't believe maven actually deploys anything that is platform-specific, so you ought to need to do it only once during the actual final creation of the release.

F-WRunTime:
Updated to remove mvn push in Mac build steps

.github/workflows/release.yml
```diff
@@ -106,6 +118,9 @@ jobs:
           docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'cd ~ && kompile test.k --backend llvm'
           docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'cd ~ && kompile test.k --backend haskell'
           docker image push ${DOCKERHUB_REPO}:${version_tag}
+          docker cp ~/.m2/settings.xml k-package-docker-test-jammy-${GITHUB_SHA}:~/.m2/settings.xml
```
ehildenb, line 121: I don't understand tihs `docker cp ...` command? Are we copying a file from the host into the docker image? Why? Shouldn't everything happen inside the docker container?

ehildenb: Even more confusing, why is this coming from the host machines home directory, rather than some specific file in teh K repo?

dwightguth: This file contains secrets like the maven username and password, so it gets initialized by the CI and it's getting copied into the docker image where it's being used.

dwightguth (approved):

Issue #3086, opened 2022-12-13
[K-Bug] Rule containing Unicode token is not correctly kompiled


tothtamas28:
### What component is the issue in?

Front-End

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v5.5.46-0-gb2a816206d

### Operating System

Linux

### K Definitions (If Possible)

```k
// emoji.k

module EMOJI-SYNTAX
    syntax Emoji ::= r"🙁|🙂" [token]
endmodule

module EMOJI
    imports EMOJI-SYNTAX
    rule <k> 🙁 => 🙂 ... </k>
endmodule
```

#### Haskell backend does not kompile rule correctly

```
$ kompile emoji.k --backend haskell
```

Then in `compiled.txt`:

```
rule `<generatedTop>`(`<k>`(``#token("\U0001f641 =","Emoji")=>#token(" ...","Emoji")``~>_DotVar1),_DotVar0) requires #token("true","Bool") ensures #token("true","Bool") [...]
```

Accordingly in `definition.kore`:

```
alias rule0LHS{}(SortGeneratedCounterCell{},SortK{}) : SortGeneratedTopCell{}
where rule0LHS{}(Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'DotVar1:SortK{}) :=
  \and{SortGeneratedTopCell{}} (
    \top{SortGeneratedTopCell{}}(), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(inj{SortEmoji{}, SortKItem{}}(\dv{SortEmoji{}}("\U0001f641 =")),Var'Unds'DotVar1:SortK{})),Var'Unds'DotVar0:SortGeneratedCounterCell{})) []

axiom{} \rewrites{SortGeneratedTopCell{}} (
  rule0LHS{}(Var'Unds'DotVar0:SortGeneratedCounterCell{},Var'Unds'DotVar1:SortK{}),
  \and{SortGeneratedTopCell{}} (
[...]

PR #3341, opened 2023-04-14
Fix incorrect #token construction for Unicode characters

Scott-Guest:
During parsing, the location data is computed to refer to indices within a UTF-8 byte string, but then are used as indices into a Java string when constructing a #token.

This PR fixes this bug, which was partially responsible for #3086


Scott-Guest:
For context, the input is presumed to be UTF-8 when read in here: https://github.com/runtimeverification/k/blob/f723b25866b4c975baba906058169fc533c4fbb6/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java#L420

The startLoc and endLoc are then read from the Flex scanner output here: https://github.com/runtimeverification/k/blob/f723b25866b4c975baba906058169fc533c4fbb6/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java#L444

Flex generates 8-bit scanners (i.e. treats every 8-bits as one character), so the location data refers to a byte index

dwightguth (approved):

Issue #2707, opened 2022-07-07
K Tutorial on literate programming needs to be viewed from source


charala1:
Viewed on a browser, the section doesn't make sense because tags aren't visible. We need to either
a) make it explicit that it should be viewed from source (and include a link), or
b) fix what the user sees and forget about the file itself being correct in terms of being parse-able

Here is what it currently looks like: https://kframework.org/k-distribution/k-tutorial/1_basic/08_literate_programming/

radumereuta:
@shd101wyy can we do something in the web interface?

PR #3349, opened 2023-04-20
Display line numbers for all code blocks

yiyi-wang-rv:
Currently blocked by https://github.com/runtimeverification/k-web-theme/pull/32

Hi @fiedlr , could you please help check if the demo website https://yiyi-wang-rv.github.io/k looks okay to you? Thank you.


fiedlr:
@yiyi-wang-rv Looks good except that the dot is missing when there is only one tag, like `k` vs. `.k .exclude` . Either add the dot here too, or when there is only one tag, hide it as we discussed. Is this going to be in another PR?

yiyi-wang-rv:
Hi @fiedlr . Cool. Is it okay if I just stop here and you guys continue to change the markdown files? Thank you!

fiedlr:
@yiyi-wang-rv Yes, but can you finish the hiding when there's only one tag, please?

yiyi-wang-rv:
@fiedlr Could you please help push the changes to this branch directly. I will continue tomorrow. Thank you!

----

Oh I misunderstood. I will continue tomorrow

k-distribution/tests/regression-new/markdownSelectors/test.md
```diff
@@ -28,7 +28,7 @@ module TEST
   rule <k> 1 => 2 </k>
 ```
 
-```{.k .discard .numberLines startFrom="0"}
```
fiedlr: Please keep it here as it tries the new syntax.

yiyi-wang-rv:
> @yiyi-wang-rv Looks good except that the dot is missing when there is only one tag, like k vs. .k .exclude . Either add the dot here too, or when there is only one tag, hide it as we discussed. Is this going to be in another PR?

Isn't this what we are supposed to change in the markdown files?

Like change

````
```k
```
````

to

````
```{.k}
```
````

Are we going to hide tags if there is only one tag?

File: k-distribution/tests/regression-new/markdownSelectors/test.md (hunk 1 above)
Comment (yiyi-wang-rv) [reply k-distribution/tests/regression-new/markdownSelectors/test.md]: Okay, I will revert all changes to markdown files.

**Comment by fiedlr** [comment]
@yiyi-wang-rv I thought 

```
```k
```

is a sugar for

```
```{.k} 
```

yiyi-wang-rv:
@fiedlr Okay I reverted changes to markdown files. 

I will make changes to hide tags when there is only one tag.

k-distribution/tests/regression-new/markdownSelectors/test.md
fiedlr: No, I meant only in this `test.md` file

fiedlr:
@yiyi-wang-rv Yes, I thought that we would hide the panel if there is only one tag, e.g., it is either 

```
```k
```

or

```
```{.k} 
```

Let's leave it for tomorrow, it must be late for you.

k-distribution/tests/regression-new/markdownSelectors/test.md
yiyi-wang-rv: Ah okay. I will make the changes.

yiyi-wang-rv:
I would suggest we keep our current behavior. 

* Removing the tag if there is only one tag will break what we want in issues like: https://github.com/runtimeverification/k/issues/2707. I think we want to display the tag even though there is only one tag. And there might be code blocks in different languages like `c`, `python`, etc. Hiding the tag will make it unclear what language the code block is.  
* Appending `.` to a tag that doesn't start with `.` will make people confused. For example, on page https://kframework.org/k-distribution/k-tutorial/1_basic/08_literate_programming/ , for the first code block if we change  the display of `k` to `.k`, then it contracts with its explanation above:

  > A K code block begins with a line of text containing the keyword <code>```k</code>, and ends when it encounters another <code>```</code> keyword.

----

@fiedlr Please help review https://yiyi-wang-rv.github.io/k to check if it looks okay to you. I just updated it. Thank you!

web/package-lock.json
```diff
@@ -42,6 +42,7 @@
         "markdown-it": "^12.0.2",
         "markdown-it-anchor": "^6.0.1",
         "markdown-it-attrs": "^4.1.0",
+        "markdown-it-footnote": "^3.0.3",
```
fiedlr, line 45: Why are you adding this?

fiedlr (approved):

web/package-lock.json
yiyi-wang-rv: The `k-web-theme` submodule is shared among many other websites. So some websites use that footnote feature.

yiyi-wang-rv:
Sorry we will need to merge https://github.com/runtimeverification/k-web-theme/pull/32 for `k-web-theme` first, then update the `k-web-theme` to the latest `master` here in `k`. I will add the `automerge` label later :+1:

PR #3351, opened 2023-04-20
Fixing missing directory in temp container for m2 settings.xml

F-WRunTime:
There is a break in release due to the missing file. Fixing this issue here.


F-WRunTime:
@ehildenb per a comment you made on a previous PR there is an option to dump the settings.xml to a different location. 
`settings-path: Maven related setting to point to the directory where the settings.xml file will be written. Default is ~/.m2.`
We can do something like this for setup-java:
```
settings-path: ${{ env.GITHUB_WORKSPACE }}
```
Then on the docker part 
```
          docker cp ${WORKSPACE}/settings.xml k-package-docker-test-jammy-${GITHUB_SHA}:~/.m2/settings.xml
```

ehildenb (approved):

PR #3352, opened 2023-04-20
Update Matrix links for use in any client

jryans:
Riot is an old name for a Matrix client (now called Element). This updates the various Matrix links to use the generic `matrix.to` share link format instead of embedding specific client in the link.


jryans:
Hmm, let me try again starting from develop... 😅

PR #3356, opened 2023-04-20
Add check for manual declaration of isSort

### LINKED ISSUE #3329: Incorrect code generated when redeclaring sort predicate
This definition crashes the LLVM code generator by generating invalid LLVM somehow, related to the spurious redeclaration of `isNonAddr`.

I think this should maybe be caught in the frontend - when is it valid to define your own `isSort` predicates?

```k
module TYPEASSERTION
    imports BOOL-SYNTAX
    imports BASIC-K

    syntax Bool ::= isNonAddr(KItem) [function]
    syntax NonAddr ::= nonAddr(KItem)
endmodule
```

Scott-Guest:
Closes #3329 

This PR adds a check to report an error if the user does both of the following:
- declares a sort `syntax S`
- declares a symbol `syntax Bool ::= isS ( _ )`

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkIsSort.k
+++ b/k-distribution/tests/regression-new/checks/checkIsSort.k
@@ -1,3 +1,4 @@
+// Copyright (c) K Team. All Rights Reserved.
 module CHECKISSORT-SYNTAX
   imports BOOL-SYNTAX
   imports BASIC-K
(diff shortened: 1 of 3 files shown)


k-distribution/tests/regression-new/checks/checkIsSort.k
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) K Team. All Rights Reserved.
+module CHECKISSORT-SYNTAX
+  imports BOOL-SYNTAX
+  imports BASIC-K
+
+  syntax Bool ::= isNonAddr(KItem) [function]
+  syntax NonAddr ::= nonAddr(KItem)
```
radumereuta, line 15: Can you also add a declaration that is correct?
This way we can test for a positive case.

radumereuta (approved):

PR #3364, opened 2023-04-26
Updating mint-llvm Makefile and test from Haskell to LLVM Backend

Robertorosmaninho:
- Setting the llvm backend instead of haskell
- Importing INT to make it executable. Ex.: 
```
krun -cPGM='2*2'
<k>
  4p8 ~> .
</k>
```

Previous behavior was: 
```
krun -cPGM='2*2'
[Error] Inner Parser: Scanner error: unexpected character sequence '2'.
        Source(/home/robertorosmaninho/rv/k/k-distribution/tests/regression-new/mint-llvm/./.krun-2023-04-26-08-54-53-SItcx3JHTL/tmp.PGM.pretty.QKfhm9bCxk)
        Location(1,1,1,2)
        1 |     2*2
          .     ^
[Error] kparse: /home/robertorosmaninho/rv/k/k-distribution/bin/kast --output 
kore --definition ./test-kompiled --module TEST 
.krun-2023-04-26-08-54-53-SItcx3JHTL/tmp.PGM.pretty.QKfhm9bCxk
[Error] krun: kparse --definition ./test-kompiled --module TEST 
.krun-2023-04-26-08-54-53-SItcx3JHTL/tmp.PGM.pretty.QKfhm9bCxk
```


radumereuta (changes requested):
Can you also add some tests, please?
Something that tests a positive case, maybe a negative case, and something with an overflow, just to make sure we're using the correct implementation.

dwightguth:
@Robertorosmaninho there are already tests you can reconstruct in the llvm backend repo version of this test, which is test30.k. You will need to reconstruct the unparsed version of those tests, since they exist as KORE in the llvm backend repo, but that's not very hard to do.

dwightguth:
Looks good otherwise though.

Robertorosmaninho:
Actually, @dwightguth, the [test30.k](https://github.com/runtimeverification/llvm-backend/blob/master/test/defn/k-files/test30.k) is almost equal to the [test.k](https://github.com/runtimeverification/k/blob/25964b04ffbe41f3f8f757588e6a40b74bf47777/k-distribution/tests/regression-new/mint-llvm/test.k) that I'm already modifying in this PR. So, I'm not sure which more tests we should have here different than this already created to test the mint implementation.

dwightguth:
@Robertorosmaninho the idea is that we want the tests that execute the program to exist in both the frontend and the backend, but currently, the frontend test only compiles the definition and doesn't execute anything.

Robertorosmaninho:
@dwightguth  @radumereuta, the outputs of these tests are in `MINT{8}` instead of being cast again to `INT` as the llvm tests. Should I keep it this way (not modifying the test logic) or change it to have the same behavior as `test30.k`?

k-distribution/tests/regression-new/mint-llvm/13.test.out
```diff
@@ -0,0 +1,3 @@
+<k>
+  true ~> .
+</k>
```
radumereuta, line 3: Is this supposed to be true?
This is also true: `127 > -1`
and `- 1 == 255p8`

radumereuta (approved):
I'm happy with the front end tests.
Before merging, please make sure the test I mentioned is correct.

k-distribution/tests/regression-new/mint-llvm/13.test.out
dwightguth: Yes, this is correct. The MInt module provides two different comparison operators. One is signed comparison, the other is unsigned comparison. The reason both exist is exactly because of this case. But the semantics of this test specifically uses signed comparison.

For comparison, 255 > 127 would return false.

dwightguth (approved):

PR #3366, opened 2023-04-26
Fix can't find mvn

radumereuta:
I'm hoping this is why it can't find `mvn`.

Right now the jammy deploy fails with
```
+ docker exec -t k-package-docker-test-jammy-5dde91161befd1f004ba0af7b8527e0616ae799c bash -c 'mv /tmp/settings.xml ~/.m2/settings.xml'
+ docker exec k-package-docker-test-jammy-5dde91161befd1f004ba0af7b8527e0616ae799c /bin/bash -c 'mvn --batch-mode deploy'
/bin/bash: line 1: mvn: command not found
Error: Process completed with exit code 127.
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -118,6 +118,7 @@ jobs:
           docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'cd ~ && kompile test.k --backend llvm'
           docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'cd ~ && kompile test.k --backend haskell'
           docker image push ${DOCKERHUB_REPO}:${version_tag}
+          docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'apt-get install --yes maven'
           docker exec -t k-package-docker-test-jammy-${GITHUB_SHA} bash -c 'mkdir -p ~/.m2'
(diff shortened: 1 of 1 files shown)


tothtamas28 (approved):

ehildenb:
This change seems nice for uniformity, but I don't think it will fix the issue. I don't see `maven` listed in the Docker image we're using here for testing: https://github.com/runtimeverification/k/blob/5dde91161befd1f004ba0af7b8527e0616ae799c/package/docker/Dockerfile.ubuntu-jammy#L8, or in the runtime dependencies of K framework package: https://github.com/runtimeverification/k/blob/5dde91161befd1f004ba0af7b8527e0616ae799c/package/debian/control.jammy#L13

I don't think we really _want_ to put it in either of those places, because we want to test that the package works without having the build system present. So really, we should be doing the deployment of the java artifacts from a different docker container, probably from the docker container that is used for main testing: https://github.com/runtimeverification/k/blob/master/.github/workflows/Dockerfile

I would suggest:

- Pull out the `mvn deploy` into it's own job, instead of mixing it into the Dockerhub image build job.
- Have that use the `with-docker` action, which you can see examples of using that in other places in the repo, and that will use the linked Dockerfile at `.github/workflows/Dockerfile`, which does have `maven` installed.

ehildenb (approved):

PR #3368, opened 2023-04-26
Bring back maven deployment process to GitHub

ehildenb:
This PR does:

- Reverts #3369 (to bring back broken maven deployment).
- Separate the Maven deployment process into its own step.
- Run Maven deployment on public github runner, to sidestep some issues with Docker.


ehildenb:
The `!!! debugging` commit needs to be reverted prior to merge.

PR #3373, opened 2023-04-27
Remove simplification rules which lift to ML

ana-pantilie:


Baltoli:
Closing as there's limited context and Ana is no longer at RV; can be reopened if needed.

Issue #2625, opened 2022-05-26
Add a whitelist of attributes


radumereuta:
This would allow us to give better error messages in case of typos.
Add a group attribute that takes a comma separated list of names.
At first it behaves identically to a list of unstructured attributes `[group(foo,bar,baz)]` == `[foo,bar,baz]`. This happens only at parse time. We still need the internal attributes to be structured.
Add a warning for unstructured attributes.
Go downstream and fix all the warnings we are maintaining.
Then turn the warning into an error.

Steps:

- [x] Opt-in error for attributes not implemented.
- [x] Try with KEVM, document issues here.
- [x] Try with KAVM, document issues here.
- [x] Try with KCC, document issues here.
- [x] Fix issues that have come up.
- [x] Switch to opt-out behavior, get working downstream.
- [x] Update documentation (in `user-manual.md` and tutorials)
- [x] Remove option to opt-out.

ehildenb:
Related: https://github.com/runtimeverification/k/issues/843 and https://github.com/runtimeverification/k/issues/1585, #3193

radumereuta:
Is `group` the best word for this?
How about `tag`? I feel it's more appropriate and intuitive.
@ehildenb @dwightguth @Baltoli @Scott-Guest @Robertorosmaninho

Scott-Guest:
`group` seems appropriate if the only usage is for defining syntax groups. Are there other use cases for user defined attributes?

Robertorosmaninho:
Probably `group` is best @radumereuta, as we already use `tag` a lot on the LLVM Backend 😄

Scott-Guest:
I've created an initial whitelist consisting of everything in `Att.scala`, the K User Manual, and the minimum I needed to add to get a simple test passing:

| Attribute | Read | Written | Documented |
|-----------|------|---------|------------|
| alias | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| alias-rec | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| all-path | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| anywhere | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| applyPriority | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| assoc | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
[...]

PR #3376, opened 2023-04-28
Create attribute whitelist and group attribute

Scott-Guest:
First step of #2625

- Update `Att.scala` to include all documented attributes in the K User Manual
- Add a `group(name1,...,nameN)` attribute to specify syntax groups
- Add a `--pedantic-attributes` option to enforce that all attributes are either provided through `group` or defined in `Att.scala`


Scott-Guest:
Tests are still failing because the whitelist is not complete. For the small test I added, still need to address:

- boolOperation
- cellName
- defaultBracket
- equalEqualK
- index
- initializer
- mlAnd
- mlBinder
- mlEquals
- mlImplies
- mlOp
- mlQuantifier
- mlUnary
- noThread
- notEqualEqualK
- returnsUnit
- stream
- topcell
- unblock
- unparseAvoid

Are any of these meaningful except as syntax groups?

k-distribution/tests/regression-new/pedanticAttributes/Makefile
```diff
@@ -0,0 +1,3 @@
+KOMPILE_FLAGS=-w none --pedantic-attributes
+
+include ../../../include/kframework/ktest-fail.mak
```
radumereuta: new line at eof is nicer.

k-distribution/tests/regression-new/pedanticAttributes/pedanticAttributes.k.out
```diff
@@ -0,0 +1,6 @@
+[Error] Compiler: Unrecognized attributes: [badAtt]
```
radumereuta: You could specify in the error message that: user defined attributes can be added with `group(my, atts)`

kernel/src/main/javacc/Outer.jj
```diff
@@ -665,15 +666,29 @@ void Attributes(ASTNode node) : {}
...
+    { String key; String val = ""; StringBuilder sb = new StringBuilder(); }
+    <KEY> { key = image(); }
+    ( "(" { SwitchTo(TAG_STATE); }
+      ( val = String() ")"
+      | TagContent(sb) ")" { val = sb.toString() + special(); }
+      )
+      { SwitchTo(ATTR_STATE); }
+    )?
+    { node.addAttribute(key, val); }
+  )
```
radumereuta: Huh? Why did you modify the outer parser?
It should be able to just parse `group(my, att)`.
I wouldn't insert more logic into parsing attributes.
If you want to modify what the compiler passes see right after parsing, you can modify this pass: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java#L324

radumereuta (changes requested):
A few code comments.

> * boolOperation - probably group
> * cellName - probably internal
> * defaultBracket - probably internal
> * equalEqualK - probably group
> * index
> * initializer
> * mlAnd - probably group
> * mlBinder - probably group
> * mlEquals - probably group
> * mlImplies - probably group
> * mlOp - probably group
> * mlQuantifier - probably group
> * mlUnary - probably group
> * noThread
> * notEqualEqualK
> * returnsUnit - probably internal
> * stream - probably internal
> * topcell - probably internal
> * unblock
> * unparseAvoid - probably internal
You can find more about these tokens by searching the entire code base. If you find them somewhere in the code it's likely they have some meaning. If you can't find them anywhere then they may be used only for grouping. `unaprseAvoid` for example is used by the pretty printer. So you need to be tomorrow about your search.

kernel/src/main/javacc/Outer.jj
Scott-Guest: Not that big of a deal, but I wanted to avoid accepting quoted inputs because they're unnecessary here and their semantics feel a bit odd in this context.

For example, `group("my, foo")` would be parsed the same as `group(my, att)`, but things like `group("my", "att")` aren't accepted. I then also have to do manual error checking for cases like `group(my,,,, att)`.

On the second thought though, I guess we already deal with these same things for multi-argument attributes like `seqstrict`, so I'll pull the logic into `KILtoKORE` as you suggest.

radumereuta: In general, you want to allow your parser to accept more and give better, more informed error messages by analyzing the resulting AST.
If you restrict the language in the parser, you will get _unexpected token x_. And compilation stops.
If you analyze the AST you could say _tags are not allowed to be null_. And as an added bonus, you can return multiple error messages.

Scott-Guest:
Not sure what happened here, but the commit history is all out of whack and I seemed to have unintentionally merged the wrong branch. Going to clean this up on a new branch and open a fresh pull-request.

radumereuta:
@Scott-Guest I recommend using a Git GUI.
I use GitKraken and it saves me a lot of hassle for 99% of my git interactions.
I still have to use the CLI from time to time for more advanced stuff, but the day to day is so much easier.
You get to see the changes preview and select exactly the changes you want. You get to see stashes, branches, dependencies and updates.
It also makes it way easier to revert, cherry pick, and merge branches.

PR #3382, opened 2023-04-28
Add temp-dir option

### LINKED ISSUE #3237: K tools by default should not create hidden temporary directories in CWD
By default, directories like `.kprove-*` should either not be created, or if necessary created under `/tmp` and cleaned up once no longer needed.

As an option, the user should be able to tell the tool where to create these directories, e.g.

```
[--temp-directory DIR]
```

Keeping around the directories can then either be implicit if the flag is provided, or another flag can be introduced for this purpose.

radumereuta:
Fixes: #3237


radumereuta:
Adding support for the script tools (krun, kore-print, kparse, kparse-gen) is somewhat harder since they duplicate a lot of code.
Is that needed?
In fact, is this entire issue still needed?
I would like to avoid adding code that needs to be maintained if nobody is going to use it.

kernel/src/main/java/org/kframework/main/GlobalOptions.java
```diff
@@ -151,4 +151,7 @@ public boolean includesExceptionType(ExceptionType e) {
     public boolean debug() {
         return debug || debugWarnings;
     }
+
+    @Parameter(names={"--temp-dir"}, description="Put temp files in this location. Default is ./.<tool>-xxx")
```
tothtamas28: I think the default should be `/tmp/.<tool>-xxx`.

tothtamas28:
> In fact, is this entire issue still needed?

You mean whether we should close #3237 without addressing it? We can move the issue back to triage and discuss it during the K Dev meeting. But I personally think the default suggested in the issue is more user friendly than the current behavior, hence in my opinion its existence is justified.

> I would like to avoid adding code that needs to be maintained if nobody is going to use it.

The issue is about a default behavior. As such, all users are going to use it.

> Adding support for the script tools (krun, kore-print, kparse, kparse-gen) is somewhat harder since they duplicate a lot of code.
> Is that needed?

I'd argue it is needed for consistency. It does not have to be addressed in a single PR though. Also, if there's technical dept that makes changes like this difficult to carry out, then we can a open an issue to start addressing that as well.

k-distribution/src/main/scripts/lib/k-util.sh
```diff
@@ -16,6 +16,7 @@
 fold_lines='fold -s'
 profile=false
 verbose=false
+tempDirParent=/tmp
```
radumereuta: Is there a standard way to get the default temp dir location?

tothtamas28 (approved):

PR #3387, opened 2023-05-03
make field in KoreBackend protected instead of private

dwightguth:
This is a very simple PR that should be pretty easy to review, so I tagged several people in it. It doesn't need more than one of you to review it.

All this is doing is marking one field in KoreBackend as protected instead of private. This is necessary because the Maude backend needs to be able to access this field. In general other backends would want access to this field as well if they choose to implement any preprocessing transformations over the K definition prior to it being emitted to KORE. As such, it makes sense that this field be protected so that any backend that wants to can read the value. Since it is final and a basic type, the derived class can't actually modify the value at all, so it should be very safe.


Scott-Guest (approved):
Seems reasonable, LGTM

dwightguth:
This is blocking https://github.com/runtimeverification/maude-backend/pull/13

Robertorosmaninho (approved):

PR #3390, opened 2023-05-04
Add attribute whitelist and group(_) attribute

Scott-Guest:
First step of #2625

- Update `Att.scala` to include all attributes used in the frontend code or documented in the K User Manual
- Add a `group(name1,...,nameN)` attribute for user-defined groups
- Add a `--pedantic-attributes` flag to enforce that all attributes are either provided through `group(_)` or defined in `Att`

In the process, I also changed the type of attribute keys from `String` to a case class `Att.Key` with a `String` field. This is useful because
- keeping the `Att.Key` constructor private forces every new attribute to define its key in `Att`, thereby maintaining the whitelist
- IDE navigation can now point you to every usage of an attribute

The diff is quite large, but almost all changes just replace string literals `"my-att"` with `Att.MY_ATT()`.


Scott-Guest:
@radumereuta I know you mentioned expanding the `group(_)` attributes in `KILtoKORE`. I ended up having to place it a little earlier because `KILtoKORE` expects an already-constructed `Context` which will only be correctly built once `group(_)` is already expanded.

kernel/src/main/java/org/kframework/compile/ExpandGroupAttribute.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class ExpandGroupAttribute {
+
+    private static Att expandGroup(Att att, HasLocation node) {
+        if (!att.contains(Att.GROUP(), String.class)) {
+            return att;
+        }
+        String groups = att.get(Att.GROUP()).trim();
+        if (groups.isEmpty()) {
+            throw KEMException.compilerError("group(_) attribute expects a comma-separated list of arguments.");
```
radumereuta: This could use location information.

kernel/src/main/java/org/kframework/compile/ExpandGroupAttribute.java
```diff
@@ -0,0 +1,55 @@
...
+                throw KEMException.compilerError("User-defined group '" + group +
+                        "' conflicts with built-in attribute.");
+            }
+            if (!group.matches("[a-z][a-zA-Z0-9-]*")) {
+                throw KEMException.compilerError("Invalid argument '" + group + "' in group(_) attribute. " +
+                        "Expected a lower case letter followed by any number of alphanumeric or '-' characters.", node);
+            }
+            att = att.addGroup(group);
+        }
+        return att.remove(Att.GROUP());
```
radumereuta: This looks a bit overkill.
A simple split and trim should be enough (like [here](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1868))
No need to change it though.

kernel/src/main/java/org/kframework/compile/ExpandGroupAttribute.java
```diff
@@ -0,0 +1,55 @@
...
+        if (groups.startsWith(",") || groups.endsWith(",")) {
+            throw badCommaException;
+        }
+        for (String group : groups.split("\\s*,\\s*")) {
+            if (group.isEmpty()) {
+                throw badCommaException;
+            }
+            if (Att.getWhitelistedOptional(group).isPresent()) {
+                throw KEMException.compilerError("User-defined group '" + group +
+                        "' conflicts with built-in attribute.");
```
radumereuta: loc info.

radumereuta (approved):
Looks very good.
See my inline comments. I think you're missing a few location info for some error messages.
This is very large, and I would like it if someone else at least skims over it.

ehildenb (approved):

k-distribution/tests/regression-new/checkWarns/missingKlabel.k
```diff
@@ -7,7 +7,7 @@ module MISSINGKLABEL
 
   syntax MyMap ::= MyMap MyMap
       [ left, function, hook(MAP.concat), klabel(_MyMap_), symbol, assoc, comm
-      , unit(.MyMap), element(_M|->_), index(0), format(%1%n%2), unused
```
dwightguth: the element attribute is actually being used by the llvm backend and needs to be added to the whitelist.

k-distribution/tests/regression-new/group/test.k
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) K Team. All Rights Reserved.
+module TEST-SYNTAX
+  imports INT-SYNTAX
+
+  syntax Int ::=
+      "(" Int ")" [bracket]
+    | "-" Int     [function, unmin]
```
dwightguth: What exactly is the point behind this test? It would fail if the option you just added were passed.

kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -40,130 +62,245 @@ class Att private (val att: Map[(String, String), Any]) extends AttributesToStri
...
   private val intClassName = classOf[java.lang.Integer].getName
 
-  def from(thatAtt: java.util.Map[String, String]): Att =
+  // Marker for group(...) attributes
+  case class GroupMarker() extends AttValue
+  private val groupMarkerClassName = classOf[GroupMarker].getName
+
+  // All String fields with UPPER_CASE naming are considered part of the whitelist
+  private val whitelist: Set[Key] =
+    Att.getClass.getDeclaredFields
```
dwightguth, line 292: There is a significant problem here: I like that you've ensured that every single structured attribute used by the codebase exists in the list above, but not all of those attributes are designed to be added by the user. Some of them are definitely designed to be used as annotations during the compilation process, and would lead to incorrect behavior of the compiler if the user wrote them manually.

I believe we ought to be able to address this with either a subclass, or some kind of naming convention to represent the keys that users are expected to be allowed to write. I'll leave that to your discretion. I will comment above with a complete list of attributes that I definitely believe the user SHOULD BE allowed to write in their definition.

Baltoli (approved):

kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -40,130 +62,245 @@ class Att private (val att: Map[(String, String), Any]) extends AttributesToStri
...
-  val TOP_RULE = "topRule"
-  val TOTAL = "total"
-  val TRANSITION = "transition"
-  val TRUSTED = "trusted"
-  val UNBOUND_VARIABLES = "unboundVariables"
-  val UNIT = "unit"
-  val UNIQUE_ID = "UNIQUE_ID"
-  val USER_LIST = "userList"
+  // Note: Any field Att.UPPER_CASE_NAME of Att.Key type is automatically added to the
+  // whitelist of attribute keys (through reflection).
```
dwightguth: See my comment below. Here is the list referenced below of attributes that SHOULD be part of the whitelist:

* alias
* alias-rec
* all-path
* applyPriority
* assoc
* avoid
* bag
* binder
* bracket
* cell
* cellCollecction
* cellName
* color
* colors
* comm
* concrete
* constructor
* context
* cool
* element
* exit
* format
* freshGenerator
* function
* functional
* group
* heat
* hybrid
* hook
* idem
* impure
* initial
* initializer
* injective
* internal
* kast
* klabel
* kore
* label
* latex
* left
* lemma
* locations
* macro
* macro-rec
* maincell
* memo
* mlBinder
* mlOp
* multiplicity
* noThread
* non-assoc
* non-executable
* not-lr1
* one-path
* owise
* parser
* prec
* prefer
* priority
* private
* public
* result
* returnsUnit
* right
* seqstrict
* simplification
* smtlib
* smt-hook
* smt-lemma
* stream
* strict
* symbol
* symbolic
* tag
* token
* topcell
* total
* trusted
* type
* unblock
* unboundVariables
* unit
* unparseAvoid
* unused
* wrapElement

dwightguth: Of this list, the following attributes exist on the list because they're used in practice currently, but we should probably evaluate in a follow-up to determine it they still serve a purpose or if they can be removed entirely:

* bag
* noThread
* topcell
* unblock

The following attributes are definitely used in practice in the K prelude, but we may wish to restrict their usage in user written code:

* internal
* mlBinder
* mlOp
* stream
* unparseAvoid

dwightguth (dismissed):

k-distribution/tests/regression-new/group/test.k
Scott-Guest: This test just confirms that the `group(_)` attributes are appropriately handled and can actually be referenced by e.g. syntax priorities (the other tests error out so they don't confirm this).

Good point that this will fail with `--pedantic-attributes` though, which will eventually be required, so I'll clean it up

kore/src/main/scala/org/kframework/attributes/Att.scala
Scott-Guest: I agree, however, one thing @radumereuta pointed out is that we want to ensure we can still dump code after some execution then reload it, so we don't want to outright ban these internal attributes from appearing in the source.

Perhaps we could add an `internal(_)` attribute that is sort of analogous to how `group(_)` works?

On a similar note, even if we error check these internal attributes, it's also just a very leaky abstraction to have user and internal attributes in the same namespace. Every attribute we add internally to the compiler then becomes a reserved word in the language syntax and needs to be documented. 

In the frontend, we could easily fix all these issues by creating three different key classes:
- `Att.BuiltinKey`, for documented built-ins that can appear in user code
- `Att.InternalKey`, for compiler-internal attributes that can only appear in source code via `internal(my-internal-att)`
- `Att.UserGroup`, for user-defined groups that are defined via `group(my-group)`

We then give `Att.InternalKey` and `Att.UserGroup` separate namespaces, but still ensure that neither conflicts with any `Att.BuiltinKey` (enforced statically for `Att.InternalKey` and dynamically for `Att.UserGroup`).

However, when emitting to KORE all the attributes still need to be emitted as `\key{value}`, so this would take a bit more work to handle across the backends. Perhaps when emitting, we could prefix builtin and internal attribute keys with some char that is not permitted to appear in user-defined groups?

dwightguth: You raise some good points, but also some things I'm less sure of. For example, it's probably not going to be practical in general to kompile a definition, emit it as k after being compiled, and then recompile it and end up with the same compiled definition you got the first time. You can't do that now and I doubt you'll be able to in the future. So we don't need to worry about making sure that that can work.

Similarly, there is no reason to emit partially compiled definitions, mess with them manually, and then pass them back to the compiler to compile the rest of the way. This is a huge nightmare from an abstraction perspective because it causes the entire internal state of the compiler at every step to become part of the compiler's public API, which we definitely don't want. I would not try to make sure this is possible either, as it greatly complexifies the way tooling gets written.

As such, I don't really see the benefit of being able to parse internal attributes added by the compilation process. We do want to be able to parse attributes added by the /parsing/ process, since we probably want to be able to parse a definition, compile it, and then use the compiled and parsed definitions to modify the parsed definition and compile it again, but you will note that attributes which fall into this category I specifically added to the whitelist I wrote for this exact reason.

I agree that mixing user attributes and internal attributes is messy, but I would point out that since I am proposing we /not/ be allowed to parse internal attributes if they appear in rules, it would not do what you mention and interfere with backwards compatibility since the internal attributes would be entirely behind the abstraction boundary of the compiler. You would get the same issue, a parsing error, whether an attribute didn't exist or was internal, and adding a new attribute doesn't change the grammar of k used for parsing in any way.

As such, while I'm not opposed to maintaining a specific structured representation for internal attributes, I don't believe that it's that complicated or problematic to do what I originally mentioned.

kore/src/main/scala/org/kframework/attributes/Att.scala
Scott-Guest: Checking all usages of `Att.MY_ATT` and searching all the source files with
``` 
grep -rl 'my-att' . --include \*.hpp --include \*.h --include \*.cpp --include \*.java --include \*.scala --include \*.jj --include \*.jjt --include \*.hs --include \*.x --include \*.y --include \*.l --include \*.k --include \*.md
```
from what I can tell,
- `bag` is both read and written, and seems necessary on a quick glance
- `noThread` appears in some K source files, is added programmatically to productions in the frontend, but is never actually consumed anywhere
- `topcell` never appears in any K source file, is added programmatically, but is never actually consumed anywhere
- `unblock` is currently only used to identify the rule `stdinUnblock`, so we could just search for the `klabel` instead. (However, there is a TODO comment indicating that we should add more `unblock` rules in the future, which would require the attribute.)

k-distribution/tests/regression-new/checkWarns/missingKlabel.k
Scott-Guest: `element` is actually already in the whitelist. 

This test was failing because the error checking iterates over the attribute `HashMap` without sorting. My changes then ended up causing the missing `_M|->_`  label to be reported before the missing `.MyMap` label.

I forgot to actually fix this - I'll submit a separate PR quickly to stabilize those error messages.

kore/src/main/scala/org/kframework/attributes/Att.scala
dwightguth: Some context:

The bag attribute is automatically added to certain cell collections in a configuration declaration in order to inform the difference between a map and a bag, both of which are non-idempotent, associative, commutative collections. We can probably just use the hook attribute on the sort declaration instead.

noThread is added to certain sentences generated from a configuration declaration, so if we parsed a definition, emitted it back to k, and parsed it again, it would crash if we reject this attribute. But it was created to support a feature of the ocaml backend which has not existed for a considerable amount of time and can probably easily be removed entirely.

topcell is similar to nothread except that it was never used anywhere. Originally it was planned to have meaning when the compiler pipeline was rewritten entirely in the early days of K 4.0, but when it came time to implement that part of the compiler, it was never actually used. It should also be easy to delete.

unblock can probably be replaced with an instance of the label attribute. That todo is old enough and pertains to code that has not been actively used in production in many years, so it can just be ignored.

kore/src/main/scala/org/kframework/attributes/Att.scala
Scott-Guest: > As such, I don't really see the benefit of being able to parse internal attributes added by the compilation process. We do want to be able to parse attributes added by the /parsing/ process, since we probably want to be able to parse a definition, compile it, and then use the compiled and parsed definitions to modify the parsed definition and compile it again, but you will note that attributes which fall into this category I specifically added to the whitelist I wrote for this exact reason.

This makes sense to me. @radumereuta, does this cover what you were referring to?

A structured `internal(_)` attribute could still be a good way to handle the "we may wish to restrict their usage in user written code" attributes you listed in your other comment, but it's not essential then.

> I agree that mixing user attributes and internal attributes is messy, but I would point out that since I am proposing we /not/ be allowed to parse internal attributes if they appear in rules, it would not do what you mention and interfere with backwards compatibility since the internal attributes would be entirely behind the abstraction boundary of the compiler.

I agree that, once the whitelist is in effect, banning something like `[internal-att]` from appearing in the source code is completely fine. For all the user knows, `internal-att` is just not a known built-in, so there's no abstraction leak.

However, the issue is that we also would need to ban `[group(internal-att)]` because internally its represented the same as `[internal-att]`. This does actually break the abstraction because adding new internals could break backwards compatibility with existing user groups.

dwightguth: Oh, I see. Then we should probably, in a follow-up change, put groups and nongroups in separate attribute pools internally.

k-distribution/tests/regression-new/checkWarns/missingKlabel.k
Scott-Guest: #3395

kore/src/main/scala/org/kframework/attributes/Att.scala
radumereuta: I think Dwight has a point about not allowing internal attributes through the parser.
It should work though through other means, like JSON ingest.

Scott-Guest:
After my latest changes, a `KeyType` member was added to `Att.Key` which can be one of the following:
- `BuiltIn`, for user-facing attributes
- `Internal`, for compiler-internal attributes
- `UserGroup`, for user-defined groups provided via `group(_)`
- `Raw`, for yet-to-be-processed or unrecognized attributes

The pipeline now works as follows:
- The parser inserts a `BuiltIn` if the key is recognized, or else inserts a `Raw` key.
- Immediately after every call to the parser, a pass `ProcessGroupAttributes` is run which:
    - Replaces `[group(g1,...,gN)]` with `[g1,...,gN]`
    - If `--pedantic-attributes` is disabled, replaces every `Raw` key with a `UserGroup` (or errors if it conflicts with an `Internal`)
- In `KILtoKORE`, `ProcessGroupAttributes` is run again. Two passes are necessary because
    - the first pass ensures the `Context` passed to `KILtoKORE` is constructed correctly
    - the second pass ensures that rule attributes are expanded, which would be missed in the first pass
      as rules are still unparsed
- In `CheckAtt`, errors are emitted for any unhandled `Raw` keys
- In `GenerateSentencesFromConfigDecl`, the same process is applied to cell-attributes `<cell foo="bar">`

A few things worth mentioning:
- `transition` was made a built-in rather than an internal (differing from the list @dwightguth provided) because of its use here: https://github.com/runtimeverification/k/blob/master/k-distribution/pl-tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.md
- `index` was added to the whitelist. I can't for the life of me seem to find where this gets used, but removing it from `syntax Map ::= Map Map` in `domains.md` breaks the test `map-symbolic-tests-haskell/assignment-11-spec.k`
- `smt-prelude` was removed from a few tests (it appears to be dead-code anyways)
- `lookup` and `choice` attributes were removed (also dead code, and conflicted with user groups in the PL tutorials)

Additionally, we now report errors if user groups conflict with internal attributes, regardless of the `--pedantic-attributes` flag. In a follow-up submit, I'll address this and make it so that internal and group attributes are namespaced separately.

dwightguth:
I'm aware that transition is still existing in some places. I didn't put it on the whitelist because it is a dead attribute that no longer does anything and needs to eventually be deleted everywhere.

dwightguth:
I'd guess that the map test in question is probably just fragile. It's not a big deal to leave it there so that we don't have to try to fix the test now, but it's probably worth having someone investigate the test and try and see why changing one attribute of one production leads to a different test output

Scott-Guest:
Blocked until runtimeverification/k-exercises/pull/36 is merged

Baltoli (approved):
Dwight will need to hit the final approval as there are unresolved change requests but this looks good to me; I note that the blocking PR to k-exercises is merged as well.

ehildenb (approved):

PR #3391, opened 2023-05-04
Make new-format rules without antileft the default

Baltoli:
Related to: https://github.com/runtimeverification/k/issues/3035

This PR inverts the previous behaviour of ModuleToKore such that the new-format rewrite rules are generated by default, rather than being gated behind the `--disable-kore-antileft` flag.

Changes in this PR:
* Rename and invert uses of `--disable-kore-antileft` to produce `--enable-kore-antileft`
* Reorder one symbolic Map test on the Haskell backend that seems to apply two rules in a different order if the new format rules are used

This PR's CI run will demonstrate that the new changes allow the full regression test suite to run. I have also completed downstream testing with the following semantics via their Nix flakes, and can add more to this list if required:
* [x] C
* [x] EVM

If this change causes unexpected failures, we can address it in one of two ways depending on the failure:
* Add the `--enable-kore-antileft` flag to a semantics' build system, and fix gradually
* Revert this PR and revisit the entire change

### DIFF
--- a/k-distribution/tests/regression-new/issue-3035-antileft/haskell/Makefile
+++ b/k-distribution/tests/regression-new/issue-3035-antileft/haskell/Makefile
@@ -2,6 +2,6 @@ DEF=test
 EXT=test
 TESTDIR=.
 KOMPILE_BACKEND=haskell
-KOMPILE_FLAGS=--syntax-module TEST --disable-kore-antileft
+KOMPILE_FLAGS=--syntax-module TEST
 
 include ../../../../include/kframework/ktest.mak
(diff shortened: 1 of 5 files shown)


dwightguth (approved):
Given that we are going to need the antileft patterns in order to do proof object generation, which we are already starting to dig into, I would not say that the enable flag ought to be deprecated. We're going to want to keep it around. Code looks good otherwise though.

Baltoli:
From the original issue:

> That `\not{SortGeneratedTopCell{}}(priorityLE2{}())` encodes the priority of the rewrite rule into Kore, and it was deemed necessary a long time ago for constructing proof objects. This is no longer one of our goals [...]

Obviously happy to remove the deprecation and keep the old behaviour, I just wanted to make sure everyone is on the same page here. We can discuss in the K meeting.

PR #3397, opened 2023-05-05
Add back Maven packages changes

dwightguth:
This is a revert of the revert in #3369, which reverted all our changes to try to get maven github packages deployment working.

I brought it back in this PR and modified the workflow so that if the push to github packages fails, the workflow will ignore the failure and continue on.

This is a temporary change; I intend to make it so that failures fail the release once I get it working; however, this allows us to make further changes in the near future to try to experiment with fixing the package deployment without preventing releases from going through in the meantime.

I also made some changes that I hope will fix the reason why it was failing previously.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,6 @@ jobs:
           pkg-name: kframework_amd64_ubuntu_jammy.deb
           build-package: package/debian/build-package jammy
           test-package: package/debian/test-package
-
       - name: 'Upload Package to Release'
         env:
           GITHUB_TOKEN: ${{ secrets.JENKINS_GITHUB_PAT }}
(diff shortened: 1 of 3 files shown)


ehildenb:
Please resume from this one instead: https://github.com/runtimeverification/k/pull/3368

F-WRunTime:
I think we should possibly look to combine the two branches 'fix-deploy' with this one. 

- Focus on the docker environment setup (this was this area I left off with)
- Use a temporary 'test' maven deployment repository [here](https://github.com/runtimeverification/test-maven-deployment)
- Comment out all the other steps and run this workflow on each push to the 'maven' branch to test if the workflow is working as expected during a maven specific deploy stage. 
- Re-introduce the commented out sections once everything is working as expected.

F-WRunTime (approved):

PR #3400, opened 2023-05-09
Setting `--directory` as deprecated

### LINKED ISSUE #3276: [K-Improvement] Hide the confusing `--directory` option from various help message
## Motivation

The `--directory` is a little quirky for historic reasons, it needs you to specify a directory with exactly one subdirectory with a `-kompiled` suffix.
This is confusing to beginners, and even gets confused with `--definition` by regular users (e.g. me). 

We should hide this option from help messages and remove it from documentation, while perhaps printing a warning.

## Alternative

Another possibility is to completely drop the option.

## Documentation

Documentation such as the usermanual and tutorial would need to be updated to not use this option

## Testing

No tests would need to change.

Robertorosmaninho:
This PR sets the flag `--directory` as deprecated and prints a warning when it's used on the following tools:
- kprove
- krun
- kast
- kparse
- kompile

For `kompile`, the recommended flag to use instead is `--output-definition`.
For other tools, the recommend flag is `--definition`.

Fixes: #3276

### DIFF
--- a/k-distribution/include/kframework/ktest-fail.mak
+++ b/k-distribution/include/kframework/ktest-fail.mak
@@ -28,7 +28,7 @@ kompile: $(TESTS)
 dummy:
 
 %.k %.md: dummy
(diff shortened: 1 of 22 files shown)


nishantjr (approved):
Looks great. Thanks @Robertorosmaninho

PR #3401, opened 2023-05-09
Fixing Kore label munging confusion

### LINKED ISSUE #3023: Kore label munging confusion
According to the munging table in [StringUtil](https://github.com/runtimeverification/k/blob/master/kore/src/main/java/org/kframework/utils/StringUtil.java#L551), and also to [KoreToK](https://github.com/runtimeverification/k/blob/fc61802320bf64d4e0b1f2462e7e91ea2d3296bd/kore/src/main/scala/org/kframework/parser/kore/parser/KoreToK.scala#L37), the munging of `-` should be `'Hyph'`. But I take this module in file `test.k`:

```
module MY-MODULE
  imports BOOL

  syntax Pgm ::= "foo"
endmodule
```

And it produces this Kore with `kompile --backend haskell test.k --main-module MY-MODULE`, and I inspect the `definition.kore`, and see this symbol present:

```
symbol Lblfoo'Unds'MY-MODULE'Unds'Pgm{} ...
```

I guess instead it should be:

```
symbol Lblfoo'Unds'MY'Hyph'MODULE'Unds'Pgm{} ...
```

I discovered this trying to reproduce `kast_to_kore` and `kore_to_kast` here: https://github.com/runtimeverification/pyk/pull/73

Robertorosmaninho:
Fixes #3023

The K Frontend had mentions of `Hyph` representation for `-` as a way to translate it to KORE to be used in older backends. Although, as described in the[ KORE Syntax](https://github.com/runtimeverification/haskell-backend/blob/master/docs/kore-syntax.md#identifiers), the `-` is allowed in symbols and module identifiers. Therefore, we chose to delete the mention of `Hyph` and use only `-` in the source code.

This PR deletes the initial use of `Hyph` and its further translation.
A new PR will be needed to delete the `Hyph` translation on the [llvm-bakend](https://github.com/runtimeverification/llvm-backend/blob/ef72b033ca29c3efbd3ebedc27a7ac53cc082e09/lib/ast/AST.cpp#L43).


Baltoli:
Let me know when the tests are passing and the required backend changes have been made @Robertorosmaninho

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1888,7 +1888,6 @@ private boolean isListOfVarsAttribute(String name) {
     private static String[] asciiReadableEncodingKoreCalc() {
         String[] koreEncoder = Arrays.copyOf(StringUtil.asciiReadableEncodingDefault, StringUtil.asciiReadableEncodingDefault.length);
         koreEncoder[0x26] = "And-";
-        koreEncoder[0x2d] = "-";
```
dwightguth, line 1891: I'm not sure removing this line is correct.

Robertorosmaninho: From my understanding, this code is just adapting the `asciiReadableEncodingDefault`, and as we are already setting it as `-` [there](https://github.com/runtimeverification/k/blob/e91f273282c632cbd3b4f735ee8139251aa6ca25/kore/src/main/java/org/kframework/utils/StringUtil.java#L551), we don't have to change it in this function.

dwightguth: Thanks for the explanation

dwightguth (approved):

PR #3407, opened 2023-05-10
Introduces `--execute-to-branch` to `krun` when using LLVM Backend

Robertorosmaninho:
This PR modifies the `krun` tool to add the option `--execute-to-branch`, which is now supported by the LLVM Backend when the user wants to print the current state when the search execution finds a branch. 

When passing the option `--execute-to-branch`, it will execute the definition until it branches in either the LLVM or the Haskell Backend. In other words, this option is propagated to `llvm-krun` or `kore-exec` regarding the backend in which the definition was kompiled.

The PR that implements the feature on the LLVM Backend is https://github.com/runtimeverification/llvm-backend/pull/756.
The issue requesting this feature is https://github.com/runtimeverification/llvm-backend/pull/756#:~:text=close%20these%20issues.-,Support%20for%20search%20execution%20until%20first%20occurrence%20of%20branching,-Notifications


dwightguth (changes requested):
A couple comments. First, I'd love to be tagged for review in changes to the llvm backend and k frontend in the future. I might not always review them, and that's okay, but it helps bring them to my attention, which is good, because there's several things I'd like to point out here:

1. The usage message is supposed to be in alphabetical order, but the new flag has been placed in the wrong spot.
2. The Haskell backend also supports this flag and if we're adding this flag to the frontend, it should get propagated to both backends and it shouldn't be labeled as specific to the llvm backend.

k-distribution/src/main/scripts/bin/krun
```diff
@@ -121,6 +122,7 @@ $KRUN options:
                                      Haskell backend
   -h, --help               Display this help and exit
       --definition DIR     Exact path to the kompiled directory.
+      --execute-to-branch  Execute until the proof branches.
```
Baltoli: The execution here is not necessarily a proof; in the case of the LLVM backend it's a non-deterministic program

Baltoli (approved):
One minor comment; otherwise looks good. Needs to wait for https://github.com/runtimeverification/llvm-backend/pull/756 to be fully propagated before we can merge.

k-distribution/src/main/scripts/bin/krun
Robertorosmaninho: "execution path" would be a good replace for "proof" here?

Baltoli: > Execute until a non-deterministic state is reached, then exit.
Requires the `--enable-search` option to have been passed to `kompile` on the LLVM backend.

Robertorosmaninho: Thanks!

dwightguth (approved):

PR #3412, opened 2023-05-12
Restrict `Bytes` token syntax

### LINKED ISSUE #3396: Restrict `Bytes` token syntax
`Bytes` tokens, currently defined as:

https://github.com/runtimeverification/k/blob/badf9c880ed615474d02f2df1b539673cfaf0243/k-distribution/include/kframework/builtin/domains.md?plain=1#L1757

should be restricted to only allow the following lexical elements:

* Any printable ASCII character (i.e. `0x20` to `0x7E`), except `"` (`0x22`) and `\` (`0x5C`)
* `\"` - `0x22`
* `\\` - `0x5C`
* `\t` - `0x09`
* `\n` - `0x0A`
* `\f` - `0x0C`
* `\r` - `0x0D`
* `\xHH` where `H` stands for a hexadecimal digit (`[0-9a-fA-F]`) - `0xHH`

Scott-Guest:
Closes #3396.

Restricts the `Bytes` token syntax to

- No longer permit Unicode escapes
- Only allow printable ASCII characters and common character escapes


Scott-Guest:
I added some negative tests in `bad-bytes-literal`, but I'm not sure if they're worth keeping given that they lock down the nitty-gritty of how the parse error is reported.

tothtamas28 (approved):
LGTM. Also passes `Bytes` integration tests in `pyk`.

dwightguth (approved):

dwightguth (approved):

PR #3417, opened 2023-05-15
Create a hidden category for advanced options

### LINKED ISSUE #3399: kompile -h shows too many options
After looking at other tools that do similar things, `kompile -h` shows too many options, which makes it intimidating and hard to read.
Maybe we should have two tiers of options. We used to have `-X` for experimental.
Could we have a `man kompile` option in the system? This way, we could reduce the amount of text presented.

radumereuta:
Fixes: #3399
For the Java tools, we use JCommander, a cli options library.
I customized the [DefaultUsageFormatter](https://github.com/cbeust/jcommander/blob/master/src/main/java/com/beust/jcommander/DefaultUsageFormatter.java)
to show `hidden` options when `--help-hidden` is provided.

I marked as hidden the options I thought to be advanced for someone starting with K.
Bellow are the options you will see when you use `-h`. Use `-hh` to print the rest. You can use both to get the entire list.

| kompile -h | kompile -hh |
| -- | -- |
|	--backend	|	--allow-anywhere-haskell	|
|	--concrete-rules	|	--bison-file	|
|	--coverage	|	--bison-lists	|
|	--debug	|	--bison-parser-library	|
|	--debug-warnings	|	--bison-stack-max-depth	|
|	--emit-json	|	--cache-file	|
|	--enable-llvm-debug	|	--debug-type-inference	|
|	--enable-search	|	--directory, -d	|
|	--gen-bison-parser	|	--enable-kore-antileft	|
|	--gen-glr-bison-parser	|	--floats-as-po	|
|	--help, -h	|	--haskell-backend-command	|
|	--help-hidden, -hh	|	--heuristic	|
|	--md-selector	|	--hook-namespaces	|
|	--no-haskell-binary	|	--ignore-missing-smtlib-warning	|
|	--output-definition, -o	|	--iterated	|
|	--syntax-module	|	--iterated-threshold	|
|	--top-cell	|	--llvm-kompile-output	|
|	--verbose, -v	|	--llvm-kompile-type	|
|	--version	|	--maps-as-int-array	|
|	--warnings, -w	|	--no-exc-wrap	|
|	--warnings-to-errors, -w2e	|	--no-llvm-kompile	|
|	-E	|	--no-prelude	|
|	-I	|	--post-process	|
|	-O1	|	--profile-rule-parsing	|
|	-O2	|	--read-only-kompiled-directory	|
|	-O3	|	--shutdown-wait-time	|
|	-W	|	--smt	|
|	-Wno	|	--smt-prelude, --smt_prelude	|
|		|	--timeout	|
|		|	--z3-cnstr-timeout	|
|		|	--z3-impl-timeout	|
|		|	--z3-jni	|
|		|	--z3-tactic	|
|		|	-ccopt	|

This works for all the other Java tools. I wasn't that aggressive with those since they didn't have that many options to start with.


kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -112,36 +112,36 @@ public String syntaxModule(FileUtil files) {
...
+    @Parameter(names="--bison-parser-library", description="Generate a shared library rather than an executable for Bison parsers", hidden = true)
     public boolean genBisonParserLibrary;
 
     public static final String DEFAULT_TRANSITION = "transition";
 
     @Parameter(names="--top-cell", description="Choose the top configuration cell when more than one is provided. Does nothing if only one top cell exists.")
     public String topCell;
 
-    @Parameter(names="--debug-type-inference", description="Filename and source line of rule to debug type inference for. This is generally an option used only by maintaners.")
+    @Parameter(names="--debug-type-inference", description="Filename and source line of rule to debug type inference for. This is generally an option used only by maintaners.", hidden = true)
```
Baltoli:
```suggestion
    @Parameter(names="--debug-type-inference", description="Filename and source line of rule to debug type inference for. This is generally an option used only by maintainers.", hidden = true)
```

kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java
```diff
@@ -0,0 +1,397 @@
...
+import com.beust.jcommander.*;
+import com.beust.jcommander.internal.Lists;
+
+import java.util.*;
+import java.util.ResourceBundle;
+
+/**
+ * Customized `usage` for kompile based on DefaultUsageFormatter.
+ * The only changes are related to printing hidden options when
+ * --help-hidden is provided.
```
Baltoli: Is the rest of the code taken from an example implementation?

Baltoli (approved):
Seems good to me; one minor question

kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java
radumereuta: This is the default implementation for the JCommander library. 
I took the code from here: https://github.com/cbeust/jcommander/blob/master/src/main/java/com/beust/jcommander/DefaultUsageFormatter.java
and modified the parts related to printing hidden attributes.
One thing that I'm not sure is how to handle the license. The original file is Apache 2.0.

Baltoli: Seems fine to do this, but you'll need to add an attribution notice I think: https://snyk.io/learn/apache-license/#apache-license-permissions

radumereuta: Added a notice comment. Please have another look.

dwightguth:
I'm uncertain about where some of these options have been placed, but I can't argue that you've chosen the most commonly used ones. Idk. I'm worried we're going to get a lot more confused users who won't know about the flags we've hidden, some of which, while not as commonly used, are still pretty important in certain circumstances.

radumereuta:
> I'm uncertain about where some of these options have been placed, but I can't argue that you've chosen the most commonly used ones. Idk. I'm worried we're going to get a lot more confused users who won't know about the flags we've hidden, some of which, while not as commonly used, are still pretty important in certain circumstances.

If you think I should move some options from one side to the other, please let me know. I was just trying to simplify the experience for new users. There are so many options it's confusing and intimidating reading it.
I will not merge this until I get feedback from more people.

@Scott-Guest @Robertorosmaninho please have a look as well and provide feedback on which side the options should be.

Robertorosmaninho:
> @Robertorosmaninho please have a look as well and provide feedback on which side the options should be.

I would put the `--no-haskell-binary` on the left column as well, as it's very important for MacOS users.

Scott-Guest:
As @Robertorosmaninho mentioned, I would move `--no-haskell-binary` to the LHS, with the caveat that we also update its help message to indicate that it may be needed for MacOS users.

Beyond that, it seems reasonable to me. I would be fine moving `-O1`/`-O2`/`-O3` to the LHS as well. They're pretty commonplace flags for a compiler, and it's nice for the user to see that they're available right away in case they're getting frustrated with performance issues. I don't feel that strongly about this though.

dwightguth:
I second the ones mentioned, and also recommend moving --gen-glr-bison-parser

dwightguth:
Also, we may want to move -E. It's pretty useful for generating minimal examples and self contained k definitions.

radumereuta:
Guys, please re-review.
I've added back the options you suggested and modified the table from the first comment to reflect the current state.

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKompileOptions.java
```diff
@@ -10,6 +10,6 @@ public class HaskellKompileOptions {
     @Parameter(names="--haskell-backend-command", description="Command to run the Haskell backend execution engine.", hidden = true)
     public String haskellBackendCommand = "kore-exec";
 
-    @Parameter(names="--no-haskell-binary", description="Do not force the haskell backend to use the binary format. Use the textual format instead. This is a development option. Sometime the binary format can cause issues.", hidden = true)
+    @Parameter(names="--no-haskell-binary", description="Do not force the haskell backend to use the binary format. Use the textual format instead. This is a development option. Sometime the binary format can cause issues especially on MacOS.")
```
Scott-Guest: The current help message reads a bit wonky IMO, so if we're re-writing it anyways, I'd prefer something like

`Use the textual KORE format in the haskell backend instead of the binary KORE format. This is a development option, but may be necessary on MacOS due to known issues with the binary format.`

Scott-Guest (approved):
LGTM, but left a suggestion for the help message

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKompileOptions.java
Robertorosmaninho: Up on this.

Robertorosmaninho (approved):
LGTM, although I agree with @Scott-Guest comments!

kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java
```diff
@@ -0,0 +1,401 @@
...
+package org.kframework.kompile;
+
+import com.beust.jcommander.*;
+import com.beust.jcommander.internal.Lists;
+
+import java.util.*;
+import java.util.ResourceBundle;
+
+/**
+ * NOTICE:
```
Baltoli: I think you might still need a reference to the original code being Apache 2.0 licensed, and possibly the text of that license. Not 100% sure but worth double checking

Baltoli (approved):
Looks good to me with one minor comment; I think we've arrived at a pretty sensible set of options to hide

kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java
radumereuta: Updated. How about now?

Baltoli (approved):

Issue #3365, opened 2023-04-26
[K-Bug] KEVM conformance test hanging due to optimisations on llvm backend


anvacaru:
### What component is the issue in?

llvm-backend

### Which command

- [ ] kompile
- [ ] kast
- [X] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

K version:    v5.6.42 Build date:   Tue Apr 11 19:35:49 EEST 2023

### Operating System

Linux

### Steps to Reproduce

1. In evm-semantics, check out [this branch](https://github.com/runtimeverification/evm-semantics/pull/1743).
2. Build dependencies using `make kevm-pyk; make deps`
3. Build the semantics for llvm-backend using debugging symbols `make build-llvm KEVM_OPTS=--enable-llvm-debug`
4. Run the test ` make tests/ethereum-tests/BlockchainTests/GeneralStateTests/stTimeConsuming/CALLBlake2f_MaxRounds.json.run` and notice it is passing.
5. Rebuild the semantics without the debugging symbols `make build-llvm -B`
6. Run the same test and notice that is hanging and does not finish.

### Expected Results

I expect the debug symbols to not affect the execution of the test and to either pass or fail.

Baltoli:
The `--enable-llvm-debug`  flag doesn't just enable debug symbols, it changes the optimisation level used by clang. The first debugging step here should be to take a look at the LLVM IR that the code generator has produced with and without `--enable-llvm-debug` to see if we're introducing an infinite loop or similar

Scott-Guest:
I was able to reproduce the hang, but as far as I can tell the generated IR is the same in both cases:

```
#!/bin/bash
set -e
set -x

rm -f yes-debug.kore no-debug.kore yes-debug.ll no-debug.ll yes-debug-strip.ll no-debug-strip.ll
make clean
make kevm-pyk
make deps

make build-llvm KEVM_OPTS=--enable-llvm-debug
cd .build/usr/lib/kevm/llvm
cp definition.kore ../../../../../yes-debug.kore
llvm-kompile-codegen definition.kore dt/dt.yaml dt 1 > ../../../../../yes-debug.ll
cd ../../../../..

make build-llvm -B
cd .build/usr/lib/kevm/llvm
cp definition.kore ../../../../../no-debug.kore
llvm-kompile-codegen definition.kore dt/dt.yaml dt 0 > ../../../../../no-debug.ll
cd ../../../../..

[...]

PR #3422, opened 2023-05-17
Remove ResolveHeatCoolAttribute -O2/-O3 behavior

Scott-Guest:
When `-O2` or `-O3` is provided, `ResolveHeatCoolAttribute` does not include an `isKResult` side condition for cooling rules. This leads to hangs as noted in https://github.com/runtimeverification/k/issues/3365.

In this PR,
- We remove this behavior for `-O2` and `-O3`, and always insert the `isKResult` side condition.
- As a result, `ResolveHeatCoolAttribute` is only ever passed the modes `{HEAT_RESULT, COOL_RESULT_CONDITION}`, so we also remove the modes all together. (The third mode `COOL_RESULT_INJECTION` was already dead code anyways).

In a future submit, I will add a flag `--non-deterministic-cooling` to selectively re-enable this behavior for particular rule labels during search. However, on the branch I am testing this on (https://github.com/runtimeverification/k/tree/unrestricted-heat-cool) it seems to always cause hangs, so further investigation is still needed. I am submitting without the flag for now in order to unblock https://github.com/runtimeverification/k/issues/3365.


Baltoli:
@Scott-Guest is the CI failure here transient or an actual failing test? Let me know if it's the former and I'll review

Scott-Guest:
@Baltoli I believe it was just a transient issue with the Macs, CI is passing entirely now.

Baltoli (approved):

PR #3430, opened 2023-05-19
Improving kore-print script

### LINKED ISSUE #3429: [K-Bug] kore-print doesn't support `-v`/`--verbose` mode

Front-End

### Which command

- [ ] kompile
- [ ] kast
- [X] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

 v5.6.102-0-g28420c1f47

### Operating System

Linux

_No response_

### Steps to Reproduce

`kore-print --verbose`

### Expected Results

`kore-print --verbose` to print the intermediate steps.
- Also, the `kore-print` script needs some minor fixes, such as deleting unused declared vars and "double quote in variable uses".

Robertorosmaninho:
Fixes: #3429 
- This simple PR improves the `kore-print` script
- Implement the necessary infrastructure to make the `-v`/`--verbose` mode work on this tool to print intermediate steps

### DIFF
--- a/k-distribution/src/main/scripts/bin/kore-print
+++ b/k-distribution/src/main/scripts/bin/kore-print
@@ -15,15 +15,14 @@ outputFile=-
 outputMode=pretty
 
 # setup temp files
-now=`date +"%Y-%m-%d-%H-%M-%S"`
-tempDir="$(mktemp -d .kore-print-${now}-XXXXXXXXXX)"
+now=$(date +"%Y-%m-%d-%H-%M-%S")
+tempDir="$(mktemp -d .kore-print-"${now}"-XXXXXXXXXX)"
 tempFiles=("$tempDir")
 trap 'rm -rf ${tempFiles[*]}' INT TERM EXIT
 
 KORE_PRINT=$(basename "$0")
 KAST_UTIL="$(dirname "$0")/kast"
 
-ktool="${KORE_PRINT}"
 source "$(dirname "$0")/../lib/kframework/k-util.sh"
 
 print_usage () {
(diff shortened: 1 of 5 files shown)


k-distribution/src/main/scripts/bin/kore-print
```diff
@@ -139,6 +140,12 @@ do
       trap - INT TERM EXIT
       ;;
 
+      -v| --verbose)
+      verboseFlag=" --verbose"
+      verbose=1
+      shift
+      ;;
```
ehildenb:
```suggestion
      -v| --verbose)
      verboseFlag="--verbose"
      verbose=1
      shift
      ;;
```

Is the extra space here on purpose? Maybe it's needed and I don't understand.

k-distribution/src/main/scripts/bin/kore-print
```diff
@@ -11,19 +11,20 @@ dir=.
...
+tempDir="$(mktemp -d .kore-print-"${now}"-XXXXXXXXXX)"
 tempFiles=("$tempDir")
 trap 'rm -rf ${tempFiles[*]}' INT TERM EXIT
 
 KORE_PRINT=$(basename "$0")
 KAST_UTIL="$(dirname "$0")/kast"
 
-ktool="${KORE_PRINT}"
+export ktool="${KORE_PRINT}"
 source "$(dirname "$0")/../lib/kframework/k-util.sh"
```
ehildenb: Why are we exporting `ktool`?

If we actually need to do this, then maybe we should use `KTOOL` instead? But it seems like we shouldn't need to do this to me....

Robertorosmaninho: we must export `ktool` to be used on the `k-util.sh` script where it is needed only [here](https://github.com/runtimeverification/k/blob/28420c1f477d8599de61618e47cd870688617255/k-distribution/src/main/scripts/lib/k-util.sh#L26)

k-distribution/src/main/scripts/bin/kore-print
Robertorosmaninho: Just a typo. Thanks!

Baltoli (approved):

PR #3431, opened 2023-05-22
Update projects description

radumereuta:
to encourage people to share their work on our projects page.
I've also replaced the Stackoverflow link with Telegram.
I'm not sure if anyone is monitoring Stackoverflow. We don't have the Slack notifications anymore since they were broken at some point.

### DIFF
--- a/web/pages/index.md
+++ b/web/pages/index.md
@@ -37,7 +37,7 @@ call/cc.
 
 - [Discord Server](https://discord.com/invite/CurfmXNtbN): Most direct way to get support.
 - [Matrix Room](https://matrix.to/#/#k:matrix.org): Another way to get support.
-- [Stackoverflow](https://stackoverflow.com/questions/tagged/kframework): for general questions to the K user community.
+- [Telegram](https://t.me/rv_inc): for newsletters and general announcements.
 
 ## Resources
 
--- a/web/pages/projects.md
+++ b/web/pages/projects.md
@@ -4,6 +4,9 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
 
 # Projects using K
 
+A list of projects using the K framework. If you are working on something interesting, and you want to share it with the community,
+let us know on our [socials](https://kframework.org/#support), and we will feature you on this list.
+
 <br>
 
 ## Featured


web/pages/index.md
```diff
@@ -37,7 +37,7 @@ call/cc.
 
 - [Discord Server](https://discord.com/invite/CurfmXNtbN): Most direct way to get support.
 - [Matrix Room](https://matrix.to/#/#k:matrix.org): Another way to get support.
-- [Stackoverflow](https://stackoverflow.com/questions/tagged/kframework): for general questions to the K user community.
+- [Telegram](https://t.me/rv_inc): for newsletters and general announcements.
```
ehildenb, line 40: I think we should keep StackOverflow, and just add Telegram to thi list perhaps?

Or are we not monitoring stack overflow anymore?

radumereuta: The StackOverflow plugin failed hard the last couple of times. It sent ~100 messages completely unrelated to K, and we had to disable it.
I feel somewhat indifferent about this.
Maybe @nishantjr has a stronger opinion. From what I remember, he is very active there.

Baltoli: FWIW we haven't had a stack overflow question in months; I'm not opposed in principle to keeping the link but we should make sure that we're monitoring it appropriately if we do.

Baltoli (approved):

web/pages/index.md
radumereuta: The last question we had is from Mar 27, 2022. More than a year ago.

Baltoli: huh you're right; I read March and parsed it as this year!

radumereuta:
Discussed it in the Wednesday meeting, and we agreed to remove StackOverflow from the list.
It hasn't had any traffic in more than a year and the notifications are down on Slack.

web/pages/index.md
nishantjr: I had liked the idea of Stackoverflow because it would let us build up a knowledge base. 
But we've not got as much traction as I had hoped. I open up stackoverflow fairly often and am following the tag, so don't mind answering questions there or forwarding them manually to the K channel. I'm also fine with dropping the link.

PR #3432, opened 2023-05-23
Implementing support for pretty printing non-terminals name

Robertorosmaninho:
Fixes https://github.com/runtimeverification/llvm-backend/issues/766

Given the syntax production: `syntax Foo ::= foo(x: String, y:String)` and the input `foo("a", "b")`, this PR will allow the output to be:
```
<k>
  foo(... x: "a", y: "b") ~> .
</k>
```
If, and only if, all the non-terminals have names. If not, as in the case of `syntax Bar ::= bar(x: String, String)` and the same input, we'll continue having this output:
```
<k>
  bar("a","b") ~> .
</k>
```


Robertorosmaninho:
With this new modification, the output of the test [issue-1098](https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/issue-1098/test.k) must be updated from:
```
x = ( 0 ~> . ) ; .RecordDescr ~> .
```
to
```
key: x = value: ( 0 ~> . ) ; .RecordDescr ~> .
```

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1535,15 +1535,35 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
...
         int nt = 1;
         boolean hasFormat = true;
+        boolean printName = true;
+        boolean printEllipses = false;
+        for (int i = 0; i < prod.items().size(); i++) {
+          if (prod.items().apply(i) instanceof NonTerminal && ((NonTerminal) prod.items().apply(i)).name().isEmpty()) {
+            printName = false;
+            break;
+          }
+        }
```
radumereuta:
```suggestion
        boolean printName = stream(prod.items()).noneMatch(pi -> pi instanceof NonTerminal && ((NonTerminal) pi).name().isEmpty());
        boolean printEllipses = false;
```
This can be simplified greatly if you use streams.

radumereuta (changes requested):
This allows too many things to be pretty printed this way.
For example, `syntax Exp ::= x:Int ")"` is considered by these changes, and they shouldn't.

If you look at the parsing code for record productions, it is guarded by [`isPrefixProduction`](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/definition/outer.scala#L608)
You should try and handle only the same cases, otherwise, the pretty printed version is not going to be re-parsable.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -1536,15 +1579,29 @@ private Att addKoreAttributes(Production prod, SetMultimap<KLabel, Rule> functio
...
         int nt = 1;
         boolean hasFormat = true;
+        boolean printName = stream(prod.items()).noneMatch(pi -> pi instanceof NonTerminal && ((NonTerminal) pi).name().isEmpty());
+        boolean printEllipses = false;
+
         for (int i = 0; i < prod.items().size(); i++) {
           if (prod.items().apply(i) instanceof NonTerminal) {
-            format = format.replaceAll("%" + (i+1) + "(?![0-9])", "%" + (nt++));
+            String replacement;
+            if (printName && isPrefixProduction(prod.items())) {
```
radumereuta:
```suggestion
            if (printName && prod.isPrefixProduction()) {
```
The original code is accessible from here. No need to duplicate it.

radumereuta (changes requested):

radumereuta (approved):
I've tested it and it works for both
```
f(x:Int)
(y:Int)
```
But does not work for `z:Int ")"` which is the behavior we want.

PR #3438, opened 2023-05-25
Throw error message when finding duplicate attributes

radumereuta:
Do a check early on at outer parsing since we lose the information after that.
For rules, the location information is only approximate since we parse attributes in two stages and fiddle with the indexes.
Line numbers should indicate +1 at most.

I'm expecting this to break some semantics. I already found some issues with our regression tests.

### DIFF
--- a/k-distribution/tests/regression-new/checks/duplicateAttRule.k
+++ b/k-distribution/tests/regression-new/checks/duplicateAttRule.k
@@ -1,0 +1,7 @@
+// Copyright (c) K Team. All Rights Reserved.
+module DUPLICATEATTRULE
+
+  syntax Foo  ::= a()
+  rule a() => .K [concrete, concrete]
+
+endmodule
--- a/k-distribution/tests/regression-new/checks/duplicateAttRule.k.out
+++ b/k-distribution/tests/regression-new/checks/duplicateAttRule.k.out
@@ -1,0 +1,5 @@
+[Error] Outer Parser: Duplicate attribute: concrete
+	Source(duplicateAttRule.k)
+	Location(5,29,5,37)
+	5 |	  rule a() => .K [concrete, concrete]
+	  .	                            ^~~~~~~~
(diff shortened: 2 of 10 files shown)


Scott-Guest (commented):
LGTM overall, but a few questions:

- For the inaccurate line numbers, could you elaborate more on what exactly is causing this, and why it would be difficult to solve?
  - In my opinion, this inaccuracy is pretty ugly and seems worth addressing (possibly in a follow-up submit) rather than just leaving it be, but I'll defer to your judgement.
  - On a related note, we might want to consider adding location information to attributes themselves. I know we are trying to tighten up the attribute language and will likely be adding more error checks like this in the future.

- Just for my own knowledge, when we introduce possibly breaking changes like this, is the best practice just to submit and require that each semantics makes the necessary changes when they update their K dependency? Or should we try to fix up the semantics before this is merged?

Baltoli:
@Scott-Guest:

> Just for my own knowledge, when we introduce possibly breaking changes like this, is the best practice just to submit and require that each semantics makes the necessary changes when they update their K dependency? Or should we try to fix up the semantics before this is merged?

There's no formal process for this yet; we've been kicking around the idea of having a downstream semantics testing workflow for a while but have not yet gone through the necessary steps to implement it on the devops side.

In general, if a PR is going to introduce potentially breaking changes, I would expect the author to acknowledge the semantics that they've tested it with, and reviewers should be looking for the potential for breakage (as you are here). The workflow for fixing up the semantics depends on the scale of the changes; here it's pretty small and the fix is obvious, so it's not the end of the world if we have to fix a semantics to get a K release through. If it's less obvious (e.g. the priority predicate change I made the other week, or your attribute grouping work) then we can do a staged migration over a series of commits.

Baltoli:
For what it's worth, the C semantics builds fine with this change applied.

radumereuta:
I've tested some of the evm kompile steps and they are passing. I haven't run all the proofs though. There might be more issues hiding there.

@Scott-Guest I think I finally managed to get the location info correct for rules as well. Have a look.

Robertorosmaninho (approved):

Scott-Guest:
@radumereuta Missed this before it was merged, but the location information logic looks good to me!

PR #3441, opened 2023-05-30
Emit user-defined groups as group(_)

### LINKED ISSUE #3439: Prevent internal attributes from leaking
When we emit to KORE, user-defined groups `[group(my-att)]` end up being represented the same as `[my-att]`. Accordingly, we must report an error if a user-defined group conflicts with any built-in attribute. Unfortunately, this includes any undocumented internal-only attributes.

To address this, we can instead emit the user groups literally as `\group{}("my-att")`, but this will require updating any consumers downstream.

Scott-Guest:
Fixes #3439.

Whenever emitting attributes (to JSON, KORE, etc.), emit user-defined groups as `[group(my-att-1, my-att-2)]` even though the front-end internally represents them as `[my-att-1, my-att-2]`. 

This prevents user-defined groups from conflicting with internal attributes.


Scott-Guest:
Blocked on #2625

If there is some built-in attribute `my-att` unintentionally missing from the whitelist, then this change will cause it to be emitted to KORE as `\group{}("my-att")`, breaking anything downstream depending on `my-att`.

To be safe then, we should wait until the whitelist is complete and the option to opt-out is removed before merging this.

radumereuta (approved):

PR #3451, opened 2023-06-05
Update K frontend to download packages from github instead of s3

dwightguth:
This should be the final PR needed to fully migrate the K frontend away from our maven s3 bucket. We do 3 things:

* Make it so the release fails if github package deployment fails; we didn't do this previously only because we were testing the workflow.
* Change the path we look for packages to download from looking at s3 to looking at github packages.
* Clean up the POM and remove unnecessary stuff. This includes both removing references to s3 and also removing the unused rv-maven-public-release repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -412,7 +412,6 @@ jobs:
 
       - name: 'Push Maven Packages'
         shell: bash {0}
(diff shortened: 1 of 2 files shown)


dwightguth:
Closing. We decided to go a different path than GitHub packages.

PR #3454, opened 2023-06-06
Improve `--version` implementation on K script

### LINKED ISSUE #3453: `kompile --version` is slow
Due to slow JVM startup, `kompile --version` takes around 0.3s. This is not ideal for tools like `kbuild` that rely on K version information but are expected to be fast.

Consider introducing a new tool, `kversion` or `kver`, that emits the same output as `kompile --version`, but without starting the JVM. This way, the time it takes to obtain the K version can be reduced to the 0.001s scale.

Robertorosmaninho:
Fixes #3453


k-distribution/src/main/scripts/lib/k
```diff
@@ -1,4 +1,36 @@
...
+K_BASE="$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)"
+SNAPSHOT_JAR="$(find "${KERNEL_DIR}"/target/ -name '*-SNAPSHOT.jar')"
+
+if [[ -z "${SNAPSHOT_JAR}" ]]; then
+    echo "Could not find the kernel jar file. Please run 'mvn package' in the K source directory." 1>&2
+    exit 1
+fi
+
+for flag in "$@"; do
+    if [[ "$flag" == "--version" || "$flag" == "-v" ]]; then
```
radumereuta: `-v` means verbose. This should only test for `--version`

Robertorosmaninho: my bad! thanks, @radumereuta.

k-distribution/src/main/scripts/lib/k
```diff
@@ -1,4 +1,36 @@
...
+        version=$(unzip -p "${SNAPSHOT_JAR}" META-INF/MANIFEST.MF | grep 'Implementation-Git-Describe' | awk -v RS='\r' '{print $2}')
+        versionTimestamp=$(unzip -q -c "${SNAPSHOT_JAR}"| grep "Implementation-Date:" | awk -v RS='\r' '{print $2}')
+        versionTimestampInMilliseconds=$(echo "${versionTimestamp} / 1000" | bc)
```
radumereuta, line 13: Try to get the info directly from `get-version-string.sh`
This is the script begin called by mvn when packaging the K tools.

radumereuta (approved):
Tested and works as expected.
lgtm

One small nit pick is that the date is printed in my locale instead of the usual en_US format.
```console
$ time kserver --version
K version:    v5.6.133-6-g5d78759a56
Build date:   Tue Jun 20 16:26:27 EEST 2023

real	0m0,280s
user	0m0,587s
sys	0m0,062s
$ time kompile --version
K version:    v5.6.133-6-g5d78759a56
Build date:   Ma iun 20 16:26:27 EEST 2023

real	0m0,040s
user	0m0,033s
sys	0m0,012s
```
You can merge when tests pass.

PR #3458, opened 2023-06-09
Fix EOF parsing error message crash

### LINKED ISSUE #3446: [K-Bug] "Internal: Uncaught exception thrown of type NoSuchElementException" while parsing

Front-End

### Which command

- [ ] kompile
- [X] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v5.6.90

### Operating System

Linux

`lambda.k`:

```
module LAMBDA
  import DOMAINS
  
  syntax Lambda ::=
    Id
    | "(" Lambda ")" [bracket]
    > left: Lambda Lambda
    > "/\\\\" Id "." Lambda

endmodule
```

### Steps to Reproduce

```
$ kompile --debug lambda.k
[Warning] Compiler: Could not find main syntax module with name LAMBDA-SYNTAX
in definition.  Use --syntax-module to specify one. Using LAMBDA as default.
$ cat bad.lambda # note the missing ending parenthesis
(/\\ x . const)
  ((/\\ x . x x) (/\\ x . x x)
$ kast bad.lambda
file a bug report at https://github.com/runtimeverification/k/issues
(NoSuchElementException: No value present)
```

### Expected Results

No internal error.

radumereuta:
Fixes: #3446
Easy fix. Failed to account for the case when the end location didn't exist because of EOF.
Before: crash
After:
```
[Error] Inner Parser: Parse error: unexpected end of file following token '('.
	Source(/home/radu/work/test/./2.test)
	Location(1,2,1,2)
	1 |	(
	  .	 ^
```


Baltoli (changes requested):
Can we add the example from the original issue as a regression test? Otherwise the change looks sensible to me!

Baltoli (approved):

Scott-Guest (approved):
LGTM!

Robertorosmaninho (approved):
LGTM!

Issue #3450, opened 2023-06-05
`kprove` with fresh variables


Baltoli:
A user on Discord reports the following:

`test.k`:
```k
module TEST-SYNTAX
    syntax Pgm ::= "quux"
endmodule

module TEST
    imports TEST-SYNTAX
    imports INT

    configuration <k> $PGM:Pgm </k>
        <c1> . </c1>
        <c2> . </c2>

    rule <k> quux => . </k>
        <c1> . => !C:Int </c1>
        <c2> . => !C:Int </c2>
endmodule
```

`test-spec.k`
```k
requires "test.k"

module TEST-SPEC-SYNTAX
    imports TEST-SYNTAX

endmodule

module VERIFICATION
    imports K-EQUAL
    imports TEST-SPEC-SYNTAX
    imports TEST

endmodule

module TEST-SPEC
    imports VERIFICATION

    claim <k> quux => . </k>
        <c1> . => ?C </c1>
        <c2> . => ?C </c2>
endmodule
```

`kprove` output:
```
kore-exec: [384063] Warning (WarnStuckClaimState):
    The configuration's term unifies with the destination's term, but the implication check between the conditions has failed. Location: /mnt/c/Users/brown/Code/ktest/test-spec.k:18:11-20:27
Context:
    (InfoReachability) while checking the implication
  #Not ( {
    _DotVar1
  #Equals
    _DotVar1 +Int 1
  } )
#And
  <generatedTop>
    <k>
      .
    </k>
    <c1>
      _DotVar1:Int ~> .
    </c1>
    <c2>
      _DotVar1:Int ~> .
    </c2>
  </generatedTop>
[Error] Prover: backend terminated because the configuration cannot be
rewritten further. See output for more details
```

Should this proof go through? Can you write proofs over fresh variables like this?

ehildenb:
It seems that what's happening is that for each occurrence of the fresh integer variable `!C`, we are incrementing the `<generatedCounter>` cell (which has contents `_DotVar1` at the beginning of the proof.

So perhaps, in the code which handles fresh counters, we should inspect if the variables used have the same name, and if so, give it the same incremented value (rather than incrementing twice).

Currently, the rule:

```
    rule <k> quux => . </k>
        <c1> . => !C:Int </c1>
        <c2> . => !C:Int </c2>
```

Is probably getting compiled to:

```
    rule <k> quux => . </k>
        <c1> . => _DotVar1 </c1>
[...]

PR #3459, opened 2023-06-09
Add `generatedCounter` implicitly to claims

Baltoli:
This PR adds an additional pass to the spec-module compilation pipeline; this pass adds a claimed `_ => ?_` rewrite to the `generatedCounter` cell if the claim does not already mention that cell. Doing so fixes the issue identified in #3450, where the prover fails to assert that this cell _does not change_ in a claim containing fresh variables.

The code triggering the original bug has been added as a test case, as well as a derived test that explicitly states the effect of the rule on the counter cell.

There is one place that this PR requires a test change; in the PL tests there is a spec that relies on a circularity implicitly containing `<generatedCounter> _ </generatedCounter>` rather than `_ => ?_`. This can be addressed by either removing the explicit "counter goes up by precisely 2" part of the spec, or by asserting that the counter doesn't change in the circularity. This PR chooses the latter, but either will work.

Fixes https://github.com/runtimeverification/k/issues/3450


kernel/src/main/java/org/kframework/compile/AddImplicitComputationCell.java
```diff
@@ -51,10 +50,10 @@ public Sentence apply(Module m, Sentence s) {
 
         if (s instanceof RuleOrClaim) {
             RuleOrClaim rule = (RuleOrClaim) s;
-            return rule.newInstance(apply(rule.body(), m), rule.requires(), rule.ensures(), rule.att());
+            return rule.newInstance(apply(rule.body(), m, rule instanceof Claim), rule.requires(), rule.ensures(), rule.att());
```
radumereuta, line 53: Nitpicking here.
If we are to differentiate between Rule and Claim, why not do it at the top level?
Have two different `apply` functions instead.

kernel/src/main/java/org/kframework/compile/AddImplicitCounterCell.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private K apply(K term, Module m) {
+        List<K> items = IncompleteCellUtils.flattenCells(term);
+        if(alreadyHasGeneratedCounter(items)) {
+            return term;
+        }
+
+        KRewrite rew = KRewrite(
+                KApply(KLabel("#SemanticCastToInt"), KVariable("_")),
+                KApply(KLabel("#SemanticCastToInt"), KVariable("?_")));
```
radumereuta: We already have constants for these variables in `ResolveAnonVar.FRESH_ANON_VAR`.

kernel/src/main/java/org/kframework/compile/AddImplicitComputationCell.java
Baltoli: I think we'd end up duplicating more code in this call stack if we did the differentiation at the top level; not opposed in principle to doing so but the current phrasing was the cleanest I came up with.

The issue is really that the initial sentence is no longer fully needed when we `apply` as we're on;y looking at its body, so you'd need an `applyRule` and `applyClaim` method that only differ in how they call `shouldConsider`.

kernel/src/main/java/org/kframework/compile/IncompleteCellUtils.java
```diff
@@ -65,7 +65,7 @@ public static List<K> getChildren(KApply cell) {
     private static KApply makeDots(boolean isOpen) {
         return isOpen ? dots : noDots;
     }
-    private static K makeBody(List<K> children) {
+    public static K makeBody(List<K> children) {
```
dwightguth: I would make this package-private.

dwightguth (approved):
Looks good but one minor comment.

radumereuta (approved):

PR #3461, opened 2023-06-12
Rangemap

mariaKt:
This PR adds support for a new builtin type, range map, in K. Previous PRs have introduced the new type ([here](https://github.com/runtimeverification/llvm-backend/pull/620)) and added hooks in the LLVM backend ([here](https://github.com/runtimeverification/llvm-backend/pull/656)).

The bulk of the changes are in `domains.md`, defining the relevant module.


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -465,6 +465,230 @@ module MAP-SYMBOLIC
...
 
+Range maps
+----
+
+Provided here is the syntax of an implementation of immutable, associative,
+commutative range maps from `Int` to `KItem`. This type is hooked to an
+implementation of range maps provided by the backend.
+Although the underlying range map data structure supports any key sort, the
+current implementation by the backend only supports `Int` keys due to
+limitations of the underlying ordering function.
```
radumereuta, line 477: Is this supported by the haskell backend as well? If not, then I think it should be mentioned here that only the llvm backend implements this feature.

I think we also have some module attributes to protect against wrongful inclusion.

ehildenb:
Hmmmm, what's the use case here? It seems to me that it's probably mostly useful for modelling byte-arrays, but I think the `#buf` abstraction in KEVM does that better.

Or perhaps the idea here is a general notion of `Array` datatype? In that case, I think we should probably make an abstract Array instead of the AC rangemap.

How is this datastructure being used?

k-distribution/include/kframework/builtin/domains.md
dwightguth: In a follow-up PR, we can probably make an inefficient version of the RangeMap sort written entirely in K for use by the Haskell backend. I don't see the need to adjust the module attributes currently, although I agree about the documentation.

dwightguth (approved):

dwightguth:
The use case is fairly straightforward. You want to be able to map ranges of integers to values and then look up the value associated with a particular integer. Neither an array nor #buf does this in a fashion that would allow you to perform both operations in log time and linear space in the number of ranges. We're talking about cases where there might be hundreds of thousands of contiguous integers that all map to the exact same value spread out across an incredibly sparse larger range of integers... Neither an array nor any existing data type in K can do this without using vastly more memory than is practical.

mariaKt:
@ehildenb Specifically, we are dealing with this when using the C semantics with programs containing large static arrays. Currently, we are using the traditional K map to represent the memory cell, which is proving to be inefficient with such programs.

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -465,6 +465,230 @@ module MAP-SYMBOLIC
 endmodule
 ```
 
+Range maps
```
```suggestion
Range Maps
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -465,6 +465,230 @@ module MAP-SYMBOLIC
...
+```k
+  syntax RangeMap ::= RangeMap RangeMap                        [left, function, hook(RANGEMAP.concat), klabel(_RangeMap_), symbol, assoc, comm, unit(.RangeMap), element(_r|->_), index(0), format(%1%n%2)]
+```
+
+### Range map unit
+
+The `RangeMap` with zero elements is represented by `.RangeMap`.
+
+```k
+  syntax RangeMap ::= ".RangeMap"                         [function, total, hook(RANGEMAP.unit), klabel(.RangeMap), symbol, latex(\dotCt{RangeMap})]
```
Baltoli, line 516: Do we actually support LaTeX generation any more?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -465,6 +465,230 @@ module MAP-SYMBOLIC
...
+that two different range maps will return the same element, even if they are
+similar.
+
+```k
+  syntax KItem ::= choice(RangeMap)                      [function, hook(RANGEMAP.choice), klabel(RangeMap:choice)]
+```
+
+### Implementation of range maps
+
+TBD
```
Baltoli: If we're going to merge this into `domains.md`, we should have a more descriptive comment here explaining why the implementation is TBD.

Baltoli (approved):
A couple of small comments but this looks really good @mariaKt - I have tried to review tests from each group, and what's there is great, but as there are ~100 tests it's possible I missed something.

k-distribution/include/kframework/builtin/domains.md
Baltoli: Leave in place; address wholesale for the other structures if removal is needed

PR #3464, opened 2023-06-14
Add booster to the k derivation

goodlyrottenapple:
Fixes: #3465 

This PR adds the booster and updates haskell-backend after some recent ghc upgrades broke the docker build in K. The old style nix CI run is also brittle and unused so I am removing it in this PR (fixes https://github.com/runtimeverification/k/issues/3465) as the same functionality tested in the Nix job also appears in the Nix flake job.


.github/workflows/Dockerfile.stack-deps
```diff
@@ -8,7 +8,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 RUN    apt-get update        \
     && apt-get upgrade --yes \
     && apt-get install --yes \
-            curl
+            curl libtinfo-dev
```
ehildenb:
```suggestion
            curl             \
            libtinfo-dev
```

ehildenb (approved):

flake.nix
```diff
@@ -24,6 +31,7 @@
           llvm-backend-build-type = "Release"; })
         mavenix.overlay
         llvm-backend.overlays.default
+        # booster-backend.overlays.default
         haskell-backend.overlay # used only to override the z3 version to the same one as used by the haskell backend.
```
ehildenb, line 35: @goodlyrottenapple why is this comment added?

PR #3469, opened 2023-06-15
Remove unit applications in modules

### LINKED ISSUE #3421: Investigate constant-folding collection functions
If we see a function application like:
```
_List_(.List, L)
```
we should be able to constant-fold it in the compiler frontend to:
```
L
```

More generally, this would be nice for general collections with unit + append operations (possibly user-defined as well?)

gtrepta:
fixes: #3421


dwightguth (commented):
The function you want from RemoveUnit is literally just `flattenLists`. Apply that to the body, requires, and ensures and then return the transformed sentence. You do not need the rest of the pass (gatherVars, substituteWithUnit, etc).

radumereuta:
The regression tests are failing
```
[Error] Internal: unimplemented case in RemoveUnit: unit but no assoc
2023-06-23T19:45:06.4073374Z      [exec] make[4]: *** [/opt/workspace/k-distribution/pl-tutorial//../include/kframework/ktest.mak:78: kool-untyped-kompiled/timestamp] Error 113
2023-06-23T19:45:06.4074111Z      [exec] make[3]: *** [/opt/workspace/k-distribution/pl-tutorial//../include/kframework/ktest-group.mak:23: 1_untyped] Error 2
2023-06-23T19:45:06.4074797Z      [exec] make[2]: *** [/opt/workspace/k-distribution/pl-tutorial//../include/kframework/ktest-group.mak:23: 2_kool] Error 2
2023-06-23T19:45:06.4075964Z      [exec] make[2]: *** Waiting for unfinished jobs....
```

To test locally, open `kserver` on a terminal, and on another terminal type this: `time make -j 32 -O`
Locally, I build K like this: `mvn package -DskipTests -T1C`
Another useful cleaning command is `mvn clean -Dhaskell.backend.skip`

kernel/src/main/java/org/kframework/compile/RemoveUnit.java
```diff
@@ -0,0 +1,63 @@
...
+      @Override
+      public K apply(KApply k) {
+        Att att = m.attributesFor().getOrElse(k.klabel(), () -> Att.empty());
+
+        // Ignore optional cells, which have a unit attribute but no assoc
+        if (   att.contains(Att.CELL())
+            && att.contains(Att.MULTIPLICITY())
+            && att.get(Att.MULTIPLICITY()).equals("?")) {
+          return super.apply(k);
+        }
```
radumereuta, line 61: Why is this check necessary?
Can't we just remove throwing the exception below?
This is an optimization step and is not critical, so I don't think we should enforce `unit` with `assoc` present.
If `unit` must have `assoc`, then maybe it should be a verification step earlier in the kompile pipeline.

radumereuta (changes requested):
Please add some documentation to the file. A short description above the class should suffice.

Also left a comment. I'm not sure what's the best solution there. I would ignore unsafe applications since it's an optimization step.

kernel/src/main/java/org/kframework/compile/RemoveUnit.java
dwightguth: We don't have a way to flatten unit from lists that doesn't depend on the production being associative. That's why the check exists. I tend to agree that it's better to special case this one exception.

dwightguth (approved):

kernel/src/main/java/org/kframework/compile/RemoveUnit.java
radumereuta: Then the exception below should contain location information to the production.
So change `attributesFor` to `productionsFor`.
Then add a test in `regression-new/checks`.

radumereuta (approved):
lgtm

PR #3475, opened 2023-06-20
Introducing the `hs-backend-booster` to K Framework

### LINKED ISSUE #3470: Include booster backend in pom.xml build
Similar to how the `haskell-backend` is built, we need to:

- Clone the booster repo
- Build with `stack build` in a similar way to `haskell-backend/pom.xml`
- Modify the `k-distribution/src/main/assembly/bin.xml` to look for the appropriate binaries using the assemblies path.
- Add a CI test that the booster is launchable with the installed Ubuntu package.

Robertorosmaninho:
This PR introduces the Haskell Backend Booster to the K Framework repository!
Fixes #3470 with:
- [x] Clone the booster repo
- [x] Build with stack build in a similar way to haskell-backend/pom.xml
- [x] Modify the k-distribution/src/main/assembly/bin.xml to look for the appropriate binaries using the assemblies path.
- [x] Add a CI test that the booster is launchable with the installed Ubuntu package.


k-distribution/src/main/assembly/bin.xml
```diff
@@ -106,5 +106,11 @@
       <fileMode>775</fileMode>
       <directoryMode>775</directoryMode>
     </fileSet>
+    <fileSet>
+      <directory>${project.basedir}/../hs-backend-booster/target/stack-install</directory>
+      <outputDirectory>/bin</outputDirectory>
+      <fileMode>775</fileMode>
+      <directoryMode>775</directoryMode>
+    </fileSet>
```
Robertorosmaninho, line 114: Hey @goodlyrottenapple, besides these binaries, is there anything that should be included in the package? E.g., the old Haskell Backend includes `kore`, but I couldn't find anything correspondent to the old `kore` in the booster repo...

goodlyrottenapple: no the bin folder should be the only thing needed

Robertorosmaninho: Thanks!

.github/workflows/Dockerfile.stack-deps
```diff
@@ -27,3 +27,8 @@ ADD --chown=$USER:$GROUP haskell-backend/src/main/native/haskell-backend/stack.y
 ADD --chown=$USER:$GROUP haskell-backend/src/main/native/haskell-backend/kore/kore.cabal                     /home/$USER/.tmp-haskell/kore/
 ADD --chown=$USER:$GROUP haskell-backend/src/main/native/haskell-backend/kore-rpc-types/kore-rpc-types.cabal /home/$USER/.tmp-haskell/kore-rpc-types/
 RUN cd /home/$USER/.tmp-haskell && stack build --only-snapshot
+
+ENV LC_ALL=C.UTF-8
+ADD --chown=$USER:$GROUP hs-backend-booster/src/main/native/hs-backend-booster/stack.yaml                    /home/$USER/.tmp-booster/
+ADD --chown=$USER:$GROUP hs-backend-booster/src/main/native/hs-backend-booster/hs-backend-booster.cabal      /home/$USER/.tmp-booster/
```
goodlyrottenapple: we don't use a `*.cabal` file, instead we use `package.yaml` in the booster

goodlyrottenapple:
we need to do this for the booster in nix:

https://github.com/runtimeverification/k/blob/ccd26567ca350cdf36c2815b8fe24044b56df4c6/nix/k.nix#L22-L23

namely, disable building of the booster via maven, since we manually build and link the booster in nix.

jberthold (dismissed):
LGTM

radumereuta:
From Dwight:
Try to build the Debian package on your local machine and see why it doesn't work.
To build the package: `dpkg build-package` (approximate)
Inspect the directory structure and see what's missing.

We have to make sure that we're testing the booster on all distros.

jberthold:
There seems to be a serious problem with the mvn setup (CI does not even reach the build step). I'll test this locally, although I am not very fluent in mvn things.

hs-backend-booster/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.runtimeverification.k</groupId>
+    <artifactId>parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>hs-backend-booster</artifactId>
+  <packaging>jar</packaging>
```
jberthold, line 12: This seems to cause issues with the docker build setup in CI (`dependency:go-offline` fails due to not finding a jar with these coordinates). The `haskell-backend` directory indeed had some java code that ends up in a `jar` in the artefact store but `hs-backend-booster` does not have any.
Also, the `dependencies` section below is not necessary, the `hs-backend-booster` does not have any.

~From https://www.baeldung.com/maven-packaging-types we might want to use `pom` here (as a hack)?~
EDIT: does not work.

jberthold: [test-docker-setup.sh.txt](https://github.com/runtimeverification/k/files/11894154/test-docker-setup.sh.txt)
FYI script that I used to test locally

Robertorosmaninho:
In order to add the booster to the package we use to distribute K, we must add it to `k-distribution/pom.xml`. However, when building the docker image, the command `mvn --dependency:go-offline` tries to download every dependence from an AWS repository: https://s3.us-east-2.amazonaws.com/runtimeverificationmaven/snapshots

```
[ERROR] Failed to execute goal on project k-distribution: Could not resolve dependencies for project com.runtimeverification.k:k-distribution:jar:1.0-SNAPSHOT: Could not find artifact com.runtimeverification.k:hs-backend-booster:jar:1.0-SNAPSHOT in runtime.verification.snapshots (https://s3.us-east-2.amazonaws.com/runtimeverificationmaven/snapshots) -> [Help 1]
```
I've already tried locally:
- remove the  `--dependency:go-offline` and just use `mvn package`
- add the flag `-DexcludeArtifactIds=hs-backend-booster` to eliminate the need to download its dependency (this flag is exclusive from --dependency:go-offline)
- add the flag `-U/--update-snapshots`, which should: "Forces a check for missing releases and updated snapshots on remote repositories"

None of these works so far...

radumereuta:
The `hs-backend-booster` doesn't contain any java/scala code to be compiled and to produce a jar.
Maybe that's why it fails.
The other modules contain some java files with the cmd line options.
For the booster the kompilation is a bit more complicated though.
@Robertorosmaninho @jberthold thoughts? Might be worth a try.

Robertorosmaninho:
As @jberthold mentioned before, even packing it as `pom`, it doesn't work either. Maybe @dwightguth or @ehildenb knows better how to deal with it.

Robertorosmaninho:
As far as I understand, to add the booster to the package we distribute, we must add it as a dependence for the `k-distribution`.

However, when we set it as a dependence, the maven-cache script tries to download its cache from AWS, and it fails since we have yet to deploy it there. The same occurs to mount the image to run other tests as they also build K using the same flags `mvn dependency:go-offline`.

Therefore, the way I see it, we must deploy it to test it, which involves merging it, failing anyway, or temporarily turning off these CI tests. When the booster deploys on AWS, we activate the CI and test it properly.

Robertorosmaninho:
Do not automerge: NixOS 23.05 not working with this branch.

dwightguth (approved):
lgtm

hs-backend-booster/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+    <plugins>
+      <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>1.7</version>
+        <executions>
+          <execution>
+            <id>build-haskell-booster</id>
+            <phase>compile</phase>
+            <configuration>
+              <skip>${haskell.backend.booster.skip}</skip>
```
radumereuta: Can we make this shorter?
Something like `booster.skip`?
It's going to make it harder to type. And I tend to do that quite often when I need to test different versions.

Robertorosmaninho: Sure!

radumereuta (approved):

PR #3484, opened 2023-06-26
Ensure module attributes are checked against the whitelist

Scott-Guest:
Part of #2625 

This PR updates the attribute whitelist logic to also account for attributes on modules.


kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -166,6 +166,7 @@ object Att {
   final val FUNCTION = Key("function", KeyType.BuiltIn)
   final val FUNCTIONAL = Key("functional", KeyType.BuiltIn)
   final val GROUP = Key("group", KeyType.BuiltIn)
+  final val HASKELL = Key("haskell", KeyType.BuiltIn)
```
radumereuta, line 169: Is this being used anywhere as an attribute?

radumereuta (approved):
lgtm, with one question.

kore/src/main/scala/org/kframework/attributes/Att.scala
Scott-Guest: It's used in `domains.md`, e.g., on `MAP-KORE-SYMBOLIC`.

Dwight requested that I whitelist it for use with the Maude backend - both the Haskell backend and Maude backend are `symbolic`, so we need a separate attribute to distinguish modules only for use by the Haskell backend.

Issue #3344, opened 2023-04-17
Unicode support is inconsistent due to conflating Unicode strings and byte strings


Scott-Guest:
Unicode characters, bytes, and related escapes `\u,\U,\x` are handled inconsistently throughout K, causing a host of issues.

Specifically, in the LLVM backend,
- strings are stored as a `char[]`, so bytes are atomic
- Unicode characters are made to fit into a `char[]` by means of UTF-8 encoding
- `\xhh` is stored directly as the byte `0xhh`
- As a result, `"\U0001F642"` is exactly the same as its UTF-8 encoding `"\xF0\x9F\x99\x82"` and both have length 4

However, in pyk, the Java frontend, and the Haskell backend,
- strings are stored in a Unicode-aware string data type, so Unicode code points are atomic
- Unicode character are stored directly as their code points
- `\xhh` is stored as the Unicode code point `U+00hh`
- As a result, `"\U0001F642"` is distinct from `"\xF0\x9F\x99\x82"`, with the former length 1 and the latter length 4.

Per @dwightguth, we have a couple options here:
1. We can leave this difference alone. This ~~might cause~~ has caused issues for the Haskell backend booster though which calls out to the LLVM backend for concrete simplification.
2. We can try to implement a distinction between `\x` escapes for Bytes vs Strings. This might break some existing code in the LLVM backend, but it's probably fixable.
3. We can change the Haskell backend's meaning of `\x`. This would technically be a breaking change but it's unclear to me whether it would actually break anything in practice.
4. We can try to fully implement strings as Unicode in the LLVM backend. This is a much bigger change, part of which would consist of option 2 above, but it would be a lot more involved than that and would almost certainly break some LLVM backend code. Again, it might be fixable. This would be a much bigger undertaking though

~~For now, no one has actually been bitten by this issue, so we are going with Option 1.~~

Related issues:

* [x] #3396

goodlyrottenapple:
We do in fact store strings using the `ByteString` type in booster, which is a Haskell wrapper around C strings as we were bitten by this when passing data to the llvm backend: https://github.com/runtimeverification/hs-backend-booster/issues/92

Scott-Guest:
The frontend also treats `\xhh` the same as `U+00hh`: https://github.com/runtimeverification/k/blob/master/kore/src/main/java/org/kframework/utils/StringUtil.java#L244

[...]

PR #3489, opened 2023-07-05
Ensure that non-ASCII characters are escaped when emitting to KORE

Scott-Guest:
Part of #3344

Previously, any non-ASCII character present in a `String` or `Bytes` was erroneously emitted as-is to the KORE (`*.kore` files are presumed ASCII-only). This PR ensures that such characters are escaped instead.

That is, a string like `"!µƩ🌍\\"` used to be emitted as `\dv{SortString{}}("!µƩ🌍\\")}` but will now instead be emitted as `\dv{SortString{}}("!\xb5\u01a9\U0001f30d\\")`

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -2001,10 +2001,10 @@ public class ModuleToKORE {
                 sb.append("\\dv{");
                 convert(k.sort(), sb);
                 sb.append("}(");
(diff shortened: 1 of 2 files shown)


radumereuta (approved):
What are the implications for the backends now? Do they even care?

Scott-Guest:
> What are the implications for the backends now? Do they even care?

@radumereuta The backends and the KORE syntax specification already assume that all non-ASCII characters are given as escape sequences, so this is really just a bug fix in the frontend, and there're no further implications from this change alone.

(There is still a lot of work to be done re #3344 to actually get the semantics of these escape sequences handled correctly in the backends though).

PR #3494, opened 2023-07-06
Introducing Booster Backend

Robertorosmaninho:
Same as #3475


Baltoli:
Looks like the same issue unfortunately @Robertorosmaninho; perhaps try switching it to run on a hosted runner rather than public?

Robertorosmaninho:
> Looks like the same issue unfortunately @Robertorosmaninho; perhaps try switching it to run on a hosted runner rather than public?

How can I do it @Baltoli?

Baltoli:
`runs-on: [self-hosted, linux, normal]`

PR #3506, opened 2023-07-10
Upgrade debian version to 12 `bookworm`

Baltoli:
This PR bumps the version of Debian Stable we use for K packaging to the latest version (12; Bookworm). The changes entailed are as follows:
* Bulk renaming of packages, workflow steps etc. from `bullseye` to `bookworm`.
* Bump LLVM dependencies from 11 -> 14 in Debian.
* Change version tag and debian compat version in packaging.
* Bump JDK version from 11 -> 17 in Linux packaging generally, as 11 has been removed from Debian 12. Version 17 is available in all repositories where it's requested here (Focal, Jammy, Bookworm). If this is likely to cause unforeseen issues, we can implement a solution similar to what we do for LLVM and parameterize Dockerfiles on their required JDK version.

Commented below is a link to a (now removed) workflow run that demonstrates we can build the new Debian package successfully.


Baltoli:
Example run of bookworm packaging succeeding here: https://github.com/runtimeverification/k/actions/runs/5510999632/jobs/10045987280?pr=3506

README.md
```diff
@@ -76,7 +76,7 @@ On Ubuntu Linux 22.04 (Jammy):
 
 ```shell
-sudo apt-get install build-essential m4 openjdk-11-jdk libfmt-dev libgmp-dev libmpfr-dev pkg-config flex bison z3 libz3-dev maven python3 python3-dev cmake gcc g++ clang-14 lld-14 llvm-14-tools zlib1g-dev libboost-test-dev libyaml-dev libjemalloc-dev
+sudo apt-get install build-essential m4 openjdk-17-jdk libfmt-dev libgmp-dev libmpfr-dev pkg-config flex bison z3 libz3-dev maven python3 python3-dev cmake gcc g++ clang-14 lld-14 llvm-14-tools zlib1g-dev libboost-test-dev libyaml-dev libjemalloc-dev
```
dwightguth, line 79: it's a little weird telling people to install java 17 when we in fact support Java 11 or 17. I feel like this needs to be made a little more explicit in the places in the Readme we're modifying here.

dwightguth (commented):
Code looks okay; I'm a little concerned that we might need to make slightly more adjustments to the documentation to make sure it's clear enough what versions of Java we support.

README.md
Baltoli: We actually already have language in the README indicating that we support "11 or greater", in the section where we elaborate on the Java install instructions.

The stanza here is the "short version" instruction which is intended for users to copy-and-paste to install K's dependencies. The change here doesn't break the copy-paste run, but we can revert to 11 if you'd prefer.  In the section where we elaborate on how to install Java, I reverted the example instruction to version 11 to match the text.

dwightguth (approved):

Issue #3491, opened 2023-07-06
`kompile` pipeline performance improvements


Baltoli:
The increased usage of `kompile` in interactive contexts means that we need to improve its performance. Hotspots are currently scattered throughout, so we need to audit and figure out how to improve things.

* [x] #3517
* [ ] #3444 is likely to be relevant; ModuleToKore is a bit of a black box and splitting it up will help us
* [ ] There are some algorithmically inefficient helper functions somewhere in the frontend reports @radumereuta 
* [x] https://github.com/runtimeverification/llvm-backend/issues/792
* [x] https://github.com/runtimeverification/llvm-backend/pull/794
* [x] https://github.com/runtimeverification/llvm-backend/pull/796

palinatolmach:
For further performance profiling, the [`kompile-prof` folder](https://github.com/palinatolmach/foundry-test/tree/main/kompile-prof) contains files that are used when running `kompile` within `foundry-kompile`, when executed on the Foundry project located in the same [repository `foundry-test`](https://github.com/palinatolmach/foundry-test/tree/main) (that's the same project I used for performance tests before, the verbose logs for the kompilation are available [here](https://gist.github.com/palinatolmach/7342ff3e01e0b68269e0685ea7c9c616)). The zipped `kompile-prof` folder is attached here as well: [kompile-prof.zip](https://github.com/runtimeverification/k/files/12013831/kompile-prof.zip)

To execute `kompile` on these files used during `foundry-kompile`, run:

`cd kompile-prof && \` 
`kompile foundry.k --main-module FOUNDRY-MAIN --syntax-module FOUNDRY-CONTRACTS --md-selector 'k & ! node & ! concrete' --hook-namespaces 'JSON KRYPTO BLOCKCHAIN' --emit-json --backend haskell --verbose -I .`

(please add `--no-haskell-binary` if you're running it on Apple Silicon)

radumereuta:
Thank you Palina, I just ran some tests and there seem to be a spike in the `--emit-json` part.
Here are my test results:
```
Outer parsing [168 modules]                                  =  0.939s
Parse configurations [0/5 declarations]                      =  1.674s
  New scanner: FOUNDRY-MAIN-RULE-CELLS                       =  2.029s
Parse rules [1/2888 rules]                                   =  3.094s
Validate definition                                          =  1.573s
Apply compile pipeline                                       = 20.787s
  Emit parsed & compiled JSON                                = 43.295s
[...]

PR #3507, opened 2023-07-11
JSON creation - use object factory

radumereuta:
Make the JSON creation faster by using a factory when creating objects.
Tested with an extract from the EVM semantics found here: #3491
Before and after:
```
  Emit parsed & compiled JSON                                = 43.295s
  Emit parsed & compiled JSON                                =  1.985s
```
It's still slow, but it's progress.


Baltoli (approved):

Baltoli:
That's a big speedup!

radumereuta:
> That's a big speedup!

I was surprised by that as well. What a bit of profiling can do...

Robertorosmaninho (approved):
Awesome speedup!

PR #3512, opened 2023-07-12
Fix release build

Baltoli:
This PR addresses two issues with the release build:
* The Arch linux Maven dependency dockerfile wasn't copying the new booster `pom.xml`.
* On Ubuntu Focal, JDK 17 is incompatible at runtime with the default version of Maven. A comment noting this is added to the README, and all relevant install instructions have been updated.

I've tested the fixes by running them in the PR environment, then rebasing out. As best as I can tell without running the full release build, the issues are fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ On Ubuntu Linux 20.04 (Focal):
 
 ```shell
 git submodule update --init --recursive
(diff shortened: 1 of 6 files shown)


jberthold (approved):

README.md
```diff
@@ -188,6 +188,10 @@ See the notes below.
     This will provide the information about the JDK Maven is using, in case
     it is the wrong one.
 
+    Note that on Ubuntu Focal, the default Maven version is not compatible with
+    newer versions of the JDK; for K's Maven build to work, please ensure you
+    have version 11 installed.
+
```
radumereuta, line 194: 11-17 are fine to use. v18 is breaking the nailgun server.

Baltoli: 17 breaks maven 3.7, which is shipped only on Focal - it's a separate issue to Nailgun

PR #3514, opened 2023-07-12
Haskell backend dependency chain updates

ehildenb:
~Blocked on: https://github.com/runtimeverification/hs-backend-booster/pull/233~

This PR switches us to getting submodules/flake locks from release tags instead of submodules. In particular:

- The direct dependency on haskell backend is removed. Instead, the haskell backend version is read from the hs-backend-booster version, which ensures we get a single version of haskell backend always (instead of having conflicting versions). [This PR](https://github.com/runtimeverification/hs-backend-booster/pull/233) switches the booster over to getting auto-updates from the haskell backend, which is also recorded [here now](https://github.com/runtimeverification/devops/blob/master/deps/runtimeverification/hs-backend-booster.json).
- The release-tag is taken as the source of truth for the correct version of both the haskell bakcend and the booster backend, and that is synced to both the flake and the submodules.


jberthold (approved):
LGTM

Baltoli (approved):
Looks good to me!

Issue #3444, opened 2023-06-01
Audit ModuleToKore for non-trivial operations


Baltoli:
We currently do a lot of "extra" work inside ModuleToKore that would be better implemented as explicit compiler passes; Pyk doesn't need to do this, which indicates that ModuleToKore could be simplified down to a trivial Kast -> Text transformation.

The first step towards simplifying the code is to dig through it and find places where we're doing this extra work, and make it an explicit compiler pass _before_ ModuleToKore runs.

ehildenb:
Examples:

- Adding attributes: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L249
- Generating assoc/comm/idem/unit axioms: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L440
- Generating Ceil axioms: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L566
- Inserting priorities: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L698
- Generating no-junk/overloads: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L711
- Throwing errors in ModuleToKore can become checks over the Kast: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1274
- Adding Kore attributes: https://github.com/runtimeverification/k/blob/a017998b1e2c8c8e6c84a4349cd2374005fe595b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1493
[...]

PR #3522, opened 2023-07-18
AddKoreAttribute as a Frontend Kompiler Pass

Robertorosmaninho:
Part of #3444

This PR creates a new compiler pass that adds the Kore attributes to productions before `ModuleToKore`.  This is part of the effort to refactor the `ModuleToKore` infrastructure to simplify it by decoupling functions and actions that can be done before the translation and that aren't related to (kore) code emission.

Besides extracting the `AddKoreAttrubutes` function from `ModuleToKore.java`, this PR also simplifies function calls such as `isFunctional` and `isConstructor` only to test if the attribute is present on the given production.

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -373,7 +373,7 @@ public class ModuleToKORE {
     private void translateSymbol(Map<Att.Key, Boolean> attributes,
                                  KLabel label, Production prod, StringBuilder sb) {
         sb.append("  ");
-        if (isFunction(prod) && prod.att().contains(Att.HOOK()) && isRealHook(prod.att())) {
+        if (isFunction(prod) && prod.att().contains(Att.HOOK()) && module.isRealHook(prod.att(), immutable(options.hookNamespaces))) {
             sb.append("hooked-");
         }
         sb.append("symbol ");
@@ -744,17 +744,6 @@ public class ModuleToKORE {
         sb.append("(), ");
         convert(lesser.klabel().get(), lesser, sb);
         sb.append("())] // overloaded production\n");
-    }
-
-    private boolean isRealHook(Att att) {
-      String hook = att.get(Att.HOOK());
-      if (hook.startsWith("ARRAY.")) {
-        return false;
-      }
-      if (options.hookNamespaces.stream().anyMatch(ns -> hook.startsWith(ns + "."))) {
-        return true;
-      }
-      return Hooks.namespaces.stream().anyMatch(ns -> hook.startsWith(ns + "."));
     }
 
     private static boolean isBuiltinProduction(Production prod) {
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -186,17 +186,7 @@ public class ModuleToKORE {
         }
         translateSorts(tokenSorts, attributes, collectionSorts, semantics);
 
-        SetMultimap<KLabel, Rule> functionRules = HashMultimap.create();
-        for (Rule rule: iterable(module.sortedRules())) {
-            K left = RewriteToTop.toLeft(rule.body());
-            if (left instanceof KApply) {
-                KApply kapp = (KApply) left;
-                Production prod = production(kapp);
-                if (prod.att().contains(Att.FUNCTION()) || rule.att().contains(Att.ANYWHERE()) || ExpandMacros.isMacro(rule)) {
-                    functionRules.put(kapp.klabel(), rule);
-                }
-            }
-        }
+        SetMultimap<KLabel, Rule> functionRules = module.getFunctionRules();
 
         semantics.append("\n// symbols\n");
         Set<Production> overloads = new HashSet<>();
@@ -205,13 +195,13 @@ public class ModuleToKORE {
                 overloads.add(greater);
             }
         }
-        translateSymbols(attributes, functionRules, overloads, semantics);
+        translateSymbols(attributes, semantics);
 
         // print syntax definition
         syntax.append(semantics);
         for (Tuple2<Sort, scala.collection.immutable.List<Production>> sort : iterable(module.bracketProductionsFor())) {
             for (Production prod : iterable(sort._2())) {
-                translateSymbol(attributes, functionRules, overloads, prod.att().get(Att.BRACKET_LABEL(), KLabel.class), prod, syntax);
+                translateSymbol(attributes, prod.att().get(Att.BRACKET_LABEL(), KLabel.class), prod, syntax);
             }
         }
         for (Production prod : iterable(module.sortedProductions())) {
@@ -261,7 +251,7 @@ public class ModuleToKORE {
                 genFunctionalAxiom(prod, semantics);
             }
             if (prod.att().contains(Att.CONSTRUCTOR())) {
-                genNoConfusionAxioms(prod, noConfusion, functionRules, semantics);
+                genNoConfusionAxioms(prod, noConfusion, semantics);
             }
         }
 
@@ -368,8 +358,7 @@ public class ModuleToKORE {
         }
     }
 
-    private void translateSymbols(Map<Att.Key, Boolean> attributes, SetMultimap<KLabel, Rule> functionRules,
-                                  Set<Production> overloads, StringBuilder sb) {
+    private void translateSymbols(Map<Att.Key, Boolean> attributes, StringBuilder sb) {
         for (Production prod : iterable(module.sortedProductions())) {
             if (isBuiltinProduction(prod)) {
                 continue;
@@ -377,11 +366,11 @@ public class ModuleToKORE {
             if (prod.klabel().isEmpty()) {
                 continue;
             }
-            translateSymbol(attributes, functionRules, overloads, prod.klabel().get(), prod, sb);
+            translateSymbol(attributes, prod.klabel().get(), prod, sb);
         }
     }
 
-    private void translateSymbol(Map<Att.Key, Boolean> attributes, SetMultimap<KLabel, Rule> functionRules, Set<Production> overloads,
+    private void translateSymbol(Map<Att.Key, Boolean> attributes,
                                  KLabel label, Production prod, StringBuilder sb) {
         sb.append("  ");
         if (isFunction(prod) && prod.att().contains(Att.HOOK()) && isRealHook(prod.att())) {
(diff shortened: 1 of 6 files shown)


kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,331 @@
...
+    private Production production(KApply term) {
+        return production(term, false);
+    }
+    private Production production(KApply term, boolean instantiatePolySorts) {
+        KLabel klabel = term.klabel();
+        if (klabel.name().equals(KLabels.INJ))
+            return instantiatePolySorts ? INJ_PROD.substitute(term.klabel().params()) : INJ_PROD;
+        Option<scala.collection.Set<Production>> prods = module.productionsFor().get(klabel.head());
+        if (!(prods.nonEmpty() && prods.get().size() == 1))
+            throw KEMException.compilerError("Expected to find exactly one production for KLabel: " + klabel + " found: " + prods.getOrElse(Collections::Set).size());
```
Robertorosmaninho: For some reason, this Exception is thrown due to finding 0 production for KLabel.

kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
```diff
@@ -413,7 +413,8 @@ private void translateSymbol(Map<Att.Key, Boolean> attributes, SetMultimap<KLabe
         sb.append(") : ");
         convert(prod.sort(), prod, sb);
         sb.append(" ");
-        Att koreAtt = addKoreAttributes(prod, functionRules, overloads);
+        Att koreAtt = prod.att();//addKoreAttributes(prod, functionRules, overloads);
```
Robertorosmaninho: I'm having issues here, probably due to the Module transformer, as even though I can add the Attributes to the production, they don't appear updated here.

Robertorosmaninho:
I'm currently blocked on the two pending issues mentioned above.
I'm testing it with this [test](https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/cell-sort-haskell/test.k) as it was one of the first from `regression-new` to fail within this implementation.

The `definition.kore` reflect the second issue, as the symbols don't contain the appropriate attributes inserted by the AddKoreAttributes pass.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,331 @@
...
+        Option<String> hook = prod.att().getOption(Att.HOOK());
+        if (hook.isEmpty()) return false;
+        if (!hook.get().equals("MAP.in_keys")) return false;
+        return (!prod.klabel().isEmpty());
+    }
+
+    public synchronized Sentence add(Sentence s) {
+        Production prod = (Production) s;
+
+        List<Rule> sortedRules = new ArrayList<>(JavaConverters.seqAsJavaList(module.sortedRules()));
```
dwightguth: You want to call `mutable` from org.kframework.Collections which we import statically usually.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,331 @@
...
+        if (options.backend.equals("haskell")) {
+            module.sortedProductions().toStream().filter(this::isGeneratedInKeysOp).toStream().foreach(
+                    p -> {
+                        if (!options.disableCeilSimplificationRules) {
+                            genMapCeilAxioms(p, sortedRules);
```
dwightguth: Why is this being put here? This should be a separate refactoring. If you are somehow depending on these axioms being present, they should be placed in a separate pass in a separate pr that is refactored first and run first.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
dwightguth: I can't tell you why that might be without knowing what klabel it's throwing the error for. In general it's always better to share the entire exact error message rather than to just summarize in a way that leaves information out.

dwightguth:
As far as the issue with the module transformer is concerned, it is a known issue that passes that only edit the attributes of sentences in a module get cached incorrectly by the code that helps cache module transformations, as a result of which the changes to the attribute get discarded.

You have two options:

1. In a separate PR, try to fix the issue with module attribute caching. I don't remember exactly why I had trouble doing this when I tried, but I remember it wasn't as easy as I hoped it would be.
2. Bundle this pass together in the same module transformer as another change that edits something other than the attributes of every module it edits attributes in. This will bypass the issue with caching and correctly give back the right info.

Robertorosmaninho:
> As far as the issue with the module transformer is concerned, it is a known issue that passes that only edit the attributes of sentences in a module get cached incorrectly by the code that helps cache module transformations, as a result of which the changes to the attribute get discarded.
> 
> You have two options:
> 
> 1. In a separate PR, try to fix the issue with module attribute caching. I don't remember exactly why I had trouble doing this when I tried, but I remember it wasn't as easy as I hoped it would be.

That's interesting, could you provide me with more details about it?
- When/how did you face this issue before?
- What have you tried?
- How do we cache these module transformations?

Robertorosmaninho:
Blocked on #3523.
We need to fix the Transformation caching to update the modules when we only modify the sentence attributes!

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
Robertorosmaninho: Addressing it in #3529

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+
+            if (left instanceof KApply) {
+                KApply kapp = (KApply) left;
+                    Production prod2 = production(kapp);
+                    if (prod2.att().contains(Att.FUNCTION()) || rule.att().contains(Att.ANYWHERE())
+                            || ExpandMacros.isMacro(rule)) {
+                        functionRules.put(kapp.klabel(), rule);
+                    }
+            }
+        }
```
Robertorosmaninho: This will be optimized in a future PR. The idea is to move this and the production's functions to the module definition and call it here and in the ModuleKore file

radumereuta (approved):
lgtm, but I would like it if someone else could have a look.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+    private KList getAssoc(scala.collection.Set<Tuple2<Tag, Tag>> assoc, KLabel klabel) {
+        return KList(stream(assoc).filter(t -> t._1().name().equals(klabel.name())).map(t -> KApply(KLabel(t._2().name()))).collect(Collectors.toList()));
+    }
+
+    static boolean hasHookValue(Att atts, String value) {
+        return atts.contains(Att.HOOK()) &&
+                atts.get(Att.HOOK()).equals(value);
+    }
+
+    private void genMapCeilAxioms(Production prod, Collection<Rule> rules) {
```
dwightguth: This is now dead.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+        KLabel klabel = term.klabel();
+        if (klabel.name().equals(KLabels.INJ))
+            return instantiatePolySorts ? INJ_PROD.substitute(term.klabel().params()) : INJ_PROD;
+        Option<scala.collection.Set<Production>> prods = module.productionsFor().get(klabel.head());
+        if (!(prods.nonEmpty() && prods.get().size() == 1))
+            throw KEMException.compilerError("Expected to find exactly one production for KLabel: " + klabel + " found: " + prods.getOrElse(Collections::Set).size());
+        return instantiatePolySorts ? prods.get().head().substitute(term.klabel().params()) : prods.get().head();
+    }
+
+    private boolean isGeneratedInKeysOp(Production prod) {
```
dwightguth: this is now dead.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+
+    private boolean isFunction(Production prod) {
+        return prod.att().contains(Att.FUNCTION());
+    }
+
+    private KList getAssoc(scala.collection.Set<Tuple2<Tag, Tag>> assoc, KLabel klabel) {
+        return KList(stream(assoc).filter(t -> t._1().name().equals(klabel.name())).map(t -> KApply(KLabel(t._2().name()))).collect(Collectors.toList()));
+    }
+
+    static boolean hasHookValue(Att atts, String value) {
```
dwightguth: this is now dead.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+                                )
+                        )
+                        , BooleanUtils.TRUE
+                        , BooleanUtils.TRUE
+                        , Att.empty().add(Att.SIMPLIFICATION())
+                );
+        rules.add(ceilMapRule);
+    }
+
+    private boolean isRealHook(Att att) {
```
dwightguth: let's share this rather than duplicating it.

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
```diff
@@ -0,0 +1,319 @@
...
+        }
+        return Hooks.namespaces.stream().anyMatch(ns -> hook.startsWith(ns + "."));
+    }
+
+    public static final Production INJ_PROD = Production(KLabel(KLabels.INJ, Sort("S1"), Sort("S2")), Sort("S2"), Seq(NonTerminal(Sort("S1"))), Att());
+
+    private Production production(KApply term) {
+        return production(term, false);
+    }
+    private Production production(KApply term, boolean instantiatePolySorts) {
```
dwightguth: same here: don't duplicate

dwightguth (changes requested):

kernel/src/main/java/org/kframework/compile/AddKoreAttributes.java
Robertorosmaninho: Addressed in https://github.com/runtimeverification/k/pull/3522/commits/4e062801eea8b1d6fd165783f4fa7cb7bbc6b208

dwightguth (approved):

PR #3523, opened 2023-07-19
Compare Attributes on Transformers

Robertorosmaninho:
This PR is part of #3444.

It introduces an additional check to verify when updating the Module is necessary.
This solves an old issue on the caching transformation that wasn't updating the module when only attributes were modified on the `localSentences` of the module.

### DIFF
--- a/kore/src/main/scala/org/kframework/definition/outer.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer.scala
@@ -368,17 +368,14 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     case _ =>
   }
 
-  def checkAtts (other: Set[Sentence]) : Boolean = {
-    val sortedSentences = sentences.toSeq.sortBy(_.source.get)
-    val sortedOther = other.toSeq.sortBy(_.source.get)
+  def checkAtts(other: Set[Sentence]): Boolean = {
+    if (sentences != other)
+      return false
 
-    sortedSentences.zip(sortedOther).forall {
-      case (s1, s2) => if (s1.att != s2.att) return false
-                       else true
-      case        _ => false
-    }
-
-    true
+    val atts = sentences.collect({ case p: Production => p.att }).toSet
+    val otherAtts = other.collect({ case p: Production => p.att }).toSet
+    val diff = atts -- otherAtts
+    diff.isEmpty
   }
 
   def checkUserLists(): Unit = localSentences foreach {
--- a/kore/src/main/scala/org/kframework/definition/outer.scala
+++ b/kore/src/main/scala/org/kframework/definition/outer.scala
@@ -372,10 +372,20 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     if (sentences != other)
       return false
 
-    val atts = sentences.collect({ case p: Production => p.att }).toSet
-    val otherAtts = other.collect({ case p: Production => p.att }).toSet
-    val diff = atts -- otherAtts
-    diff.isEmpty
+    val productionsSorted = sentences.collect({ case p: Production => p }).toSeq.sorted
+    val otherProductionSorted = other.collect({ case p: Production => p }).toSeq.sorted
+
+    if (productionsSorted != otherProductionSorted)
+      return false
+
+    for (i <- productionsSorted.indices) {
+      val p1 = productionsSorted(i)
+      val p2 = otherProductionSorted(i)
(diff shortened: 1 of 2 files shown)


Robertorosmaninho:
The current generic way to transform the Module if any Attribute was changed introduced some unintentional behavior. E.g., Error message losing Location Att in [macroWithFunction.k](https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/checks/macroWithFunction.k).

I'll introduce a new function: `fromSentenceTransformerAtt` where we intentionally force the transformation to consider the update on Attributes, whereas the former  `fromSentenceTransformer` keeps its original behavior modifying the Module if the sentences are different regardless of its attributes state.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -368,6 +368,13 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     case _ =>
   }
 
+  def checkAtts (other: Set[Sentence]) : Boolean = {
+    val atts = sentences.collect({ case p: Production => p.att }).toSet
+    val otherAtts = other.collect({ case p: Production => p.att }).toSet
+    val diff = atts -- otherAtts
+    diff.isEmpty
```
dwightguth: this is... not how I would recommend doing this. You are checking that the set of all attributes on one module is a subset of the set of all attributes on another module, but you are not actually checking that the same sentences have the same attributes.

dwightguth (changes requested):

kore/src/main/scala/org/kframework/definition/transformers.scala
```diff
@@ -60,6 +58,19 @@ object ModuleTransformer {
     case f: ModuleTransformer => f
     case _ => new ModuleTransformer(f, name)
   }
+
+  def map(sentences: Set[Sentence])(f: (Module, Sentence) => Sentence, m: Module, name: String): Set[Sentence] =
```
Baltoli: I might consider giving this a different name; while it _is_ a map operation it's also responsible for catching and handling exceptions.

Robertorosmaninho: Would `applyFunctionToSentences` be a better name?

Baltoli: I was imagining something along the lines of `mapWithTrace`; the name `map` + the type signature _do_ suggest that the function applies a function to sentences, but doesn't identify that it also works on the stack trace if necessary. Not a huge deal though!

kore/src/main/scala/org/kframework/definition/outer.scala
Robertorosmaninho: I probably should check this in this function, but given this use
`if (newSentences != m.localSentences || !m.checkAtts(newSentences))`
we already know the two sets are "equal", no? And therefore, we are safely comparing that the sets of attributes are equal for the same sentences in the two modules. 

Although, I agree that there are probably better ways to compare them.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -372,10 +372,20 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
...
+    val productionsSorted = sentences.collect({ case p: Production => p }).toSeq.sorted
+    val otherProductionSorted = other.collect({ case p: Production => p }).toSeq.sorted
+
+    if (productionsSorted != otherProductionSorted)
+      return false
+
+    for (i <- productionsSorted.indices) {
+      val p1 = productionsSorted(i)
+      val p2 = otherProductionSorted(i)
+      if (!p1.att.equals(p2.att))
```
Robertorosmaninho, line 384: `Ordering` was already defined for `Production`, and in this case, it seems to be to only subclass from `Sentence` that we care about to check and modify the attributes if necessary. 

The comparison between `Attributes` is also already defined in the Attributes object.

dwightguth (approved):
This seems sufficient for the time being. Obviously it won't work if we later have a transformer that modifies only attributes of sentences other than productions, but we can address that situation if we encounter it.

Robertorosmaninho:
> This seems sufficient for the time being. Obviously it won't work if we later have a transformer that modifies only attributes of sentences other than productions, but we can address that situation if we encounter it.

Hey @dwightguth, I created an issue to track this here: #3530

PR #3529, opened 2023-07-21
[ModuleToKore Refactoring] Creatign `GenerateMapCeilAxioms` compiler pass

Robertorosmaninho:
This PR is part of #3444.

It creates the compiler pass `GenerateMapCeilAxioms` that call `GenMapCeilAxioms` before ModuleToKore.
This modification is necessary to simplify the `ModuleToKore` structure as this computation only happens when `Haskell Backend` is used and can be computed earlier.


Baltoli (commented):
This seems good to me, but I'll let @radumereuta or @dwightguth give the final approval

dwightguth (approved):

PR #3537, opened 2023-07-26
Refactoring collect token sorts and attributes

Robertorosmaninho:
Part of #3444

This PR is part of the effort to refactor the [`ModuleToKore.java`](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java) to be simpler and only contains the AST traversing to print out the kore syntax. The core of this modification is to create a new compiler pass for pre-compute the `SortTokens` and `Attributes` on the main module before the call of `ModuleToKore`.


kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -309,6 +308,45 @@ case class Module(val name: String, val imports: Set[Import], localSentences: Se
     }
     functionRules
   }
+
+
+  val hasAttributesMap: mutable.Map[Att.Key, java.lang.Boolean] = mutable.Map[Att.Key, java.lang.Boolean]()
```
radumereuta: One of the benefits of using scala is that it ensures immutability.
I would strongly avoid using mutable classes here.

dwightguth: Agree 100%. We should be using the existing attributes, or else a lazy val that computes some information about the module.

dwightguth (changes requested):
See my comment

kore/src/main/scala/org/kframework/definition/outer.scala
Robertorosmaninho: The problem is that we depend on the result of `RuleInfo` and `ExpandMacros` to know whether we should have an attribute on that list. Also, we depend on the `heatCoolEq` value.

Robertorosmaninho: These informations comes outside the information available in the module itself.

radumereuta: If this information is required in only one step, you can compute it locally.
Pretty much what RuleInfo does already. (IIRC)

Baltoli:
Superseded by more recent work on ModuleToKore.

PR #3552, opened 2023-08-01
Z3 tests radu

### LINKED ISSUE #3551: Parsing should not crash on failure to launch Z3 process
We're seeing a lot of flakiness on CI because we're hitting some OS file limit when launching too many z3 processes.

I guess this could be because th efrontend is launching a bunch of Z3 processes in a process pool, and triggering that limit.

One simple solution to this would be for the frontend to block when it fails to launch a Z3 process, and try again in a small time interval later, maybe a set amount of times. It could pause for 0.1s up to 5 times, trying to launch z3 processes.

We will also investigate being smarter about using `kserver` in pyk when launchnig a bunch of proofs in parallel, but the frontend not crashing on this case would also be nice.

F-WRunTime:
For testing.
I do not recommend merging this.
On a healthy system, we should not get the "too many open files" error.
The proper way to handle this, is to implement a custom Z3 type inferencer and remove the dependency completely.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
+++ b/kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
@@ -314,9 +314,12 @@ public class ParseInModule implements Serializable, AutoCloseable {
         if (scanner != null) {
             scanner.close();
         }
+        int infCount = 0;
         for (TypeInferencer inferencer : inferencers) {
             inferencer.close();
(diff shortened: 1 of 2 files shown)


radumereuta:
This is not the correct solution.
Making z3 restart when it fails just makes kompile crash on other opened files.
We should still try to reduce the number of file descriptors used though.

PR #3557, opened 2023-08-03
Fresh constants for configuration initializers

### LINKED ISSUE #3520: Segfault when using fresh variables in initial configuration
With the current design of the configuration initialisation mechanism, fresh variables won't work in the initial configuration as they rely on a cell being initialised.

I think the best thing to do here is just emit an error in the frontend if we detect fresh vars in the initial configuration; it _might_ be possible to reorder calls and guarantee that `initGeneratedCounterCell` is called first, but I strongly suspect not.

Reproducing example:
```
module REPRODUCER-SYNTAX
  imports INT-SYNTAX
endmodule

module REPRODUCER
  imports DOMAINS
  imports REPRODUCER-SYNTAX

  configuration
    <k> $PGM:Int </k>
    <heap> !_:Int </heap>
endmodule
```

gtrepta:
fixes #3520


k-distribution/tests/regression-new/freshConfig-fail/test.k.out
```diff
@@ -0,0 +1,4 @@
+[Error] Compiler: Can't resolve fresh configuration variable not of sort Int
+  while executing phase "Resolve fresh variables in cell initializers" on sentence at
+	<none>
+	<none>
```
radumereuta: What happened to the location here?

k-distribution/tests/regression-new/freshConfig/Makefile
```diff
@@ -0,0 +1,7 @@
+DEF=test
+EXT=test
+TESTDIR=.
+KOMPILE_FLAGS=--enable-search
+KRUN_FLAGS=--pattern "<k> V:K </k>"
```
radumereuta: Are these flags needed for this example? Or copy/paste leftovers?

k-distribution/tests/regression-new/freshConfig-fail/test.k
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) K Team. All Rights Reserved.
```
radumereuta: Can't we have this test in regression-new/checks?

k-distribution/tests/regression-new/freshConfig/test.k
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) K Team. All Rights Reserved.
```
radumereuta: Can we name the folder `issue-3520-freshConfig`?
So that we maintain some sort of convention. It could make it easier in the future to track down changes.

radumereuta (commented):
I like the solution a lot.
Added some minor comments/questions.

k-distribution/tests/regression-new/freshConfig-fail/test.k.out
gtrepta: The rule sentence for the cell initializer doesn't get generated with the source location. I tried adding it in [this commit](https://github.com/runtimeverification/k/pull/3557/commits/7c8b27303ca78712247718b76f332710b4b78edd) but it was breaking things.

k-distribution/tests/regression-new/freshConfig/Makefile
gtrepta: They're needed. I wanted the output to match after checking that each expected fresh constant is distinct, without their actual values being a factor.

k-distribution/tests/regression-new/checks/freshConfig-fail/test.k
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) K Team. All Rights Reserved.
```
radumereuta: You don't need to create a new folder, just rename the file and the module name like the other examples.

radumereuta (changes requested):
I modified ResolveAnonVar to carry over the location information from variables
```diff
@@ -104,16 +106,16 @@ public class ResolveAnonVar {
             @Override
             public K apply(KVariable k) {
                 if (ANON_VAR.equals(k)) {
-                    return newDotVariable("");
+                    return newDotVariable("", k);
                 }
                 if (FRESH_ANON_VAR.equals(k)) {
-                    return newDotVariable("?");
+                    return newDotVariable("?", k);
                 }
                 if (FRESH_ANON_CONSTANT.equals(k)) {
-                    return newDotVariable("!");
+                    return newDotVariable("!", k);
                 }
                 if (FRESH_LIST_VAR.equals(k)) {
-                    return newDotVariable("@");
+                    return newDotVariable("@", k);
                 }
                 return super.apply(k);
             }
@@ -121,9 +123,10 @@ public class ResolveAnonVar {
     }
 
     private int counter = 0;
-    KVariable newDotVariable(String prefix) {
+    KVariable newDotVariable(String prefix, K k) {
         KVariable newLabel;
-        Att att = Att().add(Att.ANONYMOUS());
+        Att att = Att().add(Att.ANONYMOUS()).add(Source.class, k.att().get(Source.class))
+                .add(Location.class, k.att().get(Location.class));
         if (prefix.equals("?")) {
             att = att.add(Att.FRESH());
         }
```
And got the error message to show like this:
```
[Error] Compiler: Can't resolve fresh configuration variable not of sort Int
  while executing phase "Resolve fresh variables in cell initializers" on sentence at
	<none>
	<none>
	Location(16,14,16,16)
	16 |	  <freshFoo> !_:Foo </freshFoo>
	   .	             ^~
```

radumereuta (approved):
looking great

PR #3559, opened 2023-08-04
Updating Maven credential references

F-WRunTime:
I believe we're just missing some variable references but otherwise looks good. 

Fixing previous release [here](https://github.com/runtimeverification/k/actions/runs/5756671238/job/15608802939#step:6:81)


.github/workflows/release.yml
```diff
@@ -395,13 +395,16 @@ jobs:
...
+          MAVEN_USERNAME: devops@runtimeverification.com
+          MAVEN_PASSWORD: ${{ secrets.CLOUDREPO_PASSWORD }}
         with:
           java-version: '11'
           distribution: 'adopt'
           overwrite-settings: true
           server-id: runtime.verification.snapshots
-          server-username: MAVEN_USERNAME
-          server-password: MAVEN_PASSWORD
+          server-username: $MAVEN_USERNAME
```
dwightguth: Are you sure this is correct usage? I am reasonably certain that these configuration fields are asking for a environment variable name rather than an actual value.

.github/workflows/release.yml
```diff
@@ -422,7 +425,7 @@ jobs:
           docker exec -t k-release-ci-${GITHUB_SHA} bash -c 'mkdir -p /home/github-runner/.m2'
           docker cp ~/.m2/settings.xml k-release-ci-${GITHUB_SHA}:/tmp/settings.xml
           docker exec -t k-release-ci-${GITHUB_SHA} bash -c 'mv /tmp/settings.xml /home/github-runner/.m2/settings.xml'
-          docker exec -e MAVEN_USERNAME -e MAVEN_PASSWORD -t k-release-ci-${GITHUB_SHA} bash -c "mvn --batch-mode deploy"
+          docker exec -e ${MAVEN_USERNAME} -e ${MAVEN_PASSWORD} -t k-release-ci-${GITHUB_SHA} bash -c "mvn --batch-mode deploy"
```
dwightguth: The reason this was the way it was was to avoid leaking secrets in the terminal output.

dwightguth:
Did you update the secret when we changed the password?

.github/workflows/release.yml
F-WRunTime: I updated this to reference password via direct call to secret.MAVEN_PASSWORD. GH runners should obscure this information when making references to a secret.

F-WRunTime: I'm thinking actually we'd have to set -e with an '='

F-WRunTime:
> Did you update the secret when we changed the password?

Good question. I don't think I did. This may have been the root of the issue. 

I do think however the `docker run` command needs values set for the environment variables

F-WRunTime:
Regularly hitting this failure of the Jammy Package build will likely need to extract this kore*.tar.gz file 
https://github.com/runtimeverification/k/actions/runs/5811624885/job/15756779786#step:4:6662

.github/workflows/release.yml
dwightguth: I am pretty sure that if you don't specify a value it uses the value from the host environment.

PR #3563, opened 2023-08-07
Use jdk17 in nix derivation

goodlyrottenapple:


radumereuta:
This needs to be rebased onto `develop` instead of master.

PR #3569, opened 2023-08-11
Upload failure tar file to review build issue

F-WRunTime:
Thoughts on other files that would be of help or an alternative approach. 

This is happening more frequently with little info to go on as to what is causing the issue.


dwightguth:
I'd rather add things to this list on an as needed basis instead of trying to guess at what artifacts we might want. This is good for now.

dwightguth (approved):

PR #3578, opened 2023-08-14
Remove some old unused attributes

gtrepta:


kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -183,15 +183,13 @@ object Att {
...
   final val CELL_NAME = Key("cellName", KeyType.BuiltIn)
-  final val CIRCULARITY = Key("circularity", KeyType.BuiltIn)
   final val COLOR = Key("color", KeyType.BuiltIn)
   final val COLORS = Key("colors", KeyType.BuiltIn)
   final val COMM = Key("comm", KeyType.BuiltIn)
   final val CONCRETE = Key("concrete", KeyType.BuiltIn)
   final val CONSTRUCTOR = Key("constructor", KeyType.BuiltIn)
   final val CONTEXT = Key("context", KeyType.BuiltIn)
   final val COOL = Key("cool", KeyType.BuiltIn)
-  final val DEPENDS = Key("depends", KeyType.BuiltIn)
```
Scott-Guest: `circularity` and `depends` should be kept - they're not in use yet but are under development

ehildenb: Here is an example that uses `circularity`: https://github.com/runtimeverification/pyk/blob/fd06d71322d85f41ecb652a3ac33d438c1bf0e4d/src/tests/integration/k-files/imp-simple-spec.k#L136

And one that uses `depends`: https://github.com/runtimeverification/pyk/blob/fd06d71322d85f41ecb652a3ac33d438c1bf0e4d/src/tests/integration/k-files/imp-simple-spec.k#L141

ehildenb:
Other than keeping `circularity` and `depends`, it looks good. I would ask @dwightguth to also have a look though, maybe some of these attributes mean something for LLVM backend.

dwightguth (approved):

dwightguth:
Looks good. The test failure is a known transient issue, just rerun the tests.

Scott-Guest (approved):

PR #3584, opened 2023-08-17
Fix Java 19

dwightguth:
This PR fixes K on Java 19 by dealing with the problems relating to nailgun and kserver that were causing it to crash. We do this primarily by updating to a newer version of nailgun with a fix for the issue that it had, but we need to call a different function to exit if Nailgun is active in order to make use of the fix, so there is a slight refactoring to how we call the exit function in K now.

Not tested yet; don't merge.

### DIFF
--- a/kernel/src/main/java/org/kframework/main/Main.java
+++ b/kernel/src/main/java/org/kframework/main/Main.java
@@ -64,6 +64,16 @@ public class Main {
         return isNailgun;
     }
 
+    private static NGContext context = null;
+
+    public static void exit(int exitCode) {
+      if (context != null) {
+          context.exit(exitCode);
+      } else {
+          System.exit(exitCode);
+      }
+    }
+
     public static void nailMain(NGContext context) {
         long startTime = System.nanoTime();
         KServerFrontEnd kserver = KServerFrontEnd.instance();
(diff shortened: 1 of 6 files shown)


kernel/src/main/java/org/kframework/main/Main.java
```diff
@@ -64,17 +64,28 @@ public static boolean isNailgun() {
...
         if (!kserver.isLocal()) {
             context.assertLoopbackClient();
         }
         isNailgun = true;
+        Main.context = context;
         if (context.getArgs().length >= 1) {
             String[] args2 = Arrays.copyOfRange(context.getArgs(), 1, context.getArgs().length);
             int result = kserver.run(context.getArgs()[0], args2, new File(context.getWorkingDirectory()), (Map) context.getEnv(), startTime);
-            System.exit(result);
+            context.exit(result);
```
radumereuta: What's the difference between `Main.exit` and `context.texit`?

radumereuta (approved):
lgtm

kernel/src/main/java/org/kframework/main/Main.java
dwightguth: context.exit can be called if you are 100% sure that you are running nailgun. It's probably best to be consistent though; I will fix this.

PR #3586, opened 2023-08-18
add TextToKore.parseModule

dwightguth:
Adds an entry point to TextToKore to parse a single module

### DIFF
--- a/kore/src/main/scala/org/kframework/parser/kore/parser/TextToKore.scala
+++ b/kore/src/main/scala/org/kframework/parser/kore/parser/TextToKore.scala
@@ -59,10 +59,22 @@ class TextToKore(b: Builders = DefaultBuilders) {
     parsePattern(io.Source.fromFile(file), line)
   }
 
-  /** Parses the file and returns [[kore.Pattern]]. */
+  /** Parses the string and returns [[kore.Pattern]]. */
   @throws(classOf[ParseError])
   def parsePattern(str: String): Pattern = {
     parsePattern(io.Source.fromString(str), 0)
+  }
+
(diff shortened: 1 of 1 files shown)


Scott-Guest (approved):
LGTM

kore/src/main/scala/org/kframework/parser/kore/parser/TextToKore.scala
```diff
@@ -73,23 +85,32 @@ class TextToKore(b: Builders = DefaultBuilders) {
...
+  /** Parses from the stream and returns [[kore.Pattern]]. */
   @throws(classOf[ParseError])
   def parsePattern(src: io.Source, line: Integer): Pattern = {
     try {
       scanner.init(src, line)
       parsePattern()
     } catch {
       case e: java.io.EOFException => throw new ParseError("ERROR: Unexpected end of file while parsing", e)
-      case exc: ParseError => throw exc
-      case exc: Throwable => throw new ParseError("ERROR: Unexpected error while parsing: " + exc.getMessage, exc) // shouldn't be reachable
```
Robertorosmaninho, line 92: Why were these cases deleted?

dwightguth: the first one isn't needed if not for the second one, and the second one is a bit too broad. It's generally a bad idea to catch anything that extends java.lang.Error, for example, because these exceptions typically represent serious failures of the JVM and are often not recoverable at all.

Robertorosmaninho: Got it! Thanks!

Issue #3481, opened 2023-06-22
Check that attributes are placed correctly on modules vs sentences


radumereuta:
We could consider splitting the attributes into `<modules> | <rule> | <syntax> ...`
Some of them can be in multiple categories.
We could also check if attributes require a value or are allowed to have values. This right now is handled by the step which uses the attribute.

gtrepta:
Here's a list of BuiltIn attributes found for each Sentence type that had attributes on them when running `mvn verify`.

|Context|Module|Production|Rule|SyntaxSort|
|-|-|-|-|-|
|avoid|concrete|alias|alias|cellCollection|
|color|haskell|applyPriority|anywhere|element|
|colors|kore|assoc|avoid|hook|
|format|not-lr1|avoid|color|locations|
|hybrid|private|binder|colors|token|
|klabel|symbolic|bracket|concrete|unit|
|latex||cell|context||
|left||cellCollection|cool||
|non-assoc||cellName|format||
|prefer||color|heat||
|right||colors|hybrid||
|seqstrict||comm|initializer||
|strict||element|klabel||
|||exit|label||
|||format|latex||
|||freshGenerator|left||
|||function|macro||
|||hook|non-assoc||
|||hybrid|non-executable||
|||idem|owise||
|||impure|prefer||
|||index|priority||
|||initializer|result||
|||injective|right||
|||internal|seqstrict||
|||klabel|simplification||
|||latex|smt-lemma||
|||left|stream||
|||macro|strict||
|||maincell|structural||
|||mlBinder|symbolic||
|||mlOp|unboundVariables||
|||multiplicity|||
|||no-evaluators|||
|||noThread|||
|||non-assoc|||
|||parser|||
|||prec|||
|||prefer|||
|||private|||
|||public|||
|||returnsUnit|||
|||right|||
|||seqstrict|||
|||smt-hook|||
|||smtlib|||
|||stream|||
|||strict|||
|||symbol|||
|||token|||
|||topcell|||
|||total|||
|||type|||
|||unit|||
|||unparseAvoid|||
|||unused|||
|||wrapElement|||

gtrepta:
Here are Internal attributes (minus Source/Location, which appear on all of them).

|Context|Module|Production|Rule|
|-|-|-|-|
|userList|digest|bracketLabel|UNIQUE_ID|
|||cellFragment|cool-like|
|||cellOptAbsent|org.kframework.definition.Production|
|||org.kframework.definition.Production|projection|
[...]

PR #3587, opened 2023-08-18
Make generated config attributes internal

gtrepta:
More stuff for #3481 . The `CheckAtt` step happens after the definition has been populated with sentences for the configuration, so some of the attributes it's checking weren't actually added by the user. Changing these attributes to internal will make some checks I'm adding for attribute/sentence restrictions a little cleaner.

There are `cell` and `internal` attributes used here in the `RULE-CELLS` module in `kast.md` which could be addressed later.
https://github.com/runtimeverification/k/blob/3d7841ae51afe44a64931a0692be5c41d6c7e67c/k-distribution/include/kframework/builtin/kast.md?plain=1#L287-L295

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
+++ b/kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
@@ -563,8 +563,6 @@ public class GenerateSentencesFromConfigDecl {
     private static KApply optionalCellInitializer(boolean initializeOptionalCell, Att cellProperties, String initLabel) {
         if (initializeOptionalCell) {
             return KApply(KLabel(initLabel), INIT);
-        } else if (cellProperties.contains(Att.INITIAL())) {
-            return KApply(KLabel(initLabel));
         } else {
             return KApply(KLabels.CELLS);
         }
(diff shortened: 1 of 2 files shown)


dwightguth:
I think the reason why these attributes weren't marked as internal is because it is intended that a configuration declaration is identical in semantics to the same set of productions and rules that it generates. That requires users to be able to write the productions that it corresponds to themselves.

dwightguth:
Also, the initial attribute isn't dead. It controls how the initializer for cells with multiplicity attribute is generated.

dwightguth (changes requested):
Marking this as requesting changes. See my comments above. I don't want to just close this PR on you, but I don't think this particular set of changes makes sense

gtrepta:
@dwightguth 

> a configuration declaration is identical in semantics to the same set of productions and rules that it generates. That requires users to be able to write the productions that it corresponds to themselves.

Why that requirement?

> Also, the initial attribute isn't dead. It controls how the initializer for cells with multiplicity attribute is generated.

Can you show me an example where the if block I deleted actually gets reached? It was introduced in a commit that added the ocaml backend, and it was never covered when I changed it to just throw an exception.

gtrepta:
Closing this for now.

dwightguth:
@gtrepta one reason why we might want to do what I described is if we ever want to parse a k definition, serialize the parsed definition to disk, then reload and finish compiling it later. I could very easily see pyk choosing to do this at some point in the future.

I'm not entirely sure where an example might exist, but essentially, if you have a cell of multiplicity `?`, the attribute controls whether the cell starts out initialized or starts out empty.

PR #3588, opened 2023-08-18
print no-junk with flat multi-or

dwightguth:
The Maude backend Kore-RPC server doesn't start with as much stack space as kompile, and this was leading it to not have enough stack space to parse the definition.kore that was generated by kompile, primarily because of the deeply nested \or patterns. In general, we want to try to avoid situations with potentially arbitrarily deeply nested KORE because it makes things more difficult for recursive descent parsers.

Luckily, we already have a solution to this problem. The llvm and haskell backend both support the \left-assoc symbol to be used in conjunction with \or, which allows a flat multi-or-like representation that still ultimately parses to the same term in the end. This allows the recursive descent parser to parse the axioms in question with a constant amount of stack space, no matter how many productions or sorts exist in the definition.

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -713,6 +713,9 @@ public class ModuleToKORE {
         sbTemp.append("  axiom{} ");
         boolean hasToken = false;
         int numTerms = 0;
+        sbTemp.append("\\left-assoc{}(\\or{");
+        convert(sort, sbTemp);
+        sbTemp.append("} (");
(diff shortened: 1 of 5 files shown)


dwightguth:
This PR is done but it can't be merged yet because it would break pyk. We need to have a discussion with that team and the haskell backend team because there are multiple competing solutions that could be implemented across all our tools to make this work and some of them would make this PR need to change.

dwightguth:
The PR for pyk is about to be merged so this is ready for review now.

radumereuta (approved):
Do we have any unit tests for this parser?
It looks like it would be easy to add.

radumereuta:
Tests are looking great. Thank you!

PR #3590, opened 2023-08-21
Ordering instance for Sentences

### LINKED ISSUE #3530: [K-Improvement] Create `Ordering` instance for `Sentence`'s subclasses
## Motivation
We may need in the future to sort all sentences from a module to test a property or apply a transformation.

Following the last comment of #3523.

gtrepta:
fixes: #3530


Robertorosmaninho:
@gtrepta This `Ordering.by` is the same as using `Compare` to compare 2 elements to sort them?

gtrepta:
> This Ordering.by is the same as using Compare to compare 2 elements to sort them?

I'm not sure what the Compare you're referring to is. But Ordering.by allows you to define a sorting strategy for a (normally ADT) type `T` in terms of a type `S`, and you give it a function `T => S`.

ex. For the case class `Tag(name: String)` you can make an ordering for it using its `name` field with `Ordering.by[Tag, String](t: Tag => t.name)`, so you use Scala's implicit ordering for `String` to sort the `Tag` datatype.

https://www.scala-lang.org/api/2.12.14/scala/math/Ordering.html

Robertorosmaninho:
> > This Ordering.by is the same as using Compare to compare 2 elements to sort them?
> 
> I'm not sure what the Compare you're referring to is. But Ordering.by allows you to define a sorting strategy for a (normally ADT) type `T` in terms of a type `S`, and you give it a function `T => S`.
> 
> ex. For the case class `Tag(name: String)` you can make an ordering for it using its `name` field with `Ordering.by[Tag, String](t: Tag => t.name)`, so you use Scala's implicit ordering for `String` to sort the `Tag` datatype.
> 
> https://www.scala-lang.org/api/2.12.14/scala/math/Ordering.html

Got it! But I'm not sure yet how we (or Scala implicitly) decide how to compare and sort the type K.
Also, we're missing the sorting of Attributes. TBH, this is a key feature that we really want to have here.

Robertorosmaninho (changes requested):
We really need to consider the Att on the objects to sort them. Even though it can be costly, it will be worse to sort them later regarding the Att.

By considering the Att, we can safely compare two objects for equality, and we can safely enable decoupling the Add Kore Attribute function on ModuleToKore.

gtrepta:
@Robertorosmaninho 

Ordering for K is here:
https://github.com/runtimeverification/k/blob/ab225b2c18d9bf3e02ca21f206a8647dccaa3f8b/kore/src/main/scala/org/kframework/kore/interface.scala#L33-L61

I can create an ordering for attributes as well. But, are you thinking about the attributes being factored in when ordering sentences? ~~Because I can't think of an occurrence where it would come down to that.~~ Just saw the change request, I'll implement it.

Robertorosmaninho:
> I can create an ordering for attributes as well. But, are you thinking about the attributes being factored in when ordering sentences? ~Because I can't think of an occurrence where it would come down to that.~ Just saw the change request, I'll implement it.

To give you more context, we had to revert these 2 PRs(https://github.com/runtimeverification/k/pull/3523, https://github.com/runtimeverification/k/pull/3522) due to the slowdown in the `kompile` time. One of the main reasons was that we had to do a linear pass to all productions and compare every attribute list just to check if we had or not to apply the compiler pass.

Robertorosmaninho:
Nevertheless, @gtrepta, I would like to have a third opinion on this, perhaps @dwightguth or @radumereuta can help us on this.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -409,18 +409,56 @@ trait Sentence extends HasLocation with HasAtt with AttValue {
   def label: Optional[String] = att.getOptional(Att.LABEL)
 }
 
+object Sentence {
+  implicit val ord = new Ordering[Sentence] {
+    def compare(a: Sentence, b: Sentence): Int = {
+      (a, b) match {
+        case (_:SyntaxSort, _) => -1
```
dwightguth, line 426: This doesn't feel quite right. Shouldn't it have a case where both are the same and it compares their contents?

dwightguth (changes requested):

kore/src/main/scala/org/kframework/definition/outer.scala
gtrepta: I had them there but removing them didn't break the unit tests I wrote, so I figured when they're the same type then the ordering for that specific type is used instead of going up the type heirarchy.

radumereuta:
Can you remind me why the ordering is needed? To make comparisons faster?
This code touches the core data structures of kompile, and they are very sensitive to performance losses.
Before we merge this, it needs to be tested with the c-semantics and the evm-semantics.
Run `kompile -v`. The step `Apply compile pipeline` should not be slower.

gtrepta:
@radumereuta The compile pipeline didn't run any slower on those semantics.

Robertorosmaninho:
Hey @gtrepta, this test is failing rn, any ideas on the reason?
```
[exec] make[1]: Entering directory '/opt/workspace/k-distribution/tests/builtins/collections'
     [exec] set -o pipefail; (cat test-set2list-2.col.in 2>/dev/null || true) | /opt/workspace/k-distribution/bin/krun test-set2list-2.col --no-exc-wrap  --definition ./collections-test-kompiled | diff - test-set2list-2.col.out
     [exec] 2,3c2
     [exec] <   ListItem ( 1 )
     [exec] <   ListItem ( 2 )
     [exec] ---
     [exec] >   ListItem ( 4 )
     [exec] 5c4,5
     [exec] <   ListItem ( 4 ) ~> .
     [exec] ---
     [exec] >   ListItem ( 2 )
     [exec] >   ListItem ( 1 ) ~> .
     [exec] make[1]: Leaving directory '/opt/workspace/k-distribution/tests/builtins/collections'
```

gtrepta:
> Hey @gtrepta, this test is failing rn, any ideas on the reason? ...

It's most likely some nondeterminism from printing/converting sets, the expected output only recognizes one of multiple possible outcomes. I've seen plenty of tests break in this fashion, normally we just update the expected output in the PR that it's breaking in.

Robertorosmaninho (approved):
LGTM! Thanks for your contribution on this @gtrepta!

kore/src/main/scala/org/kframework/definition/outer.scala
gtrepta: Looks like we do need those extra cases. I pushed the fix.

dwightguth (approved):

PR #3596, opened 2023-08-23
Rule label cannot contain ` or whitespace

dwightguth:
Adds a check to the compiler that reports an error if a sentence label contains the backtick character or whitespace. This ought to allow the Maude backend to compile rule labels directly to Maude identifiers without some clumsy encoding system.

I am reasonably certain that it should not break anything to add this restriction.


Robertorosmaninho (approved):
LGTM!

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -58,6 +59,18 @@ public void check(Sentence sentence) {
         } else if (sentence instanceof Production) {
             check((Production) sentence);
         }
+        checkLabel(sentence);
+    }
+
+    private static final Pattern whitespace = Pattern.compile("\\s");
```
radumereuta, line 65: Why didn't you include the backtick here?
You have to traverse the label twice now.

dwightguth: it's still linear time, the string is unlikely to be long, and I doubt the impact on performance will be substantial. We can always address this later if need be. I suspect the amdahl's law on this will be pretty minimal though.

PR #3597, opened 2023-08-23
Internal left/right attributes for c++ unparser

gtrepta:
for #3481 

This creates internal left/right attributes that allow parameters which get passed down to the backends.


kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -364,26 +366,32 @@ object Att {
...
 
-  private val builtinKeys: Set[String] = keys.filter(_._2.keyType == KeyType.BuiltIn).keySet
-  private val internalKeys: Set[String] = keys.filter(_._2.keyType == KeyType.Internal).keySet
+  private val internalKeys: Map[String, Key] =
+    Att.getClass.getDeclaredFields
+      .filter(f => f.getType.equals(classOf[Key]) && pat.matcher(f.getName).matches())
+      .map(f => f.get(this).asInstanceOf[Key])
+      .filter(_.keyType == KeyType.Internal)
+      .map(k => (k.key, k))
+      .toMap
```
Scott-Guest: I'd prefer to avoid the code duplication here. We could keep the orignal `keys` and just remove the calls to `keySet` instead:

```
  private val builtinKeys: Map[String, Key] = keys.filter(_._2.keyType == KeyType.BuiltIn)
  private val internalKeys: Map[String, Key] = keys.filter(_._2.keyType == KeyType.Internal)
  ```

Scott-Guest (approved):
One nitpick but otherwise LGTM

kore/src/main/scala/org/kframework/attributes/Att.scala
gtrepta: There are now both Internal and BuiltIn versions of the left/right keys, so building the original `keys` causes one to kick out the other because they're keyed on the same string.

It's only one duplicate, and it's very small, which is not a big deal imo.

Scott-Guest: Ah I completely missed that, all good then

gtrepta: I played around with it a bit and came up with something that does the job. 704d28a116c50b081799a928143cadc89abf6cef

dwightguth (approved):

PR #3602, opened 2023-08-24
Make no junk right associative

dwightguth:
The recent changes broke the haskell backend because they changed the associativity of the no junk axioms. We are hoping that this will address those issues.


dwightguth:
@radumereuta @goodlyrottenapple this is ready for review but I don't want it merged until it's confirmed to fix Sam's issue.

radumereuta (approved):

goodlyrottenapple:
https://github.com/runtimeverification/haskell-backend/actions/runs/5975579143/job/16211825982 <- can confirm this fixed the issue

Issue #3517, opened 2023-07-13
Build `kompile` performance tracker


Baltoli:
LLVM has a [webpage](https://llvm-compile-time-tracker.com/) where the change in compilation performance over time is tracked against a set of benchmarks. Currently, we don't have a good idea of how the `kompile` pipeline performs, although we are [doing some work](https://github.com/runtimeverification/k/issues/3491) on improving our insight into this.

It would be useful to implement our own version of this tool so that significant performance regressions (or, improvements!) can be identified systematically.

A rough idea of what we'd need to do to implement this is:
* [x] Identify a set of benchmark definitions that can be fed into `kompile` with minimal setup or tinkering (C, EVM, IMP, Foundry proofs...).
* [x] Work out a metric to measure (mean wall clock time over _N_ repetitions as a simple first cut).
* [x] Write scripts to do the benchmarking and write the results into a file somewhere (for now, can probably just be an append-only log in a git repo for simplicity)
* [x] Write scripts to generate some static HTML from those results and commit them to a Github pages branch.
* [ ] Run the above on a `workflow_dispatch` from K release jobs (or commits to master, etc.)

Robertorosmaninho:
The current status for this task is as follows:
 - We decided to use [IMP](https://github.com/runtimeverification/imp-semantics), [Elrond](https://github.com/runtimeverification/elrond-semantics), [KEVM](https://github.com/runtimeverification/evm-semantics), and K's [IMP lesson 5](https://github.com/runtimeverification/k/tree/master/k-distribution/pl-tutorial/1_k/2_imp/lesson_5) and [LAMBDA lesson 9](https://github.com/runtimeverification/k/tree/master/k-distribution/pl-tutorial/1_k/1_lambda/lesson_9).
 - For now, we will work within two metrics: Elapsed time to build the semantics (for LLVM Backend) and its maximum [RSS](https://en.wikipedia.org/wiki/Resident_set_size), both got from Unix `time`. The experiments will initially run ten times, and we'll take the mean of each metric.
 - We'll use the framework [Bencher](bencher.dev) to log, compare and display the result of the benchmarks.

ehildenb:
Instead of cloning the remote semantics repositories (which is network intensive), we can instead do `kompile -E ...` on these definitions, and save the resulting k files directly in the K repo.

The definitions may need to be tweaked to remove the things that caused them to break when removed from their own build systems.

ehildenb:
[...]

PR #3603, opened 2023-08-24
Adding profiling tests

Robertorosmaninho:
Part of #3517
Introduce tests to work as a benchmark for profiling `kompile` build time and space consumed.
These tests were obtained with `kompile -E` on each semantic/test and are the following:
 - IMP-Semantics
 - EVM-Semantics
 - AVM-Semantics
 - WASM-Semantics
 - ELROND-Semantics
 - PL-Tutorial-IMP: Lesson 5
 - PL-Tutorial-LAMBDA: Lesson 9
 
 A new make command is introduced here: `make profile`, which runs `make clean && /usr/bin/time --format x --output y make kompile`.

Where `y` is the file name `.profiling-results.json` which contains the information in this format `x`:
```
{
        "IMP-SEMANTICS": {
                "build-time": {
                        "value": 2.85
                },
                "max-resident-set-size": {
                        "value": 552348
                }
        }
}
```

This output file will be consumed later by a framework to measure and compare the performance results.


.github/workflows/test-pr.yml
```diff
@@ -130,6 +130,57 @@ jobs:
           name: kore-exec.tar.gz
           path: |
             **/kore-exec.tar.gz
+      - name: 'On Success, Upload the package built to the Summary Page'
```
F-WRunTime, line 133: For every .deb doing this on our org will start to consume far more space than we have available. Is this necessary? If so for what purpose? 
We should consider either utilizing the deb, which I think is not removed from the docker image its installed to or push these debs elsewhere, if they're needed. On Every PR publishing a deb is too much space usage on GH

F-WRunTime: If we just need the deb installed why not publish the docker image that is already built in previous steps and tested and use this in the later job?

F-WRunTime: Another option is set:
```retention-days: 1 ```

Comment (Robertorosmaninho) [reply .github/workflows/test-pr.yml:133]: This last option seems more reasonable. I'm not sure what you mean with

> which I think is not removed from the docker image it's installed 

But I don't think we should consider publishing images from PR tests as well.

Comment (Robertorosmaninho) [reply .github/workflows/test-pr.yml:133]: Would it be okay for you if I set the retention of the artifact for just one day? If less than this is possible, it would also be fine, as we only need it for the current execution. If not, you can say better than I do, but I don't think we have enough PR testing volume to consume all the space that we have available in one day...

Comment (F-WRunTime) [reply .github/workflows/test-pr.yml:133]: I think starting with 1 day is reasonable and see how it goes. I can elaborate further on my semi-ramble piece about not being removed from the image. 

I was a bit all over what I meant is the deb is installed to the docker image for testing with the 'test-package' action. This is usable if we published it to our internal docker image. We have the dockerhub private registry and one we use internally with our build servers. We can speak in a private forum to discuss using this.

Comment (F-WRunTime) [reply .github/workflows/test-pr.yml:133]: Publishing images would be useful as it would be a retention in a space where we have more wiggle room. And an easier control of the space usage. It's semi-difficult to sort through uploaded artifacts and remove. A quicker ramp-up to getting into an environment with k installed for a dev who needs to get in quick, and look into something; This also actually would enable easier down stream testing of PR specific changes in other repositories by making available internally our PR builds.

**Review by F-WRunTime — APPROVED** [review]

File: .github/workflows/test-pr.yml [hunk 2]
```diff
@@ -130,6 +130,54 @@ jobs:
...
+        if: success()
+        uses: actions/upload-artifact@v3
+        with:
+          name: kframework.deb
+          path: kframework.deb
+          retention-days: 1
```
dwightguth, line 140: I would add `if-no-files-found: error` to this

.github/workflows/test-pr.yml
```diff
@@ -130,6 +130,54 @@ jobs:
...
+          workspace=$(pwd)
+          docker run --name k-profiling-tests-${GITHUB_SHA}               \
+            --rm -it --detach                                             \
+            --workdir /opt/workspace                                      \
+            -v "${workspace}:/opt/workspace"                              \
+            ${BASE_OS}:${BASE_DISTRO}
+      - name: 'Install K from Package'
+        run: |
+          set -euxo pipefail
+          docker exec -t k-profiling-tests-${GITHUB_SHA} /bin/bash -c 'export DEBIAN_FRONTEND=noninteractive'
```
dwightguth: something seems wonky here. You are running `export` inside a shell that's about to terminate. The exported value is going to get ignored in the commands below. You want to either combine the commands with `&&` or else create a script.

README.md
```diff
@@ -103,7 +103,7 @@ On macOS using [Homebrew](https://brew.sh/):
 
 ```shell
-brew install bison boost cmake flex fmt gcc gmp openjdk jemalloc libyaml llvm@15 make maven mpfr pkg-config python stack zlib z3
+brew install bison boost cmake flex fmt gcc gmp gnu-time openjdk jemalloc libyaml llvm@15 make maven mpfr pkg-config python stack zlib z3
```
dwightguth: time is just used for the profiling, right? which we are not doing on MacOS, so we shouldn't need to change this list of dependencies.

dwightguth (changes requested):
We don't need IMP twice.

.github/workflows/test-pr.yml
Robertorosmaninho: My bad, I'll create a script to clean these calls!

README.md
Robertorosmaninho: Got it!

dwightguth (approved):

PR #3611, opened 2023-08-30
Add Bencher to track Benchmarks Performance Regression

### LINKED ISSUE #3517: Build `kompile` performance tracker

Robertorosmaninho:
Fixes #3517
Follow up of #3603

This PR Introduces Bencher.dev to track performance regressions on our benchmarks.
The Benchmarks already run in the Test-PR Workflow as "Performance Tests" CI. This PR introduces a third-party program that will collect the JSON results of these executions and display them in a graphical online tool. 

It will also set a threshold for the time/space that each test, and if an execution passes this limit, the tool will post a comment on the PR  warning the developer which tests the regression was observed. The CI will also fail, which won't prevent the PR from being merged.


Robertorosmaninho:
Add -v to kompile

k-distribution/tests/profiling/pl-tutorial/imp/Makefile
```diff
@@ -1,5 +1,6 @@
 DEF=test
 EXT=test
+ITER=5
```
radumereuta, line 3: I would not iterate calling kompile multiple times. Running it once with the verbose option already tells us a lot of things.
And the way it's set up right now, it's reusing the caches from the previous call.
```
Importing: Source(/opt/workspace/k-distribution/tests/profiling/pl-tutorial/imp/test.k)
Outer parsing [64 modules]                                   =  1.212s
Parse configurations [0/3 declarations]                      =  0.519s
  New scanner: IMP-RULE-CELLS                                =  0.459s
Parse rules [0/89 rules]                                     =  0.569s
```
If you really want to iterate though, you can either delete the cache directory (*-kompiled) or you can add
`--profile-rule-parsing <file>` as an option to kompile.

radumereuta (changes requested):

k-distribution/tests/profiling/pl-tutorial/imp/Makefile
Robertorosmaninho: Thanks for pointing out that issue, Radu!
Unfortunately, I'm unfamiliar with this compiler flag, and the instructions in `kompile -hh` don't mention the need for a file.
```
$ kompile -hh
...
 --profile-rule-parsing
      Generate time in milliseconds to parse each rule in the semantics.
```

This message error seems a bit unclear to me as well:

```
$ kompile -v --profile-rule-parsing test.k --main-module TEST

Parse command line options                                   =  0,014s
outer parsing.
```

Robertorosmaninho: Also, the idea of iterating it multiple times is to ensure we're collecting a fair compilation time. That's a very common practice when executing performance benchmarks and essential when writing papers. Bencher can store the results of each execution, calculate their mean, and only output the result in the visual graph.

.github/workflows/test-pr.yml
```diff
@@ -168,7 +182,7 @@ jobs:
       - name: 'Profiling Performance Tests'
         run: |
           docker exec -t k-profiling-tests-${GITHUB_SHA} /bin/bash -c 'spawn-kserver kserver.log'
-          docker exec -t k-profiling-tests-${GITHUB_SHA} /bin/bash -c 'cd k-distribution/tests/profiling && make -j`nproc` --output-sync'
+          docker exec -t k-profiling-tests-${GITHUB_SHA} /bin/bash -c 'cd k-distribution/tests/profiling && make'
```
radumereuta, line 185: When you run things in series, which you should here, you don't need to spawn the kserver.
It can mess with the timing because it keeps caches in memory.

k-distribution/tests/profiling/pl-tutorial/imp/Makefile
radumereuta: The output for the help command is handled automatically by the `jcommnader` library here: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java
You could try to change the description to something like:
`Store in this file time taken in ms to parse each rule in the semantics.`

Robertorosmaninho:
@radumereuta I've addressed your comments, and I'm now removing the kompiled directory after each iteration. Although, I think the appropriate solution would be a new frontend flag like `--no-cache` or `--do-not-use-cache`. 

The flag `--profile-rule-parsing` indeed does this, but it does more than we want as well, so it's a workaround as much as deleting the directory.  I choose the last option as it's more strict for our purpose here.

radumereuta (approved):
lgtm
Deleting the directory is the safest option. It's too easy to mess something up in the code and get erroneous results.
I would like it if someone else also had a look at the PR.
Great work, though.

F-WRunTime (approved):
I can't comment on the lower level Makefiles but the changes to the workflow look reasonable

dwightguth (approved):

dwightguth:
Looks good, but where is the link to the results?

Robertorosmaninho:
> Looks good, but where is the link to the results?

The link for each result is generated after every execution of `bencher run` on the `Performance Tests` CI under the `Profiling Performance Tests` step as:
```
View results:
- AVM-SEMANTICS (Elapsed Real time (wall clock) in Seconds): https://bencher.dev/console/projects/k-framework/perf?metric_kind=build-time&branches=7fba2f4e-5e87-4782-92a7-9dfc4c8aaa1d&testbeds=d9eea46c-dd6c-4d0e-a830-30581a4e4446&benchmarks=3744f607-9385-4366-a682-5128ce2df25f
- AVM-SEMANTICS (Maximum Resident Set Size in KB): https://bencher.dev/console/projects/k-framework/perf?metric_kind=max-resident-set-size&branches=7fba2f4e-5e87-4782-92a7-9dfc4c8aaa1d&testbeds=d9eea46c-dd6c-4d0e-a830-30581a4e4446&benchmarks=3744f607-9385-4366-a682-5128ce2df25f
```


.
But for a general purpose, we can define this as the [Main Benchmark Page](https://bencher.dev/console/projects/k-framework/perf?reports_per_page=4&reports_page=1&branches_per_page=8&branches_page=1&testbeds_per_page=8&testbeds_page=1&benchmarks_per_page=8&benchmarks_page=1&clear=true&metric_kind=build-time&branches=7fba2f4e-5e87-4782-92a7-9dfc4c8aaa1d&testbeds=d9eea46c-dd6c-4d0e-a830-30581a4e4446&benchmarks=3744f607-9385-4366-a682-5128ce2df25f%2Ca1fd17f7-c861-4001-8be3-f5c7d92e09b7%2C1b9a644c-5b60-4cb4-ab1a-3f12ad73213c%2Cdc92b77a-2913-4af5-8bf7-5cd85d2112b3%2C92ab8514-2848-46b4-8b04-1e7d84407067%2C29feeefd-7ac2-4aca-9b7b-ac95826f2a41&tab=benchmarks).

PR #3612, opened 2023-08-30
Fix edge case in parsing comments in bubbles

### LINKED ISSUE #3501: Allow `"/"` only as the first character of a `BUBBLE` token
Redefine [`BUBBLE`](https://github.com/runtimeverification/k/blob/7e7628ca60a207ac7f5c2624dc11ba1190aef491/kernel/src/main/javacc/Outer.jj#L583) as

```
<BUBBLE: ~[" ", "\t", "\r", "\n"] (~["/", " ", "\t", "\r", "\n"])*>
```

## Motivation

Consider

```k
module TEST-SYNTAX
    imports STRING-SYNTAX
endmodule

module TEST
    imports TEST-SYNTAX
    rule/*1*/ "foo" => "bar"
    rule/*2*/ "bar" => "baz"
endmodule
```

Kompiling this definition gives the following error:

```
[Error] Inner Parser: Parse error: unexpected token 'rule' following token
'"bar"'.
        Source(/home/ttoth/git/k/test.k)
        Location(8,5,8,9)
        8 |         rule/*2*/ "bar" => "baz"
          .         ^~~~
```

which can be confusing as having only `rule/*1*/` in the definition parses fine.

The issue is that the first rule is outer parsed as:

```
"foo" => "bar"
    rule/*2*/ "bar" => "baz"
```

This is because JavaCC tries to match the longest possbile token, hence instead of lexing `rule` as `RULE` (and then `/*2*/` as `COMMENT`), it will lex `rule/*2*/` as `BUBBLE`. Note that the first `rule` token is lexed as `RULE` as expected, because in the `DEFAULT` lexer state there is no permissive token like `BUBBLE` to overlap with `COMMENT`.

## Analysis

There are two things to consider:
* Given a token that so far lexed as `COMMENT`, it will still lex as `COMMENT`: `/**/` could only be lexed as `(BUBBLE, "/**"), (BUBBLE, "/")` otherwise, but `"/**"` is shorter than `"/**/"`.
* Except for a comment (or comments) at the end of the bubble not separated by whitespace, all other characters that so far belonged to the bubble will still do: even though `bubble/*comment*/` will no longer be lexed as a single `BUBBLE`, if there are suitable tokens following, the comment will be appended to the bubble.

gtrepta:
Fixes #3501

### DIFF
--- a/k-distribution/tests/regression-new/issue-3501-bubble-blockcomments/test.k
+++ b/k-distribution/tests/regression-new/issue-3501-bubble-blockcomments/test.k
@@ -1,0 +1,14 @@
+// Copyright (c) K Team. All Rights Reserved.
+module TEST-SYNTAX
+    imports STRING-SYNTAX
+
+    syntax Strings ::= List{String,","}
+endmodule
+
+module TEST
+    imports TEST-SYNTAX
+    rule/*1*/ "foo" => "bar"
+    rule/*2*/ "bar" => "baz"
+
+    rule "foo"/*1*/,"bar" => "baz"
+endmodule
--- a/k-distribution/tests/regression-new/issue-3501-bubble-blockcomments/Makefile
+++ b/k-distribution/tests/regression-new/issue-3501-bubble-blockcomments/Makefile
@@ -1,0 +1,4 @@
+DEF=test
+EXT=test
+
+include ../../../include/kframework/ktest.mak
--- a/kernel/src/main/javacc/Outer.jj
+++ b/kernel/src/main/javacc/Outer.jj
@@ -580,7 +580,9 @@ NonTerminal SortID() : { NonTerminal nonTerminal; }
 | "context"       { matchedToken.kind = CONTEXT; } : CONTEXT_STATE
 
 // Note that "BUBBLE" must be last so it doesn't match keywords
-| <BUBBLE: (~[" ", "\t", "\r", "\n"])+>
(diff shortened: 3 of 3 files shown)


dwightguth (changes requested):
Needs a test. The linked issue contains one. You also need a test for a block comment in the middle of a bubble.

gtrepta:
Looks like the performance tests have the added benefit of catching problems in downstream semantics

https://github.com/runtimeverification/k/actions/runs/6029210633/job/16359659191?pr=3612#step:7:42

k-distribution/tests/regression-new/issue-3501-bubble-blockcomments/test.k
```diff
@@ -0,0 +1,14 @@
...
+    syntax Strings ::= List{String,","}
+endmodule
+
+module TEST
+    imports TEST-SYNTAX
+    rule/*1*/ "foo" => "bar"
+    rule/*2*/ "bar" => "baz"
+
+    rule "foo"/*1*/,"bar" => "baz"
+endmodule
```
radumereuta, line 14: This test can be added here: https://github.com/runtimeverification/k/blob/master/kernel/src/test/java/org/kframework/parser/outer/OuterParsingTests.java
with lower performance costs.
Or at most in the `/regression-new/checks` folder.

radumereuta (changes requested):

PR #3615, opened 2023-09-04
Fix token constructor error

### LINKED ISSUE #3460: [K-Bug] “Cannot define constructor: sort has domain values”

haskell-backend

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

5.6.90

### Operating System

Linux

```
module BUG-SYNTAX
  import ID-SYNTAX
endmodule

module BUG
  imports BUG-SYNTAX
  syntax Id ::= Foo(Id)
endmodule
```

### Steps to Reproduce

```
$ kompile --backend haskell bug.k
kore-exec: [39759] Error (ErrorVerify):
    Error:
      module 'BUG':
      symbol 'LblFoo'LParUndsRParUnds'BUG'Unds'Id'Unds'Id' declaration (definition.kore 93:10):
        Cannot define constructor LblFoo'LParUndsRParUnds'BUG'Unds'Id'Unds'Id for sort SortId{}: sort has domain values.
definition.
Examine output to see errors.
```

Note that this is probably a legitimate error, but the error message seems to be asking me to report a bug.

It might be that the error should have been caught earlier.

This does not occur in the LLVM backend. I don't know if that's expected.

### Expected Results

Either no error, or a proper error message instead of an internal error.

radumereuta:
Fixes #3460

Apparently we already had a check for this error, but it didn't take into consideration productions of the form:
`syntax Id [token]` since these are represented as `SyntaxSort` and not `Productions`.

This is a fix for the old issue: #2241

### DIFF
--- a/k-distribution/tests/regression-new/checks/tokenCheck.k
+++ b/k-distribution/tests/regression-new/checks/tokenCheck.k
@@ -1,5 +1,6 @@
 // Copyright (c) K Team. All Rights Reserved.
 module TOKENCHECK-SYNTAX
+  imports ID-SYNTAX
   syntax X ::= "abc" [token]
 
 endmodule
@@ -17,4 +18,5 @@ module TOKENCHECK
   syntax Int ::= "foo" [macro]
   rule foo => 3
 
+  syntax Id ::= Foo(Id)
 endmodule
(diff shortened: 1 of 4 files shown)


Robertorosmaninho:
Hey @radumereuta, these modifications made some tests fail. Could you check if the tests are intended to fail like that, or how we can rewrite them not to fail?

Robertorosmaninho (approved):
LGTM! Thanks @radumereuta !

PR #3627, opened 2023-09-07
Adding  CI to release workflow to create a `develop` baseline performance on Bencher

Robertorosmaninho:
This modification is necessary to create a baseline for newer branches to compare the performance of their changes against what we already have in develop.

This modification is necessary to create and populate the develop's branch as the baseline for newer branches. 
This script `post_results_to_develop.py` works by getting the last benchmarks reports of the last branch merged into develop, aggregating them in one file, and posting them on the develop branch of Bencher.

Hopefully, this is a temporary fix, as this was already accepted as a bug/possible feature by Bencher's maintainer https://github.com/bencherdev/bencher/issues/187. 

This PR, unlike https://github.com/runtimeverification/k/pull/3625, doesn't need to rerun all `Performance Tests` on the develop or master branch.


.github/workflows/develop.yml
```diff
@@ -37,3 +37,61 @@ jobs:
           if git add --update && git commit --no-edit --allow-empty --message "Set Version: $(cat package/version)"; then
             git push origin master
           fi
+
+  post-performance-tests:
```
F-WRunTime, line 41: I think the goal with performance to start is to start cataloging our progress, if we fail on this in anyway, command fails, connection to bencher causes an error we don't want this to stop our release process. 

We should consider adding `continue-on-error: true`

.github/workflows/develop.yml
```diff
@@ -37,3 +37,61 @@ jobs:
...
+          docker exec -t ci-${GITHUB_SHA} /bin/bash -c ' type -p curl >/dev/null || (apt update && apt install sudo curl -y) \
+            curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+            && sudo chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+            && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+            && sudo apt update \
+            && sudo apt install gh -y'
+      - name: 'Getting Performance Tests Results'
+        run: |
+          set -euxo pipefail
+          docker exec -t ci-${GITHUB_SHA} /bin/bash -c 'export GH_TOKEN="${GITHUB_TOKEN}"'
```
F-WRunTime: It looks like you're already setting the environment variable in the 'run' statement.

Robertorosmaninho: Yes! Thanks!!

radumereuta (commented):
Something is not clear for me.
Where do you store the results?
I would like to avoid modifying the current PR and/or adding new commits to develop after a PR was merged.
I don't think this is the case here, but I don't understand exactly how this works.

F-WRunTime (approved):

Robertorosmaninho:
> Something is not clear for me.
> Where do you store the results?

We're currently storing the results in the Bencher cloud! It's free, and we don't have to worry about keeping a docker container to use and maintain as a server.
 
> I would like to avoid modifying the current PR and/or adding new commits to develop after a PR was merged.
> I don't think this is the case here, but I don't understand exactly how this works.

Here, we aren't modifying the PR or adding something to the develop branch after the PR is merged. Instead, we're just:

1. querying the name of the last PR merged (using the last commit on develop) using GIT API and then querying the results of the last execution of `Performance Tests` on this PR using the Bencher API.
2. This result is then `POSTED` in the develop branch data graph on the Bencher DB using the Bencher API as well.

Doing so, we don't have to rerun the `Performance Tests` again to collect the results for the develop branch.

After this PR, develop will have its own graph on Bencher that will work as a baseline for other PRs.

radumereuta (approved):
Thank you @Robertorosmaninho that clarifies things for me.
lgtm

PR #3628, opened 2023-09-07
Fix gdate dependence: Use `date -r`  instead of `gdate -d` on macOS to get k version

### LINKED ISSUE #3610: Missing `gdate` as MacOS dependence

None

### Which command

- [ ] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v6.0.75

### Operating System

MacOS (Apple Silicon/AMD)

_No response_

### Steps to Reproduce

`gdate` is used in a few scripts to get the build version, for example, but it is mainly by the `llvm-backend`, as we can see if we grep it:
```
grep -Iir "gdate" .
./llvm-backend/target/build/install/lib/scripts/utils.sh:  DATE_CMD="gdate"
./llvm-backend/target/build/bin/utils.sh:  DATE_CMD="gdate"
./llvm-backend/src/main/native/llvm-backend/nix/llvm-backend.nix:      --replace 'gdate' 'date'
./llvm-backend/src/main/native/llvm-backend/bin/utils.sh:  DATE_CMD="gdate"
./k-distribution/src/main/scripts/lib/k:            versionDate=$(gdate -d "@${versionTimestampInMilliseconds}" "+%a %b %d %H:%M:%S %Z %Y")
```

### Expected Results

On README.md we should have:

"On macOS using [Homebrew](https://brew.sh/):

```shell
brew install bison boost cmake flex fmt gcc gdate gmp openjdk jemalloc libyaml llvm@15 make maven mpfr pkg-config python stack zlib z3
```

**Note**: in this case llvm@15 is required."

Robertorosmaninho:
Fix the frontend part of #3610 as `gdate` will not be a dependence for k version.
It still needs to be fixed on the LLVM Backend.


k-distribution/src/main/scripts/lib/k
```diff
@@ -13,7 +13,7 @@ for flag in "$@"; do
         versionTimestampInMilliseconds=$(echo "${versionTimestamp} / 1000" | bc)
 
         if [ "$(uname)" = 'Darwin' ]; then
-            versionDate=$(gdate -d "@${versionTimestampInMilliseconds}" "+%a %b %d %H:%M:%S %Z %Y")
+            versionDate=$(date -r "${versionTimestampInMilliseconds}" "+%a %b %d %H:%M:%S %Z %Y")
```
radumereuta, line 16: Interesting trick.

radumereuta (approved):
I think this looks fine, but can you explain why `date -r` works and why `date -d` doesn't?

Robertorosmaninho:
Yes, the problem is that `date` in Linux is from GNU, and in macOS, it is from FreeBSD. The API is different, and so its flags are different as well!

PR #3631, opened 2023-09-11
Add K-ST

wkyml:
We develop K-ST, a formal executable semantics for ST in the K framework.
Defined with respect to the IEC 61131-3 standard and PLC vendor manuals, K-ST is a high-level reference semantics that can be used to evaluate the correctness and consistency of different ST implementations.

Detailed information is in https://arxiv.org/abs/2202.04076.


web/pages/projects.md
```diff
@@ -83,3 +83,4 @@ let us know on our [socials](https://kframework.org/#support), and we will featu
 - [Elrond (2020-2021)](https://github.com/runtimeverification/elrond-semantics)
 - [P4 (2016-2021)](https://github.com/kframework/p4-semantics)
 - [K-Javalette (2022)](https://github.com/bbyalcinkaya/k-javalette)
+- [K-ST(2023)](https://github.com/wkyml/K-ST)
```
radumereuta:
```suggestion
- [K-ST (2023)](https://github.com/wkyml/K-ST)
```

radumereuta (changes requested):
Ah, I just noticed that you already opened a PR.
If you'll add the space there to keep everything consistent, I can approve it.

web/pages/projects.md
wkyml: We have made the necessary revisions. Thank you very much!

radumereuta (approved):

Issue #3634, opened 2023-09-11
kast: NoSuchElementException: key not found: bl_Map_


virgil-serbanuta:
This was originally produced by a kore-print command, which said
```
Please rerun your program with the --debug flag to generate a stack trace, and
file a bug report at https://github.com/runtimeverification/k/issues
```
so here it goes:

Versions:
```
$ kast --version
K version:    v6.0.87-0-g3997d65c22-dirty
Build date:   Lu sep 11 21:21:03 EEST 2023
```

To reproduce: [llvm.zip](https://github.com/runtimeverification/k/files/12580536/llvm.zip)
```
kast --input kore --definition /mnt/data/runtime-verification/elrond-wasm-real-elrond/.kbuild/kmxwasm-semantics/0a9b8db/target/v6.0.87-0-g3997d65c22-dirty/llvm --output json output.kore --debug
```

Stack trace:
```
java.util.NoSuchElementException: key not found: bl_Map_
        at scala.collection.MapLike.default(MapLike.scala:236)
        at scala.collection.MapLike.default$(MapLike.scala:235)
        at scala.collection.AbstractMap.default(Map.scala:65)
        at scala.collection.MapLike.apply(MapLike.scala:144)
        at scala.collection.MapLike.apply$(MapLike.scala:143)
        at scala.collection.AbstractMap.apply(Map.scala:65)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:308)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:302)
        at org.kframework.kore.KTransformer.apply(transformers.scala:11)
        at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
        at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
        at org.kframework.kore.TransformK.apply(TransformK.java:23)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:320)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:302)
        at org.kframework.kore.KTransformer.apply(transformers.scala:11)
        at org.kframework.kore.KTransformer.apply$(transformers.scala:10)
        at org.kframework.kore.AbstractKTransformer.apply(transformers.scala:123)
        at org.kframework.kore.TransformK.apply(TransformK.java:23)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:320)
        at org.kframework.unparser.KPrint$2.apply(KPrint.java:302)
[...]

PR #3636, opened 2023-09-12
Fix parsing left/right-assoc kore terms in kast

gtrepta:
fixes: #3634


gtrepta:
@radumereuta This is a very small fix, but would you like a test for this?

radumereuta:
Last time I had to modify that file, I added a test: #2766
Maybe you can modify an existing one to account for this case. That way we don't add to the CI time but we test for the problem.

gtrepta:
@radumereuta I found a good place for it. Ready for approval.

radumereuta (approved):

virgil-serbanuta:
Thanks!

PR #3637, opened 2023-09-13
Installing dependences to bencher execution on develop

Robertorosmaninho:
Follow up to #3627
This fix is necessary due to the lack of dependencies such as `python3`, `curl`,  and `bencher` on the test container: [error](https://github.com/runtimeverification/k/actions/runs/6173858635/job/16757140239). 

Running this container is lighter than using the `with-docker` version of CI, and we can pass the various environment variables needed by GitHub and Bencher API!

### DIFF
--- a/k-distribution/tests/profiling/setup_profiling.sh
+++ b/k-distribution/tests/profiling/setup_profiling.sh
@@ -6,8 +6,11 @@ export DEBIAN_FRONTEND=noninteractive
 export BENCHER_VERSION="0.3.10"
 apt-get update
 apt-get upgrade --yes
-apt-get install --yes make time sudo wget
-apt-get install --yes ./kframework.deb
+apt-get install --yes python3 git curl make time sudo wget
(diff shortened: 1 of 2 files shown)


k-distribution/tests/profiling/setup_profiling.sh
```diff
@@ -6,8 +6,11 @@ export DEBIAN_FRONTEND=noninteractive
 export BENCHER_VERSION="0.3.10"
 apt-get update
 apt-get upgrade --yes
```
F-WRunTime, line 8: Missed this last time but you may want to be warry of blind upgrades, they could have unintended changes for tools / builds. Just a note.

F-WRunTime (approved):

k-distribution/tests/profiling/setup_profiling.sh
Robertorosmaninho: Should I remove them?

PR #3638, opened 2023-09-13
Improving  kompile's `--help` and `--help-hidden` messages using descriptors

### LINKED ISSUE #3619: [K-Bug] Missing file argument in kompile's flag

Front-End

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v6.0.83-0-gc2ad66e022

### Operating System

MacOS (Apple Silicon/AMD)

### Steps to Reproduce

When I tried to use `--profile-rule-parsing` as it's defined on our hidden-help flag

```
$ kompile -hh
...
 --profile-rule-parsing
      Generate time in milliseconds to parse each rule in the semantics.
```

I got the following confusing error message:
```
$ kompile -v --profile-rule-parsing test.k --main-module TEST

Parse command line options                                   =  0,014s
outer parsing.
```

After talking with @radumereuta at https://github.com/runtimeverification/k/pull/3611#discussion_r1312889719 I saw the this flag actually expects a new file to write the time taken to parse each rule.

### Expected Results

I would expect 3 things here:
- [x] Check if any other flag expects a file or other type of input that isn’t currently described on `-hh` or `-h`.
- [x] Modify the help message to be `--profile-rule-parsing <file>` and probably contain a more helpful description like `Store in this file time taken in ms to parse each rule in the semantics` as suggested by @radumereuta.
- [x] If possible, modify the error message to be different when a flag doesn’t receive input. 

The last is *optional*, but it’s something to consider. Currently `kompile` works as `Usage: kompile [options] <file>`, in this case, the `--profile-rule-parsing` consumed the K file instead of `kompile`. This could be avoided if we swap the order to `kompile <file> [options]`, and then it will be easier to identify when we don’t pass input to kompile and when we don’t pass input for a kompile’s flag. Although, I admit that’s counterintuitive considering other compiler’s CLI design.

PS.: Also from @radumereuta:
"The output for the help command is handled automatically by the`jcommnader` library here: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kompile/KompileUsageFormatter.java"

Robertorosmaninho:
Fixes #3619

This PR enhances the usability of `kompile --help` and `kompile --help-hidden` by providing descriptors to the flags to help the user understand when an argument needs to be provided to the flag.

Ex.:

Before
``` 
kompile --help
Usage: kompile [options] <file>
  Options:
    --backend
      Choose a backend. <backend> is one of [llvm|haskell|kore]. Each creates 
      the kompiled K definition.
      Default: llvm

```

Now:
```
kompile --help
Usage: kompile [options] <file>
  Options:
    --backend <backend>
      Choose a backend. <backend> is one of [llvm|haskell|kore]. Each creates 
      the kompiled K definition.
      Default: llvm
```

### DIFF
--- a/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
+++ b/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
@@ -43,6 +43,6 @@ public class LLVMKompileOptions {
     @Parameter(names="--llvm-kompile-type", description="Specifies the llvm backend's output type. Valid choices are main (build an interpreter), library (build an interpreter, but don't link a main function), search (same as main, but the interpreter does search instead of single-path execution), static (same as library, but no '-l' flags are passed during linking. Used for making a partially linked object file) and python (build a Python bindings module for this definition)", descriptionKey = "type", hidden = true)
     public String llvmKompileType = "main";
 
-    @Parameter(names="--llvm-kompile-output", description="Name of the output binary from the llvm backend.", descriptionKey = "binary", hidden = true)
+    @Parameter(names="--llvm-kompile-output", description="Name of the output binary from the llvm backend.", descriptionKey = "file", hidden = true)
     public String llvmKompileOutput = null;
 }
--- a/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
+++ b/llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
@@ -15,7 +15,7 @@ public class LLVMKompileOptions {
     @Parameter(names="--enable-llvm-debug", description="Enable debugging support for the LLVM backend.")
     public boolean debug = false;
 
-    @Parameter(names="-ccopt", description="Add a command line option to the compiler invocation for the llvm backend.", listConverter=SingletonListConverter.class, hidden = true)
+    @Parameter(names="-ccopt", description="Add a command line option to the compiler invocation for the llvm backend.", descriptionKey = "options", listConverter=SingletonListConverter.class, hidden = true)
     public List<String> ccopts = new ArrayList<>();
 
     public static class SingletonListConverter implements IStringConverter<List<String>> {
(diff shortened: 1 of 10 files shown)


radumereuta (approved):
Looking nice!

radumereuta (commented):
Ah, there are more options for kprove and kast. Did you check those as well?

llvm-backend/src/main/java/org/kframework/backend/llvm/LLVMKompileOptions.java
```diff
@@ -40,9 +40,9 @@ public List<String> convert(String str) {
...
+    @Parameter(names="--llvm-kompile-type", description="Specifies the llvm backend's output type. Valid choices are main (build an interpreter), library (build an interpreter, but don't link a main function), search (same as main, but the interpreter does search instead of single-path execution), static (same as library, but no '-l' flags are passed during linking. Used for making a partially linked object file) and python (build a Python bindings module for this definition)", descriptionKey = "type", hidden = true)
     public String llvmKompileType = "main";
 
-    @Parameter(names="--llvm-kompile-output", description="Name of the output binary from the llvm backend.", hidden = true)
+    @Parameter(names="--llvm-kompile-output", description="Name of the output binary from the llvm backend.", descriptionKey = "binary", hidden = true)
```
radumereuta: That's a file, right?

Robertorosmaninho: I think so, but as we're already using `executable` or `command` for other binaries I think that's on to use `binary` as well.

radumereuta: And it makes sense when it's input. But this is output. You don't really care about what comes out. Unless it's a directory or a file.
I'm just nitpicking here.

Robertorosmaninho: Got it! I'll modify it!
Thanks!

Robertorosmaninho:
I'll implement these changes in `kast` and `kprove` in future PRs!

PR #3642, opened 2023-09-14
Adding arguments to `kast` flags

Robertorosmaninho:
Follow up of #3619 and #3638!

This PR enhances the usability of `kast --help` and `kast --help-hidden` by providing flag descriptors to help the user understand when an argument needs to be provided to the flag.

Ex.:

Before
``` 
kast --help
Usage: kast [options] <file>
  Options:
    --color
      Use colors in output. Default is on.
      Possible Values: [OFF, ON, EXTENDED]
```

Now:
```
Usage: kast [options] <file>
  Options:
    --color <mode>
      Use colors in output.
      Default: ON
      Possible Values: [OFF, ON, EXTENDED]
```

Among others, this PR introduces the following arguments to the respective parameters:
<!DOCTYPE html>

parameter | argument
-- | --
--color |  \<mode>
--expression, -e | \<expression>
--input, -i | \<mode>
--module, -m | \<module>
--output, -o | \<mode>
--output-file | \<file>
--output-flatten | \<KLabels>
--output-omit | \<KLabels>
--output-tokast | \<KLabels>
--output-tokenize | \<KLabels>
--sort, -s | \<sort>
--steps | \<steps>
--definition | \<path>
--directory, -d | \<path>

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/PrintOptions.java
+++ b/kernel/src/main/java/org/kframework/unparser/PrintOptions.java
@@ -31,9 +31,9 @@ public class PrintOptions {
     public PrintOptions(Void v) {
     }
 
-    @Parameter(names = "--color", description = "Use colors in output.", descriptionKey = "mode",
+    @Parameter(names = "--color", description = "Use colors in output. Default is on.", descriptionKey = "mode",
             converter=ColorModeConverter.class)
-    private ColorSetting color = ColorSetting.ON;
+    private ColorSetting color;
 
     public ColorSetting color(boolean ttyStdout, Map<String, String> env) {
         boolean colorize = outputFile == null && ttyStdout;
--- a/kernel/src/main/java/org/kframework/unparser/PrintOptions.java
+++ b/kernel/src/main/java/org/kframework/unparser/PrintOptions.java
@@ -31,8 +31,9 @@ public class PrintOptions {
     public PrintOptions(Void v) {
     }
 
-    @Parameter(names = "--color", description = "Use colors in output. Default is on.", converter=ColorModeConverter.class)
-    private ColorSetting color;
+    @Parameter(names = "--color", description = "Use colors in output.", descriptionKey = "mode",
+            converter=ColorModeConverter.class)
+    private ColorSetting color = ColorSetting.ON;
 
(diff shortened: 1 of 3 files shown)


radumereuta (approved):
PR lgtm but there are some failing tests

kernel/src/main/java/org/kframework/unparser/PrintOptions.java
```diff
@@ -31,8 +31,9 @@ public PrintOptions(OutputModes output) {
     public PrintOptions(Void v) {
     }
 
-    @Parameter(names = "--color", description = "Use colors in output. Default is on.", converter=ColorModeConverter.class)
-    private ColorSetting color;
+    @Parameter(names = "--color", description = "Use colors in output.", descriptionKey = "mode",
+            converter=ColorModeConverter.class)
+    private ColorSetting color = ColorSetting.ON;
```
Robertorosmaninho: That's funny, it seems that the error is from this modification. The help message says that the default is color=ON, but when I really implemented it, it raised the error on [max-fail-spec.k](https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/imp-haskell/max-fail-spec.k)

Robertorosmaninho: Notice that the output also changed depending on `--color ON` and `--color OFF`:

With color:
```
  <T>
    <k>
      .
    </k>
    <state>
      a |-> A:Int
      b |-> B:Int
      max |-> B:Int
    </state>
  </T>
#And
  #Not ( {
      A
    #Equals
      B
    }
  #And
    {
      true
    #Equals
      B >=Int B
    } )
#And
  {
    true
  #Equals
    A <=Int B
  }
```

Without color:
```
  #Not ( {
      A
    #Equals
      B
    }
  #And
    {
      true
    #Equals
      B >=Int B
    } )
#And
  <T>
    <k>
      .
    </k>
    <state>
      a |-> A:Int
      b |-> B:Int
      max |-> B:Int
    </state>
  </T>
#And
  {
    true
  #Equals
    A
```

Robertorosmaninho: I've just filled an issue to track it: https://github.com/runtimeverification/k/issues/3643

PR #3644, opened 2023-09-14
Updating `Guice` to 4.0 and deleting temporary injections

Robertorosmaninho:
This modification aimed to update the [Guice](https://github.com/google/guice) version from `4.0-beta5` to `4.0`, deleting the constructors with one void parameter to
```
Prevents instantiation by Guice when not explicitly configured in a Module.
```
And applying the `requireAtInjectOnConstructors` on every `constructor()` inside a class that extends the `AbstractModule` for consistency. These modifications made it necessary to explicitly create constructors with zero parameters and within the `@Injector` annotation.


dwightguth (changes requested):
There's some context involved here. Guice 4.0 includes an option to disable injection of zero argument constructors implicitly. The plan was to enable this option and then explicitly inject everything that was previously being injected implicitly. You probably don't want to delete those constructors unless you do this, otherwise you might accidentally inject the wrong values for options if your guice configuration has a bug, rather than reporting an error. I wholeheartedly agree with updating to the release of guice though.

Robertorosmaninho:
> There's some context involved here. Guice 4.0 includes an option to disable injection of zero argument constructors implicitly. The plan was to enable this option and then explicitly inject everything that was previously being injected implicitly. You probably don't want to delete those constructors unless you do this, otherwise you might accidentally inject the wrong values for options if your guice configuration has a bug, rather than reporting an error. I wholeheartedly agree with updating to the release of guice though.

Hey @dwightguth, to make sure that we're on the same page here. Do you mean the [requireAtInjectOnConstructors](https://google.github.io/guice/api-docs/4.0/javadoc/com/google/inject/Binder.html#requireAtInjectOnConstructors--), right?

Robertorosmaninho:
Hey @dwightguth, I still have some `configure()` to adapt, but I would like feedback here just to check that I'm on the correct path. Could you please take a look at it?

dwightguth (commented):
This looks great assuming you can get it to work with every class extending AbstractModule being modified.

Robertorosmaninho:
Hey @dwightguth, all the modules that extend the `AbstractModule` were modified to work with this. I made the modifications after my comment!

dwightguth (approved):

PR #3648, opened 2023-09-15
Adding arguments to `kprove` flags

Robertorosmaninho:
Follow up of #3619,  #3638, and #3642!

This PR enhances the usability of `kprove --help` and `kprove --help-hidden` by providing flag descriptors to help the user understand when an argument needs to be provided to the flag.

Ex.:

Before
``` 
Usage: kprove [options] <file>
  Options:
    --branching-allowed
      Number of branching events allowed before a forcible stop.
      Default: 2147483647
```

Now:
```
Usage: kprove [options] <file>
  Options:
    --branching-allowed <value>
      Number of branching events allowed before a forcible stop.
      Default: 2147483647
```

Among others, this PR introduces the following arguments to the respective parameters:

parameter | argument
-- | --
--branching-allowed | \<number>
--claims | \<labels>
--color | \<mode>
--debug-script | \<file>
--default-claim-type | \<type>
--definition | \<path>
--depth | \<number>
--emit-json-spec | \<file>
--exclude | \<labels>
--output, -o | \<mode>
--output-file | \<file>
--output-flatten | \<KLabels>
--output-omit | \<KLabels>
--output-tokast | \<KLabels>
--output-tokenize | \<KLabels>
--spec-module, -sm |  \<name>
--trusted | \<labels>
--haskell-backend-command | \<command>
--haskell-backend-home | \<directory>
--smt-timeout | \<milliseconds>

### DIFF
--- a/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
+++ b/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
@@ -55,19 +55,19 @@ public class KProveOptions {
     @ParametersDelegate
     public PrintOptions print = new PrintOptions();
 
-    @Parameter(names="--branching-allowed", descriptionKey = "value", arity = 1,
+    @Parameter(names="--branching-allowed", descriptionKey = "number", arity = 1,
             description="Number of branching events allowed before a forcible stop.")
     public int branchingAllowed = Integer.MAX_VALUE;
 
-    @Parameter(names={"--spec-module", "-sm"}, descriptionKey = "module",
+    @Parameter(names={"--spec-module", "-sm"}, descriptionKey = "name",
             description="Name of module containing specification to prove")
     public String specModule;
 
-    @Parameter(names="--depth", descriptionKey = "value",
+    @Parameter(names="--depth", descriptionKey = "number",
             description="The maximum number of computational steps to prove")
     public Integer depth;
 
-    @Parameter(names="--trusted", descriptionKey = "claims",
+    @Parameter(names="--trusted", descriptionKey = "labels",
             description="Mark this comma separated list of claims as [trusted]")
     public List<String> trusted = null;
 
--- a/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
+++ b/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
@@ -71,10 +71,10 @@ public class KProveOptions {
             description="Mark this comma separated list of claims as [trusted]")
     public List<String> trusted = null;
 
-    @Parameter(names="--exclude", descriptionKey = "claims", description="Exclude this comma separated list of claims")
+    @Parameter(names="--exclude", descriptionKey = "labels", description="Exclude this comma separated list of claims")
     public List<String> exclude = null;
 
-    @Parameter(names="--claims", descriptionKey = "claims", description="Only keep this comma separated list of claims")
+    @Parameter(names="--claims", descriptionKey = "labels", description="Only keep this comma separated list of claims")
     public List<String> claims = null;
 
     @Parameter(names="--debugger",
--- a/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
+++ b/kernel/src/main/java/org/kframework/kprove/KProveOptions.java
@@ -55,33 +55,40 @@ public class KProveOptions {
     @ParametersDelegate
     public PrintOptions print = new PrintOptions();
 
-    @Parameter(names="--branching-allowed", arity = 1, description="Number of branching events allowed before a forcible stop.")
+    @Parameter(names="--branching-allowed", descriptionKey = "value", arity = 1,
+            description="Number of branching events allowed before a forcible stop.")
     public int branchingAllowed = Integer.MAX_VALUE;
 
-    @Parameter(names={"--spec-module", "-sm"}, description="Name of module containing specification to prove")
+    @Parameter(names={"--spec-module", "-sm"}, descriptionKey = "module",
+            description="Name of module containing specification to prove")
     public String specModule;
 
-    @Parameter(names="--depth", description="The maximum number of computational steps to prove")
+    @Parameter(names="--depth", descriptionKey = "value",
+            description="The maximum number of computational steps to prove")
     public Integer depth;
 
-    @Parameter(names="--trusted", description="Mark this comma separated list of claims as [trusted]")
+    @Parameter(names="--trusted", descriptionKey = "claims",
+            description="Mark this comma separated list of claims as [trusted]")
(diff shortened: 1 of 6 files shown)


kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,33 +55,40 @@ public synchronized File specFile(FileUtil files) {
     @ParametersDelegate
     public PrintOptions print = new PrintOptions();
 
-    @Parameter(names="--branching-allowed", arity = 1, description="Number of branching events allowed before a forcible stop.")
+    @Parameter(names="--branching-allowed", descriptionKey = "value", arity = 1,
+            description="Number of branching events allowed before a forcible stop.")
```
radumereuta, line 59: Maybe "number" would be better instead of "value".

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,33 +55,40 @@ public synchronized File specFile(FileUtil files) {
...
     public int branchingAllowed = Integer.MAX_VALUE;
 
-    @Parameter(names={"--spec-module", "-sm"}, description="Name of module containing specification to prove")
+    @Parameter(names={"--spec-module", "-sm"}, descriptionKey = "module",
+            description="Name of module containing specification to prove")
     public String specModule;
 
-    @Parameter(names="--depth", description="The maximum number of computational steps to prove")
+    @Parameter(names="--depth", descriptionKey = "value",
+            description="The maximum number of computational steps to prove")
```
radumereuta, line 67: Same. "number" or "int" would be clearer in this case.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,33 +55,40 @@ public synchronized File specFile(FileUtil files) {
...
     public Integer depth;
 
-    @Parameter(names="--trusted", description="Mark this comma separated list of claims as [trusted]")
+    @Parameter(names="--trusted", descriptionKey = "claims",
+            description="Mark this comma separated list of claims as [trusted]")
```
radumereuta, line 71: Maybe "labels" would work better here?
Because that's what you expect, a list of tags or labels.
And I'm wondering the same thing for `--spec-module` if it should ask for a "name" instead of "module".

kernel/src/main/java/org/kframework/utils/options/SMTOptions.java
```diff
@@ -29,25 +30,30 @@ public Class<SMTSolver> enumClass() {
...
     @Parameter(names="--maps-as-int-array", description="Abstracts map values as an array of ints.", hidden = true)
     public boolean mapAsIntArray = false;
 
-    @Parameter(names={"--smt-prelude", "--smt_prelude"}, description="Path to the SMT prelude file.", descriptionKey = "path", hidden = true)
+    @Parameter(names={"--smt-prelude", "--smt_prelude"},
+            description="Path to the SMT prelude file.", descriptionKey = "path", hidden = true)
     public String smtPrelude;
 
-    @Parameter(names="--smt-timeout", description="Timeout for calls to the SMT solver, in milliseconds.", hidden = true)
+    @Parameter(names="--smt-timeout", descriptionKey = "millisecond",
```
radumereuta:
```suggestion
    @Parameter(names="--smt-timeout", descriptionKey = "milliseconds",
```

radumereuta (commented):
A few suggestions.
Ping me again when you answered all of them and tests passed.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
Robertorosmaninho: Okok

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
Robertorosmaninho: Number seems better than int

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
Robertorosmaninho: `--claims` and `--exclude` has a similar description, what about those?
Also, the `--spec-module` is following the `--module` of kast and kompile, even though I liked the suggestion. Should we break the pattern here?

kernel/src/main/java/org/kframework/utils/options/SMTOptions.java
Robertorosmaninho: Thanks

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
radumereuta: We should make all of them the same.
That is if you agree with the naming I'm proposing.
Then the option would read:
`--module <name>` and `--spec-module <name>`

Robertorosmaninho:
@radumereuta, I've implemented your comments! However, I would appreciate a new round on the comment I didn't mark resolved!

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
Robertorosmaninho: Alright! Let's make the change 😄

Robertorosmaninho: Same for `--claims` and `--exclude`?

radumereuta: Yes. `--claims`, `-exclude` and `--trusted` are in the same logical group.

kernel/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -55,33 +55,40 @@ public synchronized File specFile(FileUtil files) {
     @ParametersDelegate
     public PrintOptions print = new PrintOptions();
 
-    @Parameter(names="--branching-allowed", arity = 1, description="Number of branching events allowed before a forcible stop.")
+    @Parameter(names="--branching-allowed", descriptionKey = "number", arity = 1,
+            description="Number of branching events allowed before a forcible stop.")
     public int branchingAllowed = Integer.MAX_VALUE;
 
-    @Parameter(names={"--spec-module", "-sm"}, description="Name of module containing specification to prove")
+    @Parameter(names={"--spec-module", "-sm"}, descriptionKey = "name",
```
radumereuta, line 62: Now that you made it `name`, you should change the other instances for kompile and kast as well.

radumereuta (approved):
lgtm

PR #3650, opened 2023-09-18
Finished page title generation

### LINKED ISSUE #3359: Documentation pages should have unique page titles
Overall, the K Framework documentation is very helpful, and it appears that more tutorials are being added soon which is great to see. 😄

However, I do find it a bit hard to navigate at the moment because all of the pages on the https://kframework.org site use a `title` tag containing "K | Runtime Verification Inc". If you open several pages in different tabs, it becomes impossible to tell which page is which since they all have identical page titles.

Most (all?) pages do have a reasonable first header that could be used as the page title. I suspect there's some metadata and / or option that could be used with your page generator to produce better page titles for each page.

KcatMa:
Fixed #3359 

The live pages deployed on: https://kcatma.github.io/k


yiyi-wang-rv:
Hi @radumereuta , this PR is ready but I am not sure why the CI failed. Could you please help take a look when you have time? Thank you!

yiyi-wang-rv (approved):

PR #3659, opened 2023-09-22
Add llvm-kompile-matching to K package

### LINKED ISSUE #3640: llvm-backend-matching fails to parse axiom with `\right-assoc`
Using K framework 6.0.94 (but also 6.0.92 and the llvm-backend master at the time of writing), I encounter the following error when trying to generate the jump table (`./dt` directory) with `llvm-backend-matching`:

```
$ llvm-backend-matching ./definition.kore qbaL ./fubar 0
Exception in thread "main" org.kframework.parser.kore.parser.ParseError: ERROR: Line 13256: Column 13: Expected \top,\bottom,\and,\or,\implies,\iff,\exists,\forall,\ceil,\floor,\equals,\mem, but '\ri'
  axiom{} \right-assoc{}(\or{SortTxGasPriceCell{}} (\exists{SortTxGasPriceCell{}} (X0:SortInt{}, Lbl'-LT-'txGasPrice'-GT-'{}(X0:SortInt{})), \bottom{SortTxGasPriceCell{}}())) [constructor{}()] // no junk
            ^
	at org.kframework.parser.kore.parser.TextToKore.error(TextToKore.scala:774)
	at org.kframework.parser.kore.parser.TextToKore.parsePattern(TextToKore.scala:529)
	at org.kframework.parser.kore.parser.TextToKore.parseDeclarations(TextToKore.scala:284)
	at org.kframework.parser.kore.parser.TextToKore.parseModule(TextToKore.scala:161)
	at org.kframework.parser.kore.parser.TextToKore.parseModules(TextToKore.scala:173)
	at org.kframework.parser.kore.parser.TextToKore.parseDefinition(TextToKore.scala:135)
	at org.kframework.parser.kore.parser.TextToKore.parse(TextToKore.scala:72)
	at org.kframework.parser.kore.parser.TextToKore.parse(TextToKore.scala:52)
	at org.kframework.backend.llvm.matching.Matching$.writeDecisionTreeToFile(Matching.scala:14)
	at org.kframework.backend.llvm.matching.Matching$.main(Matching.scala:106)
	at org.kframework.backend.llvm.matching.Matching.main(Matching.scala)
```
The `definition.kore` file in this run was produced by `kompile` of the same version of K, in a command `kompile --llvm-kompile-type c ...` compiling an evm-semantics proof module. I then tested this with a stand-alone K definition (the well-known `imp.k`) and got the same result.

To reproduce:

```
$ kup install k.openssl.procps
$ kompile --llvm-kompile-type c some-definition.k
$ llvm-backend-matching some-definition-kompiled/definition.kore qbal ./fubar 0
```

What puzzles me is that the line numbers in the stack trace above do not match the code on master (or on the tagged version 6.0.94) at all. The code on master has [a `TextToKore::parsePattern()` with a case for `\right-assoc`](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/parser/kore/parser/TextToKore.scala#L560-L561), and [the error call is on line 621, not 774](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/parser/kore/parser/TextToKore.scala#L621).

gtrepta:
Should help with #3640


ehildenb:
Doesn't this script already live in the LLVM backend repo? Are we able to add it to the K assembly file instead, so that it still only lives in one place?

ehildenb:
I believe this is the assembly file specifying what gets included in teh K package: https://github.com/runtimeverification/k/blob/master/k-distribution/src/main/assembly/bin.xml

gtrepta:
> Doesn't this script already live in the LLVM backend repo? Are we able to add it to the K assembly file instead, so that it still only lives in one place?

It doesn't. There's a nix flake in the llvm-backend that adds an `llvm-backend-matching` to the PATH here:

https://github.com/runtimeverification/llvm-backend/blob/c64bbbb9ba6b141c183bbb0094ad8385683ad525/nix/llvm-backend-matching.nix#L20-L21

But the problem with this and with adding it in the llvm-backend in general is that it wont have the jars from the K package in its class path.

radumereuta (approved):

gtrepta:
@ehildenb I moved the scripts to the `llvm-backend` subfolder at least. I think a good time to clean up the nix files would be when this gets downstream to the backend booster team and they're using it.

ehildenb (approved):

PR #3660, opened 2023-09-25
kast --debug-tokens

### LINKED ISSUE #3647: Add kast option to show the list of tokens after scanner
It's hard to debug why things don't parse.
It would be useful to have an option for `kast` to show you the list of the tokens after parsing.
I think this can be done in ParsingInModule. Instead of calling the parser, just call the scanner, pretty print and exit.
The output should include:
- matched token
- location
- production used
Probably json output would be best for this. Although a simple list would work as well.
The production should contain the location information as well.

This would have been useful when debugging the mir semantics a few days ago.
@dkcumming do you have more suggestions that would help you with debugging?

radumereuta:
Add an option to kast to show the list of matched tokens, their location and the token used to match it.
Review with hidden whitespace.
@dkcumming how does this look?


k-distribution/tests/regression-new/issue-3647-debugTokens/a.test.kast.out
```diff
@@ -0,0 +1,8 @@
+`Match`    (location), Regex Terminal
+`a`        (1,1,1,2), r"[a-z][a-zA-Z0-9]*"
```
dwightguth: it looks like line 1 is intended to be a table header, but you don't have a line with something like `----------------` to make that clear. You also don't appear to be completely lining up the third column of the table, and I would remove the word "Regex" from the table header since not all of the values in that column are actually regexes.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -159,6 +162,26 @@ public void initialize() {
         }
     }
 
+    /**
+     * Print the list of tokens matched by the scanner, the location and the Regex Terminal
+     */
+    public String tokenizeString(String input, Source source) {
+        StringBuilder sb = new StringBuilder("`Match`    (location), Regex Terminal\n");
```
dwightguth: it doesn't look like this table will line up in the first column either if the token matched is longer than 10 characters.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -583,11 +583,11 @@ public ParserMetadata(String input, Scanner scanner, Source source, int startLin
     }
 
     // the list of tokens in the sentence
-    Scanner.Token[] words;
+    public Scanner.Token[] words;
```
dwightguth: this should not be public because arrays are mutable. Same with the two arrays below. You need a getter that returns a Collections.unmodifiableList

kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java
```diff
@@ -50,7 +50,7 @@
  */
 public class Scanner implements AutoCloseable {
 
-    private final Map<TerminalLike, Tuple2<Integer, Integer>> tokens;
+    public final Map<TerminalLike, Tuple2<Integer, Integer>> tokens;
```
dwightguth: you need a getter here too that wraps with unmodifiableMap because tokens is a mutable Map.

dwightguth (changes requested):

dkcumming:
Wow wonderful. I build this and tested it on `kmir` and it is already looking really helpful from the first test. This feedback of terminals is great to orient where the lexer is! I wonder if some idea of a trace of non-terminals seen on the way to the matched terminal would be possible in the future - this would provide a complete picture. However that isn't a criticism just a thought - this is great and will be very helpful :)

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
radumereuta: True. If the token is longer than 10 chars, it's not going to align.
This is a best-effort display. Tokens can be hundreds of chars long and can contain new line.
Not worth trying to align everything.

Robertorosmaninho (approved):
LGTM! Just a quick question: the size of the `Match` is limited (in this beautiful table format) to 10 characters (+-), or the size of the table will be adjusted by the most extended token?

dwightguth:
The getters are still janky and we still need to format the table better. I know you don't seem to view formatting as important, but a token longer than ten characters is far from unheard of and it will seriously impact the readability of this feature to be aligned incorrectly. The same problem will occur with extremely large files in the location field. It's not hard to just compute the longest width needed and format appropriately, and it's worth it because it will have a significant impact on readability. We can cap the total width at 80 characters and make the width of the token field smaller if needed to avoid going over that.

By the way, the header of the third column is misaligned.

The methods you want for the getters of the `int[]` are `Ints.asList` and `Collections.unmodifiableList`

radumereuta:
@dwightguth please have another look.
I've fixed the getter issues. `List.of(words);` already uses unmodifiable lists underneath.

I've also reordered the columns. I made the location the first column since it doesn't vary in size that much.
The matched tokens and the Terminal though, can vary greatly. I don't think it's worth trying to do smarter things here.
This output is intended to be used in cmd line if small, if not dump the output in a file and navigate w/o wrap lines.

I can try and engineer something _perfect_. But I don't think it's worth the effort. I'm also trying to keep the code simple.
Please have a look at the updated example.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -587,11 +588,11 @@ public List<Scanner.Token> getWords() {
     }
 
     public List<Integer> getLines() {
-      return Arrays.stream(lines).boxed().collect(Collectors.toList());
+      return Ints.asList(lines);
```
dwightguth: this needs to be wrapped in Collections.unmodifiableList because Ints.asList returns a mutable list.

dwightguth:
We have already spent more time on this back and forth than it would have taken to just compute the max length of each column and do some basic arithmetic. This is now actually worse than it was before because it's much /less/ readable with the location field first. I agree this is not something that makes or breaks this feature, but we are going to end up using this again in the future and having seriously misaligned columns and difficult to read tables is going to have a significant impact on readability. I genuinely don't understand the resistance to something that should be at most fifteen minutes of work.

As a side note, see my comment on the getters, which are still not right.

k-distribution/tests/regression-new/issue-3647-debugTokens/a.test.kast.out
```diff
@@ -0,0 +1,15 @@
...
+(1,5,1,6),    "2"          r"[\\+-]?[0-9]+"
+(1,7,1,8),    "+"          "+"
+(1,9,1,21),   "aaaaaaaaaaaa" r"[a-z][a-zA-Z0-9]*"
+(12,1,12,2),  "+"          "+"
+(12,3,12,11), "10000000"   r"[\\+-]?[0-9]+"
+(13,1,13,2),  "+"          "+"
+(13,3,13,8),  "\"str\""    r"[\\\"](([^\\\"\\n\\r\\\\])|([\\\\][nrtf\\\"\\\\])|([\\\\][x][0-9a-fA-F]{2})|([\\\\][u][0-9a-fA-F]{4})|([\\\\][U][0-9a-fA-F]{8}))*[\\\"]"
+(14,1,14,2),  "+"          "+"
+(14,3,14,36), "\"long str that breaks alighnment\"" r"[\\\"](([^\\\"\\n\\r\\\\])|([\\\\][nrtf\\\"\\\\])|([\\\\][x][0-9a-fA-F]{2})|([\\\\][u][0-9a-fA-F]{4})|([\\\\][U][0-9a-fA-F]{8}))*[\\\"]"
+(15,1,15,1),  ""           "<eof>"
```
ehildenb: Can we output this as a valid markdown table?

ehildenb: https://www.markdownguide.org/extended-syntax/

image

dwightguth (approved):

PR #3662, opened 2023-09-25
Check if macros were expanded in `ensures` and `requires`

### LINKED ISSUE #3585: [K-Bug] `kompile` throw a raw llvm error message when a rule in a `requires` is not defined

llvm-backend

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v6.0.58

### Operating System

Linux

```k
module SIGNED-INT
    imports INT
    imports BOOL

    syntax Int ::= "minSInt8"       [macro]
                 | "maxSInt8"       [alias]
                 | "minSInt16"      [macro]
                 | "maxSInt16"      [alias]

    syntax TypedArg ::= #int8  ( Int )   [symbol]
                      | #int16 ( Int )   [symbol]

    syntax Int ::= #getValue ( TypedArg ) [function]

    syntax Bool ::= #rangeSInt ( Int, Int ) [alias]

    rule #int8  ( X ) => X
    rule #int16 ( X ) => X

    rule minSInt8  => 0
    rule maxSInt8  => 1
    rule minSInt16 => 2
    rule maxSInt16 => 3

    rule #rangeSInt ( 8 , X ) => minSInt8  <=Int X andBool X  <Int maxSInt8 
    // rule #rangeSInt ( 16, X ) => minSInt16 <=Int X andBool X  <Int maxSInt16 // NOTE I should be defined !!

    rule #getValue(   #int8 ( X )) => X       requires #rangeSInt(8 ,   X)
    rule #getValue(   #int16( X )) => X       requires #rangeSInt(16,   X)
endmodule
```

### Input

`kompile sint.k --main-module SIGNED-INT --backend llvm`

### Output

```sh
Function return type does not match operand type of return inst!
  ret %block* %4
 i1llc: error: 'tmp.OdhC0jH8Xr': input module cannot be verified
[Error] Critical: llvm-kompile returned nonzero exit code: 1
Examine output to see errors.
[Warning] Compiler: Could not find main syntax module with name
SIGNED-INT-SYNTAX in definition.  Use --syntax-module to specify one. Using
SIGNED-INT as default.
```

### Input

`cd sint-kompiled`
`llvm-kompile-codegen definition.kore dt/dt.yaml dt --no-optimize -o definition.ll`
`opt -verify definition.ll -o /dev/null`

### Output

```sh
opt: definition.ll:540:7: error: value doesn't match function result type 'i1'
  ret %block* %4
      ^
```

### Expected Results

A more helpful error message, showing that the rule that is in the `requires` clause called `#rangeSInt(16,   X)` should be defined.

Robertorosmaninho:
Fixes #3585

This is a simple PR that extends the verification of expanded macros from checking just the rule `body` also to check its `ensures` and `requires`.

### DIFF
--- a/k-distribution/tests/regression-new/checks/expandMacroEnsures.k
+++ b/k-distribution/tests/regression-new/checks/expandMacroEnsures.k
@@ -1,3 +1,5 @@
+// Copyright (c) K Team. All Rights Reserved.
+
 module EXPANDMACROENSURES
     imports INT
     imports BOOL
--- a/k-distribution/tests/regression-new/checks/expandMacroEnsures.k
+++ b/k-distribution/tests/regression-new/checks/expandMacroEnsures.k
@@ -1,0 +1,11 @@
+module EXPANDMACROENSURES
+    imports INT
+    imports BOOL
+
+    syntax Int  ::= #foo ( Int ) [function]
+    syntax Bool ::= #baz ( Int ) [alias]
+
+//    rule #baz ( X ) => X // NOTE I should be defined !!
+
+    rule #foo( X ) => X       ensures  #baz( X )
+endmodule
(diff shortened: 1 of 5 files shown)


dwightguth (changes requested):
Needs a test, but otherwise looks fine.

dwightguth (approved):

PR #3663, opened 2023-09-25
KProve: `--infer-k-cell` option

### LINKED ISSUE #3656: `kprove --emit-json-spec ...` should allow not inferring the `<k>` cell being present
Tried with K version `v6.0.102` installed via `kup`.

Given a K definition that looks like this:

```
module TEST
    imports BOOL
    imports INT
    imports K-EQUAL-SYNTAX

    syntax Pgm ::= Bool | Int

endmodule

module TEST-SPEC
    imports TEST

    claim [claim1]: true => false
endmodule
```

We can kompile this and convert the specification to a JSON spec in this way:

```
kompile --backend haskell test.k --main-module TEST
kprove --dry-run --emit-json-spec test-spec.json test.k --spec-module TEST-SPEC --claims TEST-SPEC.claim1
```

Then if we inspect the generated file `test-spec.json`, we see something like this:

```
% cat test-spec.json | jq .                                                                                                                                                                  
{                                                                                                                                                                                                                    
  "format": "KAST",                                                                                                                                                                                                  
  "version": 3,                                                                                                                                                                                                      
  "term": {                                 
    "node": "KFlatModuleList",             
    "mainModule": "TEST-SPEC",         
    "term": [            
      {                                       
        "node": "KFlatModule",                                                                            
        "name": "TEST-SPEC",       
        "imports": [                        
          {                                                                                               
            "node": "KImport",
            "name": "TEST",         
            "isPublic": true                                                                              
          }                 
        ],                                                                                                
        "localSentences": [        
          {                                                                                               
            "node": "KClaim",         
            "body": {                
              "node": "KApply",                      
              "label": {                             
                "node": "KLabel",                                                                         
                "name": "<generatedTop>",                                                                                                                                                                            
                "params": []
              },                                                                                          
              "arity": 2,                            
              "args": [                               
...
```

So it's taken the K claim `true => false`, which does not have a configuration attached to it, and turned it into a reachability claim over the configuration by inferring that the user wanted it wrapped in a `<k>` cell like `<k> true => false ... </k>`.

In several places in pyk, including in KEVM, we actually read these K claims to build equality proofs instead of reachability proofs (discharged with the RPC server `simplify` endpoint instead of the `execute` endpoint). This is faster and actually stronger, instead of just checking that the simplification _implies_ the target term, we are checking that they are _equal_.

It would be good for `kprove` to either:

- Not infer a `<k>` cell if the user doesn't write one for a `claim ...`, or,
- Allow passing an argument `--no-infer-k-cell` which doesn't infer a `<k>` cell if the user doesn't write one.

Related: https://github.com/runtimeverification/k/issues/4211

gtrepta:
Fixes #3656

### DIFF
--- a/k-distribution/tests/regression-new/issue-3656-inferCells/Makefile
+++ b/k-distribution/tests/regression-new/issue-3656-inferCells/Makefile
@@ -1,0 +1,6 @@
+DEF=test
+TESTDIR=.
+KOMPILE_BACKEND=haskell
+KPROVE_FLAGS=--dry-run --emit-json-spec test-spec.json
+
+include ../../../include/kframework/ktest.mak
--- a/k-distribution/tests/regression-new/issue-3656-inferCells/test-spec.k
+++ b/k-distribution/tests/regression-new/issue-3656-inferCells/test-spec.k
@@ -1,0 +1,5 @@
+module TEST-SPEC
+    imports TEST
+
+    claim [claim1]: true => false
+endmodule
--- a/k-distribution/tests/regression-new/issue-3656-inferCells/test.k
+++ b/k-distribution/tests/regression-new/issue-3656-inferCells/test.k
@@ -1,0 +1,10 @@
+module TEST
+    imports BOOL
+    imports INT
+    imports K-EQUAL-SYNTAX
+
+    configuration <k> $PGM:Pgm </k>
+
+    syntax Pgm ::= Bool | Int
+
+endmodule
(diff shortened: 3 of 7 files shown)


ehildenb:
From discussion: it's not important that it's rendered to Kore because we only need the compiled JSON, not the Kore.

- If we do need to provide a rendering to Kore, the same way that simplification rules are rendered should be good.
- It's not needed to render to Kore anyway, because we are taking the spec.json.

PR #3664, opened 2023-09-25
Updating Maven and Testing it

Robertorosmaninho:
This simple PR updates the Maven version to fix the failing Release CI that builds Homebrew formulae.


dwightguth (approved):
Looks fine assuming the tests pass. If they don't, ask for my review again after fixing them.

PR #3666, opened 2023-09-26
Ensure type inference checks rhsSort <: lhsSort for macro rules

### LINKED ISSUE #3661: [K-Bug] `kompile` throw a raw llvm error message when a rule in a invalid `requires` and doesn't throw any in a invalid `ensures`

Front-End

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

K version:    v6.0.113-0-gcb848f3950 Build date:   Mon Sep 25 14:32:32 -03 2023

### Operating System

MacOS (Apple Silicon/AMD)

```
module TEST
    imports INT
    imports BOOL

    syntax Int  ::= #foo ( Int ) [function]
    syntax Bool ::= #bar ( Int ) [alias]

    rule #bar ( X ) => X // This should fail with type error: Int != Bool

    rule #foo( X ) => X       ensures  #bar( X ) // This passes, but it should?
    rule #foo( X ) => X       requires  #bar( X ) // This produces the llvm error
endmodule
```

### Steps to Reproduce

```
kompile test.k --main-module TEST --syntax-module TEST --backend llvm  --no-exc-wrap
```

### Expected Results

The front end should print an error in both cases saying that the ` rule #bar ( X ) => X` is returning an `Int` instead of a `Bool`.

Scott-Guest:
Closes #3661


Scott-Guest:
Related #3667

radumereuta (approved):

Issue #3669, opened 2023-09-27
`test-pr` workflow doesn't run tests on MacOS


gtrepta:
The `test-pr` workflow currently only runs tests on Linux and Nix. Tests aren't ran on a MacOS system until the `release` workflow, which led to us not catching an issue with a PR for #3530 until after it was merged.

ehildenb:
Sounds like we need to:

- Check what kind of load our macs are under (@F-WRunTime )
- Write a new testing block for the macs similar to https://github.com/runtimeverification/k/blob/25367c4e7f1551c8b2901103b9afa363a3a64a6e/.github/workflows/test-pr.yml#L75 for MacOS.

It sounds like just running _any_ test wouldn't have caught this problem, we need to run the whole test-suite.

Baltoli:
I will start to take a look at dropping Homebrew packaging and instead only building K locally in the test CI.

Baltoli:
Currently blocked on disabling Z3 in type inference for good.

PR #3670, opened 2023-09-27
Ordering instance for Sentences (2)

### LINKED ISSUE #3530: [K-Improvement] Create `Ordering` instance for `Sentence`'s subclasses

gtrepta:
fixes: #3530 

This branch was merged and then reverted because of an issue that didn't get caught earlier in CI. I have since pushed a fix for the issue and tested it to make sure that it will work. So, this should be good to go. But if we really want to be sure, then we can consider this PR blocked on #3669.


radumereuta (approved):
lgtm, hopefully this time is going to be successful.
I would like it if someone else approves it.

Robertorosmaninho (approved):
I've tested this locally (on MacOS13 M1), and it passes all tests, including the one that was failing previously on the release. Be free to label `automerge` on it if you like @gtrepta.

PR #3673, opened 2023-09-27
Implement new sort inference algorithm

### LINKED ISSUE #3601: Develop a more principled type inference algorithm
In order to improve performance and eventually add support for more powerful types, we should develop a robust type inference algorithm and only rely on the current Z3 implementation as a fallback.

Scott-Guest:
Closes #3601.

This PR introduces a new sort inference algorithm, aiming to eventually replace the Z3-based approach and pave the way for parametric rules and sorts. 

The design is heavily inspired by [The Simple Essence of Algebraic Subtyping: Principal Type Inference with Subtyping Made Easy](https://infoscience.epfl.ch/record/278576) by Lionel Parreaux. A high-level description explaining the relevant background can be found in `docs/developers/sort_inference.md`.

The current PR implements a limited form of the proposed algorithm, still falling back to the Z3-based algorithm for any terms containing:
- ambiguities
- strict casts
- parametric sorts

For reviewers, I would begin by reading
- `docs/developers/sort_inference.md` for a high-level conceptual understanding
- the comment at the top of `SortInferencer.java` briefly explaining how the high-level design is actually implemented with our data structures
- the paper if anything is unclear


ehildenb:
Let's make this into a blog post! Thanks for the diligent research :)

- K is a cool tech we're developing, lots of moving parts.
- Parsing is a huge part of that, which involves sort inference.
- We're a research company, so we did research on sort inference, found SimpleSub algorithm good for our cases.
- Novel from that, we also are able to handle parsing ambiguities! Letting users define grammar makes parsing very difficult, but we solved it!

dwightguth:
Most of this I understand, but I don't completely understand the theory behind the work you're doing with ambiguities. However, that's okay. As long as it works in practice, I understand the algorithm itself well enough to be able to say that it should be fine. Let's see what the implementation looks like and how it performs and behaves once it's complete.

Scott-Guest:
### Road Map
The first step is to implement the core of SimpleSub, but working with our data structures

- [x] Record constraints as bounds on type variables (ignoring strict casts, parametric sorts, and ambiguities)
- [x] Collapse bounds / make compact types
- [x] Perform co-occurrence analysis and simplify

After this, we need to make the necessary adjustments to change the inferred SimpleSub types to actual K sorts

- [x] Actually compute type joins and meets in bounds, erroring when impossible
- [x] Monomorphize inferred parametric sorts
- [x] Hook inference engine into the main pipeline, inserting type annotations
- [x] Determine which test cases pass, and whether we can use the current implementation with Z3 as a fall-back

Finally, we need to handle the more serious complications related to K:
- [ ] Handle ambiguities
- [ ] Add support for strict casts
- [ ] Add support for MInt
- [ ] Add support for general parametric sorts

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -241,7 +241,21 @@ public String syntaxModule(FileUtil files) {
...
+    CHECKED,
+  }
+
+  @Parameter(
+      names = "--type-inference-mode",
+      description =
+          "Choose between the Z3-based and SimpleSub-based type inference algorithms, or run both"
+              + " and check that their results are equal. Must be one of [z3|simplesub|checked].",
+      hidden = true)
+  public TypeInferenceMode typeInferenceMode = TypeInferenceMode.SIMPLESUB;
```
Scott-Guest: The default is currently feature-on so that performance runs are accurate, but this should be switched feature-off before merging.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Convert to an equivalent Sort, instantiating variables and actually computing a type join/meet
+   * as appropriate.
+   *
+   * @param polarity - The polarity where this CompactSort occurs.
+   * @param instantiation - A map indicating how variables should be instantiated
+   * @param subsorts - The Sort poset
+   * @return An equivalent Sort
+   */
+  public Either<LatticeOpError, Sort> asSort(
```
Scott-Guest, line 104: The call to `POSet.lowerBounds` in this function account for 79.5% of the type inferencer's total runtime in the Elrond performance test. The current approach is quite naive, so it may be worth optimizing in the future.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortInferenceError.java
```diff
@@ -0,0 +1,58 @@
...
+            + ". Expected: "
+            + rhs,
+        pr);
+  }
+}
+
+final class MonomorphizationError extends SortInferenceError {
+  // TODO: Produce better error messages!
+  public MonomorphizationError(HasLocation loc) {
+    super("Term is not well-sorted due to monomorphization failure.", loc);
```
Scott-Guest: This error message is quite poor, but I don't think it's actually possible to hit it currently (see my other comments on the "Over-Simplification" conceptual issue).

Scott-Guest:
## Conceptual Issue - Unrealizable Sub-Term Types
The current algorithm consists of two phases:
1. Infer a SimpleSub-esque type which may include
    - type intersections / unions such as `Int ∧ Bool` which do not actually exist in the subsort poset
    - type variables
2. Realize the SimpleSub-esque type as a K sort by
    - computing type intersections / unions as meets / joins in the subsort poset, erroring if they do not exist
    - monomorphizing type variables
    
In particular, this process is run on the inferred "function type" of the overall term (i.e. converting `TermSort<CompactSort>` to `TermSort<Sort>`). Thus, we only check that the type intersections / unions are valid for the sorts of the variables and the sort of the overall term. 

However, I am concerned we could run into the following situation:
- Both the type of the overall term and the type of every variable can actually be realized as a K sort
- The type of some sub-term of the overall term cannot actually be realized as a K sort (i.e. it involves a type intersection / union that does not exist), and we never check if this is valid

For example, consider
```
syntax Foo ::= "foo" [token]
syntax Bar ::= "bar" [token]
syntax Top1 ::= Foo | Bar
syntax Top2 ::= Foo | Bar
rule foo => bar
```
The rewrite here has type `Foo ∨ Bar`, but there is no such least upper-bound (`Top1` and `Top2` are both upper bounds yet incomparable), so this should report a type error. However, the produced `TermSort<Sort>` contains no variables and the overall sort is just `#RuleContent`, so `SortInferencer` never reports any issue.

For rewrites, this is still caught by a later check,
https://github.com/runtimeverification/k/blob/699568bace6836f9ba4c28b6556d06e998fea2ca/kernel/src/main/java/org/kframework/compile/AddSortInjections.java#L538
so this has not actually caused issues in practice yet. However, it's likely possible to construct other examples not involving the rewrite itself which this check will miss.

### Solution
Every `Production`'s declared sorts only involves sort parameters and primitive sorts. Thus, the only way some sub-term / `ProductionReference` can have a sort which can't be realized is if the instantiation of its sort parameters can't be realized. That is, it suffices to check that every sort parameter's instantiation can be realized in every polarity where it occurs.

## Conceptual Issue - Over-Simiplification
Because the top-level sort of the term is almost always `#RuleContent`, and the sort of the overall term is the only sort in positive position, every sort variable only occurs in negative position and thus will be simplified away, meaning the monomorphization logic never gets hit.

From our view of K `Term`s as Simple-Sub functions, this makes sense. Consider
```
𝜆𝑥 . if 𝑥 < 0 then 0 else 𝑥
```
with `nat <: int`, `< : int → int → bool`, and `0 : nat`. The inferred type is `𝛼 ⊓ int → nat ⊔ 𝛼` i.e. either `int → int` or `nat → nat`. If we wrap this term in some other function `ignore : ⊤ → ⊤`
```
𝜆𝑥 . ignore (if 𝑥 < 0 then 0 else 𝑥)
```
then the inferred type instead because `int → ⊤`. In a sense, the type simplification only cares if each type variable is relevant to the externally visible type of the function, without regard for the types of any sub-terms. The runtime type of `if 𝑥 < 0 then 0 else 𝑥` in the `ignore` example will still be `nat` if we pass in a `nat` for `x`, but this is irrelevant for the functions actual signature.

### Solution
For now, this doesn't actually matter because we fully monomorphize types, but if we eventually want to support parametric rules we will need to ensure that the exact sort of (at least some of the) sub-terms is expressible over the sorts of the variables.

For every sort that we need to express, we can just consider the usages of any type variable there during co-occurrence analysis to avoid simplifying them aways. However, I'm unsure which sorts actually matter, i.e. which of the following is relevant:
- every rewrite's sort?
- the sort of the overall rule?
- the sort of every single sub-term?

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
```diff
@@ -0,0 +1,125 @@
...
+    sorts.addAll(
+        ctors.stream()
+            .map(h -> new org.kframework.kore.ADT.Sort(h.name(), Seq()))
+            .collect(Collectors.toSet()));
+    Set<Sort> bounds = polarity ? subsorts.upperBounds(sorts) : subsorts.lowerBounds(sorts);
+    bounds.removeIf(
+        s ->
+            subsorts.lessThanEq(s, Sorts.KLabel())
+                || subsorts.lessThanEq(s, Sorts.KBott())
+                || subsorts.greaterThan(s, Sorts.K()));
```
Scott-Guest: I am not confident that filtering the bounds this way is correct, but it was necessary lest we get multiple unrelated bounds like `KLabel` and `K`.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
dwightguth: I'm not sure I understand the rationale behind leaving it off by default. Surely we want it on so that it gets tested by our integration testing moving forward, right?

dwightguth (commented):
I'm not sure I fully understand this code, but it seems well thought out, well documented, and well architected, and I'm going to assume that you've run the entire test suite with the `CHECKED` mode as the default, which would at least tell you that it's inferring the exact same terms as the old algorithm, so assuming that's correct, this seems reasonable.

See my comment though; if we've adequately tested this on the entire test suite and know that it's identical in behavior to the old system on those tests, and if it's easy to turn off in the event of bugs in other semantics, and it's faster than the old code, we ought to turn this on by default.

dwightguth:
There might be room to add a bit more documentation on fields and classes that don't currently have javadocs associated with them, though.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
Scott-Guest: @dwightguth CI passes with `CHECKED` mode, so we do infer all the same sorts (including on the performance tests which should cover EVM, Elrond, etc.).

However, given the conceptual issues I mentioned in another comment, I would like a bit more time to either
- address those issues in a follow-up PR, or
- prove to myself that they aren't real issues 

before turning the feature on.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
Scott-Guest: #3834 combined with some special casing for zero/one elements reduce this to 12.5%

docs/developers/sort_inference.md
Baltoli, line 1: This is a really nice overview of the process; excellent work - made it much easier to follow what we're doing conceptually.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
Baltoli: It will also let us selectively roll the feature out to projects one-by-one and keep an eye on them to make sure we don't see bugs or performance regressions there. I agree that in the end we definitely want this on by default, but it's a big enough feature that for now we should take care rolling it out.

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -401,20 +411,44 @@ private Tuple2<Either<Set<KEMException>, Term>, Set<KEMException>> parseStringTe
...
+            inferencers.add(currentInferencer);
+          }
+        }
+        Either<Set<KEMException>, Term> z3Rez =
+            new TypeInferenceVisitor(currentInferencer, startSymbol, isAnywhere).apply(rez3);
+        if (infModeForTerm == KompileOptions.TypeInferenceMode.CHECKED) {
+          if (!((rez.isLeft() && z3Rez.isLeft())
+              || (rez.isRight()
+                  && z3Rez.isRight()
+                  && rez.right().get().equals(z3Rez.right().get())))) {
```
Baltoli: I don't know if there's a more ergonomic way to write this condition? It makes sense to me but took a couple of goes to make sure I was parsing it properly.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+public record CompactSort(Set<SortVariable> vars, Set<SortHead> ctors) {
+
+  public CompactSort(SortVariable var) {
+    this(
+        new HashSet<>() {
+          {
+            add(var);
+          }
+        },
```
Baltoli: Do we use this syntax elsewhere in the frontend? I hadn't seen it before and had to look up what was happening, but that may be my lack of Java knowledge showing. If we don't use it elsewhere it might be clearer to write it some other way.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
```diff
@@ -0,0 +1,129 @@
...
+  public static CompactSort makeCompact(BoundedSort sort, boolean polarity) {
+    if (sort instanceof BoundedSort.Constructor ctor) {
+      if (ctor.head().params() == 0) {
+        Set<SortHead> ctors = new HashSet<>();
+        ctors.add(ctor.head());
+        return new CompactSort(new HashSet<>(), ctors);
+      }
+      throw new AssertionError("Parametric sorts are not yet supported!");
+    }
+    BoundedSort.Variable var = (BoundedSort.Variable) sort;
```
Baltoli: Is there a way that the compiler can figure out that `sort` must be a `Variable` here? I can see by inspection that this is safe, but maybe worth commenting or rephrasing so that the compiler can check it for us.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
```diff
@@ -0,0 +1,129 @@
...
+          }
+        },
+        new HashSet<>());
+  }
+
+  /**
+   * Compact a BoundedSort, converting all transitive bounds into a single intersection / union.
+   *
+   * @param sort - The BoundedSort to make compact
+   * @param polarity - The polarity where sort occurs
```
Baltoli: true is positive, false is negative like in the paper?

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/InferenceDriver.java
```diff
@@ -0,0 +1,132 @@
...
+      rhsVar.lowerBounds().add(lhs);
+      for (BoundedSort rhsUpper : rhsVar.upperBounds()) {
+        constrain(lhs, rhsUpper, pr);
+      }
+      return;
+    }
+
+    // If they are primitive sorts, we can check the sort poset directly
+    BoundedSort.Constructor lhsCtor = (BoundedSort.Constructor) lhs;
+    BoundedSort.Constructor rhsCtor = (BoundedSort.Constructor) rhs;
```
Baltoli, line 114: See comment above - I presume the `instanceof` typing rules aren't powerful enough to propagate the information that `lhs` and `rhs` must both be constructors here.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortInferenceError.java
```diff
@@ -0,0 +1,70 @@
...
+            + pr.production()
+            + ". Expected: "
+            + rhs,
+        pr);
+  }
+}
+
+/** An error indicating that some type variable cannot be monomorphized as an actual K sort. */
+final class MonomorphizationError extends SortInferenceError {
+  // TODO: Produce better error messages!
```
Baltoli, line 63: I recall you mentioned that your prototyping didn't lead to easy-to-produce nice errors; maybe worth expanding on this comment with some more context.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortVariable.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) K Team. All Rights Reserved.
+package org.kframework.parser.inner.disambiguation.inference;
+
+/** A variable which could be instantiated with any Sort. */
```
Baltoli: There's a comment elsewhere explaining that we use these for reference semantics, but perhaps mention it here as well just so it's clear why we have an empty class hanging around.

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortInferencer.java
```diff
@@ -0,0 +1,594 @@
...
+ *       CompactType in SimpleSub).
+ *   <li>Convert the inferred CompactSort into a normal K Sort
+ *       <ul>
+ *         <li>Monomorphize each sort variable, allowing it to take any value between its recorded
+ *             bounds and possibly producing multiple valid monomorphizations.
+ *         <li>For each type intersection/union, actually compute the corresponding meet/join on the
+ *             subsort poset, erroring if no such meet/join exists.
+ *       </ul>
+ *   <li>Insert a SemanticCast around every variable in the Term to record the results.
+ * </ol>
```
Baltoli, line 80: Very nice!

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortInferencer.java
```diff
@@ -0,0 +1,594 @@
...
+    if (t instanceof Constant) {
+      return false;
+    }
+    return ((TermCons) t).items().stream().anyMatch(SortInferencer::hasAmbiguity);
+  }
+
+  private static boolean hasStrictCast(Term t) {
+    if (t instanceof Ambiguity) {
+      return ((Ambiguity) t).items().stream().anyMatch(SortInferencer::hasStrictCast);
+    }
```
Baltoli:
```suggestion
    if (t instanceof Ambiguity amb) {
      return amb.items().stream().anyMatch(SortInferencer::hasStrictCast);
    }
```

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/SortInferencer.java
```diff
@@ -0,0 +1,594 @@
...
+    if (t instanceof Constant) {
+      return false;
+    }
+    return ((TermCons) t).items().stream().anyMatch(SortInferencer::hasStrictCast);
+  }
+
+  private static boolean hasParametricSorts(Term t) {
+    if (t instanceof Ambiguity) {
+      return ((Ambiguity) t).items().stream().anyMatch(SortInferencer::hasParametricSorts);
+    }
```
Baltoli:
```suggestion
    if (t instanceof Ambiguity amb) {
      return amb.items().stream().anyMatch(SortInferencer::hasParametricSorts);
    }
```

Baltoli (approved):
I've nitpicked the code a bit, but broadly speaking I agree with Dwight - this looks really well thought through, and I'm confident enough in the surrounding infrastructure that we should be OK to get it merged and try it out on live projects using a flag.

Excellent work Scott; well done for seeing this through to the point of merging!

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
Scott-Guest: The first stack overflow post I found on the topic says ["Every time someone uses double brace initialisation, a kitten gets killed."](https://stackoverflow.com/questions/1958636/what-is-double-brace-initialization-in-java), so probably worth changing!

kernel/src/main/java/org/kframework/parser/inner/disambiguation/inference/CompactSort.java
Scott-Guest: The only way I see to rephrase this is with pattern matching, which unfortunately doesn't land until Java 21.

However, Java 17 does have a preview feature that we could enable with `--enable-preview` which would suffice here, and I think this is worth considering: #3836

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
Scott-Guest: After discussing with @dwightguth, we decided to leave it off by default, but ensure it still gets tested to an extent on CI by adding `--type-inference-mode simplesub` in `ktest.mak` for the regression tests.

radumereuta:
I've tested the c-semantics with the new and old algorithm and here are some findings:
```
total time ms
1639 semantics/c.k:477   parse: 375 typeInf:1061
 568 semantics/c.k:477   parse: 339 typeInf:  18 // new alg

254 semantics/implementation/x86_64-linux-gnu/config.k:41   parse:  18 typeInf: 206
 48 semantics/implementation/x86_64-linux-gnu/config.k:41   parse:  13 typeInf:   1 // new alg
```
The new implementation is much faster.

PR #3674, opened 2023-09-28
[DRAFT] Introducing new workflow: Building K from source in MacOS

Robertorosmaninho:


Baltoli:
@Robertorosmaninho see [this issue](https://github.com/kframework/homebrew-k/pull/21) for a fix for the issue you're seeing when building K here; the Homebrew formula is in general a pretty good source for the best way to build K from source on macOS. Let me know if I can help at all with this

.github/workflows/test-pr-macos.yml
```diff
@@ -0,0 +1,156 @@
...
+           export PATH="$(brew --prefix llvm@15)/bin:$PATH"
+           export PATH="$(brew --prefix openjdk)/bin:$PATH"
+           echo $(which llvm-config)
+           export LLVM_CONFIG=$(which llvm-config)
+           export LLVM_DIR="$(brew --prefix llvm@15)/lib/cmake/"
+           mvn package -T1C
+         # brew install direnv
+         # cp macos-envrc .envrc
+         # eval "$(direnv hook bash)"
+         # direnv allow .
```
Baltoli, line 69: Looks like you're reaching that conclusion, but `direnv` is not something I'd expect to see in CI - it's just a local developer convenience tool.

Robertorosmaninho: Actually, I don't quite follow what you mean by
> the Homebrew formula is in general a pretty good source for the best way to build K from source on macOS

Could you please elaborate on that?

Baltoli: The code is here: https://github.com/kframework/homebrew-k/blob/master/Formula/kframework.rb

If you've not seen a Homebrew formula before, it's basically a Ruby DSL that allows you to specify dependencies and build steps for a package. We maintain one for K; if there are macOS-specific issues with the build then it's plausible that the Homebrew code has already dealt with it somehow

PR #3676, opened 2023-09-28
change case classes and parser to support multi-ary \and/\or.

dwightguth:
At this stage, actually parsing a multi-ary term will lead to an internal representation that is handled incorrectly by all the code that interprets KORE. However, this change ought to pass CI and be sufficient to be able to modify the llvm backend without anything breaking.


kore/src/main/scala/org/kframework/parser/kore/Default.scala
Robertorosmaninho, line 1: We're maintaining the version with `_1` and `_2` just to have backward compatibility for now, and then they will be removed in future PRs, right? So when we have the complete implementation of the multi-ary `\and` ans `\or`, they will no longer be necessary/used anywhere, right?

dwightguth: That's correct. the only place where we will still keep the binary version is in the constructor. The binary fields will be removed in the next PR into this repo in the sequence:  https://github.com/runtimeverification/k/pull/3678/commits/dd60e0698184e964258e110507e8f59d49885998

Robertorosmaninho: Awesome! Thanks for the clarification!

Robertorosmaninho (approved):
LGTM! Looking forward to the next steps!

PR #3678, opened 2023-09-28
update scala kore parser to use multi-ary and and or everywhere

dwightguth:
Blocked on https://github.com/runtimeverification/llvm-backend/pull/842

This is the second wave of changes to the frontend to update the scala parser to uses multi-ary and and or. This now completes the changes to the frontend except that we will need one minor third PR to make the class we made public private again after making the second wave of changes to the llvm backend.


Robertorosmaninho (approved):

Robertorosmaninho:
Is it possible to have some documentation on these pattern matchings? They're starting to be confusing, at least for me, as I don't touch this part of the code often.

PR #3679, opened 2023-09-28
Finish all changes to the K frontend, LLVM backend, and Haskell backend relating to multi-ary \and/\or

dwightguth:
Blocked on https://github.com/runtimeverification/llvm-backend/pull/844

Note: this is also blocked on the changes to the Haskell backend

This change finalizes the sequence of changes used to implement multi-ary \and and \or in the frontend. We remove the parsing of \left-assoc and \right-assoc of \and and \or, and we emit multi-ary and and or where previously we were using \right-assoc in ModuleToKore. We also revert the last of the temporary changes we made.


Robertorosmaninho (approved):
LGTM

Issue #3675, opened 2023-09-28
Install locations in packaging


gtrepta:
The ubuntu package installs into some strange paths and has files that don't belong in it.

- We install `/usr/lib/scripts/util.sh` and `/usr/lib/version`, which have no indication on first glance of what package they're a part of.
- We install a bunch of headers into `/usr/include/fmt`, sharing ownership over those files with the `libfmt-dev` package.
- Some files that go to `/usr/bin` either shouldn't be there (`ng`, `booster-dev`, `eventlog-parser`?) or need to be named better (`rpc-client`)

A rough outline of folders created under `/usr` by the package:
```
usr
├── include
│   ├── fmt
│   ├── kframework
│   ├── kllvm
│   └── kllvm-c
├── lib
│   ├── cmake
│   │   └── kframework
│   ├── kframework
│   ├── kllvm
│   └── scripts
└── share
    ├── doc
    │   └── kframework
    └── kframework

```

IMO it should look like this:

```
usr
├── include
│   ├── kframework
├── lib
│   ├── cmake
│   │   └── kframework
│   ├── kframework
└── share
    ├── doc
    │   └── kframework
    └── kframework

```

dwightguth:
The /usr/lib/kframework and /usr/include/kframework directories were created for this purpose. We can move scripts and fmt under the respective directories.

F-WRunTime:
Here is the debian Directory Tree definitions / standards  @gtrepta 

https://www.debian.org/releases/buster/amd64/apcs02.en.html

gtrepta:
#3683 Has been drafted to move `/usr/lib/version`.

My next step is to move the llvm-backend under `/usr/lib/kframework`. This will clean up a lot of the filesystem. However, this will require putting the llvm-backend's binaries under `/usr/lib/kframework/bin` and symlinking to them from `/usr/bin`. I have to open a PR for the llvm-backend first to make it more tolerant towards these symlinks (https://github.com/runtimeverification/llvm-backend/pull/848).

Further work ahead:
- The [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s05.html) states that `/usr/include` is for C language files. So, I think it's a good idea to move `include/kframework` to `lib/kframework/include`
- Audit the binaries installed to `/usr/bin` for any that should be moved to `/usr/lib/kframework/bin`

Baltoli:
We should maybe move the `ng` binary out of the default binary search paths.

PR #3683, opened 2023-09-29
Packaging: Move `version` into lib/kframework

gtrepta:
For #3675 

We're installing `version` into `/usr/lib/version`. This change moves it to `/usr/lib/kframework/version`.

### DIFF
--- a/k-distribution/src/main/assembly/bin.xml
+++ b/k-distribution/src/main/assembly/bin.xml
@@ -69,7 +69,7 @@
     </fileSet>
     <fileSet>
       <directory>${project.parent.basedir}/package</directory>
-      <outputDirectory>/lib</outputDirectory>
+      <outputDirectory>/lib/kframework</outputDirectory>
       <includes>
         <include>version</include>
       </includes>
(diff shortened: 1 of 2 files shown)


ehildenb (approved):

ehildenb:
Are there any bash scripts that read this file directly? Does `krun --version` call the JVM or just read the file directly, for example?

gtrepta:
> Are there any bash scripts that read this file directly? Does `krun --version` call the JVM or just read the file directly, for example?

The scripts are calling `kompile --version`

Robertorosmaninho:
Actually, `krun`, `kast`, `kompile`, and other tools call this script [k](https://github.com/runtimeverification/k/blob/master/k-distribution/src/main/scripts/lib/k), which is responsible for printing the `--version` by reading it from this file. We don't start the JVM anymore to get the K version.

This modification caused this issue: https://github.com/runtimeverification/k/issues/3690

gtrepta:
Uuuugh. I tested the packaging locally hoping to catch something like this, but the tests must not be covering it.

PR #3692, opened 2023-10-04
Test K version output on packaging

### LINKED ISSUE #3690: [K-Bug] v6.0.129 not version and so prints as "v"

Front-End

### Which command

- [X] kompile
- [X] kast
- [X] krun
- [X] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v6.0.129

### Operating System

Linux

_No response_

### Steps to Reproduce

daniel@daniel-MS-7E06 ~$ kup install k
removing 'github:runtimeverification/k/v6.0.128#packages.x86_64-linux.k'
 ✅ Successfully updated 'k' to version k (master).
daniel@daniel-MS-7E06 ~$ krun --version
cat: /nix/store/9swhc7ywxifkg5x05s65wzpf81i5vygr-k-6.0.129-e30530e4e57a860b57d66f366c8192f9b37a2657-maven/lib/version: No such file or directory
K version:    v
Build date:   Wed Oct 04 06:08:27 AEST 2023
daniel@daniel-MS-7E06 ~$ 

### Expected Results

daniel@daniel-MS-7E06 ~$ krun --version
K version:    v6.0.129
Build date:   Wed Oct 04 01:49:55 AEST 2023

Robertorosmaninho:
Follow-up to #3691.

This PR creates a simple test on the `test-package` to check whether a modification on K breaks the `--version` output, as the current regression test couldn't catch recent problems with it.

The test is pretty simple: it uses the old system of getting `--version` from the JVM execution of `kserver` and checks it against the `kompile --version` that uses the k script implementation that avoids JVM startup to output the tool's version. If they are the same, the test succeeds and fails otherwise! 

This PR should fail until #3691 gets merged!

### DIFF
--- a/package/test-package
+++ b/package/test-package
@@ -23,5 +23,16 @@ sort_int = kllvm.ast.CompositeSort("SortInt")
 assert str(sort_int) == "SortInt{}"
 HERE
 
+# Make sure that the version of the various components is correct comparing to
+# the KServer version that uses JVM.
+KOMPILE_VERSION=$(kompile --version | grep -oP 'v[0-9]+\.[0-9]+\.[0-9]+-[0-9]+-g[0-9a-z]+(-dirty)?')
+KSERVER_VERSION=$(kserver --version | grep -oP 'v[0-9]+\.[0-9]+\.[0-9]+-[0-9]+-g[0-9a-z]+(-dirty)?')
+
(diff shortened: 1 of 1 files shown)


Robertorosmaninho:
Fixed with https://github.com/runtimeverification/k/pull/3691

PR #3717, opened 2023-10-16
Move PL tutorial link out of main "learn K" section

### LINKED ISSUE #3704: Update, remove or document `pl-tutorial`
The PL tutorial is leading to confusion from external users who are unable to complete the lessons as written with modern K: https://github.com/runtimeverification/k/issues/3698

We should do one of the following things:
* Update the lessons to reflect the current set of features supported by modern K
* Remove the broken code (rather than just [dropping it silently from CI](https://github.com/runtimeverification/k/commit/6c122465e238d5df71a8b4c2fe71d79ab2089216))
* Document externally that this tutorial is out of date and should not be relied on

Baltoli:
Partially addresses https://github.com/runtimeverification/k/issues/3704 by de-emphasising the PL tutorial on the main K webpage. Happy to bikeshed the specific placement and wording used here.


web/pages/index.md
```diff
@@ -41,6 +40,7 @@ call/cc.
 
 ## Resources
 
+- A set of <a href="/k-distribution/pl-tutorial/README.md">reference implementations and tutorials</a> for common programming language features and paradigms is available, although parts of these implementations may not be fully up to date with modern K features.
```
radumereuta, line 43: Could we mention somewhere which definitions are not up to date?
We may be able to tell by looking at the Makefiles, but someone from outside won't.
The best solution I can think of right now is to add a note at the beginning of each file:
```
// NOTE: this definition is not up to date with the latest version of K.
// It is here for documentation/academic purposes.
```
Without such a message, you're still going to get those messages of confusion from the users.

Baltoli: Yeah, the only way to tell what's up to date is to look at what _actually_ gets run in CI via the Makefile groups. I'll go through and add something like that message to the ones that don't get run in CI.

Baltoli:
I've added the suggested message to the unsupported lessons (identified per the lesson Makefiles).

dwightguth (approved):

PR #3721, opened 2023-10-17
Make bison parser enabled by default

### LINKED ISSUE #1741: Bison parser is default
The goal here is to try to improve usability and other behavior to the point where it may be possible to make the Bison parser the default program parser for K.

dwightguth:
Fixes #1741 

This enables the bison parser by default everywhere. We can still get the old jit parser behavior by passing a flag.

Not ready for review yet; still needs more testing.


kernel/src/main/java/org/kframework/parser/inner/kernel/KSyntax2Bison.java
```diff
@@ -396,60 +397,79 @@ private static void processProduction(Production prod, Module module, Module dis
           "  node *n = malloc(sizeof(node) + sizeof(node *));\n" +
           "  n->str = false;\n" +
           "  n->location = @$;\n" +
-          "  n->hasLocation = " + (hasLocation ? "1" : "0") + ";\n" +
```
radumereuta, line 399: @Baltoli with Java 15, you can do multiline strings which hopefully will make this look a bit nicer.

PR #3722, opened 2023-10-17
Support for memset hook.

mariaKt:
This PR adds support in the frontend for the `memset` hook, implemented [here](https://github.com/runtimeverification/llvm-backend/pull/866) for the llvm backend.


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -2109,6 +2109,21 @@ is not a valid index.
   syntax Bytes ::= replaceAtBytes(dest: Bytes, index: Int, src: Bytes) [function, hook(BYTES.replaceAt)]
 ```
 
+### Multiple bytes update
+
+You can modify a `Bytes` to return a `Bytes` which is equal to `dest` except
+the `count` bytes starting at `index` are replaced with `count` bytes of value
+`Int2Bytes(1, v, LE/BE)` in O(count) time. This does not create a new `Bytes`
+object and will instead modify the original on concrete backends.
+This will throw an exception if index` + `count` is not a valid index.
```
```suggestion
This will throw an exception if `index` + `count` is not a valid index.
```

Baltoli (changes requested):
Could we get an integration test for this change? It can just be a K definition that does the same things as the unit tests from the backend, but makes sure that the hook exists and is wired up to the backend correctly.

Baltoli (approved):
Looks good!

PR #3733, opened 2023-10-19
Add `lowerBounds` and `upperBounds` methods to `POSet`

Scott-Guest:
This PR does the following:
- Cleans up all the warnings in `POSet.scala`
- Adds methods to `POSet` to compute all lower bounds or upper bounds of a given set of elements. 
- Modifies the least upper bound computation in `AddSortInjections.java` to use these new methods

This is just to remove some duplication when computing joins and meets in the new type inference engine.


Baltoli (approved):
This seems good to me; thanks Scott!

radumereuta (approved):

PR #3734, opened 2023-10-20
Slimmer nix runtime closure

### LINKED ISSUE #3732: Slimmer Nix
- At runtime, don't need jdk, just need jre.
- Use clang in frontend instead of gcc to avoid needing both in closure.
- Avoid using two versions of Java because llvm-backend-matching is now in frontend.

goodlyrottenapple:
Fixes #3732 

Closure size M1 Mac:
image

Closure size Linux:
image

Note that the Linux version also includes gdb

I also did not attempt to remove gcc because for some reason clang depends on it anyway in the llvm-backend, hence we would have to remove it there first.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,6 +64,7 @@
             k-framework = { haskell-backend-bins, llvm-kompile-libs }:
               prev.callPackage ./nix/k.nix {
                 inherit (prev) llvm-backend;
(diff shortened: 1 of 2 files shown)


Baltoli (approved):
Nice one!

Issue #3699, opened 2023-10-10
Update JDK to LTS 21


Robertorosmaninho:
Blocked in #3591 

Java 21 is the new LTS version of Java. Many newer features were introduced in this version, which will probably be included in the more recent versions of the images we currently use. 

We should keep track of the possibility of using it as soon as all our dependencies work with it. We're currently investigating the bumping to JVM 17, which is blocked by the focal support across our tools. Similar issues will block us in the future from getting JVM 21.

Scott-Guest:
While Java 17 has record types, pattern matching is very limited and really only a nicer syntax for `instanceof` checks. The language feature is also only a preview which must be explicitly enabled with a flag. 
```
int perimeter(Shape shape) {
   return switch (shape) {
       case Circle c -> 2 * Math.PI * c.getRadius();
       case Rectangle r -> 2 * r.getWidth() + 2 * r.getLength();
   };
}
```
Java 21 adds real pattern matching which can actually destructure data:
```
int perimeter(Shape shape) {
   return switch (shape) {
       case Circle(int radius) -> 2 * Math.PI * radius;
       case Rectangle(int width, int length) -> 2 * width + 2 * length;
   };
}
```

PR #3736, opened 2023-10-20
Bump JDK requirement to 17

Baltoli:
This PR bumps the compiler version used by Maven to build the K distribution from 11 to 17; no substantial further changes are required (other than documentation) as we already install version 17 or a rolling release in all of our packaging code. One minor required change was to pin the JRE version on Ubuntu and Debian to 17, rather than using the default version (11).

Once this change is merged, we can investigate whether there are any places in the K source where we could take advantage of new Java features.

Part of https://github.com/runtimeverification/k/issues/3699

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ The following dependencies are needed either at build time or runtime:
 *   [fmt](https://fmt.dev/)
 *   [gcc](https://gcc.gnu.org/)
 *   [gmp](https://gmplib.org/)
-*   [jdk](https://openjdk.java.net/) (version 11 or greater)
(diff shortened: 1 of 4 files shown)


radumereuta (approved):
Did you check the Dockerfiles?

Baltoli:
Yes - the Dockerfiles we use are already parameterized on `JDK_VERSION`, with a default value of 17:

https://github.com/runtimeverification/k/blob/e490da8a4673bc9c0a6cd892ccb0c7a431ffecd9/.github/actions/with-docker/action.yml#L21-L24

dwightguth (approved):

PR #3741, opened 2023-10-23
Fix for macOS nix / kup breakage

Baltoli:
On my Intel mac machine, the minimal JRE package is broken:
```console
$ /nix/store/6bqyc34isvgrrgl3nrqilzqbx7r1dyq5-zulu19.30.11-ca-jdk-minimal-jre-19.0.1/bin/java
[1]    31710 killed
```

This causes the default `kup`-installed K to break; the fix is just to identify this platform and package the full JRE. The only downside here is a slightly bigger closure for users on this platform, which is likely to be deprecated at some point in the future anyway.

I have also added a new minimal "smoke test" to the Nix CI tests; previously we were not checking that `kompile` etc. actually run on Intel macOS.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -230,9 +230,11 @@ jobs:
       - name: 'Build K Framework'
         run: GC_DONT_GC=1 nix build --print-build-logs .
 
+      - name: 'Smoke test K'
+        run: GC_DONT_GC=1 nix build --print-build-logs .#smoke-test
+
       # These tests take a really long time to run on other platforms, so we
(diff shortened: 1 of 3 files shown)


goodlyrottenapple (approved):
I [looked it up](https://github.com/NixOS/nixpkgs/blob/5550a85a087c04ddcace7f892b0bdc9d8bb080c8/pkgs/top-level/all-packages.nix#L15645C4-L15645C4) and `jre` is just a synonym for `jdk` anyway :D

Baltoli:
Thought I was being clever using `jre` instead of `jdk` 😩

PR #3743, opened 2023-10-23
Check for Java 17 when running K tools

Baltoli:
This script is responsible for checking the installed Java version when we run a K tool; it appears to have been missed out when we bumped the required version of Java from 8 to 11, and then to 17.


Robertorosmaninho (approved):
Good spotted! Thanks for this! LGTM!

dwightguth (approved):

PR #3745, opened 2023-10-24
Apply Java 17 automatic migration: enhanced switch

Baltoli:
This PR applies a **purely mechanical** upgrade of `switch` statements in the K repository to use the new "enhanced" syntax from Java 17. I have not altered any code here by hand; the results are obtained only via IntelliJ's refactoring tools.

### DIFF
--- a/kernel/src/main/java/org/kframework/unparser/Formatter.java
+++ b/kernel/src/main/java/org/kframework/unparser/Formatter.java
@@ -46,30 +46,19 @@ public class Formatter {
                 if (c == '%') {
                     char c2 = format.charAt(i + 1);
                     i++;
-                    switch(c2) {
-                    case 'n':
-                        indenter.newline();
-                        break;
-                    case 'i':
+                    switch (c2) {
+                    case 'n' -> indenter.newline();
+                    case 'i' -> {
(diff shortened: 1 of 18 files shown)


kernel/src/main/java/org/kframework/unparser/Formatter.java
```diff
@@ -96,7 +85,7 @@ public static void format(Term term, Indenter indenter, ColorSetting colorize) {
                             Term inner = tc.get(nt);
                             boolean assoc = false;
                             if (inner instanceof TermCons) {
-                                TermCons innerTc = (TermCons)inner;
+                                TermCons innerTc = (TermCons) inner;
```
radumereuta, line 88: This can be changed to something like this:
```java
if (inner instanceof TermCons innerTc) {
...
}
```
In fact, there are a lot of places where this can be done. It might deserve another PR.

radumereuta (approved):

kernel/src/main/java/org/kframework/unparser/Formatter.java
Baltoli: Yes, I'm going to follow up and do this automatically.

PR #3746, opened 2023-10-24
Apply Java 17 automatic migration: records

Baltoli:
This PR applies automatic fixes suggested by IntelliJ to refactor Java `class`es into `record`s, which reduces boilerplate code in a lot of different places.

The changes made are large, but I have split the changes up so that each commit deals only with a single class at a time, along with any downstream changes required by making that class a record. In some cases, I have applied small fix-its suggested by IntelliJ as well as the record transformation.

I'm happy to bikeshed whether any of these cases should remain a class rather than a record, and the commit history means that reverting an individual case is easy should we decide to do so.


haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
```diff
@@ -294,8 +271,7 @@ private String saveKoreSpecToTemp(ModuleToKORE converter, Module rules) {
                 String koreOutput = converter.convertSpecificationModule(module, rules,
                         haskellKRunOptions.defaultClaimType, sb);
                 files.saveToTemp("spec.kore", koreOutput);
-                String specPath = files.resolveTemp("spec.kore").getAbsolutePath();
-                return specPath;
+                return files.resolveTemp("spec.kore").getAbsolutePath();
```
radumereuta, line 274: I sometimes do this to put breakpoints and inspect the output of that function call.
Directly returning the result makes it difficult to inspect at runtime.
So, I wouldn't try to fix these in the future.

kernel/src/main/java/org/kframework/compile/checks/CheckSortTopUniqueness.java
```diff
@@ -18,20 +18,7 @@
...
-    /**
-     * Check that the given module has no sort that has multiple top sorts.
-     * @param errors to be updated when violations occur. Multiple violations will be accumulated in `errors`.
-     * @param module to be checked.
-     */
-    public CheckSortTopUniqueness(Set<KEMException> errors, Module module) {
-        this.errors = errors;
-        this.module = module;
-    }
+public record CheckSortTopUniqueness(Set<KEMException> errors, Module module) {
```
radumereuta, line 21: This comment disappeared.

Scott-Guest (approved):
I skimmed through all the commits, and LGTM!

I like all the small fixes from IntelliJ - we should consider using Qodana (https://www.jetbrains.com/qodana/) to enforce these on CI.

kernel/src/main/java/org/kframework/kprove/KProve.java
```diff
@@ -18,43 +18,28 @@
...
-        this.files = files;
-        this.kprint = kprint;
-        this.kproveOptions = kproveOptions;
-        this.compiledDefinition = compiledDefinition;
-        this.loader = loader;
-        this.proofDefinitionBuilder = proofDefinitionBuilder;
-        this.rewriterGenerator = rewriterGenerator;
-        this.sw = sw;
-    }
+    public KProve {}
```
radumereuta, line 42: Does this work?

kernel/src/main/java/org/kframework/parser/inner/ParseCache.java
```diff
@@ -13,71 +13,11 @@
...
-
-        public Source getSource() {
-            return source;
-        }
+public record ParseCache(Module module, boolean strict,
+                         Map<String, ParsedSentence> cache) implements Serializable {
+    public record ParsedSentence(K parse,
+                                 Set<KEMException> warnings,
+                                 Set<KEMException> errors, int startLine,
+                                 int startColumn, Source source) implements Serializable {
```
radumereuta, line 21: Wow!

radumereuta (approved):
One comment has been lost. Not sure if it's still valid. You should check.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -78,15 +78,8 @@ private interface EarleyProductionItem {
...
-  private static final class EarleyNonTerminal implements EarleyProductionItem {
-    public EarleyNonTerminal(int sort, List<Sort> sorts) {
-      this.sort = sort;
-      this.sorts = sorts;
-    }
-
-    final int sort;
-    final List<Sort> sorts;
-
+  private record EarleyNonTerminal(int sort, List<Sort> sorts)
```
dwightguth, line 81: This is no longer static.

kernel/src/main/java/org/kframework/parser/inner/kernel/EarleyParser.java
```diff
@@ -101,15 +94,8 @@ public String toString() {
...
-  private static final class EarleyTerminal implements EarleyProductionItem {
-    public EarleyTerminal(Scanner scanner, int kind) {
-      this.scanner = scanner;
-      this.kind = kind;
-    }
-
-    private final Scanner scanner;
-    final int kind;
-
+  private record EarleyTerminal(Scanner scanner, int kind)
```
dwightguth, line 97: Same here

kernel/src/main/java/org/kframework/main/Main.java
```diff
@@ -123,14 +116,12 @@ public int runApplication() {
...
-            int retval = frontEnd.get().main();
-            return retval;
+            return frontEnd.get().main();
         } catch (ProvisionException e) {
             for (Message m : e.getErrorMessages()) {
-                if (!(m.getCause() instanceof KEMException)) {
+                if (!(m.getCause() instanceof KEMException ex)) {
                     throw e;
                 } else {
-                    KEMException ex = (KEMException) m.getCause();
```
dwightguth: This works in a negated check and an else block?

kernel/src/main/java/org/kframework/parser/inner/ParseCache.java
```diff
@@ -13,71 +13,11 @@
...
-        public int getStartColumn() {
-            return startColumn;
-        }
-
-        public Source getSource() {
-            return source;
-        }
+public record ParseCache(Module module, boolean strict,
+                         Map<String, ParsedSentence> cache) implements Serializable {
+    public record ParsedSentence(K parse,
```
dwightguth, line 18: This is not static.

kernel/src/main/java/org/kframework/utils/RunProcess.java
```diff
@@ -26,27 +25,16 @@ public static Thread getOutputStreamThread(Supplier<InputStream> in, PrintStream
...
-        public final byte[] stdout;
-        public final byte[] stderr;
-        public final int exitCode;
-
-        public ProcessOutput(byte[] stdout, byte[] stderr, int exitCode) {
-            this.stdout = stdout;
-            this.stderr = stderr;
-            this.exitCode = exitCode;
-        }
+    public record ProcessOutput(byte[] stdout, byte[] stderr, int exitCode) {
```
dwightguth, line 32: Static?

dwightguth (commented):
My main comments are that it looks like the automated transformations removed final and static from the classes that were made into records. I know records are automatically final, so that part is fine, but I'm less certain that inner records are automatically static. If you can find evidence that they are, ignore all those comments, otherwise, the code probably needs to be audited to add the static keyword back in and make sure I didn't miss any places where there were inner records added.

Baltoli:
IntelliJ explicitly prompts for the removal of `static` on inner records @dwightguth, and [this Oracle page](https://docs.oracle.com/en/java/javase/17/language/records.html#GUID-6699E26F-4A9B-4393-A08B-1E47D4B2D263) backs up that conclusion (ctrl-f for "implicitly static")

kernel/src/main/java/org/kframework/kprove/KProve.java
Baltoli: This is the new (slightly niche) syntax to apply an annotation to the implicitly-generated constructor of a record

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
Baltoli: This seems to be [supported in IntelliJ](https://stackoverflow.com/questions/5010362/can-i-find-out-the-return-value-before-returning-while-debugging-in-intellij); my preference is for the cleaner code in the common case at the expense of an extra step on the occasional time that this code needs run in a debugger, but that's not a strong preference.

kernel/src/main/java/org/kframework/main/Main.java
Baltoli: Remarkably, yes! https://blogs.oracle.com/javamagazine/post/java-pattern-matching-instanceof-flow-scoping

kernel/src/main/java/org/kframework/compile/checks/CheckSortTopUniqueness.java
Baltoli: Will address!

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellRewriter.java
radumereuta: I wasn't aware of the feature. I guess we can remove them for now.

kernel/src/main/java/org/kframework/compile/checks/CheckSortTopUniqueness.java
Baltoli: The comment on this constructor just duplicates the sentiment in the top-level record comment, so I'm happy with the state of the change as it is after a second look.

Baltoli:
I have addressed and resolved all the comments left on this PR; none of them required any further changes to the code and so I'm going to merge the PR once CI passes with the merge conflicts resolved.

PR #3755, opened 2023-10-25
Apply "silent code cleanup" from IntelliJ

Baltoli:
This PR applies all automatic cleanups that IntelliJ is confident enough to apply silently with no further user input; a rough summary of the different kinds of change are:
* Removal of redundant casts (`T t = (T) v` where `v` already has type `T`).
* Removal of unnecessary calls to `toString()`
* Insertion of `final` where possible
* Collapsing of redundant boolean conditions `if (cond) return true else return false`
* Syntactic fixes (`;;` etc.)
* Use of `StandardCharsets.UTF_8` to simplify exception checks when encoding strings
* Removal of redundant explicit key checks when removing values from hashmaps

As ever, I'm happy to bikeshed these changes - I can redo the analysis and disable some subset of the checks if anyone has objections to the specific results we get.

~~To avoid merge conflicts, this PR is based on #3746 and should only be merged when that PR is merged.~~


kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -398,8 +398,8 @@ private static Tuple4<Set<Sentence>, Sort, K, Boolean> computeSentencesOfWellFor
                     fragmentItems.add(NonTerminal(childOptSort));
 
                     sentences.add(Production(Seq(), childOptSort, List(NonTerminal(childSort))));
-                    if (!m.definedKLabels().contains(KLabel("no"+childSort.toString()))) {
-                        sentences.add(Production(KLabel("no"+childSort.toString()), childOptSort, List(Terminal("no"+childSort.toString())),
+                    if (!m.definedKLabels().contains(KLabel("no"+ childSort))) {
+                        sentences.add(Production(KLabel("no"+ childSort), childOptSort, List(Terminal("no"+ childSort)),
```
radumereuta: Spaces before and after `+`?

kernel/src/main/java/org/kframework/compile/SortCells.java
```diff
@@ -784,7 +784,7 @@ public void apply(KApply k) {
                 }
                 @Override
                 public void apply(KVariable var) {
-                    if (!inKCell && cellFragmentVars.containsKey(var)) {
+                    if (!inKCell) {
```
radumereuta, line 787: What?! Why was this removed?

radumereuta (approved):
Overall, it looks good. I added a question though. Please double check.

kernel/src/main/java/org/kframework/compile/SortCells.java
Baltoli: From the [Java documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#remove-java.lang.Object-), you don't need to check if a key is present before removing it from a `HashMap`, so the explicit check is redundant.

kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
Baltoli: Yep, I'll fix this

Baltoli:
```suggestion
                    if (!m.definedKLabels().contains(KLabel("no" + childSort))) {
                        sentences.add(Production(KLabel("no" + childSort), childOptSort, List(Terminal("no" + childSort)),
```

PR #3757, opened 2023-10-25
Adopt Google Java style

Baltoli:
This PR is an initial attempt to build some consensus around a consistent, automated code style for the K Frontend's Java code. Other parts of K already do this for other languages, and so I think there's a pretty clear precedent that we should want to set something similar up for our largest, oldest codebase:
* The LLVM backend's C++ code is formatted with [`clang-format`](https://github.com/runtimeverification/llvm-backend/blob/master/.clang-format)
* The Haskell backend (and booster)'s Haskell code is formatted with [Fourmolu](https://github.com/runtimeverification/hs-backend-booster/blob/main/fourmolu.yaml)
* Pyk's Python code is formatted with [Black](https://github.com/runtimeverification/pyk/blob/41706ef5081585a4c71f7fccb4dea2d7a3796421/Makefile#L88-L89)

At a high level, the notional benefits of doing this are:
* Improved readability of code
* More focused PR review; no consideration needs to be given to formatting when reading and reviewing code
* Eliminates "diff noise" from unrelated formatting changes
* Consistency across the whole codebase; there are nearly 100 unique contributors to K which means that small variations in style have been accrued over time

I have done some research into available auto-formatters for Java, focusing on the following points:
1. Is the style agreeable to developers, and do we as a team feel happy with the results it produces?
2. Can the style be applied locally using our existing tools[^1]?
3. Can the style be enforced in CI using our existing Github Actions infrastructure?
4. Is there a community consensus around using this style in practice?

Beginning with (4) the standout contender for a "standardised" tool that's used by lots of Java developers in practice is Google's `google-java-format` tool, which is designed to apply the [Google Java Style](https://google.github.io/styleguide/javaguide.html) automatically. This PR focuses on using this tool and evaluating its tradeoffs; if we later decide it's totally unworkable then we can revisit this evaluation with a different tool.

My summary of the pros and cons of this format are:

#### Pros

* Implemented and maintained by a large company who rely on it for their own tooling.
  * Includes support for new language features and syntax.
  * It is unlikely that Google will stop using this tool on their Java code, and so the chances that it ends up bit-rotting into uselessness are very small.
* Active secondary ecosystem for IDE and CI tooling.
  * Official IntelliJ plugin
  * Third-party, actively maintained Maven integration
  * Third-party, actively maintained Github action
* **Personally speaking**, I like the style that's been applied; a skim through the diff suggests that some of the expressions I've found hardest to parse in the frontend are now a bit clearer
* Easy enough to get working locally; the only real issue I had was that I seemingly hadn't updated IntelliJ in two years and the JDK version it has internally was too old.
* Works well in CI; this PR adds an action to the test workflow that checks formatting.

#### Cons

* Not configurable beyond 2/4 space indentation; this is by design on Google's part[^2].
* One big diff will end up getting applied to the whole project[^3].
* Others may not prefer the style used.

I hope that this is a fair summary of the tradeoffs of doing this kind of tooling update; I'd like to get opinions from everyone who works on this code to make sure that we're all as happy as possible with what we end up doing. Things to think about in particular:
* Do you like the code style, from skimming the diff?
* Does the "format locally with IntelliJ, check + enforce in CI" workflow work for you, or would you rather have CI make commits to your branch so that you don't have to worry about formatting[^4]?
* Are there other tools you're aware of to do the same job?

[^1]: By this I really just mean IntelliJ; I'm not aware of anyone using different tools to develop the K Frontend. If you're using some other setup then please let me know!
[^2]: This is also the situation we have with Black for Python, however, and that works fine!
[^3]: Though this is true for _any_ formatter, and we probably just need to accept this pain as a once-off cost of doing business.
[^4]: The former is how we do automatic formatting for the other projects, so I am treating it as the default option here.


Baltoli:
Comments to address:
- [x] Fail fast; the tests should all depend on the formatting check
- [x] Make sure that the message is clear on what to do in case of a failure

kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
```diff
@@ -20,256 +27,385 @@
...
-            java.util.Set<Sentence> filtered = new HashSet<Sentence>();
+  // If a user guarantees that their semantics will never _dynamically_ try to rewrite an anywhere
+  // rule on the
+  // Haskell backend (with the --allow-anywhere-haskell flag), but cannot prove this statically, we
+  // allow them to
+  // strip out all those rules before sending the definition to the backend. If this transformation
+  // is applied
+  // unsoundly (i.e. an anywhere rule would have been attempted if it had not been stripped), the
+  // behaviour of the
+  // Haskell backend on that program is essentially undefined.
```
radumereuta: This looks weird. But it doesn't matter in this case.

kernel/src/main/java/org/kframework/compile/GenerateSortPredicateRules.java
```diff
@@ -1,61 +1,69 @@
...
 /**
- * Generates sort predicates from the subsort hierarchy of the module. This module assumes that the backend implements
- * the following rules:
- * isSort(#token(Sort, _)) => true
- * isK(K) => true
- * isKItem(K1(K2)) => true
- * isKItem(#token(_, _)) => true
+ * Generates sort predicates from the subsort hierarchy of the module. This module assumes that the
+ * backend implements the following rules: isSort(#token(Sort, _)) => true isK(K) => true
+ * isKItem(K1(K2)) => true isKItem(#token(_, _)) => true
```
radumereuta: This looks weird, and it might matter in this case.

Baltoli: Yep, I agree that looks weird. I suspect that for comments it's possible to write them in such a way as to avoid this; will investigate

Baltoli:
@radumereuta I resolved both your comments re. comment layout. For anyone reading / reviewing this change, a point I picked up while doing so is that the formatter **will** reformat JavaDoc comments (`/**`) but **will not** reformat ordinary block comments (`/*`).

If comment layouts are broken by the formatter, changing the type of comment is likely to be the answer.

Scott-Guest:
I don't care much about the specifics of the formatting (only that something consistent is enforced!), so I'm happy with the proposed changes.

In terms of the JavaDoc comments @radumereuta mentioned, you can also wrap things in a `<pre>` tag and they won't be formatted: https://github.com/google/google-java-format/wiki/FAQ#why-doesnt-the-formatter-touch-pre-or-table:~:text=Formatting%20can%20be%20disabled%20in%20javadoc%20comments%20using%20the%20%3Cpre%3E%20tag.

Baltoli:
Going to mark this as ready for review; I'd like to get it merged next week so that I can continue with automated cleanups, and we don't end up with a load of merge conflicts.

dwightguth (approved):
I'm not super thrilled about the order of imports, but this seems like a reasonable way of applying a standard format everywhere, and given that this is used by Google engineers it almost certainly has integrations with pretty much any editor we might choose to use, which was my main concern. So I'm going to approve this.

ehildenb (approved):

kernel/src/main/java/org/kframework/compile/CloseCells.java
```diff
@@ -10,281 +16,290 @@
...
 /**
  * Remove any use of dots in cells, by replacing them with variables and appropriate connectives.
  * This expects parent cells to have been added by earlier passes, it will only add variables
  *
- * The input to this pass is represents cells as described by {@link IncompleteCellUtils}.
- * In the output cells no longer have dots. Leaf cells have a single argument which is
- * the body, and parent cells are applied directly to the child cells and variables
- * as arguments of the KApply (though not necessarily in the right order) as
- * expected by {@link SortCells}.
+ * <p>The input to this pass is represents cells as described by {@link IncompleteCellUtils}. In the
```
radumereuta, line 23: What is the purpose of `<p>` here?
Was it added automatically? Or you added this manually?

kernel/src/main/java/org/kframework/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -24,620 +33,806 @@
...
+   * multiplicity ? desugar into an initializer production, an initializer rule, a cell production,
+   * and an empty production indicating the absence of that cell. Cells with children additionally
+   * generate a *CellFragment sort with the same arity as the cell production, but the arguments
+   * made optional by generating additional sorts. Cells which have parents and are not multiplicity
+   * * generate a CellOpt sort which is a supersort of the cell sort and has an additional
+   * production name like {@code <cell>-absent}. (For a cell with multiplicitly ? this is necessary
+   * to distinguish a fragment that did capture the state of the cell when it wasn't present, from a
+   * cell fragment that didn't even try to capture the cell).
+   *
+   * <p>Currently the implementation does not handle initializer rules; we will address this
```
radumereuta, line 60: There was a bit of a structure in this comment. Maybe we can recapture it with `<p>`

kernel/src/main/java/org/kframework/compile/GenerateSortPredicateRules.java
```diff
@@ -1,61 +1,71 @@
...
- * isSort(#token(Sort, _)) => true
- * isK(K) => true
- * isKItem(K1(K2)) => true
- * isKItem(#token(_, _)) => true
+/*
+ * Generates sort predicates from the subsort hierarchy of the module. This module assumes that the
+ * backend implements the following rules:
+ *
+ *  isSort(#token(Sort, _)) => true isK(K) => true
+ *  isKItem(K1(K2)) => true isKItem(#token(_, _)) => true
```
radumereuta: This looks mangled. It should be contained in a `<pre>` tag.

kernel/src/main/java/org/kframework/compile/ResolveIOStreams.java
```diff
@@ -26,392 +33,415 @@
...
-                    "\nIt should be one of " + streams);
+/** Created by daejunpark on 9/6/15. */
+public record ResolveIOStreams(Definition definition, KExceptionManager kem) {
+
+  /**
+   * Update modules that declare stream cells in configuration, by using builtin *-STREAM modules.
+   *
+   * <p>Steps: 1. Update the init rules of the stream cells. 2. Update rules that refer to 'stdin'
+   * stream. 3. Import rules from *-STREAM modules (with modification of cell names).
+   */
```
radumereuta, line 48: Mangled comments.

radumereuta (commented):
This is so large. I think I managed to go through half of it. I would like it if we check all the comments so they don't lose the important structures. Sometimes they really matter.
Code looks fine.

kernel/src/main/java/org/kframework/compile/CloseCells.java
Baltoli: Added automatically; the JavaDoc coding standards say that to separate paragraphs in the _output_ from the tool, you should add a `<p>` tag like this. It's basically a question of the difference between what we see here in the input (comment) and the output (JavaDoc HTML).

Baltoli:
I have addressed @radumereuta's comments re. formatting of comments, and I'm confident that any more such cases will be easy to fix. In the interests of not having such a large diff hanging around for ages, as planned I'm going to merge this now - if there are any issues with tooling etc. please let me know and we can work out a solution.

PR #3759, opened 2023-10-26
Add cachix binary cache publish step to release

goodlyrottenapple:
This change will push K to the k-framework-binary cachix cache and pin it to the commit hash. See https://github.com/runtimeverification/k/actions/runs/6664139255/job/18111232575 for a successful test run


.github/workflows/release.yml
```diff
@@ -43,17 +43,25 @@ jobs:
           cd ${CURDIR}
           gh release upload --repo runtimeverification/k --clobber v${version} ${tarball}
 
-  nix-release:
-    name: 'Nix Release'
-    runs-on: ubuntu-20.04
-    environment: production
```
goodlyrottenapple, line 49: This was only ever used for integration tests in the haskell-backend https://github.com/runtimeverification/haskell-backend/blob/858004428c6e9ff4a2e5733aee3b4d8909a3e13f/test.nix#L26-L36
but has long since been replaced with a different mechanism

.github/workflows/master-push.yml
```diff
@@ -5,58 +5,6 @@ on:
...
-        env:
-          GC_DONT_GC: 1
-          CACHIX_AUTH_TOKEN: '${{ secrets.CACHIX_PUBLIC_TOKEN }}'
-        with:
-          packages: jq
-          script: |
-            k=$(nix build .#k --json | jq -r '.[].outputs | to_entries[].value')
-            drv=$(nix-store --query --deriver ${k})
-            nix-store --query --requisites --include-outputs ${drv} | cachix push k-framework
-
```
goodlyrottenapple, line 59: this should not have been in master push. It has instead been replaced by the `cachix-release` job in `release.yml`

.github/workflows/release.yml
```diff
@@ -62,10 +70,24 @@ jobs:
...
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          packages: jq
+          script: |
+            export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
+            kup publish k-framework-binary .#k --keep-days 180
+            kup publish k-framework-binary .#k.openssl --keep-days 180
+            kup publish k-framework-binary .#k.procps --keep-days 180
+            kup publish k-framework-binary .#k.openssl.procps --keep-days 180
+            kup publish k-framework-binary .#k.procps.openssl --keep-days 180
```
Baltoli, line 90: This seems like a possible source of difficulty in the future; is there a reason these two permutations are needed? Shouldn't we normalise the list of extra libraries somehow?

goodlyrottenapple: We need both unless we normalise `kup install k.procps.openssl` into `kup install k.openssl.procps` by sorting. However, `#k.procps.openssl` and `#k.openssl.procps` evaluate to the same derivation anyway so we basically just tell cachix to pin both keys to the same package

goodlyrottenapple: note that in a different package these could potentially be non-commutative

Baltoli: Sure, that's all I wanted to check - I wasn't sure what the commutative properties were like here!

Baltoli (approved):
Looks good to me but will defer to @ehildenb for a final review

.github/workflows/release.yml
```diff
@@ -62,10 +70,24 @@ jobs:
...
+        uses: workflow/nix-shell-action@v3
+        env:
+          CACHIX_AUTH_TOKEN: '${{ secrets.CACHIX_PRIVATE_KFB_TOKEN }}'
+          GC_DONT_GC: '1'
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          packages: jq
+          script: |
+            export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
+            kup publish k-framework-binary .#k --keep-days 180
```
goodlyrottenapple, line 86: @ehildenb @Baltoli not sure how long we want to keep these pinned? They won't be GCd as long as they are pinned in the cache... ive chosen a somewhat arbitrary 180 days. Might even be too long?

Baltoli: I guess the limiting thing here is the maximum time between updates for a downstream project; if something were to go longer than 180 days without an update then is the downside risk just that their CI / build process will need to build K from source if going via Nix? (as opposed to _breaking_ a build). My gut feeling is that we should go for a shorter time, but that also depends on what the risk is on the other side - will we get charged extra by cachix? Will we fill up a disk somewhere more quickly?

(sorry for the lengthy answer, just trying to understand the tradeoff space here)

goodlyrottenapple: yes, kup will still install fine but will have to rely on the old cache and therefore will build for a bit longer. This can be mitigated in a few different ways. If the downstream CI is not ephemeral, then this should not matter anyway, otherwise we could either manually re-pin without the `--keep-days` argument which means that version will be pinned indefinitely, or we/they can set up another cachix cach and copy whatever they need into that cache...

dwightguth: Are we paying for cachix? I remember in the past there was an issue where we would run out of space and they would start to delete old stuff. Is that going to be an issue that interacts poorly with these changes?

.github/workflows/master-push.yml
ehildenb: @goodlyrottenapple we need to remove the `needs: ...` clause from later in the job under `gh-release`.

ehildenb (approved):

.github/workflows/release.yml
ehildenb: We are paying for cachix, yes. We have a plan that allows us quite a bit of storage across several repositories, @F-WRunTime has more details.

.github/workflows/master-push.yml
goodlyrottenapple: Ah thanks for the spot!

Issue #3731, opened 2023-10-19
Haskell backend exception breaks release build on Arch Linux


dwightguth:
The most recent K release build has failed: https://github.com/runtimeverification/k/actions/runs/6573891973/job/17858853532

### Symptoms

The "Build and Test Package" step of the K release on Arch Linux failed with the following error:

```
kore-exec: [800979013] Error (ErrorException):
    arithmetic underflow
Created bug report: kore-exec.tar.gz
Error:  krun: kore-exec ./lambda-kompiled/haskellDefinition.bin --module LAMBDA --pattern /tmp/.krun-2023-10-19-13-05-57-aaq4ojhJqy/tmp.in.oOrNWXIbzg --output /tmp/.krun-2023-10-19-13-05-57-aaq4ojhJqy/result.kore --searchType FINAL --search /tmp/.krun-2023-10-19-13-05-57-aaq4ojhJqy/tmp.pattern.YheUC7bp6U --smt none
Error:  krun: Backend crashed during rewriting with exit code 1
0a1,5
> {
>   T:Type
> #Equals
>   bool -> bool
> }
make[3]: *** [/usr/include/kframework/ktest.mak:103: tests/nested-lets.lambda] Error 1
```

This is in `k-distribution/pl-tutorial/1_k/5_types/lesson_5.5`.

### Investigation

Still digging into whether this issue can reproduce consistently, but the error in question might relate to the Data.Compact package, which we have had problems with before on certain platforms.

dwightguth:
This is evidently an intermittent issue, which might make it difficult to track down. Let's leave it until we at least can identify how frequently it occurs.

Baltoli:
We no longer support Arch Linux explicitly.

Issue #3723, opened 2023-10-18
Arch packaging step fails when setting up Docker


Baltoli:
### Symptoms

The Arch packaging step in the release CI job fails intermittently with an error like:
```
failed to prepare mspe7tld30tquuk7bg91h3uuw: invalid argument
```
when running the user and group creation step.

### Investigation

Maybe related?
* https://github.com/docker/buildx/issues/2021

Example runs:
* https://github.com/runtimeverification/k/actions/runs/6548843829/job/17784426991
* https://github.com/runtimeverification/k/actions/runs/6589987603/job/17905643103
* https://github.com/runtimeverification/k/actions/runs/6615629899/job/17968450307

### Fix

F-WRunTime:
Dropped Arch Linux No longer need this issue

Issue #3705, opened 2023-10-12
Homebrew Bottle testing step is flaky


Baltoli:
The "test brew bottle" step of the K release workflow is prone to flakiness because of a well-recognised issue with resource limits. An example failing run is [here](https://github.com/runtimeverification/k/actions/runs/6494720180/job/17640135384); note that the workflow has failed because of a number of "too many open files" errors.

We have a few options here, I think:
* A "devops" solution that increases the number of permitted open files on the machines we're running this on (e.g. https://stackoverflow.com/a/74117560/10121160). This may well not be possible on the hosted runners we use for this step.
* As a first cut, we could just reduce the concurrency of this step - at the moment the PL tutorial is being [run with the macOS equivalent of `-j$(nproc)`](https://github.com/runtimeverification/k/blob/master/.github/workflows/release.yml#L300-L302). Reducing the number of concurrent jobs will reduce the number of concurrent open files, in theory.
* ...however, this will increase the time taken to run what is _already_ a bottleneck for the release job. Do we need to be running the full PL tutorial as a smoke test[^1]? Is there a better subset of tests we could be using?
* Does finishing https://github.com/runtimeverification/k/pull/3674 help us? If we do a source build and test on macOS in CI before release, we can avoid running this full test suite at release time.
* Running this on our own hardware probably isn't a solution; it's not possible to buy new Intel macs any more[^2].

Fixing flaky release workflow steps is _really_ important for getting a reliable release process; the jobs are stateful enough that they can't be easily re-run, and so if a specific change needs to get into K for a downstream project then it's left waiting for _any other change_ to come along subsequently and get lucky with its runner.

[^1]: This does also apply to the Linux packaging runs, but they're not currently bottlenecks because they can use the stack-deps Docker image rather than having Homebrew build the backend's world from source every time.
[^2]: At some point this family of devices will get deprecated - how long do we want Intel macs to be our "first-class" CI and release type for macOS?

Baltoli:
This can also be a jumping-off point for a more general review of our release and packaging strategy; we have a great deal of complexity in our release process, but it's not necessarily clear _why_ we do lots of it beyond "it's always been the case". If we want to improve the downstream release process and get features rolling out reliably, improving the release build is an important step to take.

Baltoli:
Low-hanging fruit is to stop running tests without `--output-sync` passed to `make`; I've seen a couple of failures now in the PL tutorial where it's impossible to tell what's broken because the concurrent output is interleaved.

[...]

PR #3760, opened 2023-10-26
Temporary hotfixes for release flakiness

Baltoli:
This PR disables the Arch Linux packaging step, and reduces the parallelism being used by the macOS step. Doing so should address the current primary sources of flakiness in the K release job:
- https://github.com/runtimeverification/k/issues/3723
- https://github.com/runtimeverification/k/issues/3731
- https://github.com/runtimeverification/k/issues/3705

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -142,29 +142,36 @@ jobs:
           cp kframework_amd64_debian_bookworm.deb kframework_${version}_amd64_debian_bookworm.deb
           gh release upload --repo runtimeverification/k --clobber v${version} kframework_${version}_amd64_debian_bookworm.deb
 
-  arch:
-    name: 'Arch Linux Package'
-    runs-on: [self-hosted, linux, normal]
-    timeout-minutes: 90
-    steps:
-      - uses: actions/checkout@v3
-      - name: 'Build and Test Package'
-        uses: ./.github/actions/test-package
-        with:
-          dockerfile: package/arch/Dockerfile
-          os: archlinux
-          distro: base
-          build-package: package/arch/build-package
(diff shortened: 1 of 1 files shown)


Baltoli:
@ehildenb mentioned that some people (@nishantjr?) were using arch, but was in favour of removing.

Baltoli:
@goodlyrottenapple mentions that there is a `release.nix` job that _might_ be useless. We can remove it if no longer needed.

Baltoli:
Who is using Debian packaging? @dwightguth what is the historical reason we package for both Debian and Ubuntu?

.github/workflows/release.yml
```diff
@@ -142,29 +142,36 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.JENKINS_GITHUB_PAT }}
-        run: |
-          set -x
-          version=$(cat package/version)
-          cp kframework_arch_x86_64.pkg.tar.zst kframework_${version}_arch_x86_64.pkg.tar.zst
-          gh release upload --repo runtimeverification/k --clobber v${version} kframework_${version}_arch_x86_64.pkg.tar.zst
+  # The arch linux packaging step is disabled because of recurrent flakiness on
+  # the runners we use:
+  #   - https://github.com/runtimeverification/k/issues/3723
+  #   - https://github.com/runtimeverification/k/issues/3731
```
ehildenb, line 148: You need to also change the `release` step to not have `needs: [arch]`.

ehildenb (approved):

.github/workflows/release.yml
Baltoli: Good catch, thanks!

PR #3763, opened 2023-10-26
Drop PL tutorial in macOS release

Baltoli:
Reducing the parallelism applied to this job in #2937 does not fix the flakiness, so we should disable the offending part of the test suite for now so that a K release can go through properly. This workflow still smoke-tests K to ensure nothing is *obviously* broken.

Longer term, I think the solution is to run this and related jobs on our own M1/2 machines as we begin to deprecate Intel macOS.

Related to: #3705


Baltoli:
Closing in favour of a more permanent solution.

PR #3764, opened 2023-10-27
Move mac release CI to hosted ARM runners

Baltoli:
This PR moves our macOS packaging and release workflow over to use our self-hosted runners, which will hopefully address the recurrent flakiness we've had due to resource exhaustion (#3705).

I have tested these changes by adapting the release workflow to run on PRs, confirming that an `arm64_ventura` bottle gets successfully built. I don't anticipate any changes to the orchestration logic being required, but I will keep an eye on the release that this PR generates just to make sure that it doesn't break the release worse than it already is. In that event, this PR can easily be reverted or hotfixed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,7 +175,7 @@ jobs:
 
   macos-build:
     name: 'Build MacOS Package'
-    runs-on: macos-13
+    runs-on: self-macos-13
     timeout-minutes: 120
     environment: production
     needs: [set-release-id, source-tarball]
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -257,7 +257,7 @@ jobs:
 
   macos-test:
     name: 'Test MacOS Package'
-    runs-on: macos-13
+    runs-on: self-macos-13
```
goodlyrottenapple, line 260: shouldn't these be `self-macos-12` ?

Baltoli: image

We have an M2 runner on `macos-13`

goodlyrottenapple (approved):

PR #3786, opened 2023-11-02
Declare all Bool combinator equations simplifications (to disambiguate cases)

jberthold:
Rules like this pair
```
rule false andBool _:Bool => false
rule _:Bool andBool false => false
```
were previously in `BOOL-COMMON` and were said to demonstrate the implemented semantics of the hooks.
As `kore-rpc-booster` does not have its own hooks, the rules now come into effect themselves.
This causes issues because of the obvious overlap of rule matches (consider subject term `false andBool false` which matches both rule LHSs). 
This PR addresses the `kore-rpc-booster` issue by declaring the rules for `Bool` combinators as `simplification`s (whose application is optional - simplifications are trusted to be sound regardless of their (priority) order, whereas function equations with overlapping cases or uncertain matches lead to aborting evaluation in `kore-rpc-booster` to avoid unsoundness).

Also removed some trailing whitespace from the `domains.md` file.

### DIFF
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -303,7 +303,7 @@ effectively constant.
 
 You can remove the key/value pairs in a map that are present in another map in
 O(N*log(M)) time (where M is the size of the first map and N is the size of the
-second), or effectively linear. Note that only keys whose value is the same 
+second), or effectively linear. Note that only keys whose value is the same
 in both maps are removed. To remove all the keys in one map from another map,
 you can say `removeAll(M1, keys(M2))`.
 
@@ -489,7 +489,7 @@ module RANGEMAP
 
 ```k
   syntax Range ::= "[" KItem "," KItem ")"    [klabel(Rangemap:Range), symbol]
-  
+
   syntax RangeMap [hook(RANGEMAP.RangeMap)]
 ```
 
@@ -686,7 +686,7 @@ endmodule
 Sets
 ----
 
-Provided here is the syntax of an implementation of immutable, associative, 
+Provided here is the syntax of an implementation of immutable, associative,
 commutative sets of `KItem`. This type is hooked to an implementation of sets
 provided by the backend. For more information on matching on sets and allowable
 patterns for doing so, refer to K's
@@ -741,7 +741,7 @@ An element of a `Set` is constructed via the `SetItem` operator.
 
 You can compute the union of two sets in O(N*log(M)) time (Where N is the size
 of the smaller set). Note that the base of the logarithm is a relatively high
-number and thus the time is effectively linear. The union consists of all the 
+number and thus the time is effectively linear. The union consists of all the
 elements present in either set.
 
 ```k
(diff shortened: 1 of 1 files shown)


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1093,7 +1093,7 @@ You can:
 * Check inequality of two boolean values.
 
 Note that only `andThenBool` and `orElseBool` are short-circuiting. `andBool`
-and `orBool` may be short-circuited in concrete backends, but in symbolic 
+and `orBool` may be short-circuited in concrete backends, but in symbolic
 ackends, both arguments will be evaluated.
```
jberthold: As a side note, this statement about shortcut evaluation is probably not correct, given we have rules/simplifications like `rule _:Bool andBool false => false`  that focus only on the 2nd argument.

jberthold:
...not sure why but the build was re-triggered when I edited the PR description. The first build was green so this is presumably ready for review comments.

jberthold:
## Comparison of timings for `evm-semantics` tests
image
Notable speed-up for booster execution on some of the proofs, no dramatic slowdown on any of them, `kore-rpc` timings unchanged.

## Timings for foundry proofs (`kontrol` repo integration tests) unchanged
image
(EDIT: this close correspondence is suspicious, I re-ran the standard-K versions to confirm it)

k-distribution/include/kframework/builtin/domains.md
```suggestion
backends, both arguments will be evaluated.
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1115,43 +1116,42 @@ The remainder of this section consists of an implementation in K of the
...
-  rule _:Bool andBool false => false
-
-  rule true andThenBool K::Bool => K
-  rule K::Bool andThenBool true => K
-  rule false andThenBool _ => false
-  rule _ andThenBool false => false
+  rule true andBool B:Bool => B:Bool [simplification]
+  rule B:Bool andBool true => B:Bool [simplification]
+  rule false andBool _:Bool => false [simplification]
+  rule _:Bool andBool false => false [simplification]
```
ehildenb, line 1128: Why do we mark them _all_ as simplification rules? I guess this is so that we will always optimistically try them, allowing any of them to apply? My intuition would be to provide a set of complete rules for booleans (non simplification), then add simplification rules as well as needed.

So maybe we have a sepraate module with a set of complete and mutually exclusive rules for boolean (for concrete simplifications)?

```
rule true andBool B => B
rule false andBool _ => false
rule notBool true => false
rule notBool false => true
rule true orBool _ => true
rule false orBool B => B
....
```

Then have this separate module with _all_ the rules as simplificatino rules.

Maybe there should also be a way for the user to declare that this set of rules are _confluent_, so that the booster can just assume that?

jberthold: That would be possible, too. I don' thave a full set of measurements for it but I [tested the performance for a variant where all rules that focus on the 2nd argument were declared `[simplification]`](https://github.com/runtimeverification/k/pull/3786/commits/12d5ffd20fa199d571eb0fb555a1c94c22e95b12), leaving the other rules (the ones you quote above, focussing on the 1st arg) as they were.

OTOH all these rules become obsolete as soon as a backend with an implemented hook is used. Therefore I kept all rules together in one module and made them all `[simplification]` for homogenous appearance.
If the variant with both simplification rules and defining equations is desired, I can change the code.

jberthold:
Closed in favour of #3789

PR #3788, opened 2023-11-02
[DNM] Simplifying homebrew workflows

Baltoli:


Baltoli:
Still seems to be hitting resource exhaustion on the mac runners by using all the threads to run the test suite:
* Upstream changes from branch `core-factor` to allow the test runner parallelism to be configured
* Have the `tagged` branch of the Homebrew formula look up an env variable to check the loading factor and pass to the maven verify step

PR #3789, opened 2023-11-03
Avoid overlap of equations for Bool combinators simplifications

jberthold:
This is a variant of #3786 which turns all equations that focus on the 2nd argument of a `Bool` combinator as `simplification`s. In doing this, we retain a complete set of defining equations but avoid the problem described below.

Equations are kept in groups by combinator, rather than splitting them into defining and simplification equations in different modules.

-------------------
Rules like this pair
```
rule false andBool _:Bool => false
rule _:Bool andBool false => false
```
were previously in `BOOL-COMMON` and were said to demonstrate the implemented semantics of the hooks.
As `kore-rpc-booster` does not have its own hooks, the rules now come into effect themselves.
This causes issues because of the obvious overlap of rule matches (consider subject term `false andBool false` which matches both rule LHSs). 
This PR addresses the `kore-rpc-booster` issue by declaring some of the rules for `Bool` combinators as `simplification`s (whose application is optional - simplifications are trusted to be sound regardless of their (priority) order, whereas function equations with overlapping cases or uncertain matches lead to aborting evaluation in `kore-rpc-booster` to avoid unsoundness).

Also removed some trailing whitespace from the `domains.md` file.

### DIFF
--- a/k-distribution/include/kframework/builtin/domains.md
+++ b/k-distribution/include/kframework/builtin/domains.md
@@ -303,7 +303,7 @@ effectively constant.
 
 You can remove the key/value pairs in a map that are present in another map in
 O(N*log(M)) time (where M is the size of the first map and N is the size of the
-second), or effectively linear. Note that only keys whose value is the same 
+second), or effectively linear. Note that only keys whose value is the same
 in both maps are removed. To remove all the keys in one map from another map,
 you can say `removeAll(M1, keys(M2))`.
 
@@ -489,7 +489,7 @@ module RANGEMAP
 
 ```k
   syntax Range ::= "[" KItem "," KItem ")"    [klabel(Rangemap:Range), symbol]
-  
+
   syntax RangeMap [hook(RANGEMAP.RangeMap)]
 ```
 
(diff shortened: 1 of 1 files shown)


jberthold:
@ehildenb @Baltoli please choose and approve your preferred version and we close the resp. other PR.

ehildenb:
@jberthold did you have a chance to check the performance of this one on KEVM and Kontrol? This will make it so that the overlapping rules are applied in different phases, the normal equation phase or the simplification phase, so could have a weird effect on performance.

ehildenb (approved):

jberthold:
> @jberthold did you have a chance to check the performance of this one on KEVM and Kontrol? This will make it so that the overlapping rules are applied in different phases, the normal equation phase or the simplification phase, so could have a weird effect on performance.

I ran this version with `kore-rpc` (no significant changes) and with an experimental branch of `booster` before making the other variant a PR, and re-ran some of the tests with `booster` from master (speedup comparable to that of the other PR).

AFAICT `kore-rpc` does not apply any of those rules but instead runs the hooked low-level implementation. For `booster` the phase separation separates the overlapping equations which resolves the issue of indeterminate matching when the first argument is a variable.

jberthold:
I re-measured the `evm-semantics` proofs with `booster` and `K` from this branch - consistent results with what I expected.

timing

BTW these measurements are a bit noisy because of the re-compilation (I am getting a screenful of `z3` processes when I run with 4-fold parallelism), it would be great if we could avoid recompiling for every test run.

PR #3793, opened 2023-11-06
Maven plugin for using google-java-format

gtrepta:
This just adds a maven phase for running the `google-java-format` tool from the command line.

Can be invoked with `mvn process-sources`


Baltoli (approved):
I tested this locally and it seems to work nicely; @goodlyrottenapple's changes to the Maven build also look like they're working properly.

@gtrepta feel free to merge!

PR #3795, opened 2023-11-07
Fix version of mavenix

goodlyrottenapple:
Fix for a regression in kup binary packages, introduced when we switched back to a custom version of mavenix. This was done because we want to be able to manually specify the path to mavenix.lock (see https://github.com/nix-community/mavenix/commit/0cbd57b2494d52909b27f57d03580acc66bf0298). However, the next commit (https://github.com/nix-community/mavenix/commit/c2f3618ce413ad541027a163a85985760ce1ee75) is what causes the regression, by symlinking to many small nix store paths which take a very long time to download from the binary cache...


radumereuta:
@goodlyrottenapple you need to rebase your changes onto `develop` instead of master.

Issue #3715, opened 2023-10-13
`#let` desugaring into `#lambda` should be total (sometimes)


goodlyrottenapple:
Currently the front-end turns `#let <X> = <Y> in ... <X>...` into `#lambda__` expressions in kore, e.g:

```
rule [range-memUpdate-before]:
      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) =>
        #let W = minInt(W2, S1 -Int S2) #in
          #range(B1, S2, W) +Bytes #range(B1 [ S1 := B2 ], S1, W2 -Int W)
      requires 0 <=Int S2 andBool S2 <Int S1
      [simplification]
```
gets turned into:

```
rule
  #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) =>
    #lambda__3(minInt(W2, S1 -Int S2),B1,S2,B1,S1,B2,S1,W2) 
  requires 0 <=Int S2 andBool S2 <Int S1

rule 
  #lambda__3(W,B1,S2,B1,S1,B2,S1,W2) =>
    #range(B1,S2,W) +Bytes #range(B1 [ S1 := B2 ], S1, W2 -Int W)
```

In the case above where the LHS of a let is a variable, the corresponding `#lambda__3` should be marked as a total function, to allow simplification in the booster back-end.

Robertorosmaninho:
@dwightguth [resolveFun.java](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/ResolveFun.java) The modification should be in this file. The modification should be simple.

Baltoli:
@goodlyrottenapple how important is this for you to have implemented?

goodlyrottenapple:
Not super high priority as rules using `#let` can be re-written without it to work with booster

Baltoli:
Ah, this looks like it shouldn't be a huge change and it would be nice to avoid having to rewrite semantics code to deal with it. Will take a look at implementing it just now

PR #3798, opened 2023-11-08
Add `total` to `#let`-bindings where LHS is a variable

Baltoli:
This PR addresses an issue whereby the Booster reverts to the old Haskell backend when it encounters a rule written using `#let` bindings. The reason for this is that the de-sugared form of `#let` is expressed in terms of an intermediate `#lambda` function that is not marked as being `total`.

To address the issue, this PR adds the `total` attribute to the desugared intermediate lambda function for let-bindings and anonymous functions when the left-hand side of the original binding is a variable.

Points to consider when reviewing this PR:
* ~~Is the identification of when the LHS a variable correct in its handling of semantic casts and sorting? My presumption is that it is given that the "first child of `#SemanticCastTo` `KApply`" logic is replicated elsewhere.~~
  * https://github.com/runtimeverification/k/pull/3798#discussion_r1386805154
* Is there a way to integration-test this change other than "it doesn't break the rest of the test suite"? At the moment I'm not sure there is an obvious way to do so, but I'm open to suggestions.

Fixes https://github.com/runtimeverification/k/issues/3715


kernel/src/main/java/org/kframework/compile/ResolveFun.java
```diff
@@ -101,15 +106,15 @@ public K apply(KApply k) {
...
-          } else if (arg instanceof KApply
-              && ((KApply) arg).klabel().name().startsWith("#SemanticCastTo")
-              && ((KApply) arg).items().get(0) instanceof KVariable) {
-            nameHint1 = ((KVariable) ((KApply) arg).items().get(0)).name();
+          if (arg instanceof KVariable var) {
+            nameHint1 = var.name();
+          } else if (arg instanceof KApply app
+              && app.klabel().name().startsWith("#SemanticCastTo")
+              && app.items().get(0) instanceof KVariable) {
+            nameHint1 = ((KVariable) app.items().get(0)).name();
```
radumereuta: I don't think that's enough. You may have something like `(Var):Sort1` or `Var:Sort1:Sort2`
Weird but it can happen. The proper way to check if you have a variable under a specific term is with a visitor.
Maybe you can use something like this: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/compile/ResolveSemanticCasts.java#L104
and remove the code duplication from `isLHSTotal`.

Baltoli: Great, that's what I was looking for - will implement in terms of a visitor. Thanks Radu!

Baltoli: I have implemented the corrected behaviour in terms of a direct recursion rather than a visitor; the core logic is the same but I think the code is less noisy as a result. Happy to refactor if you'd prefer @radumereuta

Baltoli: (either way, the duplication is resolved)

radumereuta (approved):

PR #3802, opened 2023-11-09
Remove mavenix build

goodlyrottenapple:
Switch from using mavenix to using `buildMavenPackage` back-ported from the upcoming nixpkgs 23.11 stable release. This may increase build times somewhat and updating maven dependencies will no longer be automatic, but arguably the automation never works and It's difficult to figure out why it's failing.

I've updated the README troubleshooting section with the two cases of manual intervention, necessary with the new packaging solution, when maven dependencies change.


README.md
```diff
@@ -416,6 +416,23 @@ Common build-time error messages:
...
+                got:    sha256-fFlRqlLDZnVuoJniPvXjqdYEjnKxmFCEniavau/1gcQ=
+    error: 1 dependencies of derivation '/nix/store/79hazjbxp8829wpjvhh9c7kzc1m0ii22-k-6.1.0-dirty.drv' failed to build
+    ```
+    copy the `got:` hash (`sha256-fFlRqlLDZnVuoJniPvXjqdYEjnKxmFCEniavau/1gcQ=`) and replace it in `flake.nix`:
+    ```nix
+    k-framework = { haskell-backend-bins, llvm-kompile-libs }:
+        prev.callPackage ./nix/k.nix {
+            mvnHash = "sha256-fFlRqlLDZnVuoJniPvXjqdYEjnKxmFCEniavau/1gcQ=";
+            ...
+    ```
```
goodlyrottenapple: This will be the new "manual" way of updating the build whenever new maven dependencies get introduced.

goodlyrottenapple:
The `Nix: Maven` CI job will need to be removed from required checks by one of the admins.

Baltoli (approved):

F-WRunTime:
Checks updated, automerge added

dwightguth:
This is quite inconvenient that we don't have an automated solution for this. We can't be the only people building maven projects in Nix, surely? What do the maintainers of such packages in nixpkgs do?

PR #3804, opened 2023-11-09
make Pattern implement Comparable and also have an Ordering instance

dwightguth:
This is a change to Interface.scala in the K frontend repo that will be useful for facilitating the sorting of a list of commutative patterns in the Maude backend in order to implement a stable ordering for commutative terms.


Robertorosmaninho (changes requested):
The code LGTM, but could we have a small test for this on the k frontend? Just to avoid future regressions (before Maude is merged into k) that we'll only identify when you run the tests on the Maude backend. I wonder if the current ordering tests are enough to capture this; perhaps @gtrepta can confirm it!

gtrepta:
Yeah, the tests for sentence ordering might be a good example to follow.

https://github.com/runtimeverification/k/blob/47f589388860eeccc07262371f2b5fb552f84c59/kore/src/test/scala/org/kframework/definition/OuterTest.scala#L142

gtrepta (approved):
LGTM

Robertorosmaninho (approved):
LGTM

PR #3806, opened 2023-11-10
Remove mavenix take 2

goodlyrottenapple:
Switch from using mavenix to using `buildMavenPackage` back-ported from the upcoming nixpkgs 23.11 stable release. This may increase build times somewhat and updating maven dependencies will no longer be automatic, but arguably the automation never works and It's difficult to figure out why it's failing.

I've updated the README troubleshooting section with the two cases of manual intervention, necessary with the new packaging solution, when maven dependencies change.

The hash mismatch should only happen when deps change, but if we want to automate this in ci then we can do something similar to: https://github.com/runtimeverification/hs-backend-booster/blob/175f10b16ad50dd2bb50fcfe68cbcacd79760317/scripts/update-haskell-backend.sh#L21C3-L22


nix/build-maven-package.nix
```diff
@@ -0,0 +1,74 @@
...
+  fetchedMavenDeps = stdenv.mkDerivation ({
+    name = "${pname}-${version}-maven-deps";
+    inherit src sourceRoot patches;
+
+    nativeBuildInputs = [ maven ];
+
+    buildPhase = ''
+      runHook preBuild
+    '' + lib.optionalString buildOffline ''
+      mvn org.apache.maven.plugins:maven-dependency-plugin:3.6.1:go-offline -Dmaven.repo.local=$out/.m2 ${mvnDepsParameters}
```
goodlyrottenapple, line 20: using latest version of the go-offline plugin which does not build snapshot jars of the local project

nix/build-maven-package.nix
```diff
@@ -0,0 +1,74 @@
...
+    # keep only *.{pom,jar,sha1,nbm} and delete all ephemeral files with lastModified timestamps inside
+    installPhase = ''
+      runHook preInstall
+
+      find $out -type f \( \
+        -name \*.lastUpdated \
+        -o -name resolver-status.properties \
+        -o -name _remote.repositories \
+        -o -name \*.snapshots.xml \
+        -o -name \*.snapshots.xml.sha1 \) \
```
goodlyrottenapple, line 42: we still get `....snapshots.xml` files which contain dates of the latest snapshot, which we need to remove to make the build (hopefully) deterministic

Baltoli (approved):
Looks good but as discussed let's wait until a local change to a Java file lands in the master branch and gets merged into this branch.

nix/build-maven-package.nix
dwightguth: Um, you're not 100% requiring every single dependency to be pulled in by this, are you? go-offline has bugs that means it pulls in most of the dependencies, but not 100%. So if this is going to cause something to break if one or two dependencies isn't included in this, this solution isn't going to work.

goodlyrottenapple: No, there are indeed dependencies that the `go-offline` plugin does not detect. The missed ones have to be manually included here:
https://github.com/runtimeverification/k/blob/fc1859b723b985582d4f02566c5be7cec1352916/flake.nix#L66-L71

I've added a section in the README troubleshooting explaining this:
https://github.com/runtimeverification/k/blob/fc1859b723b985582d4f02566c5be7cec1352916/README.md?plain=1#L435-L436

PR #3807, opened 2023-11-11
Remove unused `strict` field from `ParseInModule`

Scott-Guest:
This PR removes the `strict` field in `ParseInModule`.

The field is safe to remove because, after eliminating some dead code,
- The only use of `strict` in `ParseInModule` is in `parseStringTerm`
- `strict` is (almost) always set to `true`
- The only place where `strict` is set to `false` is through a few calls to `RuleGrammarGenerator.getCombinedGrammar` in `KPrint`. However, these usages only ever call `getParsingModule` and `getExtensionModule` on the constructed `ParseInModule` object, neither of which invokes `parseStringTerm`, so the value of `strict` here is actually irrelevant.

Note that no actual functionality is lost either - the only live usage of `strict` in `parseStringTerm` is now fully controlled by the `inferSortChecks` argument rather than `strict && inferSortChecks`.


Baltoli (approved):
Nice one!

PR #3821, opened 2023-11-17
Add proof hint generation flag

Baltoli:
This flag exists in `llvm-kompile` already, but we want consumers that compile via the K frontend to be able to use the proof hint instrumentation feature.


theo25 (approved):
Looks good to me

gtrepta (changes requested):
Can we add a regression test that uses this flag, runs the interpreter with proof output, and just checks that the `HINT` header is present in the output?

gtrepta (approved):

PR #3823, opened 2023-11-20
Add `k-which-python` binary in nix

### LINKED ISSUE #92: Add option to print path to K's Python dependency
When using `kup`-installed K to run Poetry-based projects, we need to point the Poetry virtual environment at a version of Python that's been installed by Nix; pyenv and the system Python don't have the correct headers / include paths to work with the Nix C++ compiler that builds the Pythin binding modules.

We should implement a `kup` command that just prints out the path to the Python derivation that K depends on. For example:
```console
$ kup which-k-python
/nix/store/bc45k1n0pkrdkr3xa6w84w1xhkl1kkyp-python3-3.10.12/bin/python3
```

Then, projects that want to use `kup` for a Python-based semantics can run something like:
```console
$ poetry env use --no-cache $(kup which-k-python)
```
to make sure that their project uses the right version of Python.

Related:
* https://github.com/runtimeverification/pyk/issues/357
* https://github.com/runtimeverification/kup/issues/82

goodlyrottenapple:
Needed to implement a fix for https://github.com/runtimeverification/kup/issues/92


Baltoli:
Works for my original use case!

Baltoli (approved):

PR #3825, opened 2023-11-21
Move nailgun to lib/kframework/bin

gtrepta:
for #3675


gtrepta:
@goodlyrottenapple It looks like my change messed up nix somehow, I think it has to do with this snippet here

https://github.com/runtimeverification/k/blob/41dfa2474fea88b1313e8ac2b12c25dd3e722f9e/nix/k.nix#L106-L108

The error happens on CI [here](https://github.com/runtimeverification/k/actions/runs/6949674298/job/18908269582?pr=3825#step:8:646)

k-distribution/src/main/scripts/lib/setenv
```diff
@@ -9,7 +9,7 @@ else
 fi
 K_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
-export PATH="$K_LIB_DIR/../../bin:$PATH"
+export PATH="$K_LIB_DIR/../../bin:$K_LIB_DIR/bin:$PATH"
```
goodlyrottenapple: since you are including the lib/bin path, it might be better to just patch ng there, i.e.

```
preFixup = lib.optionalString (!stdenv.isDarwin) '' 
  patchelf --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" "$out/lib/bin/ng" 
''; 
```

dwightguth (commented):
Instead of changing the path in the setenv script, I would recommend just fully qualifying the path to the binary when it gets invoked.

goodlyrottenapple (approved):
LGTM

dwightguth (approved):

PR #3829, opened 2023-11-27
Add raw term production

Baltoli:
The [LLVM backend code that emits binary KORE terms into a proof trace](https://github.com/runtimeverification/llvm-backend/blob/6046ece44da6217b2cca99b3521fd20f26d5c107/runtime/util/ConfigurationSerializer.cpp#L441-L452) wraps terms that are not represented as symbols at runtime in an injection into `KItem` to allow them to be serialized using the code-generated recursion scheme used elsewhere in the backend.

However, doing so is not sound (see https://github.com/runtimeverification/llvm-backend/issues/904 for the full context). The additional injection changes the semantics of the serialized terms, and it is ambiguous whether client parser code should strip the injection from terms it reads from the trace.

The solution to this problem that I propose in this PR is to add a symbol to the K standard library whose only purpose is to signal "_the term under this symbol is an injection into `KItem` that should be stripped when parsing a trace_"; I have also implemented a small check that verifies user code does not attempt to write rules using this symbol.

Subsequent follow-up changes will be required to fully implement this solution:
* Change to the LLVM backend proof trace code to emit `#rawTerm` instead of injections, along with a version bump for the format.
* Updates / additions to the Pyk bindings to optionally detect and strip these terms when deserializing binary KORE to patterns.

There may be a better way to implement something equivalent to these changes; I'm open to suggestions!


dwightguth (commented):
I am confused by this. What exactly is going  on here?

Baltoli:
> I am confused by this. What exactly is going on here?

https://github.com/runtimeverification/llvm-backend/issues/904 is the surrounding context; this isn't quite ready for a review yet. Open to suggestions for alternative implementations!

Baltoli:
Closing after discussing with @dwightguth; a better solution is to have the backend insert a symbol automatically as a preprocessing step.

PR #3816, opened 2023-11-15
Address IntelliJ warnings in `Interface.scala`

Scott-Guest:
This PR addresses all the IntelliJ warnings in `Interface.scala`.

In particular, we add type annotations to `Sort.ord` and `Pattern.ord`. This fixes an issue noted by @radumereuta where IntelliJ couldn't properly resolving the implicit `Ordering` instances, causing erroneous compilation errors and blocking use of the debugger.

PR #3840, opened 2023-12-04
Remove `EAsSet[E]` and `WAsSet[W]` traits in `Transformers.scala` to address IntelliJ error

Scott-Guest:
This PR removes the `EAsSet[E]` and `WAsSet[W]` traits in `Transformers.scala`, and instead just duplicates a few lines of code between `SetsGeneralTransformer` and `SetsTransformerWithErrors`.

As a result, we fix an annoying issue where IntelliJ incorrectly reports an error on every inheritor of these classes, falsely claiming that `errorUnit()` has an incorrect return type.

Given that the cost here is low, and similar issues have bitten us in the past by blocking use of the debugger (#3816), I think this is a worthwhile change. If no one is opposed, I plan to address any similar incorrectly-reported IntelliJ errors as I encounter them.

### DIFF
--- a/kore/src/main/scala/org/kframework/parser/Transformer.scala
+++ b/kore/src/main/scala/org/kframework/parser/Transformer.scala
@@ -118,7 +118,8 @@ abstract class GeneralTransformer[E, W] extends ChildrenMapping[E, W] {
   def apply(c: Constant): (Either[E, Term], W) = simpleResult(c)
 }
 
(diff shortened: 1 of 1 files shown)


Baltoli (approved):
Looks good; strongly pro making this kind of incremental change as we work on the code. Thanks Scott!

radumereuta (approved):

PR #3841, opened 2023-12-04
Adopt scalafmt for Scala code style

### LINKED ISSUE #3827: Adopt Scala auto-formatter
Similarly to #3757, we should consider adopting a Scala auto-formatter.

[scalafmt](https://github.com/scalameta/scalafmt) seems to be a popular choice and is easily integrated with IntelliJ. The default settings look fine for me, perhaps setting `maxColumn = 100` to align with the Java formatter.

Scott-Guest:
Closes #3827.

This PR adopts [scalafmt](https://scalameta.org/scalafmt/) for Scala code formatting. 

Specifically,
- Both `google-java-format` and `scalafmt` are now handled by the `spotless` Maven plugin
    - Run `mvn spotless:check` to check formatting (or `mvn verify`)
    - Run `mvn spotless:apply` to auto-format all source files (or `mvn process-sources` or `mvn package`)
- CI runs `mvn spotless:check` to enforce formatting

Intellij has built-in support for `scalafmt`. Enable it by opening `Settings > Editor > Code Style > Scala` and setting
- Formatter: `Scalafmt`
- Configuration: `k/src/main/config/.scalafmt.conf`

The documentation on the different configuration options can be found here: https://scalameta.org/scalafmt/docs/configuration.html.

Currently, I've formatted with my preferred settings, but committed two versions:
- [506616c](https://github.com/runtimeverification/k/pull/3841/commits/506616cb75c8378503393b5503ce0a15d2d3c017), which does not enforce vertical alignment
- [602fd2e](https://github.com/runtimeverification/k/pull/3841/commits/602fd2e47ae4fbd97b1be4430a48428fd76345ff), which does enforce vertical alignment

I'd appreciate if everyone who works on the frontend could skim through the files and let me know your thoughts!


Baltoli (approved):
I've skimmed through and this seems good to me; the pros of having a consistently-enforced Scala style outweigh any possible cons from a particular piece of code looking suboptimal.

I'd register a _weak_ preference **in favour** of column alignment.

k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala
```diff
@@ -77,47 +92,65 @@ class KoreTest {
...
+          Some(eq)
+        case Implies(
+              _,
+              And(_, _ +: Top(_) +: Seq()),
+              And(_, (eq @ Equals(_, _, Application(_, _), _)) +: _ +: Seq())
+            ) =>
+          Some(eq)
         case eq @ Equals(_, _, Application(_, _), _) => Some(eq)
-        case _ => None
+        case _                                       => None
```
radumereuta, line 128: This looks weird. Probably because it's trying to align with the previous line.

k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala
```diff
@@ -77,47 +92,65 @@ class KoreTest {
...
+      case Implies(_, p1, p2)   => symbols(p1) ++ symbols(p2)
+      case Mem(_, _, p1, p2)    => symbols(p1) ++ symbols(p2)
 //      case Next(_, p) => symbols(p)
-      case Not(_, p) => symbols(p)
-      case Or(_, ps) => ps.flatMap(symbols)
+      case Not(_, p)           => symbols(p)
+      case Or(_, ps)           => ps.flatMap(symbols)
       case Rewrites(_, p1, p2) => symbols(p1) ++ symbols(p2)
-      case _ => Seq()
+      case _                   => Seq()
```
radumereuta, line 153: I take it back. This looks nice.

kore/src/main/scala/org/kframework/POSet.scala
```diff
@@ -1,56 +1,61 @@
...
   /**
-   * Internal private method. Computes the transitive closer of the initial relations.
-   * It also checks for cycles during construction and throws an exception if it finds any.
+   * Internal private method. Computes the transitive closer of the initial relations. It also
+   * checks for cycles during construction and throws an exception if it finds any.
    *
-   * The implementation is simple. It links each element to the successors of its successors.
-   * TODO: there may be a more efficient algorithm (low priority)
+   * The implementation is simple. It links each element to the successors of its successors. TODO:
+   * there may be a more efficient algorithm (low priority)
```
radumereuta: This doesn't look right. Can you add a new line so it doesn't move the TODO at the end?

kore/src/main/scala/org/kframework/parser/kore/Interface.scala
```diff
@@ -314,7 +317,8 @@ object Iff {
 }
 
 trait Exists extends Pattern {
-  def s: Sort // this is the sort of the whole exists pattern, not the sort of the binding variable v
+  def s
+      : Sort // this is the sort of the whole exists pattern, not the sort of the binding variable v
```
radumereuta: Weird alignment. Maybe if you move the comment on the line above the type annotation will be kept on the same line as the declaration.

radumereuta (approved):
A few nitpicks regarding comments.
I'm in favor of code alignment.
lgtm otherwise.

k-distribution/src/test/scala/org/kframework/backend/kore/KoreTest.scala
Robertorosmaninho: Indeed, it is much more readable!

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -15,31 +16,49 @@ trait Rewriter {
...
   /**
-   * (disregard this javadoc comment for now)
-   * Takes one rewriting step.
-   * - for regular execution, it returns the next K or False (i.e. an empty Or)
-   * - for symbolic execution, it can return any formula with symbolic constraints
-   * - for search, it returns an Or with multiple ground terms as children
+   * (disregard this javadoc comment for now) Takes one rewriting step.
+   *   - for regular execution, it returns the next K or False (i.e. an empty Or)
+   *   - for symbolic execution, it can return any formula with symbolic constraints
+   *   - for search, it returns an Or with multiple ground terms as children
```
Robertorosmaninho: Perhaps we should look for all the comments to see where we can add a new line to keep the semantic of separate ideas.

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -15,31 +16,49 @@ trait Rewriter {
...
+   *   The configuration to begin searching from.
+   * @param depth
+   *   No. of transitions to consider before termination (Depth of Tree to traverse). Empty
+   *   represents unbounded.
+   * @param bound
+   *   No. of states to consider as final results. Empty represents unbounded.
+   * @param pattern
+   *   The rule (pattern + side condition) that we're trying to find a substitution for.
+   * @return
+   *   A list of substitutions, denoting all the configurations matching the given rule.
```
Robertorosmaninho, line 42: I don't want to be pedantic, but for me, the comments should be above the definition they're referring to, so maybe moving them above would look nicer!

kore/src/main/scala/org/kframework/unparser/Unparse.scala
```diff
@@ -18,64 +19,65 @@ object ToKast {
...
-   * <li>Precedence level 0 is for the top of a term or within an argument list.
-   * <li>Precedence level 1 is for arguments of a KRewrite.
-   * <li>Precedence level 2 is for entries in a KSequence,
-   * </ul>
-   * The only case where braces may be needed is around a KSequence which is
-   * an argument of a KRewrite.
+   * <ul> <li>Precedence level 0 is for the top of a term or within an argument list. <li>Precedence
+   * level 1 is for arguments of a KRewrite. <li>Precedence level 2 is for entries in a KSequence,
+   * </ul> The only case where braces may be needed is around a KSequence which is an argument of a
+   * KRewrite.
```
Robertorosmaninho: Needs to be reworked here as well, we completely  lost the readability we had before

Robertorosmaninho (changes requested):
This looks really nice! Thanks for your work @Scott-Guest!

I would advocate for the column align as well!

I made some comments about comments😅!
I'm not sure if we want to enforce the column alignment on them as well, but if we want to, we need to make sure they look as readable as they were before and that their meaning is the same! 

I'll request a change at least for the `kore/src/main/scala/org/kframework/unparser/Unparse.scala` if you don't mind!

Baltoli:
@dwightguth queries editor integrations - other than IntelliJ. ~~Will be inconvenient if no Vim~~ there's a CLI version

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
Scott-Guest: Unfortunately, the only configuration we have here is
- `docstrings.wrap = yes`, which ensures docstring comments fit in `maxColumn` by fully reformatting them as above
- `docstrings.wrap = no`, which allows docstrings to be arbitrarily long and left unformatted

IMO we should keep everything under the `maxColumn`, so I don't think it's worth addressing this

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
Scott-Guest: Adding `<p>` tags should fix this, I'll make the changes

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
Robertorosmaninho: Got it! That is unfortunate indeed! Could we manually modify it to be above the definition in this case and check if it happens again somewhere? I think this is the most evident and significant case, but I can also be wrong!

Scott-Guest: On second thought, I'm actually strongly opposed to moving the comments above.

These are ScalaDoc comments which generate nice static web pages of the documentation. 

The ScalaDoc parser treats anything following `@param paramName` as the documentation for the parameter, so manually moving the description above will just lead to the generated ScalaDoc page being off-by-one for each parameter.

That is, manually moving it above like
```
/**
 * This is a function
 * 
 * This is the first argument
 * @param myArg1
 * 
 * This is the second argument
 * @param myArg2
 */
```
will generate the ill-formatted ScalaDoc
```
This is a function

This is the first argument

myArg1 - This is the second argument

myArg2 -
```

Robertorosmaninho: Got it! It isn't much readable on the code, but it wouldn't make sense indeed to the generated Scala Doc

Robertorosmaninho (approved):
LGTM!

ehildenb (approved):

Robertorosmaninho:
@Scott-Guest, do you know why the `Check Java Code Formatting` isn't being executed for this PR?

Baltoli:
> @Scott-Guest, do you know why the `Check Java Code Formatting` isn't being executed for this PR?

An admin (@ehildenb, @dwightguth, @F-WRunTime) will need to amend the branch protection rules

gtrepta (approved):
This looks good to me regarding the scala code itself. I'm bearish on formatting comments, I think it's outside the scope of what can be expected from a formatter and there can be specific ways a coder would want to format their comments that a formatter might contend against. Those are my two cents, it sounds like everyone else is fine with it.

Baltoli:
> and there can be specific ways a coder would want to format their comments that a formatter might contend against

Yeah, I agree with this basically @gtrepta and I think it's a reasonable concern - for the formatters we've chosen:
- I believe google-java-format is much stricter about formatting _JavaDoc_ comments (`/**`), but will leave non-JavaDoc (`/*`) comments much more untouched. There are also escape hatches like `<pre>` for things like diagrams in JavaDocs.
- There are loads more knobs to turn in Scalafmt; if it's really consistently mangling comments it's not a huge imposition to try and find a slightly different configuration to improve the situation.

Scott-Guest:
@gtrepta @Baltoli If need be, we can also configure spotless to disable formatting of certain sections using `// spotless:on` and `// spotless:off` 

This is done at a higher level than the formatters (removing the section of code out of the file, running the formatter, then re-inserting it), so it works even with google-java-format which doesn't natively have this ability.

PR #3855, opened 2023-12-07
Enable new TI in kprove

radumereuta:
Move the option to enable new type inference to InnerParsingOptions so it's available in kprove as well.


Scott-Guest (approved):
LGTM

radumereuta:
@Scott-Guest @Baltoli can you have another look?
I added a few more changes.

Issue #3849, opened 2023-12-06
Check that sort parameters can be instantiated during type inference


Scott-Guest:
In `AddSortInjections`, we check that every sort parameter can actually be instantiated with the least upper bound of its required types:

https://github.com/runtimeverification/k/blob/d95999edbfde902a998d2bf56aa4883307d8f347/kernel/src/main/java/org/kframework/compile/AddSortInjections.java#L538

We should move this check into type inference itself, that way we can report an error on each rule, rather than crashing at the first error we hit.

Scott-Guest:
Holding off on this until the Z3-based algorithm is removed. Blocked on #3848

Issue #3848, opened 2023-12-06
Support all terms in the SimpleSub sort inferencer


Scott-Guest:
Currently, the SimpleSub sort inferencer falls back to the Z3-based algorithm for terms containing:

- [x] strict casts

- [ ] ambiguities

- [ ] parametric sorts / MInt

We should address each of these limitations and eventually remove the Z3-based algorithm.

Scott-Guest:
### MInt 
`MInt` fits neatly into my general plan for parametric sorts, namely, by
- Introduce a super-sort `Nat` of the sorts `1, 2, ...`
- Define `syntax MInt{+S}`, where the `+` indicates a covariant type parameter
- Replace every negative usage of `MInt{S}` with `MInt{S ∧ Nat}`, e.g.,
    ```
    syntax {Width} MInt{Width} ::= MInt{Width ∧ Nat} "*MInt" MInt{Width ∧ Nat}
    ```

PR #3870, opened 2023-12-12
Support strict casts in the SimpleSub inferencer

Scott-Guest:
Part of #3848. 

- Add support for strict casts in the SimpleSub-based sort inference algorithm
- Improve the error message for `CHECKED` mode
- Add missing constraints when handling anywhere rules
    - Not strictly required at this point, but ensures that every parameter is inferred as will be needed for #3849.


Scott-Guest:
~Blocked on #3877~

kernel/src/main/java/org/kframework/parser/inner/ParseInModule.java
```diff
@@ -445,7 +445,7 @@ private Tuple2<Either<Set<KEMException>, Term>, Set<KEMException>> parseStringTe
           boolean equalRight =
               rez.isRight() && z3Rez.isRight() && rez.right().get().equals(z3Rez.right().get());
           if (!(bothLeft || equalRight)) {
-            throw KEMException.criticalError("Z3 and SimpleSub algorithms differ!");
+            throw typeInferenceCheckError(rez3, z3Rez, rez);
```
radumereuta, line 448: Do you have any examples of such differences?
It would be interesting to document them and add regression tests.

Scott-Guest: I'm not aware of any actual differences - the improved error message was just useful while debugging when I unintentionally broke something and caused a difference

radumereuta (approved):

Baltoli (approved):

PR #3874, opened 2023-12-12
Migrate from Checkstyle to Spotless

Scott-Guest:
This PR migrates all of our formatting checks from the Checkstyle plugin to the Spotless plugin. 

As a result, `mvn package` (or `mvn spotless:apply` or `mvn process-sources`) automatically fixes any formatting violations, e.g., inserting copyright headers for you.

***The diff is large, but the only non-automated code changes are in `pom.xml` and `src/main/config/checkstyle-*.xml`***

All the behavior is the exact same except:
- Previously, copyright headers could match the regex `Copyright \(c\) (Runtime Verification, Inc|K Team). All Rights Reserved.`, but now we only accept the `Runtime Verification, Inc` version due to limitations of Spotless. All existing headers were converted automatically
   - If there is an issue with this legally, I believe there is a mechanism to avoid changing the copyright of any existing files, but all new files would still require `Runtime Verification, Inc` 
- We more broadly enforce spaces over tabs, unix line endings, newlines at the end of each file, and no trailing whitespace. Previously, this was only enforced for K and Java files.


docs/developers/sort_inference.md
```diff
@@ -50,8 +50,8 @@ Informally then, the polarity restriction enforces that type intersections can o
 - `𝜏 <: 𝜏1 and 𝜏 <: 𝜏2 iff 𝜏 <: 𝜏1 ⊓ 𝜏2`
 - `𝜏1 <: 𝜏 and 𝜏2 <: 𝜏 iff 𝜏1 ⊔ 𝜏2 <: 𝜏`
 
-In total then, any type variable with bounds `L <: 𝛼 <: U` can be encoded as a set-theoretic type by 
-- replacing every negative instance of `𝛼` with `𝛼 ⊓ U` 
+In total then, any type variable with bounds `L <: 𝛼 <: U` can be encoded as a set-theoretic type by
+- replacing every negative instance of `𝛼` with `𝛼 ⊓ U`
```
radumereuta: We probably don't want to enforce trailing whitespace in Markdown since it has a semantic meaning:
https://www.markdownguide.org/basic-syntax/#line-break-best-practices

k-distribution/pl-tutorial/2_languages/1_simple/2_typed/1_static/simple-typed-static.md
```diff
@@ -1,13 +1,13 @@
...
 ---
 
 # SIMPLE — Typed — Static
 
-Author: Grigore Roșu (grosu@illinois.edu)  
+Author: Grigore Roșu (grosu@illinois.edu)
 Organization: University of Illinois at Urbana-Champaign
 
-Author: Traian Florin Șerbănuță (traian.serbanuta@unibuc.ro)  
+Author: Traian Florin Șerbănuță (traian.serbanuta@unibuc.ro)
```
radumereuta: This is intentional. You can check how the rendering looks.
Screenshot from 2023-12-15 18-18-13
Screenshot from 2023-12-15 18-18-32

radumereuta (changes requested):

docs/developers/sort_inference.md
Scott-Guest: Good catch - looks like tabs are also significant (https://www.markdownguide.org/hacks/#indent-tab), so I'll remove the tabs-to-spaces as well. 

Rewrote commit history and yanked those changes out.

radumereuta (approved):
lgtm now, although very large.
Maybe someone else can have a look. If not, you can merge in a few days.

Baltoli (approved):
Looks good to me also (once my laptop had loaded the diff page!)

PR #3877, opened 2023-12-13
Remove KLabel variables

Scott-Guest:
This PR removes all logic related to KLabel variables because
- They are unsupported by any modern backend
- Without KLabel variables, we can assume every variable has sort `S` with `KBott < S <= K`, which addresses some minor annoyances in the sort inferencer.


kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -24,25 +25,30 @@
...
+ *       byte "\x05".
+ *   <li>InjectedKLabel: The byte "\x06" followed by the representation of the klabel.
+ *   <li>KLabel: The representation of the string of the klabel, followed by the byte "\x01" if the
+ *       klabel is a variable, and "\x00" if it's a concrete klabel.
+ *   <li>String: A 4-byte offset in the string intern table. The intern table is computed as the
+ *       term is traversed. An offset of 0 means that the string has not appeared before in this
+ *       term, and is followed by a 4-byte length of the string followed by the String in UTF-16. An
+ *       offset of 1 means the string refers to the most recent previous string in the intern table.
+ *       An offset of 2 means the next-most-recent, and so forth.
+ * </ul>
```
Scott-Guest, line 51: This is just a formatting fix - noticed this got mangled by the auto-formatter.

radumereuta (approved):

Issue #1936, opened 2021-04-23
Remove --transition flag


ehildenb:
This option is only used on the Java backend in very specific circumstances with `search`. It is very counterintuitive to use when `search`ing, because you would expect _all_ rules to be enabled as `transition` by default, but instead this forces you to list the relevant rules.

It requires updating the pl-tutorial to remove mentions of this option.

kheradmand:
@ehildenb Is there a way to achieve same functionality in the Haskell backend? I.e. is there an easy way to say that some rules are not transitions? Say, in the following setting, we are only interested in checking the interleaving of global writes and do not want to check local-local or local-global write interleaving. 

```
configuration 
	<global> .Map </global>
	<threads>
		<thread multiplicity=“+” type=“Set”>
			<k> .K </k>
			<local> .Mem </local>
		</thread>
	</threads>


rule <k> write_local(k,v) => .K … </k>
       <local> M => M[k <- v] </local>

rule <k> write_global(k,v) => .K … </k>
       <global> M => M[k <- v] </global>
```

I think we could do this easily with transition labels. Is is there an easy way to do this with the haskell backend?

ehildenb:
We need to remove Java backend from the rest of the tutorial. Related: https://github.com/runtimeverification/k/issues/1271

ehildenb:
Subsumed: https://github.com/runtimeverification/k/issues/3167

Scott-Guest:
Re-opening as there are still references to --transition throughout the PL tutorials and in the frontend code

PR #3897, opened 2024-01-08
Cleanup any mention of --transition flag

gtrepta:
fixes: #1936


Scott-Guest:
I textually searched the codebase for `transition`, and there are a few lingering references missed by this PR:
- https://github.com/runtimeverification/k/blob/9fa1bf94cc86899b8f50cd47c7b477600149c0f7/k-distribution/pl-tutorial/1_k/4_imp%2B%2B/README.md?plain=1#L17
- https://github.com/runtimeverification/k/blob/9fa1bf94cc86899b8f50cd47c7b477600149c0f7/k-distribution/pl-tutorial/1_k/4_imp%2B%2B/lesson_8/imp.md?plain=1#L75
- https://github.com/runtimeverification/k/blob/9fa1bf94cc86899b8f50cd47c7b477600149c0f7/k-distribution/pl-tutorial/2_languages/1_simple/1_untyped/simple-untyped.md?plain=1#L1072
- https://github.com/runtimeverification/k/blob/9fa1bf94cc86899b8f50cd47c7b477600149c0f7/k-distribution/pl-tutorial/2_languages/1_simple/1_untyped/simple-untyped.md?plain=1#L1164

k-distribution/pl-tutorial/1_k/4_imp++/lesson_3/README.md
```diff
@@ -31,99 +31,6 @@ the `krun` tool with the option `--search`. For example:
```
Scott-Guest, line 31: We've removed almost everything from this lesson - the title is "Tagging; Transition Kompilation Option", and we no longer explain tagging nor `--transition`. We should consider just deleting it entirely.

The only remaining content is pretty short and could be easily moved into other lessons:
- defining the variable increment rule
- describing `--search`

k-distribution/pl-tutorial/1_k/4_imp++/lesson_8/imp.md
```diff
@@ -563,107 +563,3 @@ This rule acts like a ``tail recursion'' optimization, but for blocks. */
...
-model. Depending upon for what you want to use the generated language model,
-you may need to kompile the definition using various options. We here discuss
-these options.
-
-To tell the **K** tool to exhaustively explore all the behaviors due to the
-non-determinism of addition, division, and threads, we have to kompile
-with the command:
-
-```shell script
-kompile imp.k --transition="addition division lookup assignment increment read print"
```
Scott-Guest, line 579: Most of these `addition`, `division`, `lookup`, etc. tags are unneeded now, so we should remove the corresponding `group(_)` attributes on the rules. Similarly in the other lessons too.

k-distribution/pl-tutorial/1_k/4_imp++/lesson_3/README.md
```diff
@@ -31,99 +31,6 @@ the `krun` tool with the option `--search`. For example:
...
-
-Here is what happens. `krun` can only explore as much of the transition
-system associated to a program as `kompile` allowed the generated language
-model to yield. Since most of the K users are interested in language models
-that execute efficiently, that is, in faster interpreters for the defined
-languages, by default `kompile` optimizes the generated language model for
-execution. In particular, it inserts no backtracking markers, which `krun`
-uses when called with the `--search` option in order to systematically generate
-the entire transition system associated to a program. This is why `krun`
-showed us only one solution when run with the `--search` option on `div.imp`.
```
Scott-Guest, line 44: Does `--search` still only return one solution here, or does it actually show all the different interleavings nowadays?

If it only returns one solution, I think we still need to include a good chunk of the prose here to explain why the other solutions are missing.

Scott-Guest (commented):
Mostly LGTM, but see my comments.

One thing that's still unclear to me - the outdated lessons state that `--search` will not explore all heating / cooling interleavings unless specifically instructed to do so with `--transition`. Is that still the case, and is there some modern alternative to `--transition`? Or does `--search` explore all interleavings by default nowadays?

k-distribution/pl-tutorial/1_k/4_imp++/lesson_3/README.md
gtrepta: I tested it and indeed `--search` returns all solutions now. I believe this was written before the search feature in the llvm backend was a thing.

Baltoli (approved):
This looks good; happy to merge with Scott's comments addressed given that there is ongoing work to slice out and review the PL tutorial anyway so any snags there aren't critical.

gtrepta:
Closing as it's overridden by #3901 . Will open this in the pl-tutorial repo.

Issue #3896, opened 2024-01-08
[K-Bug] `replace` equation in `STRING-COMMON` module is missing a case


sskeirik:
### What component is the issue in?

Front-End

### Which command

- [ ] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

All

### Operating System

Linux

### Steps to Reproduce

In the `replace` equation in the `STRING-COMMON` module linked below:

https://github.com/runtimeverification/k/blob/001ffdc18ef5f3297ad7640b3163045adcc46a3c/k-distribution/include/kframework/builtin/domains.md?plain=1#L1884-L1887

The result of the term `replace(Source, ToReplace, Replacement, Count)` is undefined whenever the following conditions are met:

1.  `Count > 0`
2.  `ToReplace` does not appear in `Source`

In that case, `findString(Source, ToReplace, 0)` will return `-1` which will cause the subterm to evaluate as follows:

```
substrString(Source, 0, findString(Source, ToReplace, 0))

=>

substrString(Source, 0, -1)
```

which is undefined.

The reason why this bug hasn't manifested previously is that most backends provide native versions of all `String` operations --- and so these equations aren't used.

In general, these equations will only be used for backends which do not provide native versions of some `String` operation --- otherwise, they exist to document the expected behavior of hooked functions.

Still, it would be good to correct this, at least for documentation purposes.

### Expected Results

It would be ideal if `replace(Source, ToReplace, Replacement, I)` with `I > 0` returned `Source` whenever `ToReplace` does not occur in `Source`.

To achieve this, this equation for `replace` can be split into two cases:

1.  If `ToReplace` occurs in `Source` --- then the behavior is identical to the above
2.  If `ToReplace` does not occur in `Source` --- then return `Source`

PR #3910, opened 2024-01-18
Fix missing case in semantics for string replace function

gtrepta:
fixes: #3896


k-distribution/include/kframework/builtin/domains.md
```diff
@@ -1881,11 +1881,15 @@ of the above operations in K.
...
+  // implement their own behavior without contradicting these semantics. For instance, a symbolic
+  // backend can return #Bottom for that case, while a concrete backend can throw an exception.
   rule replace(Source:String, ToReplace:String, Replacement:String, Count:Int) =>
        substrString(Source, 0, findString(Source, ToReplace, 0)) +String Replacement +String
        replace(substrString(Source, findString(Source, ToReplace, 0) +Int lengthString(ToReplace), lengthString(Source)), ToReplace, Replacement, Count -Int 1)
-        requires Count >Int 0
-  rule replace(Source:String, _, _, 0) => Source
+        requires Count >Int 0 andBool findString(Source, ToReplace, 0) >=Int 0
+  rule replace(Source:String, _, _, Count) => Source
+        requires Count >=Int 0 [owise]
```
sskeirik, line 1892:
```suggestion
  rule replace(Source:String, _, _, 0) => Source [owise]
```

For the owise rule, the variant above will have proper behavior I think.

The version with the constraint `requires Count >= 0 [owise]` appears to make `replace` not uniquely defined because the term:

`replace(Source, ToReplace, Replacement, 1)`

Can evaluate into one of two possibilities:

1.   if `findString(Source, ToReplace, 0) >=Int 0`, then it can evaluate into:

     ```
     substrString(Source, 0, findString(Source, ToReplace, 0)) +String Replacement +String
       replace(substrString(Source, findString(Source, ToReplace, 0) +Int lengthString(ToReplace), lengthString(Source)), ToReplace, Replacement, Count -Int 1)
     ```
     
2.   also since `Count >=Int 0`, it can evaluate into:

     ```
     Source
     ```
     
If we're all on the same page, then I'm happy to approve after the patch is applied.

gtrepta: Because it's an owise rule though, it should always evaluate to case 1, unless the findString condition doesn't hold, then it will evaluate to case 2.

What should `replace` do if `Count >Int 0` and `findString(...) ==Int -1`? I believe your proposed solution makes it undefined.

sskeirik: Doh! You're absolutely right! Thanks for fact-checking me.

sskeirik (approved):
Looks good!

Baltoli (approved):
Looks good to me, and it seems like you're both in agreement here!

Issue #3880, opened 2023-12-15
[K-Bug] Incorrect parsing and crash with ambiguous contextual function


Scott-Guest:
### What component is the issue in?

Front-End

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v6.0.137

### Operating System

Linux

### K Definitions (If Possible)

`test-exp1.k`
<pre>
module TEST-EXP1
imports INT
syntax Exp1 ::= foo() [function]
              | Exp1Val()

syntax Exp2 ::= foo() [function]
              | Exp1 

configuration &lt;k&gt; $PGM:Exp2 &lt;/k&gt;
              &lt;val&gt; Exp1Val():<b>Exp1</b> &lt;/val&gt;

rule [[ foo() => VAL ]]
     &lt;val&gt; VAL &lt;/val&gt;
endmodule
</pre>

`test-exp2.k`
<pre>
module TEST-EXP2
imports INT
syntax Exp1 ::= foo() [function]
              | Exp1Val()

syntax Exp2 ::= foo() [function]
              | Exp1 

configuration &lt;k&gt; $PGM:Exp2 &lt;/k&gt;
              &lt;val&gt; Exp1Val():<b>Exp2</b> &lt;/val&gt;

rule [[ foo() => VAL ]]
     &lt;val&gt; VAL &lt;/val&gt;
endmodule
</pre>

#### TEST-EXP1
`kompile test-exp1.k --main-module TEST-EXP1 --syntax-module TEST-EXP1`

Output:
```
```
Parsed:
```
rule #withConfig(`foo()_TEST-EXP1_Exp1`(.KList)=>#SemanticCastToExp1(VAL),`<val>`(#noDots(.KList),#SemanticCastToExp1(VAL),#noDots(.KList))) requires #token("true","Bool") ensures #token("true","Bool")
```
`krun -cPGM='foo()' --definition test-exp1-kompiled`

Output:
```
/home/sguest/work/k/k-distribution/target/release/k/bin/../lib/kframework/k-util.sh: line 114: 3983362 Segmentation fault      (core dumped) "$@"
[Error] krun: ./test-exp1-kompiled/interpreter 
/tmp/.krun-2023-12-15-12-33-40-Lcx1VQpGb6/tmp.in.M633PiUsc0 -1 
[...]

PR #3925, opened 2024-01-25
Handle contextual functions in `PushTopAmbiguityUp`

Scott-Guest:
Part of #3880

`PushTopAmbiguityUp` ensures that the LHS of any top-level rewrite is not an `Ambiguity` by pushing any such `Ambiguity` above the containing `#RuleContent`. This simplifies sort inference by making the declared sort of the LHS readily available, as needed for function, macro, etc. rules.

This PR fixes some issues in `PushTopAmbiguityUp`, namely
- Correctly handle contextual functions. This was partly responsible for #3880, but is not the only issue there.
- Ensure that we don't introduce nested ambiguities. I couldn't come up with a contrived enough test case to prove this is necessary, but conceptually this would be an issue if the `#RuleBody` is itself an `Ambiguity` containing a top-level rewrite whose LHS is an `Ambiguity`.
- General code clean-up - documentation, better naming, factoring out some duplication


Baltoli (approved):
Looks good!

PR #3929, opened 2024-01-26
Add flag to use strict K semantics in LLVM backend

Baltoli:
This PR adds a frontend flag to activate the strict semantics for the Bytes sort implemented in https://github.com/runtimeverification/llvm-backend/pull/962, as well as a regression test based on the original issue reported in https://github.com/runtimeverification/llvm-backend/issues/930.

Blocked until https://github.com/runtimeverification/llvm-backend/pull/962 is propagated to K.


Baltoli:
Closing following a discussion on https://github.com/runtimeverification/llvm-backend/pull/962; we will invert the logic and make the strict behaviour default.

PR #3936, opened 2024-01-31
3926 smoke test llvm booster integration

### LINKED ISSUE #3926: Add integration test for the Booster to the K test suite
We want to be able to catch the situation where an LLVM backend update causes the booster to crash (e.g. via ABI incompatibility) before the problem reaches downstream semantics; currently this is not the case. A K integration test that spins up a booster server will go some way to fixing the issue.

jberthold:
As a smoke test, compile the `IMP` semantics from `tests/smoke/imp.k` and run two basic tests with `kore-rpc-booster` to check that LLVM backend and booster server work together.

Fixes #3926 

This also enables the tests in `tests/smoke` as these were previously not executed by `mvn verify`.

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -55,6 +55,7 @@ RUN    apt-get update              \
         lld-${LLVM_VERSION}        \
         llvm-${LLVM_VERSION}-tools \
         locales                    \
+        lsof                       \
         maven                      \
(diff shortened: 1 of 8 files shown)


geo2a (approved):

Baltoli (approved):
Looks good - thanks for this @jberthold!

PR #3942, opened 2024-02-01
Remove `kast` and `kore` attributes

Scott-Guest:
This PR removes the `kast` and `kore` attributes due to the fact that
- All modern backends are KORE-based
- Any potential future non-KORE based backend will have different considerations than the previous backends which motivated these attributes, so there's no point keeping existing usages around

Additionally, we remove logic in the frontend pertaining to an old usage of `kore` as an attribute on rules which is undocumented and unused in all our semantics.

A textual search

https://github.com/search?q=org%3Aruntimeverification+%2Fmodule.*%5C%5B.*%28kore%7Ckast%29.*%5C%5D%2F&type=code

shows a few lingering usages of `kore` on modules that need to be removed before this is merged:
- runtimeverification/pyk#855
- runtimeverification/evm-semantics#2279


dwightguth:
I believe that the `kore` attribute on rules was designed to facilitate writing axioms directly in k using ml connectives with more complete control over how they were encoded. It doesn't sound like we ever significantly benefitted from it, and the one place I know it used to be used has since been trimmed from k, so its removal should indeed be fine.

Baltoli (approved):
Looks good!

Baltoli:
Pyk, EVM prerequisites are fixed so go ahead and merge this when the conflicts are resolved!

Issue #3933, opened 2024-01-29
Consolidate special collection attributes


tothtamas28:
Module-to-KORE conversion defines a few special [collection sorts](https://github.com/runtimeverification/k/blob/7d5882aa7a28139cb733bfadedf36794a360a60b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L203):

```
SET.Set
MAP.Map
LIST.List
ARRAY.Array
RANGEMAP.RangeMap
```

When generating the KORE declarations for these sorts, a few special attributes are included:

```
hooked-sort SortSet{} [
  concat{}(Lbl'Unds'Set'Unds'{}()),
  element{}(LblSetItem{}()), hook{}("SET.Set"),
  unit{}(Lbl'Stop'Set{}()),
  ...
]
```

The source of these attributes is the corresponding concat production: https://github.com/runtimeverification/k/blob/7d5882aa7a28139cb733bfadedf36794a360a60b/k-distribution/include/kframework/builtin/domains.md?plain=1#L721

In particular, `concat` is taken from the `klabel`, whereas `element` and `unit` are taken from the respective K attribute ([source](https://github.com/runtimeverification/k/blob/7d5882aa7a28139cb733bfadedf36794a360a60b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L439)).

---

There are two approaches this for improving on this.

1. A single multiary attribute on the concat production

```k
syntax Set ::= Set Set [concat(SetItem, .Set)]
```

2.  A family of nullary attributes, each on the corresponding production

```k
syntax Set [collection]  // When this is defined, the corresponding productions are checked for existence and well-sortedness
syntax Set ::= Set Set        [concat]
syntax Set ::= SetItem(KItem) [element]
syntax Set ::= ".Set"         [unit]
```

If collections require special handling on the backend, then it should also be ensured that the corresponding attributes are included exactly on the special collection sorts.

Baltoli:
We should maintain the property that looking up the unit / element / unit is O(1) in the backend

PR #3943, opened 2024-02-01
Clean up `CheckAtt`

Scott-Guest:
This PR refactors `CheckAtt` to more clearly separate the checks on different attributes.

Most of the change here is just shuffling code between functions, but we additionally
- Implemented `HasAtt` and `HasLocation` for `Module` so we can use the same `checkUnrecognizedAtts` and `checkRestrictedAtts` functions for both modules and sentences 
    - In doing so, fixed a bug where we failed to check restricted attributes on modules at all 
- Corrected the whitelist to allow `one-path` and `all-path` on `Module` as well as `Claim`
- Slightly changed a few error messages
- Removed a manual check for attribute values on `binder`, as this is now covered by the general check

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
@@ -189,11 +189,6 @@ public class CheckAtt {
     if (!prod.att().contains(Att.BINDER())) {
       return;
     }
-    if (!prod.att().get(Att.BINDER()).equals("")) {
-      errors.add(
-          KEMException.compilerError(
-              "Attribute value for 'binder' attribute is not supported.", prod));
-    }
     if (prod.nonterminals().size() < 2) {
       errors.add(
           KEMException.compilerError(
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
@@ -11,6 +11,7 @@ import org.kframework.attributes.Att;
 import org.kframework.attributes.Att.Key;
 import org.kframework.attributes.HasLocation;
 import org.kframework.builtin.Sorts;
+import org.kframework.definition.HasAtt;
 import org.kframework.definition.Module;
 import org.kframework.definition.Production;
 import org.kframework.definition.ProductionItem;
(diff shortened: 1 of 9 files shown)


kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -21,46 +22,77 @@
...
   private final Module m;
-  private final boolean isSymbolicKast;
 
-  public CheckAtt(
-      Set<KEMException> errors, KExceptionManager kem, Module m, boolean isSymbolicKast) {
+  public CheckAtt(Set<KEMException> errors, KExceptionManager kem, Module m) {
     this.errors = errors;
     this.kem = kem;
     this.m = m;
-    this.isSymbolicKast = isSymbolicKast;
```
Scott-Guest, line 38: The `isSymbolicKast` change is really part of #3942

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -71,85 +103,109 @@ private void checkLabel(Sentence sentence) {
       if (label.contains("`") || whitespace.matcher(label).find()) {
         errors.add(
             KEMException.compilerError(
-                "Rule label '" + label + "' cannot contain whitespace or backticks.", sentence));
+                "Label '" + label + "' cannot contain whitespace or backticks.", sentence));
```
Scott-Guest, line 106: According to `Att.scala`, the `label` attribute can appear on any `Sentence` not just a `Rule`

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -71,85 +103,109 @@ private void checkLabel(Sentence sentence) {
...
+    if (rule.isMacro()) {
+      kem.registerCompilerWarning(
+          ExceptionType.FUTURE_ERROR,
+          errors,
+          "The attribute ["
+              + rule.att().getMacro().get()
+              + "] has been deprecated on rules. Use this label on syntax declarations instead.",
+          rule);
+    }
+  }
```
gtrepta, line 121: We discussed yesterday turning this into an error. Can we do that in this PR?

gtrepta (approved):
LGTM, with one comment about macro rules.

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
Scott-Guest: I can make a follow up PR - might have fall out and want to get this merged now for my work on #3933

Scott-Guest: Bruce handled it in #3956

PR #3955, opened 2024-02-05
Fix inconsistent symbol naming for syntactic lists

### LINKED ISSUE #3940: Inconsistent symbol names between KAST and KORE for syntactic list
K version: `v6.1.105-0-gd4cb1a0100`

Consider the following definition.

```k
module INTS
  imports INT-SYNTAX
  syntax Ints ::= List{Int, ","} [klabel(ints), symbol]
endmodule
```

`kompile` generates the following productions in `compiled.txt`:

```k
syntax Ints ::= Int "," Ints [klabel(ints),          symbol, userList(*)]
syntax Ints ::= ".Ints"      [klabel(.List{"ints"}), symbol, userList(*)]
```

However, in `definition.kore`, the corresponding symbols are as follows:

```
symbol Lblints{}(SortInt{}, SortInts{}) : SortInts{} [...]
symbol Lbl'Stop'List'LBraQuot'ints'QuotRBraUnds'Ints{}() : SortInts{} [...]
```

Here, the expected symbol name for the empty list would be
```
Lbl'Stop'List'LBraQuot'ints'QuotRBra'
```

as the `symbol` attribute is present on the corresponding production.

This discrepancy is also present in `compiled.json`, where the `klabel` field (not the attribute) of the object is

```json
{"node":"KLabel","name":".List{\"ints\"}_Ints","params":[]}
```

As a side remark, right now it seems impossible to generate a human-readable KORE symbol name for the empty list. We should consider generating a simpler label (e.g. `.ints` for this example), and also provide an attribute to specify the `klabel` explicitly.

Baltoli:
This PR makes a small refactoring to the way that we generate symbol names from syntactic list declarations. Specifically, it:
* Respects the `symbol` attribute when generating a label for the terminator element.
* Adds a new attribute `terminator-klabel(_)` by which the actual label for the terminator can be specified.

Documentation, attribute checks and a test for the new attribute are updated.

Fixes https://github.com/runtimeverification/k/issues/3940

### DIFF
--- a/k-distribution/tests/regression-new/checks/checkTerminatorKLabel.k
+++ b/k-distribution/tests/regression-new/checks/checkTerminatorKLabel.k
@@ -1,0 +1,7 @@
+module CHECKTERMINATORKLABEL
+  imports INT
+  syntax KItem ::= foo()
+                 | bar() [terminator-klabel(bar)]
+
+  syntax Ints ::= List{Int, ","} [klabel(ints), terminator-klabel(.ints), symbol]
+endmodule
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -373,6 +373,34 @@ the JSON Kast format, building terms using the user-provided pretty
 `symbol, klabel(...)` is easier and less error-prone if the auto-generation
 process for klabels changes.
 
+#### Syntactic Lists
+
+When using K's support for syntactic lists, a production like:
+```k
+syntax Ints ::= List{Int, ","} [klabel(ints), symbol]
+```
+will desugar into two productions:
+```k
+syntax Ints ::= Int "," Ints [klabel(ints), symbol]
+syntax Ints ::= ".Ints"      [klabel(List{"ints"}), symbol]
+```
+
+Note that the label for the _terminator_ of the list has been generated
+automatically from the label on the original production. It is possible to
+control what the terminator's label is using the `terminator-klabel(_)`
+attribute. For example:
+```k
+syntax Ints ::= List{Int, ","} [klabel(ints), terminator-klabel(.ints), symbol]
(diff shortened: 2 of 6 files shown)


k-distribution/tests/regression-new/checks/checkTerminatorKLabel.k
```diff
@@ -0,0 +1,7 @@
+module CHECKTERMINATORKLABEL
+  imports INT
+  syntax KItem ::= foo()
+                 | bar() [terminator-klabel(bar)]
+
+  syntax Ints ::= List{Int, ","} [klabel(ints), terminator-klabel(.ints), symbol]
+endmodule
```
Scott-Guest, line 7: This missing newline should have been caught by spotless, I'll look into it

Scott-Guest (approved):

gtrepta (commented):
I've noticed something strange while testing the symbol names on this branch. When declaring the production like this:
```
syntax Ints ::= List{Int,","} [klabel(ints)]
```

The symbol name for the terminator is `.List{"_,__TEST_Ints_Int_Ints"}_Ints`. Isn't it expected to be `.List{"ints"}_Ints`?

Baltoli:
Ah, good catch @gtrepta - that's happening because the text between `{` `}` is the _actual klabel_ of the non-terminator case. If the production isn't a `symbol`, the internal logic for what its klabel is uses the mangled name rather than the value of `klabel(_)`. I think this is fine off the top of my head, but if you can think of something it'll break then we can amend the behaviour.

gtrepta:
Ah, it makes sense why it's happening now. I think it's fine, I was just a little confused by it.

Issue #3947, opened 2024-02-01
Drop support for the `macro` attribute on rules


tothtamas28:
Having the `macro` attribute on rules has been deprecated for over two years (https://github.com/runtimeverification/k/pull/2272), consider dropping support.

This would save an implicit compiler pass in `ModuleToKORE` that checks each label production for the `macro` attribute on its function rules: https://github.com/runtimeverification/k/blob/50b6d0a969eee48834a3581ef70c812d55d8ff45/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1694

Baltoli:
This is just a question of dropping the loop over rules - because the frontend has accepted the rule, the production must necessarily have the `macro` attribute. We should **prohibit** macro[-rec] / alias[-rec] on rules rather than just warning for them.

PR #3956, opened 2024-02-05
Prohibit macro-like attributes on _all_ rules

Baltoli:
Previously, we only warned in these cases. Making this situation an error rather than a warning means that we can also simplify a lot of code that checked the macro-like attributes on specific sentence types as well.

Fixes https://github.com/runtimeverification/k/issues/3947


gtrepta (approved):

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
```diff
@@ -108,15 +108,14 @@ private void checkLabel(Sentence sentence) {
...
-              + "] has been deprecated on rules. Use this label on syntax declarations instead.",
-          rule);
+  private void checkMacro(Sentence sentence) {
+    if (!(sentence instanceof Production) && sentence.isMacro()) {
+      errors.add(
+          KEMException.compilerError(
+              "The attribute ["
+                  + sentence.att().getMacro().get()
+                  + "] may only appear on syntax productions.",
+              sentence));
```
Scott-Guest: I'd prefer if we us the general attribute whitelist infrastructure here instead of a manual error (i.e. update each `Key` in `Att.scala` to have `onlyon[Production]`).

Scott-Guest (approved):
LGTM except the manual error. Looks like the tutorial integration tests need to have a few usages of `[macro]` removed too.

kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
Baltoli: Very good point!

Baltoli:
~~Blocked on #3962 going through~~

Issue #3891, opened 2023-12-22
Frontend code cleanup + audits


Baltoli:
# Introduction

As discussed in our [year-end review](https://github.com/runtimeverification/k/issues/3871), we want to be making a push towards simplicity and removal of un-needed code across the various K projects in 2024.

The biggest offender in this respect is the K Frontend; this project has been maintained continuously for the best part of 15 years at this point and so presents the greatest opportunity to remove code that is no longer needed. Actionable steps that we identified in the review meeting were to:
* Run some kind of Java code coverage tool over the set of large semantics & regression tests that we maintain; this will give us an idea if there are non-statically dead code paths that never get run in practice.
* Schedule a review meeting with relevant parties to comb through the code tree together in a meeting.
  * Important to make sure that we record the outcome of this as documentation.
  * I have invited @dwightguth and @Scott-Guest initially; we can grow this group if desirable.

Other bits of knowledge transfer that we might want to do as part of this process:
* https://github.com/runtimeverification/k/issues/1278
* https://github.com/runtimeverification/k/issues/1415 (@Scott-Guest has tidied up and codified a lot of this as part of type inferencing)
* https://github.com/runtimeverification/k/issues/1443

# K Frontend Deep Dive (17 Jan 24)

We went through the K Frontend in some detail in a meeting; a brief summary of the main points that we identified there were:
* The [kore](https://github.com/runtimeverification/k/tree/master/kore) module is poorly named as a consequence of history; it should be renamed now that KORE refers to an actual language.
  * We may be able to fully remove this module in the near future; the main reason that it exists is to hive away the Scala code in the frontend.
  * Once we can https://github.com/runtimeverification/k/issues/3699, it should be possible to remove the Scala code entirely by using the new Java pattern matching features.
  * This is blocked only on updating our build and release infrastructure to Ubuntu Noble (due [April 2024](https://discourse.ubuntu.com/t/noble-numbat-release-schedule/35649)), so I think we can wait for that update to go through before taking a further look at the Scala code.
* There are a number of different data structures defined in the frontend across several different locations:
  * The ["kore" ADT](https://github.com/runtimeverification/k/tree/master/kore/src/main/scala/org/kframework/kore), which is actually a representation of K's _inner syntax_[^1]. The compiler pipeline that lowers a K definition is implemented over this representation.
[...]

PR #3967, opened 2024-02-08
Delete statically dead code

Scott-Guest:
As part of the  cleanup discussed in #3891, this PR removes statically dead code as identified by IntelliJ.

IntelliJ's analysis reports some false positives, e.g. with dependency injection and usages in Scala code, so I had to manually review and delete everything. It also seems to miss some dead Scala code.

The "Batch N" commits just delete dead code, while the following commits perform some easy simplifications I noticed.


Baltoli (approved):
Nice!

Issue #3909, opened 2024-01-18
Review `--output latex` options


Baltoli:
Take a look at how well supported the LaTeX output of the compiler is - does it work in practice?

We don't have the _prelude_ that provides meaning to `--output latex`; you can easily emit the latex that corresponds to terms but users need to actually supply the formatting.

Baltoli:
Issue about latex prelude exists already? Low priority - is anyone using it?

ehildenb:
https://github.com/runtimeverification/k/issues/1560

Baltoli:
We might just want to drop the frontend implementation of this for now; if there's an appetite to do similar in the future then we can reimplement in Pyk.

PR #3968, opened 2024-02-08
Remove latex attribute functionality

gtrepta:
fixes: #3909 

This removes functionality related to the `latex` attribute. It also adds a compiler warning for any usage of `latex` to give downstream semantics time to remove it.


gtrepta:
There's some strange nondeterminism that's breaking one of the tests.

The expected output:
https://github.com/runtimeverification/k/blob/df280ae43d6d8e3647ac78442ff62a39c8a9ad4d/k-distribution/tests/regression-new/set-symbolic-tests/inset-11-spec.k.out

The output on this branch:
```
kore-exec: [337301] Warning (WarnStuckClaimState):
    The configuration's term doesn't unify with the destination's term and the configuration cannot be rewritten further. Location: inset-11-spec.k:9:11-9:88
Context:
    (InfoReachability) while checking the implication
  <k>
    inSetResult ( Y:MyId in SET
    SetItem ( X:MyId )
    SetItem ( Z:MyId ) ) ~> .
  </k>
#And
  {
    false
  #Equals
    X:MyId in SET
  }
#And
  {
    false
  #Equals
    Z:MyId in SET
    SetItem ( X:MyId )
  }
[Error] Prover: backend terminated because the configuration cannot be rewritten further. See output for more details.
```

Removing the latex attributes from the Set productions in domains.md caused this. I wouldn't expect it to do that, and I don't think I should. Regardless, I'm making a note here before I update the expected outputs.

Scott-Guest (approved):
LGTM. There's a reference to `latex(_)` in the list of attributes in the user manual that should be removed too.

For the flaky test, I ran into a similar issue before when I removed an unused attribute from the Map collection. Probably fine to just update the expected output and let the haskell backend team know it's flaky?

PR #3974, opened 2024-02-09
add back "dead" code that was actually live

dwightguth:
There were a couple methods recently deleted from the K frontend that were actually being used by the Maude backend. I have added them back and annotated them so that they will not be automatically deleted by further dead code elimination in the future.


Scott-Guest (approved):
LGTM, will make sure to check the Maude backend for future dead code elimination

Baltoli (approved):

PR #3975, opened 2024-02-12
Remove special-casing for non-unique symbols

### LINKED ISSUE #3964: Uniqueness of `symbol` not enforced
`kast.md` defines `#Top` as https://github.com/runtimeverification/k/blob/01b4a3826b9c0037698efd553fcb519a0fc61b24/k-distribution/include/kframework/builtin/kast.md?plain=1#L141

and `#True` as https://github.com/runtimeverification/k/blob/01b4a3826b9c0037698efd553fcb519a0fc61b24/k-distribution/include/kframework/builtin/kast.md?plain=1#L143

(Similarly for `#Bottom` and `#False`.)

Both productions have the same `klabel` and are `symbol`-s. However, according to the [K User Manual](https://kframework.org/docs/user_manual/#klabel(_)-and-symbol-attributes),

> The symbol provided _must_ be unique to this definition. This is enforced by K.

In this case, it is not enforced. As a consequence, there are productions in `compiled.json` that have the same `klabel` (field, not the attribute), which makes processing productions more difficult.

---

Possible fixes:
1. [If this is a bug] Add the missing check for `symbol` uniqueness. Drop `#True` and `#False` as they do not seem to be indispensable.
2. [If this is a feature] Adjust the documentation. Add a new kompiler pass that eliminates the duplicate production from `compiled.json`.

Baltoli:
This PR is a partial fix for https://github.com/runtimeverification/k/issues/3964; it makes the logic for which productions are allowed to have duplicated symbol declarations attached.

With a minimal amount of effort, all but one case of this special-casing can be removed:
* One integration test still used the old `when` syntax for `requires` clauses; as far as I can tell no actual code uses this syntax any more.
* Another test has two output stanzas permuted by some kind of ordering nondeterminism.

The remaining case (empty K sequences) is harder to fix, as both syntaxes are used in real K code. We have a few options here:
* Just permit this single special-casing to exist, and document that both `.` and `.K` map to the KORE `dotk{}()`.
* Pick one of the two syntaxes as our preferred option, then add a deprecation warning to the other and fix instances we find in real code. Eventually remove the other one and make the check fully consistent with no special cases.
* Implement a more general mechanism whereby two productions can produce the same KORE.

Downstream testing to make sure nobody is actually using these old cases:
* [x] C Semantics
* [x] KEVM


Baltoli:
We should open a question to #k about _which_ syntax is preferred. Consensus in the frontend meeting is _weakly_ towards `.K` but open to convincing.

Should be easy enough, if a bit tedious, to implement this when we decide.

Scott-Guest (approved):

gtrepta (approved):

Issue #3966, opened 2024-02-07
Questions regarding symbol overloading


tothtamas28:
The [K User Manual](https://github.com/runtimeverification/k/blob/5fdec964181fec112631968bb32c3f187eac967e/docs/user_manual.md?plain=1#L324) hints on the possibility of overloading symbols:

> If we need to reference a certain syntax production externally, we have to manually define the klabels using the `klabel` attribute. One example of where you would want to do this is to (...) **enable overloading of a given symbol**.

> In general, it's recommended to use the `symbol` attribute whenever you use `klabel` unless you explicitly have a reason not to (eg. **you want to overload symbols**, or you're using a deprecated backend).

Accordingly, in `ModuleToKORE`, the set of overloading productions is computed:

https://github.com/runtimeverification/k/blob/5fdec964181fec112631968bb32c3f187eac967e/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L240

based on the poset of overloading productions:

https://github.com/runtimeverification/k/blob/5fdec964181fec112631968bb32c3f187eac967e/kore/src/main/scala/org/kframework/definition/outer.scala#L72

If I understand correctly, this essentially defines `p1 < p2` as:

1. `p1 != p2`
2. `p1` and `p2` have the same `klabel` attribute
3. `p1` and `p2` have the same positive number of nonterminals
4. `p1.sort <= p2.sort`
5. For every nonterminal index `i`, `p1.nonterminal[i].sort <= p2.nonterminal[i].sort`

Then to each production that is an overload of some other production w.r.t. this definition the `anywhere` attribute is attached. Another way this attribute can appear on a production is if the corresponding symbol is rewritten by a rule with the attribute.

 In the resulting KORE, both original and overloaded symbols are present. The `anywhere` attribute appears on KORE symbol declarations corresponding to overloading productions.

## Questions

1. How does overloading work from the user's perspective? Is it a parser feature? Should we document it?
2. Currently nothing stops the user from accidentally overloading a production (by reusing a `klabel` they are not aware of). Should we add an attribute to enable the user to signal purpose of overloading?
[...]

PR #3978, opened 2024-02-12
Add one-argument version of `symbol(_)`

Baltoli:
This PR is the first change identified as necessary in #3966; it takes a necessary but insufficient step towards rationalising the confusing `klabel` and `symbol` attributes.

The change made here is to permit a new arity-1 version of the `symbol` attribute, such that `symbol(X)` behaves identically to `klabel(X), symbol`. Doing so will allow us to begin the process of deprecating the `klabel` attribute once we have an alternative formulation for overloaded symbols.

The code changes required are small; I have added a check to ensure that new code doesn't mix the old and new idioms, as well as a test and the beginning of new documentation for `symbol(_)` that can be updated as we make related changes here.


Baltoli:
~~Blocked on https://github.com/runtimeverification/k/pull/3979~~

Scott-Guest (approved):

gtrepta (approved):

Issue #3973, opened 2024-02-09
Automated experiments on the compiler pipeline


gtrepta:
The compiler pipeline in the frontend has grown gradually over the lifetime of the project, and the purpose for each stage and their dependencies on each other has become unclear. Some stages have little to no documentation, and are possibly unneeded or more complex than they need to be.

An investigation into the pipeline by poking at each of these stages, removing them or moving them around and seeing if it breaks any tests and why those tests break can shed some light onto the pipeline.

One technique for doing this can be with git bisect. A sequence of commits on top of `develop` that each make a single, small change to the pipeline can be fed to `git bisect run` with a script that rebuilds the frontend and runs the tests. `git bisect` will then find the first commit that breaks something, giving us changes before it that can be made, and a breaking change that we can investigate to help fill in missing documentation or make a refactoring.

gtrepta:
I had a look at the duplicate pairs of `generateSortPredicateSyntax` and `generateSortProjections` in the pipeline, seeing how many stages I could remove between them.

For reference, here's how the stages are set up currently:

https://github.com/runtimeverification/k/blob/1e4aa2be6b49ca766e01d78eae0e418a749bb12b/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java#L248-L259

And here's after I moved what I could:

```
            .andThen(constantFolding)
            .andThen(propagateMacroToRules)
            .andThen(guardOrs)
            .andThen(resolveFreshConfigConstants)
            .andThen(generateSortPredicateSyntax)
            .andThen(generateSortProjections)
            .andThen(expandMacros)                                    //
            .andThen(AddImplicitComputationCell::transformDefinition) // Only these three remain
            .andThen(resolveFreshConstants)                           //
            .andThen(generateSortPredicateSyntax)
            .andThen(generateSortProjections)
            .andThen(checkSimplificationRules)
```

The regression tests still pass with this setup. The definitions in the profiling tests also kompile without issue, and identical `definition.kore` files get generated as on `develop`. So it feels pretty safe to make a PR, but I'll try some more testing downstream. The symbolic backends aren't exercised very much here.

gtrepta:
I've compared the `definition.kore` files generated on evm-semantics and wasm-semantics and they're identical. I also compared all of the definitions in the regression tests, and they differ, but it only appears to be the ordering and numbering of some sentences. I feel confident enough to open a PR for this, and leave the decision to merge it in up for review.

gtrepta:
[...]

PR #3980, opened 2024-02-12
Pipeline rearrange

gtrepta:
#3973 

This factors a few compiler pipeline stages out from between a pair of duplicate stages. If we're able to eliminate the duplicate stages, this change will help narrow down what changes need to be made to do that.


Baltoli (approved):

ehildenb:
@gtrepta was this tested downstream with KEVM and KMIR and KWasm? Feel free to add back `automerge` if so!

kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java
```diff
@@ -244,18 +244,18 @@ public Function<Definition, Definition> steps() {
...
             .andThen(resolveFreshConfigConstants)
+            .andThen(generateSortPredicateSyntax)
+            .andThen(generateSortProjections)
+            .andThen(expandMacros)
+            .andThen(AddImplicitComputationCell::transformDefinition)
             .andThen(resolveFreshConstants)
             .andThen(generateSortPredicateSyntax)
             .andThen(generateSortProjections)
+            .andThen(checkSimplificationRules)
             .andThen(subsortKItem)
```
ehildenb, line 259: I see that this pass is done 2 times as well.

gtrepta: `subsortKItem`? Yeah, we identified that in the frontend meeting today, we'll certainly look into it.

gtrepta:
> @gtrepta was this tested downstream with KEVM and KMIR and KWasm? Feel free to add back `automerge` if so!

@ehildenb KEVM and KWasm checked out, their definition.kore files with this change were identical to the ones without the change. KMIR's build system is using some temporary folders which cause different location attributes in the emitted kore, making it harder to determine whether or not they're semantically identical. I'll see if I can isolate the stage of the build that just calls kompile and then make the comparison from there.

ehildenb:
@gtrepta KEVM and KWasm should be good enough I guess, though it would be good to get in the habit of also checking KMIR for stability. I guess adding the relevant files upstreamed here (with `kompile -E` like we've done with KEVM/KWasm) can be done in a follow-up PR to make later changes easier.

Issue #3946, opened 2024-02-01
Audit the backends for relevant KORE attributes


tothtamas28:
Related: https://github.com/runtimeverification/k/issues/3500
Related: https://github.com/runtimeverification/k/issues/3540

Currently, when generating `compiled.json` and `definition.kore`, the Frontend emits a lot of attributes, many of which are probably not needed by the backends. Ideally, only those attributes should be generated that are used by at least one backend (except perhaps for `source` and `location` which are always useful for debugging and traceability).

Compile the list of attributes used by each backend, grouped by the syntactic element they are attached to.

- [x] LLVM backend - @Baltoli 
  - String lookups when attributes are required; grep for `.attributes()`, `getStringAttribute()`
- [x] Haskell backend - @goodlyrottenapple 
  - Supports cases not supported by the booster; will need some more auditing.
- [x] Booster backend - @goodlyrottenapple 
  - Relatively easy to check for attribute usage.
- [ ] Maude backend - @dwightguth / @sskeirik 

---

For guiding search in the backend code, a simple script that exports the list of KAST / KORE attributes for a kompiled definition:

[export_atts.py](https://github.com/runtimeverification/k/files/14126043/export_atts.py.txt)

Copy into `pyk` root, then

```
$ kompile imp.k --emit-json
$ poetry run python3 export_atts imp-kompiled
KORE attributes exported: imp-kompiled/atts.kore.csv
KAST attributes exported: imp-kompiled/atts.kast.csv
```

Generated artifacts: [atts.kore.csv](https://github.com/runtimeverification/k/files/14126077/atts.kore.csv), [atts.kast.csv](https://github.com/runtimeverification/k/files/14126076/atts.kast.csv)

---

Results: https://docs.google.com/spreadsheets/d/1_lX1NqjphKBdTtWwiOQC8Xt7yY565xlgS9iGy1JkR6A

Baltoli:
@Scott-Guest collected a whitelist of attributes in the static code (complete) to correspond to these dynamic lists (maybe incomplete, but a useful starting point).

goodlyrottenapple:
Here is the list of attributes being used in the haskell backends. The ones in **bold** appear to be unused/deprecated:

| Attribute                                         | Haskell Backend | Booster Backend | Attribute Type |
|---------------------------------------------------|:---------------:|:---------------:|:--------------:|
| priority                                          |        ✅        |        ✅        |      Axiom     |
[...]

PR #3983, opened 2024-02-13
Only emit attributes used by some backend

### LINKED ISSUE #3960: Do not use Java class names as attributes
There are a few attributes that leak Frontend internals by using a class name directly as their key:

* `org.kframework.attributes.Location`
* `org.kframework.attributes.Source`
* `org.kframework.definition.Production`
* `org.kframework.kore.Sort`

and potentially also:

* `java.lang.Integer`
* `java.lang.String`

Some of these attributes end up in `compiled.txt`, `compiled.json` and `definition.kore`. In order to make these attributes parseable, the outer lexer recently had to be modified to allow dots in the key (https://github.com/runtimeverification/k/pull/3953, https://github.com/runtimeverification/pyk/pull/864).

There are two levels on which this issue can be addressed.
1) Do not allow these attributes to cross system boundary.
2) Do not use these attributes internally (i.e. between compiler passes) either. Restore the outer lexer to fail on a dot in an attribute key to enforce this to some extent.

Scott-Guest:
Part of #3960

This PR updates the attribute whitelist and `ModuleToKORE` so that we only emit attributes which are actually used by at least one backend. This partially resolves #3960 as we no longer emit `org.kframework.definition.Production` nor `org.kframework.kore.Sort`.

The list was mostly compiled based on #3946, and I've confirmed that the `evm-semantics` test suite passes. 

To be explicit, the following attributes will no longer be emitted to KORE after this PR:
- all-path
- anonymous
- applyPriority
- avoid
- bag
- bracketLabel
- cell
- cellCollection
- cellFragment
- cellName
- cellOptAbsent
- cellSort
- color
- contentStartColumn
- contentStartLine
- context
- denormal
- digest
- dummy_cell
- exit
- filterElement
- fresh
- group
- haskell
- heat
- hybrid
- impure
- index
- initial
- initializer
- internal
- latex
- left
- locations
- maincell
- mlBinder
- mlOp
- multiplicity
- non-assoc
- not-lr1
- not-lr1-modules
- notInjection
- one-path
- org.kframework.definition.Production
- org.kframework.kore.Sort
- originalPrd
- parser
- prec
- predicate
- prefer
- prettyPrintWithSortAnnotation
- private
- projection
- public
- recordPrd
- recordPrd-empty
- recordPrd-item
- recordPrd-main
- recordPrd-one
- recordPrd-repeat
- recordPrd-subsort
- recordPrd-zero
- refreshed
- result
- returnsUnit
- right
- seqstrict
- smt-prelude
- sortParams
- stream
- strict
- syntaxModule
- tag
- temporary-cell-sort-decl
- type
- unboundVariables
- unparseAvoid
- unused
- userList
- userListTerminator
- withConfig
- wrapElement

I'd appreciate if someone from each backend could take a quick glance over this list and confirm that, as far as you know, none of these attributes are needed in KORE:
- [ ] LLVM backend - @Baltoli 
- [ ] Haskell backend - @goodlyrottenapple, @jberthold
- [ ] Booster backend - @goodlyrottenapple, @jberthold
- [ ] Maude backend - @dwightguth, @sskeirik


tothtamas28:
Could you also add a `KeyRange` column here?

https://docs.google.com/spreadsheets/d/1_lX1NqjphKBdTtWwiOQC8Xt7yY565xlgS9iGy1JkR6A/edit#gid=1840102280

Scott-Guest:
I've confirmed that the `evm-semantics` tests pass with this change

Baltoli (approved):
Nice!

PR #3985, opened 2024-02-13
Add attribute-based overloading

Baltoli:
This PR is the next step towards the changes suggested in https://github.com/runtimeverification/k/issues/3966; having added a new syntax for the `symbol(_)` attribute in #3978, we can now address the second use of `klabel(_)` (marking overloaded sets of symbols).

The changes in this PR are as follows:
* Add a new attribute `overload(_)` that behaves identically to the current semantics of `klabel(_)` with respect to the construction of the overload partial ordering, but that does not participate in the _naming_ of symbols.
* Add checks for the new attribute:
  * Disallow `klabel(_)` and `overload(_)` on the same production; overload sets must be converted fully to use the new mechanism.
  * Warn when an overload set of only one production is generated.
* Document the semantics of overloading in the user manual based on our conversations in #3966.
* Add a new integration test that checks a simple example of overloading works as expected to disambiguate parses.

Once this PR is merged, we can begin the long process of deprecating `klabel(_)` and `symbol`!


Baltoli:
Because we emit a _different_ overload attribute to KORE; this is actually blocked on a change @Scott-Guest is working on to strip frontend attributes before generating KORE.

tothtamas28:
The Haskell backend seems to depend on an existing `overload` attribute: https://github.com/runtimeverification/k/issues/3946#issuecomment-1930372019.

Baltoli:
Indeed - that overload attribute is the _KORE_ overload attribute (which IIRC means something like "this axiom was generated from the overload partial ordering, and needs special treatment for unification" - a different semantics to the K attribute, which is something like "this symbol should participate in overload selection").

The plan is for @Scott-Guest to finish his work on #3983, which will conclude with a compiler pass that strips out attributes meaningful to K but not KORE before `ModuleToKore` is performed. This means we can keep the K `overload(_)` syntax in the frontend as well as the existing semantics of `overload` in KORE.

tothtamas28:
> strips out attributes meaningful to K but not KORE before `ModuleToKore` is performed

There are steps in `ModuleToKORE` that depend on the overload lattice, will it still be reproducible after stripping the K `overload` attribute?

Baltoli:
Ah, that's a good point - it would be reproducible in the current Java frontend because we've computed the lattice already, but it wouldn't be in the Python reimplementation. So maybe we should just rename `overload(_)` to something else to avoid the collision?

tothtamas28:
> So maybe we should just rename `overload(_)` to something else to avoid the collision?

I think `overload(_)` is a great name for this feature, and since it's user facing, naming it well is more important than the legacy backend attribute. So I think we should rename the backend attribute.

Also, in general, in order to avoid confusion, we should try to maintain uniqueness of attribute names. (There are a few cases currently that violate this, e.g. `RIGHT` / `RIGHT_INTERNAL`). Order of importance:

```
used in K input > used in KORE output > internal
```

Scott-Guest:
To clarify, the changes I'm making don't strip out any attributes *before* `ModuleToKORE`, they just prevent those attributes from being emitted *during* `ModuleToKORE`, so I don't think the naming conflict would have caused any issues.

Philosophically though, I agree we should keep attribute names unique. If we can get the existing examples cleaned up, it'd be easy to have the whitelist enforce this.

Baltoli:
Ah, I had a slight conceptual misunderstanding about the point at which the stripping would happen. I do agree we should keep them unique though, so it's nice this conversation has sort of pushed us to do it anyway

docs/user_manual.md
```diff
@@ -349,11 +349,96 @@ this way; it would be an error to apply `symbol(foo)` to another production in
...
+
+    syntax IValType ::= "i32" | "i64"
+    syntax FValType ::= "f32" | "f64"
+    syntax AValType ::= IValType | FValType
+    syntax Number ::= Int | Float
+    syntax IVal ::= "<" IValType ">" Int    [overload(<_>_)]
+    syntax FVal ::= "<" FValType ">" Float  [overload(<_>_)]
+    syntax  Val ::= IVal | FVal
+    // the problematic constructor:
+    syntax  Val ::= "<" AValType ">" Number [overload(<_>_)]
```
Scott-Guest: I don't quite follow this example - the `syntax  Val ::= "<" AValType ">" Number` constructor is indeed overly-permissive, but that is always true regardless of any overloading.

Even with a total order, you can still construct similarly problematic examples:
```
module NUMBERS-SYNTAX
    imports INT-SYNTAX
    imports FLOAT-SYNTAX
    syntax SubValType ::= "sub"
    syntax SuperValType ::= "super" | SubValType

    syntax Sub ::= "subVal"
    syntax Super ::= "superVal" | Sub

    syntax SubVal ::= "<" SubValType ">" Sub       [overload(<_>_)]
    syntax SuperVal ::= "<" SuperValType ">" Super [overload(<_>_)] | SubVal
endmodule

module NUMBERS
    imports NUMBERS-SYNTAX
    syntax SuperVal ::= creature32 ( Super ) [function, total]
    rule creature32(S:Sub) => <super> S
    rule creature32(S::Super) => <sub> S
endmodule
```

To me, the over-permissiveness of `syntax  Val ::= "<" AValType ">" Number` is the whole motivation behind overloading in the first place - we want more precise sorting invariants than what K's syntax declaration are able to express.

Still choosing to include the overly-permissive constructor even after using overloading just seems like an odd design decision in the example rather than something fundamental we need to explain to users about overloading. 

Reading through runtimeverification/hs-backend-booster#492, it was mentioned that the problematic constructor was useful because it allows us to write unambiguous rules which handle all overloads like
```
syntax Bar ::= bar(Number) [function, total]
               |  foo(Val) [function, total]
rule foo(< _ > N) => bar(N) // ambiguous without the problematic constructor
```
but I think a more principled and only slightly-inconvenient solution would be to just manually define eliminators
```
syntax AValType ::= valType(Val) [function, total]
rule valType(< V:IValType > _) => V
rule valType(< V:FValType> _ ) => V

syntax Number ::= number(Val) [function, total]
rule number(< _ > N:Int) => N
rule number(< _ > N:Float) => N
```
and write
```
rule foo(V) => bar(number(V))
```

Baltoli: Your analysis all makes sense to me; I guess I read Jost's writeup as a cautionary example that needed to be explained to users, and without thinking it through properly made the incorrect inference that because a non-total overload ordering is prone to producing this kind of broken case, total orderings cannot. I guess we can just cut this section from the manual entirely?

docs/user_manual.md
```diff
@@ -3151,3 +3196,8 @@ All of these hooks will also eventually need documentation.
   API.
 [^unique-symbol]: Except for in a very limited number of special cases from the
   K standard library.
+[^overload-issue]: See https://github.com/runtimeverification/hs-backend-booster/issues/492
```
Scott-Guest: This `[^overload-issue]` footnote is dead code now (and while we're at it, `[^1]` is dead too).

docs/user_manual.md
```diff
@@ -349,11 +349,55 @@ this way; it would be an error to apply `symbol(foo)` to another production in
...
+the overload relation as follows. We say that `P` is a more specific overload
+of `Q` if:
+* `P` and `Q` have the same `overload(_)` attribute. Note that the argument
+  supplied has no semantic meaning other than as a key grouping productions 
+  together.
+* Let `S_P` be the sort of `P`, and `S_p1` etc. be the sorts of its arguments
+  (c.f. for `Q`). The tuple `(S_P, S_p1, ..., S_pN)` must be lexicographically
+  _strictly less than_ `(S_Q, S_q1, ..., S_qN)` according to the definition's
+  subsorting relationship. That is, a term from production `P` must be
+  substitutable where one from `Q` is expected, but cannot be identically
```
Scott-Guest: I don't follow this point - wouldn't substitutability fail to hold because the overload ordering is covariant rather than contravariant in the parameter sorts?

Perhaps a better phrasing is that a more precise overload is a "restriction" of a less precise overload - when the inputs are restricted to a more precise sort, we can also give the output a more precise sort.

docs/user_manual.md
```diff
@@ -349,11 +349,55 @@ this way; it would be an error to apply `symbol(foo)` to another production in
...
+`LVal`.
+
+Formally, the compiler organises productions into a partial order that defines
+the overload relation as follows. We say that `P` is a more specific overload
+of `Q` if:
+* `P` and `Q` have the same `overload(_)` attribute. Note that the argument
+  supplied has no semantic meaning other than as a key grouping productions 
+  together.
+* Let `S_P` be the sort of `P`, and `S_p1` etc. be the sorts of its arguments
+  (c.f. for `Q`). The tuple `(S_P, S_p1, ..., S_pN)` must be lexicographically
```
Scott-Guest: This isn't a lexicographic ordering - overloading requires that **every** component of `P`'s tuple is a subsort of the corresponding component of `Q`'s tuple, but lexicographic ordering only requires this for the first component where the tuples disagree.

Formally, renaming `S_P` to `S_p0` for ease of notation, the overload ordering defines `P < Q` if
- `S_pi <: S_qi` for all `0 ≤ i ≤ N`
- `(S_p0, S_p1, ..., S_pN) != (S_q0, S_q1, ..., S_qN)`

However, a lexicographic ordering would only require
- There exists some `0 ≤ k ≤ N` such that `S_pi = S_qi` for all `0 ≤ i < k` and `S_pk < S_qk`

Scott-Guest (approved):
Code LGTM, but a lingering conceptual question - does overloading have any semantic effect or does it only affect the disambiguation process?

For example, in
```
module NUMBERS-SYNTAX
    imports INT-SYNTAX
    imports FLOAT-SYNTAX
    syntax IValType ::= "i32" | "i64"
    syntax FValType ::= "f32" | "f64"
    syntax IVal ::= "<" IValType ">" Int   [overload(<_>_)]
    syntax FVal ::= "<" FValType ">" Float [overload(<_>_)]
endmodule
```
the two overloads are incomparable and thus will not affect disambiguation. Are the `overload` attributes here dead code then (and we should beef up our warning accordingly), or do they have an observable effect?

It would be good to clarify this in the user manual.

docs/user_manual.md
Baltoli: Good spot!

docs/user_manual.md
Baltoli: Yep, you're right

Baltoli:
> Code LGTM, but a lingering conceptual question - does overloading have any semantic effect or does it only affect the disambiguation process?

I believe it affects [unification](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2019-08-27-Unification-modulo-overloaded-constructors.md#L4) in the old Haskell backend, but figuring that algorithm + implementation combo out properly is probably beyond the scope of our current work on this and we'll need to defer to @jberthold et. al. to figure out what should happen here in the future.

PR #3987, opened 2024-02-14
Mark MAP.lookup* rules as preserving-definedness, add in_keys rules

jberthold:
* The symbolic `Map.lookup` rules preserve definedness by ensuring that the decomposed map argument is well-defined (no duplicate keys). If not (i.e., the LHS was undefined), the RHS will include `ensures #Bottom`.
* Likewise, the symbolic `Map.lookupOrDefault` rules preserve definedness by ensuring that the decomposed map argument is well-defined (no duplicate keys). If not (i.e., the LHS was undefined), the RHS will include `ensures #Bottom`.
* Rules are added to express the `MAP.in_keys` semantics in K instead of using inline kore as the existing simplification rules did.
  - `K in_keys(.Map)` simplifies to `K =/= K` to carry undefined key arguments over to the RHS
  - As in the `lookup` rules, `K in_keys( <decomposed> )` cases use an `ensures` clause to carry undefined `Map` arguments to the RHS.
  - The addition of these rules enables the backend to prune some redundant conditions in related tests  for operations on symbolic maps, test expectations have been updated accordingly.

This PR addresses an issue reported in https://github.com/runtimeverification/hs-backend-booster/issues/498/ Alternatively, this can be addressed by implementing _hooks_ in booster for the `MAP.lookup*` and `MAP.in_keys` functions (see https://github.com/runtimeverification/hs-backend-booster/pull/506)


jberthold:
Obsolete after implementing the relevant hooks.

PR #3991, opened 2024-02-14
Implement check for uses of deprecated symbols

Baltoli:
This PR implements a new `deprecated` attribute that can be applied to productions; if that production is subsequently used in a rule the compiler will emit a warning.

No code uses this feature yet, but I intend to use it to implement the canonicalisation of `.` and `.K` (#3964, #3975) by marking `.` as deprecated and doing some error-driven development.

### DIFF
--- a/kore/src/main/java/org/kframework/utils/errorsystem/KException.java
+++ b/kore/src/main/java/org/kframework/utils/errorsystem/KException.java
@@ -138,6 +138,7 @@ public class KException implements Serializable, HasLocation {
     IGNORED_ATTRIBUTE,
     REMOVED_ANYWHERE,
     DEPRECATED_DIRECTORY_FLAG,
+    DEPRECATED_SYMBOL,
     MISSING_HOOK,
     FIRST_HIDDEN, // warnings below here are hidden by default
     USELESS_RULE,
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -679,6 +679,18 @@ syntax Foo ::= foo() [unused]
 configuration <foo unused=""> .K </foo>
 ```
 
+### `deprecated` attribute
+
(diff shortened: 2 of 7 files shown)


Baltoli:
It's a logically separate change that belongs in a different PR, but I've verified locally that I can use this to mark `.` as deprecated and get warnings when it's used.

Scott-Guest (approved):

Scott-Guest:
Nice! This will definitely useful for future cleanups.

kore/src/main/java/org/kframework/utils/errorsystem/KException.java
```diff
@@ -138,6 +138,7 @@ public enum ExceptionType {
     IGNORED_ATTRIBUTE,
     REMOVED_ANYWHERE,
     DEPRECATED_DIRECTORY_FLAG,
+    DEPRECATED_SYMBOL,
```
gtrepta, line 141: What's the rationale for making a new error instead of using `FUTURE_ERROR`?

gtrepta (approved):
LGTM. Just one question about a decision.

kore/src/main/java/org/kframework/utils/errorsystem/KException.java
Baltoli: I guess my reading is that `FUTURE_ERROR` means that a decision made by the _compiler_ will break this in the future, whereas we're saying here that it's a decision made by the _semantics_ developer. It's a pretty thin line though, I agree

gtrepta: That makes sense, I see the distinction. Maybe more broadly I'm unclear on what the use case for these ExceptionTypes is, `-W` options? Message templates? Just an idle pondering. Maybe in the future we can visit this.

Issue #4003, opened 2024-02-18
Make each attribute its own class


Scott-Guest:
I've spent quite a bit of time trying to get #3993 working, and have yet to find a solution which simultaneously:
- Supports type safe lookup and insertion, accounting for forbidden / optional / required values of any possible type
- Allows us to maintain a single overloaded `add` function even in the presence of type erasure
- Doesn't cause falsely reported error messages in the Java code

Additionally, I've noticed a few pain-points with the current set up:
- We conflate two separate concepts
  - `key(value)` string pairs actually used as attributes in K or KORE source code
  - arbitrary data attached to syntax for use internally in the frontend, e.g., the `Sort` or `Production` attributes
- Attribute values are often structured data but stored as strings, meaning
  - The value parsing logic is decoupled from the attribute definition
  - The value has to be re-parsed every time it's read and re-stringified every time it's written (else we run into issues like #3960)
- `Att.scala` makes somewhat heavy usage of Scala-specific features, even though we plan to eventually deprecate Scala in the codebase

One possible solely-Java solution would be to:
- Make `Att` store a heterogenous `Set` of classes rather than a `Map` of key-value pairs. 
   - Internally, this would probably be a `Map<Class<?>, Object>` mapping `Foo.class` to at most one instance of `Foo` for fast lookup, but the external interface would be `Set`-like. 
- Arbitrary data can just be added as an instance of any class.
- Actual attributes `MyAtt` which should be considered for parsing in the frontend or emitting to KORE can implement an `Attribute` interface:
```
class MyAtt implements Attribute {
static public String key();
public String value();
static public Either<KEMException, MyAtt> fromValue(Optional<String> val);
public EnumSet<Syntax> allowedSyntax(); // Empty if not visible to users
public boolean shouldEmitToKORE();
}
```
Type-safe value creation/lookup is then handled by each attribute's constructors/methods rather than the `Att.scala` class.

Baltoli:
Don't have a monolithic PR; do smaller changes by keeping both implementations for a period of time.

Scott-Guest:
Putting this on the backlog. It's low-priority and more work than we have bandwidth for at the moment.

PR #4007, opened 2024-02-19
Use attributes that change semantics when computing rule hashes

### LINKED ISSUE #3984: Unique rule IDs do not consider attributes with semantic impact
I noticed that different rewrite rules and equations sometimes get the same `UNIQUE_ID` if they only differ in their attributes, such as `concrete()` or `symbolic()`.

To demonstrate this:

File `test.k`
```
module TEST
  imports INT

  syntax Foo ::= foo(Int, Int) | bar(Int)

  rule [rule1]:
      foo(X, Y) => bar(X +Int Y) [concrete(X)]
  rule [rule2]:
      foo(X, Y) => bar(X +Int Y) [concrete(Y)]

endmodule
```

```shell
jost@freshcode-1:same-unique-id$ kompile test.k
jost@freshcode-1:same-unique-id$ grep "rule[12]" test-kompiled/definition.kore  | grep -e "UNIQUE'Unds'ID{}([^)]*)"
  [UNIQUE'Unds'ID{}("47a45c5c997535c369cb063e28057d89a0eeacf3c14936fedd9927b91ba75caf"), concrete{}(VarX:SortInt{}), label{}("TEST.rule1"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(7,7,7,33)"), org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/jost/work/RV/code/hs-backend-booster/scratch/same-unique-id/test.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]")]
  [UNIQUE'Unds'ID{}("47a45c5c997535c369cb063e28057d89a0eeacf3c14936fedd9927b91ba75caf"), concrete{}(VarY:SortInt{}), label{}("TEST.rule2"), org'Stop'kframework'Stop'attributes'Stop'Location{}("Location(9,7,9,33)"), org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/jost/work/RV/code/hs-backend-booster/scratch/same-unique-id/test.k)"), org'Stop'kframework'Stop'definition'Stop'Production{}("syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions), symbol]")]
```
The two rules labelled `rule1` and `rule2` differ in their attributes (and rule label) but have the same `UNIQUE_ID` value.

Obviously not all attributes should be considered when computing the unique ID, but those which have an effect on how the rule is applied should IMHO have an influence on the ID.

gtrepta:
fixes #3984

This fixes an issue where the `UNIQUE_ID` attribute was relied on to trace applications of different rules, but some rules that were identical aside from some certain attributes had the same id and thus were difficult to distinguish from each other.

### DIFF
--- a/kernel/src/test/java/org/kframework/compile/NumberSentencesTest.java
+++ b/kernel/src/test/java/org/kframework/compile/NumberSentencesTest.java
@@ -1,0 +1,50 @@
+// Copyright (c) Runtime Verification, Inc. All Rights Reserved.
+package org.kframework.compile;
+
+import static org.junit.Assert.*;
+import static org.kframework.kore.KORE.*;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.Test;
+import org.kframework.attributes.Att;
+import org.kframework.builtin.BooleanUtils;
+import org.kframework.definition.Rule;
+
+public class NumberSentencesTest {
+
+  private static final List<Att> testAtts =
+      Arrays.asList(
+          Att.empty(),
+          Att.empty().add(Att.ANYWHERE()),
+          Att.empty().add(Att.CONCRETE(), "A"),
+          Att.empty().add(Att.CONCRETE(), "B"),
+          Att.empty().add(Att.NON_EXECUTABLE()),
+          Att.empty().add(Att.OWISE()),
+          Att.empty().add(Att.PRIORITY(), "50"),
+          Att.empty().add(Att.PRIORITY(), "100"),
+          Att.empty().add(Att.SIMPLIFICATION()),
+          Att.empty().add(Att.SYMBOLIC(), "A"),
+          Att.empty().add(Att.SYMBOLIC(), "B"));
+
+  @Test
+  public void testHash() {
+    Rule r =
+        new Rule(
+            KRewrite(KApply(KLabel("foo")), KApply(KLabel("bar"))),
+            BooleanUtils.TRUE,
+            BooleanUtils.TRUE,
+            Att.empty());
+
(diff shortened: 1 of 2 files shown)


kernel/src/test/java/org/kframework/compile/NumberSentencesTest.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final List<Att> testAtts =
+      Arrays.asList(
+          Att.empty(),
+          Att.empty().add(Att.ANYWHERE()),
+          Att.empty().add(Att.CONCRETE(), "A"),
+          Att.empty().add(Att.CONCRETE(), "B"),
+          Att.empty().add(Att.NON_EXECUTABLE()),
+          Att.empty().add(Att.OWISE()),
+          Att.empty().add(Att.PRIORITY(), "50"),
+          Att.empty().add(Att.PRIORITY(), "100"),
```
gtrepta, line 27: Something I noticed here. The priority attribute is supposed to be an integer, but it's a string after outer parsing and isn't converted to an integer until `ModuleToKORE` stringifies the definition. So `NumberSentences` should only see a string here.

Scott-Guest: Nice catch, there are similar pain points with any of the attribute values that require parsing/un-parsing - I'd like to fix it with #4003

Scott-Guest (approved):
LGTM.

I can't fully confirm whether the list of semantics-affecting attributes is exhaustive, but it seems reasonable, and we can patch later if anyone notices an issue.

PR #4008, opened 2024-02-19
Remove `Att` methods based on `Class` rather than `Att.Key`

Scott-Guest:
This PR removes all methods from `Att` which refer to attributes by their value type (e.g. `Source.class`, `Location.class`, etc.) rather than by their `Att.Key`.

As a result, we maintain the nice property that usages of an attribute exactly correspond with usages of the whitelisted `Att.Key`.


gtrepta (approved):
Looks like a good change to me.

PR #4011, opened 2024-02-20
Refactor allowed syntax checks for attributes

### LINKED ISSUE #3982: Check cell properties for restricted attributes
The attribute infrastructure fails to check cell properties for restricted attributes, allowing nonsense such as
```
configuration <myCell macro=""> .K </myCell>
```

Scott-Guest:
Fixes #3982 

Previously, each attribute specified an allowed set of `Class[_]`s where that attribute may occur. However, both a cell and a non-cell are just `Production`s, so no `Class[_]` restriction is sufficient to distinguish cells.

This PR then refactors to instead have an allowed set of `Att.Syntax` types, each with a `isA(term: AnyRef): Boolean` method to check membership.

In process, we also:
- Restrict `label(_)` to rules, claims, contexts, and context aliases rather than allowing it on any sentence
- Restrict `group(_)` to productions
- Restrict `cell`, `cellName`, `exit`, `stream`, `multiplicity`, and `type` to cells

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckAtt.java
@@ -87,7 +87,7 @@ public class CheckAtt {
     Set<Key> badKeys =
         stream(term.att().att().keySet())
             .map(Tuple2::_1)
-            .filter(k -> k.visibility().isPermitted(term))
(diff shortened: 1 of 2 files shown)


Scott-Guest:
Parking this until we can move CheckAtt to immediately after parsing

PR #4015, opened 2024-02-20
Run actual smoke tests in Nix

### LINKED ISSUE #4012: Add smoke test for `kore-rpc-booster`
To avoid crashes due to diverging dependency versions between the LLVM and the Booster backend on Nix:

```
 kevm-test> user error (dlopen: /nix/store/9v5d40jyvmwgnq1nj8f19ji2rcc5dksd-glibc-2.37-45/lib/libc.so.6: version `GLIBC_2.38' not found (required by tests/specs/benchmarks/verification/haskell/llvm-library/interpreter.so))
```

Baltoli:
In https://github.com/runtimeverification/k/issues/4012, we noticed that the smoke test for the booster integration that @jberthold added previously was never actually getting run under the Nix test configuration; the "smoke tests" run there were just compiling an empty module.

This PR runs the _actual_ K smoke tests, which subsume the behaviour that was already being tested along with some actual features.

I have verified that the issue in #4012 would get reproduced by these tests, ~~so this is blocked until @goodlyrottenapple's changes to realign Nixpkgs are merged.~~


jberthold (approved):
LGTM.
Tried to get the build over the line... there seem to be network problems with the self-hosted runners today.

PR #4016, opened 2024-02-20
No longer emit old overload attribute

Baltoli:
The backends have now all been changed to accept the new `symbol-overload` KORE attribute (https://github.com/runtimeverification/llvm-backend/pull/984, https://github.com/runtimeverification/haskell-backend/pull/3723) and so we can safely stop the frontend from emitting the old `overload` syntax.

Downstream testing on the C semantics (which makes heavy use of overloading) has passed, so I'm confident this change is correct.


gtrepta (approved):
LGTM.

PR #4017, opened 2024-02-20
[DNM] Make `.K` the preferred option for empty K sequences

Baltoli:
This change makes `.K` the preferred syntax for empty K sequences by:
* Adding the `deprecated` attribute to the `.` production.
* Updating the **frontend's** unparser to prefer printing empty K sequences as `.K`

Before the changes described below are merged, downstream semantics should be proactively updated to use `.K` rather than `.` as far as possible to minimise the impact on semantics developers.[^1] These changes have been implemented in:
* [ ] C https://github.com/runtimeverification/c-semantics/pull/1095
* [x] EVM https://github.com/runtimeverification/evm-semantics/pull/2309
* [x] MIR _already up to date; no uses of `.`_
* [x] WASM
  * https://github.com/runtimeverification/wasm-semantics/pull/581
  * https://github.com/runtimeverification/mx-semantics/pull/204
* [x] PL Tutorial https://github.com/runtimeverification/pl-tutorial/pull/13
* [x] Pyk https://github.com/runtimeverification/pyk/pull/902

Once downstream semantics have been updated, there is a subtlety with the unparser that needs to be addressed carefully. Unparsing can be performed either by the frontend, or by the LLVM backend's `kprint` binary called via `kore-print`. This means that there are two separate places that need to be updated to get the new, correct `.K` unparsing behaviour, and a dependency update job needs to go through to get one of them into the frontend.

I have therefore opened three separate PRs:[^2]
- https://github.com/runtimeverification/k/pull/4017: apply the deprecation warning and update test outputs that don't use the LLVM backend unparser
- https://github.com/runtimeverification/k/pull/4019: update test outputs that _do_ use `kprint`
- https://github.com/runtimeverification/llvm-backend/pull/990: make the LLVM backend's `kprint` print `.K`

Merging all of these together will take some care, but I think the best thing to do is:
* Merge the LLVM backend PR; it's standalone and passes LLVM backend CI.
* Then, cherry-pick the following PRs into the ensuing submodule update job:
  * https://github.com/runtimeverification/k/pull/4017
  * https://github.com/runtimeverification/k/pull/4019
* This should produce a submodule update PR that updates all the regression tests at once, and marks `.` as deprecated.

I have verified that this CI process passes on this stand-in PR: https://github.com/runtimeverification/k/pull/4020 (shouldn't be merged as-is because it bypasses the submodule job).

Related to https://github.com/runtimeverification/k/issues/3964

[^1]: It's unscientific, but the best way to do this is to build each semantics using this PR's branch of K and seeing what deprecation warnings you get.
[^2]: Making the separate PRs makes it slightly easier to test the two different sets of changes independently.


Baltoli:
Smoke tests need to be fixed also.

PR #4036, opened 2024-02-23
Fix handling of unary `symbol(_)` on user syntactic lists

### LINKED ISSUE #3996: Kompiler fails on syntactic list production with unary `symbol` attribute
K version: `v6.2.27`

```k
// int-list.k
module INT-LIST
    imports INT-SYNTAX
    syntax Ints ::= List{Int, ","} [symbol(int-list)]
endmodule
```

Kompile:

```
$ kompile int-list.k --syntax-module INT-LIST
```

Output:

```
[Error] Compiler: The 1-argument form of the `symbol(_)` attribute cannot be
combined with `klabel(_)`.
        Source(/home/ttoth/git/pyk/int-list.k)
        Location(3,21,3,54)
        3 |         syntax Ints ::= List{Int, ","} [symbol(int-list)]
          .                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The root cause is probably that when de-sugaring the list production, the kompiler generates productions with attributes

```k
[klabel(_), symbol(_)]
````

We should check the compiler pipeline for instances where `Att.KLABEL` is written, and eliminate them.

### PR DESCRIPTION
**Pull request by Baltoli**
Previously, if users applied the `symbol(_)` attribute to their syntactic list productions, it would collide with the implicitly added `klabel(_)` for the list's terminator production.

This PR fixes the issue by updating the syntactic list desugaring process to instead add a `symbol(_)` attribute. Additionally:
* The `terminator-klabel(_)` attribute is renamed to `terminator-symbol(_)` in line with the desugaring process's behaviour.
* The associated documentation is updated, and a new test is added to check that terminator symbols don't collide with other symbols.
* A regression test for #3996 is added.

Fixes https://github.com/runtimeverification/k/issues/3996

### CODE REVIEW

**Comment by Baltoli** [comment]
There are a couple more places in the PL tutorial that need to be fixed here; marking as a draft

File: kernel/src/main/java/org/kframework/kompile/Kompile.java [hunk 1]
```diff
@@ -545,7 +545,7 @@ public void structuralChecks(
       Option<Module> kModule,
       Set<Att.Key> excludedModuleTags) {
     checkAnywhereRules(modules);
-    checkOverloads(modules);
+    checkOverloads(mainModule);
```
Scott-Guest, line 548: Not quite following - what's the motivation for only checking `mainModule`?

Scott-Guest (approved):

kernel/src/main/java/org/kframework/kompile/Kompile.java
Baltoli: The short answer is that there was previously a subtle mismatch between how we calculated the overload lattice, and how we checked which productions are valid non-singleton overloads..

It's a little bit contrived, but here's an example I minimised from the PL tutorial:
```k
// test.k

module TEST-SYNTAX
  imports INT-SYNTAX
  imports ID-SYNTAX

  syntax Val ::= Int
  syntax Vals

  syntax Exp ::= Val | Id
  syntax Exps ::= Vals
endmodule

module TEST
  imports TEST-SYNTAX
  imports DOMAINS
  imports A

  syntax Vals ::= List{Val, ","} [overload(exps)]
endmodule

module A
  imports TEST-SYNTAX

  syntax Exps ::= List{Exp, ","} [overload(exps)]
endmodule
```

Here, if you check `A` in isolation you get a false positive that the production for `Exps` doesn't overload anything, despite the fact that we **calculate** the overload lattice globally and so this overload is actually correct.

PR #4047, opened 2024-02-26
Revamp installation instructions

geo2a:
* remove the outdated `kup update` command
* remove the 30min installation time estimate --- should be much faster with the binary cache (* needs measuring) 
* reword a little


k-distribution/INSTALL.md
```diff
@@ -7,33 +7,35 @@ Installing the K Framework
 Fast Installation (preferred)
 -----------------------------
 
-If you're on a system that supports [Nix](https://nixos.org/download.html),
-use this command to install via Nix:
+The preferred way to install K is the `kup` tool, which is based on [Nix](https://nixos.org/download.html).
+No knowledge of Nix is required to use `kup`. If you are on Apple Silicon, `kup` is currently the only way to install K.
```
Baltoli: This isn't the case any more:
```suggestion
No knowledge of Nix is required to use `kup`.
```

k-distribution/INSTALL.md
```diff
@@ -7,33 +7,35 @@ Installing the K Framework
...
+
+Install `kup` and `K` by running the following:
 
 ```shell
 bash <(curl https://kframework.org/install)
 kup install k
 ```
 
-You can update K with:
+And list available versions with:
```
Baltoli:
```suggestion
List available versions with:
```

Baltoli (approved):

Issue #4041, opened 2024-02-24
Java 17 (still) not supported in `kore` submodule


Scott-Guest:
The `kore` submodule seems unable to compile any Java 17 features. Compare the two commits in #4039 , where a `record` class compiles just fine in the `kernel` submodule, but the same exact file fails to parse when moved to `kore`.

I thought I solved this with #4032, however, after doing a fresh clone I'm still encountering the same errors. This will be a blocker for #4038, #4030, and #4022.

Scott-Guest:
I'm working to produce a minimal maven project example, but am pretty at a loss why this is happening. I'd appreciate some assistance from anyone more familiar with Maven.

Baltoli:
That's really odd; will take a look and see if I can figure out what's going on!

Scott-Guest:
~~Finally narrowed this down! While it's not mentioned in any changelog I can find, this appears to be a bug in `scala-maven-plugin` that was fixed in version 4.0.0 (we were using 3.3.1).~~

Ugh looks like this is still not sufficient - bumping the version fixes a minimized reproduction I have, but doesn't fix #4039.

Scott-Guest:
After digging through Scala's source code a bit, my understanding is that
- `scala-maven-plugin` is backed by `zinc`, which is a wrapper around the Scala compiler for incremental compilation.
- If only Java source files are provided, `zinc` just delegates to `javac`. Otherwise, it delegates to `scalac`.
- To support Java-Scala interop, `scalac` needs the type signatures and symbol information for the Java code:
   - In `JavaThenScala` mode, where Scala code may depend on Java code but not vice-versa, `scalac` just delegates to `javac` and unpickles the symbol information from the compiled `.class` files.
   - In the default `Mixed` mode, where Scala code and Java code may both depend on each other, `scalac` instead has to parse the Java source code using its own internal Java parser.

This explains why the minimal example I was investigating was fixed by just bumping the `scala-maven-plugin` version - I had only included Java source files, so it was just delegating to `javac` which of course properly supported Java 17. 

However, for K itself, we have inter-dependent Scala and Java code, so the errors were actually arising due to lack of support for Java 17 in `scalac`'s internal Java parser.

[...]

PR #4055, opened 2024-02-28
Miscellaneous Scala clean up

Scott-Guest:
This PR makes a couple quick miscellaneous changes:
- Updates `maven-compiler-plugin` and `scala-maven-plugin` to their latest versions
- Set the Java version with `<release>` rather than `<source>` and `<target>`
   - `<source>` and `<target>` only set the source code and generated class file formats, but do not take into account API changes between Java releases
   - This `<release>` tag should automatically be respected by `scala-maven-plugin` as well (once #4041 is fixed)
- Disable various Scala language extensions and enable `-Werror`
   - Not strictly necessary, but I figure this will minimize complications when we port to Scala 2.13.x and later Java
- Delete the `KORE.Att()` sugar for `Att.empty()`
   - Unrelated, the indirection just seemed pointless
- Delete `ScalaSugar`
   - It's only purpose was for implicit conversions, which requires a Scala language extension

The commit history is clean, and I recommend reviewing this one commit at a time.


Baltoli (approved):

Baltoli:
https://github.com/NixOS/nixpkgs/pull/292323 for posterity, I implemented a feature in our backported maven builder that we can hopefully get upstreamed back into Nixpkgs

Baltoli:
Going to hold off on merging this until @goodlyrottenapple can weigh in; the fixed-output derivation hash failure introduced in https://github.com/runtimeverification/k/pull/4055/commits/8fe2f88a86d74eb9e2e311a33f677f133b814c3c is something we've seen before with this Nix / Maven setup. As I recall, the issue is that any change to the `kore` package[^1] ends up breaking the hash because the corresponding deployed jar file ends up being deployed and re-downloaded when the FOD is built. I may be misremembering, though!

We discussed pulling this module out in #4063; doing so might be worth doing in order to make the Scala KORE parser a proper first-class dependency (and therefore avoid this problem).

[^1]: I _think_ this effect is localised to that module because the required hash doesn't change in https://github.com/runtimeverification/k/pull/4055/commits/ed0fe1f99db95931262b51c7a2715f7b85a6eabe; that commit only changes files outside of the `kore` package.

gtrepta (approved):
lgtm

Baltoli:
Confirmed with @goodlyrottenapple that extracting the code we currently push as a JAR to maven repositories to its own repo, and releasing versioned JARs rather than snapshots. In the interests of letting @Scott-Guest's Scala upgrade work go through, I'm going to merge this PR even though doing so will mean dealing with hash mismatches for a few days.

PR #4056, opened 2024-02-28
Update `maven-compiler-plugin` and `scala-maven-plugin`

Scott-Guest:
Part of #4041

This just updates the `maven-compiler-plugin` and `scala-maven-plugin` versions, which in turn also required an update to `maven-dependency-plugin`.


Scott-Guest:
Superseded by #4055

PR #4069, opened 2024-03-01
Add translation layer from pattern matching exceptions

Baltoli:
Related to: https://github.com/runtimeverification/llvm-backend/issues/999

We are working on removing the backwards dependency from the LLVM backend to K; as part of that change we are removing usages of the frontend error handling mechanisms from the backend: https://github.com/runtimeverification/llvm-backend/pull/1001

Doing so requires that we implement a translation layer from the new error types in https://github.com/runtimeverification/llvm-backend/pull/1001 to the frontend infrastructure; by doing so we can recover the frontend's original behaviour in response to errors / warnings thrown or created during pattern matching.

This PR can be reviewed as-is in combination with the backend PR above; I will take care of the required lockstep merging to get both changes landed to the frontend together.

### DIFF
--- a/kore/src/main/java/org/kframework/utils/errorsystem/KEMException.java
+++ b/kore/src/main/java/org/kframework/utils/errorsystem/KEMException.java
@@ -2,6 +2,7 @@
 package org.kframework.utils.errorsystem;
 
 import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


tothtamas28 (approved):

Baltoli:
Merged into and superseded by: https://github.com/runtimeverification/k/pull/4071

PR #4074, opened 2024-03-04
add pushList function to LIST module in k prelude

dwightguth:
Adds a function to the k prelude LIST module which prepends an item to a list. This can be implemented more efficiently than the generic list append operator that appends two lists.


ehildenb:
LGTM, but I'm not sure about the `pom.xml` change, so will let @Baltoli take a look at it.

ehildenb:
Also, please add a description to the PR body.

Baltoli (approved):

PR #4080, opened 2024-03-06
Restore pretty-printing for exhaustiveness

Baltoli:
Previously, this logic lived in the LLVM backend's matching compiler. When we eliminated the backwards dependency edge between the backend and the frontend, this meant we couldn't use the frontend pretty-printing logic to make the error message in the pattern matcher.

This commit reinstates equivalent pretty-printing logic in the frontend.

Blocked on https://github.com/runtimeverification/llvm-backend/pull/1007 reaching the frontend.

### DIFF
--- a/kore/pom.xml
+++ b/kore/pom.xml
@@ -16,6 +16,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.runtimeverification.k</groupId>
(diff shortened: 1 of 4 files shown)


Baltoli:
https://github.com/runtimeverification/k/pull/4076 subsumed by

PR #4085, opened 2024-03-07
Drop support for deprecated `--directory` flag

### LINKED ISSUE #4083: Delete `--directory` deprecated option from `krun` and sort the options in `--help`
`--directory` has been deprecated for a while, and right now, it's returning an error when used. IIRC, the codebase is already using `--definition`, and the semantics should also be using it, as the first one isn't working anymore. Also, let's bind `-d` to `--definition` as we won't have the `--directory` anymore!

Also, the lexicographical order of help options is wrong, as well:

```
      --dry-run            Do not execute backend, but instead print the
                           command that would be executed to stdout.
      --haskell-backend-command CMD  use CMD instead of kore-exec to invoke
                                     Haskell backend
  -h, --help               Display this help and exit
      --definition DIR     Exact path to the kompiled directory.
      --execute-to-branch  Execute until a non-deterministic state is reached, then exit.
                           Requires the --enable-search option to have been passed to kompile
                           on the LLVM backend.
```

We should fix it to be easy to find the `--definition` option.

Scott-Guest:
Closes #4083

- Drop support for `--directory` and remove all corresponding logic
- Make `-d` an alias for `--definition`
- Alphabetize all the help options

### DIFF
--- a/k-distribution/src/main/scripts/bin/kore-print
+++ b/k-distribution/src/main/scripts/bin/kore-print
@@ -42,7 +42,7 @@ $KORE_PRINT arguments:
       --color [on|off]     Enable/disable ANSI color codes. Overrides default,
                            which is determined based on whether stdout is a
                            terminal
-      --definition DIR     Exact path to the kompiled directory.
+  -d, --definition DIR     Exact path to the kompiled directory.
   -h, --help               Display this help and exit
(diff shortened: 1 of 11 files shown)


Robertorosmaninho (approved):
LGTM! Thanks for working on this!

gtrepta (approved):

Baltoli (approved):

PR #4089, opened 2024-03-11
Remove explicit deletion of frontend-only attributes

tothtamas28:
Related:
* #3946
* #3983

Removes explicit deletion of frontend-only attributes `left`, `right` and `originalPrd`.

### DIFF
--- a/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -1768,8 +1768,6 @@ public class ModuleToKORE {
           }
         }
         att = att.add(Att.PRIORITIES(), KList.class, KList(lessThanK));
-        att = att.remove(Att.LEFT());
-        att = att.remove(Att.RIGHT());
         att =
             att.add(
                 Att.LEFT_INTERNAL(),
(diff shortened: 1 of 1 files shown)


tothtamas28:
Removal of `priorities` breaks `TestMiniKEVM` in `pyk`.

The attribute is not listed in #3946 as one consumed by any of the backends. Before I look deeper into the failure, @geo2a can you confirm that the attribute is indeed not used by the old Haskell backend?

**Edit**: Never mind, removing a few more attributes fixed the test again. There might be a hash set of symbols in the backend the iteration order of which influences RPC simplification / execution.

tothtamas28:
https://github.com/runtimeverification/k/actions/runs/8233491604/job/22513069992?pr=4089

```
k> set -o pipefail; (cat sum.imp.in 2>/dev/null || true) | /nix/store/5bs426yn36znn5kfj86c3n4ywxkf6rlg-k-6.3.0-5bafa19b3acdcaa62c71a01963a04560f413e9ed/bin/krun sum.imp --no-exc-wrap  --definition ./imp-kompiled | diff - sum.imp.out
k> 6,7c6,7
k> <     ( n |-> 0 )
k> <     ( sum |-> 5050 )
k> ---
k> >     n |-> 0
k> >     sum |-> 5050
```

Is it possible some of these attributes are in fact required by some of the backends for pretty-printing?

Baltoli:
Yes, I think these attributes are indeed consumed by the pretty-printer to do parenthesisation correctly.

tothtamas28:
> Yes, I think these attributes are indeed consumed by the pretty-printer to do parenthesisation correctly.

@Baltoli, could you update the [attributes table](https://docs.google.com/spreadsheets/d/1_lX1NqjphKBdTtWwiOQC8Xt7yY565xlgS9iGy1JkR6A/edit#gid=39879701) with the missing attributes? (Please also include a type, e.g. `axiom`, `symbol`, etc.)

Baltoli:
Done; `left`, `right` and `priorities` are all used by the LLVM backend's unparser to correctly decide when to insert parentheses.

Scott-Guest (approved):

PR #4095, opened 2024-03-13
Fix version parsing script

### LINKED ISSUE #4059: Cannot install K on `ubuntu-latest` GitHub Actions runner
The following job installs K from the Jammy package on an `ubuntu-latest` GitHub Actions runner: https://github.com/runtimeverification/pyk/actions/runs/8079767562/workflow?pr=929#L124

When `kompile --version` is run, it fails with an error: https://github.com/runtimeverification/pyk/actions/runs/8079767562/job/22074896470?pr=929

```
$ java --version
openjdk 17.0.10 2024-01-16
OpenJDK Runtime Environment (build 17.0.10+7-Ubuntu-122.04.1)
OpenJDK 64-Bit Server VM (build 17.0.10+7-Ubuntu-122.04.1, mixed mode, sharing)
$ kompile --version
Error: K requires Java 17 to run but the detected version is 11.
Please either add Java 17 bin directory to the PATH or set the JAVA_HOME
environment variable accordingly.
```

## Analysis

The issue is potentially  with how `checkJava` processes the version string.

First, it tries to look up the version by running [`org.kframework.main.JavaVersion`](https://github.com/runtimeverification/k/blob/0179b261340c73decda52ce8971942ea6a69b4da/kernel/src/main/java/org/kframework/main/JavaVersion.java):

https://github.com/runtimeverification/k/blob/0179b261340c73decda52ce8971942ea6a69b4da/k-distribution/src/main/scripts/lib/checkJava#L29

Then, as a fallback, it runs `java -version`:

https://github.com/runtimeverification/k/blob/0179b261340c73decda52ce8971942ea6a69b4da/k-distribution/src/main/scripts/lib/checkJava#L39

But the two strings have a very different format, so probably the following processing steps fail at some point:

https://github.com/runtimeverification/k/blob/0179b261340c73decda52ce8971942ea6a69b4da/k-distribution/src/main/scripts/lib/checkJava#L51

### LINKED ISSUE #4061: Move `checkJava` logic into Java code
When running `kompile` (and potentially other tools of the K toolkit), a script [`checkJava`](https://github.com/runtimeverification/k/blob/0179b261340c73decda52ce8971942ea6a69b4da/k-distribution/src/main/scripts/lib/checkJava#L4) is run to check version compatibility of the installed JRE with K.

If possible (i.e. if each tool that runs `checkJava` eventually ends up running Java code), consider moving this logic into the Java code base, as introspecting the running JVM for its version is simpler and less error-prone than parsing version strings from diverse sources.

Related: #4059

Baltoli:
This PR applies some minor cleanup to the `checkJava` script's parsing of `java --version` output. The new code will produce a correct error message for Java >= 11, and better handles slightly different formats of version data produced by different JVM implementations.

Fixes #4059
Fixes #4061


k-distribution/src/main/scripts/lib/checkJava
```diff
@@ -36,7 +36,7 @@ else
     setarch
     JAVA=$NG
   else
-    version=$("$_java" -version 2>&1)
+    version=$("$_java" --version)
```
Baltoli, line 37: For some reason, `java --version` prints to `stdout` and `-version` to `stderr`. The two-dash version appears to be the more consistent format as well.

k-distribution/src/main/scripts/lib/checkJava
```diff
@@ -46,15 +46,20 @@ else
   fi
 fi
 
-
+# The Java version information we're parsing here looks something like this:
+#
+#   openjdk 21.0.2 2024-01-16
```
Baltoli, line 49: Previously, `java -version` would - for some implementations - produce something like:
```
openjdk version "21.0.2" 2024-01-16
```
on the first line.

k-distribution/src/main/scripts/lib/checkJava
```diff
@@ -46,15 +46,20 @@ else
...
+#   OpenJDK Runtime Environment Homebrew (build 21.0.2)
+#   OpenJDK 64-Bit Server VM Homebrew (build 21.0.2, mixed mode, sharing)
+#
+# To get the major version, we need to look at the first line's second field,
+# then use a regex to extract the major version component. If the JVM changes
+# the format of the version that gets printed, this script needs to be updated
+# as well.
 if [[ "$_java" ]]; then
```
gtrepta: This condition on `_java` has been here since the file was created. But AFAICT, it will always evaluate to true. Any idea why it's here?

gtrepta (approved):
LGTM. Just one comment.

k-distribution/src/main/scripts/lib/checkJava
Baltoli: Good point; there's one branch where it's not set at the top of the file, so maybe the intention was to leave it unset there and just emit the error later on?

Issue #4030, opened 2024-02-22
Port all Scala code to Java


Scott-Guest:
While Scala is a nice language, the consensus of the frontend team is that maintaining a mixture of Scala and Java causes more friction than it is worth:
- IntelliJ struggles to reason about the Scala code and reports false errors (#3867)
- We have to frequently convert between Java and Scala types (see the usages of [`collections.scala`](https://github.com/runtimeverification/k/blob/08477b587642c382e3ff11d7983431bfa3e8c124/kore/src/main/scala/org/kframework/collections.scala#L17)) and write boilerplate code to offer two versions of the same methods (e.g. `getOption` and `getOptional` in `Att.scala`).
- As a company, my impression is that more of us are familiar with the intricacies of Java than of Scala.

As a result, we have decided that
- All new code should be written in Java
- In the longterm, all existing Scala code should be ported to Java as well

The largest blocker will be the [`matching`](https://github.com/runtimeverification/llvm-backend/tree/34a33b014fbd41fce1059feff37f462c82d5a4c5/matching) library in the LLVM backend, which makes heavy usage of pattern matching and likely can't be ported until we update to Java 21 (#3499).

However, [searching the frontend](https://github.com/search?q=repo%3Aruntimeverification%2Fk+path%3A*.scala&type=Code), there are only 47 Scala files, and many of them can be straightforwardly ported even without pattern matching support in Java.

A high level roadmap:
- [ ] Port Scala case classes not involved in pattern matching to Java record classes inheriting from a sealed interface
    - [`Rich.scala`](https://github.com/runtimeverification/k/blob/08477b587642c382e3ff11d7983431bfa3e8c124/kore/src/main/scala/org/kframework/kore/Rich.scala), [`LabelInfoFromModule.scala`](https://github.com/runtimeverification/k/blob/08477b587642c382e3ff11d7983431bfa3e8c124/kore/src/main/scala/org/kframework/compile/LabelInfoFromModule.scala), [`Location.scala`](https://github.com/runtimeverification/k/blob/08477b587642c382e3ff11d7983431bfa3e8c124/kore/src/main/scala/org/kframework/attributes/Location.scala), etc.
    - The main KORE data structures likely can't be ported due to their use in `matching`
- [ ] Port Scala classes/algorithms not making use of pattern matching
[...]

PR #4103, opened 2024-03-14
Port `POSet` to Java

Scott-Guest:
Part of #4030.

Port the `POSet` class from Scala to Java. 

This is mostly a straightforward conversion from Scala functional idioms (`map`, `filter`, `reduce`, etc.) to the corresponding Java Stream methods. To port `lazy val`s, we also implement a `Lazy<T>` wrapper which caches the result of a `Supplier<T>`. 

The only remaining reference to Scala here is a single constructor which takes a `scala.collection.Set<Tuple2<T, T>>` and internally converts it to a `java.util.Set<Pair<T, T>>`. This can be removed once we more pervasively switch away from Scala collection types everywhere in the codebase.


k-distribution/tests/regression-new/checks/checkCircularList.k.out
```diff
@@ -1,2 +1,2 @@
 [Error] Compiler: Had 1 parsing errors.
-[Error] Compiler: Illegal circular relation: Exp < ExpList < Exp
+[Error] Compiler: Illegal circular relation: ExpList < Exp < ExpList
```
Scott-Guest, line 2: The output is unstable because it's based on the iteration order of `relations.entrySet()`, but I don't see a clean way to address this.

Baltoli: We can revisit this in the future if it becomes a worse problem; this code is pretty stable and so long as we aren't doing further big changes here we should be fine!

kore/src/main/java/org/kframework/POSet.java
Baltoli, line 1: I'm pleasantly surprised by how nicely the Java code comes out to be here!

kore/src/main/java/org/kframework/SerializableSupplier.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Runtime Verification, Inc. All Rights Reserved.
+package org.kframework;
+
+import java.io.Serializable;
+import java.util.function.Supplier;
+
+@FunctionalInterface
+public interface SerializableSupplier<T> extends Supplier<T>, Serializable {}
```
Baltoli, line 8: Where does the need for `Serializable` come in? I presume you've thought this all through, just curious what the technical reason is!

Baltoli (approved):
Looks good; I'm impressed by how readable the Java translation is and how minimal the API surface change is. Nice work!

k-distribution/tests/regression-new/checks/checkCircularList.k.out
gtrepta: If we care more that there's a circularity, and less what the circularity is, we could `sed` out the example given in the error message.

kore/src/main/java/org/kframework/SerializableSupplier.java
Scott-Guest: The root of the issue is that lambdas and method references are not serializable by default. Instead, they are only serializable when constructed in a context whose target type is `Serializable` (https://stackoverflow.com/questions/25391656/serialization-of-a-lambda-after-its-creation).

In particular, this means that the `Supplier<T>` in each `Lazy<T>` in the `POSet` is not guaranteed to be serializable, giving errors when trying to serialize the overall `POSet`. We fix this by instead requiring `SerializableSupplier<T>`.

As for why the overall `POSet` needs to be serializable in the first place, this was admittedly error driven development - the original `POSet` class extends `Serializable`, and removing it causes tests to fail!

PR #4104, opened 2024-03-15
Factor out format processing for records into `defaultFormat`

### LINKED ISSUE #4086: Do not modify user-defined `format` for prefix productions
## Preliminaries

A prefix production is a production of the form

```k
syntax S ::= foo ( S1, S2, ..., Sn )
```

https://github.com/runtimeverification/k/blob/bea8cf38cf5963aa47bef31250edcfcadf2905c5/kore/src/main/scala/org/kframework/definition/outer.scala#L839

A lesser known K syntactic feature is that labels are allowed on non-terminals:

```
syntax S ::= foo ( x1: S1, x2: S2, ..., xn: Sn  )
```

When generating the KORE `format` attribute in `ModuleToKORE`, such productions are special cased.
1. If all non-terminals are labelled, the labels are prepended with a `:` to the corresponding `%i` escape sequence.
    https://github.com/runtimeverification/k/blob/bea8cf38cf5963aa47bef31250edcfcadf2905c5/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1708
2. If the KAST `format` contains `(`, then `... ` is inserted after
   https://github.com/runtimeverification/k/blob/bea8cf38cf5963aa47bef31250edcfcadf2905c5/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1730

These two adjustments ensure that the pretty format is parsable:

```k
module PREFIX
    imports INT
    syntax Int ::= mul ( x: Int, y: Int ) [function, symbol(mul), no-evaluators]
endmodule
```

```bash
$ echo 'Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))' | kprint prefix-kompiled /dev/stdin
mul (...  x: 1 , y: 2 )
$ echo 'mul(... x: 1, y: 2)' | kast /dev/stdin --definition prefix-kompiled --sort Int --input rule --output kore
Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))
```

## The problem

These adjustments to the `format` attribute makes sense for the default `format`: https://github.com/runtimeverification/k/blob/bea8cf38cf5963aa47bef31250edcfcadf2905c5/kernel/src/main/java/org/kframework/unparser/Formatter.java#L136

However, they are applied even to a user-defined `format`:

```
syntax Int ::= mul ( x: Int, y: Int ) [function, symbol(mul), no-evaluators, format("%3 (")]
```

```
symbol Lblmul{}(SortInt{}, SortInt{}) : SortInt{} [format{}("x: %1 (... "), ...]
```

This is not parseable (the user pretty much gives up on that by defining a custom `format`), but it's also not what the user intends.

## Possible fixes

1. Change `defaultFormat(int)` to `defaultFormat(Production)` or even `Production.defaultFormat()`, and handle the labels and ellipses for prefix productions there. Then the special casing will only apply for non user-defined `format` attributes. Also, this simplifies `ModuleToKORE` significantly. We can also emit a warning if the user puts `format` on a prefix production. 
2. Emit an error if the user puts `format` on a prefix production.

tothtamas28:
Fixes #4086

* Move `defaultFormat` to `Production`
* Implement special case for record productions


tothtamas28:
The PR also fixes a small discrepancy between `kore-print` and `kast`. Additionally, it removes the extra space after `(...`.

```k
module PREFIX
    imports INT-SYNTAX
    syntax Int ::= mul ( x : Int, y : Int ) [symbol(mul), function, no-evaluators]
endmodule
```

Before:

```bash
$ echo 'Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))' | kore-print /dev/stdin --definition prefix-kompiled
mul (...  x: 1 , y: 2 )
$ echo 'Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))' | kast /dev/stdin --definition test-kompiled --input kore --output pretty
mul ( 1 , 2 )
```

After:

```bash
$ echo 'Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))' | kore-print /dev/stdin --definition prefix-kompiled
mul (... x: 1 , y: 2 )
$ echo 'Lblmul{}(\dv{SortInt{}}("1"),\dv{SortInt{}}("2"))' | kast /dev/stdin --definition prefix-kompiled --input kore --output pretty
mul (... x: 1 , y: 2 )
```

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -926,6 +926,24 @@ case class Production(
...
+    if (isPrefixProduction && nonterminals.size > 0 && nonterminals.forall(_.name.isDefined)) {
+      items.zipWithIndex
+        .map {
+          case (Terminal("("), i)              => s"%${i + 1}..."
+          case (Terminal(_), i)                => s"%${i + 1}"
+          case (NonTerminal(_, Some(name)), i) => s"$name: %${i + 1}"
+          case (RegexTerminal(_, _, _), _) =>
+            throw new IllegalArgumentException(
+              "Default format not supported for productions with regex terminals"
+            )
```
Baltoli, line 940: Do we validate this earlier and throw a KEM error if there's a regex terminal without a format attribute?

kore/src/test/scala/org/kframework/definition/OuterTest.scala
Baltoli, line 1: nice!

Baltoli (approved):
One minor question, LGTM otherwise

kore/src/main/scala/org/kframework/definition/outer.scala
tothtamas28: Yes:

```
[Error] Compiler: Expected format attribute on production with regular
expression terminal. Did you forget the 'token' attribute?
        Source(/home/ttoth/git/pyk/test.k)
        Location(2,20,2,28)
        2 |         syntax Foo ::= r"##foo"
          .                        ^~~~~~~~
```

PR #4108, opened 2024-03-18
Reapply "Miscellaneous Scala clean up" (#4078)

Baltoli:
This reverts commit ec70555ac575bdba60ae59c389de3bebbdc47509, which was itself a revert of #4055 because of breakages in our Nix builds because of issues around the `scala-kore` dependency. This PR reinstates the original change now that `scala-kore` has been properly released and versioned, and also makes a fix (see comment below) for an issue with Nix / Maven integration that was producing spurious hash invalidations.


Baltoli:
Not quite ready yet; there's one small issue with the `dependency:sources` command that needs to be fixed

Baltoli:
Fixed in: https://github.com/runtimeverification/k/pull/4108/commits/5ffcf6b09dd7df3030c4d6b62be09fa952161fb4

The issue was that `mvn dependency:sources` was being more aggressive than we want it to be in downloading things, and was pulling JARs from our Maven repository remotely into its cache. This meant that (as you can see from Scott's commits above) this PR was falling foul of the FOD hash-poisoning issue we've seen previously.

The fix is to use `dependency:get -Dclassifier=sources` to get the dependency sources instead; I've verified that the `maven-deps` derivation that gets generated after this change does not include any remotely-downloaded RV snapshots.

Scott-Guest (approved):

PR #4110, opened 2024-03-19
Port `TopologicalSort` to Java

Scott-Guest:
Part of #4030

Ports the `TopologicalSort` class to Java. 

Once again, this is a direct translation, replacing functional idioms with either direct iteration or equivalent `Stream` functions. The only semantic change is returning an `Optional` rather than reporting an error when there's a cycle.


kore/src/main/scala/org/kframework/collections.scala
```diff
@@ -19,7 +19,6 @@ object Collections {
   def immutable[T](s: java.util.Set[T]): Set[T]           = s.asScala.toSet
   def immutable[T](s: java.util.List[T]): Seq[T]          = s.asScala
   def immutable[K, V](s: java.util.Map[K, V]): Map[K, V]  = s.asScala
-  def immutable[T](s: Array[T]): Seq[T]                   = s
```
Scott-Guest, line 22: Removed this because it overlaps with the `immutable[T](s: java.util.List[T]): Seq[T]` overload, making the latter unusable (and it's dead code anyways).

Baltoli (approved):

PR #4111, opened 2024-03-19
Improve `k-which-python` script

Baltoli:
Pyk supports both Python 3.10 and 3.11; for projects that use Pyk and _mandate_ one of these versions, the default behaviour of `k-which-python` may not work for them. This PR improves the script to optionally take a Python version at the command line; the old script behaviour is maintained when no version is passed.


nix/k.nix
```diff
@@ -28,8 +28,6 @@ let
   runtimePath = lib.makeBinPath runtimeInputs;
 
   which-python = ''
-      #!${runtimeShell}
-      echo "${lib.makeBinPath [python3]}/python3"
       '';
```
goodlyrottenapple: `which-python` is redundant now and should be removed.

goodlyrottenapple (approved):

PR #4118, opened 2024-03-20
Add total attribute to projection functions for named nonterminals which are total

dwightguth:
As part of https://github.com/runtimeverification/evm-semantics/pull/2349 we are trying to change how we implement the gas schedule in KEVM in order to improve performance. In order to do this, we introduce a new sort, ScheduleTuple, which has a single constructor with many arguments. We rely on projection functions for named nonterminals in order to extract those arguments. However, this confuses the booster because it is unable to determine that these functions are total. In general, they are not total, however, for the case where the sort they are defined over has a single constructor, they are total. Thus, we introduce logic into GenerateSortProjections which tags these productions as total in the cases where they are in fact total.

This has been tested and shown to fix the performance regression in the booster introduced by the above change. It should be easy to see why this change is sound since it applies the `total` attribute only in the case where the production has a single constructor, and the rule matches unconditionally on any instance of that constructor.


ehildenb:
We should have a test that this works as intended. Ideally something that isn't provable that becomes provable because of this. Also testing that it doesn't suddenly make _everything_ about hte record total would be nice.

kernel/src/main/java/org/kframework/compile/GenerateSortProjections.java
```diff
@@ -28,13 +28,16 @@
 public class GenerateSortProjections {
 
   private Module mod;
+  private final Module mainMod;
   private final boolean cover;
 
-  public GenerateSortProjections(boolean cover) {
+  public GenerateSortProjections(boolean cover, Module mainMod) {
+    this.mainMod = mainMod;
```
Scott-Guest, line 35: Not setting `this.mod` in the constructor here is bug-prone. The `gen` overloads all assume `this.mod` is non-null, but not all of them actually initialize it, e.g., `GenerateSortProjections(false, mainMod).gen(mySort)` would fail.

A better approach might be to remove `this.mod` entirely and have every `gen` overload take an explicit `mod` argument.

Scott-Guest (approved):
Overall LGTM, but ~see my comment and~ I agree with @ehildenb re tests

kernel/src/main/java/org/kframework/compile/GenerateSortProjections.java
Scott-Guest: Ah just realized this was already an issue prior to your change. Feel free to ignore and I'll fix it in a follow up PR.

Baltoli:
I'd like to see a test also but I couldn't figure out a skeleton of what one might look like; I'm not sure how the legacy Haskell backend treats `function, total`.

dwightguth:
I'm open to adding a test, but I don't entirely know how to construct one. The example I'm working with is a performance regression and can't be added as a test because it doesn't actually make anything fail.

Baltoli:
Indeed - if there was a way to get the legacy backend to fail a proof because a particular function is partial (and succeed were it labelled total), then a test would be easy to construct for your change here. @ehildenb any idea how you might construct such an example?

ehildenb:
Hmmmm, the thing I would think of is to make it so the initial configuration of a proof contains a symbol that rewrites in the first step to the previously-partial (now total) function, and the old backend should take one step, and then branch on the `#Ceil(...)` condition associated with the partial function. On the branch where it assumes it's defined, proof passes, on the other branch it fails. I don't think that putting it in the initial state will work, because the backend assumes `#Ceil(...)` of the initial state.

Once it's marked as total, proof passes.

Not 100% sure this would work, but that would be my approach.

dwightguth:
The test has been added. Without this change, the test branches on the Ceil condition; with it, it is able to prove that the term in question is defined and applies the rule unconditionally.

Please re-review.

ehildenb (approved):

Baltoli (approved):

PR #4123, opened 2024-03-21
Compute connected components of `POSet`

Baltoli:
This is a simple implementation of an algorithm that associates elements in a `POSet` with unique integer identifiers for the connected component that they belong to; I want this implementation for a check I'm refactoring but thought it would be easier to review this as a standalone PR.


kore/src/main/java/org/kframework/POSet.java
```diff
@@ -186,6 +186,29 @@ public Optional<T> maximum() {
     return maximumLazy.get();
   }
 
+  private Map<T, Integer> computeConnectedComponents() {
```
tothtamas28, line 190: Please add a Javadoc comment that promptly describes the algorithm and why the greedy approach works (`directRelations` is transitively closed, what the loop ensures on `data`, etc).

kore/src/main/java/org/kframework/POSet.java
```diff
@@ -186,6 +186,29 @@ public Optional<T> maximum() {
...
+    for (var entry : directRelations) {
+      int best = nextComponent++;
+      best = Integer.min(data.getOrDefault(entry.getLeft(), best), best);
+      best = Integer.min(data.getOrDefault(entry.getRight(), best), best);
+
+      data.put(entry.getLeft(), best);
+      data.put(entry.getRight(), best);
+    }
+
+    return data;
```
tothtamas28, line 222: Would be nice to map the values to `0...#comps-1` first.

kore/src/test/java/org/kframework/POSetTest.java
```diff
@@ -80,4 +80,17 @@ public void glb() {
     assertEquals(
         Optional.empty(), new POSet<>(Pair.of(b2, b1), Pair.of(b3, b2), Pair.of(b4, b2)).minimum());
   }
+
+  @Test
+  public void connectedComponents() {
```
tothtamas28, line 85: Consider turning this into a `Parameterized` test with multiple test inputs.

tothtamas28 (approved):
Pretty clever algorithm!

kore/src/test/java/org/kframework/POSetTest.java
Baltoli: Hard to do this as the property being asserted changes depending on the poset structure, but I've added some more tests.

Baltoli:
Addressed review comments and fixed a bug in my initial greedy algorithm, which wasn't quite strong enough.

tothtamas28:
> fixed a bug in my initial greedy algorithm

What was the bug? Because I was able to convince myself that it was correct. :sweat_smile:

Baltoli:
I had wrongly assumed that the iteration over edges would be in topological order, but it's actually in _construction_ order. Consider this test case:
```
new POSet<>(Pair.of(b1, b2), Pair.of(b4, b5), Pair.of(b2, b3), Pair.of(b3, b4));
```

Here, the ordering is `b1 <: b2 <: b3 <: b4 <: b5`, so they're all in the same component. The greedy algorithm would proceed as follows:
```
b1 -> 0
b2 -> 0

b4 -> 1
b5 -> 1

b2 -> 0
b3 -> 0

b3 -> 0
b4 -> 0
```
but this leaves `b5` in component 1 incorrectly.

kore/src/test/java/org/kframework/POSetTest.java
tothtamas28: This solution is fine.

Alternatively, you can represent the expected result as a partition of the elements (e.g. `Collection<? extends List<? extends T>> expected`) and assert

* `for all X in expected . for all x1 in X . for all x2 in X: id(x1) == id(x2)`
* `for all X in expected . for all Y in expected . X.size() > 0 && Y.size() > 0 &&  id(X.get(0)) != id(Y.get(0))`
*  (optionally) `id` is onto `[0, N)`

tothtamas28:
Nice catch!

> but this leaves `b5` in component 1 incorrectly.

Right, but `directRelations` is transitively closed, so `b5` would end up getting id 0, latest when considering `(b1, b5)` or `(b2, b5)` I think.

But I can see the issue for an Y-shaped relation:

```
      3 - 4
     /
1 - 2
     \ 
      5 - 6
```

and e.g. with ordering:

```
(3, 4): 0
(5, 6): 1
(2, 6): 1
(1, 6): 1
(2, 5): 1
(1, 5): 1
(1, 2): 1
...
```

Here, 5 and 6 will never be re-evaluated, so they're stuck with id 1.

Baltoli:
> Right, but directRelations is transitively closed

This isn't the case; there's no `(b1, b5)` entry when we iterate - the `directRelations` set is not closed itself, but is used to construct a closed relation.

kore/src/test/java/org/kframework/POSetTest.java
Baltoli: Ah, very nice - clever setup!

Baltoli: Looking at the [docs](https://www.baeldung.com/parameterized-tests-junit-5), doing this concisely needs JUnit 5 and we're currently using 4 so I may stick with the hand-coded examples for now.

PR #4127, opened 2024-03-25
Refactor "unique overloads" check

Baltoli:
Previously, this check relied on identifying when a set of overloads with the same key had multiple greatest / least elements. This is too strong in the presence of Y-shaped overload sets; the solution is instead to check that all the productions that share an overload key are in the same connected component of the overload POSet.[^1]

Fixes an issue in the WASM semantics (https://github.com/runtimeverification/wasm-semantics/blob/master/wasm.md?plain=1#L62-L65) reported by @virgil-serbanuta, where a spurious warning was being generated: https://runtimeverification.slack.com/archives/C7E701MFG/p1711124927666709

The PR includes a reproduction of Virgil's issue, which breaks with K `6.3.58` and is fixed by this change.

[^1]: Except for user lists, which must partition into precisely two connected components: one for the cons case and one for the terminator.

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -650,21 +650,33 @@ public class Kompile {
   }
 
   private void checkDuplicateOverloads(Module module) {
-    // Collect the set of productions that are the greatest element of an overload set, grouped
-    // by their `overload(_)` key. If any key has multiple greatest elements, then two overload sets
-    // are using the same key. This is not ambiguous from the point of view of the compiler, but may
-    // be confusing to users.
-    var partialOrderHeads =
-        module.overloads().relations().keySet().stream()
+    // Group the overloaded productions in this module by their `overload(_)` attribute, then
+    // examine each group to see if the elements in that group are connected. If elements with the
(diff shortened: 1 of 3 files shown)


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -650,21 +650,30 @@ private void checkSingletonOverload(Module module) {
...
+            .values();
+
+    attributeGroups.forEach(
+        ps -> {
+          var userList = ps.stream().allMatch(p -> p.att().contains(Att.USER_LIST()));
+
+          var groups =
+              ps.stream()
+                  .collect(
+                      Collectors.groupingBy(p -> module.overloads().connectedComponents().get(p)));
```
tothtamas28: Does this compile?
```suggestion
                      Collectors.groupingBy(module.overloads().connectedComponents()::get));
```

tothtamas28 (approved):

kernel/src/main/java/org/kframework/kompile/Kompile.java
Baltoli: It does - missed opportunity by Intellij!

PR #4045, opened 2024-02-26
Deprecate `symbol` and `klabel`

Baltoli:
This PR implements @tothtamas28's warning messages for usages of the deprecated `klabel(_)` and `symbol` attributes (which can now be unambiguously replaced by `symbol(_)` and `overload(_)`).

The actual code changes are pretty minimal, but we need to do a bunch of updates on test cases and related output to keep tests passing. Additionally, we have pre-merged downstream PRs to make sure other projects are not subjected to noisy compiler output as a result of making this change:

Downstream updates:
- [x] C https://github.com/runtimeverification/c-semantics/pull/1125
- [x] EVM https://github.com/runtimeverification/evm-semantics/pull/2378
- [x] WASM https://github.com/runtimeverification/wasm-semantics/pull/666
- [x] MX Backend https://github.com/runtimeverification/mx-semantics/pull/298
- [x] Kasmer https://github.com/runtimeverification/kasmer-multiversx/pull/168
- [x] MIR (new semantics doesn't use `klabel`)
- [x] Pyk (this PR, now that K and Pyk are merged)
- [x] Blockchain Plugin https://github.com/runtimeverification/blockchain-k-plugin/pull/184

Issue #1278, opened 2020-05-12
Document how different modules are produced for parsing


ehildenb:
Right now we have lots of different ways that new modules are produced for parsing. Every time there is a change to the parsing logic, it's often complicated by the fact that not everyone understands how these implicit modules are created.

For example, the `LANGUAGE-PARSING` module is a custom module that is created for parsing programs in the language. How is it constructed? What can it be used for? Another example is the `_$SYNTAX` modules created for each module in the definition.

It would be good to list the modules that are created implicitly and added to the definition, along with (i) how they are created, and (ii) what those modules can be used for.

I think just adding the list here to this issue should be OK, or if it seems appropriate to add it to the parsing/unparsing section of pending-documentation.md it can be added there.

dwightguth:
We don't appear to have touched this issue in the past two years, so I'm closing it. Feel free to open a new issue with more up to date information if you still view this as a priority.

dwightguth:
I've been a little too aggressive here I think. I'll reopen this, but I'm moving it to the backlog as it's clearly not a high priority.

PR #4129, opened 2024-03-25
Fix "singleton overload" check for user lists

Baltoli:
When working on #4045, I identified a case in the C Semantics where an overload set was flagged as being singletons (that is, that the productions in the set were not actually overloads of one another). Despite this warning, removing the `overload(_)` attributes broke compilation by producing a parsing ambiguity.

It turns out that the issue was a difference between the main module and the _disambiguation module_ produced during the inner parsing process. The disambiguation module adds an additional production
```
Es ::= E
```
for user list sorts `Es` that subsorts the list element sort into the list sort. This means that two productions
```
S ::= f(Es)
T ::= f(E)
```
can be overloads of one another.

However, this check was only using the overloads induced by the main module and not those used by the disambiguation module. The fix is simply to compute the disambiguation module[^1] and use its overloads for the checks. Doing so requires some changes to the plumbing; the subsorting production is previously only added when `RULE-LISTS` is imported, but we need to short-circuit that code path in this context.

This PR includes a regression test minimised from the original example.

[^1]: I tried a cheaper fix that just checks parameters of overload-attributed productions to see if they appear in a user list, but this isn't sufficiently general in the case where the overload parameter is actually a sub-/supersort of the user list element, as happens in the C semantics.


Baltoli:
Related: https://github.com/runtimeverification/k/issues/1278

kernel/src/main/java/org/kframework/parser/inner/RuleGrammarGenerator.java
```diff
@@ -334,7 +334,7 @@ public static ParseInModule getCombinedGrammar(
   }
 
   public static Tuple3<Module, Module, Module> getCombinedGrammarImpl(
-      Module mod, boolean isBison, boolean forGlobalScanner) {
+      Module mod, boolean isBison, boolean forGlobalScanner, boolean subsortUserLists) {
```
tothtamas28, line 354: It'd be good to include a Javadoc comment for this signature.

tothtamas28 (approved):

PR #4135, opened 2024-03-26
Add secp256k1 to llvm-libs

goodlyrottenapple:
This change also ensures that the package `k.procps.openssl` and `k.openssl.procps` are exactly the same by sorting the list of llvm-lib paths

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,10 +86,7 @@ jobs:
           script: |
             export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
             kup publish k-framework-binary .#k --keep-days 180
-            kup publish k-framework-binary .#k.openssl --keep-days 180
-            kup publish k-framework-binary .#k.procps --keep-days 180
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -86,10 +86,7 @@ jobs:
           script: |
             export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
             kup publish k-framework-binary .#k --keep-days 180
-            kup publish k-framework-binary .#k.openssl --keep-days 180
-            kup publish k-framework-binary .#k.procps --keep-days 180
-            kup publish k-framework-binary .#k.openssl.procps --keep-days 180
-            kup publish k-framework-binary .#k.procps.openssl --keep-days 180
```
goodlyrottenapple, line 92: I don't think people really install these separately, so we only really need to cache the deps needed for blockchain plugin. perhaps we should even add an alias, something like `k.blockchain-plugin-deps`?

Baltoli: Yeah, let's do that in a separate PR. It also avoids having to update instructions in lots of different places if we change the list of dependency libraries (as is happening here).

Baltoli (approved):

PR #4138, opened 2024-03-27
Separate pre- and post-compilation checks

Baltoli:
When working on #4045, I noticed that a pair of productions like:
```k
syntax KItem ::= disambiguate(KItem, KItem) [overload(disambiguate)]
               | disambiguate(List,  Set)   [overload(disambiguate)]
```
were both being incorrectly flagged as singleton overloads, despite meeting the criteria for being an overload set.

Some investigation reveals that this is due to the overload lattice checks being applied _before_ the compiler inserts subsorting productions into `KItem` for every sort; we therefore don't know that `List <: KItem` when constructing the overload lattice.

The solution is to move the overload checks later in the compilation process, after the compilation pipeline has been run. This PR makes that change, along with some associated documentation and renaming for clarity. It also includes a regression test with the example above; I have verified that the test breaks before this PR.


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -484,14 +484,44 @@ public Rule parseRule(CompiledDefinition compiledDef, String contents, Source so
...
+
+    for (Module m : mutable(modules)) {
+      m.checkSorts();
+    }
+
+    checkOverloads(mainModule);
+
+    if (!errors.isEmpty()) {
+      kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
+      throw KEMException.compilerError("Had " + errors.size() + " errors after compilation.");
```
tothtamas28, line 521: Does the pre-kompilation check handle singular vs plural in the error message depending on `errors.size()`?

tothtamas28 (approved):

kernel/src/main/java/org/kframework/kompile/Kompile.java
Baltoli: Nope, it's just the same implementation as the pre-kompilation check - the plural for 1 error will be reported in both cases.

PR #4143, opened 2024-03-28
Add check for non-symbol productions with `overload(_)` attribute

### LINKED ISSUE #4142: Crash when applying overloading to subsort productions that do not have the `symbol(_)` attribute
When working on #4045, I noticed that there was one case where a direct translation of `klabel(X)` → `overload(X)` did not behave as we expected it to. This is because we missed a behaviour of `klabel(X)` when originally specifying the `overload(_)` migration.

That behaviour is that `klabel(X)` will create an _internal_[^1] `klabel` for productions that would not otherwise have one. The implementation of overloading in ModuleToKORE requires implicitly[^2] that overloaded symbols have an internal `klabel` associated with them. Under the previous implementation of overloading, it was not possible to generate an overloaded symbol without a `klabel`, but the new implementation permits this.

To fix the issue, we should add a check that overloaded subsorting productions have an associated `symbol(_)` attribute so that they have an internal `klabel` during KORE generation.

[^1]: These internal `klabel`s are different to the attribute `klabel(_)`. Roughly speaking, productions that have an internal `klabel` will be generated as KORE symbols, and those that don't will not. For instance, subsorting productions will not usually be KORE symbols.
[^2]: i.e. it crashes when this assumption is violated.

Baltoli:
For full context, see the original issue (#4142). This PR implements a simple check for overloaded non-symbol productions, along with a minimal test case.

Also included in the PR is an example minimised from the C Semantics that helps to explain _why_ we would want to overload on subsort productions (and therefore, why the fix for this issue is a check rather than, say, forbidding the feature).

In brief, the new integration test minimises part of the C expression grammar, in which we have the following sort hierarchy:
```
    Exp
     |
    Val
   /   \
RVal   LVal
```
We also wish to construct a corresponding hierarchy of left-associative, comma-separated syntactic lists over this hierarchy (with the exception being that there is no need for a list of lvalues):
```
    Exps
     |
    Vals
   /
RVals
```

Typically, the solution here would be to implement the list sorts as K syntactic lists. For example:
```
syntax Exps ::= NeList{Exp, ","}
```

This is, of course, just syntactic sugar, and we can write down an equivalent set of productions:
```
syntax Exps ::= Exp
              | Exps "," Exp
```

Note here that the singleton production is a _subsorting_ production that would, by default, not have an associated `klabel`, and therefore would not be generated as a KORE symbol.

However, a collision with a separate part of the C grammar does indeed prevent us from using the `NeList` sugar for this, as we require finer-grained control over the parsing priorities. The issue is with the comma operator:
```
syntax Exp ::= Exp "," Exp [left, group(comma)]
```

We require the property that each element in the overloaded list-sort hierarchy has a higher priority than the comma operator, and so we need to be able to attach groups to both the singleton and comma productions.

When combined with the necessary overloading attributes on the singleton list subsorting productions, this situation triggers the bug fixed by this PR.

Fixes https://github.com/runtimeverification/k/issues/4142


tothtamas28:
Does this change make it possible to reduce the checks / special cases for syntactic list overloads to this seemingly more general case?

Baltoli:
Not really, no - the issue here is different to the other special-casing I added this week because of the collision with the comma operator. That collision is why you end up needing to overload the subsorts as well (the earlier PRs dealt with just having to have the subsorting productions _exist_).

tothtamas28 (approved):

tothtamas28:
This use case is new to me, so I'd like to understand a bit better the consequences of adding `symbol` to a production like

```k
syntax Foo ::= Bar
```

Based on what I observe in `compiled.json` / `definition.kore`:
- It is no longer a subsort production: subsort axioms are not generated.
- It becomes a symbol production: `symbol` is generated.

So:
- The set of symbol and subsort productions is disjoint even if such productions are defined.
- In order to identify a subsort production, it must be checked if `symbol` is present.

Issue #4066, opened 2024-02-29
Merge frontend and Pyk again


Baltoli:
We [dropped Pyk](https://github.com/runtimeverification/k/pull/2858) as a direct component of the K repository in 2022 so that it could exist as a more standalone tool. Since then, our model of the relationship between these components has changed (specifically, we hope in the long term to push more of the features currently implemented in the frontend into Pyk, and for the frontend to gradually become a much smaller set of internal tools).

As part of this process, we probably want to think about re-merging the two repositories. Doing so with minimal disruption to downstream projects will require some thought.

Part of: https://github.com/runtimeverification/k/issues/4063

Baltoli:
@tothtamas28 began experimenting with public runners for Pyk, but doing so is blocked on #4061. We should think about fixing that before any merge begins.

@ehildenb questions the performance of these runners; there is a rate-limiting effect as we are only allocated 15? 60? x4vcpu runners for our open-source code. Beyond a point we will run out of compute, but not clear when that will be. Heavier jobs will benefit from being run on our own runners.

Baltoli:
We discussed this in the K meeting today and arrived at the following set of plans:
Important prerequisite:
- [x] Eliminate flakiness in Pyk CI: runtimeverification/pyk#1017 
- [ ] Get the Pyk-based version of the regression test suite passing at an appropriate success rate: https://github.com/runtimeverification/k/issues/4194
- [ ] Ensure that every downstream repository is correctly syncing its K version from Pyk
- [ ] Freeze changes to the Pyk repository
- [ ] ❗Copy Pyk into the K source tree directly such that K presents a Poetry project compatible with the existing Pyk installation
- [ ] On each downstream repo, change their Poetry code to point at K rather than the old Pyk repository.
- [ ] Once all clients are updated, remove the old Pyk repository and reopen K to changes.

> [!CAUTION]
> When we copy Pyk into the K source tree, we should take care to think through how the K build and installation process works for downstream users, and that we've produced a working version of the merged tree that can be installed locally / in a docker container / via Nix, as appropriate.

Baltoli:
First step of the merge:
- After copying the source code in, still keep K installing as it always has
- Pyk will then just be a standard Poetry project that lives in a subdirectory

We should disable autoupdates from Pyk before freezing its code.

Baltoli:
Post-merge subtleties we should address:
- Test suite filtering; we don't want to run the K tests when the Pyk code changes
[...]

PR #4144, opened 2024-03-28
Drop `.` empty K syntax fully and remove special-casing

Baltoli:
Blocked on #4066; we want to use this PR as an example / test case for merging thorny changes that might cause issues with downstream repositories. That process will require a lot more locking / unlocking with K / Pyk being separated.


Baltoli:
Closing as stale.

PR #4172, opened 2024-04-11
add spec module in definition_args

### LINKED ISSUE #2367: Unable to run claims with `kevm prove`
I have a `lemmas.k` file with a `LEMMAS` module and a `LEMMAS-SPEC` module with a claim `claim-1`. I'm trying to run this claim with the commands
```
kevm kompile-spec lemmas.k --target haskell --main-module LEMMAS --syntax-module LEMMAS -o kompiled
kevm prove lemmas.k --claim LEMMAS-SPEC.claim-1 --definition kompiled
```
but I get the error `ValueError: Claim labels not found: ['LEMMAS.LEMMAS-SPEC.claim-1']`. In other words, it automatically prefixes the claim label with the main module `LEMMAS`, which is incorrect because the claim is in the `LEMMAS-SPEC` module. It's not even possible for the claim to be in the main module because in this case `kevm kompile-spec` fails with the error message `Claims are not allowed in the definition.`

`kevm prove` used to have a `--spec-module` option, but it no longer seems to accept it, so I don't know how else to specify the spec module.

Baltoli:
Closes: https://github.com/runtimeverification/evm-semantics/issues/2367 
- adding `spec_module` to `definition_args`.


PetarMax:
This looks good to me, but it would be great if @tothtamas28 could confirm that that is the right place to add the parameter. @anvacaru, this has been tested with KEVM/Kontrol and works?

tothtamas28:
Currently, @nwatson22 has the best understanding of how command line arguments are handled, I added him as a reviewer.

anvacaru:
Yes, I've tested these changes with a similar example to the one provided in https://github.com/runtimeverification/evm-semantics/issues/2367. The issue should be fixed.

PetarMax:
I can also confirm that the correction works and that the functionality is restored.

PetarMax (approved):

nwatson22:
This looks correct to me. The `KDefinitionOptions` class in `cli/args.py` should be what the command line arguments in `definition_args` are populating, and it's already in there, so I guess this just got out of sync somehow. We have plans to overhaul this so the arguments are only specified in one place, but for now these two should be kept in sync.

nwatson22 (approved):

anvacaru:
Adding `automerge`.

ehildenb:
I don't think this was the correct fix perhaps.... I think maybe the better fix is to include `spec_args` in each of the subargument parsers in KEVM. Going to revert this because it's blocking a KEVM update.

PetarMax:
Oh that was unfortunate, thank you for fixing @ehildenb.

PR #4245, opened 2024-04-13
Add proof exploration and display options to CLI

Baltoli:
Transfer of: https://github.com/runtimeverification/pyk/pull/1043

This PR is for making it easier to work on proofs with default `pyk prove ...` functionalities.

- Factor out and reuse `SpecOptions` in `ProveOptions` class.
- Add options to `pyk prove`:
  - `--max-iterations`: control how many iterations of proof extension to do.
  - `--max-depth`: control the maximum number of steps to take in an `execute` request.
  - `--save-directory`: allow saving off the proof state after executing it.
  - `--temp-directory`: allow setting the temporary directory to use for intermediate artifacts.
  - `--show-kcfg`: display the resulting proof KCFG after executing the proof.
  - Handle the `RuntimeError` thrown by crashes of the call to `kprove --dry-run ...` directly and more gracefully, instead of printing a stack trace.
- Add command `pyk show` which reuses a lot of the same options as `pyk prove`, and calls `pyk prove` in a way that it just loads the proof from disk and displays the KCFG.
- Changes to the `KProve.prove_rpc` functionality:
  - Source many of the options from the `ProveOptions` class, which contains the above options and more.
  - Only initialize a `CTermSymbolic` (and thus an RPC server) if it's needed (will take more than 0 proof steps, and the proof isn't already passed).
  - Allow setting the prefix of the temporary file parsed to for `get_claims_modules`.
- Modify a test in `regression-new` to exercise the new `pyk prove --show-kcfg ...` option.


pyk/regression-new/kprove-haskell/sum-spec.k.out
```diff
@@ -11,3 +11,126 @@ APRProof: 3b7070ca4603e18d26d032c0fe64dd71ffadc2c63dae9d57315e8e471a371bd3
...
+┃  ├─ 5
+┃  │     <generatedTop>
+┃  │       <k>
+┃  │         K_CELL_fc656f08:K
+┃  │       </k>
+┃  │       <counter>
+┃  │         ?_COUNTER_CELL_af8c44a5:Int
+┃  │       </counter>
+┃  │       <sum>
+┃  │         S:Int +Int C:Int +Int N:Int +Int -1 *Int C:Int +Int 1 +Int N:Int +Int -2 *Int N:Int +Int -1 /Int 2
```
ehildenb, line 65: This should be paranthesized, issue created: https://github.com/runtimeverification/k/issues/4253

pyk/src/pyk/__main__.py
```diff
@@ -453,15 +468,39 @@ def create_argument_parser() -> ArgumentParser:
...
+    )
+    prove_args.add_argument(
+        '--max-depth',
+        default=None,
+        type=int,
+        help='Maximum number of steps to take in symbolic execution per basic block.',
+    )
+    prove_args.add_argument(
+        '--max-iterations',
+        default=None,
```
nwatson22: I think the `default=None` can be removed where the type is `int`, etc., just not `bool`.

nwatson22 (approved):

pyk/src/pyk/__main__.py
ehildenb: Done.

PR #4250, opened 2024-04-16
Remove proof field from Prover

nwatson22:
Porting https://github.com/runtimeverification/pyk/pull/960 to K repo.

`APRProver.step_proof` currently assumes the proof being passed in the `APRProofStep` it takes as argument is the same as `self.proof`, i.e. when it calls `self.nonzero_depth`, `self._check_subsume`, and `self._checked_for_bounded` assumes the proof being passed in is always the same for a single `APRProver` instance.

This is a problem in itself, but changing this should also move us closer an implementation that can run `step_proof` on different nodes in parallel, by reducing its dependency on external data. A next step could be initializing a new KCFGExplore every `step_proof` instead of using `self.kcfg_explore` to access the `KoreClient`.

- Removes `proof` field from `Prover` and `APRProver`.
- Makes methods that read/write `self.prover` take a prover argument 
- Proof initialization is performed separately in `init_proof` at the beginning of `advance_proof` rather than in the `Prover` constructor.
- Moves `_checked_for_terminal` and `_checked_for_bounded` into `APRProof`, as they refer to nodes of a specific proof.

### DIFF
--- a/pyk/src/pyk/proof/reachability.py
+++ b/pyk/src/pyk/proof/reachability.py
@@ -180,6 +180,14 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
             and not self.is_bounded(node_id)
         )
 
+    @property
+    def circularities_module_name(self) -> str:
+        return self.module_name + '-CIRCULARITIES-MODULE'
+
+    @property
+    def dependencies_module_name(self) -> str:
+        return self.module_name + '-DEPENDS-MODULE'
+
     def is_init(self, node_id: NodeIdLike) -> bool:
         return self.kcfg._resolve(node_id) == self.kcfg._resolve(self.init)
 
@@ -614,12 +622,8 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
         return refutation
 
 
-class APRProver(Prover[APRProofStep, APRProofResult]):
-    proof: APRProof
-
+class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
     main_module_name: str
-    dependencies_module_name: str
-    circularities_module_name: str
     execute_depth: int | None
     cut_point_rules: Iterable[str]
     terminal_rules: Iterable[str]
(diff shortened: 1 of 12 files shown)


ehildenb:
@nwatson22 you probably based this on `master` and need to rebase your changes onto `develop` instead.

pyk/src/pyk/proof/reachability.py
```diff
@@ -672,64 +678,54 @@ def _inject_module(module_name: str, import_name: str, sentences: list[KRuleLike
                 dependencies_as_rules.append(apr_subproof.as_rule(priority=20))
         circularity_rule = proof.as_rule(priority=20)
 
-        module_name = self.proof.module_name
-        self.dependencies_module_name = module_name + '-DEPENDS-MODULE'
-        self.circularities_module_name = module_name + '-CIRCULARITIES-MODULE'
-        _inject_module(self.dependencies_module_name, self.main_module_name, dependencies_as_rules)
-        _inject_module(self.circularities_module_name, self.dependencies_module_name, [circularity_rule])
+        _inject_module(proof.dependencies_module_name, self.main_module_name, dependencies_as_rules)
+        _inject_module(proof.circularities_module_name, proof.dependencies_module_name, [circularity_rule])
```
ehildenb, line 682: Why is this in `init_proof` and not the prover initialization? This seems more about the prover than the proof.

nwatson22: It's because the module name, and by extension, the dependencies_module_name and circularities_module_name are dependent on the proof IDs, and the subproofs getting injected as rules are also dependent on the proofs, so since I'm no longer initializing the Prover with a proof, I thought I should inject the modules as a separate step for each proof prior to stepping those proofs. Does this make sense or have I misunderstood something?

nwatson22:
> @nwatson22 you probably based this on `master` and need to rebase your changes onto `develop` instead.

@ehildenb I think I did this correctly but could you check to make sure it looks right now?

pyk/src/pyk/proof/reachability.py
ehildenb: Makes sense. I wonder though if that will be a slowdown.... shoudl the `Prover` track all the modules it's injected and not re-inject ones it doesn't need?

ehildenb (approved):

PR #4252, opened 2024-04-16
Pyk: Attribute token locations

### LINKED ISSUE #4187: Locations for attribute tokens.
In runtimeverification/pyk#1052, we're tracking locations for tokens, with the exception of the tokens in attributes. The reason for this is that attributes in KAST don't have their source locations.

However, it's still possible for locations on attribute tokens to be useful in the pipeline before generating the KAST from a definition text. So, we should add support for them.

Once implemented, every token should have a location. With that, we should set up the unit tests to take the expected token and its location, and verify that the source text actually has that token's text as a substring at that location in the source.

gtrepta:
fixes #4187 

The pyk outer lexer now has location information for attribute tokens.

The unit tests for the lexer now check that the token text actually appears at its location in the source text.


pyk/src/pyk/kast/outer_lexer.py
```diff
@@ -190,23 +191,29 @@ class State(Enum):
...
 class LocationIterator(Iterator[str]):
-    _line: int = 1
-    _col: int = 0
+    _line: int
+    _col: int
     _iter: Iterator[str]
+    _nextline: bool
 
-    def __init__(self, x: Iterable[str]) -> None:
+    def __init__(self, x: Iterable[str], line: int = 1, col: int = 0) -> None:
```
tothtamas28: Why the discrepancy in default argument values?

pyk/src/pyk/kast/outer_lexer.py
```diff
@@ -190,23 +191,29 @@ class State(Enum):
...
+        if self._nextline:
             self._line += 1
-            self._col = 0
+            self._col = 1
+        self._nextline = la == '\n'
         return la
 
     @property
     def loc(self) -> Loc:
+        """Returns the line,column of the last character returned by the iterator"""
```
tothtamas28: What if no characters have been returned yet?

pyk/src/pyk/kast/outer_lexer.py
gtrepta: The column is incremented when the iterator returns a character. So when it returns the first character, the location will be at column 1, which is where it's supposed to be.

pyk/src/pyk/kast/outer_lexer.py
gtrepta: You'll get `(1,0)`. There's no use case for getting the location before reading anything, so I didn't really think about it here. This is the only instance where you would get a location with a column of `0`, though, so that could work out as an indicator that no characters have been returned so far.

pyk/src/tests/unit/kast/test_outer_lexer.py
```diff
@@ -636,6 +639,10 @@ def test_lexer(test_file: Path) -> None:
     remaining = next(it, None)
 
     # Then
+    line_offsets = [0] + (lambda x: [x := text.find('\n', x) + 1 for _ in range(text.count('\n'))])(0)  # noqa: F841
+    for tok_text, loc in ((t.text, t.loc) for t in actual_tokens):
+        assert text[line_offsets[loc.line - 1] + loc.col - 1 :].find(tok_text) == 0
```
tothtamas28: Consider extracting a helper function.

pyk/src/pyk/kast/outer_lexer.py
tothtamas28: Please indicate this in the docstring.

tothtamas28 (approved):

Issue #4188, opened 2024-03-07
Model commands and options with composition


nwatson22:
For a follow-up refactor to runtimeverification/pyk#916 

Split the class `Options` into `Options` which holds the options as fields, and `OptionsHandler` which takes a parameterized `Options` type and can export argparse information for that option, handle default values, and instantiate its `Options` class.
Commands would then be instantiated by passing in `OptionsHandler`s for the appropriate sets of options 

See comment [here](https://github.com/runtimeverification/pyk/pull/916/#discussion_r1515873058)

ehildenb:
Maybe a nice design would be something like:

```
class Option:
    default: Any | None = None
    name: str
    cli: str
    help: str

    @abstractmethod
    def parser(parser: ArgumentParser) -> None:
        ...

class BoolOption(Option):
    ...

    def parser(_parser) -> None:
        _parser.add_argument(self.cli, self.name, action='store_true', default=self.default, help=self.help)
        _parser.add_argument(f`{--no-}{self.cli[3:]}', self.name, action='store_false', default=self.default, help=self.help)
```


Pretty rough sketch, but basically the idea is that we are only declaratively making options, and the parsers are generated for us, and then more functionality can be derived from the declarative arguments instead of speciffying thinsg like the parser/toml reader directly.

PR #4254, opened 2024-04-16
add toml support for pyk options

Baltoli:
Transfer of: https://github.com/runtimeverification/pyk/pull/1008


Baltoli:
@ovatman The K repository requires that changes are based on the `develop` branch rather than `master`; if you merge `master` into your PR (https://github.com/runtimeverification/k/pull/4254/commits/5fbfaf22d8efde1694fb28565c25d86ab197b9ab) then you'll end up with this versioning noise as well as your actual changes.

ehildenb:
@ovatman you'll need to rebase on `develop` branch directly and remove all the introduce intermediate commits then force push.

ovatman:
> @ovatman you'll need to rebase on `develop` branch directly and remove all the introduce intermediate commits then force push.

@ehildenb 
I did the rebasing yesterday but I didn't force push, I think that's why it didn't seem to work.

nwatson22 (approved):
This looks good to me, once it has been rebased properly and the KEVM and kontrol branches are working. I like the solution of having the mapping from command-line string format to python variable format.

There may be a way to avoid having to mention all the superclasses in `from_option_string`, by adding a new function in `Options` [here](https://github.com/runtimeverification/k/blob/28bcfb01f5143fcf037db4c64a2868a934b2fefe/pyk/src/pyk/cli/cli.py) similar to the constructor that collects and merges the `from_option_string`s for any type of `Options`.

There is a lot of boilerplate that has to be duplicated across all 3 repos, but this was also the case with version of `Options` that was merged. Hopefully eventually we will be wrapping the argparse and toml parsing logic and consolidating most of the boilerplate into pyk only (Similar to what is mentioned in #4188)

PR #4256, opened 2024-04-16
Parallel proof exploration

nwatson22:
### Additions:

- Adds `ParallelProver`, which uses `ProverPool` to handle a `ThreadPoolExecutor` that creates a new prover for each new thread used (based on `KoreServerPool`)
- `ParallelProver.parallel_advance_proof` executes a single proof in parallel.
- Requires parameter `create_prover: Callable[[], Prover[P, PS, SR]]` to know how to create a fresh `Prover`.
- User is responsible for making sure this function produces `Provers` which are independent, i.e. don't refer to the same `KoreClient`, etc. (Although they can/should send requests to the same server).
- Reworks `APRProofStep` to no longer contain a proof reference, but instead only immutable data.
- Moves the prior loops counting and zero depth bound checking to the `get_steps` stage, because they require whole-KCFG knowledge
- Adds an example of how to use parallel proving in `test_all_path_reachability_prove_parallel`.
- Adds a test `long-branches` in `imp-simple-spec` intended for profiling parallelization.

### Remaining issues:

- ~~Adds an `id()` abstract method to `ProofStep`, to be able to differentiate steps (TODO this might be able to be removed now if we can enforce `ProofStep` being a hashable dataclass)~~
- Saves the `port` as a field in `KoreClient`. This can probably be removed or improved somehow. I'm just using it downstream to get the port being used by a `Prover` in order to generate fresh `Prover`s with the same port, when we don't have prior access to that port, i.e. `KoreServer` was allowed to choose its own port. The port is already accessible through `KoreClient`, but you have to access private vars several layers deep.
- ~~Adds `kcfg_semantics` as parameter to `get_steps`, because I haven't been able to find a way to separate the prior loops counting algorithm into two parts, one with `kcfg` and one with `same_loop`, without breaking the prior loop caching. This is not ideal, because `kcfg_explore` probably shouldn't even be in `Prover`, just subclasses of `Prover` that deal with `KCFG`s.~~
- ~~To actually get the benefit of parallelization, you need to set `GHC='-N[threads]'` so the single shared `KoreServer` can start its own threads to deal with requests in parallel. This should probably be automated, but I'm not sure how best to do that.~~
- Perhaps a better design than passing around a `create_prover` function would be to figure out all relevant parameters for doing this and passing them to `ParallelProver()`.
- ~~Maybe move parameters from `parallel_advance_proof` to `ParallelProver.__init__()`~~
- The function passing design currently also makes it so that we aren't using the contextmanager features of `KoreServer`, so ~~the KoreServers probably aren't getting closed properly~~.
- I would like others to check my work re. thread safety.

### Tested on kevm and kontrol:

- https://github.com/runtimeverification/evm-semantics/pull/2400
- https://github.com/runtimeverification/kontrol/pull/520


pyk/src/pyk/proof/proof.py
```diff
@@ -273,7 +277,7 @@ def lines(self) -> list[str]:
         return CompositeSummary([BaseSummary(self.id, self.status), *subproofs_summaries])
 
     @abstractmethod
-    def get_steps(self) -> Iterable[PS]: ...
+    def get_steps(self, kcfg_semantics: KCFGSemantics) -> Iterable[PS]: ...
```
tothtamas28: This introduces a dependency from the abstract `Proof` to something `KCFG`-related. Can this be an attribute on the corresponding subclass instead?

pyk/src/pyk/proof/proof.py
```diff
@@ -303,13 +307,105 @@ def lines(self) -> list[str]:
...
+        *,
+        max_workers: int | None = None,
+    ) -> None:
+        self._create_prover = create_prover
+        self._provers = {}
+        self._executor = ThreadPoolExecutor(max_workers)
+        self._closed = False
 
+    def __enter__(self) -> ProverPool[P, PS, SR]:
+        return self
```
tothtamas28:
```suggestion
    def __enter__(self) -> ProverPool[P, PS, SR]:
        self.executor.__enter__()
        return self
```

pyk/src/pyk/proof/proof.py
```diff
@@ -303,13 +307,105 @@ def lines(self) -> list[str]:
...
 
-class StepResult: ...
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        self.close()
 
+    def close(self) -> None:
+        self._executor.shutdown()
+        self._closed = True
```
tothtamas28:
```suggestion
    def close(self) -> None:
        self._executor.__exit__()
        self._closed = True
```

pyk/src/pyk/proof/proof.py
```diff
@@ -303,13 +307,105 @@ def lines(self) -> list[str]:
...
+                prover = self._create_prover()
+                self._provers[thread_name] = prover
+            return prover.step_proof(proof_step)
+
+        return step
+
+
+class ParallelProver(Generic[P, PS, SR]):
+
+    def parallel_advance_proof(
```
tothtamas28: As the class has no state, consider pulling this out as a top-level function.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: I'm hoping this is fixed now without creating other problems. I'm still waiting on these changes to finish testing on kontrol. To fix it I needed to include in the `APRProofStep` the prior loops cache and a tuple of `Nodes` which represent the path leading up to the current node. I am a bit worried this may unintentionally revert the performance fix that introduced prior loop caching: https://github.com/runtimeverification/pyk/pull/922.

@ovatman, would you be able to help me figure out what the requirements for the caching were and if they are still satisfied with this change?

Baltoli (dismissed):
You'll need to rebase out the noise from merging master into this branch @nwatson22

Baltoli:
> User is responsible for making sure this function produces Provers which are independent, i.e. don't refer to the same KoreClient, etc. (Although they can/should send requests to the same server).

Can we add some documentation to this effect (and in general for the rest of the code affected by this change) that will end up in the generated Sphinx documentation? @tothtamas28 can help you out with the best way to do this if you need.

Baltoli:
> To actually get the benefit of parallelization, you need to set `GHC='-N[threads]'` so the single shared KoreServer can start its own threads to deal with requests in parallel. This should probably be automated, but I'm not sure how best to do that.

AFAICT the underlying command invocation is here: https://github.com/runtimeverification/k/blob/f20eb7f8d609084efab6bc3b9b69b5a7cc0ab506/pyk/src/pyk/kore/rpc.py#L1189

You'll need to pass the number of threads through to the KoreServer, then when the `kore-server` binary is started, give it a [modified environment](https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment) that includes the correct GHC environment variable (which I think is [`GHCRTS`](https://downloads.haskell.org/~ghc/7.4.1/docs/html/users_guide/runtime-control.html#rts-options-environment)?)

pyk/src/pyk/proof/proof.py
```diff
@@ -303,20 +304,110 @@ def lines(self) -> list[str]:
         return [line for lines in (summary.lines for summary in self.summaries) for line in lines]
 
 
-class ProofStep: ...
+class FailureInfo: ...
 
 
-class StepResult: ...
+class ProverPool(ContextManager['ProverPool'], Generic[P, PS, SR]):
```
Baltoli: What are the generic parameters here? I'm not super familiar with this code so it may be more obvious with context, but it would be nice to document things like this in Sphinx as well!

pyk/src/pyk/proof/proof.py
```diff
@@ -303,20 +304,110 @@ def lines(self) -> list[str]:
...
+        submit_steps(proof.get_steps())
+
+        while True:
+            if len(pending) == 0:
+                break
+            done, _ = wait(pending, return_when='FIRST_COMPLETED')
+            future = done.pop()
+            proof_results = future.result()
+            for result in proof_results:
+                proof.commit(result)
```
Baltoli, line 370: Just to check I understand the threading model here correctly, the parallel work is all being done by the pool executors on data that is not actually part of the root proof. Then, this loop only ever runs on the main thread and there's no contention possible on `.commit()` for the root proof?

Baltoli:
A few comments; I'll defer to @tothtamas28 and @ehildenb for a final review

pyk/src/pyk/proof/proof.py
nwatson22: That's correct. Commit should only be run on the main thread.

pyk/src/pyk/proof/proof.py
nwatson22: P is `Proof`, PS is proof step, and `SR` is step result. So each `Prover` is operating on a specific type of `Proof`, e.g. `APRProof` with corresponding types that represent the data being passed to and from `step_proof`.

nwatson22:
> AFAICT the underlying command invocation is here:
> 
> https://github.com/runtimeverification/k/blob/f20eb7f8d609084efab6bc3b9b69b5a7cc0ab506/pyk/src/pyk/kore/rpc.py#L1189
> 
> You'll need to pass the number of threads through to the KoreServer, then when the `kore-server` binary is started, give it a [modified environment](https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment) that includes the correct GHC environment variable (which I think is [`GHCRTS`](https://downloads.haskell.org/~ghc/7.4.1/docs/html/users_guide/runtime-control.html#rts-options-environment)?)

Thank you @Baltoli, this is now implemented.

pyk/src/pyk/proof/proof.py
```diff
@@ -76,7 +76,9 @@ def __init__(
             ensure_dir_path(self.proof_dir)
 
     @abstractmethod
-    def commit(self, result: SR) -> None: ...
+    def commit(self, result: SR) -> None:
+        """Applies the step result of type `SR` to `self`, modifying `self`."""
```
tothtamas28:
```suggestion
        """Apply the step result of type `SR` to `self`, modifying `self`."""
```

(Example: https://peps.python.org/pep-0257/#one-line-docstrings)

pyk/src/pyk/proof/proof.py
```diff
@@ -452,7 +456,12 @@ def shutdown(self) -> None: ...
     def failure_info(self, proof: P) -> FailureInfo: ...
 
     @abstractmethod
-    def step_proof(self, step: PS) -> Iterable[SR]: ...
+    def step_proof(self, step: PS) -> Iterable[SR]:
+        """Does work associated with a `PS`, a proof step. Should not modify a `Proof` or `self`, but may read
```
tothtamas28:
```suggestion
        """Do the work associated with a `PS`, a proof step. Should not modify a `Proof` or `self`, but may read
```

pyk/src/pyk/proof/implies.py
```diff
@@ -26,13 +26,13 @@
 _LOGGER: Final = logging.getLogger(__name__)
 
 
-@dataclass
-class ImpliesProofStep(ProofStep):
+@dataclass(frozen=True, eq=True)
```
tothtamas28:
```suggestion
@dataclass(frozen=True)
```

(`eq=True` is the default.)

pyk/src/pyk/proof/implies.py
```diff
@@ -409,9 +409,13 @@ def lines(self) -> list[str]:
 
 class ImpliesProver(Prover[ImpliesProof, ImpliesProofStep, ImpliesProofResult]):
     proof: ImpliesProof
+    kcfg_explore: KCFGExplore
+
+    def shutdown(self) -> None:
```
tothtamas28, line 414: Is this an abstract method on `Prover`? If so, maybe `Prover` should be a `ContextManager`, and use `__exit__` instead.

pyk/src/pyk/proof/proof.py
```diff
@@ -303,31 +314,169 @@ def lines(self) -> list[str]:
         return [line for lines in (summary.lines for summary in self.summaries) for line in lines]
 
 
-class ProofStep: ...
+class FailureInfo: ...
 
 
-class StepResult: ...
+class ProverPool(ContextManager['ProverPool'], Generic[P, PS, SR]):
```
tothtamas28: This class seems to belong to the implementation of `parallel_advance_proof`, consider sliding it below that function. Consider making it private even.

pyk/src/pyk/proof/proof.py
```diff
@@ -303,31 +314,169 @@ def lines(self) -> list[str]:
...
+                if step in explored:
+                    continue
+                explored.add(step)
+                future: Future[Any] = pool.submit(step)  # <-- schedule steps for execution
+                pending.add(future)
+
+        submit_steps(proof.get_steps())
+
+        while True:
+            if len(pending) == 0:
```
tothtamas28:
```suggestion
            if not pending:
```

pyk/src/pyk/proof/proof.py
```diff
@@ -303,31 +314,169 @@ def lines(self) -> list[str]:
...
+    """Abstract class which advances `Proof`s with `init_proof()` and `step_proof()`
+    :param P: Type of proof this `Prover` operates on
+    :param PS: Proof step: data required to perform a step of the proof.
+    :param SR: Step result: data produced by executing a PS with `Prover.step_proof` used to update the `Proof`
+    """
 
-    def __init__(self, kcfg_explore: KCFGExplore):
-        self.kcfg_explore = kcfg_explore
+    @abstractmethod
+    def shutdown(self) -> None: ...
```
tothtamas28: Consider making this class a `ContextManager[Prover]` instead.

pyk/src/tests/integration/proof/test_imp.py
```diff
@@ -926,6 +944,66 @@ def test_terminal_node_subsumption(
...
+                for subproof in proof.subproofs:
+                    subproof.admit()
+                    subproof.write_proof_data()
+
+            def create_prover() -> APRProver:
+                kore_client = KoreClient(
+                    'localhost', _kore_server.port, timeout=self.CLIENT_TIMEOUT, bug_report=bug_report
+                )
+                ct_symb = CTermSymbolic(kore_client, definition, kompiled_kore)
+                _kcfg_explore = KCFGExplore(ct_symb, kcfg_semantics=self.semantics(definition))
```
tothtamas28: Consider defining a test class for parallel proving (similar to `KoreClientTest` et al.) that provides these building blocks as fixtures.

pyk/src/pyk/proof/proof.py
nwatson22: I've made it a `ContextManager` as well, but I'm still having to use this `shutdown` method (I renamed it to `close`), because `ProverPool` keeps `Prover`s in a list and we do not want to close those `Prover`s until the whole `ProverPool` is closed, similar to `KoreServerPool`. Is there a way to better use the `ContextManager` feature under these circumstances?

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/implies.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/tests/integration/proof/test_imp.py
nwatson22: Added a test class for the parallel proving, although I wasn't sure what the best way to pass in the `Prover` parameters that are test-case-specific, such as max_depth and cut_point_rules.

pyk/src/pyk/proof/proof.py
```diff
@@ -303,31 +318,175 @@ def lines(self) -> list[str]:
...
+
+    :param create_prover: Function which creates a new `Prover`. These provers must not reference any shared
+        data to be written during `parallel_advance_proof`, to avoid race conditions.
+    :param max_iterations: Maximum number of steps to take
+    :param fail_fast: If the proof is failing after finishing a step, halt execution even if there are
+        still available steps
+    :param max_workers: Maximum number of worker threads the pool can spawn
+    """
+
+    class _ProverPool(ContextManager['_ProverPool'], Generic[_P, _PS, _SR]):
```
tothtamas28: Sorry for the confusion. I meant just moving the class definition after the function definition.

pyk/src/pyk/testing/_kompiler.py
```diff
@@ -300,6 +302,24 @@ def kcfg_explore(
         yield KCFGExplore(cterm_symbolic, kcfg_semantics=semantics)
 
 
+class ParallelTest(KCFGExploreTest):
```
tothtamas28: Inheriting from `KCFGExploreTest` seems unnecessary here, as the fixtures defined there are not used.

pyk/src/tests/integration/proof/test_imp.py
```diff
@@ -1353,3 +1368,56 @@ def test_implication_failure_reason(
 
         assert not passed
         assert actual == expected
+
+
+class TestImpParallelProof(ParallelTest, KCFGExploreTest):
```
tothtamas28: Does this work?

```suggestion
class TestImpParallelProof(ParallelTest):
```

pyk/src/pyk/proof/proof.py
nwatson22: I moved it out of function scope now.

pyk/src/tests/integration/proof/test_imp.py
nwatson22: It's inheriting from `ParallelTest` and `KProve` test now.

pyk/src/pyk/testing/_kompiler.py
nwatson22: I pulled out the server related fixtures from `KoreClientTest` into `KoreServerTest` and am using that with `ParallelTest` through `KoreClientFactoryTest`.

pyk/src/pyk/testing/_kompiler.py
```diff
@@ -265,6 +264,11 @@ def _kore_server(
             case _:
                 raise AssertionError
 
+
+class KoreClientTest(KoreServerTest):
+    KOMPILE_BACKEND: ClassVar = 'haskell'
```
tothtamas28: This belongs to `KoreServerTest`.

pyk/src/pyk/testing/_kompiler.py
```diff
@@ -300,6 +304,45 @@ def kcfg_explore(
...
 
+class KoreClientFactoryTest(KoreServerTest):
+    KOMPILE_BACKEND: ClassVar = 'haskell'
+    CLIENT_TIMEOUT: ClassVar = 1000
+
+    # definition_dir should ideally depend on server_type to avoid triggering kompilation,
+    # but it can't due to method signature in superclass.
+    # In the parameter list, always put `definition_dir` after `kore_client`.
+    @pytest.fixture
+    def create_kore_client(self, _kore_server: KoreServer, bug_report: BugReport) -> Callable[[], KoreClient]:
```
tothtamas28: Since tests do not seem to consume this fixture directly, I'd just move this into `ParallelTest` as a private helper method.

tothtamas28 (approved):

pyk/src/pyk/testing/_kompiler.py
nwatson22: Fixed.

pyk/src/pyk/testing/_kompiler.py
nwatson22: Fixed.

PR #4269, opened 2024-04-22
Parallel maven

Baltoli:


Baltoli:
Closing this as it wasn't worth pursuing; the idea was to try and locally optimise the individual steps in the Debian package build so as to minimise the end-to-end time to begin running the Pyk tests.

PR #4270, opened 2024-04-22
Add `parse-outer` command to pyk

gtrepta:
This starts implementing a command with pyk that handles the outer parsing of a K definition into a KAST JSON that can be passed to inner parsing in the java frontend. So far it can handle requires clauses of other files, do markdown selection, and handle extra include paths passed in from the command line.


pyk/src/pyk/__main__.py
```diff
@@ -389,6 +393,60 @@ def exec_json_to_kore(options: JsonToKoreOptions) -> None:
...
+    md_selector = options.md_selector
+    required_files = []
+    search_paths = [main_file_dir]
+    for include in options.includes:
+        include_path = Path(include)
+        try:
+            check_dir_path(include_path)
+        except ValueError:
+            _LOGGER.critical(f"Could not find directory '{include}' passed to -I")
+            exit(1)
```
Baltoli: Worth noting this isn't what `-I` does in the existing compiler frontend, nor in GCC / clang. Could maybe be a lower level warning rather than an error eventually?

gtrepta: Yep, I can just make this a warning.

pyk/src/tests/integration/test_pyk.py
tothtamas28, line 1: This should be a unit test.

pyk/src/pyk/__main__.py
```diff
@@ -389,6 +395,65 @@ def exec_json_to_kore(options: JsonToKoreOptions) -> None:
     sys.stdout.write('\n')
 
 
+def exec_parse_outer(options: ParseOuterOptions) -> None:
```
tothtamas28, line 388: Please extract a function into one of the `pyk.kast` modules that performs the transformation logic. Unit tests can call that function directly.

pyk/src/pyk/konvert/_ast_to_kast.py
```diff
@@ -0,0 +1,55 @@
...
+if TYPE_CHECKING:
+    from typing import Any, Final
+
+    from ..kast.outer import KAst
+    from ..kast.outer_syntax import AST
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+@singledispatch
```
tothtamas28: Nice.

pyk/src/pyk/__main__.py
```diff
@@ -394,6 +396,12 @@ def exec_json_to_kore(options: JsonToKoreOptions) -> None:
 
 
 def exec_parse_outer(options: ParseOuterOptions) -> None:
+    # TODO Desugar K language syntax to KAST compatible items
```
Baltoli: Perhaps turn this into a tracking issue and assign yourself to it

pyk/src/pyk/konvert/_ast_to_kast.py
Baltoli: This is a trick I didn't know about; very cool

pyk/src/pyk/__main__.py
gtrepta: Sorry, what transformation logic are you referring to?

tothtamas28: This function performs side effects. You can extract a function that returns `result_text` (or something more structured), that way, the command line handler becomes very simple.

pyk/src/pyk/konvert/_ast_to_kast.py
gtrepta: I had to massage the type checker a bit to accept this in a few places, but it does make for some convenient calls like the one here:

https://github.com/runtimeverification/k/blob/7cbf5551b5aca2e7a19a231611c8ceeb08bc5f0a/pyk/src/pyk/konvert/_ast_to_kast.py#L35

pyk/src/pyk/__main__.py
gtrepta: If you mean `_ast_to_kast`, I agree I should write some unit tests for that.

tothtamas28: Sounds good.

You've already set up a test (`TestParseOuter`) though, what I'm saying, if you extract a function (that takes `main_file`, etc. and returns the text you'd otherwise print), then you can call that directly without stubbing `sys.argv`.

Baltoli: Yes, I agree that the substantive work here should be extracted into a function that _returns_ something - the implementation will look pretty similar to what you have here, just with a different signature

pyk/src/pyk/__main__.py
```diff
@@ -389,6 +395,65 @@ def exec_json_to_kore(options: JsonToKoreOptions) -> None:
...
 
 
+def exec_parse_outer(options: ParseOuterOptions) -> None:
+    # TODO Desugar K language syntax to KAST compatible items
+    #      Import the prelude and add --no-prelude option
+    #      Trim away modules unreachable from the main/syntax modules (modulo some builtins)
+    #      Checks:
+    #        - Circular imports
+    #        - Modules with duplicate names
+    main_file_dir = Path(options.main_file.name).resolve().parent
```
Baltoli: Do you want to `check_file_path` this, or is the error from `.resolve()` good enough?

pyk/src/pyk/konvert/_ast_to_kast.py
Baltoli: I guess it saves you a big pile of `isinstance` checks which is nice

pyk/src/pyk/__main__.py
gtrepta: `options.main_file` is checked by the argument parser, so before we get to this `.resolve()` it should error there:

https://github.com/runtimeverification/k/blob/7cbf5551b5aca2e7a19a231611c8ceeb08bc5f0a/pyk/src/pyk/__main__.py#L602

Baltoli: Nice!

pyk/src/pyk/kast/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    from pathlib import Path
+    from typing import Final
+
+    from .outer import KDefinition
+    from .outer_syntax import Module
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+def _slurp(
```
tothtamas28, line 36: Consider sliding this below `parse_outer`.

pyk/src/pyk/konvert/_ast_to_kast.py
tothtamas28, line 1: This module belongs to `pyk.kast` I think.

pyk/src/pyk/kast/utils.py
```diff
@@ -0,0 +1,61 @@
...
+def parse_outer(
+    definition_file: Path, main_module: str, search_paths: Iterable[Path] = (), md_selector: str = 'k'
+) -> KDefinition:
+    text = definition_file.read_text()
+    if definition_file.suffix == '.md':
+        text = select_code_blocks(text, md_selector)
+
+    modules = _slurp(text, search_paths, (), md_selector)
+    final_definition = _ast_to_kast(Definition(modules), main_module=main_module)
+    return cast('KDefinition', final_definition)
```
tothtamas28: `cast` does not do any runtime checking, consider asserting the type instead.

pyk/src/pyk/kast/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    result = definition.modules
+    for require in definition.requires:
+        try_files = [include_dir / require.path for include_dir in search_paths]
+        try:
+            # Get the first source file we can find by iterating through search_paths
+            index = [file.exists() for file in try_files].index(True)
+        except ValueError:
+            _LOGGER.critical(f'{require.path} not found')  # TODO Print the source location of the requires clause
+            _LOGGER.critical(f'Lookup directories: {[str(path) for path in search_paths]}')
+            exit(1)
```
tothtamas28: `exit` should not be called here since we might want to call `parse_outer` from contexts other than the CLI. Instead, consider re-raising the exception and catching it in `__main__` to do the `exit`.

pyk/src/tests/unit/test_konvert.py
```diff
@@ -49,3 +55,40 @@ def test_unmunge(symbol: str, expected: str) -> None:
 
     # Then
     assert actual == expected
+
+
+AST_TO_KAST_TEST_DATA: Final = [
```
tothtamas28: These tests can go to their own test module, under `tests/unit/kast`.

pyk/src/pyk/konvert/_ast_to_kast.py
gtrepta: Came up with a better solution than the KSentence cast above: a10e68aeed9ce8fd4d3cfbb7cfa59bb1089980e8

pyk/src/pyk/kast/_ast_to_kast.py
```diff
@@ -0,0 +1,62 @@
...
+@_ast_to_kast.register
+def _att_to_katt(att: Att) -> KAtt:
+    if not att.items:
+        return EMPTY_ATT
+    return KAtt.from_dict({'att': dict(att.items)})
+
+
+@singledispatch
+@_ast_to_kast.register
+def _sentence_to_ksentence(s: Sentence) -> KSentence:
```
tothtamas28, line 59: This is a placeholder for later implementation, right?

Regardless, it is not safe to trick the type checker this way. Instead consider a helper function

```python3
def _transform_sentences(sentences: Iterable[Sentence]) -> list[KSentence]:
    return []
```

which you can call in `_module_to_kflatmodule`.

pyk/src/tests/unit/kast/test_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    # Given
+    expected_file = test_file.with_suffix(test_file.suffix + '.expected.json')
+    expected = KDefinition.from_dict(json.loads(expected_file.read_text()))
+    main_module = test_file.stem.upper()
+
+    # When
+    actual = parse_outer(test_file, main_module, (test_file.parent, test_file.parent / 'include'))
+
+    # Then
+    assert actual.to_json() == expected.to_json()
```
tothtamas28: Does this work?

```suggestion
    assert actual == expected
```

tothtamas28 (approved):

pyk/src/pyk/kast/_ast_to_kast.py
gtrepta: It's a placeholder in the same sense that `_ast_to_kast` is, which is that it isn't fully implemented. I can implement sentence transformations with `_sentence_to_ksentence.register`, and I think the dispatcher in `_ast_to_kast` should pass any `Sentence`s along to here.

I'm not sure what you mean by tricking the type checker. This is the same pattern as declaring `_import_to_kimport` and using it in `_module_to_kflatmodule`.

tothtamas28: > I'm not sure what you mean by tricking the type checker.

This:

```python3
def _sentence_to_ksentence(s: Sentence) -> KSentence:
    return None # type: ignore
```

Because here:

```python3
sentences = (_sentence_to_ksentence(s) for s in m.sentences)
```

you'll end up with a bunch of `None`s, whereas you'd expect `KSentence` instances.

So consider either weakening the signature to:

```python3
def _sentence_to_ksentence(s: Sentence) -> KSentence | None:
    return None
```

and filtering out `None`-s on the call site, or postponing the `singledispatch`-based implementation until all the cases are covered. Then, you can implement this signature as

```python3
def _sentence_to_ksentence(s: Sentence) -> KSentence:
    raise AssertionError(f'Unimplemented conversion for: {type(s)}')
```

gtrepta: Ok, I see what you are saying. Well, there really is no use case for returning `None` other than as a placeholder, so I can change it to raise an `AssertionError` which wont affect anything as there aren't any tests that hit it and it isn't being used in production. It should also make it helpful as I implement the conversions and add tests for them, since the raised exception will be easier to spot than an error message from the logger.

PR #4275, opened 2024-04-23
Remove precede/follow regex from `RegexTerminal`

### LINKED ISSUE #4266: Negative lookbehind/lookahead patterns in regex terminals are unsupported in practice
Currently, the syntax of regex terminals allows specifying 
- negative lookbehind with `(?<! ...)` 
- negative lookahead with `(?! ...)` 

For example, https://github.com/runtimeverification/k/blob/b2c217076fbaac0460a1980c47cbe5bf6b980516/pyk/src/tests/integration/test_parse_kast.py#L69-L74

Negative lookbehind occurs in some of the `kast.md` and `domains.md` definitions.

However, searching the frontend, these are never actually considered during lexing. That is, the `precedeRegex` and `followRegex` of `RegexTerminal`

https://github.com/runtimeverification/k/blob/6c9ec0b8ae8e7ab5c43c177c2414ff4a5490c896/kore/src/main/scala/org/kframework/definition/outer.scala#L1025

are never accessed in [`Scanner.java`](https://github.com/runtimeverification/k/blob/afa82dcd698ed78a985a41c3379e32a4f9fc69b8/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java`) nor anywhere else relevant.

Given that Flex does not support negative lookbehind/lookahead, IMO we should probably just drop the feature entirely.

Scott-Guest:
Closes #4266

Removes lingering references to the now unsupported negative lookbehinds/lookaheads in regex terminals.


k-distribution/tests/regression-new/issue-3647-debugTokens/a.test.kast.out
```diff
@@ -1,12 +1,12 @@
 |"Match"                                                                          | (location)    | Terminal            |
 |---------------------------------------------------------------------------------|---------------|---------------------|
-|"1"                                                                              | (1,1,1,2)     | r"[0-9]+"           |
+|"1"                                                                              | (1,1,1,2)     | r"[\\+-]?[0-9]+"    |
```
Baltoli, line 3: The issue here is that signed and unsigned integers now have overlapping lexical syntax at the same priority, right? (that is, this particular instance of instability will get resolved for free when you come round to implementing the overlap checker?)

Baltoli (approved):
Nice one, thanks Scott! I also think it's worth calling out that (as far as I can see) this PR is the first one to benefit from the Pyk / K merge.

k-distribution/tests/regression-new/issue-3647-debugTokens/a.test.kast.out
Scott-Guest: Yup! This will be stabilized with the overlap changes

PR #4280, opened 2024-04-25
Terminal node subsumption check correction

PetarMax:
This PR corrects a bug in which if the execution stopped precisely at a terminal node, that node would never be checked for subsumption. It also adds a test that fails on `master`/`develop` but succeeds in this PR.

### DIFF
--- a/pyk/src/pyk/proof/reachability.py
+++ b/pyk/src/pyk/proof/reachability.py
@@ -679,8 +679,9 @@ class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
         _inject_module(proof.dependencies_module_name, self.main_module_name, dependencies_as_rules)
         _inject_module(proof.circularities_module_name, proof.dependencies_module_name, [circularity_rule])
 
-        if self.kcfg_explore.kcfg_semantics.is_terminal(proof.kcfg.node(proof.target).cterm):
-            proof.add_terminal(proof.target)
+        for node_id in [proof.init, proof.target]:
+            if self.kcfg_explore.kcfg_semantics.is_terminal(proof.kcfg.node(node_id).cterm):
+                proof.add_terminal(node_id)
 
     def nonzero_depth(self, proof: APRProof, node: KCFG.Node) -> bool:
         return not proof.kcfg.zero_depth_between(proof.init, node.id)
--- a/pyk/src/pyk/proof/reachability.py
+++ b/pyk/src/pyk/proof/reachability.py
@@ -86,7 +86,6 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
     error_info: Exception | None
     prior_loops_cache: dict[int, list[int]]
 
-    _checked_for_terminal: set[int]
     _checked_for_bounded: set[int]
 
     def __init__(
@@ -124,7 +123,6 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
         self._exec_time = _exec_time
         self.error_info = error_info
 
-        self._checked_for_terminal = set()
         self._checked_for_bounded = set()
 
         if self.proof_dir is not None and self.proof_subdir is not None:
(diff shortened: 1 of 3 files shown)


pyk/src/pyk/proof/reachability.py
```diff
@@ -681,26 +679,12 @@ def _inject_module(module_name: str, import_name: str, sentences: list[KRuleLike
...
-    def _check_terminal(self, proof: APRProof, node: KCFG.Node) -> None:
-        if node.id not in proof._checked_for_terminal:
-            _LOGGER.info(f'Checking terminal: {node.id}')
-            proof._checked_for_terminal.add(node.id)
-            if self.kcfg_explore.kcfg_semantics.is_terminal(node.cterm):
-                _LOGGER.info(f'Terminal node: {node.id}.')
-                proof.add_terminal(node.id)
-            elif self.fast_check_subsumption and self._may_subsume(proof, node):
-                _LOGGER.info(f'Marking node as terminal because of fast may subsume check {proof.id}: {node.id}')
-                proof.add_terminal(node.id)
```
PetarMax, line 698: This was only used in `check_all_terminals`, and in other places we just use `is_terminal` of the underlying semantics, so I don't think we need it anymore.

pyk/src/pyk/proof/reachability.py
```diff
@@ -86,7 +86,6 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
     error_info: Exception | None
     prior_loops_cache: dict[int, list[int]]
 
-    _checked_for_terminal: set[int]
```
PetarMax, line 89: Given how proof execution is designed, I don't think there's a possibility of repeated terminalness checks for the same node, so I think we could lose this piece of state. Even if repeated checks do happen, they are very fast.

pyk/src/pyk/proof/reachability.py
```diff
@@ -681,26 +679,12 @@ def _inject_module(module_name: str, import_name: str, sentences: list[KRuleLike
         _inject_module(proof.dependencies_module_name, self.main_module_name, dependencies_as_rules)
         _inject_module(proof.circularities_module_name, proof.dependencies_module_name, [circularity_rule])
 
-        self._check_all_terminals(proof)
+        if self.kcfg_explore.kcfg_semantics.is_terminal(proof.kcfg.node(proof.target).cterm):
+            proof.add_terminal(proof.target)
```
PetarMax: We still have to check if the `target` node of the KCFG is terminal, because subsumption logic depends on it.

pyk/src/tests/integration/proof/test_imp.py
```diff
@@ -896,6 +896,36 @@ def test_all_path_reachability_prove(
             assert proof.status == proof_status
             assert leaf_number(proof) == expected_leaf_number
 
+    def test_terminal_node_subsumption(
```
PetarMax, line 899: The test, in essence, runs a claim up to its terminal node, stops, then makes one more step, which only checks subsumption.

tothtamas28 (approved):
LGTM. Please tests with `evm-semantics` and `kontrol` before merging.

pyk/src/pyk/proof/reachability.py
ehildenb: We should also check the initial state, since we should not have nodes in the graph which _are_ terminal but are not marked as such.

PetarMax: Yes, done now.

PetarMax:
> LGTM. Please tests with `evm-semantics` and `kontrol` before merging.

Yes, the draft PRs are [here](https://github.com/runtimeverification/evm-semantics/pull/2405) and [here](https://github.com/runtimeverification/kontrol/pull/536).

PetarMax:
The draft PRs are passing in principle, there are a few tests in Kontrol that I'm having trouble with in terms of expected output and different machines, but there are no unexpected changes.

PetarMax:
The draft PRs are passing now fully.

PR #4281, opened 2024-04-25
Revert constraint unsortedness

PetarMax:
Partial revert of #4249.

Due to problems in cross-machine orderings in the constraints of `CTerm`s observed by @goodlyrottenapple, this PR reverts the non-sortedness of `CTerm` constraints. The chronological order of `Split` constraints is still maintained.

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -14,7 +14,6 @@ from ..kast.manip import (
     flatten_label,
     free_vars,
     ml_pred_to_bool,
-    normalize_constraints,
     push_down_rewrites,
     remove_useless_constraints,
     split_config_and_constraints,
@@ -23,7 +22,7 @@ from ..kast.manip import (
 from ..prelude.k import GENERATED_TOP_CELL
 from ..prelude.kbool import andBool, orBool
 from ..prelude.ml import is_bottom, is_top, mlAnd, mlBottom, mlEqualsTrue, mlImplies, mlTop
-from ..utils import unique
+from ..utils import single, unique
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
@@ -47,10 +46,10 @@ class CTerm:
 
     def __init__(self, config: KInner, constraints: Iterable[KInner] = ()) -> None:
         """Instantiate a given `CTerm`, performing basic sanity checks on the `config` and `constraints`."""
-        if is_top(config):
+        if CTerm._is_top(config):
             config = mlTop()
             constraints = ()
-        elif is_bottom(config):
+        elif CTerm._is_bottom(config):
             config = mlBottom()
             constraints = ()
         else:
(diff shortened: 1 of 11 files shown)


PetarMax:
Relevant PRs for testing: [KEVM](https://github.com/runtimeverification/evm-semantics/pull/2404) (passing) and [Kontrol](https://github.com/runtimeverification/kontrol/pull/535) (minor expected output faff, but nothing out of the ordinary).

PetarMax:
Both PRs upstream are passing now.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -93,10 +93,12 @@ def _check_config(config: KInner) -> None:
             raise ValueError(f'Expected cell label, found: {config}')
 
     @staticmethod
-    def _normalize_constraints(constraints: Iterable[KInner]) -> tuple[KInner, ...]:
+    def _normalize_constraints(constraints: Iterable[KInner], maintain_order: bool = False) -> tuple[KInner, ...]:
```
tothtamas28: Consider extracting the flattening logic into a helper function. Then `maintain_order` can be removed, and `CSubst` won't have to call the private method of another class.

tothtamas28 (approved):

pyk/src/pyk/prelude/ml.py
```diff
@@ -20,12 +22,24 @@
 }
 
 
-def is_top(term: KInner) -> bool:
-    return isinstance(term, KApply) and term.label.name == '#Top'
-
-
-def is_bottom(term: KInner) -> bool:
-    return isinstance(term, KApply) and term.label.name == '#Bottom'
+def is_top(term: KInner, weak: bool = False) -> bool:
```
tothtamas28:
```suggestion
def is_top(term: KInner, *, weak: bool = False) -> bool:
```

pyk/src/pyk/prelude/ml.py
```diff
@@ -20,12 +22,24 @@
...
+    if weak:
+        flat = flatten_label('#And', term)
+        if len(flat) == 1:
+            return is_top(single(flat))
+        return all(is_top(term, weak=True) for term in flat)
+    else:
+        return isinstance(term, KApply) and term.label.name == '#Top'
```
tothtamas28: Consider extracting `_is_top` for the simple condition (as in an earlier version).

tothtamas28: (Nitpick) I'd organize this a bit differently to avoid indenting the `weak=True` logic:

```suggestion
    if _is_top(term):
        return True
    if not weak:
        return False
    flat = flatten_label('#And', term)
    if len(flat) == 1:
        return _is_top(single(flat))
    return all(is_top(term, weak=True) for term in flat)
```

pyk/src/pyk/prelude/ml.py
```diff
@@ -20,12 +22,24 @@
...
+
+
+def is_bottom(term: KInner, weak: bool = False) -> bool:
```
tothtamas28: (Symmetrically for `is_bottom`.)

tothtamas28 (approved):

PR #4286, opened 2024-04-26
Move long-running pyk tests into Docker container

### LINKED ISSUE #4277: Move long-running Pyk tests to self-hosted runners

tothtamas28:
Fixes #4277

* Create action `with-k-docker` that starts a Docker container with K installed from a `.deb` package
* Move long-running tests into Docker container and on self-hosted runners

### DIFF
--- a/.github/actions/setup-pyk-env/action.yml
+++ b/.github/actions/setup-pyk-env/action.yml
@@ -6,16 +6,6 @@ inputs:
     required: true
     type: string
     default: '3.10'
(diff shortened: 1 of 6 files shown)


Baltoli (approved):
Nice, this is really great! I think when we have both backends provided as separately packaged releases (https://github.com/runtimeverification/llvm-backend/pull/1036 for the LLVM backend), we should be approaching 20 minutes end-to-end time for K CI!

PR #4288, opened 2024-04-26
PyK wrappers for the proof trace python bindings

theo25:
This PR adds wrappers for the LLVM backend python bindings for poof hints. We add those wrappers to enable type checking and more convenient use of the bindings in general.

The PR also adds some infrastructure code to make it easier to write unit tests for these bindings.


pyk/src/tests/integration/kllvm/test_prooftrace.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # check that there is a initial configuration
+        assert pt.initial_config.is_kore_pattern()
+
+        # check that the trace after the initial configuration is 3 events long
+        assert len(pt.trace) == 3
+
+        # check that the first event is the rewrite a() => b()
+        assert pt.trace[0].is_step_event()
+        rewrite_event = cast('prooftrace.LLVMRewriteEvent', pt.trace[0].step_event)
```
gtrepta: You can do `assert isinstance(pt.trace[0].step_event, prooftrace.LLVMRewriteEvent)` instead. The type checker will accept it and the type will be checked at runtime as well.

pyk/src/tests/integration/kllvm/test_prooftrace.py
```diff
@@ -0,0 +1,56 @@
...
+        assert rewrite_event.rule_ordinal == 96
```
gtrepta: Do you know how stable these rule ordinals are? I'm worried this assertion might break if we make a change to `domains.md`.

gtrepta (approved):
Approving with a couple of minor comments.

pyk/src/pyk/kllvm/compiler.py
Baltoli, line 1: I think the hint generation should maybe live in a separate module; it's not compiling anything, but rather executing the interpreter at runtime to generate a hint.

pyk/src/pyk/kllvm/compiler.py
```diff
@@ -91,3 +91,41 @@ def compile_runtime(
 
     assert module_file.is_file()
     return module_file
+
+
+# -------------------------------
+# utility for generation of hints
+# -------------------------------
+
+HINTS_FILE_NAME: Final = 'hints.bin'
```
Baltoli: This could be a defaulted parameter to `generate_hints`, rather than a global.

pyk/src/tests/integration/kllvm/test_prooftrace.py
```diff
@@ -0,0 +1,56 @@
...
+        LblinitGeneratedTopCell{}(
+          Lbl'Unds'Map'Unds'{}(
+            Lbl'Stop'Map{}(),
+            Lbl'UndsPipe'-'-GT-Unds'{}(
+               inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}("$PGM")),
+               inj{SortFoo{}, SortKItem{}}(Lbla'LParRParUnds'TEST-PROOF-TRACE-SYNTAX'Unds'Foo{}())
+            )
+          )
+        )
+    """
```
Baltoli: You could express this using the KORE DSL Pyk gives you rather than writing a string by hand.

Baltoli (approved):
All looks good to me; a couple of small organisational comments.

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: I'm not sure about changes in `domain.md`, but I would say that they can be a little bit unstable regarding the content of the definition. For instance, I had to modify several assertions like that on Pi2 regression tests when Theo modified the hints not to print intermediate steps! I believe we're using this style here to mimic what we use there. I am unsure if it is ideal from the K/Pyk side, but for us, it could be helpful to identify earlier that these will be affected, so we have to prepare ourselves to fix it when updating the K version.

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: Thanks for the tip! Could you please point me to an example?

pyk/src/pyk/kllvm/compiler.py
Robertorosmaninho: Sure! Which module do you think suits it better?

pyk/src/tests/integration/kllvm/test_prooftrace.py
Baltoli: > I am unsure if it is ideal from the K/Pyk side

I would prefer not to introduce K tests that we know are likely to fail when unrelated code is changed; we've been squashing bugs like this recently.

> it could be helpful to identify earlier that these will be affected

Hard-coding the ordinals like this is convenient for writing test cases and has worked thus far, but perhaps now is the right time to figure out a more durable solution. Perhaps there's a way to use / improve the `compute-ordinal` script from the LLVM backend so that ordinals can be computed as part of the test script? Happy to chat through what this might look like if I'm being unclear.

pyk/src/pyk/kllvm/compiler.py
Baltoli: `kllvm.hints`? It is conceptually a new thing so I think a fresh module suits it best.

pyk/src/tests/integration/kllvm/test_prooftrace.py
Baltoli: Here's an integration test that constructs the expected output of a test using the KORE data structures; your input data here could be similar. Alternatively, you could just check the KORE text in as a test data file. Up to you what you prefer, but I think embedding KORE in a Python string is the least preferable of the three versions.

https://github.com/runtimeverification/k/blob/48b6ffa4110a30d5bd89f1952222c1d17d0d5dbe/pyk/src/tests/integration/kore/test_top_cell_initializer.py#L69-L78

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: The problem here is that the `mypy` type check complains if we don't cast. It can't recognize that LLVMRewriteEvent is a subclass of LLVM stepevent.

gtrepta: `assert isinstance` should work as a cast. `mypy` doesn't complain when I do this:
```diff
         # check that the first event is the rewrite a() => b()
         assert pt.trace[0].is_step_event()
-        rewrite_event = cast('prooftrace.LLVMRewriteEvent', pt.trace[0].step_event)
-        assert rewrite_event.rule_ordinal == 96
+        assert isinstance(pt.trace[0].step_event, prooftrace.LLVMRewriteEvent)
+        assert pt.trace[0].step_event.rule_ordinal == 96
```

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: Thank you, Bruce! I could assemble this as a replacement:
```Python
    program_pattern = KoreParser("Lbla\'LParRParUnds\'TEST-PROOF-TRACE-SYNTAX\'Unds\'Foo{}()").pattern()

    HINTS_INPUT_KORE = init_generated_top_cell(
        map_pattern(
            (
                inj(SORT_K_CONFIG_VAR, SORT_K_ITEM, k_config_var('$PGM')),
                inj(SortApp('SortFoo'), SORT_K_ITEM, program_pattern),
            )
        )
    ).text
```

However, I would like very much to replace the argument of `KoreParser` with this `kore_text`
```Python
kore_text = _kast(definition_dir=x, input='program', output='kore', expression="a()").stdout
```

The issue with this approach is that I couldn't figure out how to get the `x` definition directory :/ 
Could you help with it? Or is it okay to let it as it is?

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: Sorry, I didn't try using `assert`. Thanks for the tip!

pyk/src/pyk/kllvm/compiler.py
Robertorosmaninho: Done!

pyk/src/pyk/kllvm/compiler.py
Robertorosmaninho: Done!

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: We defined this will be addressed in a future PR in the LLVM backend, and then we'll come back here to update it appropriately.

pyk/src/tests/integration/kllvm/test_prooftrace.py
Baltoli: Rather than _parsing_ that pattern, just construct an instance of `pyk.kore.syntax.App` directly: https://github.com/runtimeverification/k/blob/54d416a4dfcc9dc09e5a19774fb91f700ee6ea09/pyk/src/pyk/kore/syntax.py#L475

Baltoli: The `pyk.kore.syntax` module defines data structures representing the KORE AST, the same as we have in the LLVM backend's AST module.

Robertorosmaninho: Done! Could you please review this and the whole PR once more? Sorry about the trouble!

PR #4292, opened 2024-04-27
Minimize boilerplate when building Commands/Options

### LINKED ISSUE #4188: Model commands and options with composition

nwatson22:


nwatson22:
Current issue I'm having:
The `Option` class is ideally where we'll input every piece of data about an individual option, e.g. `--debug`.
It would be nice to impose some type restrictions on each `Option`, so it's declared as a specific type and its `default` parameter can only be of that type. However, when I use generics to tell the `Option` what type it is, there doesn't seem to be any good way in python to read that type at runtime, and when I try to pass the type in as a normal parameter of type `type`, then it doesn't accept union types, e.g. `Int | None`.

Edit: I solved this for now by using a separate class for each type of option we might want to have, some of which have special features, e.g. `StringOption`, `BoolOption`, `EnumOption`, `ReadFileOption`

pyk/src/pyk/cli/cli.py
```diff
@@ -1,8 +1,377 @@
...
+    def __init__(
+        self,
+        name: str,
+        aliases: Iterable[str] = (),
+        cmd_line_name: str | None = None,
+        toml_name: str | None = None,
+        optional: bool = False,
+        help_str: str | None = None,
+        metavar: str | None = None,
+        default: Any = 'NoDefault',
```
tothtamas28: Consider introducing a dedicated value (a _null object_) for this, e.g.

```python3
NO_DEFAULT: Final = object()
```

pyk/src/pyk/cli/cli.py
```diff
@@ -1,8 +1,377 @@
...
+            default='NoDefault',
+            required=self.is_optional,
+            type=int,
+            help=self.help_str,
+            metavar=self.metavar,
+            dest=self.name,
+        )
+
+
+class BoolOption(Option):
```
tothtamas28: It'd be best to avoid the need for this class hierarchy. It is possible I'm missing something, but I think you can just add to `Option` all the attributes necessary to call `add_argument`.

nwatson22: I was looking to hide the argparse implementation details, like `FileType`, but I guess this is not really necessary.

PR #4299, opened 2024-04-30
Minimizing branching constraints

### LINKED ISSUE #3836: Spurious constraints returned by backend as part of branching conditions
I am observing constraints that are entailed by the current state returned as part of the branching conditions and am attaching a relevant bug report. Specifically, I have is a single execution step resulting is a three-way branching on
```
1a) ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) <Int pow256
1b) ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) <=Int ( maxUInt256 -Int VV1_y_114b9705:Int )
1c) 0 <=Int ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int )
1d) VV2_z_114b9705:Int <=Int ( ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) +Int VV1_y_114b9705:Int )
```
and
```
2a) ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) <Int pow256
2b) ( ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) +Int VV1_y_114b9705:Int ) <Int VV2_z_114b9705:Int
2c) ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) <=Int ( maxUInt256 -Int VV1_y_114b9705:Int )
2d) 0 <=Int ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int )
```
and 
```
3a) ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int ) <Int pow256
3b) ( maxUInt256 -Int VV1_y_114b9705:Int ) <Int ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int )
3c) 0 <=Int ( ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int ) -Int VV2_z_114b9705:Int )
```

However, the state before the branching contains
```
#And ( { true #Equals 0 <=Int VV0_x_114b9705:Int } #And ( { true #Equals VV0_x_114b9705:Int <Int pow256 }
#And ( { true #Equals 0 <=Int VV1_y_114b9705:Int } #And ( { true #Equals VV1_y_114b9705:Int <Int pow256 }
#And ( { true #Equals 0 <=Int VV2_z_114b9705:Int } #And ( { true #Equals VV2_z_114b9705:Int <Int pow256 }
#And ( { true #Equals VV2_z_114b9705:Int <=Int ( VV0_x_114b9705:Int +Int VV1_y_114b9705:Int )
#And ( { true #Equals VV0_x_114b9705:Int <=Int ( maxUInt256 -Int VV1_y_114b9705:Int ) }
```
which means that constraints 1a), 1c), 2a), 2c), 3a), and 3c) are entailed by the pre-branching state and would ideally not be returned as branching constraints. I believe that understanding this, however, requires assistance from Z3, and cannot necessarily be done just using our simplifications.

Bug report:
[spurious-constraints.tar.gz](https://github.com/runtimeverification/haskell-backend/files/15149578/spurious-constraints.tar.gz)

PetarMax:
In response to [issue #3836](https://github.com/runtimeverification/haskell-backend/issues/3836) in the Haskell backend, this PR minimizes the set of branching constraints produced by the backend in the following way:
1. It isolates an intersection of all of the branching constraints provided - if this intersection is non-empty, then the branching should not depend on it (as otherwise there should be branches where its negation holds);
2. It checks whether this intersection is entailed by the current path condition, and:
    a. if so, removes the intersection from all of the branching constraints;
    b. if not, does not make any changes. 

This solution is sound, but not complete, since we check the intersection as a whole rather than each constraint individually because each check is a call to the SMT solver. However, since this check is done for only for branching results and only for cases in which such an intersection exists, which happens rarely, perhaps we could improve by check individually.

I have tested this with Kontrol and I can confirm that the spurious constraints described in [issue #3836](https://github.com/runtimeverification/haskell-backend/issues/3836) no longer appear.


pyk/src/pyk/cterm/symbolic.py
```diff
@@ -94,6 +96,28 @@ def kast_to_kore(self, kinner: KInner) -> Pattern:
...
+        dummy_config = self._definition.empty_config(sort=GENERATED_TOP_CELL)
+        pc_cterm = CTerm(dummy_config, constraints=[pc])
+        intersection_cterm = CTerm(dummy_config, constraints=intersection)
+        implication_check = self.implies(pc_cterm, intersection_cterm, bind_universally=True)
+        # The intersection is not entailed, there is nothing to be done
+        if implication_check.csubst is None:
+            return constraints
+        # The intersection is entailed and can be filtered out of the branching constraints
+        else:
+            return tuple(mlAnd(c for c in cs if c not in intersection) for cs in flattened_default)
```
jberthold, line 119: Note that getting a substitution back does not necessarily mean the implication is true/valid, the validity is signaled by the flag `satisfiable` (which is a historic misnomer and should be called `valid`). An implication can be valid without any substitution required (think `Config  => Config`), or else a substitution may exist but the resulting path conditions are not fulfilled.

IDK if `pyk` compensates for that.

PetarMax: I think in the `Config => Config` case the substitution will be empty, rather than non-existent. The other case, in which we get a substitution but the implication is not valid, I'm less sure of, perhaps @ehildenb can clarify. What I did was copy how target subsumption was checked.

Also, I think that we should definitely rename `satisfiable` to `valid`.

PetarMax: Actually, that information is not propagated to the class carrying the information. All I have is:
```
class CTermImplies(NamedTuple):
    csubst: CSubst | None
    failing_cells: tuple[tuple[str, KInner], ...]
    remaining_implication: KInner | None
    logs: tuple[LogEntry, ...]
```
but the way these are constructed ensures, I think, that `csubst == None` is equivalent to the implication being valid.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -94,6 +96,28 @@ def kast_to_kore(self, kinner: KInner) -> Pattern:
...
+    def minimize_constraints(self, constraints: tuple[KInner, ...], pc: KInner) -> tuple[KInner, ...]:
+        """Minimize given branching constraints with respect to a given path condition."""
+        # Determine intersection between all returned sets of branching constraints
+        flattened_default = tuple(flatten_label('#And', c) for c in constraints)
+        intersection = flattened_default[0]
+        for i in range(1, len(flattened_default)):
+            intersection = [c for c in intersection if c in flattened_default[i]]
+            # If intersection is empty, there is nothing to be done
+            if not intersection:
+                return constraints
```
tothtamas28: Consider using `set.intersection`:

```suggestion
        _intersection = set.intersection(*(set(cs) for cs in flattened_default))
        intersection = tuple(c for c in flattened_default[0] if c in _intersection)
```

Both this and the original implementation will fail on empty `constraints` though, so additionally that case has to be handled.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -125,12 +150,18 @@ def execute(
 
         state = CTerm.from_kast(self.kore_to_kast(response.state.kore))
         resp_next_states = response.next_states or ()
-        next_states = tuple(
-            NextState(
-                CTerm.from_kast(self.kore_to_kast(ns.kore)),
-                self.kore_to_kast(ns.rule_predicate) if ns.rule_predicate is not None else None,
+        branching_constraints = tuple(
+            self.kore_to_kast(not_none(s.rule_predicate)) if s.rule_predicate is not None else None
```
tothtamas28, line 152: All these `not_none` make me feel we should to refine the type a bit.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -94,6 +96,28 @@ def kast_to_kore(self, kinner: KInner) -> Pattern:
     def kore_to_kast(self, pattern: Pattern) -> KInner:
         return kore_to_kast(self._definition, pattern)
 
+    def minimize_constraints(self, constraints: tuple[KInner, ...], pc: KInner) -> tuple[KInner, ...]:
```
tothtamas28: What does `pc` stand for? Consider using a more expressive name.

pyk/src/tests/integration/cterm/test_simple.py
```diff
@@ -35,6 +38,65 @@
 
 SIMPLIFY_TEST_DATA: Final = (('bytes-return', ('mybytes', '.Map'), (r'b"\x00\x90\xa0\n\xa1\xf1a" ~> .K', '.Map')),)
 
+equals = KLabel(name='#Equals', params=(KSort(name='Bool'), KSort(name='GeneratedTopCell')))
```
tothtamas28: Check `prelude.ml`, there might already be a helper function for `#Equals`.

tothtamas28: Also, `KLabel` is `Callable`, so instead of

```python3
KApply(label=equals, args=[...])
```

you can write

```python3
equals(...)
```

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -125,12 +150,18 @@ def execute(
...
-            NextState(
-                CTerm.from_kast(self.kore_to_kast(ns.kore)),
-                self.kore_to_kast(ns.rule_predicate) if ns.rule_predicate is not None else None,
+        branching_constraints = tuple(
+            self.kore_to_kast(not_none(s.rule_predicate)) if s.rule_predicate is not None else None
+            for s in resp_next_states
+        )
+        # Branch constraint minimization makes sense only if there is a proper branching
+        if len(branching_constraints) >= 2 and all(bc is not None for bc in branching_constraints):
+            branching_constraints = self.minimize_constraints(
```
tothtamas28, line 157: A few questions:

1. Is removing the extra constraints necessary for proving, or just an aesthetic thing?
2. Do we understand how these spurious constraints are introduced?
3. How general is this solution? I'd like to avoid the situation that we keep extending this class with new herusitics to handle spurious constraints that we might run into.
4. Can this be done on the backend more efficiently?

PetarMax: 1) Removing the extra constraints is not necessary per se for proving, but it is important, for two reasons:
    a) "objectively", these constraints are not real branching constraints, and the intended meaning of the `rule_predicate` field, coming from the backend, is to only contain real branching constraints - I believe that the backend team is in agreement on this (@jberthold?);
    b) subjectively, as part of the user experience of the person doing the proofs, I have found it important to know precisely what we are branching on, and I found these additional constraints both confusing and non-informative.
2) Yes, we understand this. They are introduced because the backend, when branching, only checks satisfiability of obtained constraints given the path condition, and does not check if there are some constraints coming from the `requires` clauses of the rules on which the branching happens that are entailed by the path condition using Z3, but only using the simplifications at hand.
3) I believe that this solution will cover the great majority of the cases we will encounter, and can't think right now of reasonable scenarios in which it wouldn't. For one, I don't see how there would be a constraint that is in the intersection that is not entailed by the path condition. @jberthold, perhaps you have some more thoughts on this?
4) @jberthold, what do you think? I think that, since this is a heuristic, it could be the case that it fits `pyk` better. The complexity of the procedure would be the same - find intersection, call SMT solver to check entailment. However, given 1a) above, perhaps this reasoning should indeed go to the backend.

jberthold: > `1.` ...the intended meaning of the `rule_predicate` field, coming from the backend, is to only contain real branching constraints - I believe that the backend team is in agreement on this (@jberthold?);
>
>  `4.` @jberthold, what do you think? I think that, since this is a heuristic, it could be the case that it fits `pyk` better. The complexity of the procedure would be the same - find intersection, call SMT solver to check entailment. However, given 1a) above, perhaps this reasoning should indeed go to the backend.

We could probably implement this filtering in the backend, but when we discussed the open github issue, we concluded that the frontend would be a better place to implement it.
Adding such a filter changes the understanding of what the branching condition is. At the moment, the backend returns, for each branch of a branching result, _all_ constraints that were _added_ on that branch (from a rule's side conditions. We are looking specifically at branches that arise from conditional rules where the condition is not determined)). As we found in the case at hand, some of the constraints are redundant but they are the basis of the state split that  happens in the branch.

> 3. ... I don't see how there would be a constraint that is in the intersection that is not entailed by the path condition. @jberthold, perhaps you have some more thoughts on this?

I don't have a ready-made example, and I imagine this to not be a useful scenario in practice, but I could imagine that one can construct a case where an execution branches into, say 3 possible next states, but then one of those 3 states is filtered out because of other, unrelated, path conditions. The two remaining next states of the branch may include a branching condition that is not entailed yet (i.e., constitutes new knowledge about the state).

PetarMax: I've talked with @ehildenb about this as well and he also thinks `pyk` is the correct place, for now, for this filtering.

PetarMax: > I don't have a ready-made example, and I imagine this to not be a useful scenario in practice, but I could imagine that one can construct a case where an execution branches into, say 3 possible next states, but then one of those 3 states is filtered out because of other, unrelated, path conditions. The two remaining next states of the branch may include a branching condition that is not entailed yet (i.e., constitutes new knowledge about the state).

I'm not sure that's possible without using rules that explicltly cut a part of the path condition (like, for example, the `#assume` of KEVM):
```
rule <k> #assume(B) => .K ... </k> ensures B
```
and such rules, in my experience, are only found as part of an analysis on top of an actual semantics, and would also likely not result in a branching.

To get more intuition on this, let's assume that the applied rules don't cut a part of the path condition, and that we have three incoming branches, two of which are feasible and have the same extra constraint `pc_1` that is not entailed by the current path condition, `pc`, and a third infeasible branch, whose overall path condition is in contradiction with `pc`. The third branch must contain`!pc_1`, as otherwise it would either hold that `pc |= pc_1` or the rules would cut paths, both of which are not possible by assumption.

We also know that that contradiction cannot be caused by `!pc_1`, because if `pc /\ !pc1` is a contradiction then `pc` would entail `pc_1`. Therefore, it must be caused by some other new constraint introduced in the third branch, say, `pc_2`. However, we would then be missing a branch in which`!pc_1 /\ !pc_2` holds, which is in contradiction with the assumption that no paths are cut.

pyk/src/pyk/cterm/symbolic.py
PetarMax: Oh, sorry, `path_condition`.

pyk/src/pyk/cterm/symbolic.py
PetarMax: Yes, understood. However, at this point, I am not sure what guarantees the backend provides when it comes to the branching constraints, and so I kept the `| None` option in the type. @jberthold, any insight, perhaps?

PetarMax: Perhaps if `len(branching_constraints) >= 2`, then we should expect `rule_predicate` to be non-empty and just do `not_none`?

pyk/src/pyk/cterm/symbolic.py
PetarMax: By construction, this function should not be called with empty `constraints` or just one set of constraints. I'll add an appropriate assertion.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -94,6 +96,28 @@ def kast_to_kore(self, kinner: KInner) -> Pattern:
...
+        # Determine intersection between all returned sets of branching constraints
+        flattened_default = tuple(flatten_label('#And', c) for c in constraints)
+        intersection = set.intersection(*(set(cs) for cs in flattened_default))
+        # If intersection is empty, there is nothing to be done
+        if not intersection:
+            return constraints
+        # Check if non-empty intersection is entailed by the path condition
+        dummy_config = self._definition.empty_config(sort=GENERATED_TOP_CELL)
+        path_condition_cterm = CTerm(dummy_config, constraints=[path_condition])
+        intersection_cterm = CTerm(dummy_config, constraints=intersection)
```
tothtamas28, line 112: Are constraints of a `CTerm` sorted? If not, they should be sorted before being passed to the constructor.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -94,6 +96,28 @@ def kast_to_kore(self, kinner: KInner) -> Pattern:
     def kore_to_kast(self, pattern: Pattern) -> KInner:
         return kore_to_kast(self._definition, pattern)
 
+    def minimize_constraints(self, constraints: tuple[KInner, ...], path_condition: KInner) -> tuple[KInner, ...]:
+        """Minimize given branching constraints with respect to a given path condition."""
+        # By construction, this function is to be called with at least two sets of constraints
+        assert len(constraints) >= 2
+        # Determine intersection between all returned sets of branching constraints
+        flattened_default = tuple(flatten_label('#And', c) for c in constraints)
```
tothtamas28: Nitpick:

```suggestion
        flattened_default = [flatten_label('#And', c) for c in constraints]
```

(This reads nicer, and `flattened_default` does not need to be immutable.)

pyk/src/tests/integration/cterm/test_simple.py
```diff
@@ -35,6 +38,65 @@
...
+
+MINIMIZE_CONSTRAINTS_TEST_DATA: Final = (
+    (
+        'no-intersection',
+        (
+            (KApply(label=equals, args=[TRUE, KApply('_<=Int_', args=(intToken(0), KVariable('X')))]),),
```
tothtamas28: There a few ways to make this read nicer:
* `KLabel` is `Callable` (returns a `KApply`)
* `KLabel.__call__` is overloaded to take `*args: KInner`
* `pyk.prelude.utils.token`
* `pyk.prelude.kint.leInt`

```suggestion
            (equals(TRUE, leInt(token(0), KVariable('X'))),),
```

tothtamas28: But it seems all constraints follow a certain pattern, so consider extracting a function for that.

pyk/src/tests/integration/cterm/test_simple.py
```diff
@@ -60,6 +122,10 @@ def config(kprint: KPrint, k: str, state: str, constraint: str | None = None) ->
             (_constraint,),
         )
 
+    @staticmethod
+    def constraint(kprint: KPrint, c: str) -> KInner:
+        return kprint.parse_token(KToken(c, BOOL))
```
tothtamas28: Where is this called?

pyk/src/pyk/cterm/symbolic.py
PetarMax: Yes, they are sorted.

pyk/src/pyk/cterm/symbolic.py
PetarMax: Yes, done.

pyk/src/tests/integration/cterm/test_simple.py
PetarMax: It's not, I'll remove it, I think I was using as a helper to print some debugging information for the test.

pyk/src/tests/unit/test-data/pyk_toml_test.toml
```diff
@@ -1,6 +1,6 @@
 [coverage]
 output = "default-file"
-definition = "/tmp"
+definition = "/var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T"
```
tothtamas28: Is this change related?

tothtamas28 (approved):

pyk/src/tests/unit/test-data/pyk_toml_test.toml
PetarMax: No, I haven't made that change, at least not manually...

PetarMax: I've reverted it, and automatically removed some spaces from that file.

PR #4300, opened 2024-04-30
Fix for `indexed_rewrite` missing rewrite rules

### LINKED ISSUE #4297: Pyk `indexed_rewrite` misses valid rewrites
There is a typo in the implementation of this function: https://github.com/runtimeverification/k/blob/7d92b56e645f311f2645db57a963501a95fb2773/pyk/src/pyk/kast/manip.py#L609-L645

The solution will be to fix the typo and add proper tests for this rewriting implementation.

This issue was originally identified by @PetarMax in Kontrol; it manifested there as random failures of the alias-unapplier to identify aliases it should undo.

Baltoli:
There was an issue (caused by a typo) in Pyk's `indexed_rewrite` function whereby the indexer would only ever consider a single rule for any given top-level symbol. For example, given the rewrites:
```
a(b()) => b()
a(c()) => c()
```
the rewriter would only ever rewrite one of `a(b())` or `a(c())`, depending on iteration order.

This manifested as an issue in Kontrol, where `#range(...)` aliases were not expanded consistently across different people's machines and CI.

This PR fixes the typo, and adds:
* Unit tests including a minimised reproduction of the underlying issue
* Integration tests of the original KEVM code

Fixes https://github.com/runtimeverification/k/issues/4297


tothtamas28 (approved):
:sparkles:

PR #4303, opened 2024-05-01
Enable `--no-haskell-binary` by default on macOS

Baltoli:
A long standing nit I've had with K (and never got around to fixing) is that you need to remember to pass `--no-haskell-binary` to `kompile` on macOS, or K will crash when trying to execute any programs. This is annoying and bad UX, so this PR amends the frontend behaviour to enable this setting by default on macOS.

### DIFF
--- a/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKompileOptions.java
+++ b/haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKompileOptions.java
@@ -22,7 +22,7 @@ public class HaskellKompileOptions {
       names = "--no-haskell-binary",
       description =
           "Use the textual KORE format in the haskell backend instead of the binary KORE format."
-              + " This flag is enabled by default and cannot be disabled when running on macOS because of a bug in the underlying Haskell libraries.",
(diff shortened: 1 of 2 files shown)


tothtamas28 (approved):

haskell-backend/src/main/java/org/kframework/backend/haskell/HaskellKompileOptions.java
```diff
@@ -22,7 +22,7 @@ public HaskellKompileOptions() {}
       names = "--no-haskell-binary",
       description =
           "Use the textual KORE format in the haskell backend instead of the binary KORE format."
-              + " This is a development option, but may be necessary on MacOS due to known issues"
-              + " with the binary format.")
+              + " This flag is enabled by default and cannot be disabled when running on macOS because of a bug in the underlying Haskell libraries.",
+      hidden = true)
```
goodlyrottenapple:
```suggestion
              + " This flag is disabled by default, except when running on macOS, where it is enabled and cannot be disabled because of a bug in the underlying Haskell libraries.",
      hidden = true)
```

PR #4305, opened 2024-05-01
Renaming satisfiable to valid for `ImpliesResult`

PetarMax:
This PR renames the misleading `satisfiable` field of `ImpliesResult` to `valid`, since the `implies` endpoint checks for validity of implication, not satisfiability.


PetarMax:
If we agree on this change, it would be good for the backend to also return `valid` in its response.

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -212,11 +212,11 @@ def implies(
         except SmtSolverError as err:
             raise self._smt_solver_error(err) from err
 
-        if not result.satisfiable:
+        if not result.valid:
             if result.substitution is not None:
-                _LOGGER.debug(f'Received a non-empty substitution for unsatisfiable implication: {result.substitution}')
+                _LOGGER.debug(f'Received a non-empty substitution for non-valid implication: {result.substitution}')
```
tothtamas28:
```suggestion
                _LOGGER.debug(f'Received a non-empty substitution for falsifiable implication: {result.substitution}')
```

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -212,11 +212,11 @@ def implies(
...
             if result.predicate is not None:
-                _LOGGER.debug(f'Received a non-empty predicate for unsatisfiable implication: {result.predicate}')
+                _LOGGER.debug(f'Received a non-empty predicate for non-valid implication: {result.predicate}')
```
tothtamas28:
```suggestion
                _LOGGER.debug(f'Received a non-empty predicate for falsifiable implication: {result.predicate}')
```

pyk/src/pyk/cterm/symbolic.py
tothtamas28: By the way, whats the expected relationship between `valid` being `True` and `substitution` / `predicate` being `None`?

tothtamas28 (commented):
LGTM.

Please wait for someone from the backend team to approve. Let's also open an issue for the renaming on the backend.

pyk/src/pyk/cterm/symbolic.py
PetarMax: I believe it's the following:
- if implication is valid, backend must return a non-none substitution and predicate
- otherwise, backend may return non-none substitution or predicate, but they will be ignored (with warnings) and the substitution coming back from `CTermSymbolic.implies` will be set to `None`

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -255,7 +255,7 @@ def implies(
         if result.predicate is None:
             raise ValueError('Received empty predicate for valid implication.')
         ml_subst = self.kore_to_kast(result.substitution)
-        ml_pred = self.kore_to_kast(result.predicate) if result.predicate is not None else mlTop()
```
PetarMax, line 258: Line 255 above guarantees that `result.predicate` is not `None` and the type checker agrees.

pyk/src/pyk/cterm/symbolic.py
jberthold: I first thought we would possibly omit empty substitutions/predicates, but indeed the endpoint returns 
`substitution: #Top` and `predicate: #Top` in those cases.

jberthold:
> If we agree on this change, it would be good for the backend to also return `valid` in its response.

We have this [on our list for a long time](https://github.com/runtimeverification/haskell-backend/issues/3601) but the cost/benefit ratio was a bit too low to actually do it independently. @goodlyrottenapple is working on a `check-implication` endpoint in booster so we will rename this in the backend shortly.

jberthold (approved):
LGTM

PR #4308, opened 2024-05-02
Reorganise CI to use pre-packaged LLVM backend release

Baltoli:
This PR refactors the way that we build Docker images with K installed to run later CI stages (i.e. the Pyk integration tests). Rather than building the entire K distribution from scratch, we instead download and install a pre-packaged release of the LLVM backend. A subsequent (much smaller) change will do the same for the Haskell backend, now that the supporting infrastructure is in place.

I don't expect that CI times will actually decrease meaningfully with this PR; the LLVM backend is pretty quick to build. Doing the same for the haskell backend will be a much bigger impact, and should bring CI times down to ~20 minutes.

The existing CI job that produces a "whole of K" distribution is still run, and we still release that package on new K versions.


Baltoli:
~~Should be almost ready to go; blocked on https://github.com/runtimeverification/llvm-backend/pull/1085 getting upstreamed to K.~~

.github/workflows/test-pr.yml
```diff
@@ -328,7 +367,7 @@ jobs:
           set -euxo pipefail
           workspace=$(pwd)
           docker run --name k-profiling-tests-${GITHUB_SHA}                    \
-            --rm -it --detach                                                  \
+            --rm -it --detach --user root                                      \
```
TheBlueLizard, line 370: What changed between now and before that requires running as root in the container? Not an ideal scenario if we're publishing our tool but only works when used by 'root' -- primary concern for this change.

F-WRunTime, line 370: What changed between now and before that requires running as root in the container? Not an ideal scenario if we're publishing our tool but only works when used by 'root' -- primary concern for this change.

Baltoli: Nothing has changed in terms of what requires root to run; it's because we're now using a K base image that was set up to use the github-user setup; previously this step ran an unmodified base Ubuntu image which ran as `root` by default.

The root user is only needed to run the benchmarking setup script, which installs packages using `apt`; it's not that we're publishing a tool that requires it.

This code could be refactored into an actual Dockerfile rather than an ad-hoc script to do the setup, but that's probably out of scope for this PR.

Scott-Guest (approved):
This all LGTM, but I am admittedly not that familiar with Debian packaging, so I'd wait for further approval to merge.

.github/actions/with-k-docker/Dockerfile
```diff
@@ -1,19 +1,26 @@
-FROM ubuntu:jammy
+ARG BASE_OS
```
tothtamas28: Do we use this image with any `BASE_OS` other than `ubuntu`? If not, consider keeping this value concrete for simplicity.

.github/actions/with-k-docker/action.yml
```diff
@@ -10,23 +10,61 @@ inputs:
...
+        --build-arg LLVM_BACKEND_DEB_PATH=${LLVM_BACKEND_DEB_PATH}
+
+  - name: 'Run Docker container'
+    shell: bash {0}
+    if: ${{ inputs.run-container == 'true' }}
+    run: |
+      set -euxo pipefail
+
+      CONTAINER_NAME=${{ inputs.container-name }}
+      TAG=runtimeverificationinc/${CONTAINER_NAME}
```
tothtamas28, line 60: Consider extracting these into the `env` to avoid defining them in both jobs.

tothtamas28 (approved):

Issue #4027, opened 2024-02-22
Represent attributes with no value explicitly


tothtamas28:
Currently an attribute `foo` or `foo()` is represented internally as `{"foo": ""}`, i.e. with an empty string as value. As a consequence, `foo()` and `foo("")` are indistinguishable.

In an attempt to work around this, `ModuleToKORE` has an internal pass that inspects all attributes and checks whether any occurrence has a value other than `""`: https://github.com/runtimeverification/k/blob/5a0ffc984a9d4b01687b7d3ccc6a0f49e99e50eb/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L188

This does not work well for attributes with an optional string value, e.g. `symbol`: if all occurrences of `symbol` in a given definition are without a value, then the resulting KORE attribute for `symbol()` becomes

```
symbol'Kywd'{}()
```

whereas if there is one then

```
symbol'Kywd'{}("")
```

So now backends have to process both forms as `symbol()`. This works in this particular case as `""` is not a valid `symbol` value, but in general this is problematic.

As a solution, attributes with missing values should have a separate representation. This is already so for textual K and KORE syntax, but not for the internal and JSON representations.

Baltoli:
The attribute class in Java should be extended to have an additional state that encodes "there is no attribute value here".

Baltoli:
cases:
- `att` -> no value
- `att()` -> empty string
- `att("")` -> empty string

PR #4313, opened 2024-05-03
Source and Location attributes for outer parsed modules

gtrepta:
The Java frontend is going to be intolerant of missing source and location attributes in various places after outer parsing, so we need to make sure any KAST that we pass to it when skipping its outer parser has them.

This implements those attributes for modules, taking a step towards getting the initial end to end compilation with pyk's outer parser.


pyk/src/tests/unit/kast/test_outer_parser.py
```diff
@@ -382,15 +384,42 @@ def test_import(k_text: str, expected: AST) -> None:
...
+            'FOO',
+            sentences=(Rule('x'),),
+            imports=(Import('BAR'),),
+            att=Att(
+                (
+                    ('foo', ''),
+                    (LOCATION, '1,1,1,46'),
+                )
+            ),
+        ),
```
Baltoli: This all looks good; could we perhaps have a unit test with multiple lines using a docstring?

Baltoli (approved):

pyk/src/pyk/kast/att.py
```diff
@@ -123,6 +123,11 @@ def pretty(self, value: str) -> str:
 
 class LocationType(AttType[tuple[int, int, int, int]]):
     def from_dict(self, obj: Any) -> tuple[int, int, int, int]:
+        if isinstance(obj, str):
```
tothtamas28: There should be a single JSON representation for each attribute. Which component emits this format?

gtrepta: The `Att` class in `outer_syntax` currently only accepts `tuple[str,str]` for its items, so I'm encoding the location in a string. The items get turned into a dictionary and passed here when converting to a `KAtt` so I needed logic to decode it.

pyk/src/tests/unit/kast/test_outer_parser.py
gtrepta: If you mean to exercise the line numbers in the location attribute, that's happening with the unit tests for the `parse_outer` function which take multiline source files.

pyk/src/pyk/kast/att.py
tothtamas28: You can handle this by adding a new method to `AttType`:

```python3
@abstractmethod
def parse(self, text: str) -> T: ...
```

(And consider renaming `pretty` to `unparse` to maintain symmetry.)

gtrepta: Do you think the check and call to `parse` should happen in the `AttType`'s `from_dict` method, or in `KAtt`'s `from_dict`?

tothtamas28: You can call this when you convert `outer_syntax.Att` to `KAtt`: you look up `AttKey` based on the left element of the tuple, then parse the right element using the corresponding `AttType`.

gtrepta: There's a problem here. `LocationType.parse` needs to return `tuple[int, int, int, int]` (because that's what the type variable `T` is assigned to in the superclass), so its result can't be passed back in to `LocationType.from_dict`, it will expect a `list`.

tothtamas28: There's no need to call `from_dict`, you can construct the `AttEntry` directly.

gtrepta: Oh, I think I see. `parse` can return the entire `AttType[T]` and not just a `T`

gtrepta: nvm. I just confused myself there.

But what about what I have implemented right now, where `KAtt.from_dict` picks between calling `from_dict` and `parse` depending on the type of the value? It seems like a pretty sensible place to determine when to parse a string and it's localized to `att.py`

pyk/src/tests/unit/kast/test_outer_parser.py
Baltoli: Great!

pyk/src/pyk/kast/att.py
tothtamas28: `from_dict` is for parsing the JSON format, i.e. `compiled.json`. Parsing attributes from the textual format is a new use case. In my opinion, adding a new method is the right thing to do.

pyk/src/pyk/kast/att.py
```diff
@@ -79,9 +87,12 @@ def from_dict(self, obj: Any) -> Any:
     def to_dict(self, value: Any) -> Any:
         return self._unfreeze(value)
 
-    def pretty(self, value: Any) -> str:
+    def unparse(self, value: Any) -> str:
         return str(value)
 
+    def parse(self, text: str) -> Any:
+        raise ValueError(f"Parsing a string into an Any attribute type is not allowed. Attempted to parse '{text}'")
```
tothtamas28:
```suggestion
        raise ValueError(f'Parsing a string into an Any attribute type is not supported. Attempted to parse: {text!r}')
```

pyk/src/pyk/kast/att.py
```diff
@@ -34,7 +34,10 @@ def from_dict(self, obj: Any) -> T: ...
     def to_dict(self, value: T) -> Any: ...
 
     @abstractmethod
-    def pretty(self, value: T) -> str | None: ...
+    def unparse(self, value: T) -> str | None: ...
+
+    def parse(self, text: str) -> T:
+        return self.from_dict(text)
```
tothtamas28:
```suggestion
    def parse(self, text: str) -> T: ...
```

pyk/src/pyk/kast/att.py
```diff
@@ -134,9 +145,16 @@ def from_dict(self, obj: Any) -> tuple[int, int, int, int]:
     def to_dict(self, value: tuple[int, int, int, int]) -> Any:
         return list(value)
 
-    def pretty(self, value: tuple[int, int, int, int]) -> str:
+    def unparse(self, value: tuple[int, int, int, int]) -> str:
         return ','.join(str(e) for e in value)
 
+    def parse(self, text: str) -> tuple[int, int, int, int]:
+        r = re.compile('(\\d+),(\\d+),(\\d+),(\\d+)')
+        m = r.fullmatch(text)
```
tothtamas28: Consider either
- Caching the compiled value in a class attribute.
- Matching without explicitly compiling.

pyk/src/pyk/kast/att.py
```diff
@@ -327,7 +345,10 @@ def from_dict(cls: type[KAtt], d: Mapping[str, Any]) -> KAtt:
         entries: list[AttEntry] = []
         for k, v in d['att'].items():
             key = Atts.keys().get(k, AttKey(k, type=_ANY))
-            value = key.type.from_dict(v)
+            if isinstance(v, str) and not isinstance(key.type, AnyType):
+                value = key.type.parse(v)
+            else:
+                value = key.type.from_dict(v)
```
tothtamas28: This method shouldn't be smart either. Instead, a new method should be introduced:

```python3
@classmethod
def parse(cls, dct: Mapping[str, str]) -> KAtt:
    ...
```

pyk/src/pyk/kast/outer_parser.py
```diff
@@ -62,11 +63,13 @@ class OuterParser:
...
     _la: Token
     _la2: Token
+    _source: str
 
-    def __init__(self, it: Iterable[str]):
+    def __init__(self, it: Iterable[str], source: str = ''):
         self._lexer = outer_lexer(it)
         self._la = next(self._lexer)
         self._la2 = next(self._lexer, _EOF_TOKEN)
+        self._source = source
```
tothtamas28, line 72: Consider attaching the source in the AST -> KAst transformation instead.

pyk/src/pyk/kast/outer_parser.py
```diff
@@ -120,8 +125,17 @@ def module(self) -> Module:
         while self._la.type is not TokenType.KW_ENDMODULE:
             sentences.append(self.sentence())
 
+        end_loc = self._la.loc + self._la.text
         self._consume()
 
+        location = (
```
tothtamas28: Consider making this an attribute of AST elements instead. That also makes parsing location attribute strings unnecessary.

pyk/src/pyk/kast/outer_parser.py
```diff
@@ -120,8 +125,17 @@ def module(self) -> Module:
...
+            (
+                Atts.LOCATION.name,
+                f'{begin_loc.line},{begin_loc.col},{end_loc.line},{end_loc.col}',
+            ),
+        )
+        source = ((Atts.SOURCE.name, self._source),) if self._source else ()
+        att = Att(att.items + location + source)
```
tothtamas28: My problem with this is now the text and the data structure are no longer one-to-one. This can be an issue for example if we want to use the parser for reading in K text as test data for kompiler pass unit tests.

tothtamas28 (changes requested):

pyk/src/pyk/kast/outer_parser.py
gtrepta: I prefer that it's available here in the OuterParser, it can be used to print the source file in the case of a parsing error (see 1ffd6820cb63aceee1c7a8ebbccd6d6fcdde2985)

pyk/src/pyk/kast/outer_parser.py
gtrepta: These attributes wont be added at parse time now, but are still going to be added to the `KAst` objects when converting from `AST`, would adding a flag to disable that be helpful?

pyk/src/pyk/kast/outer_parser.py
tothtamas28: Right, but the parser operates on a stream of characters, source file is not a notion on its abstraction level.

You can still include the source file in error messages with error translation:

```python
try:
    definition = OuterParser(path.read_text()).definition()
except ValueError as err:
    raise ValueError(f'Parse error: {path}: {err}') from err
```

pyk/src/pyk/kast/outer_parser.py
tothtamas28: I can't think of a use case yet where not adding them would be necessary, let's just add them unconditionally for now.

pyk/src/pyk/kast/att.py
```diff
@@ -46,7 +49,11 @@ def to_dict(self, value: None) -> Any:
         assert value is None
         return ''
 
-    def pretty(self, value: None) -> None:
+    def unparse(self, value: None) -> None:
+        return None
+
+    def parse(self, text: str) -> None:
+        assert text == ''
```
tothtamas28, line 56: This is not something that we can address right now, just something we eventually should:
* #4027

pyk/src/pyk/kast/outer_parser.py
```diff
@@ -62,20 +63,28 @@ class OuterParser:
...
+    def _error_location_string(self, t: Token) -> str:
+        if not self._source:
+            return ''
+        return f'{self._source}:{t.loc.line}:{t.loc.col}: '
+
     def _match(self, token_type: TokenType) -> str:
         if self._la.type != token_type:
-            raise ValueError(f'Expected {token_type.name}, got: {self._la.type.name}')
+            location_string = self._error_location_string(self._la)
+            raise ValueError(location_string + f'Expected {token_type.name}, got: {self._la.type.name}')
```
tothtamas28:
```suggestion
            raise ValueError(f'{location_string}Expected {token_type.name}, got: {self._la.type.name}')
```

tothtamas28: Consider extracting a function:

```python3
def _unexpected_token(token: Token, expected_types: Iterable[TokenType] = ()) -> ValueError:
    ...
```

This I think covers all three modified cases.

tothtamas28 (approved):

pyk/src/pyk/kast/outer_parser.py
tothtamas28: On the other hand, it indeed makes sense to include the path in the `Ast` along with the location as you did, so I accept your argument.

In that case however, instead of having an optional parameter for the `Path`, consider a polymorphic `__init__` that either takes an `Iterable[str]` or a `Path`.

gtrepta: Do you mean taking a `Path` instead of an `Iterable[str]` for the source text? Then the `OuterParser` would have to handle `.md` files.

I'm adding automerge, so if I didn't understand what you mean we can follow up on it in a later PR.

PR #4322, opened 2024-05-07
Add class `Formatter`

tothtamas28:
Implements pretty-printing KAST terms based on the `format` attribute.

### DIFF
--- a/pyk/src/tests/integration/kast/test_formatter.py
+++ b/pyk/src/tests/integration/kast/test_formatter.py
@@ -1,0 +1,108 @@
+from __future__ import annotations
+
+import textwrap
+from itertools import count
+from typing import TYPE_CHECKING
+
+import pytest
+
+from pyk.kast.formatter import Formatter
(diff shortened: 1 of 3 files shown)


pyk/src/tests/integration/kast/test_formatter.py
gtrepta, line 1: This looks like it should be a unit test.

tothtamas28: The reason I thought this has to be tested in integration is that this way we test from textual format attribute to output. Note that the test is pretty fast as-is since it uses the `IMP` definition which on a given test worker only kompiles once (per backend).

But adding unit tests won't hurt either so I'll add a few.

Baltoli (approved):

pyk/src/tests/integration/kast/test_formatter.py
tothtamas28: Done.

gtrepta (approved):

PR #4323, opened 2024-05-07
Fix main prover in parallel_advance_proof not being closed

nwatson22:
`main_prover`, the `Prover` used by the main thread in `parallel_advance_proof` was not being closed. This fixes it by wrapping the proof with a `with create_prover() as main_prover:` block.


ehildenb:
Can we test this?

nwatson22:
> Can we test this?

@ehildenb I added a test to check `close()` is called the same number of times a Prover is created.

pyk/src/pyk/proof/proof.py
```diff
@@ -344,43 +344,42 @@ def parallel_advance_proof(
     explored: set[PS] = set()
     iterations = 0
 
-    main_prover = create_prover()
+    with create_prover() as main_prover:
+        main_prover.init_proof(proof)
 
-    main_prover.init_proof(proof)
+        with _ProverPool[P, PS, SR](create_prover=create_prover, max_workers=max_workers) as pool:
```
tothtamas28, line 350: What's the relationship between `main_prover` and the provers created with `create_prover`?

nwatson22: `main_prover` is also created with `create_prover`, but the difference is that the other provers are created by `_ProverPool` as needed for each thread, but `main_prover` is created directly in the proving loop and is used on the main thread to `init_proof` and compute failure info.

tothtamas28: Do the other provers also call  `init_proof` and compute `failure_info`? If not, these are two disjoint roles, and the interfaces can be segregated.

tothtamas28 (approved):

pyk/src/pyk/proof/proof.py
nwatson22: They do not. Really the only thing that's shared is the `kcfg_explore` is used for both. (and `advance_proof`, the older sequential proving loop, is a member of `Proof`, so it uses `self` for all of these functions. I've opened an issue for this at #4333.

PR #4327, opened 2024-05-08
Add test for proof progressing on reload

### LINKED ISSUE #4190: Proof progresses upon reload
While trying to add a test for failure_info to [test_failure_info_recomputed_on_proof_reinit](https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L1070), I noticed strange behavior where the proof is written to and then read from disk.

The claim, which terminates early, has 1 terminal and 1 pending node when it's written to the disk, but after it's reloaded and is used to initialize APRProver, it has 2 terminal and 0 pending node — in other words, it looks like the pending node gets executed during APRProver initialization:

```py
        # reload proof from disk
        proof = APRProof.read_proof_data(proof_dir, proof_id)
        prover = APRProver(proof, kcfg_explore=kcfg_explore)
```

To reproduce, you can comment out skipping the [test](https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L733) that fails early with `fail_fast` in  [test_failure_info_recomputed_on_proof_reinit](https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L1070):
https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L1084-L1085

nwatson22:
- Adds test `test_proof_no_progress_on_reload` to demonstrate that nodes are no longer being marked as terminal simply by loading the proof. 
- Re-enables the `fail-fast` test in `test_failure_info_recomputed_on_proof_reinit`, which was still failing because `fail_fast`  parameter wasn't getting passed to `advance_proof`.


palinatolmach (approved):

pyk/src/tests/integration/proof/test_imp.py
```diff
@@ -1066,6 +1066,52 @@ def test_failure_info(
...
+        assert expected_pending == actual_pending
+        assert expected_failing == actual_failing
+
+        # reload proof from disk
+        proof = APRProof.read_proof_data(proof_dir, proof_id)
+        prover = APRProver(kcfg_explore=kcfg_explore)
+        prover.advance_proof(proof, fail_fast=True)
+
+        failure_info = proof.failure_info
+        assert isinstance(failure_info, APRFailureInfo)
```
PetarMax: Is this `failure_info` different from the previous? Should we compare?

pyk/src/tests/integration/proof/test_imp.py
```diff
@@ -1066,6 +1066,52 @@ def test_failure_info(
...
+        proof = APRProof.from_claim(kprove.definition, claim, logs={}, proof_dir=proof_dir)
+        proof_id = proof.id
+        kcfg_explore.simplify(proof.kcfg, {})
+        prover = APRProver(kcfg_explore=kcfg_explore)
+        prover.advance_proof(proof, fail_fast=True)
+
+        failure_info = proof.failure_info
+        assert isinstance(failure_info, APRFailureInfo)
+
+        actual_pending = len(failure_info.pending_nodes)
```
PetarMax: Could we have a stronger comparison, perhaps of the associated node ids? This would mean no progress, the current test only makes sure that there is one pending and one failing node, but not necessarily that it's the same ones.

pyk/src/tests/integration/proof/test_imp.py
nwatson22: It should be the same. It is now comparing the FailureInfo from before and after the reload.

pyk/src/tests/integration/proof/test_imp.py
nwatson22: I'm comparing the whole FailureInfo object before and after the reload now, so it is now checking the exact node ids are the same, not just the number.

PR #4334, opened 2024-05-09
Parse regular expressions

### LINKED ISSUE #4295: Specify and parse permitted regex syntax
Currently, regex terminals are passed through to Flex unchanged and unvalidated.

We should instead decide on the exact regex syntax we want to support, parse it ourselves in the frontend, then reprint it for usage by Flex. This will
- Provide better error messages
- Decouple K from Flex, allowing us to decide on our own syntax and supported features rather than "whatever Flex does"
- Provide a structured representation of the regex in the frontend for use in issues like #4271

Scott-Guest:
Closes #4295

- Implements a parser for the new regex syntax as given by the grammar in #4295, then re-emits for Flex. 
- Fix the regression-new test suite to also strip out absolute paths to `domains.md` from test outputs

All downstream semantics have already been updated to the new syntax, and if anything was missed, the new error messages should make the fix obvious.

In a follow up PR, I still plan to:
- Update the documentation with the grammar in #4295.
- Validate that all named lexical elements actually exist in the K definition. We still don't check this, and Flex will just error if the identifier doesn't exist.
- Properly handle Unicode characters in regex
  - Flex is an 8-bit scanner, so any Unicode codepoint is treated as its sequence of bytes, and some translation is needed to get Unicode-aware semantics
  - Outside of a character class, parenthesize so `r"😊*"` becomes `r"(\xF0\x9F\x98\x8A)*"`
  - Inside a character class, convert to an explicit `|` so `r"[😊ab]"` becomes `r"(\xF0\x9F\x98\x8A)|[ab]"`
  - Character ranges and negated character classes don't have a straightforward translation, so just report an error


k-frontend/src/main/java/org/kframework/definition/regex/RegexBody.java
Baltoli, line 1: Who needs Scala? 😛

Baltoli (approved):
The design here is really nice; great work!

PR #4336, opened 2024-05-10
Add textual KAST parser to `pyk`

tothtamas28:
Known differences to the JavaCC implementation:
* Tokens `.::K` and `.::KList` are not supported.
* Token `#klabel` is not supported.
* When lexing `.K` and `.KList`, the lexer reads until special character. This allows a simple LA(1) implementation. On the other hand, e.g. `.KA` will raise an error.
* This double comma won't parse:
   https://github.com/runtimeverification/k/blob/7be107f58bb97c8f50989b60247c0180f80d88c8/k-frontend/src/main/javacc/KastParser.jj#L182
* When unquoting strings (`"Int"`) and klabels (`` `<k>` ``), a more straightforward decoding is used that simply replaces `\X` by `X`. This is in line with how these tokens are lexed.

Potential improvements to the format:
* Support `?X`.
* Support `X:Int`.
* Support klabels with sort arguments.
* Write `foo` or `foo()` instead of `foo(.KList)`.
* Write something like `Int("1")` (like an application) or `"1":Int` (like a variable) instead of `#token("1", "Int")` (though there might be a reason `KToken` is represented that way.)

### DIFF
--- a/pyk/src/pyk/kast/lexer.py
+++ b/pyk/src/pyk/kast/lexer.py
@@ -1,0 +1,202 @@
+from __future__ import annotations
+
+from collections.abc import Callable, Iterator
+from enum import Enum, auto
+from typing import TYPE_CHECKING, NamedTuple
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from typing import Final
+
+
+class TokenType(Enum):
+    EOF = auto()
+    LPAREN = auto()
+    RPAREN = auto()
+    COMMA = auto()
+    KSEQ = auto()
+    DOTK = auto()
+    DOTKLIST = auto()
+    TOKEN = auto()
+    ID = auto()
+    VARIABLE = auto()
+    KLABEL = auto()
+    STRING = auto()
+
+
+class Token(NamedTuple):
+    text: str
+    type: TokenType
+
+
+def lexer(text: Iterable[str]) -> Iterator[Token]:
+    it = iter(text)
+    la = next(it, '')
+    while True:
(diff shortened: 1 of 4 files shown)


pyk/src/pyk/kast/lexer.py
```diff
@@ -0,0 +1,202 @@
...
+
+            else:
+                buf.append(la)
+                la = next(it, '')
+
+    return sublexer
+
+
+def _kseq(la: str, it: Iterator[str]) -> tuple[Token, str]:
+    # assert la == '~'
```
Baltoli: There are a few of these comments; should they be removed or reinstated?

pyk/src/pyk/kast/lexer.py
```diff
@@ -0,0 +1,202 @@
...
+    while la in _VARIABLE_CHARS:
+        buf += la
+        la = next(it, '')
+    text = ''.join(buf)
+    return Token(text, TokenType.VARIABLE), la
+
+
+# For ease of implementation, KDOT and KDOTLIST tokens are read until _SEP
+# This allows LA(1)
+# But e.g. .KA won't be lexed, even though it can be read as [KDOT, VARIABLE]
```
Baltoli, line 170: This is fine and matches my intuitive sense for how that would work.

Baltoli (approved):
Nice!

pyk/src/pyk/kast/lexer.py
tothtamas28: I reinstated them.

PR #4337, opened 2024-05-10
toml arg parse typing support

ovatman:
This PR introduces additional functionality for setting argument values' types while parsing toml files:

- `get_argument_type_setter()`: builds a dictionary of valid type setters to be used while setting values read from toml
- `*Options.get_argument_type()`: contains a valid type setter for each parameter. Doesn't specify if toml supports the type out of the box
- Also new tests are added to check related functionality.


ehildenb:
@ovatman you need to rebase your branch on `develop`, and remove all the commits that came in from the `master` branch. This repository uses `develop` as the base branch.

PR #4340, opened 2024-05-13
pyk RPC: bug-report w/o dt dir

jberthold:
The `llvm_definition/dt` directory in the bug report is not necessary (any more), and is usually very big.
Closes https://github.com/runtimeverification/mx-backend/issues/98


ehildenb:
The logs are used to display the list of rules used to make an edge in the KCFG. Why does logging this information make the backend hold onto large amounts of memory? Can that be fixed on the backend side? It should just be a list of hashes.

ehildenb:
Is that memory being held in Kore or in booster?

jberthold:
> Is that memory being held in Kore or in booster?

Both are collecting information during rewriting. In kore, we might still be collecting the intermediate terms that were once part of the log.
We were planning to phase out the log options in the requests but if they are required we need to keep them. I will reduce the amount of information. Will back out this change from the PR then.

tothtamas28 (approved):

tothtamas28:
> Will back out this change from the PR then

Please also adjust the title and description as they are going to be part of the commit message.

Issue #4153, opened 2024-04-04
Wrong `--syntax-module` should cause an error


tothtamas28:
Consider

```k
module TEST
endmodule
```

Then

```
$ kompile test.k
[Warning] Compiler: Could not find main syntax module with name TEST-SYNTAX in
definition.  Use --syntax-module to specify one. Using TEST as default.
```

which is reasonable. On the other hand,

```
$ kompile test.k --syntax-module FOO
[Warning] Compiler: Could not find main syntax module with name FOO in
definition.  Use --syntax-module to specify one. Using TEST as default.
```

But these should be separate cases:
1. Option is missing: warning about fallback to default.
2. Option is present, module not found: error.

PR #4342, opened 2024-05-13
kompilex command for pyk

gtrepta:
Implements the command `pyk kompilex` which uses pyk's outer parser to parse in a K Definition and then hands it off to the java frontend to handle the rest of the compilation.

Some bootstrapping is required because the conversions from the outer parsed structures to something compatible with the java frontend aren't fully implemented. For the integration tests, a `prelude-modules.json` file is included in the test data folder which has the K prelude already outer parsed.

This will only handle modules that are empty or import other modules. The goal is to expand this capability until the full K language is supported, and then the bootstrapping will no longer be necessary.


ehildenb:
We should probably make sure this plays nicely with: https://github.com/runtimeverification/k/pull/4345

pyk/src/pyk/utils.py
```diff
@@ -84,6 +85,14 @@ def __repr__(self) -> str:
 EMPTY_FROZEN_DICT: Final[FrozenDict] = FrozenDict()
 
 
+class FrozenDictEncoder(JSONEncoder):
+    def default(self, obj: Any) -> dict:
+        if isinstance(obj, FrozenDict):
+            return obj._dict
+        else:
+            return JSONEncoder.default(self, obj)
```
tothtamas28: I'd like to avoid this if possible. Where do the `FrozenDict` instances come from?

pyk/src/tests/integration/test_pyk.py
```diff
@@ -186,3 +186,26 @@ def test_rpc_kast(
         expected = json.loads(expected_file.read_text())
         actual = json.loads(actual_file.read_text())
         assert actual == expected
+
+
+class TestKompileX:
+    TEST_FILES = ('a.k', 'b.k', 'c.k', 'd.k')
+    TEST_DATA = [TEST_DATA_DIR / 'k-files' / file for file in TEST_FILES]
+
+    @pytest.mark.parametrize('definition_file', TEST_DATA, ids=[id for id, _ in enumerate(TEST_FILES)])
```
tothtamas28:
```suggestion
    @pytest.mark.parametrize('definition_file', TEST_DATA, ids=[definition_file.name for definition_file in TEST_DATA])
```

pyk/src/tests/integration/test-data/prelude-modules.json
tothtamas28, line 1: Is there a way to produce this file dynamically in a fixture? That would save us a lot of headache later in case of a change in `kompile`.

pyk/src/pyk/kast/att.py
```diff
@@ -309,6 +309,7 @@ class Atts:
     SOURCE: Final = AttKey('org.kframework.attributes.Source', type=_PATH)
     STRICT: Final = AttKey('strict', type=_ANY)
     SYMBOL: Final = AttKey('symbol', type=OptionalType(_STR))
+    SYNTAXMODULE: Final = AttKey('syntaxModule', type=_STR)
```
tothtamas28:
```suggestion
    SYNTAX_MODULE: Final = AttKey('syntaxModule', type=_STR)
```

pyk/src/pyk/__main__.py
```diff
@@ -427,5 +432,50 @@ def exec_parse_outer(options: ParseOuterOptions) -> None:
         sys.stdout.write(f'{result_text}\n')
 
 
+def exec_kompilex(options: KompileXCommandOptions) -> None:
```
tothtamas28, line 426: Are there self-contained logical units in this function that could be extracted into helpers?

pyk/src/pyk/__main__.py
```diff
@@ -427,5 +432,50 @@ def exec_parse_outer(options: ParseOuterOptions) -> None:
...
+    ntf = NamedTemporaryFile('w', prefix='pyk_kompilex_', delete=not options.debug)
+    ntf.write(json.dumps(kast_json, cls=FrozenDictEncoder))
+    ntf.flush()
+
+    options.main_file = ntf.name
+    options.outer_parsed_json = True
+    if options.definition_dir is None:
+        options.definition_dir = Path(f'{definition_file.stem}-kompiled')
+
+    exec_kompile(options)
```
tothtamas28:
```suggestion
    with NamedTemporaryFile('w', prefix='pyk_kompilex_', delete=not options.debug) as ntf:
        ntf.write(json.dumps(kast_json, cls=FrozenDictEncoder))
        ntf.flush()

        options.main_file = ntf.name
        options.outer_parsed_json = True
        if options.definition_dir is None:
            options.definition_dir = Path(f'{definition_file.stem}-kompiled')

        exec_kompile(options)
```

pyk/src/pyk/utils.py
gtrepta: There is a `bracketLabel` attribute in the prelude with a dictionary for its value. It gets loaded in as an `AnyType` which makes a `FrozenDict` for it. Now that you mention it though, it looks like it should return something unfrozen when `to_dict` gets called on it, I'll try looking into it some more.

pyk/src/tests/integration/test-data/prelude-modules.json
gtrepta: I don't think there's a way without adding a feature to `kompile` to enable it. Aside from a possible change Scott is about to make with regexes, I think the prelude is relatively stable and we shouldn't devote much effort here if our intent is to end up parsing it in once we're ready.

pyk/src/pyk/__main__.py
gtrepta: There are similar operations in `exec_parse_outer` for resolving include paths and the main module name. I could see if they could be made into some utilities.

gtrepta: I made a refactor to `KDefinitionOptions` to help reduce some of the duplicated logic. Let me know if it looks good to you.

pyk/src/tests/integration/test-data/prelude-modules.json
tothtamas28: > we shouldn't devote much effort here if our intent is to end up parsing it in once we're ready

Makes sense.

pyk/src/pyk/cli/args.py
```diff
@@ -99,12 +103,24 @@ def default() -> dict[str, Any]:
...
 
+    def __init__(self, args: dict[str, Any]) -> None:
+        if 'includes' in args:
+            include_paths = []
+            for include in args['includes']:
+                try:
+                    include_path = dir_path(include)
+                    include_paths.append(include_path.resolve())
+                except ValueError:
+                    _LOGGER.warning(f"Could not find directory '{include}' passed to -I")
```
tothtamas28:
```suggestion
                if include_path.is_dir():
                    include_paths.append(include_path.resolve())
                else:
                    _LOGGER.warning(f"Could not find directory '{include}' passed to -I")
```

pyk/src/pyk/__main__.py
```diff
@@ -427,5 +427,44 @@ def exec_parse_outer(options: ParseOuterOptions) -> None:
...
+    if options.pre_parsed_prelude:
+        prelude_json = json.loads(options.pre_parsed_prelude.read())
+        prelude_modules = tuple(KFlatModule.from_dict(mod) for mod in prelude_json)
+        final_definition = final_definition.let(all_modules=final_definition.all_modules + prelude_modules)
+
+    syntax_module_name = options.syntax_module or main_module_name + '-SYNTAX'
+    if syntax_module_name not in [m.name for m in final_definition.all_modules]:
+        _LOGGER.warn(
+            f'Could not find main syntax module with name {syntax_module_name} in definition. Use --syntax-module to specify one. Using {main_module_name} as default.'
+        )
```
tothtamas28: If the missing `syntax_module_name` comes from `options`, that should be an error: #4153

pyk/src/pyk/__main__.py
```diff
@@ -427,5 +427,44 @@ def exec_parse_outer(options: ParseOuterOptions) -> None:
...
+def exec_kompilex(options: KompileXCommandOptions) -> None:
+    definition_file = Path(options.main_file).resolve()
+    search_paths = [definition_file.parent, *options.includes]
+    main_module_name = options.main_module or definition_file.stem.upper()
+
+    try:
+        final_definition = parse_outer(definition_file, main_module_name, search_paths, options.md_selector)
+    except Exception as e:
+        _LOGGER.critical(e)
+        exit(1)
```
tothtamas28: I think you can just let the exception propagate to the output instead.

tothtamas28 (approved):

pyk/src/pyk/__main__.py
gtrepta: Sounds fine for now. There are a variety of user errors here that can cause an exception and I would eventually like to see these exceptions be collected and logged somehow for the user instead of having python die and show them a stack trace.

PR #4348, opened 2024-05-14
Reduce verbosity of the subsumption check message in pyk.proof.reachability

### LINKED ISSUE #4344: Reduce verbosity of the subsumption check message in `pyk.proof.reachability`
The INFO level message prints both KAST terms that will be sent to the `"implies"` endpoint at INFO level, which is too verbose and crops up in Kontrol. We should not print the KAST terms at INFO level.

anvacaru:
closes: https://github.com/runtimeverification/k/issues/4344

### DIFF
--- a/pyk/src/pyk/cterm/symbolic.py
+++ b/pyk/src/pyk/cterm/symbolic.py
@@ -159,7 +159,7 @@ class CTermSymbolic:
         return kast_simplified, logs
 
     def get_model(self, cterm: CTerm, module_name: str | None = None) -> Subst | None:
-        _LOGGER.info(f'Getting model: {cterm}')
+        _LOGGER.debug(f'Getting model: {cterm}')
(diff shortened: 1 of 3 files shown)


ehildenb:
We should instead be logging this at the `debug` level, not the `info` level but with the terms removed.

We want to be able to get the terms out if we need to debug.

ehildenb:
I also see another place where we're logging a KAST term at the `info` level: https://github.com/runtimeverification/k/blob/ecee7b169fb82f0e3633819e5d02492459157e8b/pyk/src/pyk/cterm/symbolic.py#L162

It sthould be at the `debug` level instead.

Please go through _all_ the `_LOGGER.info` calls (`git grep _LOGGER.info`) and check if any of them are logging KAST terms. If so, they need to be modified to log at debug level.

geo2a (approved):

PR #4361, opened 2024-05-16
Remove `flatten_label` recursion

JuanCoRo:
The recursion in `pyk/kast/inner.py::flatten_label` might be the cause of recent segfaults when using Kontrol. This PR changes the logic of the function not to make recursive calls to hopefully avoid the segfaults.


pyk/src/tests/unit/kast/test_inner.py
```diff
@@ -0,0 +1,44 @@
...
+    ('f', f(x, f(y, x, f(y)), z), [x, y, x, y, z]),
+    ('f', f(x, f(y, x, f(g(f(y))), z)), [x, y, x, g(f(y)), z]),
+)
+
+
+@pytest.mark.parametrize('label,kast,flist', MATCH_TEST_DATA, ids=count())
+def test_flatten_label(label: str, kast: KInner, flist: list[KInner]) -> None:
+    # When
+    flattened = flatten_label(label, kast)
+    print(flattened)
```
tothtamas28:
```suggestion
```

pyk/src/tests/unit/kast/test_inner.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Then
+    assert flattened is not None
```
tothtamas28:
```suggestion
```

This is so by construction.

pyk/src/tests/unit/kast/test_inner.py
```diff
@@ -0,0 +1,44 @@
...
+    ('f', f(f(f(x))), [x]),
+    ('f', f(g(f(x))), [g(f(x))]),
+    ('f', f(f(x, y, z)), [x, y, z]),
+    ('f', f(x, f(y, x, f(y)), z), [x, y, x, y, z]),
+    ('f', f(x, f(y, x, f(g(f(y))), z)), [x, y, x, g(f(y)), z]),
+)
+
+
+@pytest.mark.parametrize('label,kast,flist', MATCH_TEST_DATA, ids=count())
+def test_flatten_label(label: str, kast: KInner, flist: list[KInner]) -> None:
```
tothtamas28:
```suggestion
@pytest.mark.parametrize('label,kast,expected', MATCH_TEST_DATA, ids=count())
def test_flatten_label(label: str, kast: KInner, expected: list[KInner]) -> None:
```

pyk/src/tests/unit/kast/test_inner.py
```diff
@@ -0,0 +1,44 @@
...
+    # When
+    flattened = flatten_label(label, kast)
```
tothtamas28:
```suggestion
    actual = flatten_label(label, kast)
```

pyk/src/tests/unit/kast/test_inner.py
```diff
@@ -0,0 +1,44 @@
...
+from pyk.kast.inner import flatten_label
+
+from ..utils import a, f, g, x, y, z
+
+if TYPE_CHECKING:
+    from typing import Final
+
+    from pyk.kast import KInner
+
+MATCH_TEST_DATA: Final[tuple[tuple[str, KInner, list[KInner]], ...]] = (
```
tothtamas28:
```suggestion
FLATTEN_LABEL_DATA: Final[tuple[tuple[str, KInner, list[KInner]], ...]] = (
```

pyk/src/pyk/kast/inner.py
```diff
@@ -936,7 +936,13 @@ def flatten_label(label: str, kast: KInner) -> list[KInner]:
...
-    if type(kast) is KApply and kast.label.name == label:
-        items = (flatten_label(label, arg) for arg in kast.args)
-        return [c for cs in items for c in cs]
-    return [kast]
+    flattened_args = []
+    rest_of_args = [kast]  # Rest of arguments in reversed order
+    while rest_of_args:
+        current_arg = rest_of_args.pop()
+        if type(current_arg) is KApply and current_arg.label.name == label:
+            next_args = reversed(list(current_arg.args))
```
tothtamas28: This reversing can be saved by prepending to the list.

JuanCoRo: Would it be worth using something like [deque objects](https://docs.python.org/3/library/collections.html#collections.deque)? My understanding is that if `rest_of_args` grows big this can be inefficient. But I'm fine with whatever solution seems fine to you!

tothtamas28: Never mind my comment. On a closer look, this is the right approach since `reversed` just produces an iterator.

tothtamas28:
```suggestion
            next_args = reversed(current_arg.args)
```

pyk/src/pyk/kast/inner.py
```diff
@@ -936,7 +936,13 @@ def flatten_label(label: str, kast: KInner) -> list[KInner]:
...
     """
-    if type(kast) is KApply and kast.label.name == label:
-        items = (flatten_label(label, arg) for arg in kast.args)
-        return [c for cs in items for c in cs]
-    return [kast]
+    flattened_args = []
+    rest_of_args = [kast]  # Rest of arguments in reversed order
+    while rest_of_args:
+        current_arg = rest_of_args.pop()
+        if type(current_arg) is KApply and current_arg.label.name == label:
```
tothtamas28:
```suggestion
        if isinstance(current_arg, KApply) and current_arg.label.name == label:
```

tothtamas28 (approved):
Looks pretty good!

This looks correct to me and is also unit tested. However, because this is such a fundamental function, please run either the `evm-semantics` or the `kontrol` test suite before merging.

JuanCoRo:
Thank you! Will run both test suites before getting this PR out of draft.

ehildenb:
I would also say that you should run the particular example that was causing the infinite depth issue and make sure that it's working correctly with this change.

JuanCoRo:
I also ran it on all the examples breaking on the `flatten_label` recursion, and all worked fine after this fix

PR #4362, opened 2024-05-16
Check lexical identifiers in regular expressions

### LINKED ISSUE #4356: Check lexical identifiers
For `syntax lexical Name = r"foo"`, any usage of `{Name}` in another regex is semantically equivalent to `(foo)`. There are a number of checks that we should perform to ensure this is valid:

- Check that every `{Name}` referenced in a regex has a corresponding definition `syntax lexical Name = ...`
- Check that there are no cycles like `syntax lexical Foo = r"{Bar}"` `syntax lexical Bar = r"{Foo}"`
- Ban line anchors, lest the expansion produce an invalid regex
- Mangle the names to be Flex compatible, i.e., the name must match the regex `[_a-zA-Z][_\-a-zA-Z0-9]*`
   - Alternatively, we can manually do the expansion ourselves before emitting to Flex

Scott-Guest:
Closes #4356

Adds various checks surrounding `syntax lexical` declarations:
- Ban line anchors, as the expansion of any such identifier always produces an invalid regex with line anchors in non-start/non-end position
- Check that every `{Name}` referenced in a regex has a corresponding definition `syntax lexical Name = ...`
- Check that there are no cycles like `syntax lexical Foo = r"{Bar}"` `syntax lexical Bar = r"{Foo}"`
- Mangle the names to be Flex compatible, i.e., the name must match the regex `[_a-zA-Z][_\-a-zA-Z0-9]*`

There are a few miscellaneous changes mixed in, so I recommend reviewing commit-by-commit.

### DIFF
--- a/k-frontend/src/main/java/org/kframework/parser/outer/ParseRegex.java
+++ b/k-frontend/src/main/java/org/kframework/parser/outer/ParseRegex.java
@@ -312,11 +312,21 @@ public class ParseRegex {
       return result;
     }
     if (input.consume('{')) {
-      StringBuilder identifier = new StringBuilder();
+      StringBuilder identifierBuild = new StringBuilder();
       while (!input.consume('}')) {
-        identifier.append(Character.toString(input.next()));
+        identifierBuild.append(Character.toString(input.next()));
       }
-      return new RegexBody.Named(Outer.parseSort(identifier.toString()).name());
+      String identifier = identifierBuild.toString();
+      if (!identifier.matches("#?[A-Z][a-zA-Z0-9]*")) {
+        input.back();
+        throw parseError(
+            "Lexical identifier {"
+                + identifier
+                + "} is invalid. Identifiers should match the regular expression \"#?[A-Z][a-zA-Z0-9]*\".",
+            input,
+            true);
+      }
+      return new RegexBody.Named(identifier);
     }
     return new RegexBody.Char(parseCharExp(input, false));
   }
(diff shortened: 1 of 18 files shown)


k-frontend/src/main/java/org/kframework/parser/outer/ParseRegex.java
```diff
@@ -124,6 +124,26 @@ public static Regex parse(String regex) {
...
+  }
+
+  public static RegexBody parseNamed(String regex) {
+    return parseNamed(new Scanner(regex));
+  }
+
+  private static RegexBody parseNamed(Scanner input) {
+    Regex reg = parse(input);
+    if (reg.startLine() || reg.endLine()) {
+      String msg = "Named lexical syntax cannot contain line anchors.";
```
gtrepta: Can you add this to the test? It looks like the location info is generally lost here as well, any idea why?

gtrepta (approved):
LGTM. Just one comment.

k-frontend/src/main/java/org/kframework/parser/outer/ParseRegex.java
Scott-Guest: Do you have an example where the location info gets lost? It seems to work correctly on a simple test `syntax lexical StartAnchor = r"^foo"`.

gtrepta: My bad. I added the example to `checkLexicalIdentifiers.k` and ran the target, and misread the diff. The location info was there.

PR #4363, opened 2024-05-17
Eliminate `#Or` patterns in generated rules

### LINKED ISSUE #4355: Desugar `#Or` in rules
A recent change to the MX semantics (required to provide sensible compilation times on the LLVM backend: https://github.com/runtimeverification/mx-semantics/pull/265) introduces `#Or` patterns in a set of rules; these are not supported by the booster, which means we can't execute this code symbolically.

We need to add a feature to the frontend that hoists these symbols to the top of rewrite rules when we see them, then duplicates the rules to remove the `#Or`s.

Baltoli:
This PR adds a desugaring step that runs before code generation, whereby `#Or` patterns on the LHS of rules are expanded out into duplicated top-level rules.

For example, `foo(a #Or b) => c` becomes two distinct rules:
```k
foo(a) => c
foo(b) => c
```

I'm not entirely satisfied with where this pass sits in the compilation pipeline. To push `#Or` out of a rule, we need the rule's rewrite to be at the top level. Rules are only converted to top-level rules when we generate KORE; this step is not performed as part of the regular compilation pipeline. As a result, to demonstrate this code working at all, I've had to put the pass into `ModuleToKore`.

This is not ideal; we should not be adding new code to `ModuleToKore`, and _definitely_ shouldn't be adding new code that transforms a definition like this.

My proposal would be that at the end of the regular compilation pipeline, we add a pass that simply lifts rules to be top-level. This should not be a breaking change. Then, this pass can be placed _after_ that pass, such that `ModuleToKore` can have top-level rewrites as an invariant rather than an explicit action taken as part of its behaviour. Thoughts welcome on this restructuring; it will affect the formatting of `compiled.txt` but nothing else as far as I can think.

Fixes https://github.com/runtimeverification/k/issues/4355


bbyalcinkaya:
How does this handle rule labels?  Specifically, if I want to set this rule as a cut point rule, can I use the foo-or-bar label?

```
rule [foo-or-bar]:
   </k> foo #Or bar => .K </k>
```

Baltoli:
Good question. Currently, label uniqueness is enforced at the K level: you can't write the following program:
```k
rule [a]: a => b
rule [a]: b => c
```
which suggests to me that downstream consumers require these labels to be unique when they reach KORE. As currently implemented, we'll get duplicated labels, which may therefore break downstream code.

We should discuss this in the K meeting. Some options we have:
- Just keep the duplicated KORE labels and fix whatever downstream code depends on them being unique (for example, the LLVM backend does some rule-label based lookup in the debugger and when mapping between ordinals and source code - the Haskell backend / Pyk presumably does the same for cut rules, etc.)
- Modify the labels for the generated rules. This is unintuitive and could be tricky for users to figure out, but perhaps by modifying downstream code to use something like suffixed labels + regexes for lookup, we could recover the behaviour we want.
- Disallow rule labels in combination with `#Or` entirely. This is the degenerate fix and I think it's probably quite a bad idea.

Baltoli:
We should figure out exactly how the downstream tools are using the label pseudo-attribute, and see if we can adapt them to cope with having duplicates.

Baltoli:
Closing as we no longer seem to have a pressing need for this.

PR #4365, opened 2024-05-18
Support Unicode-aware regex

### LINKED ISSUE #4357: Handle Unicode in regular expressions
With #4295, our regular expression syntax permits Unicode characters. However, Flex is an 8-bit scanner, so some translation is necessary to actually have Unicode-aware semantics.

Specifically, each Unicode character is considered as its sequence of UTF-8 bytes, so we must translate as follows:
- Outside of a character class, just parenthesize to keep the bytes grouped
   - `r"😊*"` becomes `r"(\xF0\x9F\x98\x8A)*"`
- Inside a non-negated character class, factor any single character out into an explicit `|`
    - `r"[a😊b]"` becomes `r"(\xF0\x9F\x98\x8A)|[ab]"`
 
The other cases are possible to support, but fairly involved, so we can just report an error for now:
- For character class ranges involving Unicode, there's an algorithm described [here](https://lists.gnu.org/archive/html/help-flex/2005-01/msg00043.html).
- For negated character class, we can make use of the `dk.brics.automaton` support for regex complements
  - `r"[^a😊b]"` becomes the `dk.brics.automaton` regex `~((\xF0\x9F\x98\x8A)|[ab])`
  - `dk.brics.automaton` converts this to a DFA
  - We could implement a standard DFA -> regex algorithm which would produce a regex not involving complements

Scott-Guest:
Closes #4357

Convert K Unicode-based regex to an equivalent Flex byte-based regex:
- Outside of a character class, just parenthesize to keep the bytes grouped
   - `r"😊*"` becomes `r"(\xF0\x9F\x98\x8A)*"`
- Inside a non-negated character class, factor any single Unicode character out into an explicit `|`
    - `r"[a😊b]"` becomes `r"(\xF0\x9F\x98\x8A)|[ab]"`
- In all other cases (character ranges and negated character classes), report an error if there are non-ASCII characters

Additionally,
- Check that character ranges `[c1-c2]` have `codepoint(c1) <= codepoint(c2)`
- Check that numeric ranges `r{n,m}` have `n <= m`

The commit history is incremental, and I'd recommend reviewing commit-by-commit.


k-frontend/src/main/java/org/kframework/compile/checks/CheckRegex.java
```diff
@@ -124,4 +129,79 @@ private static void checkLineAnchors(Set<KEMException> errors, Sentence s) {
...
+    }
+  }
+
+  private static class CollectBadUnicodeRegex extends RegexVisitor {
+    private final Set<Integer> badNegated;
+    private final Set<Integer> badRange;
+
+    public CollectBadUnicodeRegex() {
+      this.badNegated = new LinkedHashSet<>();
+      this.badRange = new LinkedHashSet<>();
```
Baltoli, line 174: There's a great line in the documentation for `LinkedHashSet`:
> This implementation spares its clients from the unspecified, generally chaotic ordering provided by [HashSet](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html)

Baltoli (approved):
Very nice; seems like a really well thought through implementation!

PR #4370, opened 2024-05-21
Use LLVM 17 by default in flake

Baltoli:
@ehildenb bumped into https://github.com/llvm/llvm-project/issues/60972 again when [updating KEVM tests](https://github.com/runtimeverification/evm-semantics/pull/); at the time I fixed the bug originally it wasn't yet in a released version of LLVM we could use, and so we had to just ignore the issue.

Now that LLVM 17 is out, we should update K to use it by default where possible; this PR does so for the Nix flake build of K.


ehildenb:
Do we need to update the `flake.lock` file too`? `nix build lock` or something?

Baltoli:
Nope - the different LLVM versions are attributes of `nixpkgs` rather than lockable flake inputs in their own right. The `nixpkgs` we use already has LLVM 17 available, so we don't need to update anything there.

ehildenb (approved):

PR #4371, opened 2024-05-21
CI: Don't tell pytest to try to kill the integration tests on failure

gtrepta:
There's a bug with `pytest-xdist` where using multiple runners doesn't properly terminate the test run when there's a failure and `--maxfail=1` has been passed. Instead, the runner that ran the failed test exits and the remaining runners continue. What this results in on CI is the remaining runners are not able to finish the integration tests before the timeout for the job, killing pytest before it can finish and report the reason for the failed test.

This change removes the `--maxfail` option when running on CI, so when there's a test failure, all of the runners continue, so they should be able to finish before the timeout and pytest can properly finish and report the failure reason(s).

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -270,7 +270,7 @@ jobs:
           k-deb-path: kframework.deb
       - name: 'Run integration tests'
         run: |
-          docker exec -u user k-pyk-integration-${{ github.sha }} make test-integration TEST_ARGS='-n4 --timeout 300 -vv'
+          docker exec -u user k-pyk-integration-${{ github.sha }} make test-integration TEST_ARGS='-n4 --timeout 300'
       - name: 'Tear down Docker'
         if: always()
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test-pr.yml
```diff
@@ -270,7 +270,7 @@ jobs:
           k-deb-path: kframework.deb
       - name: 'Run integration tests'
         run: |
-          docker exec -u user k-pyk-integration-${{ github.sha }} make test-integration TEST_ARGS='-n4 --timeout 300 -vv'
```
Baltoli, line 273: Do we want to keep `-vv`?

gtrepta: I don't think so. I added that last week when trying to help out Georgy get more info from the CI failure (he was having trouble reproducing it locally). It just added a ton of not super useful timing information, see [this](https://github.com/runtimeverification/k/actions/runs/9178802879/job/25240227479#step:6:1935)

Baltoli: Makes sense!

Baltoli (approved):
Nice - I like that we keep fail-fast for local runs

PR #4373, opened 2024-05-21
Run homebrew release on macOS 14 / arm64

Baltoli:
We are working on dropping support for intel macOS across K-related projects; one of the last load-bearing things we relied on that platform for was running Homebrew builds and tests. As of this year, Github offer hosted macOS 14 / arm64 runners that solve the problem we ran into previously where our own hosted runners were not ephemeral enough to properly run these builds.

This PR moves the relevant parts of our release workflows over to the new hosted runners; this code isn't directly tested in the PR flow so I opened a dummy PR that runs the relevant parts of the workflow. You can see the tests passing here: https://github.com/runtimeverification/k/pull/4372

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -143,7 +143,7 @@ jobs:
 
   macos-build:
     name: 'Build MacOS Package'
-    runs-on: macos-13
+    runs-on: macos-14
     timeout-minutes: 120
     environment: production
     needs: [set-release-id, source-tarball]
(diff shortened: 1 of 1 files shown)


Baltoli:
Note that lots of matrices still have `runs-on: macos-1{2,3}` - this is by design. We will go through and _remove_ those stanzas in later follow-up changes; the Homebrew stuff here is a prerequisite to doing so.

F-WRunTime (approved):

PR #4374, opened 2024-05-22
Revert "Use LLVM 17 by default in flake"

Baltoli:
Temporary fix while Cachix is broken.

Reverts runtimeverification/k#4370


Baltoli:
Fixed by https://github.com/runtimeverification/kup/pull/113

PR #4376, opened 2024-05-22
Add custom_step heuristics for KCFGSemantics

anvacaru:
Introducing a new `custom_step` heuristic to the `KCFGSemantics` class that allows a semantics to specify a custom transformation to be applied to a state before its execution under a given condition.
If the transformation is applied, the function must return a `KCFGExtendResult` that reflects the modifications
If no transformation is applied, the function returns `None` and the term is executed.

 You can see an example of this implementation [here](https://github.com/runtimeverification/evm-semantics/pull/2441/files#diff-9f446fabb94d32f7c166a34331647fd15ac4e539283a9cec6b9b29cf1bf4a9deR115).


PetarMax:
From what I'm understanding, this allows us to intercept any execution step before it's passed to the backend, and the way we plan to use it is to introduce cut-point rules in the associated semantics.

Could we add some sort of test?

ehildenb:
Maybe a test could be:

- Run a small proof that exercises this step, but with this `custom_step` turned off (so have an option called `precompute_jumpdests` which is used to control this behavior of `custom_step`), and this rule added as a `cut_point_rule`.
- Run that same proof, but with the the `custom_step` turned on, and this rule not added as a `cut_point_rule`.
- Compare the two KCFGs to make sure they are the same, to test that the custom step is correctly behaving.

pyk/src/tests/integration/proof/test_mini_kevm.py
```diff
@@ -106,3 +164,80 @@ def test_all_path_reachability_prove(
...
+                execute_depth=max_depth,
+                cut_point_rules=cut_rules,
+            )
+            prover2.advance_proof(proof2, max_iterations=max_iterations)
+
+            kcfg_show = KCFGShow(
+                kprove, node_printer=APRProofNodePrinter(proof2, kprove, full_printer=True, minimize=False)
+            )
+            cfg_lines1 = kcfg_show.show(proof1.kcfg)
+            _LOGGER.info('\n'.join(cfg_lines1))
```
ehildenb: I don't understand how we're controlling whether the custom step fires or not. I guess we want to always have the rule as a cut-rule, but in one test use the custom step, and in the other one don't, but make sure we get the same KCFG either way?

anvacaru: Yes, I wasn't quite sure how to handle this on Friday and I had some errors that were preventing me from running the test locally. 

Now the rule is always marked as a `cut-rule` and I added a `globalvar skip_custom_test` flag in the `custom_step` that is `False` for the first proof and `True` for the second one, then I compare the kcfgs.

pyk/src/tests/integration/proof/test_mini_kevm.py
```diff
@@ -106,3 +168,87 @@ def test_all_path_reachability_prove(
...
+        max_iterations: int | None,
+        max_depth: int | None,
+        cut_rules: Iterable[str],
+        proof_status: ProofStatus,
+        expected_leaf_number: int,
+        tmp_path_factory: TempPathFactory,
+    ) -> None:
+        global skip_custom_step
+
+        skip_custom_step = False
```
tothtamas28: This looks problematic. What's the purpose of this variable?

pyk/src/tests/integration/proof/test_mini_kevm.py
```diff
@@ -106,3 +168,87 @@ def test_all_path_reachability_prove(
 
             assert proof.status == proof_status
             assert leaf_number(proof) == expected_leaf_number
+
+    @pytest.mark.parametrize(
+        'test_id,spec_file,spec_module,claim_id,max_iterations,max_depth,cut_rules,proof_status,expected_leaf_number',
+        CUSTOM_STEP_TEST_DATA,
+        ids=[test_id for test_id, *_ in CUSTOM_STEP_TEST_DATA],
+    )
+    def test_custom_step(
```
tothtamas28: I think you could use something much simpler than `MINI-KEVM` for testing this. Something like

```k
module CUSTOM-STEP
    sort State ::= "a" | "b" | "c" [token]
    rule <k> a => b </k>
endmodule
```

Then add a custom step to `c`.

pyk/src/tests/integration/proof/test_mini_kevm.py
anvacaru: The idea of the test is to run a proof twice, once with a custom_step, and once with the rule from the K semantics, and assert that the kcfgs are the same. The global var is used to skip the custom step for the second execution of the proof.

tothtamas28: But then there's a dependency between test runs. Also, this only works if you run the two tests on the same runner.

Why not just write a very simple K definition, and add two tests: one with a custom step, one without, and assert the end result for both.

anvacaru: I don't see the dependency. Both proofs are executed in the same `test_custom_proof` defined below. The alternative I see would be to define two implementations of a KCFGSemantics: one with a custom_step implementation, and one without and to use both in the same test to execute the same claim and assert the results.

tothtamas28: > I don't see the dependency

IIUC one test run sets the global variable and another one reads it.

> The alternative I see would be to define two implementations of a KCFGSemantics: one with a custom_step implementation, and one without and to use both in the same test to execute the same claim and assert the results.

That's what I was thinking. You can make this a parameterized test where the semantics is an input parameter.

anvacaru: > IIUC one test run sets the global variable and another one reads it.

No, the `skip_custom_step` is used only in this test, `test_custom_step`. There are two proofs `proof1` and `proof2` of the same claim, with two init_cterms and two provers, one for each proof. I'm running the first proof with `skip_custom_step` set to `False`, to execute the custom_step, then I'm setting the `skip_custom_step` to `True` and I execute the second proof.

tothtamas28: I see, so it's a way to configure the behavior of a single `KCFGSemantics` instance. Still, the need for a global variable indicates the existing set of fixtures is not a good fit for the test as currently implemented.

Solutions I can think of:
- Implement a custom fixture that lets you create `KCFGExplore` instances with different `KCFGSemantics` instances plugged in during test execution.
- The one we already discussed: https://github.com/runtimeverification/k/pull/4376#discussion_r1616222653

Also, please write a simple, dedicated semantics for this test.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+            (),
+            ['CUSTOM-STEP.c.d'],
+            cut=True,
+        ),
+    ),
+)
+
+
+class TestCustomStep(KCFGExploreTest, KProveTest):
+    KOMPILE_MAIN_FILE = K_FILES / 'custom-step.k'
```
tothtamas28: The test definition is small enough to inline it using `KOMPILE_DEFINITION`.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+            1,
+            (),
+            ['CUSTOM-STEP.c.d'],
+            cut=True,
+        ),
+    ),
+)
+
+
+class TestCustomStep(KCFGExploreTest, KProveTest):
```
tothtamas28: Consider making this a `CTermSymbolicTest`, and moving the `create_kcfg_explore` fixture here. We can expose the fixture in the `pyk.testing` library once there are more tests requiring it.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+    KOMPILE_MAIN_FILE = K_FILES / 'custom-step.k'
+
+    # Disabled until resolved: https://github.com/runtimeverification/haskell-backend/issues/3761
+    DISABLE_LEGACY = True
```
tothtamas28: Does this test not work with the legacy backend?

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @pytest.mark.parametrize(
+        'test_id,c,e',
+        CUSTOM_STEP_TEST_DATA_APPLY,
+        ids=[test_id for test_id, *_ in CUSTOM_STEP_TEST_DATA_APPLY],
+    )
+    def test_custom_step_exec(self, test_id: str, c: CTerm, e: KCFGExtendResult | None) -> None:
```
tothtamas28:
```suggestion
    def test_custom_step_exec(self, test_id: str, cterm: CTerm, expected: KCFGExtendResult | None) -> None:
```

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # When
+        kcfg_semantics = CustomStepSemanticsWithStep()
+        result = kcfg_semantics.custom_step(c)
```
tothtamas28:
```suggestion
        actual = kcfg_semantics.custom_step(c)
```

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,258 @@
...
+        spec_module: str,
+        claim_id: str,
+        max_iterations: int | None,
+        max_depth: int | None,
+        cut_rules: Iterable[str],
+        proof_status: ProofStatus,
+        tmp_path_factory: TempPathFactory,
+    ) -> None:
+
+        with tmp_path_factory.mktemp('custom_step_tmp_proofs') as proof_dir:
```
tothtamas28: This fixture can be used directly:

```python3
proof_dir = tmp_path_factory.mktemp('custom_step_tmp_proofs')
```

pyk/src/pyk/testing/_kompiler.py
```diff
@@ -291,6 +291,17 @@ def cterm_symbolic(
     ) -> CTermSymbolic:
         return CTermSymbolic(kore_client, definition, kompiled_kore)
 
+    @pytest.fixture
+    def create_kcfg_explore(
```
tothtamas28: Please make this fixture local to the test.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,266 @@
...
+    ) -> None:
+
+        proof_dir = tmp_path_factory.mktemp('custom_step_tmp_proofs')
+        claim = single(
+            kprove.get_claims(Path(spec_file), spec_module_name=spec_module, claim_labels=[f'{spec_module}.{claim_id}'])
+        )
+        kcfg_explore_without_step = create_kcfg_explore(CustomStepSemanticsWithoutStep())
+        kcfg_explore_with_step = create_kcfg_explore(CustomStepSemanticsWithStep())
+
+        proof_a = APRProof.from_claim(
```
tothtamas28: Are the two proofs stored at different locations? Consider creating a dedicated directory for each.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,266 @@
...
+        spec_module: str,
+        claim_id: str,
+        max_iterations: int | None,
+        max_depth: int | None,
+        cut_rules: Iterable[str],
+        proof_status: ProofStatus,
+        tmp_path_factory: TempPathFactory,
+    ) -> None:
+
+        proof_dir = tmp_path_factory.mktemp('custom_step_tmp_proofs')
```
tothtamas28: You can use `tmp_path`, the scope of the directory is only for this test.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,266 @@
...
+            proof_dir=proof_dir,
+        )
+        proof_b = APRProof.from_claim(
+            kprove.definition,
+            claim,
+            subproof_ids=[],
+            logs={},
+            proof_dir=proof_dir,
+        )
+        init_cterm_a = kcfg_explore_without_step.cterm_symbolic.assume_defined(proof_a.kcfg.node(proof_a.init).cterm)
```
tothtamas28: In the test, each step is repeated for the two proofs, please extract a private static method to reduce duplication.

pyk/src/tests/integration/proof/test_custom_step.py
anvacaru: It's the same claim executed twice, I created different directories using `tmp_path`.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,232 @@
...
+                KApply(
+                    KLabel('<k>'),
+                    [
+                        KSequence(KApply(KLabel(label))),
+                    ],
+                ),
+                KApply(KLabel('<generatedCounter>'), [KVariable('GENERATEDCOUNTER_CELL', KSort(name='Int'))]),
+            ],
+        ),
+    )
```
tothtamas28: You can make this a bit more succinct:

```suggestion
    return CTerm(
        KApply(
            '<generatedTop>',
            KApply('<k>', KSequence(KApply(label))),
            KApply('<generatedCounter>', KVariable('GENERATEDCOUNTER_CELL', 'Int')),
        ),
    )
```

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,232 @@
...
+            return KCFGExplore(cterm_symbolic, kcfg_semantics=kcfg_semantics)
+
+        return _create_kcfg_explore
+
+    @pytest.mark.parametrize(
+        'test_id,cterm,expected',
+        CUSTOM_STEP_TEST_DATA_APPLY,
+        ids=[test_id for test_id, *_ in CUSTOM_STEP_TEST_DATA_APPLY],
+    )
+    def test_custom_step_exec(self, test_id: str, cterm: CTerm, expected: KCFGExtendResult | None) -> None:
```
tothtamas28, line 149: This test is a bit superfluous this way I think. What would make sense is to test `custom_step` in integration with `KCFGExplore`, i.e. extend a `KCFG` a single step and check the result.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,232 @@
...
+        kcfg_explore.simplify(proof.kcfg, {})
+        prover = APRProver(
+            kcfg_explore=kcfg_explore,
+            execute_depth=max_depth,
+            cut_point_rules=cut_rules,
+        )
+        prover.advance_proof(proof, max_iterations=max_iterations)
+
+        kcfg_show = KCFGShow(kprove, node_printer=APRProofNodePrinter(proof, kprove, full_printer=True, minimize=False))
+        assert proof.status == proof_status
```
tothtamas28: I'd just return the proof status and check this in the call site.

pyk/src/tests/integration/proof/test_custom_step.py
```diff
@@ -0,0 +1,232 @@
...
+    def test_custom_step_exec(self, test_id: str, cterm: CTerm, expected: KCFGExtendResult | None) -> None:
+
+        # When
+        kcfg_semantics = CustomStepSemanticsWithStep()
+        actual = kcfg_semantics.custom_step(cterm)
+        # Then
+        assert expected == actual
+
+    @staticmethod
+    def _run_proof(
```
tothtamas28: Consider moving this after `test_cfg_equality`.

tothtamas28 (approved):

pyk/src/tests/integration/proof/test_custom_step.py
anvacaru: @tothtamas28  I'd like to merge this as it is for now to unblock https://github.com/runtimeverification/evm-semantics/pull/2441. Also, I'd say that the integration with `KCFGExplore` is tested using the `_run_prover` method below. I can open a follow-up PR to change this test if needed.

PR #4385, opened 2024-05-23
emit cell attribute into kore definition

dwightguth:
This PR adds the `cell` attribute back to the KORE definition in order to allow downstream usage in Pi Squared.


Baltoli:
Looks like this affects some Pyk tests that reference the attribute subsystem; let me know if you want a hand from @tothtamas28 or me to sort those out.

tothtamas28:
> Looks like this affects some Pyk tests that reference the attribute subsystem; let me know if you want a hand from @tothtamas28 or me to sort those out.

Here I think: https://github.com/runtimeverification/k/blob/ff056270cc62c50284c1865f5744b499f8d9a36b/pyk/src/pyk/konvert/_module_to_kore.py#L726

tothtamas28:
69fcd40: this seems to be another instance of runtimeverification/haskell-backend#3761.

Scott-Guest (approved):

PR #4387, opened 2024-05-24
Build and start to deploy pyk documentation

Baltoli:
This PR adapts the old Pyk [Github Pages workflow](https://github.com/runtimeverification/pyk/blob/68bffc5ed0994d2c462d91fa3415f170cb0ec91f/.github/workflows/master.yml) to build the Pyk API documentation and stash it at '/pyk' under the current web root.

I've tested this locally in a Docker container as far as is possible, but I want to make sure that a release goes through correctly before I follow this up by actually editing the site to point to the documentation with a link.


tothtamas28 (approved):
Nice!

PR #4394, opened 2024-05-27
Bypass Jekyll on GitHub Pages

tothtamas28:
Related:
* #4393 

Jekyll is not used for https://kframework.org/, this PR disables the default build step on deployment to GitHub Pages.

PR #4395, opened 2024-05-27
Create `.nojekyll` dynamically

tothtamas28:
Patches:
* #4394 

It is easier to create the file than to copy it.


.github/workflows/release.yml
```diff
@@ -471,6 +471,7 @@ jobs:
           rm -rf $(find . -maxdepth 1 -not -name public_content -a -not -name .git -a -not -path . -a -not -path .. -a -not -name CNAME)
           mv public_content/* ./
```
tothtamas28, line 472: The wildcard pattern does not match the file, so #4394 does not actually disable Jekyll.

anvacaru (approved):

PR #4396, opened 2024-05-27
Including `compute_ordinals` and `get_axiom_by_ordinal` in Pyk `kore.Definition`

Robertorosmaninho:
In this PR, we're translating the logic of these functions from the LLVM Backend `ast.Definition` to Pyk `kore.Definition`. This will allow us to perform the same kind of `rule_ordinal` number-agnostic testing for both kinds of `Definition` classes on the Math Proof Team's test suite!

### DIFF
--- a/pyk/src/pyk/kore/syntax.py
+++ b/pyk/src/pyk/kore/syntax.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from functools import cached_property
 from io import StringIO
 from typing import ClassVar  # noqa: TC003
-from typing import TYPE_CHECKING, final
+from typing import TYPE_CHECKING, List, final
 
 from ..dequote import enquoted
 from ..utils import check_type
@@ -2170,6 +2170,7 @@ class Module(Kore, WithAttrs, Iterable[Sentence]):
 class Definition(Kore, WithAttrs, Iterable[Module]):
     modules: tuple[Module, ...]
     attrs: tuple[App, ...]
+    ordinals: dict[int, Axiom]
 
     def __init__(self, modules: Iterable[Module] = (), attrs: Iterable[App] = ()):
         object.__setattr__(self, 'modules', tuple(modules))
@@ -2193,6 +2194,28 @@ class Definition(Kore, WithAttrs, Iterable[Module]):
         for module in self.modules:
             output.write('\n\n')
             module.write(output)
+
+    def get_axiom_by_ordinal(self, ordinal: int) -> Axiom:
+        return self.ordinals[ordinal]
+
+    def compute_ordinals(self) -> Definition:
+        new_modules = []
+        rule_ordinal = 0
+        for module in self.modules:
+            new_sentences: List[Sentence] = []
+            for sentence in module.sentences:
+                if type(sentence) is Axiom:
+                    ordinal_attr = App('ordinal', (), [String(str(rule_ordinal))])
+                    new_sentence = sentence.let_attrs(sentence.attrs + (ordinal_attr,))
+                    new_sentences.append(new_sentence)
+                    self.ordinals[rule_ordinal] = new_sentence
+                    rule_ordinal += 1
+                else:
+                    new_sentences.append(sentence)
+            new_modules.append(module.let(sentences=new_sentences))
+
+        new_definition = self.let(modules=new_modules)
+        return new_definition
 
 
 def kore_term(dct: Mapping[str, Any]) -> Pattern:
--- a/pyk/src/pyk/kore/syntax.py
+++ b/pyk/src/pyk/kore/syntax.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from functools import cached_property
 from io import StringIO
 from typing import ClassVar  # noqa: TC003
-from typing import TYPE_CHECKING, List, final
+from typing import TYPE_CHECKING, final
 
 from ..dequote import enquoted
 from ..utils import check_type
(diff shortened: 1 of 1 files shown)


pyk/src/pyk/kore/syntax.py
```diff
@@ -2170,6 +2170,7 @@ def axioms(self) -> tuple[Axiom, ...]:
 class Definition(Kore, WithAttrs, Iterable[Module]):
     modules: tuple[Module, ...]
     attrs: tuple[App, ...]
+    ordinals: dict[int, Axiom]
```
tothtamas28: This makes the class no longer immutable and hashable. Instead, consider making this a `cached_property`:

```python3
@cached_property
def ordinals(self) -> FrozenDict[int, Axiom]:
    ...  # calculate ordinals
```

Then `get_axiom_by_ordinal` becomes unnecessary.

tothtamas28 (changes requested):

pyk/src/pyk/kore/syntax.py
Robertorosmaninho: Hey Tamás, I'm not sure if you want to replace `get_axiom_by_ordinal` or `compute_ordinals` here. 
One is computing and modifying the definition,  other is just returning the element in the cached dict. Can you elaborate your suggestion?  

The `compute_ordinals` is one step on the `preprocess` LLVM function for context, which is known to modify the Definition content. To keep the object immutable, I can, instead of using `self.ordinals`, just populate a dict `ordinals` and then set it in `new_definition`.

tothtamas28: `ordinals: dict[int, Axiom]` should not be an attribute of the class in this form as it is
* mutable
* not hashable

Where does this value come from btw?

Robertorosmaninho: Got this part! Is `FrozenDict` better in this sense, right?

This value is computed in `compute_ordinals` by counting the Axioms. We not only have this same [logic in the LLVM backend definition](https://github.com/runtimeverification/llvm-backend/blob/0b0bdaeaf5acfed9034c40bec68b285dc5fac4dc/lib/ast/definition.cpp#L183-L198) but we also have tools to compute and use it from the `definition.kore`, see: [here](https://github.com/runtimeverification/llvm-backend/tree/master/tools/llvm-kompile-compute-ordinal) and [here](https://github.com/runtimeverification/llvm-backend/tree/master/tools/llvm-kompile-compute-loc).

tothtamas28: And the ordinal of the `n`-th `Axiom` is simply `n`, right? Because in that case, storing the `dict` isn't necessary:

```python3
@cached_property
def axioms(self) -> tuple[Axiom, ...]:
    return tuple(sent for module in self for sent in module if isinstance(sent, Axiom))

def get_axiom_by_ordinal(self, ordinal: int) -> Axiom:
    return self.axioms[ordinal]
```

Robertorosmaninho: Perfect. The problem here is that I really need a new definition containing the ordinal attribute in its axioms to work with. This was previously only an LLVM Backend attribute, but we have to have it here as well, as we're using PyK to parse the LLVM Proof Hints. Eventually, we'll upstream this attribute to the future Python `ModuleToKore` class, so this last party should be temporary.

tothtamas28: Constructing a new definition is not a problem. And since it preserves the order of axioms, you can compute the `axiom` list from either of the definitions.

Robertorosmaninho: Right, I'm just saying that using this approach, we'll pass through all the sentences of all modules twice when needed: one for computing the `axioms` and another for creating the new definition.

Robertorosmaninho: If you're okay with that, that's fine with me; I understand this modification is softer than adding a new field to `Definition`. I just want to highlight this rework when using these functions together.

tothtamas28: Is iterating over sentences twice a performance bottleneck? If so, you can return the `dict` along with the `Definition` from `compute_ordinals`. You can introduce a custom class to encapsulate the two values.

If there's no known performance issue then you can just iterate twice and go with the simpler design.

pyk/src/pyk/kore/syntax.py
```diff
@@ -2194,6 +2194,31 @@ def write(self, output: IO[str]) -> None:
             output.write('\n\n')
             module.write(output)
 
+    @cached_property
+    def axioms(self) -> tuple[Axiom, ...]:
+        return tuple(sent for module in self.modules for sent in module if isinstance(sent, Axiom))
+
+    def get_axiom_by_ordinal(self, ordinal: int) -> Axiom:
```
tothtamas28, line 2201: I guess you're going to call this a lot, consider making the name shorter:

```python3
def get_axiom(self, ordinal: int) -> Axiom:
```

or

```python3
def get_axiom(self, *, ordinal: int) -> Axiom:
```

if you want to include `ordinal` in the call for unambiguity.

tothtamas28 (approved):

pyk/src/pyk/kore/syntax.py
Robertorosmaninho: Thanks for the advice! Indeed, it looks better! However, if you don't mind, I would like to keep the same name we use in the LLVM Backend so that others can easily recognize it!

PR #4404, opened 2024-05-30
remove default values from argparse

ovatman:
Since we are handling default values for command line options in the specific options classes a default value shouldn't be assigned during cli argument parsing when option is absent. This PR removes default values from cli argument parser.


ovatman:
A final note: 
Even though argparse's documentation says the `default` parameter itself defaults to `None` (see [here](https://docs.python.org/3/library/argparse.html#default)), it defaults to `False` for `bool` parameters, causing the last commit here.

anvacaru (approved):

PR #4405, opened 2024-05-30
add --proof-hint flag to krun

dwightguth:
We add a new command line flag to the krun script which will tell krun to suppress the usual output and provide the user with a proof hint.


Robertorosmaninho (approved):
I would like to see a small test here, but I’m already approving it as it is a small thing to add.

Robertorosmaninho:
The makefile's test looks much cleaner now, thanks! Feel free to merge once all tests pass!

PR #4410, opened 2024-06-03
Fix `KLabel` for `#Exists`

tothtamas28:
Preparatory refactoring for:
* #4418

In KAST, ML quantifiers are defined with two sort parameters: https://github.com/runtimeverification/k/blob/9100689e357a7e5820362869a73684468ef9c91b/k-distribution/include/kframework/builtin/kast.md?plain=1#L154-L155

However, `pyk` falsely assumed these symbols only have a single sort parameter as in KORE: https://github.com/runtimeverification/haskell-backend/blob/master/docs/kore-syntax.md?plain=1#L194-L197

This PR fixes that.

### DIFF
--- a/pyk/src/pyk/cterm/symbolic.py
+++ b/pyk/src/pyk/cterm/symbolic.py
@@ -24,7 +24,7 @@ from ..kore.rpc import (
     UnsatResult,
     kore_server,
 )
-from ..prelude.k import GENERATED_TOP_CELL
+from ..prelude.k import GENERATED_TOP_CELL, K_ITEM
 from ..prelude.ml import is_top, mlAnd, mlEquals
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


pyk/src/pyk/cterm/symbolic.py
```diff
@@ -233,7 +233,7 @@ def implies(
                 bind_text, bind_label = ('universally', '#Forall')
             _LOGGER.debug(f'Binding variables in consequent {bind_text}: {unbound_consequent}')
             for uc in unbound_consequent:
-                _consequent = KApply(KLabel(bind_label, [GENERATED_TOP_CELL]), [KVariable(uc), _consequent])
+                _consequent = KApply(KLabel(bind_label, [K_ITEM, GENERATED_TOP_CELL]), [KVariable(uc), _consequent])
```
tothtamas28, line 238: This is awkward, but otherwise we would have to infer the type of each `uc`. Fortunately, due to the discrepancy in KAST and KORE, the first sort argument of `#Exists` is not considered for the transformation.

Baltoli: Perhaps add an inline comment as well here?

Baltoli (approved):

pyk/src/pyk/cterm/symbolic.py
tothtamas28: Done.

PR #4411, opened 2024-06-03
Claim dependencies included as subproofs

ehildenb:
This PR fixes some small issues in claim dependency process.

- Claim dependencies were not included as subproofs, which meant they weren't being re-used appropriately when needed, in `APRProof.from_claim(...)`. Fixing this enables using `APRProof.from_claim(...)` directly downstream instead of manually creating the KCFG and proof objects. Note that we have to do some translation of the claim dependencies because by default they would come in with unqualified names, which is fixed by adjusting the claim attributes directly.
- Common predicates between multiple branches were showing up in side-conditions on branching results of exploration, and now they are factored out and reported separately.
- Some adjustments to `KCFG.to_rule(s)` and `ARPProof.as_rule(s)` which allow better naming/labelling of the generated rules, so it's easier to tell where they came from.


anvacaru (approved):
LGTM!

pyk/src/pyk/kcfg/explore.py
```diff
@@ -262,7 +262,22 @@ def extract_rule_labels(_logs: tuple[LogEntry, ...]) -> list[str]:
...
+            branch_preds = [flatten_label('#And', not_none(rule_predicate)) for _, rule_predicate in next_states]
+            common_preds = list(
+                unique(
+                    [
+                        pred
+                        for branch_pred in branch_preds
+                        for pred in branch_pred
+                        if all(pred in bp for bp in branch_preds)
+                    ]
+                )
```
tothtamas28:
```suggestion
                unique(
                    pred
                    for branch_pred in branch_preds
                    for pred in branch_pred
                    if all(pred in bp for bp in branch_preds)
                )
```

pyk/src/pyk/kcfg/explore.py
```diff
@@ -262,7 +262,22 @@ def extract_rule_labels(_logs: tuple[LogEntry, ...]) -> list[str]:
...
+            )
+            branches = [mlAnd(pred for pred in branch_pred if pred not in common_preds) for branch_pred in branch_preds]
+            if len(common_preds) > 0:
```
tothtamas28: Nitpick:

```suggestion
            if common_preds:
```

(https://github.com/python/peps/blob/main/peps/pep-0008.rst?plain=1#L1470-L1485)

tothtamas28 (approved):

PR #4416, opened 2024-06-04
Publish `pyk` on PyPI

### LINKED ISSUE #4239: Publish `pyk` package
Since the K repository has a lot of submodules, installing `pyk` from a `git` dependency using Poetry is very slow.

We should push a distribution package to PyPI. Alternatively, we can publish the source archive as a GitHub release.

F-WRunTime:
Fixes #4239

~Blocked on~:
* #4438
* #4461

Add job `pyk-publish` to publish `pyk` on PyPI (under package name `kframework`).

Release workflow changes from

```
       release --- notify-dependents
                \
pyk-build-docs --- gh-pages
```

to

```
release --- pyk-publish --- notify-dependents
                         \
         pyk-build-docs --- gh-pages

```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -365,8 +365,6 @@ jobs:
 
       - name: Install Poetry
         uses: Gr1N/setup-poetry@v9
-        with:
-          python-version: '3.10'
 
       - name: Setup Caching
         uses: actions/cache@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -347,37 +347,6 @@ jobs:
             const { owner, repo } = context.repo
             await github.rest.repos.deleteRelease({ owner, repo, release_id: ${{ needs.set-release-id.outputs.release_id }} })
 
-  pyk-publish:
-    name: 'Publish pyk'
-    runs-on: ubuntu-latest
-    environment: production
-    permissions:
-      id-token: write
-    needs: [set-release-id, source-tarball]
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Install Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.10'
-
-      - name: Install Poetry
-        uses: Gr1N/setup-poetry@v9
-
-      - name: Build pyk
-        working-directory: pyk
-        run: |
-          make build
-
-      - name: Publish pyk to PyPI
-        uses: pypa/gh-action-pypi-publish@v1
-        with:
-          packages-dir: pyk/dist
-          user: __token__
-          password: ${{ secrets.PYPI_TOKEN }}
-
   release:
     name: 'Publish Release'
     runs-on: [self-hosted, linux, normal]
@@ -434,6 +403,37 @@ jobs:
           script: |
             const { owner, repo } = context.repo
             await github.rest.repos.updateRelease({ owner, repo, release_id: ${{ needs.set-release-id.outputs.release_id }}, prerelease: false })
+
+  pyk-publish:
+    name: 'Publish pyk'
+    runs-on: ubuntu-latest
+    environment: production
+    permissions:
+      id-token: write
+    needs: [set-release-id, source-tarball]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Install Poetry
+        uses: Gr1N/setup-poetry@v9
+
+      - name: Build pyk
+        working-directory: pyk
+        run: |
+          make build
+
+      - name: Publish pyk to PyPI
+        uses: pypa/gh-action-pypi-publish@v1
+        with:
+          packages-dir: pyk/dist
+          user: __token__
+          password: ${{ secrets.PYPI_TOKEN }}
 
   notify-dependents:
     name: 'Notify Dependents'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -410,7 +410,7 @@ jobs:
     environment: production
     permissions:
       id-token: write
-    needs: [set-release-id, source-tarball]
+    needs: release
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -438,7 +438,7 @@ jobs:
   notify-dependents:
     name: 'Notify Dependents'
     runs-on: ubuntu-latest
-    needs: release
+    needs: pyk-publish
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -494,7 +494,7 @@ jobs:
     name: 'GitHub Pages deployment'
     runs-on: ubuntu-24.04
     timeout-minutes: 30
-    needs: [pyk-build-docs, release]
+    needs: [pyk-build-docs, pyk-publish]
     steps:
       - name: 'Install pandoc/texlive/calibre'
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -358,7 +358,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Setup Python
+      - name: Install Python
         uses: actions/setup-python@v5.1.0
         with:
           python-version: '3.10'
@@ -367,7 +367,7 @@ jobs:
         uses: Gr1N/setup-poetry@v9
         with:
           python-version: '3.10'
-    
+
       - name: Setup Caching
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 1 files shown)


pyk/src/kframework/__init__.py
tothtamas28, line 1: The `pyk` package should not be renamed, it can still be distributed under `kframework`.

But since the module name is no longer the default one, `pyproject.toml` should be updated (not tested):

```toml
packages = [
    { include = "pyk", from = "src" },
]
```

To not mess up the template, `cruft.json` (and everything that depends on the corresponding field) should be updated.

Later, if we want to rename the package, we can do so without immediately breaking dependents by distributing it under both names during the transition.

.github/workflows/release.yml
```diff
@@ -344,6 +344,47 @@ jobs:
             const { owner, repo } = context.repo
             await github.rest.repos.deleteRelease({ owner, repo, release_id: ${{ needs.set-release-id.outputs.release_id }} })
 
+  pyk-publish:
```
tothtamas28: Renaming should be the only thing this PR takes care of.

pyk/src/kframework/__init__.py
tothtamas28: Even if only the distribution is renamed, this change is breaking to all dependents. I think the change necessary is

```toml
[dependencies]
pyk = {}
```

to

```toml
[dependencies]
kframework = {}
```

in `pyproject.toml` (and `poetry.lock`, and the CI jobs that depend on it, `update.yml` in particular). In any case, before merging this, it should be tested on `evm-semantics`.

tothtamas28 (dismissed):
The base of the PR should be `develop`.

Baltoli:
I agree with Tamas here - let's make the minimum renaming change that gets us to a package that could be published, then do the pypi steps in a follow-up PR.

F-WRunTime:
> The base of the PR should be `develop`.

Unclear why the master-pr workflow didn't change it. Thanks for pointing it out.

pyk/src/kframework/__init__.py
F-WRunTime: I'm not sure I follow, this is an issue I brought up on slack. The pyk package name is already taken on Pypi and we cannot upload packages with the same name they must be unique. 

> ```toml
> ```toml
> packages = [
>     { include = "pyk", from = "src" },
> ]
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

I don't see this definition in the src. Which file are you referring to?

tothtamas28: > The pyk package name is already taken on Pypi

In Python packaging terminology, the term _package_ is overloaded, as it means two different things:
- An _import package_:
  ```python
  import pyk
  ```
- A _distribution package_:
  ```bash
  $ pip install pyk
  ```

A distribution package can ship more than one import packages, whose names are arbitrary well-formed import package names.

PyPI expects the name of the distribution package to be unique. I haven't tried this, but my presumption is, it does not put any restriction on the names of import packages. So as long as we do not want to install two distribution packages that each ship an import package with the name `pyk`, we're fine.

tothtamas28: > I don't see this definition in the src. Which file are you referring to?

`pyproject.toml`. It's not there, as by default Poetry looks for an import package with the same name as the distribution package. If you want to change that, you need to include the `packages` key.

tothtamas28: Btw, I'm happy to do these changes. Then you can review them and try to upload the resulting (distribution) package to the test repository.

Baltoli:
This is partially superseded by https://github.com/runtimeverification/k/issues/4438, right? I presume after @tothtamas28's PR is merged that this one should be refactored to do _only_ the push to PyPi.

F-WRunTime:
> This is partially superseded by #4438, right? I presume after @tothtamas28's PR is merged that this one should be refactored to do _only_ the push to PyPi.

@Baltoli Correct. Waiting on merge from the other PR first.

Baltoli (dismissed):

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+    name: 'Publish pyk'
+    runs-on: [self-hosted, linux, flyweight]
```
tothtamas28:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v5.1.0
+        with:
+          python-version: '3.10'
+
+      - name: Install Poetry
+        uses: Gr1N/setup-poetry@v9
+        with:
+          python-version: '3.10'
```
tothtamas28: I don't think the action supports that field: https://github.com/Gr1N/setup-poetry

```suggestion
```

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+    environment: production
+    permissions:
+      id-token: write
+    needs: [set-release-id, source-tarball]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5.1.0
```
tothtamas28:
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+        with:
+          python-version: '3.10'
+    
+      - name: Setup Caching
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/poetry/virtualenvs
+          key: ${{ runner.os }}-poetry-${{ hashFiles('pyk/pyproject.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-poetry-
```
tothtamas28: There's not much to cache, `poetry build` does not require a virtual environment.

```suggestion
```

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+      
+      - name: Build pyk
+        working-directory: pyk
+        run: |
+          make build
+
+      - name: Publish pyk to PypI
```
tothtamas28:
```suggestion
      - name: Publish pyk to PyPI
```

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+        uses: pypa/gh-action-pypi-publish@v1.8.14
```
tothtamas28:
```suggestion
        uses: pypa/gh-action-pypi-publish@v1
```

.github/workflows/release.yml
```diff
@@ -509,4 +550,4 @@ jobs:
           set -x
           git add ./
           git commit -m 'gh-pages: Updated the website'
-          git push --force origin gh-pages
+          git push --force origin gh-pages
```
tothtamas28: Trailing newline got removed here.

.github/workflows/release.yml
```diff
@@ -349,6 +349,47 @@ jobs:
...
+    runs-on: [self-hosted, linux, flyweight]
+    environment: production
+    permissions:
+      id-token: write
+    needs: [set-release-id, source-tarball]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
```
tothtamas28: Nitpick:

```suggestion
      - name: Install Python
```

For consistency with the next step and https://github.com/runtimeverification/k/blob/f373630f28653b60b4079cce2242e3bab90ef42c/.github/actions/setup-pyk-env/action.yml#L12

tothtamas28:
When should this step occur? The way I see it, constraints are:
- Should happen after K is published: K is an implicit dependency of `pyk`.
- Should happen after `pyk` documentation is built: the published package refers to the API docs.
- Should happen before notifying dependents: dependency update PRs will refer to the package.

So maybe we should have:
```
       release --- pyk-publish - gh-pages - notify-dependents
pyk-build-docs -/
```

Where `notify-dependents` is a new step extracted from `release`.

@Baltoli, @F-WRunTime, what do you think?

Baltoli:
> When should this step occur?

I agree with your assessment; you've captured the relevant functional dependencies there

Baltoli (approved):

.github/workflows/release.yml
```diff
@@ -404,10 +404,41 @@ jobs:
             const { owner, repo } = context.repo
             await github.rest.repos.updateRelease({ owner, repo, release_id: ${{ needs.set-release-id.outputs.release_id }}, prerelease: false })
 
+  pyk-publish:
+    name: 'Publish pyk'
+    runs-on: ubuntu-latest
+    environment: production
+    permissions:
+      id-token: write
```
tothtamas28, line 412: I wonder, is this part necessary? @Baltoli @F-WRunTime

Baltoli: Unsure; I don't _think_ so looking at the docs but I can't say authoritatively

tothtamas28: The secret is probably defined for the `production` environment.

The other part is apparently for signing the published package: https://github.com/pypa/gh-action-pypi-publish/blob/unstable/v1/README.md?plain=1#L63-L64

Baltoli: Makes sense!

F-WRunTime: Sorry seeing this now, but yes it is necessary. I want to transition to using a verified publisher approach this sets us up for that. And yes the secret is stored in that environment. It's ultimately a better practice than storing everything in a general environment.

PR #4418, opened 2024-06-05
Add sort injections on the KAST level

tothtamas28:
Eliminates `kompiled_kore` from
* `kast_to_kore`
* `krule_to_kore`
* `kflatmodule_to_kore`
* `KPrint`
* `CTermSymbolic`

by
* Adding sort injections on the KAST level as a pass in `kast_to_kore`
* Removing an unnecessary `add_injections` step from `krule_to_kore`

### DIFF
--- a/pyk/src/pyk/konvert/_kast_to_kore.py
+++ b/pyk/src/pyk/konvert/_kast_to_kore.py
@@ -57,20 +57,14 @@ ML_PATTERN_LABELS: Final = dict(
 )
 
 
-def kast_to_kore(
-    kast_defn: KDefinition,
-    kompiled_kore: KompiledKore,
-    kast: KInner,
-    sort: KSort | None = None,
-) -> Pattern:
-    if sort is None:
-        sort = K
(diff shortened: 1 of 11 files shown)


pyk/src/pyk/konvert/_kast_to_kore.py
```diff
@@ -106,8 +99,96 @@ def transform_items(items: tuple[KInner, ...]) -> KInner:
     return top_down(transform, term)
 
 
+def _add_sort_injections(definition: KDefinition, term: KInner, sort: KSort) -> KInner:
```
Baltoli, line 102: Could this method get an explanatory comment perhaps? It's got a complex enough implementation that it would be good to document how it works, even though it's an internal implementation detail.

Baltoli (approved):
One minor comment, but this looks like a really nice cleanup to have!

pyk/src/pyk/konvert/_kast_to_kore.py
tothtamas28: Done.

PR #4420, opened 2024-06-05
Add `pyk` documentation generation script

### LINKED ISSUE #4413: [PyK Docs] Some pages on PyK documentation don't have the expected content of modules/functions
### What K Version?

7.0.108

### Operating System

Linux

### Issue Description

https://kframework.org/pyk/api/pyk.kllvm.hints.prooftrace.html, among other modules that depend on LLVM Python bindings, doesn't have content generated by Sphinx doc. The issue seems to be that they can't find the library as they don't have access to `llvm-kompile` to find it (in the case of files that import `load_static.py`) or to compile the lib (in the case of files that depend on `load.py`), for the first case we still have the problem that we should have it compiled already. 

 We can see the errors and traces in this execution: https://github.com/runtimeverification/k/actions/runs/9353842357/job/25748732638 at the Build Webpage step.

### Steps to Reproduce

Execute `make docs` in the PyK directory without any K/LLVM Backend tools in your `$PATH` as the CI workflow that executes the docs generation doesn't have any of them installed.

### Expected Results

We expect that all modules which depend on LLVM Python Bindings can have their appropriate documentation deployed to https://kframework.org/pyk.

tothtamas28:
Fixes #4413

- Adds a script `pyk/docs/generate.sh` that sets up a virtual environment in which the Sphinx docs can be built without error. In particular, the runtime bindings package needs to be available.
- Changes the test and release workflows to run this script in a K Docker container.


tothtamas28:
Release workflows sliced for test purposes: 6fdc5e1fa9b79d37ccfda498a5c4727cf85bb50c
Test run: https://github.com/runtimeverification/k/actions/runs/9404319093

Robertorosmaninho (approved):
LGTM

PR #4421, opened 2024-06-05
Adding new proof trace tests from Pi2

Robertorosmaninho:
This PR adds a set of new tests that previously lived in the Pi2 repository but no longer need to be there as the company now uses the LLVM Hints Parser instead of its own parser. Therefore, we decided to move the tests from the Pi2 repo to PyK so it can not only increase our test suit but also be as close as possible to the source code.

We decided to keep the whole Hints input instead of using PyK to build it, as we realized that not only were the traces generated using the corresponding PyK function different but also smaller than we expected, modifying the test behavior. One of the differences we spot is that `map_pattern` adds an unnecessary `\left-assoc` in hints production that modifies the final trace.

To facilitate the review of this PR, each new test corresponds to a commit!


pyk/src/pyk/testing/_kompiler.py
```diff
@@ -100,15 +100,22 @@ class KompiledTest:
     KOMPILE_MAIN_FILE: ClassVar[str | Path | None] = None
     KOMPILE_DEFINITION: ClassVar[str | None] = None
     KOMPILE_MAIN_MODULE: ClassVar[str | None] = None
+    KOMPILE_SYNTAX_MODULE: ClassVar[str | None] = None
```
tothtamas28: Extracting this from `KOMPILE_ARGS` is not a bad idea as we use this often.

However, currently, all other tests use

```python3
KOMPILE_ARGS = {'syntax_module': ...}
```

I'd prefer if this change was introduced in a separate PR that addresses those tests as well.

Robertorosmaninho: The problem in this case is that we already set 

```Python
KOMPILE_ARGS = {'llvm_proof_hint_instrumentation': True}
```

in 

```Python
class ProofTraceTest(KompiledTest):
    KOMPILE_BACKEND = 'llvm'
    KOMPILE_ARGS = {'llvm_proof_hint_instrumentation': True}
...
```

Robertorosmaninho: I can create a mock

```K
module DV-SYNTAX
endmodule
```

in this test, and then address this issue and fix the test properly in a future PR. What do you think?

tothtamas28: That's fine. You can also pass multiple keys in the subclass `KOMPILE_ARGS`.

Robertorosmaninho: Ok, I think I managed to find a good temporary solution. Can you please review it?

pyk/src/tests/integration/kllvm/test_prooftrace.py
```diff
@@ -597,6 +597,49 @@ def test_parse_concurrent_counters(self, hints: bytes, header: prooftrace.kore_h
...
+        definition_text = repr(definition).split('\n')
+
+        pt = prooftrace.LLVMRewriteTrace.parse(hints, header)
+        assert pt is not None
+
+        # 11 initialization events
+        assert len(pt.pre_trace) == 11
+
+        # 461 post-initial-configuration events
+        assert len(pt.trace) == 404
```
Baltoli: We should either remove these comments or make sure they're all properly up-to-date.

pyk/src/tests/integration/kllvm/test_prooftrace.py
```diff
@@ -67,16 +69,846 @@ def test_proof_trace(self, hints: bytes, header: prooftrace.kore_header, definit
...
+            rule [id-rule]: id(X:Foo) => X
+            rule [bar-rule]: bar(baz(X:Foo)) => id(id(bar(X)))
+        endmodule
+    """
+
+    KOMPILE_MAIN_MODULE = 'NON-REC-FUNCTION'
+
+    HINTS_INPUT_KORE = """
+        LblinitGeneratedTopCell{}(Lbl'Unds'Map'Unds'{}(Lbl'Stop'Map{}(),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\\dv{SortKConfigVar{}}("$PGM")),inj{SortFoo{}, SortKItem{}}(Lblbar'LParUndsRParUnds'NON-REC-FUNCTION-SYNTAX'Unds'Foo'Unds'Foo{}(Lblbaz'LParUndsRParUnds'NON-REC-FUNCTION-SYNTAX'Unds'Foo'Unds'Foo{}(Lbla'Unds'NON-REC-FUNCTION-SYNTAX'Unds'Foo{}()))))))
+        """
```
Baltoli, line 296: In an _ideal_ world, we'd build these inputs using the KORE data structures available in Pyk, but that's not a critical change.

Robertorosmaninho: Agreed! I tried it but experienced the error mentioned in the PR’s description. Do you know why this happens? Ideally, it should produce the same result, right?

pyk/src/tests/integration/kllvm/test_prooftrace.py
Robertorosmaninho: Done! Thanks!

pyk/src/tests/integration/kllvm/test_prooftrace.py
Baltoli: It should do but probably not worth the effort to debug it too hard!

Baltoli (approved):

pyk/src/pyk/testing/_kompiler.py
Robertorosmaninho: ping @tothtamas28

tothtamas28 (approved):

PR #4427, opened 2024-06-06
Allow specifying that an APRProof generated lemmas should be direct

ehildenb:
Currently, when we are generating lemmas from subproofs to be included in superproofs, we include each basic block of execution of the subproof directly as a rule for execution during the superproof. This means that we mirror the branching structure of the subproof directly, and aren't able to reduce the branching factor. However, when writing proofs that involve loop invariants or branching as `KClaim`, the user often provides the _exact_ final state that matters, even if it's reached in multiple different ways on different branches.

For example, in the MCD proof suite, the `Vat.frob` proof: https://github.com/runtimeverification/evm-semantics/blob/20a292412797498f9eb70486aef19cd2f99cd9f2/tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k#L7

End up reusing several of the `Vat.` arithmetic proofs, like `Vat.muluu` several times: https://github.com/runtimeverification/evm-semantics/blob/20a292412797498f9eb70486aef19cd2f99cd9f2/tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k#L288

If this subproof is included directly as its basic blocks as learned lemmas, then we arrive to a situation where we aren't actually reducing the branching factor in the superproof, even though the `Vat.muluu` specification already describes exactly the behavior that happens on all 3 branches of that proofs.

This PR introduces an `direct_subproof_rules: bool = True` field to the `APRProver` class, so the user can specify whether they want this behavior or not.

- A few logging messages are added to `APRProver.advance_proof` to help with debugging proofs that fail early.
- The lemma generation process is handled more directly in `APRProof.as_rules`, instead of in `APRProve.init_proof`, which now can just call `APRProof.as_rules` directly without any extra processing of the results or deciding how to turn it into lemmas.
- The field `direct_subproof_rules` is added to the `APRProver` class.
  - When building lemmas in `APRProver.init_proof`, we use the field `direct_subproof_rules` to determine how to build lemmas from subproofs.
  - The `APRProof.as_rules` changes its behavior on whether `direct_rule` is set or not. If `direct_subproof_rules` is set and the proof is passing, instead of building the lemmas as the basic blocks in the KCFG, we build it directly from the initial node to the target node of the KCFG. 

This improves the performance of the KEVM MCD Vat Proof suite dramatically in [this PR](https://github.com/runtimeverification/evm-semantics/pull/2467):

```
spec                                                                        |  original time |  new time
VAT-ADDUI-FAIL-ROUGH-SPEC.Vat.addui.fail.rough                              |  38.32540s     |  38.8101s
VAT-ARITHMETIC-SPEC.Vat.adduu.pass                                          |  21.32606s     |  20.8181s
VAT-ARITHMETIC-SPEC.Vat.mului.pass                                          |  72.18441s     |  71.2940s
VAT-ARITHMETIC-SPEC.Vat.muluu.pass                                          |  84.01074s     |  68.8481s
VAT-ARITHMETIC-SPEC.Vat.subui.pass                                          |  104.4093s     |  105.154s
VAT-ARITHMETIC-SPEC.Vat.subuu.pass                                          |  20.32957s     |  21.4818s
VAT-DAI-PASS-SPEC.Vat.dai.pass                                              |  10.81132s     |  10.5747s
VAT-DENY-DIFF-FAIL-ROUGH-SPEC.Vat.deny-diff.fail.rough                      |  58.88762s     |  55.3194s
VAT-FLUX-DIFF-PASS-ROUGH-SPEC.Vat.flux-diff.pass.rough                      |  45.90303s     |  45.7166s
VAT-FLUX-DIFF-PASS-SPEC.Vat.flux-diff.pass                                  |  46.53391s     |  44.7944s
VAT-FOLD-PASS-ROUGH-SPEC.Vat.fold.pass.rough                                |  380.6935s     |  107.786s
VAT-FORK-DIFF-PASS-ROUGH-SPEC.Vat.fork-diff.pass.rough                      |  10961.03s     |  466.333s
VAT-FROB-DIFF-ZERO-DART-PASS-ROUGH-SPEC.Vat.frob-diff-zero-dart.pass.rough  |  4382.788s     |  992.793s
VAT-HEAL-PASS-SPEC.Vat.heal.pass                                            |  34.03113s     |  28.5053s
VAT-MOVE-DIFF-ROUGH-SPEC.Vat.move-diff.pass.rough                           |  38.68574s     |  40.4618s
VAT-SIN-PASS-SPEC.Vat.sin.pass                                              |  10.80752s     |  10.5180s
VAT-SUBUI-FAIL-ROUGH-SPEC.Vat.subui.fail.rough                              |  35.38427s     |  36.0153s
```


ehildenb:
@lucasmt and @PetarMax this is perhaps interesting to you. If you pass `direct_subproof_rules` as an argument here (and thus to `run_prover` in KEVM), you would get the behavior for CSE proofs that it would just jump to the constructed target term of the proof, which basically has symbolic output. So it wouldn't branch, and would lose information about the actual proof execution contents, and if the overall test was still provable in that context, then we didn't need the more specific rules.

Baltoli (approved):
Looks good; clever optimisation!

PetarMax:
This is a great feature, I was thinking about it at some point but then figured it wouldn't be immediately useful because our target nodes in CSE are very free. I wonder how we could specify a more precise target.

PR #4429, opened 2024-06-07
Move simple jobs to latest Ubuntu version

Baltoli:
These jobs were previously running on an old version of Ubuntu, the image for which will be deprecated at some point. This PR just bumps them to use a newer version; the jobs aren't interacting with the OS in any meaningful way so this won't be a breaking change.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ jobs:
 
   set-release-id:
     name: 'Set Release ID'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: 'Get release_id'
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -456,7 +456,7 @@ jobs:
 
   gh-pages:
     name: 'GitHub Pages deployment'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
tothtamas28, line 459: > ```
> sudo apt install --yes wget texlive-xetex libegl1 libopengl0
> ```

Are these packages (`libegl1`, `libopengl0`) available on Noble?

tothtamas28 (approved):

.github/workflows/release.yml
Baltoli: Yes - the command runs successfully with no error on a 24.04 Docker container for me.

Issue #4406, opened 2024-05-30
Cache claim parses


ehildenb:
Currently, in KEVM, when we run any command involving a claim, we run `kprove --dry-run ...` to get the claims out of the text files. Then we check the hash of the claim to find out the proof id on disc, and load that proof if it already exists. For some large claim structures, this whole process can take ~100s, even though all we're doing is pruning a node from the proof for example, which can be done much quicker. But we need to read the claim in to know if the proof stored on disc is up-to-date or not (by parsing it and checking its hash).

We should cache parses of claims, by using the newly merged outer parser. Specifically:

- Here is where we invoke `kprove --dry-run`: https://github.com/runtimeverification/k/blob/a03b705805f72b116219997993863e0051b14195/pyk/src/pyk/ktool/kprove.py#L265
- We should first (if the `use_directory` _exists_, otherwise skip):
  - Invoke the outer parser on the selected file.
  - Read in any `require ...` files as well, and invoke the outer parser on them.
  - Take a hash of the outer-parsed definition.
  - Check if the given hash exists in the `f{use_directory}/parse_cache` (or some similar folder), and if so read in the JSON representation of the parsed definition and return that.
  - If the given hash doesn't exist in the parse cache, then actually invoke `kprove --dry-run ...` and store the parse in the cache.

We perhaps don't even need to do outer-parsing at this point, but we need to be able to slurp in the required files to make sure the parse cache is actually up to date, so maybe it does make sense to do outer parser, and it's pretty quick anyway.

Eventually, we want to improve the cache granularity to be able to detect "does reparsing need to happen for _my_ claim?". If no syntax declarations changed, and only a handful of claims changed, then we only need to reparse that handful of claims.

tothtamas28:
Makes sense to me.

Maybe we can even store the digest in the output file, and let the user manage file locations:

1. Input: `spec_file: Path, output_file: Path`
2. Outer parse `spec_file` and its imports
3. Compute digest (can be refined later on what the input to digest computation is)
4. Check if `output_file` exists.
   1. If not, write `output_file`
   2. If so, check digest stored in `output_file`, and only write file on mismatch.

So the output can have its custom format:

```json
{
  "digest": "",
  "claims": [
    {}
  ]
}
```

As our parsing capabilities improve we can make digest computation more fine-grained:

```json
[
  {
    "digest": "",
    "claim": {}
  }
]
```

ehildenb:
[...]

PR #4436, opened 2024-06-11
Refactor `parse_outer`

tothtamas28:
Preparatory refactoring for:
* #4406

Changes:
* Expose function `slurp_definitions`
* When slurping, a given file's parent is always considered for requires (and takes priority)
* Extract a few helper functions
* Rename, remove or change type for a few parameters


gtrepta (approved):
LGTM. Nice refactorings!

PR #4444, opened 2024-06-13
Extract class `ClaimIndex`

tothtamas28:
Preparatory refactoring for:
* #4406

Extracts a class `ClaimIndex` that enables decoupling the two main steps of `KProve.get_claims`:
- `KProve.get_claim_index`: create a data structure from a K spec file, expensive. We eventually want to save this part by caching the data structure.
- `ClaimIndex.labels`: filter the data structure, cheap.

`ClaimIndex` itself is a mapping from labels to `KClaim`-s. The invariant it enforces (`ClaimIndex._validate`):
- For each item `(label, claim)` in the mapping, `claim.label == label`.
- For each claim, its `depend` attribute values are in the mapping.

In addition, when attribute `main_module_name` is set, it allows looking up claims from the main module without qualifying the label (i.e. to look up `MAIN_MODULE.foo` with key `foo`).

As a next step, `APRProof.from_spec_modules` can be simplified.

### DIFF
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -586,12 +586,18 @@ class ClaimIndex(Mapping[str, KClaim]):
         return len(self.claims)
 
     def __getitem__(self, label: str) -> KClaim:
-        if label in self.claims:
-            return self.claims[label]
         try:
-            return self.claims[f'{self.main_module_name}.{label}']
-        except KeyError as err:
-            raise KeyError(f'Claim not found: {label}') from err
+            label = self.resolve(label)
+        except ValueError:
+            raise KeyError(f'Claim not found: {label}') from None
+        return self.claims[label]
+
+    def resolve(self, label: str) -> str:
+        """Resolve `label` to a valid label in the index, or raise.
+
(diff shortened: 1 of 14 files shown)


pyk/src/pyk/ktool/kprove.py
```diff
@@ -493,3 +480,138 @@ def _prove_claim_rpc(
...
+@dataclass(frozen=True)
+class ClaimIndex(Mapping[str, KClaim]):
+    claims: FrozenDict[str, KClaim]
+    main_module_name: str | None
+
+    def __init__(
+        self,
+        claims: Mapping[str, KClaim],
+        main_module_name: str | None = None,
+    ):
```
ehildenb, line 494: I guess we should go ahead and implement `ClaimIndex.dict` and `ClaimIndex.from_dict` so we can serialize it to disk.

tothtamas28: Either that or just write the `KFlatModuleList` to disk (has all the information, already has these methods). In any case, I'd like to postpone that to the PR that closes #4406.

Baltoli (approved):

Issue #4359, opened 2024-05-16
Sunset `kore-exec`, remove its uses in `kprove`


geo2a:
The transition to the RPC-based provers, that are based on the `kore-rpc` and `kore-rpc-booster` is complete. The old monolithic prover, `kore-exec`, is only used for regression/integration testing at as point, and I'd argue that there are no production uses of it. We would like to sunset `kore-exec` and make the RPC servers the only entry-points to the Haskell backend.

The `kprove` executable uses the Haskell's backend `kore-exec` to prove K specifications. However, `kprove` is also used for converting specification to JSON with `--emit-json-spec`. The second use-case is required, but I'd argue that the first is not.

The main challenge is of course the integration test suites: 
* for `kore-exec` in `haskell-backend` and,
* for `kprove` in `k`.
These will need to be adapted for use with the RPC server.

I'd like to start the discussion!

ehildenb:
KEVM is still testing for tests that aren't passing using the RPC server:

- https://github.com/runtimeverification/evm-semantics/blob/ddef90a245c8f911c0aad2d37530f42595e28570/.github/workflows/test-pr.yml#L119
- https://github.com/runtimeverification/evm-semantics/blob/ddef90a245c8f911c0aad2d37530f42595e28570/kevm-pyk/src/tests/integration/test_prove.py#L326

There are many tests that have not been ported, many of them are because we're missing `circularity` or `depends(...)` attributes on the claims (the old prover assumed _all_ rules were circularities and dependencies, the new one is explicit). But there are some deeper failures too.

Baltoli:
The pain of maintaining `kore-exec` at this point is possibly not quite enough to go through the process of fixing the failing tests in KEVM, but we might want to come back to fix this in the future.

Baltoli:
The KEVM tests are very nearly ready for this to happen; we need to do similarly for the integration test suite and tutorial code.

Baltoli:
Specifically, the regression-new test suite in Pyk

Baltoli:
#4194

Baltoli:
Query: how best to get a single failing node out of the prover tests at the point they get stuck? The translation we need is from:
```diff
< APRProof: A1-SPEC.s1
<     status: ProofStatus.FAILED
<     admitted: False
<     nodes: 4
<     pending: 0
<     failing: 1
<     vacuous: 0
<     stuck: 1
<     terminal: 0
<     refuted: 0
<     bounded: 0
[...]

PR #4445, opened 2024-06-14
Cut on circularity rule application, better error message

ehildenb:
Related: https://github.com/runtimeverification/k/issues/4359
Related: https://github.com/runtimeverification/haskell-backend/issues/3912

This PR is to support passing the final tests on the RPC prover in KEVM. In particular:

- A bug we've been hitting in Haskell backend gets a better error message, to hopefully be more informative off the bat.
- A duplicate log message in `KCFGExplore.section_edge` is removed.
- When doing `KCFGExlore.section_edge`, we were previously just dropping the rule labels of the applied rules on the new edges. Now we save them and store them appropriately on the KCFG.
- When computing the rule labels to store in the KCFG, if the unique-id is not present in teh definition, we still use the supplied unique id because the HB now returns the rule label instead if it's present (https://github.com/runtimeverification/haskell-backend/pull/3916). We just report the location of the rule as `UNKNOWN`.
- The logic for how we apply circularity rules is refined a bit:
  - When discharging a circularity proof obligation, we always consider the circularity rule itself to be a cut-point rule. That's because we need the state immediately following the application of the circularity rule to compare it to subsumption into the initial target state.
  - When discharging a circularity proof obligation, if we have not made any progress yet on the proof, we first take an execution step of depth 1 before resuming normal execute depth. This is because we cannot inject the circularity rule for consideration until we've made some progress, but want to enable it for consideration as early as possible.
  - A test of IMP that before had to manually massage the cut-point-rules list does not have to do that anymore, and is passing directly.

### DIFF
--- a/pyk/src/pyk/proof/reachability.py
+++ b/pyk/src/pyk/proof/reachability.py
@@ -70,6 +70,7 @@ class APRProofStep:
     module_name: str
     shortest_path_to_node: tuple[KCFG.Node, ...]
     prior_loops_cache: FrozenDict[int, tuple[int, ...]] = field(compare=False)
+    circularity_active: bool
 
 
 class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
@@ -176,6 +177,7 @@ class APRProof(Proof[APRProofStep, APRProofResult], KCFGExploration):
                     target=self.kcfg.node(self.target),
                     shortest_path_to_node=tuple(shortest_path),
                     prior_loops_cache=FrozenDict(self.prior_loops_cache),
+                    circularity_active=self.circularity and nonzero_depth(self, node),
                 )
             )
         return steps
@@ -684,6 +686,7 @@ class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
     fast_check_subsumption: bool
     direct_subproof_rules: bool
     kcfg_explore: KCFGExplore
+    circularity_rule_id: str | None
 
     def __init__(
         self,
@@ -706,6 +709,7 @@ class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
         self.always_check_subsumption = always_check_subsumption
         self.fast_check_subsumption = fast_check_subsumption
         self.direct_subproof_rules = direct_subproof_rules
+        self.circularity_rule_id = None
 
     def close(self) -> None:
         self.kcfg_explore.cterm_symbolic._kore_client.close()
@@ -728,6 +732,7 @@ class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
             for rule in subproof.as_rules(priority=20, direct_rule=self.direct_subproof_rules)
         ]
         circularity_rule = proof.as_rule(priority=20)
+        self.circularity_rule_id = circularity_rule.label
 
         _inject_module(proof.dependencies_module_name, self.main_module_name, dependencies_as_rules)
         _inject_module(proof.circularities_module_name, proof.dependencies_module_name, [circularity_rule])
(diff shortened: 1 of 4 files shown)


pyk/src/pyk/proof/reachability.py
```diff
@@ -728,6 +738,7 @@ def _inject_module(module_name: str, import_name: str, sentences: list[KRule]) -
             for rule in subproof.as_rules(priority=20, direct_rule=self.direct_subproof_rules)
         ]
         circularity_rule = proof.as_rule(priority=20)
+        self.circularity_rule_id = circularity_rule.label
```
tothtamas28: During the lifetime of an `APRProver`, is `init_proof` allowed to be called multiple times with different proofs? So for example, is this a valid interaction?

```python3
prover.init_proof(a)       # write circularity_rule_id of a
prover.init_proof(b)       # write circularity_rule_id of b
...
prover.step_proof(a_step)  # read circularity_rule_id of b
```

If so, then `circularity_rule_id` would be better as an attribute on `APRProof`, `APRProofStep`, etc.

ehildenb: `Proof.init_proof` is called at the beginning of `Proof.advance_proof`, and should be called before doing something like `Proof.step_proof`. So I guess this pattern would not be allowed, but also is not prevented.

The issue is that `Proof.step_proof` does not have access to the `Proof`, but a `ProofStep`, so we'd need to pass that information in addition on teh `ProofStep`. I can do that if you'd prefer, as a `str | None` basically.

tothtamas28: Method `step_proof` comes from the abstract superclass, so if we add a parameter, it'll have to be added for all classes that implement it. And `circularity_rule_id` looks specific to `APRProver`.

Would it make sense to make `circularity_rule_id` an attribute of `APRProof`? Then `APRProof.get_steps` can pass that value to `APRProofStep`, so it becomes accessible in `APRProver.step_proof`.

If not, the lifecycle of `APRProver` can still be documented and invalid interactions checked during runtime (as long as it's not inconsistent with the abstract `Prover` lifecycle, e.g. what `parallel_advance_proof` expects from the interface).

@nwatson22, what's your take on this?

PetarMax:
> When doing KCFGExlore.section_edge, we were previously just dropping the rule labels of the applied rules on the new edges. Now we save them and store them appropriately on the KCFG.

This is a bug fix then, great.

> When discharging a circularity proof obligation, we always consider the circularity rule itself to be a cut-point rule. That's because we need the state immediately following the application of the circularity rule to compare it to subsumption into the initial target state.

This is also essential and it's excellent that it's automatic now.

> When discharging a circularity proof obligation, if we have not made any progress yet on the proof, we first take an execution step of depth 1 before resuming normal execute depth. This is because we cannot inject the circularity rule for consideration until we've made some progress, but want to enable it for consideration as early as possible.

Is it reasonable to expect the application of a circularity while executing the first request from the initial node, without any branching first? I think that's what this additional step effectively enables. What would such a circularity mean?

pyk/src/pyk/proof/reachability.py
nwatson22: We don't want to store a `circularity_rule_id` in the prover, since it's proof-specific, like you say. I think adding it to `APRProof` makes sense. Another similar option might be if we could assign the circularity rule a specific label when we create it that we would be able to calculate just from the proof ID, and then we wouldn't have to explicitly include it in the proof step.

ehildenb: Ok, I've refactored it so that the `APRProof` computes/carries the rule id, and passes it into the `APRProver` via the `APRProofStep`.

ehildenb:
@PetarMax in case of a circularity, it will first do one step of execution (to make progress), which will mean it's no longer at the loop head in teh LHS of the rule. The next time it comes to the loop head, the rule will be available as a circularity, so instead of branching (into the loop and out of the loop), it will immediately attempt to take the circularity rule to the conclusion (the RHS). It still needs to check that the reached state is subsumed into the original RHS of the claim at that point.

pyk/src/pyk/proof/reachability.py
```diff
@@ -70,6 +70,9 @@ class APRProofStep:
     module_name: str
     shortest_path_to_node: tuple[KCFG.Node, ...]
     prior_loops_cache: FrozenDict[int, tuple[int, ...]] = field(compare=False)
+    circularity: bool
+    nonzero_depth: bool
+    circularity_rule_id: str
```
tothtamas28, line 75: If `not circularity`, is `circularity_rule_id` relevant? If not, then perhaps the two attributes can be merged into a `str | None`.

tothtamas28 (approved):

pyk/src/pyk/proof/reachability.py
ehildenb: Honestly not sure which approach is best, but going to merge this as is.

PetarMax (approved):

PetarMax:
Looks good to me.

PR #4464, opened 2024-06-20
Fix docs build release job

Baltoli:
This PR fixes a bug that's been breaking the Pyk documentation release jobs; the issue was that we were trying to install the standalone LLVM backend `.deb` package on top of an existing K installation that _already_ contains the LLVM backend.

The fix is to add a flag to the K Docker reusable action that allows the backend installation step to be skipped. You can see a successful run of the fixed workflow here: https://github.com/runtimeverification/k/actions/runs/9604052282/job/26489371015?pr=4464

### DIFF
--- a/.github/actions/with-k-docker/Dockerfile
+++ b/.github/actions/with-k-docker/Dockerfile
@@ -19,7 +19,7 @@ RUN    apt-get -y update    \
          /kframework.deb
 
 RUN if [[ "${INSTALL_BACKEND_DEBS}" == "true" ]]; then \
-      apt-get -y install /llvm-backend.deb \
+      apt-get -y install /llvm-backend.deb; \
     fi
 
 RUN apt-get -y clean
--- a/.github/actions/with-k-docker/Dockerfile
+++ b/.github/actions/with-k-docker/Dockerfile
@@ -3,6 +3,7 @@ ARG BASE_DISTRO
 FROM ubuntu:${BASE_DISTRO}
 
 ARG K_DEB_PATH
+ARG INSTALL_BACKEND_DEBS
 ARG LLVM_BACKEND_DEB_PATH
 
 COPY ${K_DEB_PATH} /kframework.deb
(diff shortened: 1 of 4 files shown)


.github/actions/with-k-docker/Dockerfile
```diff
@@ -15,9 +16,13 @@ RUN    apt-get -y update    \
...
          python3.10-dev     \
          python3.10-venv    \
-         /llvm-backend.deb  \
-         /kframework.deb    \
-    && apt-get -y clean
+         /kframework.deb
+
+RUN if [ "${INSTALL_BACKEND_DEBS}" = "true" ]; then \
+      apt-get -y install /llvm-backend.deb; \
+    fi
```
tothtamas28, line 23: We should clean this up a bit (in a different PR): the `LLVM_BACKEND_DEB_PATH` is required, even if `INSTALL_BACKEND_DEBS` is not `true`.

.github/actions/with-k-docker/Dockerfile
```diff
@@ -15,9 +16,13 @@ RUN    apt-get -y update    \
          make               \
          python3.10-dev     \
          python3.10-venv    \
-         /llvm-backend.deb  \
-         /kframework.deb    \
-    && apt-get -y clean
+         /kframework.deb
```
tothtamas28, line 19: So IIUC this Docker file is used with two kinds of `kframework.deb`: one that contains the LLVM backend and one that doesn't.

tothtamas28 (approved):

.github/actions/with-k-docker/Dockerfile
Baltoli: Yep! We should clean that up

.github/actions/with-k-docker/Dockerfile
Baltoli: yep, it's a slightly awkward design as it stands - will aim to clean up in a following PR per your comment below

PR #4466, opened 2024-06-21
Iterate over an Assoc's elements with top_down/bottom_up

gtrepta:
The `top_down` and `bottom_up` operations on a `Kore` structure don't iterate over the arguments in a `LeftAssoc` or `RightAssoc`. This fixes that.

### DIFF
--- a/pyk/src/pyk/kore/syntax.py
+++ b/pyk/src/pyk/kore/syntax.py
@@ -1625,8 +1625,8 @@ class Assoc(MLSyntaxSugar):
         return ()
 
     @property
-    def patterns(self) -> tuple[()]:
-        return ()
+    def patterns(self) -> tuple[Pattern, ...]:
+        return self.app.args
 
     @property
     def ctor_patterns(self) -> tuple[App]:
@@ -1651,8 +1651,7 @@ class LeftAssoc(Assoc):
         return LeftAssoc(app=app)
 
     def let_patterns(self, patterns: Iterable[Pattern]) -> LeftAssoc:
-        () = patterns
-        return self
+        return self.let(app=self.app.let(args=patterns))
 
     @property
     def pattern(self) -> Pattern:
@@ -1707,8 +1706,7 @@ class RightAssoc(Assoc):
         return RightAssoc(app=app)
 
     def let_patterns(self, patterns: Iterable[Pattern]) -> RightAssoc:
-        () = patterns
-        return self
+        return self.let(app=self.app.let(args=patterns))
 
     @property
     def pattern(self) -> Pattern:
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kore/syntax.py
```diff
@@ -1651,8 +1651,7 @@ def let(self, *, app: App | None = None) -> LeftAssoc:
         return LeftAssoc(app=app)
 
     def let_patterns(self, patterns: Iterable[Pattern]) -> LeftAssoc:
-        () = patterns
-        return self
+        return self.let(app=self.app.let(args=patterns))
```
gtrepta, line 1654: This could instead be
```
(app,) = patterns
assert isinstance(app, App)
return self.let(app=app)
```
Which would also catch the entire `App` during `top_down`/`bottom_up`. @tothtamas28 any opinions? Or, alternatively (see my other comment)

gtrepta:
https://github.com/runtimeverification/k/blob/24146661eeff13489d5138000c13d83251003a88/pyk/src/pyk/kore/syntax.py#L1616-L1617

Maybe this could be...?

```py
class Assoc(MLSyntaxSugar):
    symbol: str
    args: tuple[Pattern, ...]
```

Then the `pattern` method can just build the `App` with `symbol`. And `top_down`/`bottom_up` transformers can match on an `Assoc` and do a `let` over `symbol` if they want.

pyk/src/pyk/kore/syntax.py
tothtamas28: I'm not perfectly happy with either of these solutions:
1. The one solution does not operate on the direct child.
2. The other solution is not type safe (statically).

But then the current solution is not ideal either. ~I think (2) is more straightforward from the programmer's perspective, so I lean towards that.~ **Edit**: inlining the `App` as you suggest in https://github.com/runtimeverification/k/pull/4466#issuecomment-2181797190 addresses both these issues.

If we do this change however:
- There's some code that could be simplified, it would be good to take care of that as well.
- It's a breaking change, so it needs to be tested on `evm-semantics` and `kontrol`.

Would it be acceptable for you to work around this for now (the same way e.g. `kore_to_kast` does)? Then I'll open a PR with the change and the simplifications, and do the downstream testing.

tothtamas28:
> Maybe this could be...?
> 
> ```python
> class Assoc(MLSyntaxSugar):
>     symbol: str
>     args: tuple[Pattern, ...]
> ```

Good idea, let's go with this solution.

pyk/src/pyk/kore/syntax.py
gtrepta: Yeah, I found a workaround that I can use.

Issue #4470, opened 2024-06-21
Pass `-Wno $WARNING` through KompileOptions stack


Baltoli:
Related: https://github.com/runtimeverification/k/issues/4462

@JuanCoRo requested that this option be exposed through `KompileOptions` and onwards to KEVM; it will allow the Kontrol team to silence some persistent warnings that affect their build outputs.

JuanCoRo:
Closing since #4471 addressed this

PR #4471, opened 2024-06-21
Expose `-Wno` via `ignore-warnings` kompile option

JuanCoRo:
Address #4470


Baltoli:
@JuanCoRo you'll need to rebase out these version noise commits; PRs to K should be based on the `develop` branch rather than `master`.

Baltoli (dismissed):
rebase needed

JuanCoRo:
Thanks @Baltoli! Let me know if this looks good. Forgot to start off `develop` :fearful:

pyk/src/pyk/ktool/kompile.py
```diff
@@ -320,6 +327,9 @@ def __call__(
         if outer_parsed_json:
             args += ['--outer-parsed-json']
 
+        if ignore_warnings:
+            args += ['-Wno', ', '.join(ignore_warnings)]
```
nwatson22: Have you tested this? I thought `-Wno` when used with multiple values had to be specified again, e.g.
`kompile -Wno missing-syntax-module -Wno proof-lint`, not `kompile -Wno missing-syntax-module, proof-lint`

nwatson22 (changes requested):

pyk/src/pyk/ktool/kompile.py
JuanCoRo: Just modified it to add `kompile -Wno missing-syntax-module -Wno proof-lint`

JuanCoRo: However, seems to not make any effect:
Screenshot from 2024-06-21 21-33-19

JuanCoRo: The above seemed to be related to the kevm part of that example. After correcting it, no warnings show this time. I've updated the `-Wno` arguments to be of the form `-Wno A,B,C` since it works equally well and takes less space

nwatson22 (approved):

pyk/src/pyk/ktool/kompile.py
```diff
@@ -597,6 +610,10 @@ def args(self) -> list[str]:
         if self.outer_parsed_json:
             args += ['--outer-parsed-json']
 
+        if self.ignore_warnings:
+            warns = ','.join(self.ignore_warnings)
+            args += ['-Wno', warns]
+
```
JuanCoRo: I think this might not be necessary

pyk/src/pyk/ktool/kompile.py
```diff
@@ -512,6 +512,7 @@ class KompileArgs:
     coverage: bool
     bison_lists: bool
     outer_parsed_json: bool
+    ignore_warnings: Iterable[str] | None
```
tothtamas28: This would be better as a parameter of `Kompile.__call__`, please move it there.

tothtamas28 (changes requested):

pyk/src/pyk/ktool/kompile.py
JuanCoRo: Thank you! Just did so. Let me know if anything else needs changing!

pyk/src/pyk/cli/args.py
```diff
@@ -206,6 +206,7 @@ class KompileOptions(Options):
     bison_lists: bool
     no_exc_wrap: bool
     outer_parsed_json: bool
+    ignore_warnings: list[str] | None
```
tothtamas28: I'm not a 100% sure in this, but consider

```suggestion
    ignore_warnings: list[str]
```

This makes it consitent with `ccopts`.

@nwatson22, is this the right convention to follow?

pyk/src/pyk/cli/args.py
```diff
@@ -230,6 +231,7 @@ def default() -> dict[str, Any]:
             'bison_lists': False,
             'no_exc_wrap': False,
             'outer_parsed_json': False,
+            'ignore_warnings': None,
```
tothtamas28: Then this becomes

```suggestion
            'ignore_warnings': [],
```

pyk/src/pyk/cli/args.py
```diff
@@ -475,6 +477,9 @@ def kompile_args(self) -> ArgumentParser:
             action='store_true',
             help='Do not wrap the output on the CLI.',
         )
+        args.add_argument(
+            '--ignore-warnings', dest='ignore_warnings', default=None, action='append', help='Ignore provided warnings'
```
tothtamas28: And here:

```suggestion
            '--ignore-warnings', dest='ignore_warnings', action='append', help='Ignore provided warnings'
```

tothtamas28: For compatibility with the `kompile` CLI:

```suggestion
            '--ignore-warnings', '-Wno', dest='ignore_warnings', default=None, action='append', help='Ignore provided warnings'
```

pyk/src/pyk/ktool/kompile.py
```diff
@@ -279,6 +279,7 @@ def __call__(
         check: bool = True,
         bug_report: BugReport | None = None,
         outer_parsed_json: bool = False,
+        ignore_warnings: Iterable[str] | None = None,
```
tothtamas28: Consider grouping this with the other warnings-related options, i.e.:

```python3
...
warnings_to_errors: bool = False,
ignore_warnings: Iterable[str] | None = None,
...
```

(Similarly for the argument parser, etc.)

pyk/src/pyk/__main__.py
```diff
@@ -315,6 +315,7 @@ def exec_kompile(options: KompileCommandOptions) -> None:
         'gen_glr_bison_parser': options.gen_glr_bison_parser,
         'bison_lists': options.bison_lists,
         'outer_parsed_json': options.outer_parsed_json,
+        'ignore_warnings': options.ignore_warnings,
```
tothtamas28: To fix the test failures, now that the option has been moved to `__call__`, this value needs to be passed to the call at line 344 instead:

```python3
try:
    Kompile.from_dict(kompile_dict)(
        output_dir=kompiled_directory,
        ...
```

pyk/src/pyk/cli/args.py
tothtamas28: Actually, `append` might not be suitable here. Does `kompile` take the option once with a comma-separated list:

```
-Wno foo,bar
```

or multiple times with atomic values:

```
-Wno foo -Wno bar
```

(or perhaps even a combination of those.)

JuanCoRo:
@tothtamas28 just addressed all comments above (hopefully). Let me know if there's anything I missed while we wait on @nwatson22's input on removing the `| None` part of the `ignore_warnings` type.

pyk/src/pyk/testing/_kompiler.py
```diff
@@ -78,6 +78,7 @@ def __call__(
                 output_dir=output_dir,
                 type_inference_mode=TypeInferenceMode.CHECKED,
                 warnings_to_errors=True,
+                ignore_warnings=(),
```
tothtamas28:
```suggestion
```

pyk/src/pyk/cli/args.py
tothtamas28: If we want to be compatible with the `kompile` CLI, we need to support the way `kompile` consumes this option.

@JuanCoRo, can you check which of the above methods `kompile` supports to take multiple warnings to ignore?

Baltoli: Both versions, or a mix of the two, are supported:
```console
$ kompile --version
K version:    v7.1.21
Build date:   Sat Jun 22 00:00:14 BST 2024
$ cat test.k
module TEST
  imports INT

  syntax Int ::= f(Int) [function, total, deprecated]
  rule f(0) => 0
endmodule
$ kompile test.k
[Warning] Compiler: Non exhaustive match detected:
`f(_)_TEST_Int_Int`(#token("1","Int"))
	Source(/Users/brucecollie/code/scratch/test.k)
	Location(4,18,4,54)
	4 |	  syntax Int ::= f(Int) [function, total, deprecated]
	  .	                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[Warning] Compiler: Use of deprecated production found; this syntax may be
removed in the future.
	Source(/Users/brucecollie/code/scratch/test.k)
	Location(5,8,5,12)
	5 |	  rule f(0) => 0
	  .	       ^~~~
$ kompile test.k -Wno non-exhaustive-match -Wno missing-syntax-module,deprecated-symbol
$
```

tothtamas28: This probably calls for a custom action that instead of `append`ing to the list `extend`s it. I'm not sure how well our current CLI convention supports this though.

Anyways, this doesn't have to be addressed in this PR. @JuanCoRo, could you open an issue for tracking this?

JuanCoRo: I don't follow what the problematic behavior here is. Could you elaborate a bit for me to see it @tothtamas28? Thanks!

tothtamas28:
```
-Wno non-exhaustive-match -Wno missing-syntax-module,deprecated-symbol
```

will parse to

```python3
['non-exhaustive-match', 'missing-syntax-module,deprecated-symbol']
```

which is fine in the sense that concatenating them with `,` will give you something you can pass to `kompile`. Would still be nice to parse the atoms separately:

```python
['non-exhaustive-match', 'missing-syntax-module', 'deprecated-symbol']
```

JuanCoRo: Just opened #4479

tothtamas28 (approved):

JuanCoRo:
@nwatson22 @tothtamas28 @Baltoli do you see anything missing in this PR? Need this one merged to start merging the upstream ones. Let me know if there's anything else from my side that needs to be done!

Baltoli (approved):

pyk/src/pyk/cli/args.py
nwatson22: Yes, I'd go without the `None` type option unless there is a situation where it could actually be set to `None`, which there shouldn't be as far as I know since you have `[]` as the default.

PR #4473, opened 2024-06-24
Use prebuilt haskell backend

Baltoli:
This PR uses the existing infrastructure we have for the LLVM backend to download a pre-packaged release of the Haskell backend when running the K integration tests. It's a bit obscured currently because of degraded CI performance generally, but this should take average-case CI run times down to 20-25 minutes.


tothtamas28 (approved):
:rocket:

PR #4480, opened 2024-06-25
Fix bug report request IDs being reset when using multiple `KoreClient`s

### LINKED ISSUE #4478: Backend bug-reporting broken due to request enumeration being reset
When running a proof in Kontrol, I obtained the following transcript:
```
INFO 2024-06-25 12:49:42,378 pyk.kore.rpc - Sending request to localhost:63691: 1 - add-module
INFO 2024-06-25 12:49:42,388 pyk.utils - Added file to bug report report.tar:rpc_contracts%model%TestContract.test_function():0/001_request.json: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmp0hu85pb0
INFO 2024-06-25 12:49:42,391 pyk.utils - Added file to bug report report.tar:commands/000.sh: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpb30mcxu6
[proxy] Processing request 1
INFO 2024-06-25 12:49:44,773 pyk.utils - Added file to bug report report.tar:rpc_contracts%model%TestContract.test_function():0/001_response.json: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpi_tbcqf1
INFO 2024-06-25 12:49:44,774 pyk.utils - Added file to bug report report.tar:commands/001.sh: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmp49hcofxp
INFO 2024-06-25 12:49:44,774 pyk.kore.rpc - Received response from localhost:63691: 1 - add-module
INFO 2024-06-25 12:49:44,862 pyk.kore.rpc - Sending request to localhost:63691: 2 - add-module
INFO 2024-06-25 12:49:44,866 pyk.utils - Added file to bug report report.tar:rpc_contracts%model%TestContract.test_function():0/002_request.json: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpxatv6btc
INFO 2024-06-25 12:49:44,867 pyk.utils - Added file to bug report report.tar:commands/002.sh: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpeolr2l1s
[proxy] Processing request 2
INFO 2024-06-25 12:49:46,644 pyk.utils - Added file to bug report report.tar:rpc_contracts%model%TestContract.test_function():0/002_response.json: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpmmfm1z02
INFO 2024-06-25 12:49:46,646 pyk.utils - Added file to bug report report.tar:commands/003.sh: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpkj2k4cih
INFO 2024-06-25 12:49:46,646 pyk.kore.rpc - Received response from localhost:63691: 2 - add-module
INFO 2024-06-25 12:49:46,654 pyk.kore.rpc - Connecting to host: localhost:63691
INFO 2024-06-25 12:49:46,655 pyk.kore.rpc - Connected to host: localhost:63691
INFO 2024-06-25 12:49:47,035 pyk.kore.rpc - Sending request to localhost:63691: 1 - execute
INFO 2024-06-25 12:49:47,043 pyk.utils - Added file to bug report report.tar:rpc_contracts%model%TestContract.test_function():0/001_request.json: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmpyiceo5tt
INFO 2024-06-25 12:49:47,045 pyk.utils - Added file to bug report report.tar:commands/004.sh: /var/folders/ks/1p8zyp5j7xz_v1krhl4l17tm0000gn/T/tmp_50b95ei
[proxy] Processing request 1
```
which means that the request enumeration is reset after the first two requests, resulting in the third request being marked as `request 1`, overwriting the original `request 1` and corrupting the bug report.

nwatson22:
Fixes #4478.

We were generating response json files for bug reports in a non-unique way so that when we have multiple `JsonRpcClient`s they will overwrite each other's response files. This change ensures that each client will have a unique ID associated which will be included in the path of the response files generated by that client so there are no more name conflicts.

Tested generating bug reports for `kontrol prove` and `kevm prove`.


pyk/src/pyk/kore/rpc.py
```diff
@@ -291,23 +291,24 @@ def close(self) -> None:
         self._transport.close()
 
     def request(self, method: str, **params: Any) -> dict[str, Any]:
-        old_id = self._req_id
+        # Generate unique ID regardless of different clients, repeated requests from the same client, etc.
+        uid = hash_str((id(self), self._req_id))
```
PetarMax: Does this mean that the requests will no longer be linear? If they are not, is there any way of establishing the correct order in the bug reports?

nwatson22: My understanding is that the order is determined by the ordering of the commands files, which are bash scripts that refer to the request files, and those are still named using a counter. This would only be changing the names of the request files (and the IDs of the requests referred to in the requests themselves)
See: https://github.com/runtimeverification/k/blob/0cfdcda602d13ef43f8fbc508c0e405dd0a5784e/pyk/src/pyk/utils.py#L541

tothtamas28: I think I'm missing something. My understanding is that the request file names used to be a function of
1. `id(self)` (through `self._bug_report_id`),
2. `self._req_id`.

My questions are:
1. Which one of the two values causes duplication of file names, and how?
2. Variable `uid` seems to depend exactly on these two values. How does this ensure uniqueness?

nwatson22: `self._bug_report_id` is not uniquely identifying the `JsonRpcClient`. We are currently just setting it as the proof ID, in @PetarMax's example, `rpc_contracts%model%TestContract.test_function():0`.

1. `self._req_id` causes the duplication of file names, because there are two separate `JsonRpcClient`s working on the same proof that each track their own `_req_id` starting from 1 and incrementing only when a request is submitted through that specific `JsonRpcClient`. (Their `_bug_report_id` is the same.)
2. It should be unique because the `id`s of different `JsonRpcClient` objects are different and within a single `JsonRpcClient`, `self._req_id` is still incremented with every new request, so the combination of the two should be unique across all requests from all `JsonRpcClient`s.

tothtamas28: Okay, so IIUC the root cause is, the caller tells the client what `_bug_report_id` to use, and that's not unique. Is this correct?

nwatson22: Yes, I suppose it could be remedied instead by fixing the calling code so it is giving unique bug report IDs every time it's generating a new client.

tothtamas28: Or just redefine the role of `bug_report_id`: if given, the path within the bug report is

```
{bug_report_id}/{client_id}/request_{req_id}.json
```

and otherwise, it's

```
kore_rpc/{client_id}/request_{req_id}.json
```

or

```
{client_id}/request_{req_id}.json
```

nwatson22: I've gone with this approach. The key thing is just to make sure the client ID is guaranteed included in the path. I see it wasn't previously when a `bug_report_id` was passed.

pyk/src/pyk/kore/rpc.py
```diff
@@ -304,10 +308,12 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
...
         _LOGGER.info(f'Sending request to {server_addr}: {old_id} - {method}')
         req = json.dumps(payload)
+        base_name = self._bug_report_id if self._bug_report_id is not None else 'kore_rpc'
+        req_name = f'{base_name}/{self.client_id}/{old_id:03}'
         if self._bug_report:
-            bug_report_request = f'rpc_{self._bug_report_id}/{old_id:03}_request.json'
+            bug_report_request = f'{req_name}_request.json'
             self._bug_report.add_file_contents(req, Path(bug_report_request))
-            self._bug_report.add_command(self._transport.command(self._bug_report_id, old_id, bug_report_request))
+            self._bug_report.add_command(self._transport.command(req_name, bug_report_request))
```
tothtamas28: I wonder, is it possible to pull all this logic into `Transport` (and its subclasses)? Then `command` and `client_id`  can become private on `Transport`.

tothtamas28 (commented):
LGTM, let's consider further refactorings.

pyk/src/pyk/kore/rpc.py
```diff
@@ -304,29 +346,7 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
         server_addr = self._transport.description()
         _LOGGER.info(f'Sending request to {server_addr}: {old_id} - {method}')
```
tothtamas28: The logging call can be moved into `Transport` so that `description` can become private there.

pyk/src/pyk/kore/rpc.py
```diff
@@ -69,11 +105,15 @@ def __exit__(self, *args: Any) -> None:
     def close(self) -> None: ...
 
     @abstractmethod
-    def command(self, bug_report_id: str, old_id: int, bug_report_request: str) -> list[str]: ...
+    def command(self, req_name: str, bug_report_request: str) -> list[str]: ...
```
tothtamas28: Can become private.

pyk/src/pyk/kore/rpc.py
```diff
@@ -69,11 +105,15 @@ def __exit__(self, *args: Any) -> None:
...
 
     @abstractmethod
     def description(self) -> str: ...
 
+    @property
+    def _client_id(self) -> str:
```
tothtamas28: There's a single call to this, can be inlined.

pyk/src/pyk/kore/rpc.py
```diff
@@ -56,6 +56,42 @@ def __init__(self, message: str, code: int, data: Any = None):
...
+                    '-b',
+                    '-s',
+                    f'{req_name}_actual.json',
+                    f'{req_name}_response.json',
+                ]
+            )
+        return resp
+
     @abstractmethod
     def request(self, req: str) -> str: ...
```
tothtamas28: Can become private, `request_with_bug_report` then can be renamed to `request`. All the private abstract methods for `request` can be moved right after it.

pyk/src/pyk/kore/rpc.py
nwatson22: Fixed.

pyk/src/pyk/kore/rpc.py
nwatson22: Fixed.

pyk/src/pyk/kore/rpc.py
nwatson22: Fixed.

pyk/src/pyk/kore/rpc.py
nwatson22: Fixed.

pyk/src/pyk/kore/rpc.py
```diff
@@ -56,8 +56,50 @@ def __init__(self, message: str, code: int, data: Any = None):
 
 
 class Transport(ContextManager['Transport'], ABC):
+    _bug_report: BugReport | None
+    _bug_report_id: str | None
+
+    def __init__(self, bug_report_id: str | None = None, bug_report: BugReport | None = None) -> None:
+        self._bug_report_id = bug_report_id
+        self._bug_report = bug_report
```
tothtamas28:
```suggestion
    def __init__(self, bug_report: BugReport | None = None, bug_report_id: str | None = None) -> None:
        self._bug_report = bug_report
        self._bug_report_id = bug_report_id
```

tothtamas28: Is it valid to have a `bug_report_id` but not a `bug_report`? If it isn't, `ValueError` should be raised when it happens.

pyk/src/pyk/kore/rpc.py
```diff
@@ -56,8 +56,50 @@ def __init__(self, message: str, code: int, data: Any = None):
...
+
+    def request(self, req: str, request_id: int, method_name: str) -> str:
+        base_name = self._bug_report_id if self._bug_report_id is not None else 'kore_rpc'
+        req_name = f'{base_name}/{str(id(self))}/{request_id:03}'
```
tothtamas28:
```suggestion
        req_name = f'{base_name}/{id(self)}/{request_id:03}'
```

pyk/src/pyk/kore/rpc.py
```diff
@@ -87,7 +123,8 @@ class SingleSocketTransport(Transport):
     _sock: socket.socket
     _file: TextIO
 
-    def __init__(self, host: str, port: int, *, timeout: int | None = None):
+    def __init__(self, host: str, port: int, *, timeout: int | None = None, **kwargs: Any):
+        super().__init__(**kwargs)
```
tothtamas28: Consider just explicitly listing `kwargs` here.

tothtamas28 (approved):
Before merging, please test this on `evm-semantics` and `kontrol`.

pyk/src/pyk/kore/rpc.py
nwatson22: Added this error.

pyk/src/pyk/kore/rpc.py
nwatson22: Fixed.

PR #4482, opened 2024-06-25
Add RangeMap functionality to Pyk

Scott-Guest:
Depends on runtimeverification/llvm-backend#1096

Adds RangeMap functionality to Pyk analogous to all the existing built-in collections


pyk/src/pyk/prelude/collections.py
```diff
@@ -51,3 +52,16 @@ def map_item(k: KInner, v: KInner) -> KInner:
...
     ks = dict(ks)
     return build_assoc(map_empty(), KLabel('_Map_'), (map_item(k, v) for k, v in ks.items()))
+
+
+def rangemap_empty() -> KInner:
+    return KApply('.RangeMap')
+
+
+def rangemap_item(k: tuple[KInner, KInner], v: KInner) -> KInner:
+    return KApply('_r|->_', [KApply('RangeMap:Range', k), v])
```
tothtamas28:
```suggestion
    return KApply('_r|->_', (KApply('RangeMap:Range', k), v))
```

(This is a micro-optimization, saves copying the list to a tuple.)

pyk/src/pyk/prelude/collections.py
```diff
@@ -51,3 +52,16 @@ def map_item(k: KInner, v: KInner) -> KInner:
...
+
+
+def rangemap_of(ks: dict[tuple[KInner, KInner], KInner] | Iterable[tuple[tuple[KInner, KInner], KInner]]) -> KInner:
+    ks_dict: dict[tuple[KInner, KInner], KInner] = dict(ks)
```
tothtamas28, line 66: Does the type checker complain for this?

```suggestion
    ks_dict = dict(ks)
```

pyk/src/pyk/kore/match.py
```diff
@@ -86,6 +86,28 @@ def match_map(pattern: Pattern, *, cell: str | None = None) -> tuple[tuple[Patte
     return tuple(entries)
 
 
+def _match_range(pattern: Pattern) -> tuple[Pattern, Pattern]:
```
tothtamas28: Please move this helper after  `match_rangemap`.

tothtamas28 (approved):

pyk/src/pyk/prelude/collections.py
Scott-Guest: Without the annotation, it infers the type as `dict[object, KInner]`. This causes an error on the `rangemap_item(k, v)` call below as `k` is `object` rather than `tuple[KInner, KInner]`.

Baltoli (approved):

PR #4485, opened 2024-06-26
Remove old workaround

Baltoli:
In https://github.com/runtimeverification/k/commit/7ba37fb32f357b0deee710a83b6e58501a2b7212 (8 years ago), we disabled cell collection syntax productions when parsing rules because of OOMs in the compiler. We haven't needed to use this syntax since then, but @PetarMax recently identified a lemma that required being able to do so.


Baltoli:
I verified that KEVM builds successfully with this version of K, and @PetarMax confirmed that it fixes his issue.

tothtamas28 (approved):
Nice finding!

PR #4487, opened 2024-06-27
Inline `app` in `Assoc`

tothtamas28:
Related:
* #4466

This allows type-safe traversal of `Assoc` using `top_down` and `bottom_up`.

### DIFF
--- a/pyk/src/pyk/kore/syntax.py
+++ b/pyk/src/pyk/kore/syntax.py
@@ -22,7 +22,6 @@ if TYPE_CHECKING:
     WS = TypeVar('WS', bound='WithSort')
     WA = TypeVar('WA', bound='WithAttrs')
     ML = TypeVar('ML', bound='MLPattern')
-    AS = TypeVar('AS', bound='Assoc')
 
 
 @final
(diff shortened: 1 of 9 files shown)


Baltoli (approved):

pyk/src/pyk/kore/syntax.py
```diff
@@ -533,6 +533,167 @@ def write(self, output: IO[str]) -> None:
...
+        return '\\right-assoc'
+
+    @classmethod
+    def of(
+        cls: type[RightAssoc],
+        symbol: str,
+        sorts: Iterable[Sort] = (),
+        patterns: Iterable[Pattern] = (),
+    ) -> RightAssoc:
+        return RightAssoc(symbol=symbol, sorts=sorts, args=patterns)
```
gtrepta: Is this meant to be removed? It isn't in `LeftAssoc`.

gtrepta (approved):
Looks good. Thanks!

pyk/src/pyk/kore/syntax.py
tothtamas28: Nice catch!

PR #4492, opened 2024-07-01
Pyk: Wrap fuzzing in an interface

gtrepta:


pyk/src/pyk/ktool/kfuzz.py
```diff
@@ -79,11 +108,15 @@ def fuzz(
         raise RuntimeError('Must pass one of check_func or check_exit_code, and not both!')
 
     def test(subst_case: Mapping[EVar, Pattern]) -> None:
+        if sys.getrecursionlimit() < REC_LIMIT:
```
tothtamas28: Does `hypothesis` change the recursion limit? Does it change it back?

gtrepta: It does change it, and it changes it back. But, the way it changes it doesn't guarantee there's enough recursion depth for our traversals into kast/kore:

https://github.com/HypothesisWorks/hypothesis/blob/d8c17832141f587fec9ef89895ed01a4e9c1650d/hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py#L303-L336

The comment there states that they're adding to the recursion limit, but what it's really doing is setting the limit based on the current depth.

We're already doing this in kasmer. Maybe there's an alternative, I haven't looked into it too much.

https://github.com/runtimeverification/mx-semantics/blob/352ce4d81102a5ad33f44cdabe5c7b80c60d6302/kmultiversx/src/kmultiversx/kasmer.py#L226-L228

gtrepta: It looks like this actually isn't necessary. I didn't realize that we've eliminated a lot of recursion depth in the pyk libraries already.

pyk/src/pyk/ktool/kfuzz.py
```diff
@@ -20,15 +17,38 @@
...
 
+class KFuzz:
+    """Interface for fuzzing over property tests in K."""
+
+    definition_dir: Path
+
+    def __init__(self, definition_dir: Path) -> None:
+        self.definition_dir = definition_dir
+
+    def fuzz_with_check(
```
tothtamas28, line 31: Please add a docstring.

pyk/src/pyk/ktool/kfuzz.py
```diff
@@ -20,15 +17,38 @@
...
+        self,
+        template: Pattern,
+        subst_strategy: dict[EVar, SearchStrategy[Pattern]],
+        check_func: Callable[[Pattern], Any],
+    ) -> None:
+        _fuzz(self.definition_dir, template, subst_strategy, check_func=check_func)
+
+    def fuzz_with_exit_code(
```
tothtamas28, line 44: Here too.

tothtamas28 (approved):

PR #4499, opened 2024-07-04
Fix `match_with_constraint`

### LINKED ISSUE #4496: CTerm.match_with_constraint does not return expected CSubst.constraints
Hi @tothtamas28 !

I am encountering an issue with `CTerm.match_with_constraint` where it does not provide the expected `CSubst` that can be used to transform `self` into `other` via `CSubst.apply`. Instead, I am receiving an empty constraint. I suspect there may be an issue with the `_ml_impl` method.

Here is my test harness:

```python
def test_cterm_match_with_constraint() -> None:
    # Given
    merged_cterm = {
        'config': {
            'args': [{'name': 'X', 'node': 'KVariable'}],
            'arity': 1,
            'label': {'name': '<top>', 'node': 'KLabel', 'params': []},
            'node': 'KApply',
            'variable': False
        },
        'constraints': []
    }
    merged_cterm = CTerm.from_dict(merged_cterm)

    original_cterm = {
        'config': {
            'node': 'KApply',
            'label': {'node': 'KLabel', 'name': '<top>', 'params': []},
            'args': [{'node': 'KVariable', 'name': 'X'}],
            'arity': 1,
            'variable': False
        },
        'constraints': [{
            'node': 'KApply',
            'label': {
                'node': 'KLabel',
                'name': '#Equals',
                'params': [{'node': 'KSort', 'name': 'Bool'}, {'node': 'KSort', 'name': 'GeneratedTopCell'}]
            },
            'args': [
                {'node': 'KToken', 'token': 'true', 'sort': {'node': 'KSort', 'name': 'Bool'}},
                {
                    'node': 'KApply',
                    'label': {'node': 'KLabel', 'name': '_>=Int_', 'params': []},
                    'args': [
                        {'node': 'KVariable', 'name': 'X'},
                        {'node': 'KToken', 'token': '5', 'sort': {'node': 'KSort', 'name': 'Int'}}
                    ],
                    'arity': 2,
                    'variable': False
                }
            ],
            'arity': 2,
            'variable': False
        }]
    }
    original_cterm = CTerm.from_dict(original_cterm)

    # When
    csubst = merged_cterm.match_with_constraint(original_cterm)

    # Then
    assert csubst.apply(merged_cterm) == original_cterm
```

Could you please help me identify what might be going wrong?

Thank you!

tothtamas28:
Fixes #4496

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -163,13 +163,13 @@ class CTerm:
     def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
         """Find `CSubst` instantiating this `CTerm` to the other, return `None` if no such `CSubst` exists."""
         subst = self.config.match(cterm.config)
-        
+
         if subst is None:
             return None
-        
+
         source_constraints = [subst(c) for c in self.constraints]
         constraints = [c for c in cterm.constraints if c not in source_constraints]
-        
+
         return CSubst(subst, constraints)
 
     @staticmethod
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -184,7 +184,8 @@ class CTerm:
 
     def add_constraint(self, new_constraint: KInner) -> CTerm:
         """Return a new `CTerm` with the additional constraints."""
-        return CTerm(self.config, [new_constraint] + list(self.constraints))
+        new_constraint = flatten_label('#And', new_constraint)
+        return CTerm(self.config, new_constraint + list(self.constraints))
 
     def anti_unify(
         self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -247,25 +247,6 @@ class CTerm:
         return CTerm(self.config, new_constraints)
 
 
-def cterm_match(cterm1: CTerm, cterm2: CTerm) -> CSubst | None:
-    """Find a substitution which can instantiate `cterm1` to `cterm2`.
-
-    Args:
-        cterm1: `CTerm` to instantiate to `cterm2`.
-        cterm2: `CTerm` to instantiate `cterm1` to.
-
-    Returns:
-        A `CSubst` which can instantiate `cterm1` to `cterm2`, or `None` if no such `CSubst` exists.
-    """
-    # todo: delete this function and use cterm1.match_with_constraint(cterm2) directly after closing #4496
-    subst = cterm1.config.match(cterm2.config)
-    if subst is None:
-        return None
-    source_constraints = [subst(c) for c in cterm1.constraints]
-    constraints = [c for c in cterm2.constraints if c not in source_constraints]
-    return CSubst(subst, constraints)
-
-
 def anti_unify(state1: KInner, state2: KInner, kdef: KDefinition | None = None) -> tuple[KInner, Subst, Subst]:
     """Return a generalized state over the two input states.
 
@@ -462,5 +443,5 @@ def cterms_anti_unify(
     merged_cterm = cterms[0]
     for cterm in cterms[1:]:
         merged_cterm = merged_cterm.anti_unify(cterm, keep_values, kdef)[0]
-    csubsts = [not_none(cterm_match(merged_cterm, cterm)) for cterm in cterms]
+    csubsts = [not_none(merged_cterm.match_with_constraint(cterm)) for cterm in cterms]
     return merged_cterm, csubsts
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -184,8 +184,7 @@ class CTerm:
 
     def add_constraint(self, new_constraint: KInner) -> CTerm:
         """Return a new `CTerm` with the additional constraints."""
-        new_constraint = flatten_label('#And', new_constraint)
-        return CTerm(self.config, new_constraint + list(self.constraints))
+        return CTerm(self.config, [new_constraint] + list(self.constraints))
 
     def anti_unify(
         self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -163,13 +163,14 @@ class CTerm:
     def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
         """Find `CSubst` instantiating this `CTerm` to the other, return `None` if no such `CSubst` exists."""
         subst = self.config.match(cterm.config)
-
+        
         if subst is None:
             return None
-
-        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))
-
-        return CSubst(subst=subst, constraints=[constraint])
+        
+        source_constraints = [subst(c) for c in self.constraints]
+        constraints = [c for c in cterm.constraints if c not in source_constraints]
+        
+        return CSubst(subst, constraints)
 
     @staticmethod
     def _ml_impl(antecedents: Iterable[KInner], consequents: Iterable[KInner]) -> KInner:
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -1091,10 +1091,6 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         source = self.node(source_id)
         branch_node_ids = [self.create_node(source.cterm.add_constraint(c)).id for c in constraints]
         csubsts = [not_none(source.cterm.match_with_constraint(self.node(id).cterm)) for id in branch_node_ids]
-        csubsts = [
-            reduce(CSubst.add_constraint, flatten_label('#And', constraint), csubst)
-            for (csubst, constraint) in zip(csubsts, constraints, strict=True)
-        ]
         self.create_split(source.id, zip(branch_node_ids, csubsts, strict=True))
         return branch_node_ids
 
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -5,11 +5,8 @@ import logging
 from abc import ABC, abstractmethod
 from collections.abc import Container
 from dataclasses import dataclass, field
-from functools import reduce
 from threading import RLock
 from typing import TYPE_CHECKING, Final, List, Union, cast, final
-
-from pyk.cterm.cterm import cterm_match
 
 from ..cterm import CSubst, CTerm, cterm_build_claim, cterm_build_rule
 from ..kast import EMPTY_ATT
@@ -18,7 +15,6 @@ from ..kast.manip import (
     bool_to_ml_pred,
     extract_lhs,
     extract_rhs,
-    flatten_label,
     inline_cell_maps,
     minimize_rule_like,
     rename_generated_vars,
@@ -1062,7 +1058,7 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         source = self.node(source_id)
         targets = [self.node(nid) for nid in target_ids]
         try:
-            csubsts = [not_none(cterm_match(source.cterm, target.cterm)) for target in targets]
+            csubsts = [not_none(source.cterm.match_with_constraint(target.cterm)) for target in targets]
         except ValueError:
             return None
         return self.create_split(source.id, zip(target_ids, csubsts, strict=True))
--- a/pyk/src/pyk/kcfg/minimize.py
+++ b/pyk/src/pyk/kcfg/minimize.py
@@ -98,11 +98,11 @@ class KCFGMinimizer:
             and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0  # <-- Can we delete this check?
         )
         # Create CTerms and CSubsts corresponding to the new targets of the split
-        new_cterms_with_constraints = [csubst(a.cterm) for csubst in csubsts]
+        new_cterms = [csubst(a.cterm) for csubst in csubsts]
         # Remove the node `B`, effectively removing the entire initial structure
         self.kcfg.remove_node(b_id)
         # Create the nodes `[ A #And cond_I | I = 1..N ]`.
-        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for cterm in new_cterms_with_constraints]
+        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for cterm in new_cterms]
         # Create the edges `[A #And cond_1 --M steps--> C_I | I = 1..N ]`
         for i in range(len(ai)):
             self.kcfg.create_edge(ai[i], ci[i], a_to_b.depth, a_to_b.rules)
--- a/pyk/src/pyk/kcfg/minimize.py
+++ b/pyk/src/pyk/kcfg/minimize.py
@@ -135,8 +135,8 @@ class KCFGMinimizer:
             and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0
         )
         # Remove the node `B`, thereby removing the two splits as well
-        self.kcfg.remove_node(b_id)
         splits_from_a.pop(self.kcfg.node(b_id).id)
+        self.kcfg.remove_node(b_id)
         # Create the new split `A --[..., cond_B #And cond_1, ..., cond_B #And cond_N, ...]--> [..., C_1, ..., C_N, ...]`
         self.kcfg.create_split_by_nodes(a.id, list(splits_from_a.keys()) + list(splits_from_b.keys()))
 
--- a/pyk/src/pyk/kcfg/minimize.py
+++ b/pyk/src/pyk/kcfg/minimize.py
@@ -3,9 +3,8 @@ from __future__ import annotations
 from functools import reduce
 from typing import TYPE_CHECKING
 
-from pyk.cterm import CTerm
 from pyk.cterm.cterm import cterms_anti_unify
-from pyk.utils import not_none, partition, single
+from pyk.utils import partition, single
 
 from .semantics import DefaultSemantics
 
@@ -103,12 +102,12 @@ class KCFGMinimizer:
         # Remove the node `B`, effectively removing the entire initial structure
         self.kcfg.remove_node(b_id)
         # Create the nodes `[ A #And cond_I | I = 1..N ]`.
-        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for (cterm, _) in new_cterms_with_constraints]
+        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for cterm in new_cterms_with_constraints]
         # Create the edges `[A #And cond_1 --M steps--> C_I | I = 1..N ]`
         for i in range(len(ai)):
             self.kcfg.create_edge(ai[i], ci[i], a_to_b.depth, a_to_b.rules)
         # Create the split `A --[cond_1, ..., cond_N]--> [A #And cond_1, ..., A #And cond_N]
-        self.kcfg.create_split(a.id, zip(ai, csubsts, strict=True))
+        self.kcfg.create_split_by_nodes(a.id, ai)
 
     def lift_split_split(self, b_id: NodeIdLike) -> None:
         """Lift a split up a split directly preceding it, joining them into a single split.
@@ -129,7 +128,7 @@ class KCFGMinimizer:
         split_from_a, split_from_b = single(self.kcfg.splits(target_id=b_id)), single(self.kcfg.splits(source_id=b_id))
         splits_from_a, splits_from_b = split_from_a.splits, split_from_b.splits
         a = split_from_a.source
-        ci = list(splits_from_b.keys())
+        list(splits_from_b.keys())
         # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
         assert (  # <-- Does it will be a problem when using merging nodes, because it would introduce new variables?
             len(split_from_b.source_vars.difference(a.free_vars)) == 0
--- a/pyk/src/pyk/kcfg/minimize.py
+++ b/pyk/src/pyk/kcfg/minimize.py
@@ -96,17 +96,10 @@ class KCFGMinimizer:
         # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
         assert (
             len(split_from_b.source_vars.difference(a.free_vars)) == 0
-            and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0
+            and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0  # <-- Can we delete this check?
         )
         # Create CTerms and CSubsts corresponding to the new targets of the split
-        new_cterms_with_constraints = [
-            (CTerm(a.cterm.config, a.cterm.constraints + csubst.constraints), csubst.constraint) for csubst in csubsts
-        ]
-        # Generate substitutions for new targets, which all exist by construction
-        new_csubsts = [
-            not_none(a.cterm.match_with_constraint(cterm)).add_constraint(constraint)
-            for (cterm, constraint) in new_cterms_with_constraints
-        ]
+        new_cterms_with_constraints = [csubst(a.cterm) for csubst in csubsts]
         # Remove the node `B`, effectively removing the entire initial structure
         self.kcfg.remove_node(b_id)
         # Create the nodes `[ A #And cond_I | I = 1..N ]`.
@@ -115,7 +108,7 @@ class KCFGMinimizer:
         for i in range(len(ai)):
             self.kcfg.create_edge(ai[i], ci[i], a_to_b.depth, a_to_b.rules)
         # Create the split `A --[cond_1, ..., cond_N]--> [A #And cond_1, ..., A #And cond_N]
-        self.kcfg.create_split(a.id, zip(ai, new_csubsts, strict=True))
+        self.kcfg.create_split(a.id, zip(ai, csubsts, strict=True))
 
     def lift_split_split(self, b_id: NodeIdLike) -> None:
         """Lift a split up a split directly preceding it, joining them into a single split.
--- a/pyk/src/pyk/kcfg/minimize.py
+++ b/pyk/src/pyk/kcfg/minimize.py
@@ -131,29 +131,15 @@ class KCFGMinimizer:
         a = split_from_a.source
         ci = list(splits_from_b.keys())
         # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
-        assert (
+        assert (  # <-- Does it will be a problem when using merging nodes, because it would introduce new variables?
             len(split_from_b.source_vars.difference(a.free_vars)) == 0
             and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0
         )
-        # Get the substitution for `B`, at the same time removing 'B' from the targets of `A`.
-        csubst_b = splits_from_a.pop(self.kcfg.node(b_id).id)
-        # Generate substitutions for additional targets `C_I`, which all exist by construction;
-        # the constraints are cumulative, resulting in `cond_B #And cond_I`
-        additional_csubsts = [
-            not_none(a.cterm.match_with_constraint(self.kcfg.node(ci).cterm))
-            .add_constraint(csubst_b.constraint)
-            .add_constraint(csubst.constraint)
-            for ci, csubst in splits_from_b.items()
-        ]
-        # Create the targets of the new split
-        ci = list(splits_from_b.keys())
-        new_splits = zip(
-            list(splits_from_a.keys()) + ci, list(splits_from_a.values()) + additional_csubsts, strict=True
-        )
         # Remove the node `B`, thereby removing the two splits as well
         self.kcfg.remove_node(b_id)
+        splits_from_a.pop(self.kcfg.node(b_id).id)
         # Create the new split `A --[..., cond_B #And cond_1, ..., cond_B #And cond_N, ...]--> [..., C_1, ..., C_N, ...]`
-        self.kcfg.create_split(a.id, new_splits)
+        self.kcfg.create_split_by_nodes(a.id, list(splits_from_a.keys()) + list(splits_from_b.keys()))
 
     def lift_splits(self) -> bool:
         """Perform all possible split liftings.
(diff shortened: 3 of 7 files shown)


pyk/src/pyk/cterm/cterm.py
```diff
@@ -166,7 +166,7 @@ def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
         if subst is None:
             return None
 
-        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))
+        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))  # <--
```
tothtamas28: When calculating the `constraint` during matching, an implication is constructed from the pattern and term `constraints`.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -334,7 +334,7 @@ def add_constraint(self, constraint: KInner) -> CSubst:
     def apply(self, cterm: CTerm) -> CTerm:
         """Apply this `CSubst` to the given `CTerm` (instantiating the free variables, and adding the constraints)."""
         _kast = self.subst(cterm.kast)
-        return CTerm(_kast, [self.constraint])
+        return CTerm(_kast, [self.constraint])  # <--
```
tothtamas28: On the other hand, when applying the `CSubst`, the `constraint` is used directly as term `constraints`.

tothtamas28 (commented):
`CTerm.constraint` seems to have two different uses.

pyk/src/pyk/cterm/cterm.py
PetarMax: This implication, I believe, is what should hold in order for the instantiation to indeed be from a more general to a more specific state, but not what should be added to the state. I think that what should be added to the state is 
```
_ml_impl(subst.constraint, cterm.constraints)
```
which says that if the substitution holds, then the constraints are the more specific constraints.

PetarMax: For example, if I am matching 
```
<top> X </top>
#And { true #Equals X >Int 0 }
```
with
```
<top> Y </top>
#And { true #Equals Y >Int 5 }
```
the substitution resulting from configuration matching would be `{ X -> Y }`.

Then, the code we have now would compute
```
{ true #Equals Y >Int 5 } #Implies { true #Equals Y >Int 0 }
```
which is `#Top`, meaning that the instantiation is to a more specific `CTerm`, but this would not transform one `CTerm` into the other.

On the other hand
```
(X #Equals Y)  #Implies { true #Equals Y >Int 5 }
```
would transform one into the other if `apply` first added it and then applied the substitution.

The way `apply` is currently working, it seems to me that it would be enough to add just the constraint: `{ true #Equals Y >Int 5 }` after the substitution has been applied.

tothtamas28: > The way `apply` is currently working, it seems to me that it would be enough to add just the constraint: `{ true #Equals Y >Int 5 }` after the substitution has been applied.

Indeed, in order for

```python
# When
csubst = cpattern.match_with_constraint(cterm)

# Then
assert csubst.apply(cpattern) == cterm
```

to pass, it'd be sufficient to ensure something like `csubst.constraint == ml_and(cterm.constraints)`.

The problem is, `CSubst.constraint` is public, so before changing it's meaning we need to check all locations where it's used.

pyk/src/pyk/cterm/cterm.py
Stevengre: Hmm, wouldn’t it be better to maintain consistency? For example, using `self.constraints`? In fact, it’s not quite consistent within CTerm and CSubst. Ideally, constraints is a sugar for `#AND`. Thus,  I think that`#AND` shouldn’t exist in the elements of constraints, like [this](https://github.com/runtimeverification/k/blob/e6d4f896d7c51efd68f11c77141d457625472c43/pyk/src/pyk/cterm/cterm.py#L70). So we should add some operations to ensure this.

Another thing I think is that I may need a `constraint` property for CTerm to ensure consistency between CTerm and CSubst.

pyk/src/pyk/cterm/cterm.py
Stevengre: I found that CTerm doesn't provide `__eq__`. So the correct equivalence may like this:

```python
assert csubst.apply(cpattern).to_dict() == cterm.to_dict()
```

We may provide normalized return for `CTerm.to_dict()`, or provide a logical `__eq__` for CTerm?

tothtamas28: `CTerm` is a `dataclass`, the `__eq__` method is auto-generated.

Stevengre: Thank you, I didn’t notice that.

Stevengre:
I think the spec of the `match_with_constraint` is this, is that correct?:

```python
MATCH_WITH_CONSTRAINT_TEST_DATA: Final = (
    # (CTerm(generated_top(x)), CTerm(generated_top(x))),
    # (CTerm(generated_top(x)), CTerm(generated_top(y))),
    (CTerm(generated_top(x)), CTerm(generated_top(y), (mlEqualsTrue(geInt(y, intToken(0))),))),
)


@pytest.mark.parametrize('t1, t2', MATCH_WITH_CONSTRAINT_TEST_DATA, ids=count())
def test_cterm_match_with_constraint(t1: CTerm, t2: CTerm) -> None:
    # When
    c_subst1 = t1.match_with_constraint(t2)
    c_subst2 = t2.match_with_constraint(t1)

    # Then
    assert c_subst1 is not None and c_subst1.apply(t1) == t2
    assert c_subst2 is not None and c_subst2.apply(t2) == t1
```

pyk/src/pyk/cterm/cterm.py
tothtamas28: > I think the spec of the `match_with_constraint` is this, is that correct?

We need to inspect all of the call sites to understand how the function is assumed to behave there.

For example, here's how `match_with_constraint` is used in `KCFG` to create a cover edge: https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/kcfg/kcfg.py#L854-L863

In an `APRProof` though, `csubst` comes from an `APRProofSubsumeResult`: https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/proof/reachability.py#L193-L194

Whose `csubst` comes from a subsumption check: https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/proof/reachability.py#L772-L781

Subsumption check: https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/proof/reachability.py#L738-L748

Which calls `CTermSymbolic.implies`: https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/cterm/symbolic.py#L191-L199

Here, the constraints come from the `predicate` field from the [RPC response](https://github.com/runtimeverification/haskell-backend/blob/2faf6c1d71de054f4630e3489efb4631a0f320b5/docs/2022-07-18-JSON-RPC-Server-API.md?plain=1#L361). Also, if the implication is not `valid`, the cover edge will not be added in the first place. So my understanding is:
- In `create_cover(csubst=None)`, the created `csubst` represents a _condition_ under which the entailment holds.
- In `create_cover(csubst=csubst)`, the passed `csubst` is expected to represent some sort of _witness_ of entailment.

@PetarMax, what do you think?

If my analysis is correct, these two uses are not consistent with each other. My solution would be to eliminate the `None` case, which gives us one less calls to `match_with_constraint`.

Stevengre: Thank you for your analysis! Here’s how I understand it:

Given `subst = t1.match_with_constraint(t2)`, the comment seems to indicate that `subst.apply(t1) == t2`.

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/cterm/cterm.py#L163

A cover is an edge where its `csubst` instantiates the target to the source. The code is implemented like this:

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/kcfg/kcfg.py#L858

However, I don’t fully understand the result of `cterm_symbolic.implies`, which comes from the koreclient. But if the cover edge's `csubst` has no other implications, I think `csubst.apply(target_cterm) == node_cterm`.

Therefore, I believe `match_with_constraint` should act as a syntactic implication from cterm to self, similar to the semantic one `cterm_symbolic.implies(node.cterm, target_cterm)`.

So, the spec for `match_with_constraint` should be like:

```python
@pytest.mark.parametrize('t1, t2', MATCH_WITH_CONSTRAINT_TEST_DATA, ids=count())
def test_cterm_match_with_constraint(t1: CTerm, t2: CTerm) -> None:
    # When
    c_subst1 = t1.match_with_constraint(t2)
    c_subst2 = t2.match_with_constraint(t1)

    # Then
    if cterm_symbolic.implies(t2, t1):
        assert c_subst1 is not None and cterm_symbolic.simplify(c_subst1.apply(t1)) == cterm_symbolic.simplify(t2)
    if cterm_symbolic.implies(t1, t2):
        assert c_subst2 is not None and cterm_symbolic.simplify(c_subst2.apply(t2)) == cterm_symbolic.simplify(t1)
```

So we shouldn’t implement it this way, because it will always return `#Top`, as `cterm` implies `self`.

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/cterm/cterm.py#L169

Stevengre: Without modifying the following `apply` method, the `apply` function will remove the existing constraints of the original `CTerm` and won’t apply the `subst` to the constraints.

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/cterm/cterm.py#L334-L337

### Running Examples
Given:
```
t1 = CTerm(generated_top(x))
t2 = CTerm(generated_top(y), (mlEqualsTrue(geInt(y, intToken(0))),))
c_subst1 = t1.match_with_constraint(t2)
```
Then:
```
c_subst1 == {x |-> y, y > 0}
```
Given:
```
t1 = X #And { true #Equals X >Int 0 }
t2 = Y #And { true #Equals Y >Int 5 }
c_subst1 = t1.match_with_constraint(t2)
```
Then:
```
c_subst1 == {X |-> Y, Y > 5 /\ Y > 0}
```

### New Algorithm for This

`constraint = t2.constraints + subst(t1.constraints)`

Stevengre: I think we can simplify this algorithm further to align with https://github.com/runtimeverification/k/pull/4499#issuecomment-2292758129

`constraint = t2.constraint`

The advantage of this approach is that if `t2 -> t1`, then `t2.constraints + subst(t1.constraints) == t2.constraints`. If not, at least we can recover `t2`, which aligns with the comment’s description.

Stevengre: The previous description had some issues. Currently, `CSubst.apply` can only be applied to `CTerm`s with empty constraints; otherwise, it will throw an error. This is due to `CTerm.kast`. Therefore, the current spec is:

```python
@pytest.mark.parametrize('t1, t2', MATCH_WITH_CONSTRAINT_TEST_DATA, ids=count())
def test_cterm_match_with_constraint(t1: CTerm, t2: CTerm) -> None:
    # When
    c_subst1 = t1.match_with_constraint(t2)
    c_subst2 = t2.match_with_constraint(t1)

    # Then
    assert c_subst1 is not None and c_subst1.apply(CTerm(t1.config)) == t2
    assert c_subst2 is not None and c_subst2.apply(CTerm(t2.config)) == t1
```

pyk/src/pyk/cterm/cterm.py
```diff
@@ -166,9 +166,8 @@ def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
         if subst is None:
             return None
 
-        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))  # <--
-
-        return CSubst(subst=subst, constraints=[constraint])
+        # constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))  # <--
+        return CSubst(subst=subst, constraints=cterm.constraints)
```
tothtamas28: Currently, `match_with_constraint` is called from the following locations (tests not included):

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/cterm/cterm.py#L152

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/cterm/cterm.py#L234-L235

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/kcfg/kcfg.py#L858

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/kcfg/kcfg.py#L937

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/kcfg/kcfg.py#L1017

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/kcfg/kcfg.py#L1060

Do we understand the implications of changing the behavior of `match_with_constraint` on these callers? If not, all of these call sites must be inspected to understand why `match_with_constraint` is called there. 

Alternatively, you can always introduce a new function with arbitrary behavior without breaking existing code.

tothtamas28: In particular, I'd like to understand the semantics of `KCFG.Cover.csubst` (as in `KCFG.create_cover` it is populated with a value from `match_with_constraint`).

@PetarMax, when doing proofs, what assumptions do you have on that attribute?

Stevengre: The `Cover` is an edge from a concrete representation to a more abstract representation. `KCFG.Cover.csubst` is used to recover the source from the target. Therefore, the new implementation is correct for this use case.

Stevengre: I’ve reviewed the usage you mentioned.

### CTerm.match

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/cterm/cterm.py#L157-L158

These lines seem a bit strange to me. I don’t think `match` needs to rely on `match_with_constraint`. Implementing it this way might make things clearer.

```python
    def match(self, cterm: CTerm) -> Subst | None:
        """Find `Subst` instantiating this `CTerm` to the other, return `None` if no such `Subst` exists."""
        subst = self.config.match(cterm.config)

        if not subst or len(self.constraints) != 0 or len(cterm.constraints) != 0:
            return None

        return subst
```

I'm not sure if it is the best way to check if the CTerm is a state without constraints.

### CTerm.anti_unify

The updated `match_with_constraint` can achieve the desired functionality here, but the previous version could not. However, the check here is ineffective regardless because `match_with_constraint`'s input doesn't have a `kdef`, so it cannot perform the `implies` operation using the koreclient.

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/cterm/cterm.py#L236-L239


### create_cover

The updated `match_with_constraint` can achieve the desired functionality here, but the previous version could not.

### split_on_constraints

The updated `match_with_constraint` can achieve the desired functionality here, but the previous version could not. Because the `Split` is an edge from souce to target, and the `CSubst` on the edge is to instantiate source to target.

### lift_split_edge

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/kcfg/kcfg.py#L1016-L1019

The implementation here was designed to adapt to the incorrect implementation of `match_with_constraint`. We can eliminate the `.add_constraint(constraint)` when using the current, correct implementation of `match_with_constraint`.

### lift_split_split

Same as above.

Stevengre: In summary, I believe these usages were intended to use `match_with_constraint` according to its comment, but the current implementation is not consistent with that intent.

https://github.com/runtimeverification/k/blob/3692aa781186422363be7a49f6a157714caf84d7/pyk/src/pyk/cterm/cterm.py#L163

PetarMax: I think that here
https://github.com/runtimeverification/k/blob/d8abc6b62823435eae5a2a8df2ae1e8256ae2c43/pyk/src/pyk/kcfg/kcfg.py#L1031-L1034
I am counting on that constraint to be empty, because all those substs contain as constraints are the branching conditions.

Could `CSubst.apply` be the place where changes should be made? What if we just added the constraint there to the already existing constraints of the CTerm?

PetarMax: ...because it kinda does make sense for the constraint of the `CSubst` to be what it is (the implication), because it tells us what must hold for the additional constraints that the `CSubst` produces, but it doesn't quite work with the `apply`. What happens if we change the `apply`, @Stevengre?

Stevengre: I'm not sure if I fully understand, but it seems the current implementation is intended to only add different constraints (`csubst.constraints`). However, the previous implementation of `CSubst.apply` didn’t actually support applying it to a `CTerm` with existing constraints.

In my current implementation, `CSubst.apply` retains only the `config` of the `CTerm`, with all constraints coming from `CSubst`.

https://github.com/runtimeverification/k/blob/c8613fd40f49f1f08ab17a83beb37db49caa2109/pyk/src/pyk/cterm/cterm.py#L333-L336

Meanwhile, `match_with_constraint` now retains only the constraints from `other`. This makes the two functions consistent, allowing them to perform their intended tasks without relying on `koreclient` operations.

For the following code, we no longer need to store constraints; we can simply use `a.cterm.match_with_constraint(cterm)`.

https://github.com/runtimeverification/k/blob/d8abc6b62823435eae5a2a8df2ae1e8256ae2c43/pyk/src/pyk/kcfg/kcfg.py#L1027-L1034

Stevengre: By the way, can I go ahead and fix `anti_unification` or `anti_unify` in this PR?

PetarMax: Yes, I am not sure that that implementation of `apply` is the way to go. Can we modify it just so that it adds the implication coming from the current `match_with_constraints` to the current constraints and see if that works?

tothtamas28: > I don’t think `match` needs to rely on `match_with_constraint`. Implementing it this way might make things clearer.

I think this change can be safely made.

> The updated `match_with_constraint` can achieve the desired functionality here

I don't think there's an agreement yet what that would be though. My main concern is, `KCFG.Cterm.csubst` is mainly populated based on the predicates the backend returns, and that might be very different to the following purely syntactic requirement:

>  `KCFG.Cover.csubst` is used to recover the source from the target.

Stevengre: This is the current test for `match_with_constraints`, which I believe also tests `apply`. Is that what you want?

https://github.com/runtimeverification/k/blob/47911b5591eea6bcb01ef8296a75b416b9a6c229/pyk/src/tests/unit/test_cterm.py#L81-L97

Stevengre:
Since #4628 determines the `apply` in `CSubst` and provides a `cterm_match` that aligns with this `apply`, we can proceed to fix `match_with_constraint` using `cterm_match`.

Stevengre:
Downstream Validation:
#### Other Repos

No direct use.

#### CTerm.match

The following check carries some risk because `match_with_constraint` conducts a syntactic match and does not return an evaluated constraint. However, it is still an improvement over the original implementation, which always returns true when provided with a tighter `cterm`.

https://github.com/runtimeverification/k/blob/4fa4e82e7ae7b58087ecda5550232ee8b876a508/pyk/src/pyk/cterm/cterm.py#L158-L159

#### Cterm.anti_unify

Return a correct constraint.

#### KCFG.create_cover

Return a correct constraint.

#### KCFG.lift_split_edge

Need Refactor

#### KCFG.lift_split_split

Need Refactor

#### KCFG.merge_node

Need Simplify

pyk/src/tests/unit/test_cterm.py
```diff
@@ -75,6 +74,28 @@ def test_no_cterm_match(term: KInner, pattern: KInner) -> None:
...
+
+
+@pytest.mark.parametrize('t1, t2', MATCH_WITH_CONSTRAINT_TEST_DATA, ids=count())
+def test_cterm_match_with_constraint(t1: CTerm, t2: CTerm) -> None:
+    # When
+    c_subst1 = t1.match_with_constraint(t2)
+    c_subst2 = t2.match_with_constraint(t1)
+
+    # Then
+    assert c_subst1 is not None and c_subst1.apply(t1) == t2
```
Stevengre, line 95: Test for `match_with_constraint`

pyk/src/pyk/cterm/cterm.py
```diff
@@ -167,9 +167,10 @@ def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
         if subst is None:
             return None
 
-        constraint = self._ml_impl(cterm.constraints, map(subst, self.constraints))
+        source_constraints = [subst(c) for c in self.constraints]
+        constraints = [c for c in cterm.constraints if c not in source_constraints]
 
-        return CSubst(subst=subst, constraints=[constraint])
+        return CSubst(subst, constraints)
```
Stevengre, line 174: new `match_with_constraint`

pyk/src/pyk/cterm/cterm.py
```diff
@@ -167,9 +167,10 @@ def match_with_constraint(self, cterm: CTerm) -> CSubst | None:
...
```
Stevengre, line 173: Refactor the `match_with_constraint` by the `cterm_match`

pyk/src/pyk/cterm/cterm.py
```diff
@@ -461,5 +443,5 @@ def cterms_anti_unify(
     merged_cterm = cterms[0]
     for cterm in cterms[1:]:
         merged_cterm = merged_cterm.anti_unify(cterm, keep_values, kdef)[0]
-    csubsts = [not_none(cterm_match(merged_cterm, cterm)) for cterm in cterms]
+    csubsts = [not_none(merged_cterm.match_with_constraint(cterm)) for cterm in cterms]
```
Stevengre, line 446: `cterm_match` => `match_with_constraint`

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1062,7 +1058,7 @@ def create_split_by_nodes(self, source_id: NodeIdLike, target_ids: Iterable[Node
         source = self.node(source_id)
         targets = [self.node(nid) for nid in target_ids]
         try:
-            csubsts = [not_none(cterm_match(source.cterm, target.cterm)) for target in targets]
+            csubsts = [not_none(source.cterm.match_with_constraint(target.cterm)) for target in targets]
```
Stevengre, line 1061: `cterm_match` => `match_with_constraint`

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1091,10 +1087,6 @@ def split_on_constraints(self, source_id: NodeIdLike, constraints: Iterable[KInn
         source = self.node(source_id)
         branch_node_ids = [self.create_node(source.cterm.add_constraint(c)).id for c in constraints]
         csubsts = [not_none(source.cterm.match_with_constraint(self.node(id).cterm)) for id in branch_node_ids]
-        csubsts = [
-            reduce(CSubst.add_constraint, flatten_label('#And', constraint), csubst)
-            for (csubst, constraint) in zip(csubsts, constraints, strict=True)
-        ]
```
Stevengre, line 1097: No need to flatten, since it's already flattened during `CSubst` construction.

Stevengre: Additionally, we don't need to add the alpha constraints manually, because of the new `match_with_constraint`.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -96,26 +95,19 @@ def lift_split_edge(self, b_id: NodeIdLike) -> None:
         # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
         assert (
             len(split_from_b.source_vars.difference(a.free_vars)) == 0
-            and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0
+            and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0  # <-- Can we delete this check?
```
Stevengre, line 98: We might have this situation after merging node. I don't know the intent of this assertion.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -96,26 +95,19 @@ def lift_split_edge(self, b_id: NodeIdLike) -> None:
...
         # Create CTerms and CSubsts corresponding to the new targets of the split
-        new_cterms_with_constraints = [
-            (CTerm(a.cterm.config, a.cterm.constraints + csubst.constraints), csubst.constraint) for csubst in csubsts
-        ]
-        # Generate substitutions for new targets, which all exist by construction
-        new_csubsts = [
-            not_none(a.cterm.match_with_constraint(cterm)).add_constraint(constraint)
-            for (cterm, constraint) in new_cterms_with_constraints
-        ]
+        new_cterms_with_constraints = [csubst(a.cterm) for csubst in csubsts]
```
Stevengre: Simplified by using `create_split_by_nodes`

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -96,26 +95,19 @@ def lift_split_edge(self, b_id: NodeIdLike) -> None:
...
         self.kcfg.remove_node(b_id)
         # Create the nodes `[ A #And cond_I | I = 1..N ]`.
-        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for (cterm, _) in new_cterms_with_constraints]
+        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for cterm in new_cterms_with_constraints]
         # Create the edges `[A #And cond_1 --M steps--> C_I | I = 1..N ]`
         for i in range(len(ai)):
             self.kcfg.create_edge(ai[i], ci[i], a_to_b.depth, a_to_b.rules)
         # Create the split `A --[cond_1, ..., cond_N]--> [A #And cond_1, ..., A #And cond_N]
-        self.kcfg.create_split(a.id, zip(ai, new_csubsts, strict=True))
+        self.kcfg.create_split_by_nodes(a.id, ai)
```
Stevengre, line 110: This will automatically generate the substs for splits.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -136,31 +128,17 @@ def lift_split_split(self, b_id: NodeIdLike) -> None:
         split_from_a, split_from_b = single(self.kcfg.splits(target_id=b_id)), single(self.kcfg.splits(source_id=b_id))
         splits_from_a, splits_from_b = split_from_a.splits, split_from_b.splits
         a = split_from_a.source
-        ci = list(splits_from_b.keys())
+        list(splits_from_b.keys())
         # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
-        assert (
+        assert (  # <-- Does it will be a problem when using merging nodes, because it would introduce new variables?
```
Stevengre, line 133: Same question regarding merging node

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -136,31 +128,17 @@ def lift_split_split(self, b_id: NodeIdLike) -> None:
...
-        ci = list(splits_from_b.keys())
-        new_splits = zip(
-            list(splits_from_a.keys()) + ci, list(splits_from_a.values()) + additional_csubsts, strict=True
-        )
         # Remove the node `B`, thereby removing the two splits as well
+        splits_from_a.pop(self.kcfg.node(b_id).id)
         self.kcfg.remove_node(b_id)
         # Create the new split `A --[..., cond_B #And cond_1, ..., cond_B #And cond_N, ...]--> [..., C_1, ..., C_N, ...]`
-        self.kcfg.create_split(a.id, new_splits)
+        self.kcfg.create_split_by_nodes(a.id, list(splits_from_a.keys()) + list(splits_from_b.keys()))
```
Stevengre, line 141: Simplified by `create_split_by_nodes`

pyk/src/tests/unit/test_cterm.py
```diff
@@ -75,6 +75,32 @@ def test_no_cterm_match(term: KInner, pattern: KInner) -> None:
...
+)
+
+
+@pytest.mark.parametrize('t1, t2', MATCH_WITH_CONSTRAINT_TEST_DATA, ids=count())
+def test_cterm_match_with_constraint(t1: CTerm, t2: CTerm) -> None:
+    # When
+    c_subst1 = t1.match_with_constraint(t2)
+
+    # Then
+    assert c_subst1 is not None and c_subst1.apply(t1) == t2
```
Stevengre, line 101: Test for `match_with_constraint`

pyk/src/pyk/kcfg/minimize.py
ehildenb: @PetarMax do you recall why this check was introduced? It seems to be introduced here: https://github.com/runtimeverification/pyk/pull/1009, where it was indicated that this "prevents a bug where split lifting could cause infinite loop", but it's difficult to tell there if a test of that particular case was added.

pyk/src/pyk/kcfg/minimize.py
ehildenb: Let's rename the variable to `new_cterms` then, instead of `new_cterms_with_constraints`.

ehildenb (approved):
@Stevengre I'm approving, but please see my suggestion for renaming a variable that is now simpler (`new_cterms_with_constraints => new_cterms` perhaps?).

I think it's OK for now to wait for Petar to explain the need for that extra check, but let's leave it in for now. Once you have an example of where that check is hindering us from making progress, then it will be easier to add the appropriate test and weaken the check in a follow-up PR.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -96,26 +95,19 @@ def lift_split_edge(self, b_id: NodeIdLike) -> None:
...
-        new_csubsts = [
-            not_none(a.cterm.match_with_constraint(cterm)).add_constraint(constraint)
-            for (cterm, constraint) in new_cterms_with_constraints
-        ]
+        new_cterms = [csubst(a.cterm) for csubst in csubsts]
         # Remove the node `B`, effectively removing the entire initial structure
         self.kcfg.remove_node(b_id)
         # Create the nodes `[ A #And cond_I | I = 1..N ]`.
-        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for (cterm, _) in new_cterms_with_constraints]
+        ai: list[NodeIdLike] = [self.kcfg.create_node(cterm).id for cterm in new_cterms]
```
Stevengre, line 105: changed name

ehildenb (approved):

PR #4504, opened 2024-07-05
README: Add libunwind-dev to list of dependencies for Debian

0cjs:
Without this installed the build on Debian still works, but the following error is produced when you try to run `kompile`:

    ld.lld: error: unable to find library -lunwind
    clang: error: linker command failed with exit code 1...

There is no need to rebuild the K framework after installing `libunwind-dev`; it's just the shared library in its dependency `libunwind8` that we need to fix this. (But we use `libunwind-dev` in the README anyway for consistency.)

This is apparently not seen on MacOS, probably because -lunwind because it's a part of LLVM there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ sudo apt-get install    \
     libjemalloc-dev     \
     libmpfr-dev         \
     libsecp256k1-dev    \
+    libunwind-dev       \
     libyaml-dev         \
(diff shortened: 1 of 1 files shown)


Baltoli:
Adopted in: https://github.com/runtimeverification/k/pull/4512

PR #4509, opened 2024-07-09
Bug report sequence storage changes

nwatson22:
Changes bug reports in the case of KoreServer proofs to indicate the sequence of requests not through the `commands` directory, as this wasn't being used, but through a simpler `sequence` directory, which contains files `0`, `1`, `2`, etc. which just contain the name of the request/response file in that position.

### DIFF
--- a/pyk/src/pyk/utils.py
+++ b/pyk/src/pyk/utils.py
@@ -688,7 +688,7 @@ class BugReport:
             ntf.flush()
             self.add_file(Path(ntf.name), arcname)
 
-    def add_request_command(self, req_name: str, args: Iterable[str]) -> None:
+    def add_request(self, req_name: str) -> None:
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/utils.py
```diff
@@ -688,6 +688,10 @@ def add_file_contents(self, input: str, arcname: Path) -> None:
             ntf.flush()
             self.add_file(Path(ntf.name), arcname)
 
+    def add_request(self, req_name: str) -> None:
+        self.add_file_contents(req_name, Path(f'sequence/{self._command_id}'))
```
jberthold: Let's make it easier to sort these files:
```suggestion
        self.add_file_contents(req_name, Path(f'sequence/{self._command_id:03}'))
```

jberthold (approved):
LGTM

PR #4511, opened 2024-07-10
Add class `ClaimLoader`

tothtamas28:
Function `ClaimLoader(kprove).load_claims` exposes the same interface as `kprove.get_claims`, but additionally, it caches the `--emit-json-spec` result. The saved `.json` file contains a digest for checking if the spec is still up-to-date.

### DIFF
--- a/pyk/src/pyk/ktool/claim_loader.py
+++ b/pyk/src/pyk/ktool/claim_loader.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 from typing import TYPE_CHECKING, NamedTuple
 
 from ..kast.outer import KFlatModuleList
@@ -11,11 +12,14 @@ from .claim_index import ClaimIndex
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
     from pathlib import Path
-    from typing import Any
+    from typing import Any, Final
 
     from ..kast.outer import KClaim
     from . import TypeInferenceMode
     from .kprove import KProve
+
+
+_LOGGER: Final = logging.getLogger(__name__)
 
 
 class ClaimLoader:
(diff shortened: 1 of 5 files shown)


pyk/src/pyk/ktool/claim_loader.py
```diff
@@ -0,0 +1,97 @@
...
+        type_inference_mode: TypeInferenceMode | None = None,
+    ) -> list[KClaim]:
+        digest = self._digest(spec_file, include_dirs=include_dirs, md_selector=md_selector)
+
+        claim_file = spec_file.with_suffix('.json')
+
+        cache_hit = False
+        if claim_file.exists():
+            module_list, _digest = _ClaimModuleList.from_dict(json.loads(claim_file.read_text()))
+            cache_hit = digest == _digest
```
Baltoli: Maybe call this `loaded_digest` or something? It caught me out very briefly scanning from the top that there's `self._digest` as well

pyk/src/pyk/ktool/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+from ..utils import run_process_2
+
+if TYPE_CHECKING:
+    from typing import Final
+
+
+@final
+@dataclass(frozen=True)
+class KDistribution:
```
Baltoli, line 15: Could this maybe get a docstring as it's a top-level tool?

Baltoli (approved):
A couple of very pedantic comments, but looks good!

pyk/src/pyk/ktool/claim_loader.py
tothtamas28: Done.

pyk/src/pyk/ktool/utils.py
tothtamas28: Done.

PR #4512, opened 2024-07-10
[adopt: #4504] README: Add libunwind-dev to list of dependencies for Debian

Baltoli:
Without this installed the build on Debian still works, but the following error is produced when you try to run `kompile`:

    ld.lld: error: unable to find library -lunwind
    clang: error: linker command failed with exit code 1...

There is no need to rebuild the K framework after installing `libunwind-dev`; it's just the shared library in its dependency `libunwind8` that we need to fix this. (But we use `libunwind-dev` in the README anyway for consistency.)

This is apparently not seen on MacOS, probably because -lunwind because it's a part of LLVM there.

> This is an adopted PR from @0cjs (#4504)


gtrepta (approved):
LGTM. Do we know why this wasn't an issue before? I recall having the `-lunwind` error on my machine and installing this package to fix it a few weeks ago, it was a bit of a mystery that I simply forgot about.

Baltoli:
> Do we know why this wasn't an issue before?

Dwight only added it as an dependency of the LLVM backend recently.

PR #4518, opened 2024-07-12
Set logging level to `DEBUG` for `llvm_interpret_raw`

### LINKED ISSUE #4516: `llvm_interpret` logs are very verbose
With the merge of #4497, any calls to `llvm_interpret` or `llvm_interpret_raw` now log the stdin and stdout of the interpreter process, which normally are both very large kore terms. Some way to suppress the logging of choice file descriptors when invoking these or `run_process2` should be an option.

I discovered this when working on the fuzz testing, which will now print out ~100Mb to stdout when kasmer fuzzes over a multiversx contract.

tagging @tothtamas28

tothtamas28:
Fixes #4516

* Add parameter `loglevel` to `run_process_2`.
* Set `loglevel` to `DEBUG` for `llvm_interpret_raw`. This is a more reasonable choice than `INFO`, since we only need this information for diagnosing problems.
* Set the `logger` to the logger of the host module for `llvm_interpret_raw`. This enables finer grained configuration from the application.

### DIFF
--- a/pyk/src/pyk/utils.py
+++ b/pyk/src/pyk/utils.py
@@ -464,6 +464,7 @@ def run_process_2(
     cwd: str | Path | None = None,
     env: Mapping[str, str] | None = None,
     logger: Logger | None = None,
+    loglevel: int | None = None,
     check: bool = True,
 ) -> CompletedProcess:
     if type(args) is str:
@@ -478,6 +479,9 @@ def run_process_2(
     if not logger:
         logger = _LOGGER
 
+    if loglevel is None:
+        loglevel = logging.INFO
+
     res = _subprocess_run(
         args,
         input=input,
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/utils.py
```diff
@@ -649,11 +655,6 @@ def ensure_dir_path(path: str | Path) -> Path:
     return path
 
 
-# Implementation because of outdated Python versions: https://github.com/python/cpython/blob/1de4395f62bb140563761ef5cbdf46accef3c550/Lib/pathlib.py#L554
-def is_relative_to(_self: Path, other: Path) -> bool:
-    return _self == other or other in _self.parents
-
-
```
tothtamas28, line 656: This is an unrelated cleanup, but since it is straightforward to adjust for in dependents I decided to include it in this PR.

gtrepta (approved):
Looks good, and it works. Thanks!

Issue #4502, opened 2024-07-05
CTerm.anti_unify doesn't provide expected constraints for CTerm and CSubst.


Stevengre:
Hi @nwatson22!

I’ve encountered an issue with the `Cterm.anti_unify` function, which doesn’t behave as expected. Below is a simple example provided by @PetarMax  to illustrate the problem.

### Given

CTerm1:
```xml
<k> X </k> 
<whatever> 5 </whatever>
X > 0
```

CTerm2:
```xml
<k> Y </k>
<whatever> 3 </whatever>
Y > 5
```

### Expected Anti-Unification

Anti-Unified CTerm:
```xml
<k> A </k>
<whatever> B </whatever>
{ (A ==Int X #And B ==K 5) #Implies X >Int 0 } #And { (A ==Int Y #And B ==Int 3) #Implies Y >Int 5 }
```

CSubst1
```xml
{A -> X; B -> 5}
```

CSubst1
```xml
{A -> Y; B -> 3}
```

### Actual Result

Anti-Unified CTerm:
```xml
<k> A </k>
<whatever> B </whatever>
```

CSubst1
```xml
{A -> X; B -> 5}
```

CSubst1
```xml
{A -> Y; B -> 3}
```

### Additional Information

1. Issue #4499 is addressing more reasonable constraints for the result of `match_with_constraint`.
2. Issue #4425 is utilizing `match_with_constraint` to obtain the constraint substitution and `anti-unify` for merging nodes.

### Open Questions

1. The `Expected Anti-Unification` result may not be optimal depending on the desired KCFG structure. It’s possible to place all constraints in either the CTerm or the CSubst. However, it may be more effective to differentiate constraints: common constraints should remain in the CTerm, while specific constraints should be in the respective CSubst. Consequently, X >Int 0 and Y >Int 5 should be in CSubst1 and CSubst2, respectively.
2. This function might be used for loop invariant generation. While it works well due to `anti-unify`’s termination properties (transforming specific CTerms into general CTerms with free variables and empty constraints), tightening the generated CTerm constraints could lead to non-termination or delayed termination but better loop-inv. We may need to prove termination for loop invariant generation.

Thank you for your attention to this matter.

Stevengre:
[...]

PR #4521, opened 2024-07-15
4425 node merging to reduce branching for cse

### LINKED ISSUE #4425: Node merging to reduce branching for CSE
Currently we have the `minimize_kcfg` routine, which pulls branches up to the top of a the KCFG, and unions edges together. That can improve performance of a transition system by reducing the amount of rules that are injected into the definition, but does not reduce overall branching factor of the function. Here is a link to a diagram for references: https://docs.google.com/drawings/d/1l8i_aUMcL8E7WyxHmk2cqRIY74fQ-yvMdhhdDNcOUp0/edit.

image

In this diagram, on the left, we've already run the existing KCFG minimization routine which lifts splits to the top and collapses subsequent edges into a single edge. We want to transform it into the diagram on the right, which has pushed the split _forward_ through the edges that come after it. In order to make this transformation preserve information, we make the following adjustments:

- `Edge`s now store a _list_ of tuples `(depth, rules, condition)`, which means that "from `source` node, given `condition`, you will reach `target` node using `rules` at `depth`". Current edges in the KCFG just store a single `depth, rules`, with `condition` always being `#Top`.
- When we see a split with edges following, we run a semantics-specific heuristic called `merge_nodes(CTerm, CTerm) -> bool` on the nodes following the edges. In the case of the diagram, we pairwise run it on all of `B`, `C`, `D` (or do it iteratively for each pair and do smaller mergings). If two nodes (such as `B` and `C`) have `merge_nodes` return `True`, we do the following:
  - Compute the node `B \/ C`, which has fresh variables in places where the configurations differ, and the specific assignments to the fresh variables conditioned on the path-condition from teh split node leading to each node. So for example, if `E` represents the more abstract configuration, we would compute `E #And { PC1 #Implies SUBST1 } #And { PC2 #Implies SUBST2 }`, where `SUBST1` instantiates `E` to `B`, and `SUBST2` instantiates `E` to `C`.
  - Remove nodes `A1` and `A2` from the graph, and insert new node `B \/ C`, and insert an edge from `A` to `B \/ C` with edge data `[(d1, PC1), (d2, PC2)]`, and insert a split from `B \/ C` to `B` and `C` with conditions `PC1`, `PC2`.
  - Repeat procedure with other splits followed by edges, where the user-supplied heuristic fires on the edge targets.

This effectively "pushes splits down", allowing us to reduce the overall branching factor of the proof if we can push them all the way down to the target node of the proof.

Stevengre:
This is an implementation for #4425; the changes related to #4499 and #4502 will affect the `merge_node` functionality.

Please note that while this implementation is consistent with the design of #4425, it is not identical. It takes into account some general cases of mergeable nodes.

```mermaid
graph TD
  1 -->|"Split: X >=Int 5"| 2
  2 -->|"Edge: (5, (r1), #Top)"| 6
  1 -->|"Split: X >=Int 3 /\ X <Int 5"| 3
  3 -->|"Edge: (10, (r2,r3), X >=Int 4 /\ X <Int 5), (15, (r4), X <Int 4 /\ X >=Int 3 )"| 7
  1 -->|"Split: X >=Int 0 /\ X <Int 3"| 4
  4 -->|"Edge: (20, (r5), #Top)"| 8
  1 -->|"Split: X  <Int 0"| 5
  5 -->|"Edge: (25, (r6), #Top)"| 9
```

With mergeable pairs (2,3), (2,4), (3,4), (3,5), and (4,5).
Based on the largest connected subgraph, we identify two mergeable groups: (2,3,4) and (3,4,5), as 2 and 5 cannot be merged. Thus, the merged result should be:

```mermaid
graph TD
1 -->|"Split: X >=Int 0"| 10["2 | 3 | 4"]
10 -->|"Edge: (5, (r1), X >=Int 5),(10, (r2,r3), X >=Int 4 & X <Int 5), (15, (r4), X <Int 4 /\ X >=Int 3 ),(20, (r5), X >=Int 0 /\ X <Int 3),"| 11["6 | 7 | 8"]
11 -->|Split| 6
11 -->|Split| 7
11 -->|Split| 8
1 -->|"Split: X <Int 5"| 12["3 | 4 | 5"]
12 -->|"Edge: (10, (r2,r3), X >=Int 4 & X <Int 5), (15, (r4), X <Int 4 & X >=Int 3 ),(20, (r5), X >=Int 0 & X <Int 3),(25, (r6), X <Int 0)"| 13["7 | 8 | 9"]
13 -->|Split| 7
13 -->|Split| 8
13 -->|Split| 9
```

Open Issues:
1. CTerm is anti-unified to a form without constraints. Should we add all constraints to the CTerm?
2. Currently, CSubst can contain redundant constraints, e.g., x > 0, x > 1 should be merged to x > 1.
3. Additionally, CSubst is sensitive to the order of constraints. Should we override the `__eq__` method to address this?
4. Furthermore, the CSubst list is essentially a sugar for $\land$. Should we standardize the constraints to conjunctive normal form, or at least flatten all $\land$ by default to simplify constraint reasoning? Or, since the constraints are user-facing, should we retain the original form for better user understanding?


ehildenb:
@Stevengre you need to rebase your branch on `develop` branch of this repository, not `master`.

Also, we can assume that the `merge_nodes` heuristic will create a _partition_, that is, it will not assign the same node to be merged into two different buckets. In yoru example above, nodes 7 and 8 seem to be in both node-merge buckets, so you get those cross-splits. We don't need that functionality.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -684,6 +732,90 @@ def remove_node(self, node_id: NodeIdLike) -> None:
         for alias in [alias for alias, id in self._aliases.items() if id == node_id]:
             self.remove_alias(alias)
 
+    def remove_node_safely(
+            self,
+            node_id: NodeIdLike,
+            predecessors: dict[str, tuple[NodeIdLike, ...]] = None,
+            successors: dict[str, tuple[NodeIdLike, ...]] = None,
+    ) -> None:
```
ehildenb: Why do we have this method? Don't we already have `remove_node`? It should work fine, right?

Stevengre: Although the current remove node operation does not cause dangling issues, it directly removes edges related to the node, leading to some unexpected situations. For example:
```mermaid
graph LR
A -->|Split| B
A -->|Split| C
```
Removing node B results in:
```mermaid
graph LR
A 
C
```
However, what I expect is:
```mermaid
graph LR
A -->|Split| C
```

Stevengre:
> Also, we can assume that the merge_nodes heuristic will create a partition, that is, it will not assign the same node to be merged into two different buckets. In yoru example above, nodes 7 and 8 seem to be in both node-merge buckets, so you get those cross-splits. We don't need that functionality.

@ehildenb  If this is the case, i.e., (2,3), (2,4), (3,4), (3,5), (4,5) (2,5), then the current implementation will work as mentioned in #4425. I am just providing a more general solution to handle the special cases that might appear in the figure. I believe that the meaning of KCFGSemantics.mergable(CTerm, CTerm) is to specify that two CTerms are mergable, so this situation might arise. And this merging method aligns with the semantics of mergable.

ehildenb:
@Stevengre please rebase your branch on `develop`. And like I said, we don't need a more general solution, a _simple_ solution is better if it still achieves our goals, because it leads to more maintainable code. Please read this article: https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding. We _must_ prioritize simplicity of code over completeness. It is _significantly_ better to have simpler code that handles 80% of the cases (but handles the cases we need), than complex code that handles 100% of the cases (including cases we don't necessarily need).

Stevengre:
Thank you for your suggestion. And sorry for my late reply. I will make sure to study the article thoroughly. I will try to rebase my branch, but please give me one day to focus on my paper. Some problems still need to be fixed for it. I will complete this work by next Monday.

Stevengre:
By the way, I have one more thing to confirm. Regarding this functionality: should we assume that 1. All branch targets of the split edge will be merged? 2. Is the mergable relationship transitive?

ehildenb:
Mergability should be an equivalence relation. So it should _partition_ the nodes. Transitive, commutative, and reflexive.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: This is actually expected behavior, because `Split` nodes must maintain that they are _total_ and _mutually exclusive_. Mutually exclusive means that nodes `B` and `C` must not have any intersection of their states. Total means that node `A` must represent the same set of states as the union of nodes `B` and `C`. So putting a split like `A -> C` violates the totalness requirement, so the split must be removed.

Stevengre:
Thank you! I understand. This way, we won't have a situation where (2,4) and (4,5) are mergable but not (2,5). However, it is still possible that some split targets may be merged while others are not, right?

ehildenb:
Yes, it's still possible some split nodes are merge-able and others are not.

But you can do that by refactoring the way the splits happen. See for example here: https://docs.google.com/drawings/d/1l8i_aUMcL8E7WyxHmk2cqRIY74fQ-yvMdhhdDNcOUp0/edit

Between diagram 1 and 2, we determined that `B` and `C` are mergeable, and `D` not. So we introduce a pre-split between `B \/ C` and `D`, then split `B \/ C` into `B` and `C`. That way the merging can happen over the entire split.

Stevengre:
Thank you for sharing this article (https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding)! I will do my best to make my PRs easy to review by following the correct, fast, simple, and complete guidelines. However, I apologize as I am still new to collaborative coding and might make some mistakes. If you see any areas that need improvement, please let me know. Thank you again!

PR #4532, opened 2024-07-18
Adding `--proof-hint` flag to PyK KRun

Robertorosmaninho:
This PR aims to add the `--proof-hint` krun's flag to PyK infrastructure so we can use it in semantics that currently relies on PyK for building itself and execute programs.

Fixes: https://github.com/Pi-Squared-Inc/pi2/issues/1725


pyk/src/pyk/ktool/krun.py
```diff
@@ -265,6 +277,7 @@ def _build_arg_list(
     search_final: bool,
     no_pattern: bool,
     debugger: bool,
+    proof_hint: bool = False,
```
tothtamas28:
```suggestion
    proof_hint: bool,
```

pyk/src/pyk/ktool/krun.py
```diff
@@ -129,8 +133,14 @@ def run(
             pipe_stderr=pipe_stderr,
             bug_report=bug_report,
             debugger=debugger,
+            proof_hint=proof_hint,
         )
 
+        if proof_hint:
+            # Print the binary proof hint to stdout regardless of the output option
+            write(1, result.stdout)
```
Baltoli: I'm not sure what the intent of the code here is; it seems confusing that the proof hint mode will unconditionally escape the Python framework and dump directly to stdout.

tothtamas28: I agree, it would be better to expose this operation mode as a separate function.

pyk/src/pyk/ktool/krun.py
```diff
@@ -115,6 +118,7 @@ def run(
         pipe_stderr: bool = True,
         bug_report: BugReport | None = None,
         debugger: bool = False,
+        proof_hint: bool = False,
```
tothtamas28: Unfortunately, this approach won't work: under the hood it leads to a call like

```python3
subprocess.run(['krun', 'test.imp', '--proof-hint'], text=True, stdout=subprocess.PIPE)
```

which will raise an exception:

```python3
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 8: invalid start byte
```

This is because
* `stdout=subprocess.PIPE`, so `CompletedProcess.stdout` will be populated.
* `text=True`, so the function will try to decode the bytes output to produce `CompletedProcess.stdout`.

You drop any of these arguments and it'll work.

Please add a test to reproduce this issue.

Workarounds:
1. Reuse `_build_arg_list` and call `subprocess.run` directly.
2. If we know how the bytes output is supposed to be decoded, we can thread argument `encoding` through to `subprocess.run` ([standard encodings](https://docs.python.org/3/library/codecs.html#standard-encodings)).
3. You can pass `pipe_stdout=False` to `run_process`, so that you have `stdout=None` in `subprocess.run`.
4. We can patch `run_process` (and it's newer version, `run_process_2` ) to accept `text=False`, but that's not straightforward.

I'd prefer a solution that won't prevent us from dropping `run_process` in favor of `run_process_2` in the future, i.e. (1) or (2).

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: This is the actual behavior of the flag in the krun. `kprint` can't handle this kind of binary, so the idea is just to dump it in the console as `krun` does. The alternative, from my point of view, would be to Dump into a file but escape the Python framework anyway.

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: You're right! I dropped `text=True` locally and forgot to commit this.

Robertorosmaninho: About your workarounds:
1. Isn't that just a different way to create a third `run_process`?
2. I don't think we can do this as the encoding was created by us on the llvm backend serializer; we really need to output the binary itself, not its text representation/deserialization.
3. That's a good approach; we can have `pipe_stdout = not prof_hint`.
4. Indeed, that's not trivial and isn't worth spending time on. 

If you don't mind, I would prefer to stick to option 3 to unblock Pi2 folks, and when we need to drop `run_process` in favor of `run_process_2`, we can sit together again to brainstorm a way to make it compatible with proof hints.

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: For instance, krun only `cat .../result.kore` when `--prof-hint` is passed as opposite to call `kore-print ... result.kore` when this flag isn't in use. 

So I believe this mimics the same behavior we already have.

Baltoli: > The alternative, from my point of view, would be to Dump into a file but escape the Python framework anyway.

The point is that we manage `stdout` and `stderr` through the `CompletedProcess` object. If the hint has been captured by the subprocess library, then the way to get it to the calling app's `stdout` is not to bypass that management entirely; you should probably instead have the calling code write it out to a file yourself.

Baltoli (dismissed):
Looks good to me post revision but let @tothtamas28 review before you merge!

pyk/src/pyk/ktool/krun.py
```diff
@@ -246,7 +257,14 @@ def _krun(
         else:
             bug_report.add_command(args)
 
-    return run_process(args, check=check, pipe_stderr=pipe_stderr, logger=logger or _LOGGER, exec_process=debugger)
+    return run_process(
+        args,
+        check=check,
+        pipe_stderr=pipe_stderr,
+        pipe_stdout=not proof_hint,
```
tothtamas28: Consider exposing `pipe_stdout` instead, and checking `not proof_hint or pipe_stdout`. You can raise `ValueError` with a descriptive error message to tell the user why this is necessary.

pyk/src/pyk/ktool/krun.py
```diff
@@ -129,8 +133,13 @@ def run(
             pipe_stderr=pipe_stderr,
             bug_report=bug_report,
             debugger=debugger,
+            proof_hint=proof_hint,
         )
 
+        if proof_hint:
+            # We didn't pipe the result to stdout, so it was already printed 
+            return
```
tothtamas28: This is a new mode of operation, please add a dedicated function, say, `run_proof_hint` for this.

pyk/src/tests/unit/ktool/test_krun.py
```diff
@@ -23,6 +23,7 @@
         ('search_final', False),
         ('no_pattern', False),
         ('debugger', False),
+        ('proof_hint', False),
```
tothtamas28, line 26: Please add an integration test for checking the proof hints output for some execution.

tothtamas28 (commented):
I've left a few comments for the current approach (i.e. (3) from the list). However, I personally think (1) is a better solution, which you can implement as follows:

```python3
def run_proof_hint(...) -> bytes:
    args = _build_arg_list(...)
    ...
    proc_res = subprocess.run(args, text=False, stdout=subprocess.PIPE, ...)
    return proc_res.stdout
```

This is easy to implement and test, gives you a better interface, and does not prevent us from dropping `run_process` later.

Robertorosmaninho:
> I've left a few comments for the current approach (i.e. (3) from the list). However, I personally think (1) is a better solution, which you can implement as follows:
> 
> ```python
> def run_proof_hint(...) -> bytes:
>     args = _build_arg_list(...)
>     ...
>     proc_res = subprocess.run(args, text=False, stdout=subprocess.PIPE, ...)
>     return proc_res.stdout
> ```
> 
> This is easy to implement and test, gives you a better interface, and does not prevent us from dropping `run_process` later.

The problem with that approach is that we have to have almost all logic duplicated just to get a different way to output the program instead of only modifying the function that outputs the result itself. I can understand that this makes things more coupled right now for a possible future refactoring, but I'm sure that in future refactoring, we'll not need to have hundreds of code duplicated.

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: This isn't a comment about your option 3.

tothtamas28: What I wrote:

> I've left a few comments for the current approach (i.e. (3) from the list).

What I meant: either go with the current approach (option 3) and fix the following, or do option 1.

tothtamas28:
> The problem with that approach is that we have to have almost all logic duplicated

I don't think there's a lot to duplicate honestly, the interesting part is `_build_arg_list`.

> just to get a different way to output the program

That's a significant difference in my opinion.

> instead of only modifying the function that outputs the result itself.

It'd be awesome if we could do `stdout=subprocess.PIPE` by only modifying the function that outputs the result, but we can't unfortunately. Hence option 1 as a suggestion. 

> I can understand that this makes things more coupled right now for a possible future refactoring, but I'm sure that in future refactoring, we'll not need to have hundreds of code duplicated.

That's my point, I'd rather have some code duplication than dependency to deprecated code. I'm going to be able deal with `run_process` even if you introduce a call to it now, but it's just easier to make these considerations before publishing new API.

Baltoli:
My 2c is that this is a perfectly legitimate duplication to introduce. The proof hints pipeline behaves very differently to the other ways that krun works, so it's legitimate to fully separate their python entrypoints.

Robertorosmaninho:
@tothtamas28, can you please take a new look at this?

pyk/src/pyk/ktool/krun.py
```diff
@@ -36,6 +36,7 @@ class KRunOutput(Enum):
     LATEX = 'latex'
     KORE = 'kore'
     NONE = 'none'
+    HINTS = 'hints'
```
tothtamas28: Now that there's `run_proof_hint_process`, this seems unnecessary.

pyk/src/pyk/ktool/krun.py
```diff
@@ -115,7 +160,25 @@ def run(
         pipe_stderr: bool = True,
         bug_report: BugReport | None = None,
         debugger: bool = False,
+        proof_hint: bool = False,
```
tothtamas28: This as well.

tothtamas28 (commented):
Please add an integration test for `run_proof_hint`.

pyk/src/pyk/ktool/krun.py
```diff
@@ -205,9 +270,11 @@ def _krun(
     # ---
     check: bool = True,
     pipe_stderr: bool = True,
+    pipe_stdout: bool = True,
     logger: Logger | None = None,
     bug_report: BugReport | None = None,
     debugger: bool = False,
+    proof_hint: bool = False,
```
tothtamas28: Since there's `run_proof_hint_process`, for the sake of simplicity, it'd better not to expose this functionality in `_krun`.

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: I created it here so we can use the print pipeline from `run` as
```Python
case KRunOutput.HINTS:
      print(hints)
```

Robertorosmaninho: But I can use any flag if you prefer

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: The idea to expose `--proof-hints` in PyK is to enable it for KWASM and KEVM. KWASM uses `_krun` see [here](https://github.com/runtimeverification/wasm-semantics/blob/c4fc79c684ca9007547a74d768a4e743e3f1431f/pykwasm/src/pykwasm/scripts/kwasm.py#L48) while KEVM uses `Krun.run` see [here](https://github.com/runtimeverification/evm-semantics/blob/25be7669a6d42341d563cf53354ba0bbd29d9372/kevm-pyk/src/kevm_pyk/__main__.py#L593).

Robertorosmaninho: Therefore, I expect to see these parameters exposed so we can set them on the semantics caller.  Unless there's a simpler way to do it that I'm not aware of. In this case, I would be happy to adapt the code to use it!

pyk/src/pyk/ktool/krun.py
tothtamas28: I prefer no new value. Any new value breaks `_krun`.

pyk/src/pyk/ktool/krun.py
tothtamas28: There's no need to extend behavior here, you'll need to add new parameters / add a new command in KWASM / KEVM (whatever the maintainers there prefer) anyways.

pyk/src/pyk/utils.py
```diff
@@ -499,6 +499,57 @@ def run_process_2(
     return res
 
 
+def run_proof_hint_process(
```
tothtamas28: Please move this to `krun.py`, and make it private.

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Looking to the function calls on the links of previous message, can you point to me how krun would be aware that I want to output Hints?

If not by exposing this variable I can only think in passing the output type as HINTS.

tothtamas28: > can you point to me how krun would be aware that I want to output Hints

By calling `KRun.run_proof_hint`, and not `_krun` / `KRun.run` in the handler. The handlers get their arguments from a custom command line layer, so you'll need to add a new command or a new option for proof hints.

Robertorosmaninho: Alright then.

Robertorosmaninho:
@tothtamas28 It Wouldn't be good to already support `--proof-hint` in `pyk run`?
Using your description is no way to access the proof hints feature outside a semantic. Isn't the idea of `pyk run` to work as `krun`? So why can't I take this opportunity to already support it?

tothtamas28:
> It Wouldn't be good to already support `--proof-hint` in `pyk run`?  Isn't the idea of `pyk run` to work as `krun`? So why can't I take this opportunity to already support it?

I'd like to keep these two APIs separate because
* These are separate use cases: proof hint generation vs. regular concrete interpretation.
* They have separate users: Pi2 vs regular K users.
* One (i.e. the proof hints) is more important to maintain backwards compatibility for than the other.

tothtamas28:
> Using your description is no way to access the proof hints feature outside a semantic.

I don't see why that'd be the case: you can use the proof hints feature by calling `run_proof_hint`. You can also build a CLI for it if you'd like to.

pyk/src/pyk/utils.py
Robertorosmaninho: Done

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Done

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Deleted

Robertorosmaninho:
Hey @tothtamas28, I believe I've addressed all your comments. Can you please take a new look at it?

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Deleted

pyk/src/pyk/ktool/krun.py
```diff
@@ -115,6 +217,7 @@ def run(
         pipe_stderr: bool = True,
         bug_report: BugReport | None = None,
         debugger: bool = False,
+        proof_hint: bool = False,
```
tothtamas28: Please remove this parameter.

pyk/src/pyk/ktool/krun.py
```diff
@@ -205,9 +308,11 @@ def _krun(
     # ---
     check: bool = True,
     pipe_stderr: bool = True,
+    pipe_stdout: bool = True,
     logger: Logger | None = None,
     bug_report: BugReport | None = None,
     debugger: bool = False,
+    proof_hint: bool = False,
```
tothtamas28: Please remove this parameter.

pyk/src/pyk/ktool/krun.py
```diff
@@ -265,6 +377,7 @@ def _build_arg_list(
     search_final: bool,
     no_pattern: bool,
     debugger: bool,
+    proof_hint: bool = False,
```
tothtamas28:
```suggestion
    proof_hint: bool,
```

pyk/src/tests/integration/test_krun_proof_hints.py
```diff
@@ -0,0 +1,87 @@
...
+
+import pyk.kllvm.hints.prooftrace as prooftrace
+from pyk.kore.parser import KoreParser
+from pyk.testing import KRunTest
+
+if TYPE_CHECKING:
+    from pyk.ktool.krun import KRun
+
+
+class Test0Decrement(KRunTest):
```
tothtamas28: Consider

```suggestion
class Test0Decrement(ProofTraceTest, KRunTest):
```

Then all the new additions to `_kompiler.py` become unnecessary I think.

Robertorosmaninho: Good point! Thanks!

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: If I don't set `False` by default here, MyPy complains about the `_build_arg_list` construction inside `_krun`. So, the options here, as far as I understand, are: 
- Keep the default ` proof_hint: bool = False,`
- Add this field to `_build_arg_list` call inside `_krun` as ` proof_hint = False)`

Robertorosmaninho: Which option do you prefer?

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Done

pyk/src/pyk/ktool/krun.py
Robertorosmaninho: Done

pyk/src/pyk/ktool/krun.py
tothtamas28: Making the argument explicit on call site.

Robertorosmaninho: Done!

tothtamas28 (approved):

PR #4533, opened 2024-07-19
Further fine-tune `Int2Bytes(Int, Endianness, Signedness)` equations for booster

jberthold:
The previous change #4522 was insufficient to make the equations usable in `booster`. Matches with literal `Int` constants will still be indeterminate, so this PR removes the literals from the rule LHSs.

* Use `0 ==Int I` instead of a syntactic _match_  in the equation for argument `0`. This enables evaluating terms with `I ==Int 0` in the path condition.
* Remove equation with literal `-1` and widen requires in equation for negative arguments (previously `I <Int -1`, now `I <Int 0`). The result is the same as before with no extra equation required.


jberthold:
NB Another, independent, issue is that the function is marked `total` but `Int2Bytes(I, _, Unsigned)` with `I <Int 0` does not match any of the equations.

Baltoli (approved):

PR #4537, opened 2024-07-20
Test macos build

Baltoli:


Baltoli:
Closing as it wouldn't provide the speedup we're looking for; the macOS public runners take nearly an hour to run the K test suite.

PR #4541, opened 2024-07-22
Merge node

### LINKED ISSUE #4425: Node merging to reduce branching for CSE

Stevengre:
Blocked on: https://github.com/runtimeverification/k/pull/4617
Blocked on: https://github.com/runtimeverification/k/pull/4621
Blocked on: https://github.com/runtimeverification/k/pull/4612

This pull request addresses issue #4425 and refactors #4521 for simplification and ease of review.

- The key implementations are in `kcfg.py` and `semantics.py`.
- Unit tests are located in `pyk/src/tests/unit/kcfg/test_merge_nodes.py`.
- Changes in other Python files are either adaptations for Edge or result from the rebase. I am uncertain how to eliminate these changes caused by rebase.


Baltoli:
@Stevengre you'll want to use `git rebase -i develop` to interactively select the commits that contain your changes, and to drop the ones that don't belong to this PR. This is a useful git workflow to know how to do in general - there's a good guide [here](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) if you're not familiar.

Stevengre:
Thank you so much! I'll try it!

ehildenb:
Subsumed by: https://github.com/runtimeverification/k/pull/4628

PR #4560, opened 2024-07-29
Add extra check for lldb path in krun --debugger

gtrepta:
When invoking `krun --debugger` on MacOS, we check a file that's supposed to specify where the lldb we use is installed on the system.

Sometimes the contents of this file are `LLDB-NOTFOUND` and we're supposed to fall back to whatever lldb is on the user's PATH. There's a bug where we don't handle this case properly, and this change fixes that.


Robertorosmaninho (approved):
Lgtm! I've tested it in my local build!

PR #4561, opened 2024-07-29
Status bar support in proof loop

nwatson22:
Adds a couple things needed for supporting a proof status bar for real-time updates in kevm/kontrol: Adds a callback parameter to the two proof loops to get information about the proofs during execution of the proof loop (without having to read from disk). Also adds a field to `APRProof` which is the one-line summary which will be printed in the status bar, along with a default implementation in `Proof`.

### DIFF
--- a/pyk/src/pyk/proof/proof.py
+++ b/pyk/src/pyk/proof/proof.py
@@ -281,7 +281,7 @@ class Proof(Generic[PS, SR]):
 
     @property
     def one_line_summary(self) -> str:
-        return f'{self.status.name}'
+        return self.status.name
 
     @abstractmethod
     def get_steps(self) -> Iterable[PS]:
--- a/pyk/src/pyk/proof/proof.py
+++ b/pyk/src/pyk/proof/proof.py
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
     from pathlib import Path
     from typing import Any, Final
 
+    from rich.progress import Progress, TaskID
+
     T = TypeVar('T', bound='Proof')
 
 P = TypeVar('P', bound='Proof')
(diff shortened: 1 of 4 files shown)


pyk/src/pyk/proof/proof.py
```diff
@@ -279,6 +279,10 @@ def lines(self) -> list[str]:
         subproofs_summaries = [subproof.summary for subproof in self.subproofs]
         return CompositeSummary([BaseSummary(self.id, self.status), *subproofs_summaries])
 
+    @property
+    def one_line_summary(self) -> str:
+        return f'{self.status.name}'
```
anvacaru: Why not just `return self.status.name`?

anvacaru (approved):
Approved with a question.

pyk/src/pyk/proof/reachability.py
```diff
@@ -502,6 +502,13 @@ def summary(self) -> CompositeSummary:
             ]
         )
 
+    @property
+    def one_line_summary(self) -> str:
+        return (
+            super().one_line_summary
+            + f'/{len(self.kcfg.nodes)} nodes/{len(self.pending)} pending/{len(self.failing)} failing/{len(self.kcfg.ndbranches()) + len(self.kcfg.splits())} branches/{len(self.kcfg.vacuous)} vacuous/{len(self.kcfg.stuck)} stuck'
```
PetarMax: I think that we could also add the number of successfully completed executions (nodes that have a cover into target).

pyk/src/pyk/proof/proof.py
nwatson22: Fixed.

pyk/src/pyk/proof/reachability.py
nwatson22: Added a `passed` section that does this.

PR #4566, opened 2024-08-01
Display number of terminal nodes in `one_line_summary` for `APRProof`

nwatson22:
Adds the number of terminal nodes to `one_line_summary` for `APRProof`. Also changes the display to use `|` as dividers instead of `/`.


anvacaru (approved):
Approved! Please add a brief description to the PR!

Issue #1126, opened 2020-02-26
LLVM Backend crashes


jzw2:
I'm not sure if this is because I broke my machine or something, but when I try to kompile and krun sum.imp from the k tutorial, I get 

``` [Error] Critical: LLVM Backend crashed during rewriting. ```

I'm running on a mac, I'm not sure what other variables there are that would cause this.

dwightguth:
I'm aware of this issue; it seems to only be affecting macs on recent versions of k. I would suggest installing the latest Homebrew bottle in the meantime.

dwightguth:
Having investigated further, I believe this issue may be fixed with kframework/llvm-backend#303 once the submodule in kframework/k is updated. I'm not 100% sure this is the same issue since I have not independently confirmed it, but it seems to manifest the same way as the bug I discovered, which was a bug caused by the recent release of GMP 6.2, which is also recently released on Mac OS.

As such, if this is the cause of a bug, older versions of K will not fix the problem (unlike my previous suggestion), because it is not so much a regression as a latent bug. You can try to roll back to GMP 6.1.2 and see if that solves it, or you can try building from source and pinning the llvm backend submodule to the branch in the PR I mentioned above, or you can wait and one way or another I will have this fixed soon.

dwightguth:
I believe this should now be fixed. Please let me know if it is still occurring for you.

vasil-sd:
I'm suffering from similar bug, using latest release of K-Tools (v5.0.0-51b354c).

After short investigation, I've found that it crashes on NULL ptr deref in getTagForSymbolName.

(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffffd730801 in __GI_abort () at abort.c:79
#2  0x00000000002127a1 in getTagForSymbolName ()
#3  0x000000000021ba72 in allocatePatternAsConfiguration (Pattern=<optimized out>) at /home/jenkins-slave/workspace/k_master/llvm-backend/src/main/native/llvm-backend/runtime/configurationparser/ConfigurationParser.cpp:34

dwightguth:
[...]

PR #4567, opened 2024-08-01
Handle `KLLVM.StringBuffer.contents` as `bytes` and enable `unicode` tests in `String` contructors

Robertorosmaninho:
Blocked on [#1126](https://github.com/runtimeverification/llvm-backend/pull/1126).
Related to: 
- https://github.com/runtimeverification/llvm-backend/issues/824
- https://github.com/runtimeverification/llvm-backend/issues/1078
- https://github.com/Pi-Squared-Inc/pi2/issues/1683
- https://github.com/Pi-Squared-Inc/pi2/issues/1688

This PR aims to support the new feature introduced in [#1126](https://github.com/runtimeverification/llvm-backend/pull/1126) that makes `KLLVM.StringBuffer.contents` returns `bytes`. To ensure the symmetric behavior, we should also construct the `KLLVM.StringBuffer` object with bytes. To ensure this feature works properly, we're using `raw_unicode_escape` encoding to pass and retrieve the result.

This can be used as a first step in a new attempt to support Unicode!

### DIFF
--- a/pyk/src/tests/integration/test_string.py
+++ b/pyk/src/tests/integration/test_string.py
@@ -40,6 +40,12 @@ TEST_DATA: Final = (
     '🙂',
 )
 
+SKIPPED_BINDINGS_TESTS: Final = {
+    '𐍈',
+    '武天老師',
+    '🙂',
+}
+
 KOMPILE_DEFINITION = """
     module STRING-REWRITE
         imports STRING-SYNTAX
@@ -216,6 +222,9 @@ def test_krun(backend: str, definition_dir: Path, text: str) -> None:
 @pytest.mark.parametrize('text', TEST_DATA, ids=TEST_DATA)
 def test_bindings(runtime: Runtime, text: str) -> None:
     from pyk.kllvm.convert import llvm_to_pattern, pattern_to_llvm
+
+    if text in SKIPPED_BINDINGS_TESTS:
+        pytest.skip()
 
     # Given
     kore = kore_config(text, '')
--- a/pyk/src/pyk/kllvm/convert.py
+++ b/pyk/src/pyk/kllvm/convert.py
@@ -108,7 +108,7 @@ def sentence_to_llvm(sentence: Sentence) -> kllvm.Declaration:
 def pattern_to_llvm(pattern: Pattern) -> kllvm.Pattern:
     match pattern:
         case String(value):
-            return kllvm.StringPattern(value)
+            return kllvm.StringPattern(value.encode('latin-1'))
         case VarPattern(name, sort):
             return kllvm.VariablePattern(name, sort_to_llvm(sort))
         case App(symbol, sorts, args):
@@ -201,7 +201,7 @@ def llvm_to_sentence(decl: kllvm.Declaration) -> Sentence:
 def llvm_to_pattern(pattern: kllvm.Pattern) -> Pattern:
     match pattern:
         case kllvm.StringPattern():  # type: ignore
-            return String(pattern.contents)
+            return String(pattern.contents.decode('latin-1'))
         case kllvm.VariablePattern():  # type: ignore
             if pattern.name and pattern.name[0] == '@':
                 return SVar(pattern.name, llvm_to_sort(pattern.sort))
--- a/pyk/src/tests/integration/kllvm/test_patterns.py
+++ b/pyk/src/tests/integration/kllvm/test_patterns.py
@@ -50,7 +50,7 @@ def test_string() -> None:
 
     # Then
     assert str(pattern) == '"abc"'
-    assert pattern.contents == 'abc'
+    assert pattern.contents.decode('latin-1') == 'abc'
 
 
 def test_variable() -> None:


Robertorosmaninho:
These commits will be squashed and committed to #4565 as they block that PR and depend on the changes there. However, as this changes important parts of PyK, I would like your thoughts about it @tothtamas28, @gtrepta, and @Scott-Guest. 

I've tested it also on KEVM, and everything looks good there

pyk/src/tests/integration/kllvm/test_convert.py
```diff
@@ -21,8 +21,8 @@ def _in_module(kore_text: str) -> str:
     ('string-escape', r'"\t\n\r\f\\\""'),
     ('string-$', '"$"'),
     (r'string-\x24', r'"a\x24b"'),
-    # (r'string-\u03b1', r'"a\u03b1b"'),
-    # (r'string-\U0001F642', r'"a\U0001F642b"'),
+    (r'string-\u03b1', r'"a\u03b1b"'),
+    (r'string-\U0001F642', r'"a\U0001F642b"'),
```
tothtamas28: As far as I understand, it should be possible to enable these failing tests without changing how pyk internally represents `SortString` and `SortBytes` domain values, and only introduce changes to `pattern_to_llvm` and `llvm_to_pattern`.

If this is indeed the case, I'd prefer not to break the behavior of these functions (i.e. `String.init`, `String.let`, `String.write`, `enquote_bytes`, `stringToken`, etc).

@Scott-Guest, what do you think?

Robertorosmaninho: At first glance, unfortunately, this isn't possible:
```bash
FAILED src/tests/integration/kllvm/test_convert.py::test_pattern_to_llvm[string-\\U0001F642] - AssertionError: assert String(value='a\\U0001f642b') == String(value='a🙂b')
FAILED src/tests/integration/kllvm/test_convert.py::test_pattern_to_llvm[string-\\u03b1] - AssertionError: assert String(value='a\\u03b1b') == String(value='aαb')
```

tothtamas28:
```python3
>>> from pyk.kore.parser import KoreParser
>>> KoreParser(r'"a\u03b1b"').pattern()
String(value='aαb')
```

In general, changing the pyk encoding is only necessary if there's a domain value that we cannot represent in the current one. If there's such a value, let's add a unit test for it and debug it further (this can even go in a separate PR).

> ```
> FAILED src/tests/integration/kllvm/test_convert.py::test_pattern_to_llvm[string-\\u03b1] - AssertionError: assert String(value='a\\u03b1b') == String(value='aαb')
> ```

It looks like this can be fixed in `llvm_to_pattern` or `pattern_to_llvm`, as `r'\u03b1'` and `'α'` (i.e. `'\u03b1'`) are just two different encodings of the same thing. To make these tests pass, you can try inlining your changes in these two functions to make them local to the bindings interface.

Robertorosmaninho: The problem is that `pattern_to_llvm` returns the `kllvm.StringPattern`, which is converted to string using `str` in the test. All these are trying to enforce `utf-8`, if I let this encoding instead of `raw_unicode_escape` as did, we get things like: `AssertionError: assert String(value='aÎ±b') == String(value='aαb')`

Robertorosmaninho: This implementation was the least intrusive I could came up with, and after testing it everywhere, only `lexical.k` needed to be changed to really escape the Unicode created there "accidentally."

Robertorosmaninho: I understand that we would like to have this change atomic and only modify the binding parts, but as we're modifying the encoding on one side, we must prepare the other to receive it as we want to.

tothtamas28: > The problem is that `pattern_to_llvm` returns the `kllvm.StringPattern`, which is converted to string using `str` in the test.

If there has been a change in how `kllvm.StringPattern` is encoded, modifying these tests might be appropriate, and it's definitely favorable to introducing breaking changes to `pyk.kore.syntax.String`. The test calls `__str__` as until now the easiest way to compare these objects for equality was through the textual KORE format.

> we must prepare the other to receive it as we want to

I do not see why. Has there been a change in what we consider a valid [KORE string pattern](https://github.com/runtimeverification/haskell-backend/blob/master/docs/kore-syntax.md#string-literals)? If it's just the internal representation of the LLVM backend that changed, there's no reason to propagate it through the whole stack.

Robertorosmaninho: In some sense, we are accepting Unicode characters that we didn't before, so this should already be a valid reason for this change. However, we are motivated by the issues listed in the description of the PR, especially by the second, and the main advantage of this approach is to keep the exact same behavior if no Unicode is given.

tothtamas28: Can you give an example of a domain value in textual KORE format that cannot be faithfully represented in pyk?

Robertorosmaninho: Yes.
```
\dv{SortStringBuffer{}}("\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00R\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0a\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 R[\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 Q\x15\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x9b")
```
fails with:

```
File "/Users/mirceasebe/Library/Caches/pypoetry/virtualenvs/proof-generation-YL8vN-tV-py3.10/lib/python3.10/site-packages/pyk/kllvm/convert.py", line 204, in llvm_to_pattern
    return String(pattern.contents)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9b in position 202: invalid start byte
```

tothtamas28: Consider the following snippet.

```python3
from pyk.kore.parser import KoreParser
from pyk.dequote import bytes_encode
kore = r'"\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00R\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0a\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 R[\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 Q\x15\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x9b"'
value = KoreParser(kore).string().value
print(bytes_encode(value))
# out:  b'\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00R\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\n\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 R[\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 Q\x15\x7f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x9b'
```

I aligned the bytes output so that it's easier to visually compare. The only difference is that in the input there's `\x0a` and in the output `\n`.

But I see how this raises an exception in `llvm_to_pattern`. My point is, it has to be fixed there and not let the breakage propagate.

Robertorosmaninho: And how would this work without propagating?

Remember: for usual cases, this change doesn't modify any behavior; for bytes/Unicode, it makes them pass as expected.

Robertorosmaninho: Also, please note that this is an urgent request from our Proof Generation team, which is blocked by this issue from processing hints from KEVM since last week, which is super important to us to get them asap.

dwightguth: My understanding is that if we tried to change it purely in pattern_to_llvm and llvm_to_pattern, there would be no way to prevent mojibake because we would have no way of distinguishing between the code point 0x80 and the byte 0x80

tothtamas28: > My understanding is that if we tried to change it purely in pattern_to_llvm and llvm_to_pattern, there would be no way to prevent mojibake because we would have no way of distinguishing between the code point 0x80 and the byte 0x80

Is this not dependent on the sort though? If it's `SortBytes`, you interpret it as a byte. If it's `SortString`, you interpret it as the code point. Otherwise, `'\'` is escaped. These two sorts are special cased on the Frontend, and pyk mirrors this behavior:

https://github.com/runtimeverification/k/blob/0b4e21041a75e42c003daff1091f91cb0de61519/k-frontend/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L2215-L2239

Would it be a solution to case split on these sorts in `convert.py`? I do not know how the proof hint generator interprets these values, but if neither the K token definitions in `domains.md`, nor the textual KORE format changes, my intuition tells me it shouldn't be necessary to introduce changes in pyk either.

@Scott-Guest, please advise.

tothtamas28: In the worst case, it should be possible to serialize the `pyk.kore.syntax.String` value to textual KORE, and let the LLVM backend bindings parse it. Otherwise, one of these formats does not correctly map to textual KORE.

Scott-Guest: Prior to this PR, the situation was that
- On the LLVM backend, 
    - The KORE string pattern contents are just a sequence of bytes
    - Escapes `\xHH` are replaced with the byte `0xHH`
    - Escapes `\u` and `\U` are [replaced with the UTF-8 bytes for the codepoint](https://github.com/runtimeverification/llvm-backend/blob/9cb2137d96059c8751bb57216cc1d0cea5746161/lib/parser/KOREScanner.l#L85).
    - `U+03B1` is UTF-8 encoded as the bytes `[0xCE, 0xB1]`, so `"\u03b1"` and `"\xce\xb1"` are exactly the same term of length 2.
- In Pyk and on the Haskell backend,
    - The KORE string pattern contents are a sequence of Unicode characters
    - Escapes `\xHH` are replaced with the Unicode codepoint `U+00HH`
        - This is why `bytes_decode` and `bytes_encode` are exactly `latin-1` decoding / encoding - `latin-1` replaces each byte `0xHH` with the Unicode codepoint `U+00HH`, so in effect replaces each byte `0xHH` with the KORE string pattern `\xHH` as desired.
    - Escapes `\u` and `\U` are also replaced with their Unicode codepoint
    - `"\u03b1"` has length 1, `"\xce\xb1"` has length 2, and these terms are not the same.

Because of this difference, until we more fundamentally change the LLVM backend's representation of Strings to make them Unicode-aware, the Pyk-LLVM round-trip tests _should_ fail for Unicode values. 

Using `raw_unicode_escape` treats `"\u03b1"` literally as the sequence of characters `['\', 'u', '0', '0', '3', 'b', '1']`, i.e., `"\x5c\x75\x30\x33\x62\x31"`, which survives the round-trip, but isn't the behavior we actually want - `"\u03b1"` and `"\x5c\x75\x30\x33\x62\x31"` should be different terms.

Instead, all we really need to do is treat the LLVM backend strings as sequences of bytes in the same way we would represent a `Bytes` DV, i.e., `latin-1` decoding and encoding to produce the KORE string pattern. 

Explicitly, 
- In `llvm_to_pattern` , `latin-1` decode the `bytes` from `StringPattern.contents` to produce the corresponding `str` for `String.value`
- In `pattern_to_llvm`, `latin-1` encode the `str` from `String.value` to produce the corresponding `bytes` for `StringPattern.contents`

Robertorosmaninho: That makes sense. But with this `latin-1` encoding, we have to mark 3 tests with expected to fail:

```
FAILED src/tests/integration/test_string.py::test_bindings[\U00010348] - UnicodeEncodeError: 'latin-1' codec can't encode character '\U00010348' in position 0: ordinal not in range(256)
FAILED src/tests/integration/test_string.py::test_bindings[\U0001f642] - UnicodeEncodeError: 'latin-1' codec can't encode character '\U0001f642' in position 0: ordinal not in range(256)
FAILED src/tests/integration/test_string.py::test_bindings[\u6b66\u5929\u8001\u5e2b] - UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-3: ordinal not in range(256)
```

```Python
TEST_DATA: Final = (
    'hello', # passs
    '\n', # pass
    '\x01', # pass
    '£', # pass
    '𐍈', # fail with latin-1
    '武天老師',  # fail with latin-1
    '🙂', # fail with latin-1
)
```

If you're okay with this, I see no problem with modifying the PR. So far, we don't have a strong need for Unicode support.

Robertorosmaninho: Please advice @Scott-Guest and @tothtamas28

tothtamas28: Works for me.

Scott-Guest: Yup, that should work! Was a bit of a fluke that those tests passed before without us actually having Unicode support.

Robertorosmaninho:
@tothtamas28 and @Scott-Guest, can you please review the PR again? If you approve it, I'll cherry-pick the commit and put it in https://github.com/runtimeverification/k/pull/4565 so the tests pass as expected there, and we can merge both changes. 
The modification that makes `StringPattern.contents` return bytes is there and it's blocked by this change.

pyk/src/tests/integration/test_string.py
```diff
@@ -40,6 +40,13 @@
     '🙂',
 )
 
+BINDINGS_TEST_DATA: Final = (
+    'hello',
+    '\n',
+    '\x01',
+    '£',
+)
```
tothtamas28: Consider marking the opposite:

```python3
SKIPPED_BINDINGS_TESTS: Final = {
    '𐍈',
    '武天老師',
    '🙂',
}
```

pyk/src/tests/integration/test_string.py
```diff
@@ -213,7 +220,7 @@ def test_krun(backend: str, definition_dir: Path, text: str) -> None:
     assert actual == expected
 
 
-@pytest.mark.parametrize('text', TEST_DATA, ids=TEST_DATA)
+@pytest.mark.parametrize('text', BINDINGS_TEST_DATA, ids=BINDINGS_TEST_DATA)
```
tothtamas28: Then here:

```python3
if text in SKIPPED_BINDINGS_TESTS:
    pytest.skip()
```

tothtamas28 (approved):

pyk/src/tests/integration/test_string.py
Robertorosmaninho: Done! Thanks!

Scott-Guest (approved):

Robertorosmaninho:
Closing as this modification was merged with https://github.com/runtimeverification/k/pull/4565

PR #4575, opened 2024-08-06
Status bar improvements

nwatson22:
- Adds the number of refuted nodes to the proof status bar.
- Adds a separator which separates the one line summary into 3 parts, one which is the proof status, one to do with structure, and one to do with status of the leaves.
- Changes the definition of `branches` to mean "how many `Split` and `NDBranch` targets are there?" instead of "how many `Split`s and `NDBranch`es are there?" to more accurately  reflect the degree of branching in a proof.


PetarMax:
Great - could we please have a slightly different order, say:
```
pending-passed-failing-vacuous-refuted-stuck-terminal || (or some other separator) nodes-branches
```
Also, note that `branches` might not carry all of the information, since, for example, in minimised proofs I will likely have just one branching that goes many ways.

PetarMax (approved):

PR #4577, opened 2024-08-07
Removing minimize_constraints and adding ability to use booster implies

### LINKED ISSUE #4570: Support  the `"assume-defined"` parameter of `"implies"`
The Kore RPC API's `"implies"` endpoint allows specifying the `"assume-defined"` request [parameter](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md#request-1)  that, when set to `true`, allows handling the `"implies"` request in Booster. `pyk` should allow passing this request option.

PetarMax:
This PR improves on the handling and use of the `implies` endpoint in the following two ways:
- it allows the use of the Booster implies endpoint
- it removes constraint minimization, which calls the implies endpoint and was originally designed to streamline the path condition, but in real-world cases only introduced an overhead.


pyk/src/pyk/kore/rpc.py
```diff
@@ -1027,12 +1027,14 @@ def implies(
...
         module_name: str | None = None,
+        booster_implies: bool = False,
     ) -> ImpliesResult:
         params = filter_none(
             {
                 'antecedent': self._state(antecedent),
                 'consequent': self._state(consequent),
                 'module': module_name,
+                'assume-defined': booster_implies,
             }
```
ehildenb: Why are we changing the name from `assume-defined` to `booster_implies`? I guess I would tend towards that we should just pass the same option name through unless there is a good reason. A good reason would be that `booster_implies`, as a variable, will be used to control some other behavioral change _not_ to do with assuming definedness of the input. But I guess that isn't the case here.

PetarMax: Yes, so, our goal is to use the booster `implies` end-point, hence the name I chose - this is effectively what happens when that options is passed in as `True`. 

The corresponding option in the rpc server/backend seems to be, from what @geo2a tells me, `assume-defined`. This, for me, is not indicative of the effect - assume *what* is defined and what does that mean? So I chose the other, more informative name.

PetarMax: I've renamed it to `assume_defined`.

geo2a:
The change looks good to me. Before merging, consider evaluating the performance impact on KEVM and Kontrol. In the HB repo, we have Nix-based scripts for doing so, here's the KEVM one: https://github.com/runtimeverification/haskell-backend/blob/master/scripts/performance-tests-kevm.sh

geo2a (approved):
Approving, as I do not expect these changes to negatively affect performance.

PR #4580, opened 2024-08-08
Redesigning rewrite logging

PetarMax:
This PR adds the option to turn on/off the logging of both successful and failed rewrites. While the latter was already present, the former was observed to introduce an execution speedup in engagement proofs, and is therefore introduced.

As part of this PR, the `trace_rewrites` is renamed into `log_fail_rewrites`, which is both aligned with `log_succ_rewrites` and describes its purpose better.


ehildenb:
The reason to have rewrite success logging turned on was to make sure that we had a list of `rules` to attach to the `KCFG.Edge`. I guess this can be useful for proof objects and debugging. I guess we just need to make sure that having that option turned off doesn't cause it to crash... Here is where we extract the rule labels from the logs: https://github.com/runtimeverification/k/blob/7dfd927dc1a7a590576303f8527ed2a68f4b3691/pyk/src/pyk/kcfg/explore.py#L246. It would be good to check that this code won't crash if there are no rule labels present.

PetarMax:
> The reason to have rewrite success logging turned on was to make sure that we had a list of `rules` to attach to the `KCFG.Edge`. I guess this can be useful for proof objects and debugging. I guess we just need to make sure that having that option turned off doesn't cause it to crash... Here is where we extract the rule labels from the logs:
> 
> https://github.com/runtimeverification/k/blob/7dfd927dc1a7a590576303f8527ed2a68f4b3691/pyk/src/pyk/kcfg/explore.py#L246
> . It would be good to check that this code won't crash if there are no rule labels present.

I have been running a custom `pyk` with the logging turned off for the last two weeks in Kontrol engagement proofs, haven't spotted any issues.

ehildenb (approved):

PR #4581, opened 2024-08-08
Bug not allowing `--smt-retry-limit 0`

nwatson22:
We were checking `if self.smt_retry_limit` to see if the `--smt-retry-limit` option should be passed to `kore-rpc-booster`. This will evaluate to false when it's 0, so the option will not be passed. However there are cases where we want to pass `--smt-retry-limit 0` to the kore server. This fixes the check to be `if self.smt_retry_limit is not None` so it will only not fire if the option is not set. Also fixes this issue for other KoreServer args.

Also fixes the validation of arguments not being done of the base `KoreServer` args when using `BoosterServer`. In both cases, changes `check_if_positive` to `check_if_nonnegative` for `smt_retry_limit` to allow passing 0.

I confirmed this allows `--smt-retry-limit` to make it through to the `kore-rpc-booster` command when passed to `kontrol prove`.


pyk/src/pyk/kore/rpc.py
```diff
@@ -1227,10 +1227,14 @@ def _check_none_or_positive(n: int | None, param_name: str) -> None:
             if n is not None and n <= 0:
                 raise ValueError(f'Expected positive integer for: {param_name}, got: {n}')
 
+        def _check_none_or_nonnegative(n: int | None, param_name: str) -> None:
+            if n is not None and n < 0:
+                raise ValueError(f'Expected positive integer for: {param_name}, got: {n}')
```
PetarMax:
```suggestion
                raise ValueError(f'Expected non-negative integer for: {param_name}, got: {n}')
```

PetarMax (approved):

Issue #4481, opened 2024-06-25
Make use of parsing caches in kdist


dwightguth:
One thing I noticed recently while working on making modifications to the WASM semantics is that kdist appears to be doing something that effectively clobbers `kompile`'s parsing cache each time a rule is changed. compilation times via `kdist build` might be improved rather dramatically if the code were tweaked so that incremental parsing was enabled.

gtrepta:
https://github.com/runtimeverification/k/blob/7485b9660c628041680e012d38dba3fdeca2ec90/pyk/src/pyk/kdist/_kdist.py#L148-L150

This "cleanup" of the directory when the target is built should probably just be removed and then `kdist clean` can be relied on when a clean build is needed.

gtrepta:
I tested out just removing the `shutil.rmtree` call and it looks like it will break the wasm semantics which does a `shutil.copytree` in one of its build targets and relies on the destination folder not being there. So, I will need to go through everything downstream to make sure this change can be made.

PR #4583, opened 2024-08-12
--no-clean option for kdist build

gtrepta:
fixes: #4481 

By default, `kdist` will delete the output folder of a target before building it. This adds a `--no-clean` option to `kdist build` which will be useful for keeping the parse cache of a kompiled semantics when rebuilding.


pyk/src/pyk/kdist/__main__.py
```diff
@@ -48,13 +48,15 @@ def _exec_build(
     force: bool,
     verbose: bool,
     debug: bool,
+    clean: bool,
```
tothtamas28, line 51: I wonder, would it make sense to make `clean=False` behavior the only option? After all, if a build step requires an empty output directory, it can ensure that as the first step of the build.

gtrepta: This was my original thought. However, many downstream projects are depending on `clean=True` behavior because they use a `SourceTarget` which does `shutil.copytree`, and that call throws an error if the destination directory already exists. We can go through the downstream projects and fix that, but this change is a middle ground that we can get in for now to the people who requested it.

tothtamas28: > this change is a middle ground that we can get in for now to the people who requested it

Can you confirm if this solution is acceptable for them? If so, I fine with this workaround.

gtrepta: I confirmed that this will work.

tothtamas28 (approved):

PR #4585, opened 2024-08-13
APRProof get_steps optimizations

nwatson22:
- Calling `nonzero_depth` is (relatively) expensive, when it's done for each pending nodes and we have 50+ pending nodes. This is because it traverses the KCFG to find all paths between the init node and the current node.
- We were calling `nonzero_depth` twice unnecessarily. Caches the result in a variable instead.
- In most cases, depth will be immediately obvious to be nonzero just from looking at the first edge on the path. In this case, skips the traversal. Also skips the traversal if the nodes we are checking the if the depth between is 0 for if they are the same node.
- Tested on `test%kontrol%VetoSignallingTest.testVetoSignallingInvariantsArePreserved4(uint256,uint256,uint256)` from the lido tests. Around 400 nodes in, the sync time is 0.3-0.4 seconds without the optimization. With the optimization it is around 0.015 to 0.04 seconds (although this also depends on the number of pending nodes at any given time). This seems to eliminates the bulk of the sync time when there are a large number of pending nodes (~30+).


pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1188,7 +1188,19 @@ def shortest_distance_between(self, node_1_id: NodeIdLike, node_2_id: NodeIdLike
...
     def zero_depth_between(self, node_1_id: NodeIdLike, node_2_id: NodeIdLike) -> bool:
-        shortest_distance = self.shortest_distance_between(node_1_id, node_2_id)
+        _node_1_id = self._resolve(node_1_id)
+        _node_2_id = self._resolve(node_2_id)
+        if _node_1_id == _node_2_id:
+            return True
+        # Short-circuit and don't run pathing algorithm if there is no 0 length path on the first step.
+        node_1_successors = self.successors(_node_1_id)
+        node_2_successors = self.successors(_node_2_id)
+        path_lengths = [self.path_length([successor]) for successor in node_1_successors + node_2_successors]
```
PetarMax: I think you could inline the `node_1_successors` definitions here?

nwatson22: Fixed.

PetarMax (approved):

PR #4589, opened 2024-08-14
Fixing indentation in `run_proof_hint`

Robertorosmaninho:
We missed in the first implementation the indentation that keeps the temporary file opened. This PR fixed it.

### DIFF
--- a/pyk/src/pyk/ktool/krun.py
+++ b/pyk/src/pyk/ktool/krun.py
@@ -126,29 +126,29 @@ class KRun(KPrint):
             pgm.write(ntf)
             ntf.flush()
 
-        args = _build_arg_list(
-            command='krun',
-            input_file=Path(ntf.name),
-            definition_dir=self.definition_dir,
-            output=output,
-            parser=parser,
-            depth=depth,
-            pmap=pmap,
-            cmap=cmap,
-            term=term,
-            temp_dir=temp_dir,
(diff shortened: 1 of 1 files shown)


pyk/src/pyk/ktool/krun.py
```diff
@@ -126,29 +126,29 @@ def run_proof_hint(
...
 
-        hints_bytes = self.__run_proof_hint_process(
-            args=args, check=check, pipe_stderr=pipe_stderr, logger=_LOGGER, exec_process=debugger
-        )
+            hints_bytes = self.__run_proof_hint_process(
+                args=args, check=check, pipe_stderr=pipe_stderr, logger=_LOGGER, exec_process=debugger
+            )
 
-        return hints_bytes.stdout
+            return hints_bytes.stdout
```
tothtamas28:
```suggestion
        return hints_bytes.stdout
```

tothtamas28 (commented):
Can you add a test / modify the existing test so that this issue is caught?

Robertorosmaninho:
> Can you add a test / modify the existing test so that this issue is caught?

I'm not sure if I can reproduce the error. Regardless of the indentation, the integration tests pass, the temporary file is created, and we're able to use it.

The error seems to happen only when calling the function from outside the project, in my example, calling it from the `kevm`. How to proceed?

tothtamas28 (approved):

PR #4590, opened 2024-08-14
Reusing backend information to construct KCFGs more efficiently

PetarMax:
This PR leverages information provided by the backend as part of an `execute` request to construct more than one KCFG element: specifically, vacuous nodes, stuck nodes, edges resulting from cut-nodes, branches, and non-deterministic branches.

In addition, it moves the associated printing to the KCFG class, allowing for more informative prints: for example, we can now print the target nodes of edges and branches, the absence of which (personally) was making debugging of large proofs slightly more difficult.

The PR also removes the `always_check_subsumption` flag, which was not used and was made obsolete by the subsumption logic in `step_proof`.

### DIFF
--- a/pyk/src/pyk/kcfg/explore.py
+++ b/pyk/src/pyk/kcfg/explore.py
@@ -218,19 +218,20 @@ class KCFGExplore:
         cut_point_rules: Iterable[str] = (),
         terminal_rules: Iterable[str] = (),
         module_name: str | None = None,
-    ) -> KCFGExtendResult:
+    ) -> list[KCFGExtendResult]:
+
         def log(message: str, *, warning: bool = False) -> None:
             _LOGGER.log(logging.WARNING if warning else logging.INFO, f'Extend result for {self.id}: {message}')
 
         custom_step_result = self.kcfg_semantics.custom_step(_cterm)
         if custom_step_result is not None:
             log(f'custom step node: {node_id}')
-            return custom_step_result
+            return [custom_step_result]
 
         abstract_cterm = self.kcfg_semantics.abstract_node(_cterm)
         if _cterm != abstract_cterm:
             log(f'abstraction node: {node_id}')
-            return Abstract(abstract_cterm)
+            return [Abstract(abstract_cterm)]
 
         cterm, next_states, depth, vacuous, next_node_logs = self.cterm_symbolic.execute(
             _cterm,
@@ -240,33 +241,35 @@ class KCFGExplore:
             module_name=module_name,
         )
 
+        extend_results: list[KCFGExtendResult] = []
+
         # Basic block
         if depth > 0:
             log(f'basic block at depth {depth}: {node_id}')
-            return Step(cterm, depth, next_node_logs, self._extract_rule_labels(next_node_logs))
+            extend_results.append(Step(cterm, depth, next_node_logs, self._extract_rule_labels(next_node_logs)))
 
         # Stuck or vacuous
         if not next_states:
             if vacuous:
                 log(f'vacuous node: {node_id}', warning=True)
-                return Vacuous()
-            log(f'stuck node: {node_id}')
-            return Stuck()
-
+                extend_results.append(Vacuous())
+            else:
+                log(f'stuck node: {node_id}')
+                extend_results.append(Stuck())
         # Cut rule
-        if len(next_states) == 1:
+        elif len(next_states) == 1:
             log(f'cut-rule basic block at depth {depth}: {node_id}')
-            return Step(
-                next_states[0].state,
-                1,
-                next_node_logs,
-                self._extract_rule_labels(next_node_logs),
-                cut=True,
+            extend_results.append(
+                Step(
+                    next_states[0].state,
+                    1,
+                    next_node_logs,
+                    self._extract_rule_labels(next_node_logs),
+                    cut=True,
+                )
             )
-
         # Branch
-        assert len(next_states) > 1
-        if all(branch_constraint for _, branch_constraint in next_states):
+        elif all(branch_constraint for _, branch_constraint in next_states):
             branch_preds = [flatten_label('#And', not_none(rule_predicate)) for _, rule_predicate in next_states]
             common_preds = list(
                 unique(
@@ -283,9 +286,11 @@ class KCFGExplore:
                 )
             constraint_strs = [self.pretty_print(ml_pred_to_bool(bc)) for bc in branches]
             log(f'{len(branches)} branches: {node_id} -> {constraint_strs}')
-            return Branch(branches)
+            extend_results.append(Branch(branches))
         else:
             # NDBranch
             log(f'{len(next_states)} non-deterministic branches: {node_id}')
             next_cterms = [cterm for cterm, _ in next_states]
-            return NDBranch(next_cterms, next_node_logs, self._extract_rule_labels(next_node_logs))
+            extend_results.append(NDBranch(next_cterms, next_node_logs, self._extract_rule_labels(next_node_logs)))
+
+        return extend_results
--- a/pyk/src/pyk/kcfg/explore.py
+++ b/pyk/src/pyk/kcfg/explore.py
@@ -250,7 +250,7 @@ class KCFGExplore:
                 extend_results.append(Stuck())
         # Cut rule
         elif len(next_states) == 1:
-            if not self.kcfg_semantics.can_make_custom_step(next_states[0].state):
+            if not self.kcfg_semantics.can_make_custom_step(cterm):
                 (next_node_logs, rules) = (
                     ((), []) if extend_results else (next_node_logs, self._extract_rule_labels(next_node_logs))
                 )
--- a/pyk/src/pyk/kcfg/explore.py
+++ b/pyk/src/pyk/kcfg/explore.py
@@ -240,7 +240,7 @@ class KCFGExplore:
 
         # Basic block
         if depth > 0:
-            extend_results.append((Step(cterm, depth, next_node_logs, self._extract_rule_labels(next_node_logs))))
+            extend_results.append(Step(cterm, depth, next_node_logs, self._extract_rule_labels(next_node_logs)))
 
         # Stuck or vacuous
         if not next_states:
--- a/pyk/src/pyk/kcfg/explore.py
+++ b/pyk/src/pyk/kcfg/explore.py
@@ -250,7 +250,7 @@ class KCFGExplore:
                 extend_results.append(Stuck())
         # Cut rule
         elif len(next_states) == 1:
-            if not self.kcfg_semantics.can_make_custom_step(cterm):
+            if not self.kcfg_semantics.can_make_custom_step(next_states[0]):
                 (next_node_logs, rules) = (
                     ((), []) if extend_results else (next_node_logs, self._extract_rule_labels(next_node_logs))
                 )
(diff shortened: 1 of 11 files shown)


pyk/src/pyk/kcfg/explore.py
```diff
@@ -218,19 +218,20 @@ def extend_cterm(
         cut_point_rules: Iterable[str] = (),
         terminal_rules: Iterable[str] = (),
         module_name: str | None = None,
-    ) -> KCFGExtendResult:
+    ) -> list[KCFGExtendResult]:
```
PetarMax, line 221: To re-use the information from the backend about next states, `extend_cterm` now returns a list of states that carries info on whether the next node after a step is vacuous, stuck, branch, or non-det branch.

pyk/src/pyk/proof/reachability.py
```diff
@@ -674,6 +674,7 @@ class APRProver(Prover[APRProof, APRProofStep, APRProofResult]):
     direct_subproof_rules: bool
     assume_defined: bool
     kcfg_explore: KCFGExplore
+    next_steps: dict[NodeIdLike, KCFGExtendResult]
```
PetarMax: When we get two results back from `extend_cterm`, it is not clear that we can just apply them both, because, for example, applying the first result produces a bounded or a terminal node, and the execution should stop there. This is tested by the test suite and a number of tests that are supposed to pass will fail if we apply both.

My idea is to store this next step information in the prover and just re-use it when we explore from the node produced by the first step instead of running `extend_cterm` again. For this I've created the `next_steps` dictionary holding these next steps and the associated logic.

pyk/src/pyk/proof/reachability.py
```diff
@@ -793,14 +795,24 @@ def step_proof(self, step: APRProofStep) -> list[APRProofResult]:
...
+                terminal_rules=self.terminal_rules,
+                module_name=step.module_name,
+                node_id=step.node.id,
+            )
+
+        assert len(extend_results) == 1 or len(extend_results) == 2
+        extend_result = extend_results[0]
+        if len(extend_results) == 2:
+            assert step.node.id not in self.next_steps
+            self.next_steps[step.node.id] = extend_results[1]
```
PetarMax: The problem is here - this is wrong and I'm not sure how to fix it - I need the identifier of the node that is to be created by the first step, but I don't have it. I know what it is, it is the `_node_id` of the associated KCFG, but the KCFG is not here.

Alternatively, I could look for the predecessors of the node, but I still don't have the KCFG here.

Alternatively, I could try to move this logic into the KCFG logic, but I'm neither sure how nor that it's the way to go.

PetarMax:
I need some help with this PR, @ehildenb, @tothtamas28, @nwatson22, if you can. I've left comments throughout, please advise.

PetarMax:
I will now run some engagement proofs using this and check for speedups.

PetarMax:
I can confirm that this change accelerates long-running Z3-heavy engagement proofs, will add more measurements as I make them:
```
09m 27s ==> 06m 41s
26m 07s ==> 22m 39s
07m 39s ==> 06m 40s
26m 34s ==> 18m 22s
```

pyk/src/pyk/proof/reachability.py
```diff
@@ -46,6 +46,7 @@ class APRProofResult:
 class APRProofExtendResult(APRProofResult):
```
tothtamas28, line 46: Please add a docstring for this class.

PetarMax: Done now, please take a look.

pyk/src/pyk/kcfg/explore.py
```diff
@@ -240,33 +236,27 @@ def log(message: str, *, warning: bool = False) -> None:
...
-            return Step(
-                next_states[0].state,
-                1,
-                next_node_logs,
-                self._extract_rule_labels(next_node_logs),
-                cut=True,
-            )
-
+        elif len(next_states) == 1:
+            if not self.kcfg_semantics.can_make_custom_step(next_states[0]):
```
PetarMax, line 253:
```suggestion
            if not self.kcfg_semantics.can_make_custom_step(next_states[0]):
```

PetarMax:
```suggestion
            if not self.kcfg_semantics.can_make_custom_step(next_states[0].state):
```

pyk/src/pyk/kcfg/explore.py
```diff
@@ -240,33 +236,27 @@ def log(message: str, *, warning: bool = False) -> None:
...
-            return Step(
-                next_states[0].state,
-                1,
-                next_node_logs,
-                self._extract_rule_labels(next_node_logs),
-                cut=True,
-            )
-
+        elif len(next_states) == 1:
+            if not self.kcfg_semantics.can_make_custom_step(next_states[0].state):
```
PetarMax:
```suggestion
            if not self.kcfg_semantics.can_make_custom_step(cterm):
```

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1306,7 +1321,8 @@ def read_node_data(cfg_dir: Path, node_id: int) -> KCFG.Node:
         return KCFG.Node.from_dict(store.read_node_data(node_id))
 
 
-class KCFGExtendResult(ABC): ...
+class KCFGExtendResult(ABC):
+    info: str = field(default='')
```
tothtamas28: Consider

```suggestion
    info: str
```

and adding the `info: str = field(default='')` to each `dataclass` subclass.

Alternatively, do not make it an abstract attribute, since the value is not used for most subclasses.

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -18,6 +18,9 @@ def abstract_node(self, c: CTerm) -> CTerm: ...
     @abstractmethod
     def same_loop(self, c1: CTerm, c2: CTerm) -> bool: ...
 
+    @abstractmethod
+    def can_make_custom_step(self, c: CTerm) -> bool: ...
```
tothtamas28, line 28: Please add a docstring.

pyk/src/pyk/proof/reachability.py
```diff
@@ -44,7 +44,19 @@ class APRProofResult:
 
 @dataclass
 class APRProofExtendResult(APRProofResult):
-    extend_result: KCFGExtendResult
+    """Holds the description of how an APRProof should be extended.
+
+    Fields:
```
tothtamas28:
```suggestion
    Attributes:
```

pyk/src/pyk/proof/reachability.py
```diff
@@ -44,7 +44,19 @@ class APRProofResult:
...
+        extend_results: Holds the KCFG extension to be applied.
+        to_cache: Holds an indicator of whether or not the provided extension
+                  should be cached instead of being applied.
```
tothtamas28: What does it mean for an extension to be cached?

pyk/src/pyk/proof/reachability.py
```diff
@@ -189,7 +212,26 @@ def get_steps(self) -> list[APRProofStep]:
     def commit(self, result: APRProofResult) -> None:
         self.prior_loops_cache[result.node_id] = result.prior_loops_cache_update
         if isinstance(result, APRProofExtendResult):
-            self.kcfg.extend(result.extend_result, self.kcfg.node(result.node_id), logs=self.logs)
+            # Result has been cached, use the cache
+            if result.use_cache is not None:
+                assert len(result.extend_results) == 0
+                assert result.use_cache in self._next_steps
```
tothtamas28: To me, all these assertions indicate that `APRProofExtendResult` could be broken up into multiple subclasses with stronger typing guarantees.

PetarMax: 100%. I will think about this a bit.

pyk/src/pyk/proof/reachability.py
PetarMax: `extend_cterm` can now return two extensions, the first of which is a `Step`, and the second of which is a cut-rule, branch, or NDBranch. When that happens, the first one is to be applied, and the second one is to be cached so that it can be automatically applied instead of recomputed when we explore from the node created by that first `Step`.

pyk/src/pyk/kcfg/semantics.py
PetarMax: Currantly, the docstrings do not exist for the other functions. I will add for all.

pyk/src/pyk/proof/reachability.py
PetarMax: Done.

pyk/src/pyk/kcfg/kcfg.py
PetarMax: Would it be ok to just have it in the ones where it's needed? The compilation does not seem to complain.

tothtamas28: Yes, but then it shouldn't be present in the abstract class.

PetarMax: Yes, I removed it from there.

pyk/src/pyk/proof/reachability.py
PetarMax: I've redesigned them now.

pyk/src/pyk/proof/reachability.py
```diff
@@ -761,15 +800,15 @@ def step_proof(self, step: APRProofStep) -> list[APRProofResult]:
                 _LOGGER.warning(f'Bounded node {step.proof_id}: {step.node.id} at bmc depth {step.bmc_depth}')
                 return [APRProofBoundedResult(node_id=step.node.id, prior_loops_cache_update=prior_loops)]
 
-        # Terminal checks for current node and target node
+        # Check if the current node and target ar terminal
```
tothtamas28:
```suggestion
        # Check if the current node and target are terminal
```

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -12,15 +12,28 @@ class KCFGSemantics(ABC):
     @abstractmethod
     def is_terminal(self, c: CTerm) -> bool: ...
 
+    """Checks whether or not a given CTerm is terminal."""
```
tothtamas28:
```suggestion
    """Check whether or not a given ``CTerm`` is terminal."""
```

pyk/src/pyk/proof/reachability.py
```diff
@@ -790,20 +829,52 @@ def step_proof(self, step: APRProofStep) -> list[APRProofResult]:
...
-        extend_result = self.kcfg_explore.extend_cterm(
-            step.node.cterm,
-            execute_depth=execute_depth,
-            cut_point_rules=cut_rules,
-            terminal_rules=self.terminal_rules,
-            module_name=step.module_name,
-            node_id=step.node.id,
-        )
+        # If the step has already been cached, do not invoke the backend and only send a signal back to the proof to use the cache
+        if step.use_cache is not None:
```
tothtamas28, line 836: What happens if the prover is terminated with cached results not yet applied? Is it still possible to continue extending the KCFG?

tothtamas28 (approved):
LGTM, please test on `evm-semantics` and `kontrol` before merging.

PetarMax:
Yes, it has been extensively tested with Kontrol and engagement tests.

pyk/src/pyk/proof/reachability.py
PetarMax: Yes, it is, and cached results should be re-used if the proof has survived. If they are lost, there is no loss for the KCFG, since they are only a snapshot of guaranteed future extension results.

Issue #4579, opened 2024-08-08
Support for syntactic simplifications


PetarMax:
To improve the simplification mechanism of the backend, we require the addition of a `syntactic` attribute in simplifications, which takes a list of positive integers. These integers would then indicate to the backend that the corresponding clauses in the `requires` should be checked only syntactically. Two examples of this would be:
```
rule X <=Int Y => X <Int Y
  requires notBool (X ==Int Y)
  [simplification, syntactic(1)]
```
and
```
rule X <=Int Y => true 
  requires X <=Int Z
   andBool Z <=Int Y
  [simplification, concrete(Y, Z), syntactic(1)]
```
Note that, in the second example, there is a symbolic variable `Z` introduced in the first clause that is not present in the LHS. This appears to be supported by the front-end if the corresponding module is marked as `[symbolic]`. Optionally, we would like such variables to be supported also if the clause in they first appear is marked as syntactic.

Scott-Guest:
@PetarMax Not familiar here - what explicitly counts as a clause / how is the numbering of clauses precisely defined?

PetarMax:
I am thinking of simplifications as follows:
```
rule LHS => RHS
  requires C1 andBool C2 andBool ... andBool CN
  [simplification]
```
where the `Ci` are the clauses, separated by `andBool`, and numbered in order in which they appear in the simplification.

geo2a:
I've implemented a prototype of this feature in Booster (https://github.com/runtimeverification/haskell-backend/pull/4022). I've manually modified a `definition.kore` file and added `syntactic{}("1")` to the attributes of a simplification.

I have one question for @Scott-Guest: do you know if the K compiler preserves the `requires` clause as-is when compiling surface K to kore? This is important, as we are relying on the ordering of conjuncts.

And another question for @PetarMax: what behavior you expecting when there are more than one syntactic clause? I.e. `syntactic(1, 2)`. I'd expect we process them from left to right, i.e. if the first one binds all variables from the second, than the second will have no effect and the result will be equivalent to `syntactic(1)`. Does that seem reasonable?

PetarMax:
Left to right, yes please.

Scott-Guest:
> @geo2a: do you know if the K compiler preserves the requires clause as-is when compiling surface K to kore?

Hmm, that's a good point - regardless of the current situation, it's very fragile to rely on `requires` clauses never being manipulated, and I think we should consider an alternative syntax which doesn't require this assumption. 

[...]

PR #4597, opened 2024-08-20
Add syntactic to whitelist

Scott-Guest:
Part of #4579

Adds the `syntactic` attribute to the whitelist, and checks that it is only applied to simplification rules.

Notably, we still need to:
- Check that the arguments actually refer to existing clauses
- Tag the underlying clauses and re-construct the attribute at the end of the compilation pipeline because the `requires` clauses may be manipulated during compilation


PetarMax:
@Scott-Guest Were you meaning to add @geo2a and myself as reviewers?

geo2a (approved):
LGTM!

Just curios: does the frontend has any unit tests for this sort of functionality?

Scott-Guest:
@geo2a We have tests for the attribute whitelist infrastructure in general (buried somewhere in `regression-new`), but don't specifically test each entry. I'll add a small test for the "only on simplification rules" part though.

PR #4600, opened 2024-08-21
Refactor edge structure for merging node

Stevengre:
Merging nodes will combine several edges into one. To ensure no loss of information, we need to make sure that a single edge can store multiple pieces of data. However, modifying the Edge structure has wide-ranging implications. Even though I’ve added `rule` and `depth` as properties to minimize this impact, it still has broad effects.

### DIFF
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -202,20 +202,36 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
     class Edge(EdgeLike):
         source: KCFG.Node
         target: KCFG.Node
-        depth: int
-        rules: tuple[str, ...]
+        depths: tuple[int, ...]
+        rules_list: tuple[tuple[str, ...], ...]
+        csubsts: tuple[CSubst, ...]
+
+        @property
+        def depth(self) -> int:
+            return max(self.depths)
+
+        @property
+        def rules(self) -> tuple[str, ...]:
+            return single(self.rules_list)
 
         def to_dict(self) -> dict[str, Any]:
             return {
                 'source': self.source.id,
                 'target': self.target.id,
-                'depth': self.depth,
-                'rules': list(self.rules),
+                'depths': list(self.depths),
+                'rules_list': [list(rules) for rules in self.rules_list],
+                'csubsts': [csubst.to_dict() for csubst in self.csubsts],
             }
 
         @staticmethod
         def from_dict(dct: dict[str, Any], nodes: Mapping[int, KCFG.Node]) -> KCFG.Edge:
-            return KCFG.Edge(nodes[dct['source']], nodes[dct['target']], dct['depth'], tuple(dct['rules']))
+            return KCFG.Edge(
+                nodes[dct['source']],
+                nodes[dct['target']],
+                tuple(dct['depths']),
+                tuple(tuple(rules) for rules in dct['rules_list']),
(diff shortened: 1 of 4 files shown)


Stevengre:
This is the primary change.

https://github.com/runtimeverification/k/blob/0b3669fcf65a69b19d1195003cf9053519e0f217/pyk/src/pyk/kcfg/kcfg.py#L205-L207

I’ve tried to update all the affected methods, but I’m unsure what changes should be made to the following: KCFGShow and KCFGViewer.

https://github.com/runtimeverification/k/blob/0b3669fcf65a69b19d1195003cf9053519e0f217/pyk/src/pyk/kcfg/show.py#L132-L136
https://github.com/runtimeverification/k/blob/0b3669fcf65a69b19d1195003cf9053519e0f217/pyk/src/pyk/kcfg/tui.py#L225
https://github.com/runtimeverification/k/blob/0b3669fcf65a69b19d1195003cf9053519e0f217/pyk/src/pyk/kcfg/tui.py#L291

Additionally, will this update affect KaaS?

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -202,20 +202,36 @@ def targets(self) -> tuple[KCFG.Node, ...]:
     class Edge(EdgeLike):
         source: KCFG.Node
         target: KCFG.Node
-        depth: int
-        rules: tuple[str, ...]
+        depths: tuple[int, ...]
+        rules_list: tuple[tuple[str, ...], ...]
+        csubsts: tuple[CSubst, ...]
```
tothtamas28, line 207: I'm not sure we can manage a breaking change of this scale across our stack at this point.

Is there a reasonable way to merge edges _with_ loss of information? (E.g. take the the maximum `depth`, etc.)

Stevengre: I’ve added two properties to mitigate this issue, ensuring that previous access to `depth` and `rule` continues to work as expected.

https://github.com/runtimeverification/k/blob/0b3669fcf65a69b19d1195003cf9053519e0f217/pyk/src/pyk/kcfg/kcfg.py#L209-L215

What they should return can be further evaluated.

tothtamas28: `to_dict`  and `from_dict` still changes though, and as implemented, proofs written before the change will no longer be usable with this version.

We could still
1. Update the old format on the first read, or
2. Support both formats

but we should really think about if adding this complexity is worth it right now.

tothtamas28: If losing information is not an option for your use case, you could still add a new class `MergedEdge(EdgeLike)` that contains a `tuple[Edge, ...]`.

Stevengre: Thank you for your suggestion! I think introducing a new class is a good idea. It won’t disrupt the existing proofs and will preserve the information that would otherwise be lost. However, we’ll still need to modify KCFGShow and KCFGViewer to adapt to this change, but the adjustments required will be much smaller.

Stevengre: Moreover, we need to change the `as_rule` and `as_rules` to make this new Edge class work correctly for the CSE.

Stevengre: If that implementation is better, I will close this PR and work that then. What do you think?

ehildenb: I like the `MergeEdge` idea!

PR #4601, opened 2024-08-21
Start `kdist` workers with `spawn`

tothtamas28:
* Ensure that the start method is the same on all platforms (https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)
* Set the log level in the worker. This fixes an issue on macOS where the worker logs are not visible.


pyk/src/pyk/kdist/_kdist.py
```diff
@@ -134,10 +138,13 @@ def _build_target(
         target_id: TargetId,
         args: dict[str, Any],
         *,
+        log_level: int,
         force: bool,
         verbose: bool,
         clean: bool,
     ) -> Path:
+        logging.basicConfig(level=log_level, format=LOG_FORMAT)
```
tothtamas28, line 146: @goodlyrottenapple, could you check please if this solves the logging issue on macOS?

goodlyrottenapple: Yes! Perfect :)

gtrepta (approved):

PR #4603, opened 2024-08-22
MergedEdge class for merging node functionality

Stevengre:
Here is the alternative implementation for [this PR](https://github.com/runtimeverification/k/pull/4600).

- commit 1: introduce the new class `MergedEdge`
- commit 2: add `to_rule` function to `MergedEdge` for CSE
- commit 3: add corresponding functions like `Edge` for `MergedEdge`
- commit 4: modify `Edge.to_rule` related functions for CSE
- commit 5: modify KCFGShow for MergedEdge
- commit 6: modify KCFGViewer for MergedEdge

### DIFF
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -542,7 +542,7 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         elif type(_path[0]) is KCFG.Edge:
             return _path[0].depth + KCFG.path_length(_path[1:])
         elif type(_path[0]) is KCFG.MergedEdge:
-            return sum(edge.depth for edge in _path[0].edges) + KCFG.path_length(_path[1:])  # todo: check this
+            return min(edge.depth for edge in _path[0].edges) + KCFG.path_length(_path[1:])  # todo: check this
         raise ValueError(f'Cannot handle Successor type: {type(_path[0])}')
 
     def extend(
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -247,6 +247,38 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
 
     @final
     @dataclass(frozen=True)
+    class MergedEdge(EdgeLike):
+        """Merged edge is a collection of edges that have been merged into a single edge."""
+
+        source: KCFG.Node
+        target: KCFG.Node
+        edges: tuple[KCFG.Edge, ...]
+
+        def to_dict(self) -> dict[str, Any]:
+            return {
+                'source': self.source.id,
+                'target': self.target.id,
+                'edges': [edge.to_dict() for edge in self.edges],
+            }
+
+        @staticmethod
+        def from_dict(dct: dict[str, Any], nodes: Mapping[int, KCFG.Node]) -> KCFG.Successor:
+            return KCFG.MergedEdge(
+                nodes[dct['source']],
+                nodes[dct['target']],
+                tuple(KCFG.Edge.from_dict(edge, nodes) for edge in dct['edges']),
+            )
+
+        def replace_source(self, node: KCFG.Node) -> KCFG.Successor:
+            assert node.id == self.source.id
+            return KCFG.MergedEdge(node, self.target, self.edges)
+
+        def replace_target(self, node: KCFG.Node) -> KCFG.Successor:
+            assert node.id == self.target.id
+            return KCFG.MergedEdge(self.source, node, self.edges)
+
+    @final
+    @dataclass(frozen=True)
     class Cover(EdgeLike):
         source: KCFG.Node
         target: KCFG.Node
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -683,9 +683,11 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
     def from_json(s: str, optimize_memory: bool = True) -> KCFG:
         return KCFG.from_dict(json.loads(s), optimize_memory=optimize_memory)
 
-    def to_rules(self, priority: int = 20, id: str | None = None) -> list[KRuleLike]:
-        id = '' if id is None else f'{id}-'
-        return [e.to_rule(f'{id}BASIC-BLOCK', priority=priority) for e in self.edges()]
+    def to_rules(self, _id: str | None = None, priority: int = 20) -> list[KRuleLike]:
+        _id = 'BASIC-BLOCK' if _id is None else _id
+        return [e.to_rule(_id, priority=priority) for e in self.edges()] + [
+            m.to_rule(_id, priority=priority) for m in self.merged_edges()
+        ]
 
     def to_module(
         self,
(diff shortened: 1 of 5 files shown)


pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -501,6 +540,8 @@ def path_length(_path: Iterable[KCFG.Successor]) -> int:
             return 1 + KCFG.path_length(_path[1:])
         elif type(_path[0]) is KCFG.Edge:
             return _path[0].depth + KCFG.path_length(_path[1:])
+        elif type(_path[0]) is KCFG.MergedEdge:
+            return sum(edge.depth for edge in _path[0].edges) + KCFG.path_length(_path[1:])  # todo: check this
```
Stevengre: I need some suggestions here because I’m not sure what the `path_length` is for. It seems to be primarily used to calculate the shortest path, obtain the path constraints, and eventually print them as `APRFailureInfo`. However, even if the shortest path is resolved, other paths can still reach the error node. So, this seems to be a guide to help users adjust their specs. If it’s a sum, it encourages users to review paths containing merged edges last; if it’s set to min or 1, it helps users see this path earlier.

Stevengre:
Compared to #4600, the impact of this implementation is smaller, but it still has some effects. Specifically, for `KCFG.minimize`, the `lift_edge_edge` and `lift_split_edge` functions won’t work with the `MergedEdge`, which will prevent obtaining a summary for a function that uses a summary. However, this approach still seems simpler than #4600. I’ll start a separate PR to refactor the existing `KCFG.minimize`.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -244,6 +244,41 @@ def replace_target(self, node: KCFG.Node) -> KCFG.Edge:
...
+
+        @staticmethod
+        def from_dict(dct: dict[str, Any], nodes: Mapping[int, KCFG.Node]) -> KCFG.Successor:
+            return KCFG.MergedEdge(
+                nodes[dct['source']],
+                nodes[dct['target']],
+                tuple(KCFG.Edge.from_dict(edge, nodes) for edge in dct['edges']),
+            )
+
+        def replace_source(self, node: KCFG.Node) -> KCFG.Successor:
```
tothtamas28, line 272: Does replacing the source (target) of such a merged edge not have an implication on the source (target) of the contained edges?

pyk/src/pyk/proof/reachability.py
```diff
@@ -444,12 +444,7 @@ def as_rules(self, priority: int = 20, direct_rule: bool = False) -> list[KRule]
             or (self.admitted and not self.kcfg.predecessors(self.target))
         ):
             return [self.as_rule(priority=priority)]
-        _rules = []
-        for _edge in self.kcfg.edges():
-            _rule = _edge.to_rule(self.rule_id, priority=priority)
-            assert type(_rule) is KRule
-            _rules.append(_rule)
-        return _rules
+        return [rule for rule in self.kcfg.to_rules(self.rule_id, priority=priority) if isinstance(rule, KRule)]
```
tothtamas28: Consider

```suggestion
        def assert_rule(x: object) -> KRule:
            assert isinstance(x, KRule)
            return x
            
        return [assert_rule(rule) for rule in self.kcfg.to_rules(self.rule_id, priority=priority)]
```

tothtamas28: But you can even strengthen the return type of `to_rules` by pulling the assertion there.

pyk/src/pyk/kcfg/kcfg.py
tothtamas28: I think `min` is a reasonable choice here.

tothtamas28 (commented):
Please add some unit tests for a `KCFG` containing `MergedEdge` instances.

pyk/src/pyk/kcfg/kcfg.py
PetarMax: Are we even allowed to do this, given the semantics of merged edges?

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -244,6 +244,41 @@ def replace_target(self, node: KCFG.Node) -> KCFG.Edge:
...
+            assert node.id == self.source.id
+            return KCFG.MergedEdge(node, self.target, self.edges)
+
+        def replace_target(self, node: KCFG.Node) -> KCFG.Successor:
```
PetarMax, line 276: Same as for `replace_source`...

pyk/src/pyk/kcfg/kcfg.py
Stevengre: Done!

pyk/src/pyk/kcfg/kcfg.py
Stevengre: Firstly, `Successor` has abstract methods `replace_source` and `replace_target`, so I had to implement them in the `MergedEdge`. Although it may seem a bit odd semantically, it does provide some syntactic and operational convenience.

Secondly, there is a semantic relationship here, but we cannot verify this implication without an interpreter (SymbolicCTerm). And from what I understand, not all implications can be solved in Z3.

pyk/src/pyk/proof/reachability.py
Stevengre: Although `to_rules` always returns `KRule`, I’m not sure if the previous design was intentional. So I’ve decided to follow your suggestion.

pyk/src/pyk/kcfg/kcfg.py
Stevengre: Thank you! We can continue the discussion in the previous comments.

Stevengre:
> Please add some unit tests for a `KCFG` containing `MergedEdge` instances.

[This commit](https://github.com/runtimeverification/k/pull/4603/commits/cc1e353f4341f9e99982101c1f69d6c82dc82c18) adds these unit tests, but leaves the unit test for minimization for another PR.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: The `MultiEdge` should function just as a normal edge, just that it carries on it extra data, specifically:

- A `list` of depths/rules/constraints attached to it, indicating the different paths to take to get between the source and target.

I don't think `replace_[source|target]` really need to be treated specially here. They are maintenance functions for updating the nodes in an edge. They are used, for example, when calling `simplify` on all the nodes in the KCFG and we need to update the references on the edges. They exist just because we still store actual `Node` on the edges in the graph, rather than just references to a node dictionary.

Stevengre: So, I think that I can resolve this conversation

tothtamas28 (approved):
LGTM. Please test on `evm-semantics` before merging.

Stevengre:
- `make cov-unit` passed for `evm-semantics`
- `make cov-unit` passed for `kontrol`

tothtamas28:
> * `make cov-unit` passed for `evm-semantics`
> * `make cov-unit` passed for `kontrol`

Please also run integration tests for `evm-semantics`.

Stevengre:
After spending a long time configuring, I still couldn’t get the evm-semantics integration tests to run. So I decided to submit it as a [PR](https://github.com/runtimeverification/evm-semantics/pull/2596) and use the existing CI for testing.

Stevengre:
This [PR](https://github.com/runtimeverification/evm-semantics/pull/2596) encountered the same error as the previous [PR](https://github.com/runtimeverification/evm-semantics/pull/2571), and it also failed the Nix tests.

Therefore, I believe the introduction of the merged edge doesn’t affect evm-semantics. It might be some other updates in the pyk repository that are impacting evm-semantics, but I’m not entirely sure.

tothtamas28 (approved):

PR #4607, opened 2024-08-26
Move kcfg.minimize into a separate file and class

Stevengre:
Because there will be more pattern rewriting to minimize the KCFG, we need to extract syntax-unrelated functions from `kcfg.py` for clarity. This PR doesn’t introduce any new features; it simply moves functions related to KCFG minimization to `minimize.py`.

The downstream repo that would be affected is likely https://github.com/runtimeverification/kontrol:

- `Proof.minimize_kcfg()` == `KCFGExploration.minimize_kcfg()` will not be affected:
  - https://github.com/runtimeverification/kontrol/blob/beacbb2d8267e29804864364e7263846283e088f/src/kontrol/foundry.py#L988-L992
  - https://github.com/runtimeverification/kontrol/blob/beacbb2d8267e29804864364e7263846283e088f/src/kontrol/prove.py#L463-L464

- `kcfg.minimize` will not be affected:
  - https://github.com/runtimeverification/kontrol/blob/beacbb2d8267e29804864364e7263846283e088f/src/kontrol/prove.py#L730-L734


pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1308,6 +1132,9 @@ def write_cfg_data(self) -> None:
         self._deleted_nodes.clear()
         self._created_nodes.clear()
 
+    def minimize(self) -> None:
```
Stevengre, line 1135: kcfg.minimize() with the same effect.

pyk/src/pyk/kcfg/exploration.py
```diff
@@ -109,4 +111,4 @@ def to_dict(self) -> dict[str, Any]:
 
     # Minimizing the KCFG
     def minimize_kcfg(self) -> None:
-        self.kcfg.minimize()
+        KCFGMinimizer(self.kcfg).minimize()
```
Stevengre: proof.minimize_kcfg() with the same effect.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1308,6 +1132,9 @@ def write_cfg_data(self) -> None:
...
+        KCFGMinimizer(self).minimize()
```
Stevengre, line 1136: kcfg.minimize() with the same effect.

pyk/src/pyk/kcfg/exploration.py
tothtamas28: This change doesn't seem necessary, as `KCFG` still provides the `minimize` method.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -0,0 +1,200 @@
...
+from pyk.cterm import CTerm
+from pyk.utils import not_none, single
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from pyk.kcfg.kcfg import KCFG, NodeIdLike
+
+
+class KCFGMinimizer:
```
tothtamas28, line 15: Just to confirm: does `KCFG` have any private attributes this class needs access to?

pyk/src/tests/integration/proof/test_refute_node.py
```diff
@@ -372,7 +373,7 @@ def test_apr_proof_refute_node_multiple_constraints(
         prover.advance_proof(proof, max_iterations=4)
         # After the minimization, nodes 7-10 created by the advancement of the proof
         # will have multiple constraints in their immediately preceding splits
-        proof.kcfg.minimize()
+        KCFGMinimizer(proof.kcfg).minimize()
```
tothtamas28: Here as well. Let's keep `KCFGMinimizer` private to the `kcfg` module.

pyk/src/pyk/kcfg/exploration.py
Stevengre: Thank you for pointing this out. I'll remove this change.

pyk/src/pyk/kcfg/minimize.py
Stevengre: self.kcfg._resolve(b_id) at L134

Stevengre: Does it matter?

tothtamas28: > Does it matter?

In general, yes.
* There's no guarantee for backwards compatibiliy on a private method.
* A class can only guarantee its invariants if it's internals are not tinkered with.

However, this particular case is pretty innocuous. Still, it'd great to eliminate it (e.g. by `node(b_id).id`).

pyk/src/tests/integration/proof/test_refute_node.py
Stevengre: Sure, I'll change that!

pyk/src/pyk/kcfg/minimize.py
Stevengre: Thank you for your suggestion! I’ll make that change. But I have a question: will `node(b_id).id` be significantly slower than the current implementation?

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -0,0 +1,200 @@
...
+        splits_from_a, splits_from_b = split_from_a.splits, split_from_b.splits
+        a = split_from_a.source
+        ci = list(splits_from_b.keys())
+        # Ensure split can be lifted soundly (i.e., that it does not introduce fresh variables)
+        assert (
+            len(split_from_b.source_vars.difference(a.free_vars)) == 0
+            and len(split_from_b.target_vars.difference(split_from_b.source_vars)) == 0
+        )
+        # Get the substitution for `B`, at the same time removing 'B' from the targets of `A`.
+        csubst_b = splits_from_a.pop(self.kcfg.node(b_id).id)
```
Stevengre, line 134: Done!

pyk/src/tests/integration/proof/test_refute_node.py
Stevengre: Done! Nothing changed in [test_refute_node.py](https://github.com/runtimeverification/k/pull/4607/files#diff-c50f6d907465456da017ef879b6a11d15c1cfddc76dfc8eb3b6f122532239616) for now.

pyk/src/pyk/kcfg/minimize.py
tothtamas28: It's not significantly slower, it's just one extra map lookup plus an attribute access.

tothtamas28 (approved):
Nice refactoring!

anvacaru (approved):

PR #4610, opened 2024-08-27
Basic defunctionalization for building rules in K

ehildenb:
Currently, when we are converting a summarized K definition to a module, we have only used it for then feeding that module into the Haskell backend, which does not have the restriction that the generated rules cannot have functions on the LHS of them. Now Pi2 wants to also feed those rules into the LLVM backend, so it's needed that we can defunctionalize the LHS of those rules.

This PR achieves that by threading through the needed options to `cterm_build_rule` to defunctionalize the LHS of generated rules. In particular:

- Some bugs in the imports of https://github.com/runtimeverification/k/pull/4607 are fixed. Imports are made relative, and a circular import path is broken.
- A routine `defunctionalize(KDefinition, KInner) -> tuple[Kinner, list[KInner]]` is added to `kast.manip`, which replaces functions in the supplied term with variables and returns also a list of constraints binding those variables to their original terms.
- A unit test-harness, and many tests, are added of the `defunctionalize` routine.
- An integration test of `cterm_build_rule` defunctionalization is added over the `simple_proofs.k` test harness.
- The "functions" `.Map`, `.Set`, `.List`, and `.RangeMap` are added to the list of "not actually functions, more like constructors" for considering when deciding if something is a function or a constructor. This is exercised by the added test because `.Map` shows up, but is not defunctionalized.
- `KCFG.edge.to_rule`, `cterm_build_rule`, and `build_rule` now take an optional `KDefinition`, and if it's supplied, will defunctionalize the LHS of generated rules that it makes.
- The routine in `kast.manip` named `minimize_rule` is renamed to `minimize_rule_like`, to reflect better its type.
- `KCFG.edge.to_rule` now takes an optional `minimize: bool` parameter, which if supplied will run the `minimize_rule_like` routine over the generated rule.


pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1234,9 +1237,6 @@ def write_cfg_data(self) -> None:
         self._deleted_nodes.clear()
         self._created_nodes.clear()
 
-    def minimize(self) -> None:
-        KCFGMinimizer(self).minimize()
```
Stevengre, line 1238: If this is deleted, Kontrol will no longer function as before since it calls `KCFG.minimize()`.

Stevengre: However, I haven’t figured out how to break this circular import path without modifying the existing interface and functionality for the downstream repos. Does this circular import cause any issues? I didn’t notice any errors during my testing.

ehildenb: It does cause more issues as things are added to th efile. Circular imports are always a problem.

I intend to fix the downstream repos, here is the fix for Kontrol: https://github.com/runtimeverification/kontrol/pull/793/commits/dfd7222f28fb857fd010779e76d37443bb124149

Stevengre: Thank you! I think this is a better way.

tothtamas28: Consider:

```python3
def minimize(self) -> None:
    from .minimize import KCFGMinimizer
    KCFGMinimizer(self).minimize()
```

This way, the interface can be maintained. This (and the relative import fix) can be pulled out in a separate PR and merged right away.

pyk/src/pyk/kast/manip.py
```diff
@@ -854,3 +861,30 @@ def _no_cell_rewrite_to_dots(_term: KInner) -> KInner:
...
+    Returns:
+        A tuple of the defunctionalized term and the list of constraints generated.
+    """
+    function_symbols = [prod.klabel for prod in defn.functions if prod.klabel is not None]
+    constraints: list[KInner] = []
+
+    def _defunctionalize(_kinner: KInner) -> KInner:
+        if type(_kinner) is KApply and _kinner.label in function_symbols:
+            sort = defn.sort(_kinner)
+            assert type(sort) is KSort
```
tothtamas28:
```suggestion
            sort is not None
```

pyk/src/pyk/kast/manip.py
```diff
@@ -756,6 +756,7 @@ def build_rule(
     final_constraints: Iterable[KInner] = (),
     priority: int | None = None,
     keep_vars: Iterable[str] = (),
+    definition: KDefinition | None = None,
```
tothtamas28: Maybe `defunc_with`? This way, the role of the `KDefinition` is expressed.

pyk/src/pyk/kast/manip.py
```diff
@@ -778,6 +781,10 @@ def build_rule(
     init_constraints = [normalize_ml_pred(c) for c in init_constraints]
     final_constraints = [normalize_ml_pred(c) for c in final_constraints]
     final_constraints = [c for c in final_constraints if c not in init_constraints]
+    new_constraints: list[KInner] = []
```
tothtamas28: Does this type check?

```suggestion
```

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -218,7 +218,14 @@ def to_dict(self) -> dict[str, Any]:
...
             return KCFG.Edge(nodes[dct['source']], nodes[dct['target']], dct['depth'], tuple(dct['rules']))
 
-        def to_rule(self, label: str, claim: bool = False, priority: int | None = None) -> KRuleLike:
+        def to_rule(
+            self,
+            label: str,
+            claim: bool = False,
+            priority: int | None = None,
+            definition: KDefinition | None = None,
+            minimize: bool = False,
```
tothtamas28, line 227: If you rename the parameter in `cterm_build_rule`, this has to be updated as well.

pyk/src/pyk/kast/manip.py
```diff
@@ -854,3 +861,30 @@ def _no_cell_rewrite_to_dots(_term: KInner) -> KInner:
     subst = Subst({cell_name: _no_cell_rewrite_to_dots(cell_contents) for cell_name, cell_contents in _subst.items()})
 
     return subst(config)
+
+
+def defunctionalize(defn: KDefinition, kinner: KInner) -> tuple[KInner, list[KInner]]:
```
tothtamas28, line 865: I think this could even be unit tested, the only tedious part is setting up a definition. Here's an example from another unit test: https://github.com/runtimeverification/k/blob/809d59eb819dcb14ac5fb0eb9abe1c05eae38fde/pyk/src/tests/unit/kast/test_formatter.py#L34-L47

pyk/src/pyk/kast/manip.py
ehildenb: Done.

pyk/src/pyk/kast/manip.py
ehildenb: Done.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: Done.

pyk/src/pyk/kast/manip.py
ehildenb: Indeed it does. I guess that was leftover from a prior version. Removed.

pyk/src/pyk/kast/manip.py
ehildenb: Done, I added them to `unit/kast/test_manip.py`, and tested a bunch of scenarios. I kept the integration test I added as well.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: I'm inclined to leave it as is, as I'll be pushing through the updates today anyway. But can change it if you like.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -375,6 +376,8 @@ def cterm_build_rule(
         init_cterm: State to put on LHS of the rule (constraints interpreted as `requires` clause).
         final_cterm: State to put on RHS of the rule (constraints interpreted as `ensures` clause).
         keep_vars: Variables to leave in the side-conditions even if not bound in the configuration.
+        priority: Priority index to use for generated rules.
+        defunc_with: KDefinition to be able to defunctionalize LHS appropriately.
```
tothtamas28:
```suggestion
        defunc_with (optional): KDefinition to be able to defunctionalize LHS appropriately.
```

Renders like this: https://kframework.org/pyk/api/pyk.cterm.cterm.html#pyk.cterm.cterm.CTerm.anti_unify

pyk/src/tests/unit/kast/test_manip.py
```diff
@@ -543,3 +549,94 @@ def test_is_term_like(test_id: str, kast: KInner, expected: bool) -> None:
...
+    (
+        to_s_2(add(token(1), token(2)), to_s(to_int(init_s()))),
+        to_s_2(KVariable('F_eefa6e95', INT), to_s(KVariable('F_3294e1be', INT))),
+        [var_equals('F_eefa6e95', add(token(1), token(2))), var_equals('F_3294e1be', to_int(init_s()))],
+    ),
+)
+
+
+@pytest.mark.parametrize('input,expected_output,expected_constraints', TEST_DATA, ids=count())
+def test_defunctionalize(input: KInner, expected_output: KInner, expected_constraints: list[KInner]) -> None:
```
tothtamas28, line 636: Nice!

tothtamas28 (approved):

PR #4612, opened 2024-08-30
Mergeable heuristic in KCFGSemantics for merging node

Stevengre:
This is a simple PR for merging nodes, offering a semantics-level heuristic to partition nodes that can be merged.

### DIFF
--- a/pyk/src/pyk/kcfg/semantics.py
+++ b/pyk/src/pyk/kcfg/semantics.py
@@ -34,6 +34,11 @@ class KCFGSemantics(ABC):
 
     """Implement a custom semantic step."""
 
+    @abstractmethod
+    def is_mergeable(self, c1: CTerm, c2: CTerm) -> bool: ...
+
+    """Check whether or not the two given ``CTerm``s are mergeable. Must be transitive, commutative, and reflexive."""
+
 
 class DefaultSemantics(KCFGSemantics):
     def is_terminal(self, c: CTerm) -> bool:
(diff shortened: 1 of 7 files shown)


ehildenb:
I don't think it makes sense to merge this without also having some tests of it somehow. Do you have an example of IMP node merging worknig, or another simple definition?

Stevengre:
I tested this on an actual Solidity program a few days ago. However, this PR primarily introduces an interface to define the heuristic, which can't be tested without the merge_node function. Since the new interface results in many test changes, I plan to submit a separate PR for those.

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -34,6 +34,11 @@ def custom_step(self, c: CTerm) -> KCFGExtendResult | None: ...
 
     """Implement a custom semantic step."""
 
+    @abstractmethod
+    def is_mergeable(self, c1: CTerm, c2: CTerm) -> bool: ...
```
Stevengre, line 38: This PR only introduces an interface. Its semantics will be determined by the merge_node function, which will be provided in another PR.

ehildenb:
Well we can wait for the merging funcitonality to be ready to go, and merged, then merge this new interface and some tests of it with the node merging. This will trigger needed changes downstream, because each of the downstream semantics will have to now implement this interface, and we don't know if it's fully working yet.

Once https://github.com/runtimeverification/k/pull/4617 and https://github.com/runtimeverification/k/pull/4621 are merged, we can then have this PR introduce both the `merge_nodes` predicate, the main algorithm for merging KCFG nodes, and some tests over one of the semantics.

Stevengre:
Thank you for your suggestion! I'll follow your advice.

ehildenb:
Subsumed by: https://github.com/runtimeverification/k/pull/4628

PR #4615, opened 2024-08-30
Set pyk version with `importlib.metadata.version`

### LINKED ISSUE #4614: [K-Bug] Python code should use version number from pyproject.toml

Front-End

### Which command

- [ ] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

7.1.125

### Operating System

Linux

Note that in the above it's not just Linux, it's all OSes. You should change that dropdown to allow that, instead of forcing to select a particular OS.

### Steps to Reproduce

None; it's inherent in the system.

### Expected Results

This form is really bad for the bug I'm trying to report.

In `pyk/src/pyk/__init__.py` you should not be duplicating the version number; you should be using:

    from importlib.metadata import version
    ...
    __version__: Final = version('pyk')

to get it from `pyproject.toml`.

Scott-Guest:
Closes #4614

Uses `importlib.metadata.version` to read the version from `pyproject.toml` rather than manually `sed`ding.


tothtamas28 (approved):
Nice!

PR #4616, opened 2024-09-02
Use Euclidean division/modulus to express T-division/modulus for SMT hooks

### LINKED ISSUE #4606: Potential bug in how integer division/modulus are encoded to SMT
See Slack discussion: https://runtimeverification.slack.com/archives/C03NP1UP2KZ/p1724335226574649

We have two different forms of division/modulus available to users: https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md#integer-arithmetic

One is Euclidian division, which rounds towards nearest. The other is t-division ("truncated division"), which rounds towards zero.

In particular:

```
 syntax Int ::= ...
               /* /Int and %Int implement t-division, which rounds towards 0 */
               | Int "/Int" Int                 [function, symbol(_/Int_), left, smt-hook(div), hook(INT.tdiv)]
               | Int "%Int" Int                 [function, symbol(_%Int_), left, smt-hook(mod), hook(INT.tmod)]
               /* divInt and modInt implement e-division according to the Euclidean division theorem, therefore the remainder is always positive */
               | Int "divInt" Int               [function, symbol(_divInt_), left, smt-hook(div), hook(INT.ediv)]
               | Int "modInt" Int               [function, symbol(_modInt_), left, smt-hook(mod), hook(INT.emod)]
```

The issue is that we're giving the same `smt-hook(...)` to both types of division and modulus, even though they behave differently below zero.

Looking at smt-lib, it seems that it implements Euclidian division: https://smt-lib.org/theories-Ints.shtml. So we have the correct implmeentation for `divInt` and `modInt`, but not `/Int` and `%Int`.

Here is a random gist I found that implements functions for `truncated` variants: https://gist.github.com/Rufflewind/a880e03fb0d13644a1e8.

Perhaps we could add these truncated variants to our SMT prelude that the backend uses, and hook those methods instead?

jberthold:
Fixes #4606

The following reasoning was applied:

Given dividend `D` and divisor `d`, let the `q₋` and `r₋` in
```
D = d · qₑ + rₑ    # Euclidean division
D = d · qₜ + rₜ    # T-division
```
be Euclidean and T division results. In terms of K operations,
```
qₑ = D divInt d    rₑ = D modInt d
qₜ = D /Int d      rₜ = D %Int d
```

According to [1], we can express Euclidean division by T-division:
```
qₑ = qₜ − I
rₑ = rₜ + I · d
  where
    I = if rₜ >= 0
        then 0
        else if d > 0
             then 1
             else −1
```

To transform the equations to isolate `qₜ` and `rₜ`, we need to express the `where` clause without `rₜ`. We use

1. `rₑ = 0 <=> rₜ = 0`, and
2. `rₜ ≠ 0 => sgn(rₜ) = sgn(D)` (as per [1])

to rewrite the first `if` to `if (rₑ = 0 || D > 0) ...`.

Now we can transform the equations to isolate `qₜ` and `rₜ`, getting (written with `let`):

```
let I = if (rₑ = 0 || D > 0)
        then 0
        else if d > 0
             then 1
             else -1
 in qₜ = qₑ + I
    rₜ = rₑ - I · d
```

We need two independent equations, so lifting the shared `I` is not beneficial. we can write two `if` expressions per equation.

```
qₜ = if (rₑ = 0 || D > 0)
     then qₑ
     else if d > 0
          then qₑ + 1
          else qₑ - 1
rₜ = if (rₑ = 0 || D > 0)
     then rₑ
     else if d > 0
          then rₑ - d
          else rₑ + d
```

Written as SMT hook expressions (coherent with [2]):

* (`#1 /Int #2`): `ite (or (= 0 (mod #1 #2)) (>= #1 0)) (div #1 #2) (ite (> #2 0) (+ (div #1 #2) 1) (- (div #1 #2) 1))`
* (`#1 %Int #2`): `ite (or (= 0 (mod #1 #2)) (>= #1 0)) (mod #1 #2) (ite (> #2 0) (- (mod #1 #2) #2) (+ (mod #1 #2) #2))`

[1] https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/divmodnote-letter.pdf
[2] https://gist.github.com/Rufflewind/a880e03fb0d13644a1e8


jberthold:
I tested this with the regression tests from `evm-semantics` (using a `nix` shell on a special branch https://github.com/runtimeverification/evm-semantics/compare/jb/test-k-change-for-t-division).
Unfortunately, the `sum-to-n-foundry.k` tests (which uses `/Int` to form `n * (n + 1) / 2` fails, with an SMT solver error on the following expression: 
```
{ true #Equals S:Int +Int N:Int *Int N:Int +Int 1 /Int 2 <Int pow256 }
#And { true #Equals 0 <Int N:Int }
#And { true #Equals N:Int <Int pow256 }
#And { true #Equals S:Int <Int pow256 }
#And { true #Equals 178 *Int N:Int <=Int GAS_AMT:Int }
#And { true #Equals 0 <=Int N:Int }
#And { true #Equals 0 <=Int S:Int }
```
The proof passes when using `divInt`, indicating that `/Int` (and likewise `%Int`) should be avoided if/when we merge this.

k-distribution/include/kframework/builtin/domains.md
tothtamas28, line 1: We should set up integration tests for this to ensure that concrete and symbolic execution agree on the semantics.

What should the input to the Booster `simplify` endpoint look like to force it to solve an equation with the SMT solver?

jberthold: It could be an artificial test that rewrites (or simplifies?) a state based on a combination of side conditions which requires the SMT solver to decide.
Generally, booster cannot combine rule side conditions or path conditions without involving the SMT solver.

tothtamas28: So for example, doing an `execute` on

```
{ <k> X %Int Y =Int Z </k> }
#And
{ X =Int a }
#And
{ Y =Int b }
#And
{ Z =Int c }
```

where `a`, `b`, `c` are `Int` tokens?

jberthold: Maybe it is sufficient to make rules dependent on the sign of an argument, and then use `%Int` to check the behaviour? Like :
```
syntax KItem ::= Start(Int) | Negative | Positive
rule Start(N) => Negative requires N < 0
rule Start(N) => Positive requires N > 0
```

~and then test executing with `Start( -1 %Int -2 )` and `Start(-1 /Int -2)` (and some more cases).~
EDIT: sorry, scratch that, we need to involve the SMT in the decision about the sign of the argument.

From [D.Leijen's note linked in the issue](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/divmodnote-letter.pdf):

| (D, d) | (qT , rT )  | (qF , rF )  | (qE, rE) |
|---------|---------|---------|---------|
|(+8, +3)| (+2, +2)| (+2, +2) |(+2, +2) |
|(+8, −3) |(−2, +2)| (−3, −1)| (−2, +2)
|(−8, +3) |(−2, −2) |(−3, +1) |(−3, +1)
| (−8, −3)|  (+2, −2)| (+2, −2)| (+3, +1)|
| (+1, +2) |( 0, +1) |( 0, +1) | ( 0, +1)
| (+1, −2) |( 0, +1) |(−1, −1) |( 0, +1)
| (−1, +2) | ( 0, −1) | (−1, +1) | (−1, +1)
| (−1, −2) | ( 0, −1) | ( 0, −1) | (+1, +1)

PetarMax:
> The proof passes when using `divInt`, indicating that `/Int` (and likewise `%Int`) should be avoided if/when we merge this.

I didn't quite get this, should be avoided in which context? For example, the division that should be used in KEVM is `/Int`.

jberthold:
> > The proof passes when using `divInt`, indicating that `/Int` (and likewise `%Int`) should be avoided if/when we merge this.
> 
> I didn't quite get this, should be avoided in which context? For example, the division that should be used in KEVM is `/Int`.

By "should be avoided" , I mean that in cases where the result will be the same for `divInt` and `/Int` (as it is the case in this `sum-to-n-foundry` proof, because [`N > 0` and the divisor is 2](https://github.com/runtimeverification/evm-semantics/compare/jb/test-k-change-for-t-division#diff-ee5537c4bd83526309c95fd3461f2d3c658b2c7478c93ef8606c71d61e5a4ec1L44-L68)) the `divInt` is preferred because it won't contribute to the SMT solver's headache.

jberthold:
I made a simple test (which currently fails on master):
The [test.k](https://github.com/user-attachments/files/16842427/test.k.txt) file as above, taking `Number(N:Int)` to `Positive` or `Negative` 

<summary> <details>

```
module TEST-SYNTAX
  imports INT-SYNTAX
  syntax KItem ::= Number(Int) [symbol(Number)]
                 | "Positive"  [symbol(Positive)]
                 | "Negative"  [symbol(Negative)]
endmodule

module TEST
  imports TEST-SYNTAX
  imports INT

  rule Number(N) => Positive requires 0 <Int N
  rule Number(N) => Negative requires N <Int 0

endmodule
```

</details></summary>,

and two JSON kore files as input:
* [negative-t-remainder-direct.json](https://github.com/user-attachments/files/16842452/negative-t-remainder-direct.json) with a K cell `Number(-8 %Int -3)`
* [negative-t-remainder-substituted.json](https://github.com/user-attachments/files/16842453/negative-t-remainder-substituted.json)  with a K cell `Number(N %Int M)` and path constraints `N ==Int -8 && M ==Int -3`
 

 Both should rewrite the K cell to `Negative` but the second one rewrites to `Positive`  on current master.

Unsure where to put this test. It should probably live within the `pyk` integration tests?

tothtamas28:
> I made a simple test (which currently fails on master)

I think the test semantics can be as simple as

```k
module INT-TEST
    imports INT
endmodule
```

Then we define a list of tuples `(Rel, a, b, c)` with the intended meaning `a Rel b = c`. Based on this list, we programmatically create the test inputs:
1. For the LLVM backend, `<k> a Rel b </k>` (with `<generatedTop>` and all the necessary boilerplate), and check if the final config is `<k> c </k>`.
2. For the Booster backend, `X Rel Y andBool X =Int a andBool Y =Int b` in a `simplify` request, and check if the result is `c`.

@jberthold, is (2) a reasonable setup, i.e. would it be decided by calling the SMT solver?

jberthold:
> > I made a simple test (which currently fails on master)
> 
> I think the test semantics can be as simple as
> 
> ```
> module INT-TEST
>     imports INT
>     configuration <k> $PGM:Int </k>
> endmodule
> ```
> 
> Then we define a list of tuples `(Rel, a, b, c)` with the intended meaning `a Rel b = c`. Based on this list, we programmatically create the test inputs:
> 
> 1. For the LLVM backend, `<k> a Rel b </k>` (with `<generatedTop>` and all the necessary boilerplate), and check if the final config is `<k> c </k>`.
> 2. For the Booster backend, `X Rel Y andBool X =Int a andBool Y =Int b` in a `simplify` request, and check if the result is `c`.
> 
> @jberthold, is (2) a reasonable setup, i.e. would it be decided by calling the SMT solver?

I am not sure this will do, actually, as we want the SMT solver to operate with the `Rel` expression (as translated to SMT operations, which are changed in this PR). The point of using a rewrite rule was that it would relegate evaluating the expression to the side condition (and depending on several conditions together so the SMT solver will do it) .

ehildenb:
Could we write tests that invoke the `get-model` endpoint of the backend instead? That could be used to test the SMT solver directly.

jberthold:
> Could we write tests that invoke the `get-model` endpoint of the backend instead? That could be used to test the SMT solver directly.

That's a good idea, because we can be _certain_ that the SMT solver will be invoked.

The input would be something like
```
X ==Int A Rel B
A ==Int a
B ==Int b
```
or even simply  `X ==Int a Rel b`, and we then check that `X` has the expected value `c`, for triples `(a, b, c)` from a tabulation of `Rel`. I will try to implement that later today.

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+
+if TYPE_CHECKING:
+    from typing import Final
+
+    from pyk.kore.rpc import KoreClient
+
+
+T_DIVISION_TEST_DATA: Final = (
+    # Op   ,  a,  b, a Op b, OpKore
+    ('/Int', +8, +3, +2, "Lbl'UndsSlsh'Int'Unds'"),
```
tothtamas28: Consider extracting the KORE symbols into constants.

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+        module SMT
+            imports INT
+        endmodule
+    """
+    KOMPILE_MAIN_MODULE = 'SMT'
+    KOMPILE_ARGS = {'syntax_module': 'SMT'}
+    LLVM_ARGS = {'syntax_module': 'SMT'}
+
+    @pytest.mark.parametrize(
+        'rel, a, b, c, rel_kore',
```
tothtamas28: Nitpick:

```suggestion
        'op, a, b, c, op_kore',
```

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+        ids=[f'{a} {r} {b} == {c}' for r, a, b, c, _ in T_DIVISION_TEST_DATA],
+    )
+    def test_smt_t_division(self, kore_client: KoreClient, rel: str, rel_kore: str, a: int, b: int, c: int) -> None:
+
+        # checks whether the SMT solver returns X = c for X = a rel b
+        pattern = Equals(
+            BOOL,
+            INT,
+            TRUE,
+            eq_int(App(rel_kore, (), (int_dv(a), int_dv(b))), EVar('x', INT)),
```
tothtamas28: Nitpick:

```suggestion
            eq_int(App(rel_kore, (), (int_dv(a), int_dv(b))), EVar('X', INT)),
```

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+    LLVM_ARGS = {'syntax_module': 'SMT'}
+
+    @pytest.mark.parametrize(
+        'rel, a, b, c, rel_kore',
+        T_DIVISION_TEST_DATA,
+        ids=[f'{a} {r} {b} == {c}' for r, a, b, c, _ in T_DIVISION_TEST_DATA],
+    )
+    def test_smt_t_division(self, kore_client: KoreClient, rel: str, rel_kore: str, a: int, b: int, c: int) -> None:
+
+        # checks whether the SMT solver returns X = c for X = a rel b
```
tothtamas28:
```suggestion
        """Check whether the SMT solver returns ``X = c`` for ``X = a op b``."""

```

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+        pattern = Equals(
```
tothtamas28:
```suggestion
        # Given
        pattern = Equals(
```

pyk/src/tests/integration/test_smt_hooks.py
tothtamas28, line 1: Nice! :sparkles:

pyk/src/tests/integration/test_smt_hooks.py
```diff
@@ -0,0 +1,80 @@
...
+from pyk.kore.syntax import App, Equals, EVar
+from pyk.testing import KoreClientTest
+
+if TYPE_CHECKING:
+    from typing import Final
+
+    from pyk.kore.rpc import KoreClient
+
+
+T_DIVISION_TEST_DATA: Final = (
```
tothtamas28: Please add a test that uses this same table to check the LLVM hook implementations.

Here's a similar (though a bit more complex) test as an example: https://github.com/runtimeverification/k/blob/42e03545c319847e7df84e6fbc196b2f51c148a0/pyk/src/tests/integration/test_krun.py#L87

jberthold: I have added a second test that uses `simplify`. This will call the HS-internal hooks for legacy kore and the LLVM library for booster.

tothtamas28: Nice, this way HS hooks are covered as well.

Please still add a test that calls `interpreter` directly. Then regardless of how `simplify` is implemented, we can be sure the LLVM backend hooks are exercised (even if there's a regression in Booster).

jberthold: Personally, I think this might be a bit of a scope creep...  checking the consistency of different backends was not the purpose of my change at all (it was just consistently wrong wrt. the SMT hook).
I can implement another test but that should be in the `test_krun.py` file where everything is set up for it. The tests I added inherit from `KoreClientTest` which implies a _Haskell_ backend compilation.

tothtamas28: > I think this might be a bit of a scope creep

I don't think so, we need to ensure that all implementations of these operations agree on the semantics.

But since the hooks are already exercised through Booster `simplify`, I can approve this PR right now if you prefer. In that case, please add an issue for this so that I can add the `interpret` tests later.

Also, before merging, please test with `evm-semantics` and `kontrol` to see if there's a regression due to the changes in `smt-hook`.

> I can implement another test but that should be in the `test_krun.py` file

Tests should be organized by what they test, and here, we're not testing `krun`, but whether `smt-hook` and `hook` agree on their semantics. Accordingly, the whole test data table and the definition test can be shared between the two tests.

> where everything is set up for it

There's not much to set up I think, you just need to add a new `KompiledTest` with the same string for the definition, and add a test method that calls `interpret`.

jberthold: Yes, probably not much to set up... but it takes me a long time to learn what it is that needs to be set up. you are probably in a better position to do that (I just tried for an hour or so but don't fully grasp the hierarchy of tests and what they can do).
So I'd prefer if you could approve this PR and we merge it, for efficiency.

tothtamas28 (approved):

pyk/src/tests/integration/test_smt_hooks.py
tothtamas28: Makes sense, approved.

ehildenb (approved):

ehildenb:
Awesome job @jberthold !

ehildenb:
Have we tested this downstream with KEVM and Kontrol?

jberthold:
> Have we tested this downstream with KEVM and Kontrol?

I have [tested `evm-semantics` regression tests](https://github.com/runtimeverification/evm-semantics/compare/jb/test-k-change-for-t-division#diff-ee5537c4bd83526309c95fd3461f2d3c658b2c7478c93ef8606c71d61e5a4ec1L44-L68), and had to make one small change in the `sum-to-n-foundry` test.
There may be more fall-out in `kontrol`, I will run the `kontrol` tests locally today (setup is a bit tricky).

jberthold:
> > Have we tested this downstream with KEVM and Kontrol?
> 
> I have [tested `evm-semantics` regression tests](https://github.com/runtimeverification/evm-semantics/compare/jb/test-k-change-for-t-division#diff-ee5537c4bd83526309c95fd3461f2d3c658b2c7478c93ef8606c71d61e5a4ec1L44-L68), and had to make one small change in the `sum-to-n-foundry` test. There may be more fall-out in `kontrol`, I will run the `kontrol` tests locally today (setup is a bit tricky).

Tested `kontrol`, too, no failures observed so no change required when upgrading.

Issue #4602, opened 2024-08-21
What is `KCFG.Cover.csubst` supposed to be?


Stevengre:
I suspect that `KCFG.Cover.csubst` should be able to obtain the `source` by applying it to the `target`. However, in its current form, the `match_with_constraint` in `create_cover` doesn’t achieve this when it’s set to `None`.

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/kcfg/kcfg.py#L854-L863

There is a non-None call here, which should have the same effect as a non-empty call. However, based on [this documentation](https://github.com/runtimeverification/haskell-backend/blob/2faf6c1d71de054f4630e3489efb4631a0f320b5/docs/2022-07-18-JSON-RPC-Server-API.md?plain=1#L361), it doesn’t seem to behave the same way.

https://github.com/runtimeverification/k/blob/0bfb62cf00b71470ed80a42be339d5f87c858e54/pyk/src/pyk/proof/reachability.py#L193-L194

So what exactly does it do? How can we make it consistent? And do we even need `csubst` if `match_with_constraint` works correctly?

tothtamas28:
The only call to `create_cover` with `csubst=None`: https://github.com/runtimeverification/k/blob/eb0b1815af0a8c08773ce43c70f9f307624d195f/pyk/src/pyk/kcfg/kcfg.py#L530-L533

tothtamas28:
TL;DR: When the `APRPRover` establishes coverage between nodes, the `KCFG.Cover.csubst.constraints` are the conjuncts of the `predicate` from the [KORE RPC `implies` response](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md#correct-response-1).

---

The other code path:

`KCFG.Cover` is instantiated in `APRProof.commit`: https://github.com/runtimeverification/k/blob/eb0b1815af0a8c08773ce43c70f9f307624d195f/pyk/src/pyk/proof/reachability.py#L234-L235

`APRProofSubsumeResult` is instantiated in `APRProver.step_proof`: https://github.com/runtimeverification/k/blob/eb0b1815af0a8c08773ce43c70f9f307624d195f/pyk/src/pyk/proof/reachability.py#L811-L820

[...]

PR #4617, opened 2024-09-02
Create Split without CSubst

Stevengre:
1. generate `CSubst` from the targets for `Split`
2. fix a bug in `CSubst.apply`.

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -263,13 +263,12 @@ def cterm_match(cterm1: CTerm, cterm2: CTerm) -> CSubst | None:
     Returns:
         A `CSubst` which can instantiate `cterm1` to `cterm2`, or `None` if no such `CSubst` exists.
     """
+    # todo: delete this function and use cterm1.match_with_constraint(cterm2) directly after closing #4496
     subst = cterm1.config.match(cterm2.config)
     if subst is None:
         return None
     source_constraints = [subst(c) for c in cterm1.constraints]
-    constraints = [c for c in cterm2.constraints if c not in source_constraints] + [
-        c for c in source_constraints if c not in cterm2.constraints
-    ]
+    constraints = [c for c in cterm2.constraints if c not in source_constraints]
     return CSubst(subst, constraints)
 
 
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -337,7 +337,7 @@ class CSubst:
         constraints = [self.subst(constraint) for constraint in cterm.constraints] + list(self.constraints)
         return CTerm(config, constraints)
 
-    def __call__(self, cterm: CTerm):
+    def __call__(self, cterm: CTerm) -> CTerm:
         """Overload for `CSubst.apply`."""
         return self.apply(cterm)
 
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -253,6 +253,26 @@ class CTerm:
         return CTerm(self.config, new_constraints)
 
 
+def cterm_match(cterm1: CTerm, cterm2: CTerm) -> CSubst | None:
+    """Find a substitution which can instantiate `cterm1` to `cterm2`.
+
+    Args:
+        cterm1: `CTerm` to instantiate to `cterm2`.
+        cterm2: `CTerm` to instantiate `cterm1` to.
+
+    Returns:
+        A `CSubst` which can instantiate `cterm1` to `cterm2`, or `None` if no such `CSubst` exists.
+    """
+    subst = cterm1.config.match(cterm2.config)
+    if subst is None:
+        return None
+    source_constraints = [subst(c) for c in cterm1.constraints]
+    constraints = [c for c in cterm2.constraints if c not in source_constraints] + [
+        c for c in source_constraints if c not in cterm2.constraints
+    ]
+    return CSubst(subst, constraints)
+
+
 def anti_unify(state1: KInner, state2: KInner, kdef: KDefinition | None = None) -> tuple[KInner, Subst, Subst]:
     """Return a generalized state over the two input states.
 
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -333,8 +333,8 @@ class CSubst:
 
     def apply(self, cterm: CTerm) -> CTerm:
         """Apply this `CSubst` to the given `CTerm` (instantiating the free variables, and adding the constraints)."""
-        _kast = self.subst(cterm.kast)
-        return CTerm(_kast, [self.constraint])
+        config = self.subst(cterm.config)
+        return CTerm(config, [self.constraint])
 
 
 def cterm_build_claim(
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -38,7 +38,7 @@ class CTerm:
 
     Contains the data:
     - `config`: the _configuration_ (structural component of the state, potentially containing free variabls)
-    - `constraints`: conditiions which limit/constraint the free variables from the `config`
+    - `constraints`: conditions which limit/constraint the free variables from the `config`
     """
 
     config: KInner  # TODO Optional?
@@ -336,6 +336,10 @@ class CSubst:
         config = self.subst(cterm.config)
         constraints = [self.subst(constraint) for constraint in cterm.constraints] + list(self.constraints)
         return CTerm(config, constraints)
+
+    def __call__(self, cterm: CTerm):
+        """Overload for `CSubst.apply`."""
+        return self.apply(cterm)
 
 
 def cterm_build_claim(
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -334,7 +334,8 @@ class CSubst:
     def apply(self, cterm: CTerm) -> CTerm:
         """Apply this `CSubst` to the given `CTerm` (instantiating the free variables, and adding the constraints)."""
         config = self.subst(cterm.config)
-        return CTerm(config, [self.constraint])
+        constraints = [self.subst(constraint) for constraint in cterm.constraints] + list(self.constraints)
+        return CTerm(config, constraints)
 
 
 def cterm_build_claim(
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -1045,9 +1045,15 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         source = self.node(source_id)
         targets = [self.node(nid) for nid in target_ids]
         substs = [source.cterm.config.match(target.cterm.config) for target in targets]
+        csubsts: list[CSubst] = []
         if None in substs:
             return None
-        csubsts = [CSubst(subst, target.cterm.constraints) for subst, target in zip(substs, targets, strict=True)]
+        for subst, target in zip(substs, targets, strict=True):
+            constraints: list[KInner] = []
+            for c in target.cterm.constraints:
+                if c not in source.cterm.constraints:
+                    constraints.append(c)
+            csubsts.append(CSubst(subst, constraints))
         return self.create_split(source.id, zip(target_ids, csubsts, strict=True))
 
     def ndbranches(self, *, source_id: NodeIdLike | None = None, target_id: NodeIdLike | None = None) -> list[NDBranch]:
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -1050,7 +1050,10 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
             subst = source.cterm.config.match(target.cterm.config)
             if subst is None:
                 return None
-            constraints = [c for c in target.cterm.constraints if c not in source.cterm.constraints]
+            source_constraints = [subst(c) for c in source.cterm.constraints]
+            constraints = [c for c in target.cterm.constraints if c not in source_constraints] + [
+                c for c in source_constraints if c not in target.cterm.constraints
+            ]
             csubsts.append(CSubst(subst, constraints))
 
         return self.create_split(source.id, zip(target_ids, csubsts, strict=True))
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -1041,8 +1041,9 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         return split
 
     def create_split_by_nodes(self, source_id: NodeIdLike, target_ids: Iterable[NodeIdLike]) -> KCFG.Split | None:
+        """Create a split without crafting a CSubst."""
         source = self.node(source_id)
-        targets = [self.node(nid) for nid in list(target_ids)]
+        targets = [self.node(nid) for nid in target_ids]
         substs = [source.cterm.config.match(target.cterm.config) for target in targets]
         if None in substs:
             return None
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -9,6 +9,8 @@ from functools import reduce
 from threading import RLock
 from typing import TYPE_CHECKING, Final, List, Union, cast, final
 
+from pyk.cterm.cterm import cterm_match
+
 from ..cterm import CSubst, CTerm, cterm_build_claim, cterm_build_rule
 from ..kast import EMPTY_ATT
 from ..kast.inner import KApply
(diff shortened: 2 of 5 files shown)


pyk/src/pyk/cterm/cterm.py
```diff
@@ -333,8 +333,8 @@ def add_constraint(self, constraint: KInner) -> CSubst:
 
     def apply(self, cterm: CTerm) -> CTerm:
         """Apply this `CSubst` to the given `CTerm` (instantiating the free variables, and adding the constraints)."""
-        _kast = self.subst(cterm.kast)
-        return CTerm(_kast, [self.constraint])
+        config = self.subst(cterm.config)
+        return CTerm(config, [self.constraint])
```
Stevengre: Previously, `CSubst.apply` did not support `CTerm` with non-empty constraints. The current approach substitutes the `CTerm`'s config and completely replaces its constraints with those from `CSubst`. The reason for not combining the `CTerm`'s constraints with `CSubst`'s constraints is that, at the `CSubst` level, all Boolean expressions are treated as uninterpreted functions, which cannot be computed. 

For example, if the source has the constraint `0 < X <= 10`, the target has `0 < Y <= 5`, and `CSubst` has `0 < Y <= 5`, the combined constraint would be `0 < Y <= 10 /\ 0 < Y <= 5`. Although logically equivalent to `0 < Y <= 5`, the KCFG does not interpret these Boolean expressions, making it impossible to verify their equivalence.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1040,6 +1040,16 @@ def create_split(self, source_id: NodeIdLike, splits: Iterable[tuple[NodeIdLike,
...
 
+    def create_split_by_nodes(self, source_id: NodeIdLike, target_ids: Iterable[NodeIdLike]) -> KCFG.Split | None:
+        """Create a split without crafting a CSubst."""
+        source = self.node(source_id)
+        targets = [self.node(nid) for nid in target_ids]
+        substs = [source.cterm.config.match(target.cterm.config) for target in targets]
+        if None in substs:
+            return None
+        csubsts = [CSubst(subst, target.cterm.constraints) for subst, target in zip(substs, targets, strict=True)]
+        return self.create_split(source.id, zip(target_ids, csubsts, strict=True))
```
Stevengre, line 1053: In the substitution-based approach of `CSubst.apply`, we directly set the `CSubst` for the split to the constraints of the target.

If `CSubst.apply` were to use a combination approach, we could set the `CSubst` for the split to `target.constraints - source.constraints`.

pyk/src/tests/unit/test_kcfg.py
```diff
@@ -849,3 +866,50 @@ def test_no_cell_rewrite_to_dots() -> None:
...
+def test_create_split_by_nodes(kcfg: KCFG, source_id: int, target_ids: Iterable[int]) -> None:
+    # When
+    new_split = kcfg.create_split_by_nodes(source_id, target_ids)
+
+    # Then
+    assert new_split
+    assert new_split.source == kcfg.node(source_id)
+    assert new_split.targets == tuple(kcfg.node(target_id) for target_id in target_ids)
+    for target_id, csubst in new_split.splits.items():
+        assert csubst.apply(kcfg.node(source_id).cterm) == kcfg.node(target_id).cterm
```
Stevengre: Here, we provide a specification for a correct Split.

Stevengre:
Given this, I feel that `CSubst` might not be necessary. Instead, we could extract them from the nodes, eliminating the need to store them separately.

pyk/src/pyk/cterm/cterm.py
ehildenb: We should have a unit test of the new behaivor. And we should keep both constraints (from the original `CTerm` and the new `CSubst`). Tests can go here: https://github.com/runtimeverification/k/blob/master/pyk/src/tests/unit/test_cterm.py. Make sure to write a test scenario for every tricky case you can think of, in `test_csubst_apply` or something similar.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: We should use the combination approach, and do the delta of constraints instead. A `CSubst` applied to a `CTerm` should apply the substitution and append the constraints, so that we're guaranteed to always go to a more specific state.

pyk/src/pyk/cterm/cterm.py
Stevengre: Thank you! #4618 will offer these tests.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -333,8 +333,13 @@ def add_constraint(self, constraint: KInner) -> CSubst:
 
     def apply(self, cterm: CTerm) -> CTerm:
         """Apply this `CSubst` to the given `CTerm` (instantiating the free variables, and adding the constraints)."""
-        _kast = self.subst(cterm.kast)
-        return CTerm(_kast, [self.constraint])
+        config = self.subst(cterm.config)
+        constraints = [self.subst(constraint) for constraint in cterm.constraints] + list(self.constraints)
+        return CTerm(config, constraints)
```
Stevengre, line 357: The return `CTerm`'s constraints are `CSubst` + original `CTerm`'s.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1040,6 +1040,24 @@ def create_split(self, source_id: NodeIdLike, splits: Iterable[tuple[NodeIdLike,
...
+            subst = source.cterm.config.match(target.cterm.config)
+            if subst is None:
+                return None
+            source_constraints = [subst(c) for c in source.cterm.constraints]
+            constraints = [c for c in target.cterm.constraints if c not in source_constraints] + [
+                c for c in source_constraints if c not in target.cterm.constraints
+            ]
+            csubsts.append(CSubst(subst, constraints))
+
+        return self.create_split(source.id, zip(target_ids, csubsts, strict=True))
```
Stevengre, line 1053: Create a split edge from source to targets. This won't modify the nodes of source and targets. So users need to ensure that source and targets conform to the implication relationship. The generated CSubst contains the delta constraints between source and targets. Therefore, CSubst.apply(source) may not be equal to target, but it can definitely obtain a target that conforms to the implication relationship.

pyk/src/tests/unit/test_kcfg.py
```diff
@@ -849,3 +850,69 @@ def test_no_cell_rewrite_to_dots() -> None:
...
+        ),
+    ),
+    (CTerm(config('X')), [CTerm(config('Y')), CTerm(KApply('<bot>', [KVariable('Z')]))], None),
+    (
+        CTerm(config('X'), [ge('X', 0), lt('X', 10)]),
+        [CTerm(config('Y'), [ge('Y', 0)]), CTerm(config('Z'), [ge('Z', 5)])],
+        KCFG.Split(
+            KCFG.Node(1, CTerm(config('X'), [ge('X', 0), lt('X', 10)])),
+            [
+                (KCFG.Node(2, CTerm(config('Y'), [ge('Y', 0)])), CSubst(Subst({'X': KVariable('Y')}), [lt('Y', 10)])),
```
Stevengre: Note that this target doesn't meet the implication relationship with source. So the CTerm of target and CSubst.apply(target) are not logically equal. But source -> CSubst.apply(target) can always be true.

pyk/src/tests/unit/test_kcfg.py
```diff
@@ -849,3 +850,69 @@ def test_no_cell_rewrite_to_dots() -> None:
...
+    # Given
+    cfg = KCFG()
+    source_node = cfg.create_node(source)
+    target_nodes = [cfg.create_node(target) for target in targets]
+
+    # When
+    actual = cfg.create_split_by_nodes(source_node.id, [n.id for n in target_nodes])
+
+    # Then
+    assert actual == expected
```
Stevengre, line 976: Test cases for `create_split_by_nodes`

pyk/src/tests/unit/test_cterm.py
```diff
@@ -186,3 +186,41 @@ def test_from_kast(test_id: str, kast: KInner, expected: CTerm) -> None:
...
+@pytest.mark.parametrize(
+    'term,subst,expected',
+    APPLY_TEST_DATA,
+)
+def test_csubst_apply(term: CTerm, subst: CSubst, expected: CTerm) -> None:
+    # When
+    actual = subst(term)
+
+    # Then
+    assert actual == expected
```
Stevengre, line 226: Test cases for `csubst.apply`

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -1040,6 +1040,24 @@ def create_split(self, source_id: NodeIdLike, splits: Iterable[tuple[NodeIdLike,
...
+
+        for target in targets:
+            subst = source.cterm.config.match(target.cterm.config)
+            if subst is None:
+                return None
+            source_constraints = [subst(c) for c in source.cterm.constraints]
+            constraints = [c for c in target.cterm.constraints if c not in source_constraints] + [
+                c for c in source_constraints if c not in target.cterm.constraints
+            ]
+            csubsts.append(CSubst(subst, constraints))
```
ehildenb: What if instead we use `CTerm.match_with_constraint`: https://github.com/runtimeverification/k/blob/42e03545c319847e7df84e6fbc196b2f51c148a0/pyk/src/pyk/cterm/cterm.py#L162

```suggestion
        csubsts = [not_none(source.cterm.match_with_constraint(target.cterm)) for target in targets]
```

You can import `not_none` from `..utils`.

pyk/src/tests/unit/test_cterm.py
ehildenb: Looks great, thank you!

pyk/src/tests/unit/utils.py
```diff
@@ -16,3 +18,19 @@
...
+def ge(var: str, n: int) -> KApply:
+    return mlEqualsTrue(geInt(KVariable(var), intToken(n)))
+
+
+def config(var: str) -> KApply:
+    return KApply('<top>', KVariable(var))
+
+
+def config_int(n: int) -> KApply:
+    return KApply('<top>', intToken(n))
```
ehildenb: I think these should be renamed to make them more specific. `lt` should be `lt_ml_true` and `ge` should be something like `ge_ml_true`.

ehildenb: Also, do we need `config` and `config_int`, could we instead use `k` that's already defined above for all your examples? I think it would work, you can just import the variables already used here.

Where you have:

```
config('X') => k(x)
```

And where you have:

```
config_int(3) => k(intToken(3))
```

Then we don't need more helpers.

pyk/src/tests/unit/test_kcfg.py
ehildenb: I think, in that case, we should actually have it fail to build the split.

The semantics of a `split` are that:

- The split targets should be mutually exclusive (no overlap).
- The split targets should be total on the original (the union of the split targets == the split source).

pyk/src/tests/unit/test_kcfg.py
```diff
@@ -849,3 +850,69 @@ def test_no_cell_rewrite_to_dots() -> None:
...
+        [CTerm(config('X')), CTerm(config('Y'))],
+        None,  # todo: support split from top, because top means anything can be matched
+    ),
+    (
+        CTerm.top(),
+        [CTerm.top(), CTerm.top()],
+        KCFG.Split(
+            KCFG.Node(1, CTerm.top()), [(KCFG.Node(2, CTerm.top()), CSubst()), (KCFG.Node(3, CTerm.top()), CSubst())]
+        ),
+    ),
```
ehildenb, line 901: This one violates that the target of the splits are mutually exclusive, since nodes 2 and 3 overlap.

pyk/src/tests/unit/test_kcfg.py
```diff
@@ -849,3 +850,69 @@ def test_no_cell_rewrite_to_dots() -> None:
...
+        [CTerm(config('X')), CTerm(config('Y')), CTerm(config('Z'))],
+        KCFG.Split(
+            KCFG.Node(1, CTerm(config('X'))),
+            [
+                (KCFG.Node(2, CTerm(config('X'))), CSubst(Subst({'X': KVariable('X')}))),
+                (KCFG.Node(3, CTerm(config('Y'))), CSubst(Subst({'X': KVariable('Y')}))),
+                (KCFG.Node(4, CTerm(config('Z'))), CSubst(Subst({'X': KVariable('Z')}))),
+            ],
+        ),
+    ),
```
ehildenb: This one also violates that the targets are mutually exclusive.

pyk/src/tests/unit/test_kcfg.py
ehildenb: I think here, we should just put a comment on the ones that violate this condition, and otherwise move forward. Eventually we should enforce that split nodes are actually semantically splits, and then the comments you put here will be useful to know that these tests can be deleted/ignored.

pyk/src/tests/unit/test_kcfg.py
ehildenb: Let's put a comment that it violates that condition, or update the test to not violate that condition (by having constraints on the target nodes that make them mutually exclusive), eg:

```
[CTerm(config('X'), [gt('X', 3), le('X', 10)]), CTerm(config('Y'), [le('X', 3)]), CTerm(config('Z'), [gt('X', 10)])])],
```

pyk/src/tests/unit/test_kcfg.py
Stevengre: It's impossible to check these properties without the Haskell backend and semantics definition. As without them, the functions inside CTerms are uninterpreted. 

We can introduce a tool called `CFGChecker` in the kcfg package for automatic checking. I want the `KCFG` class to be only used for its syntax definition and syntax-level (static-semantics level) manipulation. Therefore, users have the duty to ensure the result kcfg is correct. For example, like `KCFG.create_split`, it might produce incorrect targets and even wrong csubsts.

Also, I think edge types in kcfg can be reduced to rewrite and implies (or refine). It doesn't seem necessary to distinguish these edge types, except for printing. If necessary, we can identify them at runtime.

Stevengre: I want to make it simple and I think it will be acceptable for merging nodes. Because if a mergeable split is correct, the generated split will always be correct.

pyk/src/pyk/kcfg/kcfg.py
Stevengre: `CTerm.match_with_constraint` doesn't give the same result as expected. It always returns a true imply as the constraints for CSubst. That's why we have #4499 and #4602. Since `match_with_constraints` is already in use, modifying it might cause downstream issues. Additionally, I'd like to provide merge node functionality as soon as possible. So, I give up refactoring `match_with_constraints` and instead provide this implementation here.

After we have the result of the discussion at #4602, we can refactor `match_with_constraints` and `create_split_by_nodes`.

pyk/src/tests/unit/utils.py
Stevengre: Thank you! I didn't notice https://github.com/runtimeverification/k/blob/42e03545c319847e7df84e6fbc196b2f51c148a0/pyk/src/pyk/kast/inner.py#L123-L125

Stevengre: As for `lt`, how about renaming it to `lt_ml`? I feel this is the same as adding the meaning of `true`. In fact, I don't understand the difference between `mlEqualsTrue(geInt(KVariable(var), intToken(n)))` and `geInt(KVariable(var), intToken(n))`.

Stevengre: Finished!

pyk/src/tests/unit/test_kcfg.py
Stevengre: I have left some comments for these test case.

pyk/src/tests/unit/test_kcfg.py
Stevengre: I have left some comments for these test case, and introduced a new correct case.

ehildenb (approved):

pyk/src/pyk/kcfg/kcfg.py
ehildenb: Please leave a comment here saying that when #4496 is closed (link to it) that we can change this to use `match_with_constarint`

pyk/src/tests/unit/test_kcfg.py
ehildenb: Makes sense.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +253,26 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+    Returns:
+        A `CSubst` which can instantiate `cterm1` to `cterm2`, or `None` if no such `CSubst` exists.
+    """
+    subst = cterm1.config.match(cterm2.config)
+    if subst is None:
+        return None
+    source_constraints = [subst(c) for c in cterm1.constraints]
+    constraints = [c for c in cterm2.constraints if c not in source_constraints] + [
+        c for c in source_constraints if c not in cterm2.constraints
+    ]
```
ehildenb: I think the second line here (`c for c in source_constraints if c not in cterm2.constraints`) is probably not needed. It won't add any constraints to `cterm1` when the csubst is applied that are new.

pyk/src/pyk/kcfg/kcfg.py
Stevengre: Done!

pyk/src/pyk/cterm/cterm.py
Stevengre: Yes, deleted this line!

PR #4621, opened 2024-09-03
Merge CTerm

Stevengre:
Merge two cterms into one for merging node functionality #4425.

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -271,7 +271,7 @@ def merge_cterms(t1: CTerm, t2: CTerm) -> CTerm | None:
         # cannot merge two configurations with different structure
         return None
 
-    new_subst: Subst = Subst({})
+    new_subst = Subst({})
     new_t1_subst = Subst({})
     new_t2_subst = Subst({})
 
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -21,7 +21,7 @@ from ..kast.manip import (
     split_config_from,
 )
 from ..prelude.k import GENERATED_TOP_CELL
-from ..prelude.kbool import TRUE, andBool, orBool
+from ..prelude.kbool import andBool, orBool
 from ..prelude.ml import is_bottom, is_top, mlAnd, mlBottom, mlEquals, mlEqualsTrue, mlImplies, mlTop
 from ..utils import unique
 
@@ -101,6 +101,11 @@ class CTerm:
     def is_bottom(self) -> bool:
         """Check if a given `CTerm` is trivially empty."""
         return is_bottom(self.config, weak=True) or any(is_bottom(cterm, weak=True) for cterm in self.constraints)
+
+    @property
+    def constraint(self) -> KInner:
+        """Return the set of constraints as a single flattened constraint using `mlAnd`."""
+        return mlAnd(self.constraints)
 
     @staticmethod
     def _constraint_sort_key(term: KInner) -> tuple[int, str]:
@@ -289,7 +294,7 @@ def merge_cterms(t1: CTerm, t2: CTerm) -> CTerm | None:
     for new_subst, t in [(new_t1_subst, t1), (new_t2_subst, t2)]:
         if new_subst:
             antecedent = mlAnd([mlEquals(KVariable(cell), new_subst[cell]) for cell in new_subst])
-            consequent = mlAnd(t.constraints) if t.constraints else TRUE
+            consequent = t.constraint
             new_constraints.append(mlImplies(antecedent, consequent))
 
     return CTerm(new_config, new_constraints)
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -6,7 +6,7 @@ from itertools import chain
 from typing import TYPE_CHECKING
 
 from ..kast import KInner
-from ..kast.inner import KApply, KRewrite, KToken, Subst, bottom_up
+from ..kast.inner import KApply, KRewrite, KToken, KVariable, Subst, bottom_up, var_occurrences
 from ..kast.manip import (
     abstract_term_safely,
     build_claim,
@@ -21,8 +21,8 @@ from ..kast.manip import (
     split_config_from,
 )
 from ..prelude.k import GENERATED_TOP_CELL
-from ..prelude.kbool import andBool, orBool
-from ..prelude.ml import is_bottom, is_top, mlAnd, mlBottom, mlEqualsTrue, mlImplies, mlTop
+from ..prelude.kbool import TRUE, andBool, orBool
+from ..prelude.ml import is_bottom, is_top, mlAnd, mlBottom, mlEquals, mlEqualsTrue, mlImplies, mlTop
 from ..utils import unique
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +253,48 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+    new_config = new_subst(t1_config)
+
+    new_constraints: list[KInner] = []
+    for new_subst, t in [(new_t1_subst, t1), (new_t2_subst, t2)]:
+        if new_subst:
+            antecedent = mlAnd([mlEquals(KVariable(cell), new_subst[cell]) for cell in new_subst])
+            consequent = mlAnd(t.constraints) if t.constraints else TRUE
+            new_constraints.append(mlImplies(antecedent, consequent))
+
+    return CTerm(new_config, new_constraints)
```
Stevengre, line 298: Main implementation of `merge_cterms`.

1. Return `None` if the configuration structures are different.
2. Replace cell content into a new free variable if there are free variables or different between them.
3. The constraints will be `CELL_NAME == SUBST /\ ... -> source.constraint /\ ...`

pyk/src/tests/unit/utils.py
```diff
@@ -16,3 +18,19 @@
...
+def ge(var: str, n: int) -> KApply:
+    return mlEqualsTrue(geInt(KVariable(var), intToken(n)))
+
+
+def config(var: str) -> KApply:
+    return KApply('<top>', KVariable(var))
+
+
+def config_int(n: int) -> KApply:
+    return KApply('<top>', intToken(n))
```
Stevengre: Provide some sugars.

pyk/src/tests/unit/test_cterm.py
```diff
@@ -186,3 +188,104 @@ def test_from_kast(test_id: str, kast: KInner, expected: CTerm) -> None:
...
+)
+
+
+@pytest.mark.parametrize('t1,t2,expected', MERGE_TEST_DATA, ids=count())
+def test_cterm_merge(t1: CTerm, t2: CTerm, expected: CTerm) -> None:
+    # When
+    merged = merge_cterms(t1, t2)
+
+    # Then
+    assert merged == expected
```
Stevengre, line 290: Provide several test cases for `merge_cterms`

pyk/src/tests/unit/utils.py
tothtamas28: Some of these (or something similar) is already available in: https://github.com/runtimeverification/k/blob/847da52e27d54434df0d1dad89146c8d0e371840/pyk/src/pyk/prelude/kint.py

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +253,48 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+    """
+    # check all cells in t1 and t1, if they are the same, keep them, otherwise, create a new free variable for them
+    t1_config, t1_subst = split_config_from(t1.config)
+    t2_config, t2_subst = split_config_from(t2.config)
+
+    if t1_config != t2_config:
+        # cannot merge two configurations with different structure
+        return None
+
+    new_subst: Subst = Subst({})
```
tothtamas28:
```suggestion
    new_subst = Subst({})
```

pyk/src/tests/unit/test_cterm.py
tothtamas28: Nice!

tothtamas28 (approved):

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +253,48 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+            # create a new free variable for the cell
+            new_t1_subst = new_t1_subst * Subst({cell: t1_subst[cell]})
+            new_t2_subst = new_t2_subst * Subst({cell: t2_subst[cell]})
+    new_config = new_subst(t1_config)
+
+    new_constraints: list[KInner] = []
+    for new_subst, t in [(new_t1_subst, t1), (new_t2_subst, t2)]:
+        if new_subst:
+            antecedent = mlAnd([mlEquals(KVariable(cell), new_subst[cell]) for cell in new_subst])
+            consequent = mlAnd(t.constraints) if t.constraints else TRUE
```
ehildenb: `TRUE` is in `Bool`, it should instead by `mlTop()` I believe.

How about:

```suggestion
            consequent = t.constraint
```

See: https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/cterm/cterm.py#L326

pyk/src/tests/unit/utils.py
Stevengre: I think I used some functions in kint to make my test cases simpler. If I use kint directly, test cases will be complex.

pyk/src/pyk/cterm/cterm.py
Stevengre: Done!

pyk/src/pyk/cterm/cterm.py
Stevengre: Done! Added a property function in cterm to support this call.

pyk/src/tests/unit/utils.py
tothtamas28: You're right, I didn't notice you were already using these functions inside the helpers.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +258,48 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+        else:
+            # create a new free variable for the cell
+            new_t1_subst = new_t1_subst * Subst({cell: t1_subst[cell]})
+            new_t2_subst = new_t2_subst * Subst({cell: t2_subst[cell]})
+    new_config = new_subst(t1_config)
+
+    new_constraints: list[KInner] = []
+    for new_subst, t in [(new_t1_subst, t1), (new_t2_subst, t2)]:
+        if new_subst:
+            antecedent = mlAnd([mlEquals(KVariable(cell), new_subst[cell]) for cell in new_subst])
```
ehildenb: I think this can be simplified as well? See https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kast/inner.py#L753

```suggestion
            antecedent = new_subst.ml_pred
```

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +258,48 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+
+    new_subst = Subst({})
+    new_t1_subst = Subst({})
+    new_t2_subst = Subst({})
+
+    for cell in t1_subst:
+        if t1_subst[cell] == t2_subst[cell] and not var_occurrences(t1_subst[cell]):
+            # keep the cell if it is the same & exists no free variables
+            new_subst = new_subst * Subst({cell: t1_subst[cell]})
+        else:
```
ehildenb: Why do we have the condition `not var_occurances(t1_subst[cell])`? If the two configurations have the same substitution for a given cell, can't we just apply that substitution instead of having it become part of the side-condition?

Stevengre: Previously, I thought not adding this determination would cause errors in the following example, but after providing it, I found it wouldn't. I will modify this and try whether when replacing X in CT2 with Y, <a> can still be replaced instead of using a free variable.

Given:

CT1
```xml
<a> X + 1 </a>
<b> X - 1 </b>
requires X > 10
```

CT2
```xml
<a> X + 1 </a>
<b> X </b>
requires X < 10
```

If X + Int 1 is directly retained, the result is:

```xml
<a> X +Int 1 </a>
<b> Y </b>
Y == X -> X > 10 /\ Y == X -Int 1 -> X < 10
```

pyk/src/pyk/cterm/cterm.py
Stevengre: Thank you for point it out! I've changed that!

pyk/src/pyk/cterm/cterm.py
Stevengre: Done!

pyk/src/pyk/cterm/cterm.py
```diff
@@ -253,6 +258,46 @@ def remove_useless_constraints(self, keep_vars: Iterable[str] = ()) -> CTerm:
...
+            new_t1_subst = new_t1_subst * Subst({cell: t1_subst[cell]})
+            new_t2_subst = new_t2_subst * Subst({cell: t2_subst[cell]})
+    new_config = new_subst(t1_config)
+
+    new_constraints: list[KInner] = []
+    for new_subst, t in [(new_t1_subst, t1), (new_t2_subst, t2)]:
+        if new_subst:
+            new_constraints.append(mlImplies(new_subst.ml_pred, t.constraint))
+
+    return CTerm(new_config, new_constraints)
```
ehildenb, line 298: This seems like a less powerful version of `CTerm.anti_unify`. Basically, instead of `merge_cterms(cterm1, cterm2)`, you can do:

```
cterm, csubst1, csubst2 = cterm1.anti_unify(cterm2)
```

You'll have that `csubst1(cterm) == cterm1`, and `csubst2(cterm) == cterm2`, and you can use the data returned there to compute what you need for the merged nodes.

ehildenb: Here are some tests of antiunification: https://github.com/runtimeverification/k/blob/42e03545c319847e7df84e6fbc196b2f51c148a0/pyk/src/tests/integration/proof/test_imp.py#L1296
https://github.com/runtimeverification/k/blob/42e03545c319847e7df84e6fbc196b2f51c148a0/pyk/src/tests/integration/proof/test_imp.py#L1247

Stevengre: Thank you! I looked at the code and tests of `anti_unify` and feel that `anti_unify` can indeed be directly used. It's just that the generated `csubst` I can't use. Perhaps I can close this PR.

PR #4625, opened 2024-09-05
Cleanups to handling ml predicates and substitutions

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/pull/4631~
~Blocked on: https://github.com/runtimeverification/k/pull/4630~
~Blocked on: https://github.com/runtimeverification/k/pull/4633~

While reviewing and going over https://github.com/runtimeverification/k/pull/4621 with @Stevengre , it became somewhat clear that how we handle turning substitions into ML predicates is a bit dirty. This attempts to clean this up a bit. Where potentially breaking changes to API are introduced here, I've checked if it affects the following repos when I mention "downstream" below: `evm-semantics kontrol wasm-semantics riscv-semantics mir-semantics`.

In particular:

- The function `CTerm.anti_unify` has a simplification where it reuses a function from `kast.manip` instead of reimplementing it.
- The functions `CSubst.from_pred` and `CSubst.pred` are added, as replacements for `Subst.ml_pred`. This is because `Subst.ml_pred` doesn't have a good way to produce correctly sorted predicates, because it's in module `kast.inner`.
  - `Subst.ml_pred` is removed, and tests are updated to use the new `CSubst` variant. None of the downstream repositories use `Subst.ml_pred` directly.
  - The new `CSubst.pred` correctly sorts the generated `#Equals` clauses, defaulting to `K` sort or if a `KDefinition` is supplied using it to do sort inference. It also provides options for controlling whether we include the substitution or the constraints in the generated predicate.
  - A test is added for a `CSubst.pred` case which caused a bug in the integration tests dealing with identity substitutions.
- The `CTermSymbolic.implies` function is updated to reuse `CSubst.from_pred` instead of reimplementing it.
  - On the case of duplicate entries, the first is kept and the latter are made as predicates.

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -218,17 +218,7 @@ class CTerm:
             if KToken('true', 'Bool') not in [disjunct_lhs, disjunct_rhs]:
                 new_cterm = new_cterm.add_constraint(mlEqualsTrue(orBool([disjunct_lhs, disjunct_rhs])))
 
-        new_constraints = []
-        fvs = new_cterm.free_vars
-        len_fvs = 0
-        while len_fvs < len(fvs):
-            len_fvs = len(fvs)
-            for constraint in common_constraints:
-                if constraint not in new_constraints:
-                    constraint_fvs = free_vars(constraint)
-                    if any(fv in fvs for fv in constraint_fvs):
-                        new_constraints.append(constraint)
-                        fvs = fvs | constraint_fvs
+        new_constraints = remove_useless_constraints(common_constraints, new_cterm.free_vars)
 
         for constraint in new_constraints:
             new_cterm = new_cterm.add_constraint(constraint)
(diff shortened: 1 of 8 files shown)


pyk/src/pyk/cterm/cterm.py
```diff
@@ -322,6 +313,23 @@ def from_dict(dct: dict[str, Any]) -> CSubst:
...
+        """Extract from a boolean predicate a CSubst."""
+        subst, pred = extract_subst(pred)
+        return CSubst(subst=subst, constraints=flatten_label('#And', pred))
+
+    def pred(self, sort_with: KDefinition | None = None, subst: bool = True, constraints: bool = True) -> KInner:
+        """Return an ML predicate representing this substitution."""
+        _preds: list[KInner] = []
+        if subst:
+            for k, v in self.subst.minimize().items():
+                sort = K if not (sort_with and sort_with.sort(v)) else not_none(sort_with.sort(v))
```
tothtamas28: Here, `sort_with.sort` is called twice if `sort_with` and `sort_with.sort(v)`.

tothtamas28 (approved):
Nice refactoring!

pyk/src/pyk/cterm/cterm.py
ehildenb: Done, unfortunately it's more verbose, maybe there is a better way.

tothtamas28: Looks good to me.

PR #4628, opened 2024-09-06
Feature merge node

Stevengre:
Addresses https://github.com/runtimeverification/k/issues/4425.

This PR introduces the base node merging functionality. 

This is the main feature of this merging node function. Relies on #4617 and #4612 being merged.


pyk/src/pyk/cterm/cterm.py
```diff
@@ -202,6 +202,8 @@ def anti_unify(
             - ``csubst2``: Constrained substitution to apply to `cterm` to obtain `other`.
         """
         new_config, self_subst, other_subst = anti_unify(self.config, other.config, kdef=kdef)
+        # todo: It's not able to distinguish between constraints in different cterms,
+        #  because variable names may be used inconsistently in different cterms.
```
Stevengre, line 207: After this PR, we can further optimize `anti_unify` and provide more tests for it.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -942,12 +942,18 @@ def contains_merged_edge(self, edge: MergedEdge) -> bool:
...
-        merged_edge = KCFG.MergedEdge(source, target, tuple(edges))
+        flatten_edges: list[KCFG.Edge] = []
+        for edge in edges:
+            if isinstance(edge, KCFG.MergedEdge):
+                flatten_edges.extend(edge.edges)
+            else:
+                flatten_edges.append(edge)
+        merged_edge = KCFG.MergedEdge(source, target, tuple(flatten_edges))
         self.add_successor(merged_edge)
         return merged_edge
```
Stevengre, line 960: Add `create` support for `MergedEdge`s

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -942,12 +942,20 @@ def contains_merged_edge(self, edge: MergedEdge) -> bool:
             return edge == other
         return False
 
-    def create_merged_edge(self, source_id: NodeIdLike, target_id: NodeIdLike, edges: Iterable[Edge]) -> MergedEdge:
+    def create_merged_edge(
+        self, source_id: NodeIdLike, target_id: NodeIdLike, edges: Iterable[Edge | MergedEdge]
+    ) -> MergedEdge:
```
Stevengre, line 947: Add `create` support for `MergedEdge`s

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -1,22 +1,36 @@
...
+    from pyk.kast.outer import KDefinition
+
     from .kcfg import KCFG, NodeIdLike
+    from .semantics import KCFGSemantics
 
 
 class KCFGMinimizer:
     kcfg: KCFG
+    heuristics: KCFGSemantics
+    kdef: KDefinition | None
```
Stevengre, line 24: Structure change to support `merge_node` and `anti_unify`

Stevengre, line 24: Structure support for `merge_node` and `anti_unify`

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +202,106 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
             _fold_lift, [(self.kcfg.edges, self.lift_split_edge), (self.kcfg.splits, self.lift_split_split)], False
         )
 
+    def merge_nodes(self) -> bool:
```
Stevengre, line 203: Main implementation for merging node

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -198,3 +312,7 @@ def minimize(self) -> None:
         while repeat:
             repeat = self.lift_edges()
             repeat = self.lift_splits() or repeat
+
+        repeat = True
+        while repeat:
+            repeat = self.merge_nodes()
```
Stevengre, line 304: merging node after other lifting operations.

pyk/src/pyk/kcfg/semantics.py
Stevengre, line 1: #4612

pyk/src/tests/integration/ktool/test_imp.py
Stevengre, line 1: #4612

pyk/src/tests/integration/proof/test_custom_step.py
Stevengre, line 1: #4612

pyk/src/tests/integration/proof/test_goto.py
Stevengre, line 1: #4612

pyk/src/tests/integration/proof/test_imp.py
Stevengre, line 1: #4612

pyk/src/tests/integration/proof/test_refute_node.py
Stevengre, line 1: #4612

pyk/src/tests/integration/proof/test_simple.py
Stevengre, line 1: #4612

pyk/src/tests/unit/kcfg/merge_node_data.py
Stevengre, line 1: Test cases for merging node.

Tested & Reviewed some result of the `anti_unify` function.

Mainly tested the generated structure of merging node, which is the main additional functionality of `merge_node`.

pyk/src/tests/unit/kcfg/merge_node_data.py
```diff
@@ -0,0 +1,470 @@
...
+    (MergedNo(), merge_node_test_kcfg(), check_merge_no),
+    (MergedOne(), merge_node_test_kcfg(), check_merged_one),
+    (MergedPartialOne0(), merge_node_test_kcfg(), check_merged_partial_one0),
+    (MergedPartialOne1(), merge_node_test_kcfg(), check_merged_partial_one1),
+    (MergedPartialOne2(), merge_node_test_kcfg(), check_merged_partial_one2),
+    (MergedTwo0(), merge_node_test_kcfg(), check_merged_two0),
+    (MergedTwo1(), merge_node_test_kcfg(), check_merged_two1),
+    (MergedPartialTwo(), merge_node_test_kcfg(), check_merged_partial_two),
+    (MergedFail(), merge_node_test_kcfg(), check_merged_fail),
+)
```
Stevengre, line 472: Here are the test cases.

pyk/src/tests/unit/kcfg/test_minimize.py
```diff
@@ -407,3 +409,12 @@ def x_lt(n: int) -> KApply:
...
     assert actual == expected
+
+
+@pytest.mark.parametrize('heuristics,kcfg,check', KCFG_MERGE_NODE_TEST_DATA)
+def test_merge_nodes(heuristics: KCFGSemantics, kcfg: KCFG, check: Callable[[KCFGMinimizer], None]) -> None:
+    # When
+    minimizer = KCFGMinimizer(kcfg, heuristics)
+
+    # Then
+    check(minimizer)
```
Stevengre, line 423: The way to use the test cases.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -3,20 +3,32 @@
...
 
+    from pyk.kast.outer import KDefinition
+
     from .kcfg import KCFG, NodeIdLike
+    from .semantics import KCFGSemantics
 
 
 class KCFGMinimizer:
     kcfg: KCFG
+    heuristics: KCFGSemantics
```
ehildenb:
```suggestion
    semantics: KCFGSemantics
```

Call it what it is, for easier use later. Either `kcfg_semantics` or just `semantics`.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -942,12 +942,20 @@ def contains_merged_edge(self, edge: MergedEdge) -> bool:
...
         target = self.node(target_id)
-        merged_edge = KCFG.MergedEdge(source, target, tuple(edges))
+        flatten_edges: list[KCFG.Edge] = []
+        for edge in edges:
+            if isinstance(edge, KCFG.MergedEdge):
+                flatten_edges.extend(edge.edges)
+            else:
+                flatten_edges.append(edge)
+        merged_edge = KCFG.MergedEdge(source, target, tuple(flatten_edges))
         self.add_successor(merged_edge)
```
ehildenb, line 959: Should we be checking that each `edge` in `flatten_edges` has the correct `source` and `target`? We may want to either:

- Make sure by updating their `source` and `target`, so we don't get serialization issues, or
- Assert that they already have the correct `source` and `target`.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,106 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+                while idx < len(ai2bi):
+                    if self.heuristics.is_mergeable(mergeable_edges[0].target.cterm, ai2bi[idx].target.cterm):
+                        for mergable_edge in mergeable_edges[1:]:
+                            if not self.heuristics.is_mergeable(mergable_edge.target.cterm, ai2bi[idx].target.cterm):
+                                raise ValueError(
+                                    'Mergeable edges are not partitioned, you should provide a better heuristic'
+                                )
+                        mergeable_edges.append(ai2bi.pop(idx))
+                    else:
+                        idx += 1
```
ehildenb: I think we can factor out a function in `utils.py` (one directory up) called `partition_with` which does this:

```
def partition_with(relation: Callable[[T, T], bool], items: Iterable[T]) -> list[list[T]]:
    ...
```

That would make this routine a bit shorter and easier to read.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,106 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+                single(self.kcfg.merged_edges(source_id=ai))
+                for ai in split.target_ids
+                if self.kcfg.merged_edges(source_id=ai)
+            ]
+            if len(edges) > 2:
+                a2ai_list.append(split)
+                ai2bi_list.append(edges)
+
+        # Step 3. Apply the heuristic & Obtain the merge-able KCFG sub-graphs
+        to_merge: dict[KCFG.Split, list[list[KCFG.Edge | KCFG.MergedEdge]]] = {}  # Split |-> Merge-able Edges
```
ehildenb: I don't see where this is ever really used as a dictionary, it can probably just be a list of tuples instead: `list[tuple[KCFG.Split, list[list[KCFG.Edge | KCFG.MergedEdge]]]]`

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,106 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+                ai_cterms = [ai2bi.source.cterm for ai2bi in ai2bis]
+                bi_cterms = [ai2bi.target.cterm for ai2bi in ai2bis]
+                merged_ai_cterm = reduce(cterm_anti_unify, ai_cterms)
+                merged_bi_cterm = reduce(cterm_anti_unify, bi_cterms)
+                merged_ai = self.kcfg.create_node(merged_ai_cterm)
+                merged_bi = self.kcfg.create_node(merged_bi_cterm)
+                merged_edge = self.kcfg.create_merged_edge(merged_ai.id, merged_bi.id, ai2bis)
+                merged_edges.append(merged_edge)
+                self.kcfg.create_split_by_nodes(merged_bi.id, [ai2bi.target.id for ai2bi in ai2bis])
+
```
ehildenb: Can we factor out a method in `cterm.py` called `cterm_anti_unify(cterms: Iterable[CTerm], keep_values: bool = False, kdef: KDefinition | None = None) -> tuple[CTerm, list[CSubst]]` which handles the multiple cterm case? That would make writing teh algorithm a little easier.

ehildenb:
Here is what I think is a simpler and easier-to-read algorithm that accomplishes the same (or similar). It relies on defining `cterm_anti_unify` for unifying many cterms at once, and `partition` for partitioning a list based on boolean predicates. It also assumes that we rename `heuristics` to just `semantics`.

```
    def merge_nodes(self) -> bool:
        """Merge targets of Split for cutting down the number of branches, using semantics KCFGSemantics.is_mergeable. 
 
        Side Effect: The KCFG is rewritten by the following rewrite pattern, 
            - Match: A -|Split|-> A_i -|Edge|-> B_i 
            - Rewrite: 
                - if `B_x, B_y, ..., B_z are not mergeable` then unchanged 
                - if `B_x, B_y, ..., B_z are mergeable`, then 
                    - `A -|MergedEdge|-> B_x or B_y or ... or B_z`
 
        Specifically, when `B_merge = B_x or B_y or ... or B_z`
        - `or`: fresh variables in places where the configurations differ 

        :param semantics: provides the is_mergeable heuristic 
        :return: whether any merge was performed 
        """ 

        def is_mergeable_edge(edge1: KCFG.Edge, edge2: KCFG.Edge) -> bool:
            return self.semantics.is_mergeable(edge1.target, edge2.target)
                    
        # Find matching subgraphs
        merge_subgraphs: list[tuple[KCFG.Split, list[list[KCFG.Edge]]]] = []
        for node in self.kcfg.nodes():
            splits = self.kcfg.splits(node.id)
            if len(splits) == 1:
                split = splits[0]
                _edges = [self.kcfg.edges(source_id=target_id) for target_id in split.target_ids]
                if all(len(_edge) == 1 for _edge in _edges):
                    edges = [_edge[0] for _edge in _edges]
                    merge_partitions = partition(is_mergeable_edge, edges)
                    if len(merge_partitions) < len(edges):
                        merge_subgraphs.append((split, merge_partitions))

        merged = False
        affected_nodes = []
        for split, edge_partitions in merge_subgraphs:
            _affected_nodes = (
                [split.source.id]
                + split.target_ids
                + [edge.target.id for edge_partition in edge_partitions for edge in edge_partition]
            )
            # Ensure that we're always working on disjoint subgraphs
            if any(_affected_node in affected_nodes for _affected_node in _affected_nodes):
                continue

            merged = True
            affected_nodes.extend(_affected_nodes)

            # Remove the whole subgraph
            for target_id in split.target_ids:
                self.kcfg.remove_node(target_id)

            for edge_partition in edge_partitions:
                merged_b_cterm, merged_b_csubsts = cterm_anti_unify(
                    [edge.target for edge in edge_partition], keep_values=True, kdef=kdef
                )
                merged_b_node = self.kcfg.create_node(merged_b_cterm)
                self.kcfg.create_split(
                    merged_b_node.id, zip([edge.target.id for edge in edge_partition], merged_b_csubsts, strict=True)
                )
                edges = [KCFG.Edge(split.source, merged_b_node, edge.depth, edge.rules) for edge in edge_partition]
                self.kcfg.create_merged_edge(split.source.id, merged_b_node.id, edges)

        return merged
```

This is about 30 lines shorter than before, and reads a bit more Pythonic. It works by just identifying and removing the entire subgraph (the split nodes `B_i` coming off of `A_i`), then building a new subgraph. Note it computes `merged_b` and just builds a single `MergedEdge` from `A -|MergedEdge|-> B_x or B_y or ... or B_z`, instead of keeping any of the original structure.

pyk/src/pyk/kcfg/minimize.py
Stevengre: Changed it to `semantics`.

pyk/src/pyk/kcfg/kcfg.py
Stevengre: I'm not sure about the correct `source` and `target` of `flatten_edges`. We deleted them before `create_merged_edge`. I guess $edge.source \implies merged.source$ and $edge.target \implies merged.target$. But this validation needs `CTermSymbolic` or `KCFGExplore` to return a boolean result. KCFG is just a syntactic structure. How about introducing a `KCFGChecker` to check the semantics correctness of KCFG?

pyk/src/pyk/kcfg/minimize.py
Stevengre: Implementation:

https://github.com/runtimeverification/k/blob/28c3905319d56f11b417a4b76bc4406585856d02/pyk/src/pyk/utils.py#L322-L346

Test Cases:

https://github.com/runtimeverification/k/blob/28c3905319d56f11b417a4b76bc4406585856d02/pyk/src/tests/unit/test_utils.py#L63-L83

Application:

https://github.com/runtimeverification/k/blob/28c3905319d56f11b417a4b76bc4406585856d02/pyk/src/pyk/kcfg/minimize.py#L242-L248

pyk/src/pyk/kcfg/minimize.py
Stevengre: Sure! Changed it!

pyk/src/pyk/kcfg/minimize.py
Stevengre: https://github.com/runtimeverification/k/blob/4a95d6d8ffba61d54480fc789d88843383db287c/pyk/src/pyk/cterm/cterm.py#L441-L468

Stevengre:
@ehildenb Thank you for your suggestion!  I've refactored the implementation of `merge_node`. However, partial merge and split merge are exist for the given heuristics, I provide two ways to rewrite the graph.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -434,3 +436,33 @@ def cterm_build_rule(
...
+
+def cterms_anti_unify(
+    cterms: Iterable[CTerm], keep_values: bool = False, kdef: KDefinition | None = None
+) -> tuple[CTerm, list[CSubst]]:
+    """Given two `CTerm` instances, find a more general `CTerm` which can instantiate to both.
+
+    Args:
+        cterms: `CTerm`s to consider for finding a more general `CTerm` with this one.
+        keep_values: do not discard information about abstracted variables in returned result.
+        kdef (optional): `KDefinition` to make analysis more precise.
```
ehildenb, line 449:
```suggestion
    """Given many `CTerm` instances, find a more general `CTerm` which can instantiate to all.

    Args:
        cterms: `CTerm`s to consider for finding a more general `CTerm` with this one.
        keep_values: do not discard information about abstracted variables in returned result.
        kdef (optional): `KDefinition` to make analysis more precise.
```

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+        for split in self.kcfg.splits():
+            _edges = [
+                single(self.kcfg.general_edges(source_id=ai))
+                for ai in split.target_ids
+                if self.kcfg.general_edges(source_id=ai)
+            ]
+            _partitions = [p for p in partition(_edges, _is_mergeable) if len(p) > 1]
+            if _partitions:
+                sub_graphs.append((split, _partitions))
```
ehildenb: Why do we need the case `if len(p) > 1`? We should allow partitions that are a single edge. We should be checking that `len(_partitions) < len(_edges)`, because if not then we want to not try doing any node merging here (it will just result in the same graph as before).

pyk/src/pyk/utils.py
```diff
@@ -319,6 +319,45 @@ def repeat_last(iterable: Iterable[T]) -> Iterator[T]:
...
+        return []
+
+    # each pair of elements in the same group satisfies the predicate
+    for group in result:
+        for i in range(len(group)):
+            for j in range(i + 1, len(group)):
+                if not pred(group[i], group[j]) or not pred(group[j], group[i]):
+                    raise ValueError('Cannot partition with the given predicate')
+
+    return result
```
ehildenb:
```suggestion
def partition(iterable: Iterable[T], pred: Callable[[T, T], bool]) -> list[list[T]]:
    """Partition the iterable into sublists based on the given predicate. 
 
    predicate pred(_, _) should satisfy: 
    - pred(x, x) 
    - if pred(x, y) and pred(y, z) then pred(x, z); 
    - if pred(x, y) then pred(y, x); 
    """
    groups: list[list[T]] = []
    for item in iterable:
        found = False
        for group in groups:
            group_matches = []
            for group_item in group:
                group_match = pred(group_item, item)
                if group_match != pred(item, group_item):
                    raise ValueError(f'Partitioning failed, predicate commutativity failed on: {(item, group_item)}')
                group_matches.append(group_match) 
            if found and any(group_matches):
                raise ValueError(f'Partitioning failed, item matched multiple groups: {item}')
            if all(group_matches):
                found = True
                group.append(item)
            elif any(group_matches):
                raise ValueError(f'Partitioning failed, item matched only some elements of group: {(item, group)}')
        if not found:
            groups.append([item])
    return groups
```

This seems a little simpler, has better error messages, and checks commutativity, transitivity, and disjointness.

It passes the unit tests you have, though you have to change the assertion error message failure to instead be `assert str(e).startswith('Partitioning failed')`

pyk/src/pyk/cterm/cterm.py
```diff
@@ -434,3 +436,33 @@ def cterm_build_rule(
...
+    """
+    # TODO: optimize this function, reduce useless auto-generated variables.
+    iterator = iter(cterms)
+    try:
+        merged_cterm = next(iterator)
+    except StopIteration:
+        raise ValueError('Anti-unification failed: no terms provided') from None
+
+    for cterm in iterator:
+        merged_cterm = merged_cterm.anti_unify(cterm, keep_values, kdef)[0]
```
ehildenb, line 465:
```suggestion
        merged_cterm = merged_cterm.anti_unify(cterm, keep_values=keep_values, kdef=kdef)[0]
```

Optional named argument should be named.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -434,3 +436,33 @@ def cterm_build_rule(
...
+
+        - ``cterm``: More general `CTerm` than any of the input `CTerm`s.
+        - ``csubsts``: List of `CSubst` which, when applied to `cterm`, yield the input `CTerm`s.
+    """
+    # TODO: optimize this function, reduce useless auto-generated variables.
+    iterator = iter(cterms)
+    try:
+        merged_cterm = next(iterator)
+    except StopIteration:
+        raise ValueError('Anti-unification failed: no terms provided') from None
```
ehildenb, line 462:
```suggestion
    if not cterms:
        raise ValueError('Anti-unification failed, no cterms provided.')
    merged_cterm = cterms[0]
```

pyk/src/pyk/cterm/cterm.py
```diff
@@ -434,3 +436,33 @@ def cterm_build_rule(
...
+
+    for cterm in iterator:
+        merged_cterm = merged_cterm.anti_unify(cterm, keep_values, kdef)[0]
+
+    csubsts = [not_none(cterm_match(merged_cterm, cterm)) for cterm in cterms]
+    return merged_cterm, csubsts
```
ehildenb:
```suggestion
def cterms_anti_unify(
    cterms: Iterable[CTerm], keep_values: bool = False, kdef: KDefinition | None = None
) -> tuple[CTerm, list[CSubst]]:
    """Given many `CTerm` instances, find a more general `CTerm` which can instantiate to all.

    Args:
        cterms: `CTerm`s to consider for finding a more general `CTerm` with this one.
        keep_values: do not discard information about abstracted variables in returned result.
        kdef (optional): `KDefinition` to make analysis more precise.

    Returns:
        A tuple ``(cterm, csubsts)`` where

        - ``cterm``: More general `CTerm` than any of the input `CTerm`s.
        - ``csubsts``: List of `CSubst` which, when applied to `cterm`, yield the input `CTerm`s.
    """
    if not cterms:
        raise ValueError('Anti-unification failed, no CTerms provided')
    merged_cterm = cterms[0]
    for cterm in cterms[1:]:
        merged_cterm = merged_cterm.anti_unify(cterm, keep_values, kdef)[0]
    csubsts = [not_none(cterm_match(merged_cterm, cterm)) for cterm in cterms]
    return merged_cterm, csubsts
```

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+
+        # ---- Rewrite ----
+
+        merged = False
+        for split, edge_partitions in sub_graphs:
+            merged = True
```
ehildenb:
```suggestion
        if not sub_graphs:
            return False

        # ---- Rewrite ----

        for split, edge_partitions in sub_graphs:
```

Then `return True` at the end of the function.

The reason I had `merged` variable in my original code is because I was also tracking whether the merges were on disjoint subgraphs for safety.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+
+            # Remove the original sub-graphs
+            for p in edge_partitions:
+                for e in p:
+                    self.kcfg.remove_node(e.source.id)
```
ehildenb:
```suggestion
            # Remove the original sub-graphs
            for target_id in split.target_ids:
                self.kcfg.remove_node(target_id)
```

Why the double-iteration? This should suffice I think.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+
+            # Create A -|MergedEdge|-> Merged_Bi -|Split|-> Bi, if one edge partition covers all the splits
+            if len(edge_partitions) == 1 and len(edge_partitions[0]) == len(split.target_ids):
+                merged_bi_cterm, merged_bi_subst = cterms_anti_unify(
+                    [edge.target.cterm for edge in edge_partitions[0]], True, self.kdef
+                )
```
ehildenb:
```suggestion
                merged_bi_cterm, merged_bi_subst = cterms_anti_unify(
                    [edge.target.cterm for edge in edge_partitions[0]], keep_values=True, kdef=self.kdef
                )
```

Make sure to keep names for named optional parameters.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+
+            # Create A -|Split|-> Others & Merged_Ai -|MergedEdge|-> Merged_Bi -|Split|-> Bi
+            _split_nodes = [t for t in split.target_ids if t not in {e.source.id for p in edge_partitions for e in p}]
+            for edge_partition in edge_partitions:
+                merged_ai_cterm, _ = cterms_anti_unify(
+                    [ai2bi.source.cterm for ai2bi in edge_partition], True, self.kdef
+                )
+                merged_bi_cterm, merged_bi_subst = cterms_anti_unify(
+                    [ai2bi.target.cterm for ai2bi in edge_partition], True, self.kdef
+                )
```
ehildenb:
```suggestion
                merged_ai_cterm, _ = cterms_anti_unify(
                    [ai2bi.source.cterm for ai2bi in edge_partition], keep_values=True, kdef=self.kdef
                )
                merged_bi_cterm, merged_bi_subst = cterms_anti_unify(
                    [ai2bi.target.cterm for ai2bi in edge_partition], keep_values=True, kdef=self.kdef
                )
```

Make sure to keep the named parameters' names.

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,87 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+                merged_bi = self.kcfg.create_node(merged_bi_cterm)
+                self.kcfg.create_merged_edge(split.source.id, merged_bi.id, edge_partitions[0])
+                self.kcfg.create_split(
+                    merged_bi.id, zip([e.target.id for e in edge_partitions[0]], merged_bi_subst, strict=True)
+                )
+                continue
+
+            # Create A -|Split|-> Others & Merged_Ai -|MergedEdge|-> Merged_Bi -|Split|-> Bi
+            _split_nodes = [t for t in split.target_ids if t not in {e.source.id for p in edge_partitions for e in p}]
```
ehildenb: This should always be an empty list at this point right? Every edge should be in one of the partitions, even a partition of its own.

pyk/src/pyk/kcfg/minimize.py
ehildenb: I think we should allow partitions of size 1, but we shouldn't allow _all_ partitions to be size 1 (then we should just skip that subgraph). So I think the check we want is what I have above, that `len(_partitions) < len(_edges)`, but not the check that `len(p) > 1` for each partition.

Stevengre: Changed it! But we still need to handle this special case.

pyk/src/pyk/kcfg/minimize.py
Stevengre: This will remove the partitions of length one. Following this deletion, we will need to reconstruct the Node and Edge or MergedEdge, which will introduce additional complexity.

pyk/src/pyk/kcfg/minimize.py
Stevengre: Changed them!

pyk/src/pyk/kcfg/minimize.py
Stevengre: Yes, it will be empty with the updated partition filter. I have modified and tested the corresponding implementation.

pyk/src/tests/unit/test_utils.py
```diff
@@ -58,3 +58,28 @@ def test_poset(relation: Iterable[tuple[int, int]], expected: dict[int, set[int]
...
+def test_partition(iterable: Iterable[int], pred: Callable[[int, int], bool], expected: list[list[int]] | None) -> None:
+    # When
+    try:
+        actual = partition(iterable, pred)
+
+    # Then
+    except ValueError:
+        if not expected:
+            return
+        raise
```
ehildenb:
```suggestion
    except ValueError as e:
        if not expected:
            str(e).startswith('Partitioning failed')
        raise
```

pyk/src/pyk/kcfg/minimize.py
```diff
@@ -188,6 +200,94 @@ def _fold_lift(result: bool, finder_lifter: tuple[Callable, Callable]) -> bool:
...
+
+        for split, edge_partitions in sub_graphs:
+
+            # Remove the original sub-graphs
+            for p in edge_partitions:
+                if len(p) == 1:
+                    continue
+                for e in p:
+                    # TODO: remove the split and edges, then safely remove the nodes.
+                    self.kcfg.remove_node(e.source.id)
```
ehildenb, line 254: I don't think this `TODO` is needed. Removing the node will go through and remove all the attached splits and edges as well: https://github.com/runtimeverification/k/blob/6e23a83e5380f9eae45a73e3956776f2e666da35/pyk/src/pyk/kcfg/kcfg.py#L764

ehildenb: That's why I think it's just needed to remove the target nodes of the items of the split.

Stevengre: Sure, this will remove all attached splits and edges. However, it may not only eliminate the matched splits and edges but could also inadvertently remove unexpected edges, such as those from other splits, covers, unbranches, and edges to the targets of the splits, as well as splits, covers, and unbranches from those targets.

I prefer not to remove all target nodes of the split because this would also eliminate partitions with a single element. These single-element partitions do not merge anything and should remain unchanged. To restore them after deletion, we need to retain the CTerms of those nodes and the information about the edges (Edge or MergedEdge). Therefore, it is simpler and clearer to keep them unchanged.

ehildenb (approved):

PR #4630, opened 2024-09-06
Refactoring to KCFG printer for CSubst

ehildenb:
In the KCFG printer, we have two separate places where `CSubst` are printed out, for `Split` and for `Cover`. There was lots of very-similar code between the two of them, and also some incorrect logic abotu minimization. This PR:

- Factors out helpers that make this printing logic easier.
- Makes sure that we take into account minimization properly.
- Standardizes how the printing for `CSubst` works.
- Updates expected test output.
- Includes some formatting/style reviews from https://github.com/runtimeverification/k/pull/4631 that were skipped because of automerger.


ehildenb:
I've not marked this `automerge` because several downstream test-suites will need to update their output, so I want to be ready to handle that when I merge.

tothtamas28 (approved):

PR #4634, opened 2024-09-09
Handle list random access patterns in frontend

dwightguth:
This PR adds support for the `L [ I <- X ]` syntax on the LHS of rules, which allows you to pattern match on an arbitrary element of a list.

The code is not fully polished; it is not integrated with configuration abstraction or cell collections. It is not likely to be further improved upon presently because the code it was intended to support ended up not proving worth merging. However, it is broadly useful functionality (along with the accompanying LLVM backend PR) and thus ought to be merged.


ehildenb:
Please add tests to `regression-new` of the new functionality.

k-distribution/tests/regression-new/list-set/test.k
```diff
@@ -0,0 +1,11 @@
...
+  imports LIST
+  imports INT
+ 
+  configuration <k> $PGM:K </k> <list> ListItem(0) ListItem(1) ListItem(2) </list>
+
+  syntax KItem ::= l(Int, Int)
+
+  rule <k> l(I, J) => .K ...</k>
+       <list> _ [ I <- J ] </list>
+endmodule
```
ehildenb, line 11: Can we add a test `l(5, 2)` or the like? Something out of range?

ehildenb:
@Scott-Guest will have to comment on the frontend code, but the tests look good, and the feature makes sense.

Scott-Guest (approved):
Code LGTM, but please
- Document this in "Collection patterns" part of the user manual
- Submit an issue describing the remaining work needed for configuration abstraction and cell collections

dwightguth:
https://github.com/runtimeverification/k/issues/4640

PR #4645, opened 2024-09-19
Optimize `cterms_anti_unify`

Stevengre:
Drop useless generated variables from the generalized cterm.


Stevengre:
It's complicated to multiply multiple variables. Please close this PR and reopen it when performance is needed.

PR #4647, opened 2024-09-23
Avoid using `merge_nodes` by default

Stevengre:
Default: Avoid using merge_nodes, as they offer a more abstract rule instead of the original ones.

### DIFF
--- a/pyk/src/pyk/kcfg/exploration.py
+++ b/pyk/src/pyk/kcfg/exploration.py
@@ -109,5 +109,5 @@ class KCFGExploration:
     #
 
     # Minimizing the KCFG
-    def minimize_kcfg(self) -> None:
-        KCFGMinimizer(self.kcfg).minimize()
+    def minimize_kcfg(self, merge: bool = True) -> None:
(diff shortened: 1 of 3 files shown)


ehildenb (approved):

pyk/src/pyk/kcfg/exploration.py
```diff
@@ -109,5 +109,5 @@ def to_dict(self) -> dict[str, Any]:
     #
 
     # Minimizing the KCFG
-    def minimize_kcfg(self) -> None:
-        KCFGMinimizer(self.kcfg).minimize()
+    def minimize_kcfg(self, merge: bool = True) -> None:
+        KCFGMinimizer(self.kcfg).minimize(merge=merge)
```
ehildenb:
```suggestion
    def minimize_kcfg(self, merge: bool = False) -> None:
        KCFGMinimizer(self.kcfg).minimize(merge=merge)
```

The defaults should probably match to set them both to `False`.

ehildenb (approved):

PR #4650, opened 2024-09-25
Adding is-loop functionality

PetarMax:
This PR adds a recogniser function for loops to the KCFG semantics. This allows us to avoid unnecessary KCFG traversal computation in the cases when the node examined is not a loop head. This computation turned out to be quite expensive in real-world proofs with a high branching factor.


PetarMax:
I have tested this change on the above-mentioned engagement proof and can confirm that the slowdown is no longer present.

PetarMax:
This functionality is also tested here implicitly through all of the tests for bounded executions.

ehildenb (approved):

PR #4657, opened 2024-10-08
TOML: use `default` profile values by default when another profile is active

palinatolmach:
Follow up to https://github.com/runtimeverification/kontrol/pull/825.

In Foundry, when parsing TOML config files, values set in a default profile (e.g., `[prove.default]` in `kontrol.toml`) are inherited by other profiles unless they are explicitly overridden in those profiles, which helps avoid redefining shared configuration settings between profiles ([Foundry docs](https://book.getfoundry.sh/config/)). We'd like to support the same behavior in `kontrol.toml`.

Similarly to a change made in https://github.com/runtimeverification/kontrol/pull/825 to the `foundry.toml` parsing, this PR enforces reading the values set in both active and `default` profiles and using the default ones for values that are not set in the active profile. This PR also adds a `test_prove_legacy_profiles` test that ensures that both `default` and active profile's values are used. 

I do realise, however, that this behavior (other profiles inheriting from `default`) is non-standard, so I'd appreciate any feedback on this PR. Maybe it'd be better if I added another parameter in `get_profile` that would enable this behavior only if the function is called by Kontrol?

### DIFF
--- a/pyk/src/pyk/cli/pyk.py
+++ b/pyk/src/pyk/cli/pyk.py
@@ -579,7 +579,9 @@ def parse_toml_args(
         if len(profile_list) == 0 or profile_list[0] not in toml_profile:
             return {k: v for k, v in toml_profile.items() if type(v) is not dict}
         elif len(profile_list) == 1:
-            return {k: v for k, v in toml_profile[profile_list[0]].items() if type(v) is not dict}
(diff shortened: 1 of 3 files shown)


anvacaru (approved):
This looks good to me. It shouldn't introduce any backward compatibility issues.

PR #4662, opened 2024-10-16
Marking simplifications that preserve definedness

PetarMax:
For the booster-with-remainders to be fully operational and therefore the dependency on Kore be reduced, the simplifications that contain partial functions but which we know preserve definedness must be marked as such.


ehildenb:
It would be great if we could have some tests of this.

It _may_ be that we actually can already test this, by enabling some regression-new tests that are currently disabled? If you look at `pyk/regression-new/skipped`, there are a bunch of tests in the `map-symbolic` folder that may work now?

PetarMax:
I wouldn't quite know how to test this. Are there some tests that are specifically using `booster-dev` (meaning just the booster and no Kore)?

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -414,28 +414,28 @@ module MAP-KORE-SYMBOLIC [symbolic,haskell]
 
   // Adding the definedness condition `notBool (K in_keys(M))` in the ensures clause of the following rule would be redundant
   // because K also appears in the rhs, preserving the case when it's #Bottom.
-  rule (K |-> _ M:Map) [ K <- V ] => (K |-> V M) [simplification]
-  rule M:Map [ K <- V ] => (K |-> V M) requires notBool (K in_keys(M)) [simplification]
+  rule (K |-> _ M:Map) [ K <- V ] => (K |-> V M) [simplification, preserves-definedness]
+  rule M:Map [ K <- V ] => (K |-> V M) requires notBool (K in_keys(M)) [simplification, preserves-definedness]
   rule M:Map [ K <- _ ] [ K <- V ] => M [ K <- V ] [simplification]
```
jberthold: I think we can mark this one, too (if `K in_keys(M)` then LHS is undefined, and so is RHS)
```suggestion
  rule M:Map [ K <- _ ] [ K <- V ] => M [ K <- V ] [simplification, preserves-definedness]
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -414,28 +414,28 @@ module MAP-KORE-SYMBOLIC [symbolic,haskell]
...
   // Adding the definedness condition `notBool (K1 in_keys(M))` in the ensures clause of the following rule would be redundant
   // because K1 also appears in the rhs, preserving the case when it's #Bottom.
-  rule (K1 |-> V1 M:Map) [ K2 <- V2 ] => (K1 |-> V1 (M [ K2 <- V2 ])) requires K1 =/=K K2 [simplification]
+  rule (K1 |-> V1 M:Map) [ K2 <- V2 ] => (K1 |-> V1 (M [ K2 <- V2 ])) requires K1 =/=K K2 [simplification, preserves-definedness]
 
   // Symbolic remove
-  rule (K |-> _ M:Map) [ K <- undef ] => M ensures notBool (K in_keys(M)) [simplification]
+  rule (K |-> _ M:Map) [ K <- undef ] => M ensures notBool (K in_keys(M)) [simplification, preserves-definedness]
   rule M:Map [ K <- undef ] => M requires notBool (K in_keys(M)) [simplification]
```
jberthold: ditto, we could mark this one (if `M` is undefined, it makes LHS and RHS equally undefined).
```suggestion
  rule M:Map [ K <- undef ] => M requires notBool (K in_keys(M)) [simplification, preserves-definedness]
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -414,28 +414,28 @@ module MAP-KORE-SYMBOLIC [symbolic,haskell]
...
+  rule (K1 |-> V1 M:Map) [ K2 <- undef ] => (K1 |-> V1 (M [ K2 <- undef ])) requires K1 =/=K K2 [simplification, preserves-definedness]
 
   // Symbolic lookup
-  rule (K  |->  V M:Map) [ K ]  => V ensures notBool (K in_keys(M)) [simplification]
-  rule (K1 |-> _V M:Map) [ K2 ] => M [K2] requires K1 =/=K K2 ensures notBool (K1 in_keys(M)) [simplification]
-  rule (_MAP:Map [ K  <-  V1 ]) [ K ]  => V1 [simplification]
-  rule ( MAP:Map [ K1 <- _V1 ]) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2 [simplification]
+  rule (K  |->  V M:Map) [ K ]  => V ensures notBool (K in_keys(M)) [simplification, preserves-definedness]
+  rule (K1 |-> _V M:Map) [ K2 ] => M [K2] requires K1 =/=K K2 ensures notBool (K1 in_keys(M)) [simplification, preserves-definedness]
+  rule (_MAP:Map [ K  <-  V1 ]) [ K ]  => V1 [simplification, preserves-definedness]
```
jberthold: Here, we could face an undefined `_MAP` on the LHS which yields a defined value `V1`, better remove the attribute.
```suggestion
  rule (_MAP:Map [ K  <-  V1 ]) [ K ]  => V1 [simplification]
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -414,28 +414,28 @@ module MAP-KORE-SYMBOLIC [symbolic,haskell]
...
+  rule ( MAP:Map [ K1 <- _V1 ]) [ K2 ] => MAP [ K2 ] requires K1 =/=K K2 [simplification, preserves-definedness]
 
-  rule (K  |->  V M:Map) [  K ] orDefault _ => V ensures notBool (K in_keys(M)) [simplification]
-  rule (K1 |-> _V M:Map) [ K2 ] orDefault D => M [K2] orDefault D requires K1 =/=K K2 ensures notBool (K1 in_keys(M)) [simplification]
+  rule (K  |->  V M:Map) [  K ] orDefault _ => V ensures notBool (K in_keys(M)) [simplification, preserves-definedness]
```
jberthold: Ditto, if we remove `M` from the RHS we might swallow an undefined case. Of course we keep `M` in the ensured condition but under a function, which is not the same as in the RHS itself.
```suggestion
  rule (K  |->  V M:Map) [  K ] orDefault _ => V ensures notBool (K in_keys(M)) [simplification]
```

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -414,28 +414,28 @@ module MAP-KORE-SYMBOLIC [symbolic,haskell]
...
+  rule (K1 |-> _V M:Map) [ K2 ] orDefault D => M [K2] orDefault D requires K1 =/=K K2 ensures notBool (K1 in_keys(M)) [simplification, preserves-definedness]
```
jberthold: Here I am not sure about the `ensures` clause. `M` might contain a key `K1`, making the LHS undefined, but the RHS is OK (with regard to `K1`).  We could add a `requires notBool (K1 in_keys(M))` instead of the `ensures`... :thinking: however we get bottom either way so probably OK

k-distribution/include/kframework/builtin/domains.md
```diff
@@ -835,13 +835,13 @@ module SET-KORE-SYMBOLIC [symbolic,haskell]
...
   rule .Set           -Set  _             => .Set       [simplification]
   rule SetItem(X)     -Set (S SetItem(X)) => .Set
-                               ensures notBool (X in S) [simplification]
+                               ensures notBool (X in S) [simplification, preserves-definedness]
   rule S              -Set (S SetItem(X)) => .Set
-                               ensures notBool (X in S) [simplification]
+                               ensures notBool (X in S) [simplification, preserves-definedness]
   rule (S SetItem(X)) -Set S              => SetItem(X)
-                               ensures notBool (X in S) [simplification]
+                               ensures notBool (X in S) [simplification, preserves-definedness]
```
jberthold: Similar argument as for the `Map` cases above, the set `S` could be undefined for other reasons than containing `X`...

jberthold (commented):
Sorry for all those comments about the `Map` and `Set` simplifications...
We came here for the `Int` operations so we could merge those and leave the collection simplifications for another PR, or go through them again and stay on the safe side.

PetarMax:
Yes, so, I am not enjoying what `preserves-definedness` means. I don't think I should be thinking about the definedness of the LHS when writing a simplification, it should be assumed and enforced that it is only applicable when the LHS is defined - this is like when writing rules of any operational semantics. The real issue comes when the RHS has partial functions, and for that I accept that `preserves-definedness` is needed. I am going to just do what is needed for the integer simplifications for now.

The operational semantics of the language has to be definedness-preserving, in the sense that *when the LHS of a rule is well-defined*, then so is the RHS, and we need to ensure that we start from a configuration that is well-defined. That should be all, really.

Now, to the PR at hand. What `domains.md` does not say is when `X >>Int N` and `X <<Int N` are defined. Are they defined for negative numbers? Are they defined for negative shifts? These rules seem to suggest that yes, they are, if the other operand is zero, which I find a bit odd.
```
  rule X <<Int 0 => X [simplification]
  rule 0 <<Int _ => 0 [simplification]
  rule X >>Int 0 => X [simplification]
  rule 0 >>Int _ => 0 [simplification]
```

I will reset the PR and just take care of these rules.

geo2a (approved):

ehildenb (approved):

PR #4666, opened 2024-10-17
add hidden visibility flag

dwightguth:
This PR adds a flag to kompile which passes through to the llvm backend. Fairly straightforward.


ehildenb (approved):
Looks fine, do you need these options in pyk too?

dwightguth:
No.

PR #4673, opened 2024-10-22
Adding Timer Hook Implementation for K

Robertorosmaninho:
Follow up of https://github.com/runtimeverification/llvm-backend/pull/1131

This Hook is commonly used to measure VM time in Semantics executions.

The execution of a semantics with `timerStart` and `timerStop` produces a `hook_time.txt` file with the difference of two timestamps in nanoseconds.


ehildenb:
Why does this need to exist as a dedicated hook? Shouldn't shelling out to `date` and writing the results to a file work for this too, which would use existing hooks?

Robertorosmaninho:
Hey Everett, not really. With a dedicated hook, we can select the granularity of the measurement we would like to do. These hooks provide a simple and generic interface to measure the whole execution, a transaction, a rule evaluation, and opcodes execution! We have been using it for a few months now only updating the branch that contains it. However, to automate the measurement of KEVM, for example, to get some key performance numbers to present at DevCon, we would like to have it merged and available to use inside a Dockerfile without the need to build K.

ehildenb (approved):

PR #4681, opened 2024-11-12
Allow including lemmas dynamically into APRProver

ehildenb:
In order to include lemmas dynamically in the Kontrol prover, we need an option for passing the extra module into the `APRProver` so that it can be fed into the booster's `add-module` endpoint. This also requires being able to translate functional rules to Kore in pyk using `krule_to_kore`. This PR:

- Adds the `extra_module` parameter to the `APRProver` initialization, which allows passing a new module with lemmas into the prover at initialization time. It will pass this module on to the `add-module` endpoint and make sure it's used on all future requests.
- Adds the `symbolic` and `smt-lemma` attributes to the list of attributes recognized by pyk.
- Renames `KProve.get_claims_modules` to `KProve.parse_modules`, and renames some of its parameters.
- Adds support to specifying the sort to use in `bool_to_ml_pred` for converting boolean predicates to matching logic ones.
- Makes adjustments to `_krule_to_kore` to make it support translating some functional/simplification rules to Kore which were not supported before (and adds tests):
  - Factors out `_krule_att_to_kore` method for converting attributes of `KRule` to kore.
  - Adds a pass to process the rule attributes a bit in `krule_to_kore` (eg. handle `owise => priority(200)` and `simplification` vs `priority`).
  - Sort the attributes before sending to Kore (for reliable ordering of the Kore output).
  - Refactors the production of Kore `axiom` to differentiate between semantic and functional rules, and to convert functional rules as matching-logic `implies` axioms.


pyk/src/pyk/konvert/_kast_to_kore.py
```diff
@@ -198,45 +218,105 @@ def _inject(definition: KDefinition, term: KInner, sort: KSort) -> KInner:
 
 # 'krule' should have sorts on variables
 def krule_to_kore(definition: KDefinition, krule: KRule) -> Axiom:
+
```
palinatolmach:
```suggestion
```

pyk/src/pyk/kast/manip.py
```diff
@@ -102,15 +102,15 @@ def fun(term: KInner) -> KInner:
     return fun
 
 
-def bool_to_ml_pred(kast: KInner) -> KInner:
+def bool_to_ml_pred(kast: KInner, sort: str | KSort = GENERATED_TOP_CELL) -> KInner:
     def _bool_constraint_to_ml(_kast: KInner) -> KInner:
         if _kast == TRUE:
             return mlTop()
```
tothtamas28:
```suggestion
            return mlTop(sort=sort)
```

pyk/src/pyk/kast/outer.py
```diff
@@ -1117,6 +1117,11 @@ def functions(self) -> tuple[KProduction, ...]:
         """Returns the `KProduction` which are function declarations transitively imported by the main module of this definition."""
         return tuple(func for module in self.modules for func in module.functions)
 
+    @cached_property
+    def function_labels(self) -> tuple[str, ...]:
+        """Returns the label names of all the `KProduction` which are function symbols for all modules in this definition."""
+        return tuple(func.klabel.name for func in self.functions if func.klabel is not None)
```
tothtamas28: I don't think there's a function symbol without a `klabel`, consider asserting instead.

pyk/src/pyk/konvert/_kast_to_kore.py
```diff
@@ -198,45 +218,105 @@ def _inject(definition: KDefinition, term: KInner, sort: KSort) -> KInner:
...
+    kast_rule_sorted = definition.sort_vars(KRewrite(kast_lhs, kast_rhs))
+
+    kast_lhs_body, *kast_lhs_constraints = flatten_label('#And', extract_lhs(kast_rule_sorted))
+    kast_rhs_body, *kast_rhs_constraints = flatten_label('#And', extract_rhs(kast_rule_sorted))
+    kore_lhs_body = kast_to_kore(definition, kast_lhs_body, sort=top_level_k_sort)
+    kore_rhs_body = kast_to_kore(definition, kast_rhs_body, sort=top_level_k_sort)
+
+    axiom_vars: tuple[SortVar, ...] = ()
+    kore_axiom: Pattern
+    if not is_functional:
```
tothtamas28, line 258: Consider extracting two helpers: one for the functional case, one for a normal rewrite.

pyk/src/pyk/konvert/_kast_to_kore.py
```diff
@@ -249,6 +329,25 @@ def kflatmodule_to_kore(definition: KDefinition, kflatmodule: KFlatModule) -> Mo
     return Module(name=kflatmodule.name, sentences=(imports + kore_axioms))
 
 
+def _krule_att_to_kore(att_entry: AttEntry, kast_rule_vars: dict[str, list[KVariable]]) -> App:
```
tothtamas28, line 331: There's similar logic in `_module_to_kore`, maybe it's possible to extract that into it's own module and reuse it here.

pyk/src/pyk/ktool/kprove.py
```diff
@@ -255,19 +255,19 @@ def prove_claim(
                 depth=depth,
             )
 
-    def get_claim_modules(
+    def parse_modules(
```
tothtamas28, line 258: Is the module parser exposed by `kprove` general enough to parse an arbitrary module?

tothtamas28 (approved):

pyk/src/pyk/ktool/kprove.py
ehildenb: Hmmmm, I _think_ it should be? But I'm not 100% on that.

pyk/src/pyk/kast/manip.py
ehildenb: Done.

pyk/src/pyk/kast/outer.py
ehildenb: Done.

pyk/src/pyk/konvert/_kast_to_kore.py
ehildenb: Done.

pyk/src/pyk/kast/manip.py
ehildenb: Done.

pyk/src/pyk/konvert/_kast_to_kore.py
ehildenb: Hmmmm, I looked at implementing this as a `RulePass`, but I think that would create a circular dependency between `_kast_to_kore.py` and `_module_to_kore.py`. Maybe we should have `_outer_to_kore.py` and `_inner_to_kore.py`, since a lot of `_kast_to_kore.py` is converting `KInner -> Pattern`, and a lot of `_module_to_kore.py` is converting `KOuter -> KoreDefinition`. Then we could move `_krule_to_kore` to `_outer_to_kore.py`.

For now, I think I'll leave it as is.

pyk/src/pyk/konvert/_kast_to_kore.py
ehildenb: I think I'll leave it as-is for now. But I can make an issue with the desired refactorings listed there.

ehildenb: https://github.com/runtimeverification/k/issues/4690

pyk/src/pyk/konvert/_kast_to_kore.py
ehildenb: https://github.com/runtimeverification/k/issues/4690

PR #4682, opened 2024-11-12
Fix `Formatter`

tothtamas28:
* Change function `add_backets` to use `bottom_up`. This fixes an issue where the function does not recur on the items of a `KSequence`.
* Add an artificial default bracket production. In case a term would require parentheses but there's no bracket production for its sort, this symbol will be applied. This emulates the following parametric production from module `KSEQ`:
   ```k
   syntax {Sort} Sort ::= "(" Sort ")" [bracket, group(defaultBracket), applyPriority(1)]
   ```


ehildenb (approved):

ehildenb:
Thanks for the test. I didn't look deep into the code, but looks overall good.

PR #4685, opened 2024-11-15
Take result attributes directly from the production when resolving contexts

### LINKED ISSUE #4683: [K-Bug] Broken condition for heating rules with result

Front-End

### Which command

- [X] kompile
- [ ] kast
- [ ] krun
- [ ] kprove
- [ ] kprovex
- [ ] ksearch

### What K Version?

v7.1.164-0-g459fdd7b84

### Operating System

Linux

```k
module A
  imports BOOL

  syntax KResult

  syntax Stuff ::= "a" | "b" | "c"
  syntax KItem ::= thing(Stuff, Stuff)  [seqstrict, result(MyResult)]

  rule thing(c, c) => .K

  rule b => c
  rule a => c

  syntax Bool ::= isMyResult(K)  [function, total, symbol(isMyResult)]
  rule isMyResult(_) => false  [owise]
  rule isMyResult(c) => true


endmodule
```

### Steps to Reproduce

```
echo "thing(a, b)" > a.in
kompile a.k
krun a.in
```

Note that execution stops without heating the second argument of "thing". As far as I can tell, this happens because the second heating rule (the one for "b") requires that the first argument is KResult instead of MyResult. 

### Expected Results

The second argument of "thing" should be heated, then evaluated to "c" and put back, then "thing(c, c)" should be rewritten to ".K". That is, the end result should be:
```
<k>
  .K
</k>

```

virgil-serbanuta:
A declaration like
```syntax KItem ::= thing(Stuff, Stuff)  [seqstrict, result(Evaluated)]```
should consider tems that are `Evaluated` as evaluated. Right now, this happens only for the argument being heated/cooled, while the previous arguments are checked for being `KResult`. As an example, when heating the second argument above, the heating rule looks something like this:
```
rule thing(A, B) => B ~> thingFreezer(A) requires isKResult(A) andBool notBool isEvaluated(B)
```

With this PR, the rule becomes
```
rule thing(A, B) => B ~> thingFreezer(A) requires isEvaluated(A) andBool notBool isEvaluated(B)
```

Fixes #4683

### CODE REVIEW

File: k-frontend/src/main/java/org/kframework/compile/ResolveStrict.java [hunk 1]
```diff
@@ -119,6 +119,14 @@ private void setAliases(String attribute, Set<ContextAlias> aliases, Production
   private static final ContextAlias DEFAULT_ALIAS =
       ContextAlias(KVariable("HERE"), BooleanUtils.TRUE, Att.empty());
 
+  private static final ContextAlias defaultAliasWithAtts(Att att) {
+    if (!att.contains(Att.RESULT())) {
+      return DEFAULT_ALIAS;
```
tothtamas28: If `DEFAULT_ALIAS` is not referred to elsewhere, consider inlining it here.

k-frontend/src/main/java/org/kframework/compile/ResolveStrict.java
```diff
@@ -119,6 +119,14 @@ private void setAliases(String attribute, Set<ContextAlias> aliases, Production
   private static final ContextAlias DEFAULT_ALIAS =
       ContextAlias(KVariable("HERE"), BooleanUtils.TRUE, Att.empty());
 
+  private static final ContextAlias defaultAliasWithAtts(Att att) {
```
tothtamas28: Consider

```suggestion
  private static final ContextAlias defaultAliasFor(Production production) {
```

k-distribution/tests/regression-new/issue-4683/test.k
```diff
@@ -0,0 +1,19 @@
...
+  rule thing(c, c) => .K
+
+  rule b => c
+  rule a => c
+
+  syntax Bool ::= isMyResult(K)  [function, total, symbol(isMyResult)]
+  rule isMyResult(_) => false  [owise]
+  rule isMyResult(c) => true
+
+
```
tothtamas28: Nitpick: consider removing these blank lines.

tothtamas28 (approved):
Before merging, please add a description for the PR, as that will become the commit description once it is merged into `master`.

k-frontend/src/main/java/org/kframework/compile/ResolveStrict.java
virgil-serbanuta: Done

k-distribution/tests/regression-new/issue-4683/test.k
virgil-serbanuta: Done

virgil-serbanuta:
> Before merging, please add a description for the PR, as that will become the commit description once it is merged into `master`.

I hope it looks reasonable now.

PR #4687, opened 2024-11-20
updating the `minimize_kcfg` method to accept an optional heuristics parameter for merging node.

Stevengre:
This pull request includes changes to the `pyk/src/pyk/kcfg/exploration.py` file to enhance the functionality of the `minimize_kcfg` method by incorporating heuristics. The most important changes include importing the `KCFGSemantics` module and updating the `minimize_kcfg` method to accept an optional heuristics parameter.

### DIFF
--- a/pyk/src/pyk/kcfg/exploration.py
+++ b/pyk/src/pyk/kcfg/exploration.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 from pyk.kcfg.kcfg import KCFG, NodeAttr
 from pyk.kcfg.minimize import KCFGMinimizer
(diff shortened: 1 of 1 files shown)


palinatolmach (approved):
Thanks! Can you please make the title and PR comment a bit more descriptive? If there's a test you can add to show that the `heuristics` is actually being used, that'd be great too.

Stevengre:
@palinatolmach What do you think of the current description of this PR?

As for the usage, I've tested the heuristic for `KCFGMinimizer`. Since this function is quite simple which just uses the `KCFGMinimizer`, I think I don't need to provide a same test to validate the same result.

palinatolmach (approved):

PR #4688, opened 2024-11-21
Add support for batch requests to JSON RPC server

RaoulSchaffranek:
Simbolik and kontrol-node exchange many requests, and processing them one request at a time quickly becomes a performance bottleneck. By batching requests together we can avoid the bottleneck. The batching implementation here is part of the JSON RPC spec, see: https://www.jsonrpc.org/specification#batch


pyk/src/pyk/rpc/rpc.py
```diff
@@ -71,37 +72,105 @@ class JsonRpcMethod(Protocol):
...
-    methods: dict[str, JsonRpcMethod]
+@dataclass(frozen=True)
+class JsonRpcRequest:
 
-    def __init__(self, methods: dict[str, JsonRpcMethod], *args: Any, **kwargs: Any) -> None:
-        self.methods = methods
-        super().__init__(*args, **kwargs)
+    method: str
+    params: Any
+    id: Any
```
tothtamas28:
```suggestion
    id: int | str
```

pyk/src/pyk/rpc/rpc.py
```diff
@@ -71,37 +72,105 @@ class JsonRpcMethod(Protocol):
     def __call__(self, **kwargs: Any) -> Any: ...
 
 
-class JsonRpcRequestHandler(BaseHTTPRequestHandler):
-    methods: dict[str, JsonRpcMethod]
+@dataclass(frozen=True)
+class JsonRpcRequest:
```
tothtamas28: Consider just making this a `typing.NamedTuple`:

```suggestion
class JsonRpcRequest(NamedTuple):
```

pyk/src/pyk/rpc/rpc.py
tothtamas28: Also consider making `id` the first attribute.

pyk/src/tests/unit/test_json_rpc.py
tothtamas28, line 1: I think this should rather be an integration test.

pyk/src/pyk/rpc/rpc.py
```diff
@@ -71,37 +72,105 @@ class JsonRpcMethod(Protocol):
...
 
-    def __init__(self, methods: dict[str, JsonRpcMethod], *args: Any, **kwargs: Any) -> None:
-        self.methods = methods
-        super().__init__(*args, **kwargs)
+    method: str
+    params: Any
+    id: Any
+
+    @staticmethod
+    def validate(request_dict: Any, valid_methods: Iterable[str]) -> JsonRpcRequest | JsonRpcError:
```
tothtamas28: Consider moving this to `JsonRpcRequestHandler`.

pyk/src/pyk/rpc/rpc.py
```diff
@@ -113,44 +182,44 @@ def do_POST(self) -> None:  # noqa: N802
...
-            return
+    def _batch_request(self, requests: list[dict[str, Any]]) -> JsonRpcBatchResult:
+        return JsonRpcBatchResult(tuple(self._single_request(request) for request in requests))
+
+    def _single_request(self, request: dict[str, Any]) -> JsonRpcError | JsonRpcSuccess:
+        validation_result = JsonRpcRequest.validate(request, self.methods.keys())
+        if isinstance(validation_result, JsonRpcError):
+            return validation_result
 
+        method_name = request['method']
```
tothtamas28:
```suggestion
        method_name = validation_result.method
```

tothtamas28: But if `JsonRpcRequest` is a `NamedTuple`, you could also just unpack:

```python3
method, args, id = JsonRpcRequest.validate(request, self.methods.keys())
```

pyk/src/pyk/rpc/rpc.py
tothtamas28: Then `valid_methods` can also be removed from the parameters (assuming the method is not static).

RaoulSchaffranek:
Thanks for the review @tothtamas28! Sounds all reasonable to me and I addressed your comments in https://github.com/runtimeverification/k/pull/4688/commits/f8fa54f823fe9202e2a86efd8deb40d77413b872

pyk/src/pyk/rpc/rpc.py
```diff
@@ -71,37 +72,80 @@ class JsonRpcMethod(Protocol):
...
-    def send_json_error(self, code: int, message: str, id: Any = None) -> None:
-        error_dict = {
+class JsonRpcBatchRequest(NamedTuple):
+    requests: tuple[JsonRpcRequest]
+
+
+class JsonRpcResult:
+
+    def encode(self) -> bytes:
+        raise NotImplementedError('Subclasses must implement this method')
```
tothtamas28:
```suggestion
class JsonRpcResult(ABC):

    @abstractmethod
    def encode(self) -> bytes: ...
```

tothtamas28 (approved):

PR #4694, opened 2024-11-25
modified:   .github/workflows/master-pr.yml

F-WRunTime:
Problem: Prs opened by forks do not have access to secrets. The WF to change from default banch > develop uses a secret inaccessible by forks.
- Modifying the workflow permissions to enable explicit write allowance to the API endpoint used for modifying the pull request base branch.
- Removed references to internal secrets.

Reference Material: 
[Overview of modifying permissions 
](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)

[Permission breakdown](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token) of what allows what when modifying job permissions (solution used) 

Checking the API endpoints affected by setting 'pull-requests' > 'write' [lists the api endpoint we are using in the WF](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests) to change the base branch

### DIFF
--- a/.github/workflows/master-pr.yml
+++ b/.github/workflows/master-pr.yml
@@ -11,7 +11,10 @@ jobs:
   change-base:
     name: 'Change base to develop branch'
     runs-on: ubuntu-24.04
-    permissions: "write-all"
+    permissions:
+      pull-requests: write
+      actions: write
+      contents: read
     steps:
       - name: 'Check out code'
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


F-WRunTime:
Target Branch Changed! Thank you for your support!

ehildenb:
I don't think this is actually necessary after the conversation on slack though. Virgil rebased his PR on `develop` manually, which is fine for forking workflows, since those PRs are not coming in very often.

PR #4695, opened 2024-11-26
support write MergedEdge during CSE proof

Stevengre:
As @ehildenb suggested, this PR doesn't delete the storage of the nodes during merging nodes.

- provide a silent removal of nodes: `KCFG.remove_node_silent`;
- use this removal for merging node.


Stevengre:
Idea: Could we always store all the `_node` with `_delete_flag` to show which are deleted.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -758,11 +760,14 @@ def create_node(self, cterm: CTerm) -> KCFG.Node:
...
         node_id = self._resolve(node_id)
-
-        node = self._nodes.pop(node_id)
+        self._nodes.pop(node_id)
+        self._deleted_nodes.add(node_id)
         self._created_nodes.discard(node_id)
-        self._deleted_nodes.add(node.id)
+
+    def remove_node_silent(self, node_id: NodeIdLike) -> None:
+        node_id = self._resolve(node_id)
```
ehildenb: I would call this instead `remove_edges_around(self, node_id: NodeIdLike) -> None:`

As it removes all the edges touching the given node.

ehildenb:
@Stevengre you need to still do `remove_edges_around` before doing th erest of `remove_node`.

ehildenb (approved):

PR #4696, opened 2024-11-26
Add logic to parse rules from `definition.kore`

tothtamas28:
Parse rule axioms (i.e. not `functional`, `constructor`, etc.) of a KORE definition into a more structured form:
* `RewriteRule` for rewrite rules,
* `FunctionRule` for cases of function definitions.
* `SimpliRule` for simplification rules.

The simplification rule for `inj` transitivity from the static prelude of `definition.kore`, as well as ceil simplification rules are skipped for now.

### DIFF
--- a/pyk/src/pyk/kore/rule.py
+++ b/pyk/src/pyk/kore/rule.py
@@ -9,12 +9,39 @@ from abc import ABC
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, final
 
-from .syntax import And, App, Ceil, Equals, EVar, Implies, In, Not, Rewrites, String, Top
+from .prelude import inj
+from .syntax import And, App, Axiom, Ceil, Equals, EVar, Implies, In, Not, Rewrites, SortVar, String, Top
 
 if TYPE_CHECKING:
-    from .syntax import Axiom, Definition, Pattern
+    from typing import Final
+
+    from .syntax import Definition, Pattern
 
     Attrs = dict[str, tuple[Pattern, ...]]
+
+
+# There's a simplification rule with irregular form in the prelude module INJ.
+# This rule is skipped in Rule.extract_all.
+_S1, _S2, _S3, _R = (SortVar(name) for name in ['S1', 'S2', 'S3', 'R'])
+_T: Final = EVar('T', _S1)
+# axiom {S1, S2, S3, R} \equals{S3, R}(inj{S2, S3}(inj{S1, S2}(T:S1)), inj{S1, S3}(T:S1)) [simplification{}()]
+_INJ_AXIOM: Final = Axiom(
+    vars=(_S1, _S2, _S3, _R),
+    pattern=Equals(_S3, _R, inj(_S2, _S3, inj(_S1, _S2, _T)), inj(_S1, _S3, _T)),
+    attrs=(App('simplification'),),
+)
+
+# The following attributes mark axioms that are not rule axioms.
+# Such axioms are skipped in Rule.extract_all.
+_SKIPPED_ATTRS: Final = (
+    'assoc',
+    'constructor',
+    'functional',
+    'idem',
+    'symbol-overload',
+    'subsort',
+    'unit',
+)
 
 
 class Rule(ABC):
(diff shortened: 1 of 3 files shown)


tothtamas28:
Manually tested on both the LLVM and Haskell targets of `evm-semantics`.

jberthold:
> Manually tested on both the LLVM and Haskell targets of `evm-semantics`.

Maybe also test it on `mx-semantics` and other semantics from the Wasm family? 
That was quite useful for booster development where `evm-semantics` did not use all of the K features.

pyk/src/pyk/kore/rule.py
```diff
@@ -0,0 +1,287 @@
...
+        for axiom in defn.axioms:
+            attrs = {attr.symbol for attr in axiom.attrs}
+            if any(attr in attrs for attr in filter_attrs):
+                continue
+
+            match axiom.pattern:
+                case Equals():
+                    # A single simplification from prelude module INJ:
+                    # axiom {S1, S2, S3, R} \equals{S3, R}(inj{S2, S3}(inj{S1, S2}(T:S1)), inj{S1, S3}(T:S1)) [simplification{}()]
+                    continue
```
jberthold: Maybe a good idea to double-check that it is indeed this particular simplification equation? (by checking the symbol on the LHS and maybe verifying that it does not have location information)

When developing booster, we tried to be specific about the cases that we discard and not have a catch-all case so we would get to know if the classification is incomplete.

pyk/src/pyk/kore/rule.py
```diff
@@ -0,0 +1,287 @@
...
+            req=req,
+            ens=ens,
+            priority=priority,
+        )
+
+    @staticmethod
+    def _extract_args(axiom: Axiom) -> tuple[tuple[Pattern, ...], Pattern | None]:
+        req: Pattern | None = None
+        # Cases 7-10 of get_left_hand_side
+        # Cases 0-3 of get_requires
```
jberthold, line 185: What are these comments referring to?

pyk/src/pyk/kore/rule.py
```diff
@@ -0,0 +1,287 @@
...
+    def extract_all(defn: Definition) -> list[Rule]:
+        filter_attrs = [
+            'assoc',
+            'constructor',
+            'functional',
+            'idem',
+            'symbol-overload',
+            'subsort',
+            'unit',
+        ]
```
jberthold: Maybe the code would be easier to read if you added a function to classify rules, separate from "extracting"/internalising them.  Each class has certain invariants (obviously a `Rewrites` head for rewrite rules, and an `Implies`  head for equations - simplification or function equation as per attribute, but within those there is a mandatory shape of LHS and RHS given by the compiler).

jberthold: While typing this up... this "mandatory shape" is actually a _contract_  between frontend and (any) backend that should be specified (in a specification external to the code, not "in `ModuleToKore.java`" :grimacing: ).

pyk/src/pyk/kore/rule.py
tothtamas28: These are referring to the comments from the LLVM Backend code, linked to in the module header: https://github.com/runtimeverification/llvm-backend/blob/d5eab4b0f0e610bc60843ebb482f79c043b92702/lib/ast/pattern_matching.cpp#L101-L104

pyk/src/pyk/kore/rule.py
tothtamas28: Strictly classifying rules is as expensive as extracting them, this is just a faster way for pre-classifying them.

But the mandatory shape is enforced by the respective `from_axiom` functions, and if there's a regression in `definition.kore`, these functions would fail (or at least that's the intention).

tothtamas28: I will try to do something about readability though.

pyk/src/pyk/kore/rule.py
tothtamas28: Good point.

tothtamas28: Done.

tothtamas28:
> Maybe also test it on `mx-semantics` and other semantics from the Wasm family?

Tested on the following definitions as well, no failures:
* `mx-semantics.llvm-mandos`
* `mx-semantics.llvm-kasmer`
* `mx-semantics.haskell-mandos`
* `mx-semantics.haskell-kasmer`
* `mxwasm-semantics.llvm`
* `mxwasm-semantics.haskell`
* `mxwasm-semantics.lemma-tests`

jberthold (approved):
Thanks for the adjustments and testing :+1:

PR #4699, opened 2024-12-02
Refactor rule parsing

tothtamas28:
* Remove unnecessary cases
* Make pattern matching logic more concise
* Process three cases of simplification rules: `AppRule`, `CeilRule`, `EqualsRule`

### DIFF
--- a/pyk/src/pyk/kore/rule.py
+++ b/pyk/src/pyk/kore/rule.py
@@ -211,7 +211,7 @@ class FunctionRule(Rule):
             case And(ops=(In(right=x), y)):
                 return (x,) + FunctionRule._get_patterns(y)
             case _:
-                raise AssertionError()
+                raise ValueError(f'Cannot extract argument list from pattern: {pattern.text}')
 
     @staticmethod
     def _extract_rhs(axiom: Axiom) -> tuple[App, Pattern, Pattern | None]:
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kore/rule.py
```diff
@@ -179,93 +181,133 @@ def from_axiom(axiom: Axiom) -> FunctionRule:
...
-        # get_patterns(\top()) = []
-        # get_patterns(\and(\in(_, X), Y)) = X : get_patterns(Y)
+    def _extract_args(pattern: Pattern) -> tuple[Pattern, ...]:
         match pattern:
             case Top():
                 return ()
-            case And(ops=(In(right=x), y)):
-                return (x,) + FunctionRule._get_patterns(y)
+            case And(ops=(In(left=EVar(), right=arg), rest)):
+                return (arg,) + FunctionRule._extract_args(rest)
```
jberthold, line 206: Of course this relies on the order of arguments being consistent with the order of `\in` clauses... (before and after your change), but that's probably OK...

jberthold (approved):
LGTM

pyk/src/pyk/kore/rule.py
tothtamas28: Good catch. But since everything else also relies on a particular structure of terms, I'll just keep it this way until it causes issues.

PR #4700, opened 2024-12-04
Add an instance for CTermSymbolic to custom_step

anvacaru:
Needed for https://github.com/runtimeverification/kontrol/pull/899 to expose the `simplify` endpoint in the `custom_step` function.


PetarMax:
More generally, this gives greater expressivity to `custom_step`, which can now use the backend if required for any of its computation.

PetarMax (approved):

PR #4701, opened 2024-12-04
Add method `to_axiom` to class `Rule`

tothtamas28:


pyk/src/pyk/kore/rule.py
```diff
@@ -89,22 +94,25 @@ def from_axiom(axiom: Axiom) -> Rule:
                 raise ValueError(f'Cannot parse simplification rule: {axiom.text}')
 
     @staticmethod
-    def extract_all(defn: Definition) -> list[Rule]:
-        def is_rule(axiom: Axiom) -> bool:
-            if axiom == _INJ_AXIOM:
-                return False
+    def is_rule(axiom: Axiom) -> bool:
+        if axiom == _INJ_AXIOM:
+            return False
```
jberthold, line 99: :+1:  nicer as a separate function

pyk/src/pyk/kore/rule.py
```diff
@@ -231,63 +317,67 @@ class AppRule(SimpliRule[App]):
...
             ens=ens,
+            sort=sort,
             priority=priority,
         )
 
 
 @final
 @dataclass(frozen=True)
-class EqualsRule(SimpliRule):
+class EqualsRule(SimpliRule[Equals]):
```
jberthold, line 363: Just noticed the interesting (sub-)types here! :thinking: Nice!

jberthold (approved):

jberthold:
Your round-tripping `from_axiom -> to_axiom` test fails on [some accidental artefact of simplification priorities](https://github.com/runtimeverification/k/actions/runs/12161382349/job/33917011378?pr=4701#step:6:2359):
The simplification attribute takes an optional priority argument, and apparently the compiler inserts an empty string in many cases, but not in the injection rule from `BASIC-K`.
Booster [came to accept both the empty string and the absence of an argument to `simplification`](https://github.com/runtimeverification/haskell-backend/blob/854db7fcab75a891b93ffe2969944d9f9dd97f9b/booster/library/Booster/Definition/Attributes/Reader.hs#L314-L320) as priority 50.

tothtamas28:
> The simplification attribute takes an optional priority argument, and apparently the compiler inserts an empty string in many cases, but not in the injection rule from `BASIC-K`.

And for `--backend llvm`, the `""` is not inserted. So I just adjusted the test to normalize the attribute list.

jberthold:
> > The simplification attribute takes an optional priority argument, and apparently the compiler inserts an empty string in many cases, but not in the injection rule from `BASIC-K`.
> 
> And for `--backend llvm`, the `""` is not inserted. So I just adjusted the test to normalize the attribute list.

Fun fact: (or rather conjecture, but I am pretty certain after trying to find out more about it this morning...)
The "" is _also_ not inserted for `--backend haskell` when there is no simplification that has a non-default priority.

Certainly the answer is buried somewhere in the java code... maybe in `ModuleToKore.java`

PR #4704, opened 2024-12-09
Fixes and improvements in `module_to_kore`

tothtamas28:
* Enable checking of non-rule axioms in the test
* Fix assoc, no-junk and no-confusion axiom generation
* Fix sort generation
* Fix a bug in collection attribute generation
* Add support for the `impure` attribute
* Filter out some internal attributes


pyk/src/pyk/konvert/_module_to_kore.py
```diff
@@ -463,16 +469,14 @@ def functional_axiom(production: KProduction) -> Axiom:
...
+        if prod.klabel.name in BUILTIN_LABELS:
             continue
-        if sentence.klabel.name in BUILTIN_LABELS:
+        if not Atts.FUNCTIONAL in prod.att:
             continue
-        if not Atts.FUNCTIONAL in sentence.att:
-            continue
-        res.append(functional_axiom(sentence))
+        res.append(functional_axiom(prod))
     return res
```
jberthold, line 480: There is a clear pattern of repeated code here: 
* for each production `prod` in the module
   - if not `prod.klabel`: skip
   - if `prod.klabel.name in BUILTIN_LABELS`: skip
   - if `TARGET in prod.att`: call `TARGET_axiom` (some instances have different criteria, sometimes several ones)

Since each productions should be part of exactly one group with a certain criterion, would it make sense to iterate over all productions just once, to classify them into `TARGET` groups and  and dispatch to each `TARGET_axiom` in just one place?
(That place would be https://github.com/runtimeverification/k/pull/4704/files#diff-0e99d64ff2ae75d6aa4d5eb3c1822e330e1f2d1f8b69fb19bbb399e5780e2591R107-R119 )

pyk/src/pyk/konvert/_module_to_kore.py
```diff
@@ -940,13 +985,80 @@ def _transform_rule(self, rule: KRule) -> KRule:
 
 
 @dataclass
-class AddAnywhereAtts(KompilerPass):
-    """Add the anywhere attribute to all symbol productions that are overloads or have a corresponding anywhere rule."""
+class AddImpureAtts(SingleModulePass):
+    """Add the `impure` attribute to all function symbol productions whose definition transitively contains `impure`."""
```
jberthold, line 989: While reading through this the question arose in my head whether this is really a "single-module" property.
Maybe it is only single-module because the compiler is smashing all user code into a single module? Because if imports are admitted one could easily import an `impure` "function" from a different module and call it on the RHS of a rule, or not?

pyk/src/pyk/konvert/_module_to_kore.py
```diff
@@ -1063,6 +1196,7 @@ class DiscardHookAtts(SingleModulePass):
         'SET',
         'STRING',
         'SUBSTITUTION',
+        'TIMER',
         'UNIFICATION',
     )
```
jberthold, line 1202: At ~1300 lines for this file, maybe worth refactoring into separate modules...
What comes to mind is a split into compiler passes and final axiom rendering.

jberthold (approved):

pyk/src/pyk/konvert/_module_to_kore.py
tothtamas28: Yes, this is not single module in that sense. The name is probably misleading, but `SingleModulePass` means that it transforms a definition with a single module into another definition with a single module.

tothtamas28:
I opened an issue for the refactorings: #4707

PR #4708, opened 2024-12-10
Optimize `pyk.kast.inner.collect`

tothtamas28:
The previous implementation called `bottom_up` and as a side effect created a deep copy of the input term.

### DIFF
--- a/pyk/src/pyk/kast/inner.py
+++ b/pyk/src/pyk/kast/inner.py
@@ -877,20 +877,18 @@ def var_occurrences(term: KInner) -> dict[str, list[KVariable]]:
     return _var_occurrences
 
 
-# TODO replace by method that does not reconstruct the AST
 def collect(callback: Callable[[KInner], None], kinner: KInner) -> None:
-    """Collect information about a given term traversing it bottom-up using a function with side effects.
+    """Collect information about a given term traversing it top-down using a function with side effects.
 
     Args:
         callback: Function with the side effect of collecting desired information at each AST node.
         kinner: The term to traverse.
     """
-
-    def f(kinner: KInner) -> KInner:
-        callback(kinner)
-        return kinner
-
-    bottom_up(f, kinner)
+    subterms = [kinner]
+    while subterms:
+        term = subterms.pop()
+        subterms.extend(reversed(term.terms))
+        callback(term)
 
 
 def build_assoc(unit: KInner, label: str | KLabel, terms: Iterable[KInner]) -> KInner:


pyk/src/pyk/kast/inner.py
```diff
@@ -885,12 +885,14 @@ def collect(callback: Callable[[KInner], None], kinner: KInner) -> None:
         callback: Function with the side effect of collecting desired information at each AST node.
         kinner: The term to traverse.
     """
+    subterms = [kinner]
 
-    def f(kinner: KInner) -> KInner:
-        callback(kinner)
-        return kinner
+    for term in subterms:
+        subterms.extend(term.terms)  # noqa: B909
```
tothtamas28: `B909` here is not a false positive:

```
Found a mutation of a mutable loop iterable inside the loop body.
```

But I think this is acceptable here to avoid having to maintain two lists.

tothtamas28: (It would be easier to do a top-down traversal, but since the docstring mentions traversal order I didn't want to change it.)

pyk/src/pyk/kast/inner.py
```diff
@@ -885,12 +885,14 @@ def collect(callback: Callable[[KInner], None], kinner: KInner) -> None:
         callback: Function with the side effect of collecting desired information at each AST node.
         kinner: The term to traverse.
     """
+    subterms = [kinner]
```
ehildenb: I guess the `#TODO` at the top of this function can be removed now?

tothtamas28: Indeed, good catch.

pyk/src/pyk/kast/inner.py
```diff
@@ -877,20 +877,21 @@ def _var_occurence(_term: KInner) -> None:
...
-    def f(kinner: KInner) -> KInner:
-        callback(kinner)
-        return kinner
+    for term in subterms:
+        subterms.extend(term.terms)  # noqa: B909
 
-    bottom_up(f, kinner)
+    while subterms:
+        term = subterms.pop()
+        callback(term)
```
ehildenb: Could this be done in one loop instead of two?

```
for term in subterms:
    subterms.extend(term.terms)
    callback(term)
```

ehildenb (approved):

pyk/src/pyk/kast/inner.py
tothtamas28: It could be in a top-down manner, but I don't think it's possible bottom-up.

tothtamas28: I pushed a commit, please take a look: 53a8cc2

ehildenb: Uses downstream:

- [X] https://github.com/runtimeverification/kontrol/blob/cc9afb725355053938e9bd788860f0bc4e7516e0/src/kontrol/foundry.py#L972. This use looks fine for a top-town traversal.
- [X] no direct uses in KEVM
- [X] no direct uses in KWasm
- [X] no direct uses in mx-semantics
- [X] no direct uses in mir-semantics
- [X] no direct uses in Komet
- [X] no direct uses in KRiscV

So if all the uses in this repo also look good for a top-down, we can have this change.

ehildenb: Good to resolve.

PR #4709, opened 2024-12-11
keep the constraints unchanged after creating a new cterm

Stevengre:


pyk/src/pyk/cterm/cterm.py
```diff
@@ -94,8 +94,12 @@ def _check_config(config: KInner) -> None:
...
 
     @staticmethod
-    def _normalize_constraints(constraints: Iterable[KInner]) -> tuple[KInner, ...]:
-        constraints = sorted(normalize_constraints(constraints), key=CTerm._constraint_sort_key)
+    def _normalize_constraints(constraints: Iterable[KInner], sort_constraints: bool = True) -> tuple[KInner, ...]:
+        if sort_constraints:
+            constraints = normalize_constraints(constraints)
+            # constraints = sorted(normalize_constraints(constraints), key=CTerm._constraint_sort_key)
+        else:
+            constraints = normalize_constraints(constraints)
```
Stevengre, line 102: @PetarMax Here is the fastest way to solve the disordered problem for `CSubst`.

https://github.com/runtimeverification/k/blob/c8d0bee9b64f0e83a4db805575699bae040abbf1/pyk/src/pyk/kcfg/minimize.py#L76-L110

Take the above function as an example. line 101 should create a cterm with ordered constraints that the new constraints locates at the end of all the constraints.

https://github.com/runtimeverification/k/blob/c8d0bee9b64f0e83a4db805575699bae040abbf1/pyk/src/pyk/cterm/cterm.py#L348-L352

But even the line 351 gives a correct order for the `CTerm`, the line 352 will re-sort the constraints after calling `__init__`.


https://github.com/runtimeverification/k/blob/c8d0bee9b64f0e83a4db805575699bae040abbf1/pyk/src/pyk/kcfg/kcfg.py#L755-L760

What makes this situation harder to tackle is that the `=` in line 759 will call `__init__` by default, which makes the `sort_constraints` flag useless.

tothtamas28: I think there's a copy-paste error here: even if `sort_constraints=True`, the constraints won't be sorted.

Stevengre: @tothtamas28  Yes, it's an issue that I don't know how to solve it. Because the following code will call `CTerm.__init__` by default and make the previous order-keeping efforts useless. Do you have any suggestions for it? And why should we sort the constraints, btw?

https://github.com/runtimeverification/k/blob/c8d0bee9b64f0e83a4db805575699bae040abbf1/pyk/src/pyk/kcfg/kcfg.py#L758

tothtamas28: I'm not familiar with the context of this change. But my point is, instead of

```python3
def _normalize_constraints(constraints: Iterable[KInner], sort_constraints: bool = True) -> tuple[KInner, ...]:
    if sort_constraints:
        constraints = normalize_constraints(constraints)
        # constraints = sorted(normalize_constraints(constraints), key=CTerm._constraint_sort_key)
    else:
        constraints = normalize_constraints(constraints)
```

you could have

```python3
def _normalize_constraints(constraints: Iterable[KInner]) -> tuple[KInner, ...]:
    constraints = normalize_constraints(constraints)
```

Stevengre: Thank you! I want to refactor the function this way. But I'm not sure if there is any other applications / functions relying on the `sorted` function. Do you think it's safe to just remove the `sorted`?

tothtamas28: One thing I can think of where the constraints being sorted is useful is for equality checking between `CTerm`-s. I'm not sure if it's a common use case in the prover though.

Just to make sure, please check the history of this piece of code. Then we'll have a better idea of why constraints are sorted in the first place.

Stevengre: I think the sorted function is introduced by Petar in [this PR](https://github.com/runtimeverification/k/pull/4281). What's your suggesions? @PetarMax

PR #4710, opened 2024-12-11
KCFG optimisation-on-creation

PetarMax:
This PR introduces an optional optimisation that performs, as the KCFG is created, the already implemented edge minimisation procedure, in which two consecutive edges are subsumed into one.

In this way, we generate much smaller KCFGs, speeding up very large proofs.

### DIFF
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -580,7 +580,7 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
                 self.let_node(node_id=node.id, cterm=cterm)
                 self.create_edge(in_edge.source.id, node.id, in_edge.depth + depth, list(in_edge.rules) + rule_labels)
                 logs[node.id] = logs[node.id] + next_node_logs
-                log(f'basic block at depth {depth}: node update: {node.id}')
+                log(f'basic block at depth {depth}: update: {node.id}')
                 return True
             return False
 
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -586,20 +586,22 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
                 log(f'abstraction node: {node.id}')
 
             case Step(cterm, depth, next_node_logs, rule_labels, _):
-                if optimize_kcfg and (len(edges := self.edges(target_id=node.id)) == 1):
-                    in_edge = edges[0]
+                next_node = self.create_node(cterm)
+                # Optimization for steps consists of on-the-fly merging of consecutive edges and can
+                # be performed only if the current node has a single predecessor connected by an Edge
+                if (
+                    optimize_kcfg
+                    and (len(predecessors := self.predecessors(target_id=node.id)) == 1)
+                    and isinstance(in_edge := predecessors[0], KCFG.Edge)
+                ):
+                    # The existing edge is removed and the step parameters are updated accordingly
                     self.remove_edge(in_edge.source.id, node.id)
-                    self.let_node(node_id=node.id, cterm=cterm)
-                    self.create_edge(
-                        in_edge.source.id, node.id, in_edge.depth + depth, list(in_edge.rules) + rule_labels
-                    )
-                    logs[node.id] = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
-                    log(f'basic block at depth {depth}: update: {node.id}')
-                else:
-                    next_node = self.create_node(cterm)
-                    logs[next_node.id] = next_node_logs
-                    self.create_edge(node.id, next_node.id, depth, rules=rule_labels)
-                    log(f'basic block at depth {depth}: {node.id} --> {next_node.id}')
+                    depth += in_edge.depth
+                    rule_labels = list(in_edge.rules) + rule_labels
+                    next_node_logs = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
+                self.create_edge(in_edge.source.id, next_node.id, depth, rule_labels)
+                logs[next_node.id] = next_node_logs
+                log(f'basic block at depth {depth}: {node.id} --> {next_node.id}')
 
             case Branch(branches, _):
                 branch_node_ids = self.split_on_constraints(node.id, branches)
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -586,6 +586,7 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
                 log(f'abstraction node: {node.id}')
 
             case Step(cterm, depth, next_node_logs, rule_labels, _):
+                node_id = node.id
                 next_node = self.create_node(cterm)
                 # Optimization for steps consists of on-the-fly merging of consecutive edges and can
                 # be performed only if the current node has a single predecessor connected by an Edge
@@ -596,12 +597,13 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
                 ):
                     # The existing edge is removed and the step parameters are updated accordingly
                     self.remove_edge(in_edge.source.id, node.id)
+                    node_id = in_edge.source.id
                     depth += in_edge.depth
                     rule_labels = list(in_edge.rules) + rule_labels
                     next_node_logs = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
-                self.create_edge(in_edge.source.id, next_node.id, depth, rule_labels)
+                self.create_edge(node_id, next_node.id, depth, rule_labels)
                 logs[next_node.id] = next_node_logs
-                log(f'basic block at depth {depth}: {node.id} --> {next_node.id}')
+                log(f'basic block at depth {depth}: {node_id} --> {next_node.id}')
 
             case Branch(branches, _):
                 branch_node_ids = self.split_on_constraints(node.id, branches)
--- a/pyk/src/pyk/kcfg/kcfg.py
+++ b/pyk/src/pyk/kcfg/kcfg.py
@@ -22,7 +22,7 @@ from ..kast.manip import (
 )
 from ..kast.outer import KFlatModule
 from ..prelude.kbool import andBool
-from ..utils import ensure_dir_path, not_none
+from ..utils import ensure_dir_path, not_none, single
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping, MutableMapping
@@ -557,6 +557,7 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
     def extend(
         self,
         extend_result: KCFGExtendResult,
+        optimize_kcfg: bool,
         node: KCFG.Node,
         logs: dict[int, tuple[LogEntry, ...]],
     ) -> None:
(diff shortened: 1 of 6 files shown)


ehildenb:
@PetarMax this needs to be a new `KCFGExtendResult` type, not something modifying the signature of `commit`, since that function is used for different proof types too. KCFG minimization has tobe its own routine that isn't done in conjunction with other things, to keep synchronization across threads working nicely.

PetarMax:
This works perfectly across threads. Understood re signature of `commit`, will adapt.

PetarMax:
Also, I am not sure that we can do anything beyond this edge subsumption in the context of branch parallelisation. The leaf node identifiers do get preserved, though...

PetarMax:
@ehildenb Is it looking better now?

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -569,6 +570,20 @@ def log(message: str, *, warning: bool = False) -> None:
                 f'Extending current KCFG with the following: {message}{result_info_message}',
             )
 
+        def optimize_step(
+            cterm: CTerm, depth: int, next_node_logs: tuple[LogEntry, ...], rule_labels: list[str]
+        ) -> bool:
+            in_edges = self.edges(target_id=node.id)
+            if len(in_edges) == 1:
+                in_edge = single(in_edges)
+                self.remove_edge(in_edge.source.id, node.id)
```
ehildenb: Does this code have to live here or could it be in `APRProver.step_proof` or `APRProver.commit` instead? Passing the data in the `APRProofResult` is better, it keeps our existing abstraction of only passing `*Step` and `*Result` objects back and forth between master and worker threads.

If it lives in `APRProver`, then it can access `KCFGMinimizer` (defined in `minimizepy`) and re-use the routines there instead of making a new one.

ehildenb: For example, it can reuse this routine: https://github.com/runtimeverification/k/blob/ca93ca476a73abfbff358a6911fcb0e260371d58/pyk/src/pyk/kcfg/minimize.py#L53

PetarMax: Yes, I am aware of this routine. This is not the same routine, but rather an analogous one. The key difference is that `lift_edge`/`lift_edges` require both/all edges to already exist. Here, we are "extending" a single edge by modifying its `target` in place.

We could create both edges first, but this would mean that the node counter for the KCFG will be incremented, which could result in a KCFG that has edges like `3 -> 200 ->> {201, 202}` instead of `3 -> 4 ->> {5, 6}`. By doing the extension directly, we also keep the node counter at a minimum.

I'm not sure how to move this code to the places you suggested, since the modifications to the KCFG related to the `APRProofResult/KCFGProofResult` happen in `kcfg.py`.

PetarMax: I do recognise the issue of code duplication and will think how to improve.

PetarMax: I'm not sure. 

This routine updates one edge (by removing and re-creating it) and updates its `target` node.
The `lift_edge` routine in `minimize.py` removes the central node (effectively removing two edges) and then creates an edge between the outer nodes.

There is little actual duplication, it's just a similar approach.

PetarMax:
I can confirm that this change decreases the running time of our longest, most intensive engagement test by ~20%.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -569,6 +570,20 @@ def log(message: str, *, warning: bool = False) -> None:
                 f'Extending current KCFG with the following: {message}{result_info_message}',
             )
 
+        def optimize_step(
+            cterm: CTerm, depth: int, next_node_logs: tuple[LogEntry, ...], rule_labels: list[str]
+        ) -> bool:
+            in_edges = self.edges(target_id=node.id)
+            if len(in_edges) == 1:
+                in_edge = single(in_edges)
```
tothtamas28:
```suggestion
                in_edge = in_edges[0]
```

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -557,6 +557,7 @@ def path_length(_path: Iterable[KCFG.Successor]) -> int:
     def extend(
         self,
         extend_result: KCFGExtendResult,
+        optimize_kcfg: bool,
         node: KCFG.Node,
         logs: dict[int, tuple[LogEntry, ...]],
```
tothtamas28:
```suggestion
    def extend(
        self,
        extend_result: KCFGExtendResult,
        node: KCFG.Node,
        logs: dict[int, tuple[LogEntry, ...]],
        *,
        optimize_kcfg: bool,
```

tothtamas28:
> I can confirm that this change decreases the running time of our longest, most intensive engagement test by ~20%.

20% is pretty significant, and it indicates that somehow maintaining a proof becomes more and more expensive as the number of nodes (or the depth of the tree) grows. What operation could be the cause of this?

PetarMax:
It could be the KCFG traversal, understanding pending nodes, understanding next steps, and then doing all this in parallel for a number of threads...

pyk/src/pyk/kcfg/kcfg.py
PetarMax: Yes, thank you.

pyk/src/pyk/kcfg/kcfg.py
PetarMax: Done.

ehildenb:
It would be worth it to profile it and see where the expensive routines are. I bet we can get a lot of that 20% shaved off with some targeted optimizations.

~That being said, periodic minimization seems useful. But maybe we should just _always_ minimize if the user asks? Basically, when we get a `Step` result [here](https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kcfg/kcfg.py#L586), and if minimization is turend on, we check if the prior edge was an `Edge`, and if so just extend that prior edge (remove it, and put in a longer edge with the union of the rules list and the addition of the depth counters to the new state). So we don't need to minmiize the whole KCFG every time, we instead just use `Step` to extend `Edge` when they are entered into the KCFG, rather than create a new one.~ I see now that this is basically what we're doing here!

We still need to be able to maintain the old behavior for Simbolik, which relies on being able to see every node that is stored for opcode cuts. Minimization basically doesn't play nicely with cut-rules.

pyk/src/pyk/kcfg/kcfg.py
ehildenb: Sonuds good.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -569,6 +571,20 @@ def log(message: str, *, warning: bool = False) -> None:
...
+            in_edges = self.edges(target_id=node.id)
+            if len(in_edges) == 1:
+                in_edge = in_edges[0]
+                self.remove_edge(in_edge.source.id, node.id)
+                self.let_node(node_id=node.id, cterm=cterm)
+                self.create_edge(in_edge.source.id, node.id, in_edge.depth + depth, list(in_edge.rules) + rule_labels)
+                logs[node.id] = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
+                log(f'basic block at depth {depth}: update: {node.id}')
+                return True
+            return False
```
ehildenb: Only remaining comment is that I would just inline this into where `Step` is handled, rather than have a separate function, since it's only going to be used in that one spot.

PetarMax: Ok, the only reason I separated it is because I thought we might want to do other optimisations for other non-`Step` variants and I wanted a modular way of capturing this. Happy to inline.

PetarMax:
> I see now that this is basically what we're doing here!
> 
> We still need to be able to maintain the old behavior for Simbolik, which relies on being able to see every node that is stored for opcode cuts. Minimization basically doesn't play nicely with cut-rules.

Yes, this is what we're doing :) and yes, this behaviour wouldn't be forced, it would be non-default and activated with a switch. Hmm, could we have a `minimize-for-symbolik` variant of `minimization`, which leaves the cut-rules but compresses others?

pyk/src/pyk/kcfg/kcfg.py
PetarMax: Inlined now.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -584,10 +586,20 @@ def log(message: str, *, warning: bool = False) -> None:
...
+                    self.create_edge(
+                        in_edge.source.id, node.id, in_edge.depth + depth, list(in_edge.rules) + rule_labels
+                    )
+                    logs[node.id] = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
+                    log(f'basic block at depth {depth}: update: {node.id}')
+                else:
+                    next_node = self.create_node(cterm)
+                    logs[next_node.id] = next_node_logs
+                    self.create_edge(node.id, next_node.id, depth, rules=rule_labels)
+                    log(f'basic block at depth {depth}: {node.id} --> {next_node.id}')
```
ehildenb: I don't think we should re-use the original node-id actually, because it may be referred to by another edge (for example, a back-edge/cover that points at it, or part of a `MergedEdge`). So we should still generate a fresh new node that gets put into the KCFG (which would be common logic between the minimize branch and non-minimize branch).

We should probably also have a stronger check on the node, that it _only_ has the one incoming edge. Something like:

```suggestion
                next_node = self.create_node(cterm)
                predecessors = self.predecessors(node.id)
                if optimize_kcfg and len(predecessors) == 1 and isinstance(predecessors[0], KCFG.Edge):
                    in_edge = predecessors[0]
                    self.remove_edge(in_edge.source.id, node.id)
                    self.create_edge(
                        in_edge.source.id, next_node.id, in_edge.depth + depth, list(in_edge.rules) + rule_labels
                    )
                    logs[node.id] = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
                    log(f'basic block at depth {depth}: extend {in_edge.source.id} --> {node.id} --> {next_node.id}')
                else:
                    logs[next_node.id] = next_node_logs
                    self.create_edge(node.id, next_node.id, depth, rules=rule_labels)
                    log(f'basic block at depth {depth}: {node.id} --> {next_node.id}')
```

PetarMax: Yes, that is a good point. 

I think that we can only extend the KCFG from a leaf node, is that correct? Can we extend then even from a node that's been covered or a node that's part of a `MergedEdge`, or a node that has a back edge into itself (How could this happen to begin with, please remind me?).

It also seems to me that if we guard against the cases that you mentioned, I think that we can keep the node identifier safely, can we not?

Am I missing something?

PetarMax: There must be some invariant on the predecessors, right? They can't all exist at the same time, surely...

ehildenb: The only thing we guarantee about edges is that each one has at most one successor, and that no zero-length loops are introduced (for example, a split into a cover). So I think you could have multiple predecessors.

But more than that, we should just keep good hygiene with the identifiers and keep them unique, because they can be referenced by things higher up the stack (like the `terminal` nodes list, or the `bounded` nodes list, for example).

PetarMax: Yes, I agree, that's fair.

pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -584,10 +586,25 @@ def log(message: str, *, warning: bool = False) -> None:
...
+                    and (len(predecessors := self.predecessors(target_id=node.id)) == 1)
+                    and isinstance(in_edge := predecessors[0], KCFG.Edge)
+                ):
+                    # The existing edge is removed and the step parameters are updated accordingly
+                    self.remove_edge(in_edge.source.id, node.id)
+                    node_id = in_edge.source.id
+                    depth += in_edge.depth
+                    rule_labels = list(in_edge.rules) + rule_labels
+                    next_node_logs = logs[node.id] + next_node_logs if node.id in logs else next_node_logs
+                    self.remove_node(node.id)
```
ehildenb, line 604: This may cause problems in the future if the same node is referenced by a merged edge somewhere, for example. Maybe best to just keep the node? But that also could cause problems with how we display the KCFG....

So maybe best to just note that this could be a source of issues if the nodes is already referred to somewhere else. It may also warrant making a `safe_remove_node` option, which checks that there aren't any references to the node before removing it. Can you make an issue about that and link it back to this comment?

ehildenb (approved):
I assume this final version was tested downstream to work with the examples we care about. Looks good.

PR #4712, opened 2024-12-13
Add class `KoreDefn`

tothtamas28:
* Encapsulates semantic information about a `Definition`: sorts, symbols, the sort ordering, rewrite and function rules.
* Method `project_to_rewrites` removes all sorts, symbols and function rules that are not relevant for rewriting.

### DIFF
--- a/pyk/src/pyk/kore/internal.py
+++ b/pyk/src/pyk/kore/internal.py
@@ -83,17 +83,7 @@ class KoreDefn:
 
     def project_to_symbols(self) -> KoreDefn:
         """Project definition to symbols present in the definition."""
-
-        def _symbol_sorts() -> set[str]:
-            res = set()
-            for symbol_decl in self.symbols.values():
-                if isinstance(symbol_decl.sort, SortApp):
-                    res.add(symbol_decl.sort.name)
-                res.update(sort.name for sort in symbol_decl.param_sorts if isinstance(sort, SortApp))
-            return res
-
-        symbol_sorts = _symbol_sorts()
-
+        symbol_sorts = {sort for symbol in self.symbols for sort in self._symbol_sorts(symbol)}
         sorts: FrozenDict[str, SortDecl] = FrozenDict(
             (sort, decl) for sort, decl in self.sorts.items() if sort in symbol_sorts
         )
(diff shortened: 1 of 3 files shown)


tothtamas28:
Based on an ad-hoc experiment, for `imp-semantics`, about 40% of sorts, 60% of symbols, and more than 90% of function symbols present in `definition.kore` are irrelevant for rewriting.

For `evm-semantics`, the numbers are about 40%, 60% and 75%, respectively.

jberthold:
> Based on an ad-hoc experiment, for `imp-semantics`, about 40% of sorts, 60% of symbols, and more than 90% of function symbols present in `definition.kore` are irrelevant for rewriting.
> 
> For `evm-semantics`, the numbers are about 40%, 60% and 75%, respectively.

I agree we are carrying around some (typically auto-generated) payload that we won't usually need (those `project_` functions, for one).
However, symbols being "not relevant for rewriting" (i.e. not occurring on rewrite rules) can still not be simply stripped off from the definition. They can be in the input, and need to be known to the rewrite engine to accept them in input.

BTW does the "rewriting" include function equations? Because obviously function symbols don't appear in rewrite rules on the LHS, they can only be on the RHS.

tothtamas28:
> However, symbols being "not relevant for rewriting" (i.e. not occurring on rewrite rules) can still not be simply stripped off from the definition. They can be in the input, and need to be known to the rewrite engine to accept them in input.

That's fair. In this particular case though, we need this transformation for the Lean 4 backend, where only rewrite rules are relevant (or just some of them, which will be a generalization of this transformation).

**Edit**: https://github.com/runtimeverification/k/pull/4712#issuecomment-2548954067

> BTW does the "rewriting" include function equations?

Yes, for those function symbols that appear in rewrite rules (transitively).

tothtamas28:
> However, symbols being "not relevant for rewriting" (i.e. not occurring on rewrite rules) can still not be simply stripped off from the definition. They can be in the input, and need to be known to the rewrite engine to accept them in input.

I ran into a corner case for this actually: the `GeneratedCounterCell` constructor is stripped as the cell is abstract in all rewrite rules. Moving the PR back to draft to fix this.

jberthold (approved):
LGTM

Issue #4725, opened 2025-01-08
Implement K prelude in Lean 4


tothtamas28:
Most functions in `domains.md` are hooked. We need to provide a manual implementation for those. (The function declarations can and should be generated automatically.)

PR #4734, opened 2025-01-15
Add uninterpreted `Map` implementation to `Prelude.lean`

JuanCoRo:
Tackling #4725, this PR adds an uninterpreted implementation for the `Map` sort.

This PR's goal is to iron out how we want these uninterpreted implementations to be before proceeding with the rest of the hooked functions in `domains.md`.


JuanCoRo:
We can write things like
```lean
open MapHookDef
noncomputable def map := mapImpl SortInt SortInt
noncomputable def exmpl : map.map := map.cons 3 4 map.unit
```

pyk/src/pyk/k2lean4/Prelude.lean
```diff
@@ -1,10 +1,105 @@
...
+axiom in_keysAx    : mapCAx → K → Bool
+axiom valuesAx     : mapCAx → List V
+axiom sizeAx       : mapCAx → Nat
+axiom includesAx   : mapCAx → mapCAx → Bool -- map inclusion
+axiom choiceAx     : mapCAx → K -- arbitrary key from a map
+axiom nodupAx      : forall m, List.Nodup (keysAx K m)
+
+-- Uninterpreted Map implementation
+noncomputable def mapImpl (K V : Type) : MapHookSig K V :=
+  MapHookSig.mk
```
tothtamas28: Consider using the `{ map := mapCAx, ... }` syntax.

pyk/src/pyk/k2lean4/Prelude.lean
```diff
@@ -1,10 +1,105 @@
...
+Multiple bindings for the same key are not allowed.
+Note that both keys and values will always be KItems.
+ -/
+
+-- Signature to be instantiated by map implementations
+structure MapHookSig (K V : Type) where
+  map        : Type -- Carrier, such as List (KItem × KItem)
+  unit       : map
+  cons       : K → V → map → map
+  lookup     : map → K → V
```
tothtamas28: Is this not a non-total function?

```suggestion
  lookup     : map → K → Option V
```

(The `#Ceil` rules in `domains.md` should be reviewed for all these functions to see which ones are non-total.)

tothtamas28:
In general, I'm fine with any implementation that models the functions over these data types faithfully.

In addition, it should be checked how implementation can be combined with the sort module:
* #4735

I.e. what additional changes need to be made to the program to make it compile if

```lean
inductive SortMap : Type where
  | mk (coll : List (SortKItem × SortKItem)) : SortMap
```

is replaced by

```lean
inductive SortMap : Type where
  | mk (coll : (mapImpl SortKItem SortKItem).map) : SortMap
```

(assuming that's the intended use).

pyk/src/pyk/k2lean4/Prelude.lean
JuanCoRo: As per [this description](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md#map-lookup) I thought it should be implemented as a `lookup!`, which I believe has the current type. But if that's not what we want we can have it with the `Option` type

ehildenb: I'm not sure what `lookup!` is, is that a LEAN4 thing? I guess initially I would agree that anything non-total should go to `Option`, but probably `Option` is also not quite what we want. It would be best if we could represent `#Bottom` directly, maybe the discussion with the Pi2 folks who implemented ML in LEAN4 will help clarify things.

pyk/src/pyk/k2lean4/Prelude.lean
JuanCoRo: Done!

pyk/src/pyk/k2lean4/Prelude.lean
JuanCoRo: The notation for `!` after a function indicates that it will throw a panic error if the argument it is called with is not defined, but I guess this is not the same as returning `Bottom`.

JuanCoRo: Perhaps we can have it as a sum type? Something like
```lean
axiom lookupAx : mapCAx → K → V ⊕ Empty
axiom lookup_def_exists : forall m k, in_keysAx K m k → ∃ v, lookupAx K V m k = .inl v
axiom lookup_def_empty : forall m k, ¬ in_keysAx K m k → ∃ e, lookupAx K V m k = .inr e
```
This way, if it's defined, we know it'll have the correct value, and otherwise, we get a contradiction.
Edit: added axiom for the pathological case

tothtamas28: It can just return an `Option` for now.

pyk/src/pyk/k2lean4/Prelude.lean
```diff
@@ -1,10 +1,104 @@
...
+axiom keysAx       : mapCAx → List K
+axiom in_keysAx    : mapCAx → K → Bool
+axiom valuesAx     : mapCAx → List V
+axiom sizeAx       : mapCAx → Nat
+axiom includesAx   : mapCAx → mapCAx → Bool -- map inclusion
+axiom choiceAx     : mapCAx → K -- arbitrary key from a map
+axiom nodupAx      : forall m, List.Nodup (keysAx K m)
+
+-- Uninterpreted Map implementation
+noncomputable def mapImpl (K V : Type) : MapHookSig K V :=
```
tothtamas28: Would this work as well?

```suggestion
axiom mapImpl : MapHookSig
```

JuanCoRo: I don't think so since `axiom mapImpl : MapHookSig K V` doesn't provide concrete elements for `mapImpl`. So we cannot access it's components like `mapImpl.map`:
```lean
axiom mapImpl : MapHookSig K V
#check mapImpl.map
```
returns
```
unknown constant 'MapHookDef.mapImpl.map'
```

JuanCoRo:
Re: [this comment](https://github.com/runtimeverification/k/pull/4734#issuecomment-2595744696)

Do we want `SortMap` as an inductive wrapping over the implementation? I thought we would use instead something along these lines:
```lean
noncomputable abbrev SortMap := mapImpl KItem KItem
```
defining `SortMap` right after `KItem` or as soon as possible

tothtamas28:
> I thought we would use instead something along these lines:
> 
> ```lean
> noncomputable abbrev SortMap := mapImpl KItem KItem
> ```

That would be ideal, but I couldn't make the sort module compile that way, so I went with the most straightforward approach for now.

JuanCoRo:
Ah yes! It seems Lean does not allow (definitions/abbrevs/theorems) and inductives in the same `mutual` block, just found out!

pyk/src/pyk/k2lean4/k2lean4.py
```diff
@@ -154,7 +154,7 @@ def _collection(self, sort: str) -> Structure:
                 val = Term(f'List {item}')
             case CollectionKind.MAP:
                 key, value = sorts
-                val = Term(f'List ({key} × {value})')
+                val = Term(f'(MapHook {key} {value}).map')
```
tothtamas28, line 159: Does the program still compile? How much time does it take?

JuanCoRo: Yes, it does compile. The user needs to add the prelude, add `open MapHookDef` and `set_option maxHeartbeats 10000000`

JuanCoRo: Maybe with fewer heartbeats, it also compiles.  I find it takes around the same time as before, 2-3 minutes.

tothtamas28 (approved):

PR #4742, opened 2025-01-21
Avoid clashes of generated declarations and Lean 4

JuanCoRo:
Some generated names might clash with Lean's built in declarations. For instance, generating the function symbols for `evm-semantics` produces `axiom ite`, which clashes with Lean:
```lean
axiom ite {SortSort : Type} (x0 : SortBool) (x1 : SortSort) (x2 : SortSort) : Option SortSort
```
To avoid the clash error, it is now generated as:
```lean
axiom iteAx {SortSort : Type} (x0 : SortBool) (x1 : SortSort) (x2 : SortSort) : Option SortSort
```
Safeguards are put in place for other types of declarations, such as `inductive`, `structure`, `abbrev` and `instance`.

### DIFF
--- a/pyk/src/pyk/k2lean4/model.py
+++ b/pyk/src/pyk/k2lean4/model.py
@@ -7,6 +7,9 @@ from typing import TYPE_CHECKING, final
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
+    from typing import Final
+
+_LEAN_KEYWORDS: Final = {'ite', 'end', 'where'}  # Words that cannot be a the name of a definition
 
 
 def indent(text: str, n: int) -> str:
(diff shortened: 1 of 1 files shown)


pyk/src/pyk/k2lean4/model.py
tothtamas28, line 1: This should be handled in the generator, and maybe checked in the model.

pyk/src/pyk/k2lean4/model.py
```diff
@@ -7,6 +7,9 @@
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
+    from typing import Final
+
+_LEAN_KEYWORDS: Final = {'ite', 'end', 'where'}  # Words that cannot be a the name of a declaration
```
tothtamas28: `end` and `where` are keywords, `ite` is probably just a predefined function.

tothtamas28 (commented):
Consider opening an issue for this instead. We can clear up minor issues like this once we have a semi-automatic end-to-end example worked out.

JuanCoRo:
I'll open an issue then, and leave finishing the rest of the work here for when we decide to tackle it :ok_hand:

pyk/src/pyk/k2lean4/model.py
JuanCoRo: I changed it to `_LEAN_WORDS`. Let me know if this notation is not desired, or if we should separate into two sets depending on keyword or definition.

pyk/src/pyk/k2lean4/model.py
JuanCoRo: I'll leave this as a future item if/when work on this PR resumes.

PR #4748, opened 2025-01-29
Generate `lake` project

tothtamas28:
* Make `cookiecutter` a runtime dependency
* Add logic for project generation
* Add integration test for project generatiion


pyk/src/pyk/klean/generate.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _load_defn(definition_dir: Path) -> KoreDefn:
+    from ..kore.parser import KoreParser
+
+    kore_text = (definition_dir / 'definition.kore').read_text()
+    definition = KoreParser(kore_text).definition()
+
+    defn = KoreDefn.from_definition(definition)
+    defn = defn.project_to_rewrites()
```
JuanCoRo, line 65: Could we add an option to have here a `filter_rewrites` call before `project_to_rewrites`?

JuanCoRo:
How does the workflow of generating the Lean files look like? A user will run a `k` command on a kompiled definition, and that will output a Lean 4 project containing the generated Lean files. Is this correct? Is there anything I'm missing? Thanks!

pyk/src/pyk/klean/generate.py
tothtamas28: Sure. Is it OK if this happens in the next PR that introduces a CLI for this?

tothtamas28:
> How does the workflow of generating the Lean files look like?

The user kompiles a K definition, then points the Python script to the definition directory to generate the `lake` project from it.

pyk/src/pyk/klean/generate.py
JuanCoRo: Absolutely!

pyk/src/pyk/klean/generate.py
```diff
@@ -0,0 +1,87 @@
...
+
+    return {
+        Path(library_name) / f'{name}.lean': gen_module(generator, imports)
+        for name, (generator, imports) in genmodel.items()
+    }
+
+
+def _write_modules(output_dir: Path, modules: dict[Path, Module]) -> None:
+    for path, module in modules.items():
+        (output_dir / path).write_text(str(module))
```
JuanCoRo, line 87: Would it be wise to include here the replacement of conflicting words such as `ite`? Or should that be left entirely to the user/cli-tool?

tothtamas28: > Or should that be left entirely to the user/cli-tool?

I think that should be handled in `k2lean4`.

pyk/src/pyk/klean/k2lean4.py
```diff
@@ -200,10 +200,20 @@ def _symbol_ctor(self, sort: str, symbol: str) -> Ctor:
...
+
+    @staticmethod
+    def _var_ident(var: str) -> str:
+        assert var.startswith('Var')
+        return K2Lean4._escape_ident(var[3:], kore=True)
+
+    @staticmethod
+    def _escape_ident(ident: str, *, kore: bool = False) -> str:
+        if kore:
+            ident = unmunge(ident)
```
JuanCoRo, line 213: In what scenario would `kore = False`?

tothtamas28: If `ident` is not a KORE identifier, in particular, for rule labels.

JuanCoRo (approved):
:rocket:

PR #4749, opened 2025-01-31
 Add `klean` command

tothtamas28:
Enables generating a Lean 4 project from a K definition from the command line.

### DIFF
--- a/pyk/src/pyk/klean/__main__.py
+++ b/pyk/src/pyk/klean/__main__.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from pyk.cli.utils import dir_path
+from pyk.kore.internal import KoreDefn
 
 if TYPE_CHECKING:
     from argparse import Namespace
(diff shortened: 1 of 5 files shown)


pyk/src/pyk/klean/__main__.py
```diff
@@ -0,0 +1,108 @@
...
+    level = loglevel(ns)
+    return level, rest
+
+
+def klean(args: Iterable[str]) -> None:
+    from .generate import generate
+
+    ns = _parse_args(args)
+    defn = _load_defn(ns.definition_dir)
+    defn = defn.filter_rewrites(ns.rules)
```
JuanCoRo: Since `ns.rules` can be `None`, maybe add something like the following?

```suggestion
    if ns.rules:
        defn = defn.filter_rewrites(ns.rules)
```

tothtamas28: Good catch.

JuanCoRo (approved):

PR #4750, opened 2025-01-31
Adding `--save-temps` and `--statistics` to krun

Robertorosmaninho:
This PR adds 2 `krun` flags to `pyk-krun`. This implementation is needed to expose these flags to semantics that rely on pyk like KEVM. The idea is to provide more debugging resources for developers.


gtrepta (approved):
LGTM.

It's kind of unfortunate that we need to add these arguments to so many layers of boilerplate. There's also no obvious way to add support for these new parameters on the command line for `pyk run`. I wonder if expanding the `RunOptions` class to hold all of these arguments and passing that through instead could help with these two issues. Just thinking out loud here.

Robertorosmaninho:
Hey @gtrepta, thanks for the idea! I'll take a look at rhis before merging and ping you again if solve this, ok?

gtrepta:
> Hey @gtrepta, thanks for the idea! I'll take a look at rhis before merging and ping you again if solve this, ok?

It's probably better to merge what you have now. I think it will be a non-trivial change to make and is better suited as a separate PR. I don't mean to discourage you from looking into it though, if you want to do that then by all means please do!

PR #4753, opened 2025-02-05
Generate function definitions

### LINKED ISSUE #4727: Generate function definitions
Depends on: #4726

Suppose we have a function production

```
syntax Foo ::= foo ( <params> )
```

The general structure of a function rule is the following:

```k
rule foo ( <args_i> ) => <rhs_i> requires <req_i> [priority(<prio_i>)]
```

First, we order rules by priority. This allows the following defintion:

```lean
def foo <params> := foo_pgroup_1 <params> <|> foo_pgroup_2 <params> <|> ...
```

where `foo_pgroup_i` is a function representing all rules in priority group `i`. These helpers are combined using `<|>` in increasing order of priority, i.e., if a group with lower priority fails to yield a result, then the next group is tried, etc.

In order for the function to be well-defined, within each priority group, for a given concrete input, there must be at most one matching rule. There are three cases.

## No `requires` clauses

```lean
def foo_pgroup_i <params> :=
  | <args_1> => some <rhs_1>
  | <args_2> => some <rhs_2>
  ...
  | _ => none
```
Strictly speaking, this mapping is faithful only if the patterns are non-overlapping, as the order in which patterns are defined matters in Lean. In the presence of overlapping patterns, the K function is only well defined if there's confluence of values, thus it might happen that a non-well-defined K function is translated to a well-defined Lean function.

## Only a single pattern on the LHS (modulo alpha renaming)

In this case, it has to be ensured that `requires` clauses are non-overlapping. We can generate a `theorem` stub (or multiple ones, for each pair of clauses) to encode the proof obligation. Then the function definition can be implemented using `if-then-else`:

```lean
def foo_pgroup_i <params> :=
  | <args> =>
    if <req_1> then some <rhs_1>
    else if <req_2> then some <rhs2>
    else if ...
    else none
  | _ => none
```

## The general case

I.e., there are `requires` clasues, and there are various patterns on the left hand sides of rules. Such a function can be well defined (even if the LHS patterns are overlapping), consider

```
rule foo ( X, 0 ) => 0 requires X =/=Int 0
rule foo ( 0, Y ) => 1 requires Y =/=Int 0
```

Group rules by LHS pattern. Then generate a helper function as per the previous case for each group. Combine the cases with `<|>` as for priority groups.

Similarly to the previous case though, a `theorem` stub (or multiple ones, for each pair of helpers) has to be generated that states well-definedness (i.e. at most one of the helpers returns `some`).

tothtamas28:
Closes #4727

Generate a function definition for each (interpreted) function symbol and each function rule.

The definition for a function symbols applies its rule functions in an order respecting priorities:

```lean
def «#lookup(_,_)_IMP_KItem_Id_Map» (x0 : SortId) (x1 : SortMap) : Option SortKItem :=
  [_acfc941, _0e4f65f].findSome? (· x0 x1)
```

A function rule evaluates to `none` if
* the LHS pattern doesn't match;
* a function application is the requires clause or the RHS is undefined;
* the requires clause does not hold;
 
and to `some` otherwise:

```lean
def _acfc941 : SortId → SortMap → Option SortKItem
  | X, M => do
    let _Val0 <- «_in_keys(_)_MAP_Bool_KItem_Map» ((@inj SortId SortKItem) X) M
    let _Val1 <- «Map:lookup» M ((@inj SortId SortKItem) X)
    guard _Val0
    return _Val1

def _0e4f65f : SortId → SortMap → Option SortKItem
  | X, M => some (SortKItem.«Error(_,_)_IMP_KItem_Id_Map» X M)
```

Function definitions depending on an uninterpreted function are marked `noncomputable`.

### DIFF
--- a/pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
+++ b/pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
@@ -210,3 +210,8 @@ class Inj (From To : Type) : Type where
   inj (x : From) : To
 
 def inj {From To : Type} [inst : Inj From To] := inst.inj
+
+def «_+Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 + x1)
+def «_-Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 - x1)
+def «_*Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 * x1)
+def «_<=Int_» (x0 : SortInt) (x1 : SortInt) : Option SortBool := some (x0 <= x1)
--- a/k-frontend/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
+++ b/k-frontend/src/main/java/org/kframework/backend/kore/ModuleToKORE.java
@@ -34,6 +34,7 @@ import org.kframework.builtin.Sorts;
 import org.kframework.compile.AddSortInjections;
 import org.kframework.compile.ComputeTransitiveFunctionDependencies;
 import org.kframework.compile.ExpandMacros;
+import org.kframework.compile.NumberSentences;
 import org.kframework.compile.RefreshRules;
 import org.kframework.compile.RewriteToTop;
 import org.kframework.definition.Claim;
@@ -723,6 +724,7 @@ public class ModuleToKORE {
             BooleanUtils.TRUE,
             BooleanUtils.TRUE,
             Att.empty().add(Att.SIMPLIFICATION()));
+    ceilMapRule = (Rule) NumberSentences.number(ceilMapRule);
     rules.add(ceilMapRule);
   }
 
--- a/pyk/src/pyk/klean/generate.py
+++ b/pyk/src/pyk/klean/generate.py
@@ -26,9 +26,9 @@ def generate(
     k2lean4 = K2Lean4(defn)
     genmodel = {
         'Sorts': (k2lean4.sort_module, ['Prelude']),
-        'Func': (k2lean4.func_module, ['Sorts']),
         'Inj': (k2lean4.inj_module, ['Sorts']),
-        'Rewrite': (k2lean4.rewrite_module, ['Func', 'Inj']),
+        'Func': (k2lean4.func_module, ['Inj']),
+        'Rewrite': (k2lean4.rewrite_module, ['Func']),
     }
 
     modules = _gen_modules(context['library_name'], genmodel)
--- a/pyk/src/pyk/klean/k2lean4.py
+++ b/pyk/src/pyk/klean/k2lean4.py
@@ -350,8 +350,9 @@ class K2Lean4:
             rule_str = sorted_rules[0]
             term = Term(f'{rule_str}{arg_str}')
         else:
-            rules_str = f'[{", ".join(sorted_rules)}]'
-            term = Term(f'{rules_str}.findSome? (·{arg_str})')
+            assert arg_str  # a function with multiple rules is not nullary
+            rules_str = ' <|> '.join(f'({rule_str}{arg_str})' for rule_str in sorted_rules)
+            term = Term(rules_str)
 
         val = SimpleVal(term)
         func = decl.symbol.name
--- a/pyk/src/pyk/klean/k2lean4.py
+++ b/pyk/src/pyk/klean/k2lean4.py
@@ -426,7 +426,7 @@ class K2Lean4:
 
     def _transform_bytes_dv(self, value: str) -> Term:
         bytes_str = ', '.join(f'0x{byte:02X}' for byte in bytes_encode(value))
-        return Term(f'⟨#[{bytes_str}⟩]')
+        return Term(f'⟨#[{bytes_str}]⟩')
 
     def _transform_string_dv(self, value: str) -> Term:
         escapes = {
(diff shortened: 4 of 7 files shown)


tothtamas28:
## Known limitations

### Restricting an argument to a subsort on the LHS

This fails to capture the case where a subsort of the subsort is injected directly.

```lean
def _f4c2469 : SortK → Option SortBool
  | SortK.kseq (SortKItem.inj_SortKResult KResult) SortK.dotk => some true
  | _ => none
```

Solution: unroll the pattern for each subsort (that has no further subsorts):

```lean
def _f4c2469 : SortK → Option SortBool
  | SortK.kseq (SortKItem.inj_SortBool KResult) SortK.dotk
  | SortK.kseq (SortKItem.inj_SortInt KResult) SortK.dotk => some true
  | _ => none
```


### Collection patterns on the LHS

Collection "constructors" are mapped as functions, thus cannot be used in patterns.

```lean
def _2ede380 : SortList → Option SortInt
  | _List_ _Gen0 (ListItem (SortKItem.inj_SortInt I)) => do
    let _Val0 <- «_+Int_» I 1
    return _Val0
  | _ => none
```

Solution: add a de-structuring function for each supported pattern to the correspoinding hook implementation, then:

```lean
def _2ede380 (x0 : SortList) : Option SortInt :=
  match list_pat_init_last x0
  | some (_Gen0 (SortKItem.inj_SortInt I)) => do
    let _Val0 <- «_+Int_» I 1
    return _Val0
  | _ => none
```

pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
```diff
@@ -210,3 +210,8 @@ class Inj (From To : Type) : Type where
   inj (x : From) : To
 
 def inj {From To : Type} [inst : Inj From To] := inst.inj
+
+def «_+Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 + x1)
+def «_-Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 - x1)
+def «_*Int_» (x0 : SortInt) (x1 : SortInt) : Option SortInt := some (x0 * x1)
+def «_<=Int_» (x0 : SortInt) (x1 : SortInt) : Option SortBool := some (x0 <= x1)
```
JuanCoRo, line 217: Shouldn't other functions be added? Such as `«_modInt_»`? We could also add hooked functions such as `_List_`. They would be `noncomputable` rather than the current `axiom`s, and we have the above hooks to link to them.

JuanCoRo: This could very well be a follow up PR tho!

JuanCoRo:
I'm getting the following error when trying to build the `Func.lean` file:
```
fail to show termination for
  _432555e
  sizeWordStackAux
with errors
failed to infer structural recursion:
Cannot use parameters #1 of _432555e and x0 of sizeWordStackAux:
  unexpected occurrence of recursive application
    _432555e
Cannot use parameters #2 of _432555e and x1 of sizeWordStackAux:
  the type Int does not have a `.brecOn` recursor
```
This is for the following mutual block, but the larger one is also showing a similar error:
```lean
mutual
  def _432555e : SortWordStack → SortInt → Option SortInt
    | SortWordStack.«_:__EVM-TYPES_WordStack_Int_WordStack» _Gen0 WS, SIZE => do
      let _Val0 <- «_+Int_» SIZE 1
      let _Val1 <- sizeWordStackAux WS _Val0
      return _Val1
    | _, _ => none

  def sizeWordStackAux (x0 : SortWordStack) (x1 : SortInt) : Option SortInt := [_432555e, _75897fa].findSome? (· x0 x1)
end
```
Somehow Lean is failing to see that the block is decreasing on the `SortWordStack` argument

JuanCoRo:
The `_432555e` termination proof seems a bit tricky, since Lean it's not finding it by structural recursion, and the other path we have is by some `size` function on `SortWordStack`, but that's precisely `_432555e` if I understood right.
We could define another `size` function for `SortWordStack` that is obviously terminating to Lean, and then use that decreasing measure here, but perhaps there's a more straight-forward solution?

pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
tothtamas28: Yes, these few functions are just to demonstrate how to define functions that have no associated function rules:
* Add function definition to `Preulde`.
* Add KORE symbol name to `_PRELUDE_FUNCS` in `k2lean4.py`.

tothtamas28:
Regarding the termination proof issue.

I did a few experiments, and I think it might actually be a bug in Lean. In particular, as defined, Lean is unable to show that in `sizeWordStackAux`, function `_432555e` is invoked with `x0`, `x1`, and instead introduces fresh variables in the goal. This practically makes it impossible to finish the proof.

If however the definition is modified to

```lean
  def sizeWordStackAux (x0 : SortWordStack) (x1 : SortInt) : Option SortInt := (_432555e x0 x1) <|> (75897fa x0 x1)
```

the goal is as expected, and in fact no manual proof is necessary.

I'll look into this a bit more, push a hotfix for the generator, and open an issue on Lean if necessary.

tothtamas28:
> Regarding the termination proof issue.

Not a bug in Lean:
* https://github.com/leanprover/lean4/issues/7020#issuecomment-2647911993

Pushed the fix: 8ec666f

JuanCoRo (approved):

PR #4757, opened 2025-02-19
Process complex LHS patterns in function rules

tothtamas28:
Eliminates patterns that are not supported by Lean from the LHS of function rules.

For example, since `_List_` is a function, instead of matching the last item of a list as

```lean
noncomputable def _2ede380 : SortList → Option SortInt
  | _List_ _Gen0 (ListItem (SortKItem.inj_SortInt I)) => ...
  | _ => none
```

the following definition is generated:

```lean
noncomputable def _2ede380 : SortList → Option SortInt
  | _Pat0 => match (ListHook SortKItem).split _Pat0.coll 0 1 with
    | some ([], _Gen0, [SortKItem.inj_SortInt I]) => ...
    | _ => none
```

where `split` is repsonsible for deconstructing the list into prefix, middle and suffix.

### DIFF
--- a/pyk/poetry.lock
+++ b/pyk/poetry.lock
@@ -915,6 +915,25 @@ files = [
 ]
 
 [[package]]
+name = "networkx"
+version = "3.4.2"
+description = "Python package for creating and manipulating graphs and networks"
+optional = false
+python-versions = ">=3.10"
+files = [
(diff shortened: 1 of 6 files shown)


JuanCoRo (approved):
LGTM
Documentation on the introduced changes and overall `klean` code would be appreciated!

PR #4766, opened 2025-03-13
Add Lean prelude instances and change collection implementation

JuanCoRo:
This PR paves the ground to allow `deriving BEq, DecidableEq` for (some) generated sorts.
In particular:
- Instantiates hooks' carriers with a dummy concrete type, which can be revised later
- Instantiates collection types with the hooks' dummy types
- Adds necessary `instance`s in `Prelude.lean`

### DIFF
--- a/pyk/src/pyk/klean/k2lean4.py
+++ b/pyk/src/pyk/klean/k2lean4.py
@@ -190,13 +190,13 @@ class K2Lean4:
         match coll.kind:
             case CollectionKind.LIST:
                 (item,) = sorts
-                term = Term(f'(ListHook {item}).list')
+                term = Term(f'List {item}')
             case CollectionKind.SET:
                 (item,) = sorts
-                term = Term(f'(SetHook {item}).set')
+                term = Term(f'List {item}')
             case CollectionKind.MAP:
                 key, value = sorts
-                term = Term(f'(MapHook {key} {value}).map')
+                term = Term(f'List ({key} × {value})')
         return Field('coll', term)
 
     def sort_module(self) -> Module:
--- a/pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
+++ b/pyk/src/pyk/klean/template/{{ cookiecutter.package_name }}/{{ cookiecutter.library_name }}/Prelude.lean
@@ -58,46 +58,49 @@ structure MapHookSig (K V : Type) where
 -- We use axioms to have uninterpreted functions
 namespace MapHookDef
   variable (K V : Type)
-  axiom mapCAx       : Type -- Map Carrier
-  axiom unitAx       : mapCAx
-  axiom consAx       : K → V → mapCAx → mapCAx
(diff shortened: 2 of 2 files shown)


pyk/src/pyk/klean/k2lean4.py
```diff
@@ -190,13 +190,13 @@ def _collection_field(self, sort: str) -> Field:
         match coll.kind:
             case CollectionKind.LIST:
                 (item,) = sorts
-                term = Term(f'(ListHook {item}).list')
+                term = Term(f'List {item}')
```
tothtamas28, line 193: Do the generated artifacts not compile without this change? `(ListHook Foo).list` still seems to be a valid term.

We should organize the `Prelude` so that we can change the interpretation without changing the generator.

JuanCoRo: They do not compile in general.
As far as I understand, the positivity checker errors out if we have collections where the carrier type needs the structure being defined to be already defined in order to access the carrier type.
E.g.: `SortAccountCellMap` has `coll : (MapHook SortAcctIDCell SortAccountCell).map`, which presumes `SortAccountCell` being defined, but then
`SortAccountCell.val : SortAccountCellMap`.
Changing `coll : (MapHook SortAcctIDCell SortAccountCell).map` to `coll :
List (SortAcctIDCell × SortAccountCell)` seems to resolve this issue.

Another example, consider the following definition with `ListHook`:
```lean
structure SortList : Type where
    coll : (ListHook SortKItem).list
```
This errors out with:
`(kernel) arg #1 of 'SortList.mk' contains a non valid occurrence of the datatypes being declared`
While
```lean
structure SortList : Type where
    coll : List SortKItem
```
passes just fine. I believe the above error stems out of [Lean's positivity checker](https://github.com/leanprover-community/lean/blob/f91b574c40ad862c5134d48995ad083fe48e4cd4/src/kernel/inductive/inductive.cpp#L375)

tothtamas28: It should still work with an `abbrev`, right? So something like

```lean
abbrev KList (T : Type) := List T
```

and then in the generator we can have

```python3
term = Term(f'KList {item}')
```

JuanCoRo: It still complains :pensive: 
```lean
structure SortList : Type where
    coll : KList SortKItem
```
with the above `abbrev KList (T : Type) := List T` outputs `(kernel) arg #1 of 'SortList.mk' contains a non valid occurrence of the datatypes being declared`

tothtamas28: In that case, since we're fixing the interpretation in the generator, would it make sense to remove some of `ListHookSig`, `ListHookDef` and `ListHook` from the prelude?

JuanCoRo: My thinking was to maybe define a coercion or similar from the carrier type to the hook structure.
In this way, we could (if this is possible) still use the hook's interface on the carrier type. So I would keep them there for that purpose if that's alright.

tothtamas28 (approved):

PR #4768, opened 2025-03-13
Add support for generating `deriving` clause

tothtamas28:
Adds command line flags `--derive-beq` and `--derive-decidableeq`, which enable generating `deriving` clauses for the respective instances.


JuanCoRo (approved):
:rocket:

PR #4769, opened 2025-03-13
Use `bottom_up` in fuzzer template substitution

bbyalcinkaya:
Replaces `top_down` with `bottom_up` to avoid redundant traversals in replacement terms, improving performance for large terms. While replacement terms are usually not large, this change enables an optimization in komet that results in a 3x speedup.

The optimization in komet works by embedding large configuration cells without variables directly into a replacement term. This prevents the substitution algorithm from needlessly traversing these large terms, significantly reducing overhead.
* https://github.com/runtimeverification/komet/pull/69


gtrepta (approved):
LGTM, good catch. I recall picking top_down over bottom_up for performance reasons, I don't remember how I decided that and admittedly didn't test it.

PR #4771, opened 2025-03-14
Refactor adding modules via `CTermSymbolic`

ehildenb:
This PR factors out the ability to add KAST level modules to the RPC server via `CTermSymbolic`, so that the user doesn't need to manually do the conversions needed for adding such modules. The `APRProver` is refactored to use this new `CTermSymbolic.add_module` as well.

This is part of https://github.com/runtimeverification/kontrol/issues/977, and blocking https://github.com/runtimeverification/kontrol/pull/979.


palinatolmach (approved):
lgtm!

PR #4772, opened 2025-03-15
Allow introducing semantic rules in kprove

ehildenb:
Part of: https://github.com/runtimeverification/kontrol/issues/977

Currently, `kprove` does not allow including new semantic rules in the definition that is being proven, only claims and `simplification` rules. This change allows the user to set a flag `--allow-new-rules` which disables this check. The intention is for it to be used with `--dry-run`, so that `kprove ...` can be used just to parse some modules basically.

In pyk, we add `--allow-new-rules` to `KProve.parse_module(...)` invocation, so that it can parse new modules that introduce semantic rules. This enables us to use those new rules in the haskell backend booster when doing proofs with Kontrol, which is key to loading loop invariants dynamically to support https://github.com/runtimeverification/kontrol/issues/977.


ehildenb:
I was waiting for more features to emerge that I need, but I think this will do for now!

k-frontend/src/main/java/org/kframework/kprove/KProveOptions.java
```diff
@@ -113,4 +113,10 @@ public synchronized File specFile(FileUtil files) {
       description = "Allow functional claims to be printed in kore format. Use with --dry-run.",
       hidden = true)
   public boolean allowFuncClaims = false;
+
+  @Parameter(
+      names = "--allow-new-rules",
```
tothtamas28: I'd just call it `--allow-rules` to make it shorter.

pyk/src/pyk/ktool/kprove.py
```diff
@@ -274,7 +274,7 @@ def parse_modules(
                 temp_dir=self.use_directory,
                 dry_run=True,
                 type_inference_mode=type_inference_mode,
-                args=['--emit-json-spec', ntf.name],
+                args=['--emit-json-spec', ntf.name, '--allow-new-rules'],
```
tothtamas28: Consider adding an integration test.

tothtamas28 (approved):

k-frontend/src/main/java/org/kframework/kprove/KProveOptions.java
ehildenb: Done.

pyk/src/pyk/ktool/kprove.py
ehildenb: I added a rule to `looping-spec.k`, so that `TestEmitJsonSpec` exercises adding new rules via `parse_modules`.

PR #4774, opened 2025-03-18
Pretty-print constructor rules

JuanCoRo:
This PR introduces some logic to have a prettier generation of the constructors. Especially the constructors of the Rewrite relations.
In particular:
- Adds two extra fields to the `Term` class to allow encoding the `lhs` and `rhs` of rewrite rules
- Adds the following logic to the `__str__` method of the `Signature` class:
  - If there are more than 6 binders, each binder will be in a new line
  - If the constructor of a rewrite relation is being printed, the LHS and RHS are not printed all in one line

The pretty-printing of rewrite states is not as optimal as it could be, but it compiles and allows for easier manipulation of the generated code.

### DIFF
--- a/pyk/src/pyk/klean/k2lean4.py
+++ b/pyk/src/pyk/klean/k2lean4.py
@@ -942,7 +942,7 @@ class K2Lean4:
 
         lhs_term = self._transform_pattern(lhs)
         rhs_term = self._transform_pattern(rhs)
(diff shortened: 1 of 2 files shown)


tothtamas28:
@JuanCoRo, can we close this?

JuanCoRo:
Closed in favor of https://github.com/runtimeverification/evm-equivalence/pull/6

PR #4775, opened 2025-03-19
PyK: Misc enhancements to KCFG and anti-unification

nishantjr:
*  KCFG: Shorten log message for easier readability

    This brings the more important parts of the messages into focus.

*  CTerm.anti_unify: Don't call ml_pred_to_bool unless necessary

    Since these values are only used if keep_values is set, we do not need to always
    compute them. In addition, it avoids exceptions when the predicate cannot be
    converted to a bool expression: e.g. when using #Ceil as a side condition

    This is only a work-around (it would
    still break for a #Ceil side condition, if keep_values=True), but lets us move forward without risking
    breaking other projects that rely on the original behaviour.
    
*   pyk:KCFGShow.to_module: Thread imports and defunc_with parameters through to KCFG.to_module

     This lets us use those options for printing the modules.

### DIFF
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -187,7 +187,7 @@ class CTerm:
         return CTerm(self.config, [new_constraint] + list(self.constraints))
 
     def anti_unify(
-        self: CTerm, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
+        self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
     ) -> tuple[CTerm, CSubst, CSubst]:
         """Given two `CTerm` instances, find a more general `CTerm` which can instantiate to both.
 
--- a/pyk/src/pyk/cterm/cterm.py
+++ b/pyk/src/pyk/cterm/cterm.py
@@ -187,7 +187,7 @@ class CTerm:
         return CTerm(self.config, [new_constraint] + list(self.constraints))
 
     def anti_unify(
-        self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
+        self: CTerm, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
     ) -> tuple[CTerm, CSubst, CSubst]:
         """Given two `CTerm` instances, find a more general `CTerm` which can instantiate to both.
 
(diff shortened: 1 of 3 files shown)


pyk/src/pyk/kcfg/kcfg.py
```diff
@@ -568,7 +568,7 @@ def log(message: str, *, warning: bool = False) -> None:
             result_info_message = f': {result_info}' if result_info else ''
             _LOGGER.log(
                 logging.WARNING if warning else logging.INFO,
-                f'Extending current KCFG with the following: {message}{result_info_message}',
+                f'Extending with {message}{result_info_message}',
```
ehildenb:
```suggestion
                f'Extending KCFG with: {message}{result_info_message}',
```

ehildenb: I think we've been trying to keep the colon to delineate the generic message from the specific data. Not sure if we've kept strictly to that convention, but I think it's a good one.

nishantjr: Makes sense

ehildenb:
@nishantjr it looks ilke you updated the Haskell and LLVM submodules both in your PR, is that intentional? In general, you shouldn't do that, because it missed updating all of the things like nix flake, pyproject.toml, and dependency list files. Can you back those changes out?

nishantjr:
> @nishantjr it looks ilke you updated the Haskell and LLVM submodules both in your PR, is that intentional? In general, you shouldn't do that, because it missed updating all of the things like nix flake, pyproject.toml, and dependency list files. Can you back those changes out?

Ugh, accidentally committed after a rebase. Fixed.

nishantjr:
@ehildenb I've pushed another change that exposes `KCFG.to_module`'s `imports` parameter in `KCFGShow`.

ehildenb:
@nishantjr , looking at the CI failures, I think this test is the most telling one: `TestKonvertCellMap.test_kast_to_kore[account-cell-map]`. I guess this comes from the change in how you're handling cell-maps, maybe that change isn't needed? If it is needed, please add a test demonstrating why it's needed, so we can make sure that one is passing _and_ current ones are passing.

pyk/src/pyk/kast/outer.py
```diff
@@ -1519,8 +1519,8 @@ def _wrap_elements(_k: KInner) -> KInner:
             return _k
 
         # To ensure we don't get duplicate wrappers.
-        _kast = self.remove_cell_map_items(kast)
-        return bottom_up(_wrap_elements, _kast)
+        _kast = bottom_up(_wrap_elements, kast)
+        return self.remove_cell_map_items(_kast)
```
ehildenb: This change in particular, is _my guess_ about what is causing the CI failures. If you need this change, please add a test (probably to `test_kast_to_kore.py`) that demonstrates why it's needed. Currently the failing test is in that category: `TestKonvertCellMap.test_kast_to_kore[account-cell-map]`.

nishantjr: Thanks Everett. Making this a draft again, will get back next week.

ehildenb:
I re-triggered teh CI run, it was due to the haskell server becoming unavailable for some reason. Maybe changing the way that modules are built is causing some issues? But I figured it's worth re-triggering the run.

Can you update the PR description so we get a good commit message?

nishantjr:
Thanks Everett. How does it look now? I've backed out one of the commits; will add in the test for it next week.

pyk/src/pyk/cterm/cterm.py
```diff
@@ -187,7 +187,7 @@ def add_constraint(self, new_constraint: KInner) -> CTerm:
         return CTerm(self.config, [new_constraint] + list(self.constraints))
 
     def anti_unify(
-        self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
+        self: CTerm, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
```
nishantjr:
```suggestion
        self, other: CTerm, keep_values: bool = False, kdef: KDefinition | None = None
```

To commit after CI finishes.

ehildenb (approved):

PR #4776, opened 2025-03-19
modified:   .github/workflows/master-pr.yml

F-WRunTime:
- An error occurs with collaborators without access to secrets and cannot access JENKINS_PAT.
- Try to avoid this secret by defining permission scope to the workflow and us the generated GITHUB_TOKEN

### DIFF
--- a/.github/workflows/master-pr.yml
+++ b/.github/workflows/master-pr.yml
@@ -6,6 +6,9 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
+permissions:
+  contents: read
+  pull-requests: write
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


ehildenb (approved):

F-WRunTime:
Documenting Testing: 
Seen here for a new contributor to run a WF in a repository that is public but with WFs facing the public. An approval is still required 
image

After approval... target is the 'default branch ' 
**NOTE: Approval is required for every change the user makes while they are still a "First-time contributor" tedious but this is good**

Still having issues though with proper permissions generated by non-org users to run the WF and modify a PR. 

Problem is: A PR opened targeting our org/repo is under our domain so permissions generated for a GITHUB_TOKEN occur under our permissions controls, which a contributor (note this is not a collaborator) has none. 

This is a tricky problem. 1 work around is to make the job workflow dispatch so an internal developer beyond providng their review must also run the workflow and target this PR in order to meet all requirement checks.

PR #4780, opened 2025-04-01
README: correct cachix target for K

ehildenb:
As pointed out by @nishantjr , we have an outdated Cachix instructions in the README, which this corrects. See: https://runtimeverification.slack.com/archives/C06L0RWQ2V9/p1742794416788669


nishantjr:
I can't make PR comments on text that hasn't changed, but I'd either remove or update the following section that mentions public keys  and substituter urls, as it may confuse people. That is lines 240-257.

ehildenb:
@nishantjr I just removed that information, is that what you had in mind? Perhaps I should also ping @juliankuners to confirm that it looks good to him.

nishantjr:
Yeah. Thanks! Those keys looked like they were from a previous binary cache.

tothtamas28 (approved):

PR #4783, opened 2025-04-02
Update README in Intro to Section 1

ibastys:
- Rework the intro to Section 1.


k-distribution/k-tutorial/1_basic/02_basics/README.md
```diff
@@ -30,23 +28,43 @@ module LESSON-02-A
...
+
+You should get the following output:
+
+```
+[Warning] Compiler: Could not find main syntax module with name
+LESSON-02-A-SYNTAX in definition.  Use --syntax-module to specify one. Using
+LESSON-02-A as default.
+```
+
+It's just a warning, and you can ignore it for now. Compilation succeeded.
```
Robertorosmaninho, line 58: Here, you can say that the compiler is reminding the user about the desirable convention or splitting definitions from implementation. That we'll teach them the proper way to do it in lesson 5, once the programs start to become bigger, but as these first examples are small, it ought to be fine to ignore them for now.

k-distribution/k-tutorial/1_basic/02_basics/README.md
```diff
@@ -30,23 +28,43 @@ module LESSON-02-A
...
 
-Now, save the following input file in your editor as `banana.color` in the same
-directory as `lesson-02-a.k`:
+`kompile` will also output a directory containing everything needed to execute
+programs and perform proofs using the definition given as input.
+Check your working directory and you'll see directory `lesson-02-a-kompiled`
+has beed created inside. If you're curious, have a look at its contents and 
+inspect the generated documentation.
```
Robertorosmaninho, line 64: You can say that these files not only contain all the basic tools produced by a K definition, but also contain all sorts of files that can be used as input to execute these tools. Or to tell the user how its program was parsed by the compiler frontend.

If you want to highlight one of them, I would elaborate on the `definition.kore`, which is the input for the K compiler's backend. KORE is the Intermediate Representation of K and has its own definition ([link](https://github.com/runtimeverification/haskell-backend/blob/master/docs/kore-syntax.md)), and it was built to be a high-level implementation of matching logic (the Wikipedia page link should be fine as we have recently updated it).

k-distribution/k-tutorial/1_basic/02_basics/README.md
```diff
@@ -58,41 +76,81 @@ We can now evaluate this K term by running (from the same directory):
...
+You can also execute small programs directly on the command line instead of 
+putting them in a file. For example, the same program above
 could also have been executed by running the following command:
 
 ```
 krun -cPGM='colorOf(Banana())'
 ```
 
-Now, let's look at what this definition and program did.
+You could also run it as:
```
Robertorosmaninho, line 101:
```suggestion
Another way to execute your program is:
```

Or something like this, to not repeat `could also` immediately.

k-distribution/k-tutorial/1_basic/02_basics/README.md
```diff
@@ -58,41 +76,81 @@ We can now evaluate this K term by running (from the same directory):
...
+## Structure of a K file
+
+A K file consists of one or more **modules**, of which only one is the 
+_main_ one, in the sense that no other module defined in that file **imports** 
+it. The name of this _main_ module will also be the one given as argument to 
+attribute `main-module` at compile time. 
+Modules residing in other K files are imported instead through **requires** 
+statements (similar to `include` in C/C++). You'll learn more about **imports** and 
+**requires** in [Lesson 1.5](../05_modules/README.md).
+In this lesson we continue with a closer look to a K module syntax.
```
Robertorosmaninho, line 124: I would avoid using “syntax” here if possible. We already have the keyword `syntax` and the syntax module. So, a synonym for “structure” would be better in this case.

PR #4789, opened 2025-04-03
Remove unused options from cterm_symbolic function

ehildenb:
None of the downstream semantics use the `maude_port` functionality and extra dispatch functionality of the `cterm_symbolic` at this point, so it's removed.

We also remove a check that `bug_report_id` is set, because it's only used in one place, and in that spot we anyway handle the case where it's unset:  https://github.com/runtimeverification/k/blob/32359773218dd79bc6f2c73dc973e571b3b89e1a/pyk/src/pyk/kore/rpc.py#L308

### DIFF
--- a/pyk/src/pyk/kore/rpc.py
+++ b/pyk/src/pyk/kore/rpc.py
@@ -265,9 +265,6 @@ class JsonRpcClient(ContextManager['JsonRpcClient']):
         bug_report_id: str | None = None,
         transport: TransportType = TransportType.SINGLE_SOCKET,
     ):
-        if (bug_report is None) != (bug_report_id is None):
-            raise ValueError('bug_report and bug_report_id must be passed together.')
-
         self._transport = self._create_transport(transport, host=host, port=port, timeout=timeout)
         self._req_id = 1
         self._bug_report_id = bug_report_id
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/kore/rpc.py
```diff
@@ -265,9 +265,6 @@ def __init__(
         bug_report_id: str | None = None,
         transport: TransportType = TransportType.SINGLE_SOCKET,
     ):
-        if (bug_report is None) != (bug_report_id is None):
-            raise ValueError('bug_report and bug_report_id must be passed together.')
-
```
gtrepta, line 270: Do you know why this was here before?

gtrepta (approved):
LGTM

pyk/src/pyk/kore/rpc.py
ehildenb: I'm actually not sure, it was very first introduced here: https://github.com/runtimeverification/k/pull/4480/files

But as far as I can tell, this shouldn't really be an issue. This check, I guess, would make sure that each bug report has a unique user-specified ID. But if they don't have that ID, then the user can supply it separately.

PR #4792, opened 2025-04-11
`pyk.ktool.kfuzz`: Add support for custom substitution functions in `fuzz`

bbyalcinkaya:
Adds an optional `subst_func` parameter to the `fuzz` function, allowing users to customize how substitutions are applied to the template.

By default, substitutions are applied using a bottom-up traversal on the whole template pattern. Custom substitution functions can be used to target specific parts of the configuration, enabling more optimized fuzzing.

Also includes a test case demonstrating the use of a custom substitution function.


ehildenb:
Since we made the change to `bottom_up`, does having the custom substitution function actually improve performance at this point? Are we using it in Stylus or Skribe?

bbyalcinkaya:
> Since we made the change to `bottom_up`, does having the custom substitution function actually improve performance at this point? Are we using it in Stylus or Skribe?

The performance improvement from `bottom_up` alone is very little. To actually benefit from it in Komet, we had to implement additional semantic rules for initializing the configuration and introduce extra logic in the fuzzing template construction. This added unnecessary complexity to the overall setup. We’ll likely need to implement the same workaround in Skribe, too.

I think adding the `subst_func` option offers a cleaner solution to that performance issue. It keeps the performance concerns of fuzzing localized within the fuzzer, without requiring semantic changes or special handling in the user code.

pyk/src/pyk/kore/manip.py
```diff
@@ -66,19 +84,12 @@ def elim_aliases(pattern: Pattern) -> Pattern:
 
     def inline_aliases(pattern: Pattern) -> Pattern:
         match pattern:
-            case And(_, (p, EVar(name))):
-                aliases[name] = p
+            case And(_, (p, EVar(_) as var)):
```
tothtamas28:
```suggestion
            case And(_, (p, EVar() as var)):
```

pyk/src/pyk/kore/manip.py
```diff
@@ -57,6 +57,24 @@ def add_symbol(pattern: Pattern) -> None:
...
+    """Substitute variables in a pattern using a bottom-up traversal.
+
+    Args:
+        pattern: The pattern containing variables to be substituted.
+        subst_map: A mapping from variables to their replacement patterns.
+    """
+
+    def subst(pattern: Pattern) -> Pattern:
+        match pattern:
+            case EVar(_) as var:
```
tothtamas28:
```suggestion
            case EVar() as var:
```

pyk/src/pyk/kore/manip.py
```diff
@@ -66,19 +84,12 @@ def elim_aliases(pattern: Pattern) -> Pattern:
...
-    def substitute_vars(pattern: Pattern) -> Pattern:
-        match pattern:
-            case EVar(name) as var:
-                return aliases.get(name, var)
-            case _:
-                return pattern
-
     pattern = pattern.bottom_up(inline_aliases)
-    pattern = pattern.bottom_up(substitute_vars)
+    pattern = substitute_vars(pattern, aliases)
```
tothtamas28, line 94: Nice function extraction. :sparkles:

tothtamas28 (commented):
LGTM. I'll let @gtrepta comment on the changes in the fuzzer.

gtrepta (approved):
LGTM

PR #4805, opened 2025-05-04
Update README in Lesson 1.3

ibastys:


ibastys:
Line 188: Why are there so many AST representations? 

Lines 188-189: Should we mention `kore` if we cannot give the reader any reference for it? Maybe another flag value is more relevant?

Lines 209-211: I could not find any other reference to `inj` attribute in the tutorial. Do you know where this is mentioned?

Lines 213-214: I added this sentence, but i am not 100% certain it is true.

Lines 451-452: Compiling with flag `--gen-bison-parser` or `--gen-glr-bison-parser` gives an additional warning. Could you say something about it?

Lines 478-502: I didn't get an error when parsing after compilation with bison flags on. When compiling with flag `--gen-glr...` i get two ASTs, correct?

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -340,48 +424,107 @@ syntax lexical LongSuffix = r"[lL]"
...
+
+```
+[Warning] Compiler: Could not find main syntax module with name
+LESSON-03-A-SYNTAX in definition.  Use --syntax-module to specify one. Using
+LESSON-03-A as default.
+[Warning] Inner Parser: Skipping modules [ML-SYNTAX] tagged as not-lr1 which
+are not supported by Bison.
+```
+
+We now have a second warning.
```
ibastys: Compiling with flag `--gen-bison-parser` or `--gen-glr-bison-parser` gives an additional warning. Could you say something about it?

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -340,48 +424,107 @@ syntax lexical LongSuffix = r"[lL]"
...
+        Lbltrue'Unds'LESSON-03-A'Unds'Boolean{}(),
+        Lblfalse'Unds'LESSON-03-A'Unds'Boolean{}()
+      ),
+      Lblfalse'Unds'LESSON-03-A'Unds'Boolean{}()
+    )
+  )
+)
+```
+
+In this case, we get both ASTs.
```
ibastys: I didn't get an error when parsing after compilation with bison flags on. When compiling with flag `--gen-glr...` i get two ASTs, correct?

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -4,32 +4,38 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+syntactic **sentences** can be used to define grammars for parsing both rules 
+and programs. In this context, you'll also learn about two additional types
+of productions, **brakets** and **tokens**.
 
 ## K's approach to parsing
 
-K's grammar is divided into two components: the **outer syntax** of K and the
-**inner syntax** of K.  Outer syntax refers to the parsing of **requires**,
+K's grammar is divided into two components: one **outer syntax** and one
+**inner syntax**.  Outer syntax refers to the parsing of **requires**,
```
mariaKt: Not a strong preference, but I actually prefer the previous phrasing here. Also, this is duplicated directly below.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -4,32 +4,38 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+module form the grammar of that module, and this grammar is used for parsing 
+programs in the language you defined.
 
 ## Basic BNF productions
 
-To illustrate how this works, we will consider a simple K definition which
-defines a relatively basic calculator capable of evaluating Boolean expressions
-containing and, or, not, and xor.
+To illustrate how this works, let's consider the K module below which defines
+a logical calculator for evaluating Boolean expressions containing operations
```
mariaKt: It looks to me that "logical" is not needed here, since you already describe that the calculator is for boolean expressions.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -43,87 +49,121 @@ module LESSON-03-A
...
-for defining the syntax of productions using a variant of
-[BNF Form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form).
+Observe that the productions in this module look a little different than
+what we have seen in the previous lesson. The reason is that K has two 
+mechanisms for defining productions:
+[BNF notation](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) and
+alphanumeric identifiers. In [Lesson 1.2](../02_basics/README.md) we presented 
+the latter, where the identifier was followed by a (possibly empty) list of 
+sorts in parentheses. In this lesson, we introduce the former, which is a more 
+generic mechanism for defining the syntax of productions.
```
mariaKt, line 58: This does not seem entirely accurate, for no other reason that the name "alphanumeric identifiers" is not a name for defining productions, and the description of this mechanism lacks the placement of `::=`, so it is not clear to which identifier we are referring. How about something along the lines of:
- [BNF Form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form), which is a notation system for defining the syntax of productions.
- The mechanism we presented in [Lesson 1.2](../02_basics/README.md), where the `::=` symbol is followed by an
alphanumeric identifier followed by a comma-separated list of sorts in parentheses. This is merely a special case of the more generic BNF Form.
In this lesson, we introduce the former.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -43,87 +49,121 @@ module LESSON-03-A
...
+For example, `Banana` is part of the syntax. Conversely, non-terminals, refer 
+to a sort name, like `Fruit`, and the syntax of the production they 
+belong to accepts any valid term of that sort at that position.
+
+In the previous lesson we executed successfully the program `colorOf(Banana())` 
+using `krun`. `krun` parses and interprets terms according to the grammar 
+defined. `Banana()` is a term of sort `Fruit`, hence a valid argument
+for function `colorOf`. Under the hood, the term is automatically converted
+into an AST (abstract syntax tree), and then the function `colorOf` is
+evaluated using the function rules provided in the definition.
```
mariaKt: The flow of this paragraph seems a bit unintuitive, e.g. there is no statement as to the program being of sort `Color`, and the explanation of why it is of sort `Color` is misplaced. You may consider something like:
In the previous lesson we executed successfully the program `colorOf(Banana())` 
using `krun`. That is because the program represented a term of sort `Color`:
indeed, `Banana()` is a term of sort `Fruit`, hence a valid argument for function
`colorOf`. `krun` parses and interprets terms according to the grammar 
defined by the developer. Under the hood, the term is automatically converted
into an AST (abstract syntax tree), and then the function `colorOf` is
evaluated using the function rules provided in the definition.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -153,69 +203,106 @@ yield the following output:
 )
 ```
 
-Note how the first output is largely a name-mangled version of the second
-output. The one difference is the presence of the `inj` symbol in the KORE
-output. We will talk more about this in later lessons.
+Comparing both outputs, you can observe that the former is largely a 
+name-mangled version of the latter. A notable difference is represented by the
```
mariaKt: I would suggest:
A notable difference is the presence of the `inj` symbol ...
instead.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -153,69 +203,106 @@ yield the following output:
...
+this ambiguity, making it impossible to write complex expressions in our small 
+language. The standard solution in most programming languages to this problem 
+is to use parentheses to indicate the appropriate grouping. K generalizes this 
+notion into a type of production called **bracket**. 
 
-```k
+A bracket production is any production with the `bracket` attribute. It is 
+required that such a production only have a single non-terminal, and the sort 
+of the production must equal the sort of that non-terminal. Otherwise, K does 
+not impose any restrictions on the grammar provided for a bracket.
```
mariaKt: This sentence: "Otherwise, K does not impose any restrictions on the grammar provided for a bracket" is a bit confusing to me. The meaning is that apart from the already specified restrictions there are no others, but the placement of otherwise makes this more difficult to infer.
I think it something like "K does not otherwise impose restrictions on the grammar the user provides for a bracket" or "K does not impose any other restrictions on the grammar the user provides for a bracket" would be more clear.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -228,98 +315,95 @@ module LESSON-03-D
...
+part of parsing a language. Also relevant is the lexical syntax of the 
+language, i.e., how the tokens are defined and recognized. 
+
+Sometimes we need to define more complex lexical syntax. Consider, for 
+instance, integers in C. They consist of a decimal, octal, or hexadecimal 
+number, followed by an optional suffix that specifies the type of the literal. 
+While it’s theoretically possible to define this syntax using a grammar, doing 
+so would be cumbersome and tedious. Additionally, you'd be faced with an AST 
+generated for the literal, which is not particularly convenient to work with. 
+
```
mariaKt: Extra space.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -228,98 +315,95 @@ module LESSON-03-D
...
+As an alternative, K allows you to define **token** productions, which are
+[regular expressions](https://en.wikipedia.org/wiki/Regular_expression) followed
```
mariaKt: The previous phrasing seems more accurate, technically: a token production consists of a regular expression followed y the token attribute.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -340,48 +424,107 @@ syntax lexical LongSuffix = r"[lL]"
...
+You can enable ahead-of-time parsing via the `--gen-bison-parser` flag to 
+`kompile`. This will make use of Bison's 
+[LR(1) parser](https://en.wikipedia.org/wiki/Canonical_LR_parser) generator. As
 such, if your grammar is not LR(1), it may not parse exactly the same as if
-you were to use the just-in-time parser, because Bison will automatically pick
+you were to use the just-in-time parser because Bison will automatically pick
 one of the possible branches whenever it encounters a shift-reduce or
 reduce-reduce conflict. In this case, you can either modify your grammar to be
-LR(1), or you can enable use of Bison's GLR support by instead passing
+LR(1), or you can use Bison's GLR support by passing flag
```
mariaKt, line 460: "instead" should be included here, to make it cleat that it is either one flag or the other.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: This output itself indicates that there is an ambiguity. The output shows that the program can be parsed in two different ways (which is happening because the the grammar is ambiguous) and as a result the generated AST contains an ambiguity. You can see it as an Amb node (`Lblamb`), that contains the two parses that are possible for the corresponding term: the first with `||` having higher priority and the second with `&&` having higher priority (as usual).

mariaKt: This section here should either not include the generated outputs, or be a bit altered to include this information, and the fact that in the first case we get one of the possible parses, the one where `||` has higher priority.

mariaKt:
> Line 188: Why are there so many AST representations?

There are representations at different levels, that are suited best for different purposes/optimizations. E.g., type/sort inference is easier at a higher level (closer to the definition's original structure). Matching is easier at a lower level, after name mangling. The outputs shown are merely different ways to print an AST (easier to read, easier to debug, convert to a different format as expected by another tool). I don't think there is a need to add any information about this at the tutorial though.

> Lines 188-189: Should we mention `kore` if we cannot give the reader any reference for it? Maybe another flag value is more relevant?

There is an exercise in this tutorial, asking about the KORE representation as well, so it seems to me that when the tutorial was created they thought that KORE was important enough to add at this time. For that reason, I myself would not make the decision to change this, I would like some more input.

> Lines 213-214: I added this sentence, but i am not 100% certain it is true.

That is true.

mariaKt (changes requested):
I have suggested some changes in the text, but overall this looks good. Thank you @ibastys .

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Reverted to the initial phrasing.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Removed "logical".

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Yes, you are right, my change is not accurate. I have rephrased this part to combine the initial text and your suggestion. Please have a look again.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Thanks, i used your formulation, but with a minor change. :)

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Removed.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Rephrased

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Added "instead".

k-distribution/k-tutorial/1_basic/03_parsing/README.md
dwightguth: this warning is basically a side effect of not explicitly specifying the module used for parsing as described by the first warning. It ends up falling back on a default behavior which creates a parser from a grammar not designed to be fed into bison. Under normal circumstances, a faithful recreation of this grammar would cause bison to likely crash or behave very badly, so instead of having that happen, we explicitly exclude the module that bison doesn't know how to handle from the grammar it is asked to generate a parser for. this warning appears to warn you that you have attempted to generate a grammar from a module that is known to cause problems for bison and that that module, despite being part of the grammar you requested, was actually not part of the grammar generated and all of its productions have been excluded from the language the generated parser recognizes.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
dwightguth: by default bison uses an LR1 algorithm which generates a single parse tree even if the grammar is ambiguous. Due to the limitations of LR1 parsing, it is also limited in the degree to which it can intelligently solve shift-reduce and reduce-reduce conflicts in the grammar, which means that even if you correctly describe how to disambiguate these ambiguities in the grammar, the algorithm may not be powerful enough to know how to do so.

The --gen-glr-bison-parser flag is designed to tell the parser generator to use a GLR algorithm instead, which generates the complete parse forest. This is then disambiguated using a couple different mechanisms. However, if disambiguation fails, it will return a term of this form telling you that the string you attempted to parse can be interpreted more than one different way. There are ways to then use semantic rules to disambiguate in this case, but that is well beyond the scope of this lesson. It is expected at this point in the student's progression that if they are using a GLR parser for an ambiguous grammar and parse a sentence that is ambiguous, it is likely some kind of issue they should try to resolve by determining which parse is the correct one and attempting to apply some kind of mechanic to disambiguate.

ibastys: Thanks, @mariaKt and @dwightguth . I have rephrased to capture your input here.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: I have added a compressed explanation to the second warning. Not very proud of it though :)

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: This is good! Just add "comma-separated" to describe the list of sorts.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: This is good, thank you. I noticed there is an additional "a" here: are such **a** sequences of characters, if you could please double check and remove it.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: This may have gone unnoticed due to there being many comments, so I am just pointing it out again.
https://github.com/runtimeverification/k/pull/4805#discussion_r2098892176

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: Correct me if I am wrong, but I think that instead would be better placed here:
... by passing flag `--gen-glr-bison-parser` to `kompile` instead.
Please edit, if you agree.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: Great, thank you.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
mariaKt: Great, thank you!

mariaKt (approved):
I have added a couple of comments (one was already there, but was probably missed due to the conversation being hidden). They are minor and easy to address, so I will approve this now, and feel free to merge after you do so. If there is anything unclear, please let me know.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
```diff
@@ -43,87 +47,123 @@ module LESSON-03-A
...
-corresponding to the Boolean operators for and, or, not, and xor. We have also
-given a syntax for each of these functions based on their syntax in the `C`
-programming language. As such, we can now write programs in the simple language
-we have defined.
-
-Input the following program into your editor as `and.bool` in the same
-directory:
+**terminals** and **non-terminals**. A terminal denotes a fixed sequence of
+characters that is a verbatim part of the syntax of that production. For 
+example, `Banana`, `(`, `)`, or `colorOf` are such a sequences of characters
```
mariaKt, line 88: This is good, thank you. I noticed there is an additional "a" here: are such **a** sequences of characters, if you could please double check and remove it.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Done

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Good find. It's a typo. Removed the "a".

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: Moved it to the end of the sentence.

k-distribution/k-tutorial/1_basic/03_parsing/README.md
ibastys: I chose the formulation with "K does not otherwise ..."

PR #4808, opened 2025-05-06
Factor out CTermShow class for simpler node printing

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/pull/4819~

This is a refactoring to how our display of nodes/KCFGs works, ideally to make it easier to thread options through to them. We're specifically trying to capture all the common options for how `CTerm` should be displayed in one place, and remove those options from other places. This came about as I was trying to implement some asks from the KMIR team on changing the display in the TUI viewer, and realized that currently all the options are threaded through in different ways and controlled at different points.

This is a breaking change in pyk API, and so corresponding changes into downstream repos will need to be added to the update-deps PRs.

Another big change that happens here: the `kcfg` directory no longer needs a `KPrint` because we use a `kast.pretty.PrettyPrinter` directly instead. Before, it was using `KPrint.pretty_print`, which was just constructing a `PrettyPrinter` then using it. So we skip the intermediate class. That means that several classes which took in a `KPrint` now just take in a `KDefinition`.

Changes in particular:

- A new class `CTermShow` is introduced, which enables printing out `CTerm` using a `PrettyPrinter`, via `CTermShow.show_config`, `CTermShow.show_constraints` and `CTermShow.show`. It takes several options controlling how printing can occur, including:
  - `extra_unparsing_modules`, `patch_symbol_table`, `unalias`, and `sort_ac_collections`: all the same as `PrettyPrinter`.
  - `minimize`: collapse unused cell-variables into `...`, and the recursively collapse the `...` as much as possible in the configuration.
  - `break_cell_collections`: enable line-breaking collections (`List`, `Set`, `Map`) which are stored directly in cells, as usually you want one item printer per line in this case.
  - `omit_labels`: a set of labels to always replace occurrences of with `...`, so they do not get printed.
- `NodePrinter` (and `APRProofNodePrinter`) now take a `CTermShow` instead of a `KPrint`, which also means they don't take the options controlling printing directly (like `minimize`). These options must be set when building teh `CTermShow` that is passed in for the `NodePrinter` to use.
- Similarly, `KCFGShow` (and `APRProofShow`) now take a `KDefinition` instead of a `KPrint` and build a `PrettyPrinter` directly. This means that the `NodePrinter` (and thus `CTermShow`) is responsible for setting options related to node printing like `sort_collections` ahead of time.
- The `KCFGViewer` (and `APRProofViewer`) are now taking directly a `KDefinition` instead of a `KPrint`, and similar to the `*Show` counterparts offloading decisions about how to print items to how the `CTermShow` is setup in the client application.
- Tests are updated with the new pattern (of constructing a `CTermShow` ahead of time and passing it in), which provides an example of what's needed downstream. In particular, before we were mocking an entire `KPrint` in the unit tests, but now we only need to mock a `CTermShow`, which is much more compact. We rename the file from `mock_kprint.py` to `mock.py`, so it can be for all mocks.


ehildenb:
On my local version, I'm able to remove the dependency from `kcfg -> ktool`. This relies on removing the use of `KPrint` for rendering nodes and such in the `show/tui` modules.

Original `develop`:

k-before

And after:

k

This PR is the  first step in that refactoring.

ehildenb:
Downstream repos to open branches into:

- [x] https://github.com/runtimeverification/evm-semantics
  - Update: https://github.com/runtimeverification/evm-semantics/pull/2751
  - To this PR: https://github.com/runtimeverification/evm-semantics/pull/2764
- [x] https://github.com/runtimeverification/mir-semantics
  - Update: https://github.com/runtimeverification/mir-semantics/pull/565
  - To this PR: https://github.com/runtimeverification/mir-semantics/pull/585
- [x] https://github.com/runtimeverification/wasm-semantics
  - Update: https://github.com/runtimeverification/wasm-semantics/pull/724
  - To this PR: https://github.com/runtimeverification/wasm-semantics/pull/725
- [x] https://github.com/runtimeverification/riscv-semantics
  - Update: https://github.com/runtimeverification/riscv-semantics/pull/86
  - To this PR: https://github.com/runtimeverification/riscv-semantics/pull/113
- [x] https://github.com/runtimeverification/skribe
  - Update: https://github.com/runtimeverification/skribe/pull/12
  - To this PR: https://github.com/runtimeverification/skribe/pull/14
- [x] https://github.com/runtimeverification/komet
  - Update: https://github.com/runtimeverification/komet/pull/79
  - To this PR: https://github.com/runtimeverification/komet/pull/80
- [x] https://github.com/runtimeverification/kontrol
  - Update: https://github.com/runtimeverification/kontrol/pull/1008
  - To this PR: https://github.com/runtimeverification/kontrol/pull/1032

ehildenb:
Import graphs:

Before:

k-before

After:

k

ehildenb:
Adjustments to import graph.

Before:

k-before

After:

k

In particular, `kcfg` is no longer importing `ktool`.

ehildenb:
I have one more refactoring in mind, which should eliminate the need for eth `NodePrinter` class heirarchy (in favor of customizing `CTermShow` class and passing it around), but that should wait for a follow-up and is a bit more invasive.

Once that refactoring s in, our display heirarchy should be significantly simplified.

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,112 @@
...
+    from typing import Final
+
+    from ..kast.inner import KInner
+    from ..kast.outer import KDefinition, KFlatModule
+    from ..kast.pretty import SymbolTable
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class CTermShow(PrettyPrinter):
```
tothtamas28: IIUC `CTermShow` is for printing `CTerm` instances, whereas `PrettyPrinter` is for printing `KAst` instances, so in this case composition is sufficient I think:

```suggestion
@final
class CTermShow:
    printer: PrettyPrinter
```

Then you can probably remove

```
definition
extra_unparsing_modules
patch_symbol_table
unalias
sort_collections
```

from the `__init__` argument list.

tothtamas28: In fact, I'd reduce coupling between these classes a bit further: all this class requires is a `Callable[[KInner], str]]`. So I'd store that as an attribute, and provide something like

```python3
@staticmethod
def from_pretty_printer(printer: PrettyPrinter) -> CTermShow: ...
```

that creates the appropriate closure from the `PrettyPrinter`. Similarly, we can provide

```python3
@staticmethod
def from_formatter(formatter: Formatter) -> CTermShow: ...
```

and other implementations as needed.

pyk/src/pyk/kcfg/show.py
```diff
@@ -30,31 +32,26 @@
...
     from ..kast.outer import KDefinition, KFlatModule, KImport, KSentence
-    from ..ktool.kprint import KPrint
     from .kcfg import NodeIdLike
 
 _LOGGER: Final = logging.getLogger(__name__)
 
 
 class NodePrinter:
-    kprint: KPrint
+    cterm_show: CTermShow
```
tothtamas28, line 41: Target for future refactorings: this class should be made abstract, as what details I want to show for a node (and how) varies from definition to definition (and even for a given definition there might be more than a single useful view). Displaying a `CTerm` as in this current implementation is a reasonable default implementation.

Also, the abstract interface should probably define separate abstract functions for providing a list of labels that should be shown in the tree, and for displaying the actual node content.

pyk/src/pyk/kcfg/show.py
```diff
@@ -74,12 +71,12 @@ def node_attrs(self, kcfg: KCFG, node: KCFG.Node) -> list[str]:
 
 
 class KCFGShow:
-    kprint: KPrint
+    pretty_printer: PrettyPrinter
     node_printer: NodePrinter
```
tothtamas28, line 75: What's the difference between `pretty_printer` and `node_printer`? One is for displaying edges, the other is for nodes?

pyk/src/pyk/kcfg/tui.py
```diff
@@ -173,8 +173,9 @@ def on_click(self, click: Click) -> None:
 
 
 class NodeView(Widget):
-    _kprint: KPrint
+    _definition: KDefinition
     _custom_view: Callable[[KCFGElem], Iterable[str]] | None
+    _cterm_show: CTermShow
```
tothtamas28, line 178: Target for future refactorings: abstract interfaces should be extracted.

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,112 @@
...
+        ret_strs: list[str] = []
+        for constraint in cterm.constraints:
+            constraint_strs = self.print(constraint).split('\n')
+            if len(constraint_strs) > 0:
+                constraint_strs = [f'#And {cstr}' for cstr in constraint_strs]
+                ret_strs.append(constraint_strs[0])
+                ret_strs.extend([f'  {constraint_str}' for constraint_str in constraint_strs[1:]])
+        return ret_strs
+
+    def show(self, cterm: CTerm) -> list[str]:
```
tothtamas28: Consider reordering these functions: public functions on top, private helpers after.

```
show
show_config
show_constraints
_break_cell_collections
_hide_labels
```

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,112 @@
...
+            extra_unparsing_modules,
+            patch_symbol_table,
+            unalias,
+            sort_collections,
+        )
+        self.minimize = minimize
+        self.break_cell_collections = break_cell_collections
+        self.omit_labels = tuple(omit_labels)
+
+    def let(
```
tothtamas28, line 43: If `PrettyPrinter` or the `Callable` is made an attribute, then this function can be removed altogether: the user can either set the remaining parameters (`minimize`, `break_cell_collections`, `omit_labels`) directly, or they can be made arguments of the `show*` methods.

tothtamas28:
> I have one more refactoring in mind, which should eliminate the need for eth `NodePrinter` class heirarchy (in favor of customizing `CTermShow` class and passing it around), but that should wait for a follow-up and is a bit more invasive.

I see two issues that I think should be addressed when refactoring these classes.

1. Parallel class hierarchies: `KCFGShow`, `KCFGViewer`, `NodePrinter` vs `APRProofShow`, `APRProofViewer`, `APRProofNodePrinter`. My current understanding is that the only difference between these is that when printing a proof, there's a different set of node labels we want to show.

2. Coupling: these printing utilities (`KCFGShow`, `KCFGViewer` and the `APRPRoof` variants) depend on layers of concrete implementations of how a KAST term is printed (`NodePrinter`, which depends on `PrettyPrinter`, which depends on `KDefinition`, ...). Instead, they should define their own interface in terms of abstractions (e.g. a function that takes a `Node` and a `KCFGViewer`-specific set of configuration parameters, and returns a `str` for the node content to be printed). This makes configuration of the tools for a given semantics downstreams much more feasible.

pyk/src/pyk/kcfg/show.py
ehildenb: The next refactoring does remove the actual node content display (which is fully moved into `CTermShow`). I was hoping to entirely remove the `NodePrinter` hierarchy, but it's somewhat complicated by the fact that `KCFGShow.pretty_segments` and `APRProofShow.pretty_segments` both take a `KCFG` and need to print node attrs, but we want to print different node attrs depending on if it's a `KCFG` or an `APRProof`. So I'm a bit stuck there, but at least we can totally lift the printing out of `NodePrinter` later (and I'll get that change in once this PR is propagated downstream).

pyk/src/pyk/kcfg/show.py
ehildenb: Ideally, we wouldn't have `NodePrinter` at all. This change is what I could do to lower from `KPrint` to `KDefinition` in this class now. Later refactoring will remove the `print_node` from `NodePrinter`, and then it will be more like `NodeAttrs` instead, which ideally would be inlined directly into `KCFG` class (thus removing the `NodePrinter` heirarchy). For now, I know I can remove `NodePrinter.print_node`, but not sure about inlining into `KCFG`.

pyk/src/pyk/cterm/show.py
ehildenb: done

pyk/src/pyk/kcfg/tui.py
ehildenb: I agree. I'm playing with that a bit on my machine, but I think we need a bit of a planning session around that. Currently, this change is an improvement that gets us some benefit and a slighlty cleaner interface, and I have a few more improvements planned, but I need to propagate it downstream before going too far I think.

pyk/src/pyk/cterm/show.py
ehildenb: Ok, I've done this. Instead of providing static methods on `CTermShow` instead I've added a `property PrettyPrinter.printer` and `property Formatter.printer`, so it's easy to get the needed `Callable[[KInner], list[str]]` directly out of those classes for constructing the `CTermShow`.

ehildenb:
1. I want to tackle this one. I will start by first getting rid of `print_node` in the `NodePrinter` heirarchy. Then it would be nice to handle the `attrs` directly at the `KCFG` level, but this is tricky because the attributes are different based on how you interpret the KCFG (as a raw KCFG or as an `APRProof`).
2. Also agreed. One step/improvement at a time though, this is what I have time for to get KMIR printing to look better, and I want to get it merged downstream all the way first, before introducing more potential breakages.

pyk/src/pyk/cterm/show.py
ehildenb: It would be nice for it to be possible to pass these arguments directyl to teh `show`. But I think overall, it ends up being less code to just set them once at class construction and go from there. Otherwise, all the points that are holding onto a `CTermShow` for printing need to also hold teh arguments `minimize` and others to determine how to print _that time_, instead of the user setting it up once up front.

pyk/src/pyk/cterm/show.py
tothtamas28: > Instead of providing static methods on `CTermShow` instead I've added a property `PrettyPrinter.printer` and property `Formatter.printer`

1. I think the interface should be `Callable[[KInner], str]` as it is simpler (`Formatter` is already a `Callable[[KInner], str]` so it could even be passed directly). Also, since the `split` has to be done in both providers, it could be done internally `CTermShow` where necessary (by calling a helper `_print_lines` or similar).

2. I think the `staticmethod` approach is cleaner: when reading the code it is immediately obvious what the intention is. On the other hand, if you add `printer` methods, in a way, you introduce a dependency in the wrong direction.

pyk/src/pyk/cterm/show.py
tothtamas28: Makes sense.

pyk/src/pyk/kast/formatter.py
```diff
@@ -76,6 +77,10 @@ def format(self, term: KInner) -> str:
             term = add_brackets(self.definition, term)
         return ''.join(self._format(term))
 
+    @property
+    def printer(self) -> Callable[[KInner], list[str]]:
+        return lambda k: self.format(k).split('\n')
```
tothtamas28: This feels like an implicit dependency in the wrong direction: this method is only introduced so that `CTermShow` can be instantiated with a `Fomatter`.

Instead, consider the `@staticmethod` approach, which makes the dependency between these two classes explicit.

pyk/src/pyk/kast/pretty.py
```diff
@@ -76,6 +76,10 @@ def symbol_table(self) -> SymbolTable:
             self._patch_symbol_table(symb_table)
         return symb_table
 
+    @property
+    def printer(self) -> Callable[[KInner], list[str]]:
+        return lambda k: self.print(k).split('\n')
+
```
tothtamas28: Same here.

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,97 @@
...
+    from collections.abc import Callable, Iterable
+    from typing import Final
+
+    from ..kast.inner import KInner
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class CTermShow:
+    _printer: Callable[[KInner], list[str]]
```
tothtamas28: Consider

```suggestion
    _printer: Callable[[KInner], str]
```

Reason:
* It's a simpler interface.
* Both `Formatter` and `PrettyPrniter` produce `str`. (In fact, `Formatter` is already a `Callable[[KInner], str]`)

Then `print` can be replaced by `_print_lines` which does the split.

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,97 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+    from typing import Final
+
+    from ..kast.inner import KInner
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class CTermShow:
```
tothtamas28, line 22:
```suggestion
@dataclass
class CTermShow:
```

pyk/src/pyk/cterm/show.py
```diff
@@ -0,0 +1,97 @@
...
+    _printer: Callable[[KInner], list[str]]
+    minimize: bool
+    break_cell_collections: bool
+    omit_labels: tuple[str, ...]
```
tothtamas28: I'd make these private too.

pyk/src/pyk/cterm/show.py
ehildenb: I switched to `Callable[[KInner], str]`, which makes it so that I can just pass `PrettyPrinter.print` and `Formatter.print` directly as teh arguments to `CTermShow` constructor. Because this is pretty straightforward, I didn't add the `@staticmethod def from_...` options (it's not really any shorter).

pyk/src/pyk/kast/formatter.py
ehildenb: I removed this.

pyk/src/pyk/kast/pretty.py
ehildenb: Removed.

pyk/src/pyk/cterm/show.py
ehildenb: Done.

pyk/src/pyk/cterm/show.py
ehildenb: Done.

pyk/src/pyk/cterm/show.py
ehildenb: Done.

pyk/src/pyk/cterm/show.py
ehildenb: One benefit of `list[str]` as a return is that we can represent the empty return (no lines returned), whereas `''` always it at least one line returned.

tothtamas28: > Because this is pretty straightforward, I didn't add the `@staticmethod`

Makes sense.

> I can just pass `PrettyPrinter.print` and `Formatter.print`

For `Formatter`, you can even pass the instance directly. The same can be achieved for `PrettyPrinter` by implementing `__call__` with the right signature.

tothtamas28 (approved):

PR #4811, opened 2025-05-12
Remove unused private attribute from `KProve`

tothtamas28:
Eliminates a (package level) circularity between `pyk.kcfg` and `pyk.ktool`.

### DIFF
--- a/pyk/src/pyk/ktool/kprove.py
+++ b/pyk/src/pyk/ktool/kprove.py
@@ -30,7 +30,6 @@ if TYPE_CHECKING:
 
     from ..kast.outer import KClaim, KRule, KRuleLike
     from ..kast.pretty import SymbolTable
-    from ..kcfg import KCFGExplore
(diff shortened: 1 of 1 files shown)


tothtamas28:
```
pydeps -o k.png -T png --no-show --cluster --max-bacon 2 --max-module-depth 2 pyk/src/pyk
```

Before:
before

After:
after

ehildenb (approved):

ehildenb:
~I find it weird that `ktool` is included into `kcfg`, but I guess that's because `KCFGExplore` uses `ktool`. I'd like to have kcfg be a bit more "pure" of a datastructure, so maybe we move the `explore.py` and `exploration.py` modules into `proof` instead, since that's really where they are used.~

We have some other work to remove the use of `KPrint` from `kcfg` class too: https://github.com/runtimeverification/k/pull/4808

PR #4816, opened 2025-05-14
Update Lesson 1.5 README.

ibastys:


k-distribution/k-tutorial/1_basic/05_modules/README.md
```diff
@@ -10,118 +10,132 @@ definition.
...
-differences between importing a module and simply including its sentences in 
-another module directly, but we will cover these differences later. 
-Essentially, you can think of modules as a way of conceptually grouping 
-sentences in a larger K definition.
+Let's return to the K definition in `lesson-05-d.k` above. Note that it is
+equivalent to the definition expressed by the single module `LESSON-05-C`.
+Essentially, by importing module `A` into module `B`, we include all sentences
+in module `A` into module `B`. Note that K distinguishes between importing a 
+module and simply including its sentences in another module directly. We 
+will discuss these differences in Lesson [XYZ]().
```
ibastys, line 107: Do you happen to know in which lesson is this discussed?

k-distribution/k-tutorial/1_basic/05_modules/README.md
```diff
@@ -10,118 +10,132 @@ definition.
...
 
-As you may have noticed, each module in a definition can express a distinct set
-of syntax. When parsing the sentences in a module, we use the syntax
-**of that module**, enriched with the basic syntax of K, in order to parse
-rules in that module. For example, the following definition is a parser error
-(`lesson-05-e.k`):
+As you may have noticed, each module in a definition can express a **distinct** 
+set of syntax. When parsing the sentences in a module, we use the syntax
+**of that module**, enriched with the basic syntax of K. For example, the 
+following definition is a parser error (`lesson-05-e.k`):
```
ibastys, line 124: This parser error mentioned here refers to K's own parser used at compile time, right? The outer syntax parser, which is different than the inner syntax parser generated after a successful `kompile` execution, right?

Do you think i should make this more explicit in the tutorial or is it understandable as it is right now?

k-distribution/k-tutorial/1_basic/05_modules/README.md
dwightguth: There is differences, but I actually wouldn't expect them to meaningfully affect the behavior of the module that would be doing the import except in fringe cases. I don't believe we currently have any lessons diving into that type of detail

k-distribution/k-tutorial/1_basic/05_modules/README.md
dwightguth: If you feel you need to confirm, it probably could be clearer. Both outer parsing of modules and inner parsing of sentences happen when the definition is compiled, though. We generally refer to parsing that happens at runtime as program parsing and it uses a pretty different code path than the one used to parse rules, which is what is being discussed here.

k-distribution/k-tutorial/1_basic/05_modules/README.md
ibastys: But then maybe it doesn't make sense mentioning this at all at this moment. I can just remove the sentences after "Note that K distinguishes". OK?

k-distribution/k-tutorial/1_basic/05_modules/README.md
ibastys: OK, then i might leave this as it is. I'll make another pass over the lesson to make sure nothing is ambiguous.

PR #4818, opened 2025-05-14
Update Lesson 1.4 README.

ibastys:


k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+Instead, programming language designers have developed methods for 
+disambiguating language expressions by making use of operator precedence and 
+associativity directions for them. It is often possible to remove _all_ 
+ambiguities in a grammar with these methods, as they instruct the parser to 
+accept some ASTs in favor of others.
+
+In general, grammars can be rewritten to remove unwanted parses. However,
+in K, the grammar specification and AST generation are intrinsically linked,
+so we discourage this approach. You will still learn how to do this in K 
+towards the end of this lesson. Now we continue with showing you how to
```
ibastys, line 28: Is this true?

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+Cluttering the code with brackets to remove ambiguity is not an ideal solution. 
+Instead, programming language designers have developed methods for 
+disambiguating language expressions by making use of operator precedence and 
+associativity directions for them. It is often possible to remove _all_ 
+ambiguities in a grammar with these methods, as they instruct the parser to 
+accept some ASTs in favor of others.
+
+In general, grammars can be rewritten to remove unwanted parses. However,
+in K, the grammar specification and AST generation are intrinsically linked,
+so we discourage this approach. You will still learn how to do this in K
```
ibastys: Please check if this statement i introduced is correct.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
 
 # Lesson 1.4: Disambiguating Parses
 
-The purpose of this lesson is to teach how to use K's builtin features for
-disambiguation to transform an ambiguous grammar into an unambiguous one that
-expresses the intended ASTs.
+In this lesson you will learn how to disambiguate a grammar by using K's 
+built-in features, rather than _artificial_ operators such as brackets.
```
mariaKt: Reading this and below, I think there may be a misunderstanding - correct me if I am wrong. You present the use of brackets as an alternative way to disambiguate a grammar. The use of brackets has no effect on the grammar being ambiguous. It just gives the user the ability to specify which AST was intended **by a single term**, but all alternatives are still expressed by the grammar. Disambiguating the grammar means that only the intended ASTs are expressed - the not-intended ones previously expressed due the grammar being ambiguous are not expressed by the new unambiguous grammar.
Therefore, I don't think you should present brackets here. Also, the previous statement captures the fact that disambiguating a grammar transforms it into a (unambiguous) grammar that captures only the intended ASTs, which could make sense to include.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
-
-Consider, then, the third iteration on the grammar of this definition
+Parsing unambiguous languages is asymptotically faster than parsing ambiguous
+languages. That's why in practice, very few formal languages outside the domain
+of natural language processing are ambiguous.
+
+Cluttering the code with brackets to remove ambiguity is not an ideal solution. 
+Instead, programming language designers have developed methods for 
+disambiguating language expressions by making use of operator precedence and 
+associativity directions for them. It is often possible to remove _all_
```
mariaKt: "directions for them" is not needed here.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+Parsing unambiguous languages is asymptotically faster than parsing ambiguous
+languages. That's why in practice, very few formal languages outside the domain
+of natural language processing are ambiguous.
+
+Cluttering the code with brackets to remove ambiguity is not an ideal solution. 
+Instead, programming language designers have developed methods for 
+disambiguating language expressions by making use of operator precedence and 
+associativity directions for them. It is often possible to remove _all_ 
+ambiguities in a grammar with these methods, as they instruct the parser to 
+accept some ASTs in favor of others.
```
mariaKt: This should be either "accept some ASTs instead of others" or "reject some ASTs in favor of others".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
mariaKt: Yes, it is possible to rewrite a grammar by hand  instead of using K's mechanisms for precedence and associativity - it would involve introducing more non-terminal symbols. Is this actually included?

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,32 +4,36 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
...
+In general, grammars can be rewritten to remove unwanted parses. However,
+in K, the grammar specification and AST generation are intrinsically linked,
+so we discourage this approach. You will still learn how to do this in K 
+towards the end of this lesson. Now we continue with showing you how to
+explicitly specify the relative precedence of operators in different 
+situations in order to resolve the ambiguity.
+
+Recall that in C, `&&` binds tighter than `||`, meaning that it has higher
+precedence, meaning additionally that the expression `true && false || false` 
+has only one valid AST: `(true && false) || false`.
```
mariaKt: This could be phrased a bit differently to avoid repeating the word meaning, e.g.
"Recall that in C, `&&` binds tighter than `||`, meaning that it has higher
precedence. Therefore, the expression ..." or "Recall that in C, `&&` binds tighter than `||`, i.e., it has higher precedence, meaning that the expression ...".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -171,9 +194,9 @@ endmodule
 ```
 
 In this example, unlike the one above, `&&`, `^`, and `||` have the same
-priority. However, viewed as a group, the entire group is left associative.
-This means that none of `&&`, `^`, and `||` can appear as the right child of
-any of `&&`, `^`, or `||`. As a result of this, this grammar is also not
+priority, as are part of the same block. Addionally, the entire group is
```
mariaKt: missing word "they": as **they** are part of the same block

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -185,17 +208,14 @@ definition to generate the alternative parse.
...
-which you wish to express a priority or associativity relation over are
-co-located in the same block of productions. However, in practice this is not
-always feasible or desirable, especially as a definition grows in size across
-multiple modules.
-
-As a result of this, K provides a second way of declaring priority and
-associativity relations.
+Previously we have only considered the case where all productions expressing a 
+priority or associativity relation are co-located in the same block of 
+productions. However, in practice this is not always feasible or desirable,
```
mariaKt: I think that the previous phrasing is more accurate here, i.e., you express a priority or associativity relation over a set of productions (via the symbols specified above). So this should be "... where all of the productions over which you wish to express a priority or associativity relation are co-located ...".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -215,41 +235,49 @@ endmodule
...
 ```
-As a consequence of this, `syntax [left|right|non-assoc]` should not be used to
+
+TODO: Explain how they are different.
+
+As a consequence, `syntax [left|right|non-assoc]` should not be used to
 group together labels with different priority.
 
-## Prefer/avoid
+## Prefer/avoid productions
```
mariaKt, line 270: These are attributes, so maybe replace productions with attributes.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -4,33 +4,37 @@ copyright: Copyright (c) Runtime Verification, Inc. All Rights Reserved.
 
 # Lesson 1.4: Disambiguating Parses
 
-The purpose of this lesson is to teach how to use K's builtin features for
-disambiguation to transform an ambiguous grammar into an unambiguous one that
-expresses the intended ASTs.
+In this lesson you will learn how to disambiguate a grammar by using K's 
+built-in features, rather than _artificial_ operators such as brackets. 
+You will learn how to define a precedence order on operators and 
+associativity directions for them.
```
mariaKt: I think this would be better phrased as "define the precedence and associativity of operators.", what you you think?

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -45,77 +49,88 @@ module LESSON-04-A
...
-first `Bool` non-terminal is not preceded by any terminals, and the last `Bool`
-non-terminal is not followed by any terminals. As a result of this, we apply
-the priority rule to both children of `&&`. However, in the `()` operator,
-the sole non-terminal is both preceded by and followed by terminals. As a
-result, the priority rule is not applied when `()` is the parent. Because of
-this, the program we mentioned above successfully parses.
+`true && false || false`. As noted before, this program was ambiguous because 
+the parser could either choose `&&` to be the child of `||` or vice versa. 
+Recall figures 3-A and 3-B in previous lesson. However, because a symbol 
+with lesser priority (i.e., `||`) cannot appear as the _direct_ child of a
```
mariaKt: Now that "Recall figures 3-A and 3-B in previous lesson." is not in parentheses, the following "However" seems weirdly placed here. Maybe you could move this phrase right after the beginning of the paragraph? E.g., 
To see this more concretely, let's look again at the program
`true && false || false`, and recall figures 3-A and 3-B in previous lesson.
As noted before, this program was ambiguous because 
the parser could either choose `&&` to be the child of `||` or vice versa.  However, because a symbol ...

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -45,77 +49,88 @@ module LESSON-04-A
...
+the parser could either choose `&&` to be the child of `||` or vice versa. 
+Recall figures 3-A and 3-B in previous lesson. However, because a symbol 
+with lesser priority (i.e., `||`) cannot appear as the _direct_ child of a 
+symbol with greater priority (i.e., `&&`), the parser will **reject** the 
+parse where `||` is under the `&&` operator (Fig. 3-B). As a result, we are 
+left with the unambiguous parse `(true && false) || false`. Conversely, if 
+the user wants the other parse, they can express this with brackets by 
+explicitly writing `true && (false || false)`. This still parses successfully because the
+`||` operator is no longer the **direct** child of the `&&` operator, but of
+`()` operator, even if the bracket is not explicitly depicted in the AST.
```
mariaKt, line 80: missing word: of **the** `()` operator

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -68,38 +68,31 @@ productions in the production block, expressing their relative priority.
...
-first `Bool` non-terminal is not preceded by any terminals, and the last `Bool`
-non-terminal is not followed by any terminals. As a result, we apply the 
-priority rule to both children of `&&`. However, in the `()` operator, the 
-sole non-terminal is both preceded by and followed by terminals. Hence, the 
-priority rule is not applied when `()` is the parent. Because of this, the 
-program we mentioned above successfully parses.
+than `||` and we might have possibly reached a contradiction as `||` cannot 
+appear as a direct child of `()`. What we have not mentioned is that the
+priority rule is more complex and applies only _conditionally_. Specifically, 
+it applies in cases where the first child is either the first or last
```
mariaKt: in cases where the **first** child is either: extra word, first

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -68,38 +68,31 @@ productions in the production block, expressing their relative priority.
...
+than `||` and we might have possibly reached a contradiction as `||` cannot 
+appear as a direct child of `()`. What we have not mentioned is that the
+priority rule is more complex and applies only _conditionally_. Specifically, 
+it applies in cases where the first child is either the first or last 
+production item in the parent's production. For example, in production 
+`Boolean "&&" Boolean`, the first `Boolean` non-terminal is not preceded by 
+any terminals, and the last `Boolean` is not followed by any terminals. As a 
+result, we apply the priority rule to both children of `&&`.
+In production `"(" Boolean ")"`, the non-terminal is both preceded and followed
+by terminals `"("` and `")"`. Thus, the priority rule does apply when `()` is
```
mariaKt: "does apply" seems to have the opposite meaning, and should become "is not applied".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -45,77 +49,88 @@ module LESSON-04-A
...
+the user wants the other parse, they can express this with brackets by 
+explicitly writing `true && (false || false)`. This still parses successfully because the
+`||` operator is no longer the **direct** child of the `&&` operator, but of
+`()` operator, even if the bracket is not explicitly depicted in the AST.
+Internally, `&&` operator is viewed as an **indirect** parent, which is not 
+subject to the priority restriction.
+
+You must have noticed that `()` has been defined as having greater priority
+than `||` and we might have possibly reached a contradiction as `||` cannot 
+appear as a direct child of `()`. What we have not mentioned is that the
```
mariaKt: I feel this contains too many hypotheticals, and is not necessarily simpler than the original . What do you think about something like:
"You must have noticed that `()` has been defined as having greater priority than `||`. However `||` appears as a direct child of `()`, in what appears to be a contradiction to the priority rule."

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -127,9 +142,10 @@ symbol with equal priority; and
 * a non-associative symbol cannot appear as a direct leftmost **or** rightmost
 child of a symbol with equal priority.
 
-In C, binary operators are all left-associative, meaning that the expression
+In C, binary operators are all left-associative, meaning that expression
 `true && false && false` parses unambiguously as `(true && false) && false`,
-because `&&` cannot appear as the rightmost child of itself.
+because `&&` cannot appear as the rightmost child of itself, only the AST 
+in Fig. 4-A is valid.
```
mariaKt: I prefer your previous commit here, with
- meaning that **the** expression
- i.e., only ...

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -211,45 +223,57 @@ module LESSON-04-D
...
-user-defined name. For example, `literal` in the `syntax priority` sentence is
-used to refer to all the productions marked with the `group(literal)` attribute,
-i.e., `true` and `false`. A production can belong to multiple groups using
-syntax such as `group(myGrp1,myGrp2)`.
+of a production is used to conceptually group together sets of sentences under 
+a common user-defined name. For example, `literal` in the `syntax priority` 
+sentence is used to refer to all productions marked with the `group(literal)` 
+attribute, i.e., `true` and `false`, `atom` to all productions marked with 
+`group(atom)`, i.e., braket production, and so on and so fort. A production can 
+belong to multiple groups using syntax such as `group(myGrp1,myGrp2)`.
```
mariaKt: typo: fort should be forth

mariaKt (changes requested):
I have not finished reviewing this PR yet, but I wanted to give you my feedback so far, while I continue - I am now on the prefer-avoid section.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -211,45 +223,57 @@ module LESSON-04-D
...
 ```
 syntax left a
 syntax left b
 ```
-As a consequence of this, `syntax [left|right|non-assoc]` should not be used to
+
+`syntax left a b` places `a` and `b` in the same associativity block, meaning
+that `a` and/or `b` cannot be the rightmost child of `a` and/or `b`. The latter
+sentences don't define this restriction, as `a` and `b` are not part of the
+same group.
```
mariaKt: This is not clear alone. "don't define this restriction" means  all of the following are not defined: a not being rightmost child of a, a not being rightmost child of b, b not being rightmost child of a, b not being rightmost child of a. But you don't specify which are defined.
It could be changed to something like:
"The latter sentences instead mean that a cannot be the rightmost child of a and b cannot be the rightmost child of b, but place no restriction on a being the rightmost child of b or b being the rightmost child of a."

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -301,21 +349,46 @@ module LESSON-04-F
...
 ```
 
-Here we have added the `avoid` attribute to the `else` production. As a result,
-when an ambiguity occurs and one or more of the possible parses has that symbol
-at the top of the ambiguous part of the parse, we remove those parses from
-consideration and consider only those remaining. The `prefer` attribute behaves
-similarly, but instead removes all parses which do not have that attribute.
-In both cases, no action is taken if the parse is not ambiguous.
+Here we have added the `avoid` attribute to the `else` production. As a result, 
+when an ambiguity occurs and any of the possible parses has that symbol at the
```
mariaKt: I would recommend using the word attribute instead of symbol here, in order to be consistent throughout the paragraph.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
```diff
@@ -301,21 +349,46 @@ module LESSON-04-F
...
-when an ambiguity occurs and one or more of the possible parses has that symbol
-at the top of the ambiguous part of the parse, we remove those parses from
-consideration and consider only those remaining. The `prefer` attribute behaves
-similarly, but instead removes all parses which do not have that attribute.
-In both cases, no action is taken if the parse is not ambiguous.
+Here we have added the `avoid` attribute to the `else` production. As a result, 
+when an ambiguity occurs and any of the possible parses has that symbol at the 
+top of the ambiguous part of the parse, we discard those parses, and consider
+only the remaining ones. The `prefer` attribute behaves similarly, but instead 
+removes all parses which do not have that attribute. In both cases, no action
```
mariaKt: Above, you changed remove to discard, but not here. I suggest to use the same, either remove/removes or discard/discards in both places.

mariaKt (changes requested):
I completed the review, and asked for a couple more changes.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: I was referring to this sentence "You will still learn how remove unwanted parses in K towards the end of this lesson."

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: I noticed my mistake and rephrased to convey the initial message.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Removed it.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Used the latter formulation.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Rephrased.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: I think associativity refers to operator associativity, i.e., in arithmetic, + is associative, 5 + 4 = 4 + 5.
That's why i used associativity direction. I don't have a strong opinion here.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done. Went for "discard".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Reverted.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Rephrased.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Removed "first".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
mariaKt: If I understand correctly, it seems you wanted to differentiate between the mathematical associativity (`(a + b) + c = a + (b + c)`)  and the operator associativity: K also selecting a specific order when we indicate that an operator is associative. This is common in PLs though, I think it is OK to say operator associativity.
By the way, `5 + 4 = 4 + 5` is due to `+` being commutative (it is also associative, but that means that
`(a + b) + c = a + (b + c) `). Just in case it comes up later in the tutorial.

mariaKt: If I understand correctly, it seems you wanted to differentiate between the mathematical associativity (`(a + b) + c = a + (b + c)`)  and the operator associativity: K also selecting a specific order when we indicate that an operator is associative. This is common in PLs though, I think it is OK to say operator associativity.
By the way, `5 + 4 = 4 + 5` is due to `+` being commutative (it is also associative, but that means that `(a + b) + c = a + (b + c) `). Just in case it comes up later in the tutorial.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
mariaKt: Missed this comment: https://github.com/runtimeverification/k/pull/4818#discussion_r2100958156

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
mariaKt: Missed this comment: https://github.com/runtimeverification/k/pull/4818#discussion_r2103461217

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
mariaKt: I understand. The tutorial presents ways to remove unwanted parses. but it presents using precedence/associativity/prefer/avoid to do that, not how to rewrite the grammar. This phrasing is more clear, thank you.
I believe there is a missing **to** here: "You will still learn how to remove unwanted parses".

mariaKt (changes requested):
I responded to your comments, and I pointed out a couple of comments that may have gone unnoticed. Thank you!

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Added the missing "to".

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: I have a slight preference for keeping "productions" here. `prefer` and `avoid` are attributes of productions. I thought we can name productions carrying these attributes by the name of the attribute, just like we do with `bracket` or `token` productions. :)

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Yes, such a silly mistake, confusing commutativity with associativity. :))
I rephrased based on your suggestion.

k-distribution/k-tutorial/1_basic/04_disambiguation/README.md
ibastys: Done.

mariaKt (approved):
LGTM @ibastys .

PR #4819, opened 2025-05-15
Small changes in preparation for CTermShow

ehildenb:
These are several changes pulled out of https://github.com/runtimeverification/k/pull/4808 in order to make the delta there more manageable. These changes should not have significant breakages downstream (if any).

- `minimize_term` loses the options `abstract_labels` and `keep_cells`, which were not used downstream anyway. The `pyk print ...` command loses the corresponding options too. Test data is updated as well.
- `CTerm.constraint` is added, which returns a conjunct of all the constraints.
- The `KCFGShow` class no longer has an option for wrapping/truncating lines for you. Instead, that's left to the call-site, which can process the output how it needs.


dkcumming (approved):
LGTM

Issue #4797, opened 2025-04-24
[Nix] Version of maven dependency `com.google.code.gson` is not locked


juliankuners:
Recently, the `mvnHash` in the k nix derivation had to be changed even though nothing changed in the repository. This caused CI to break on pull requests, see [here](https://github.com/runtimeverification/k/actions/runs/14641493584/job/41098410643?pr=4782#step:6:10121) and [here](https://github.com/runtimeverification/k/actions/runs/14641493584/job/41098410643?pr=4782#step:6:10121).

The hash changing is apparently caused by a non-locked version of `com.google.code.gson`. In particular, I compared upstream with version k-7.1.211 that was still stored in my nix store. The last change of `mvnHash`, except for the [recent hash change](https://github.com/runtimeverification/k/pull/4790/commits/22d49bc06b4aef3116fde839dd07fd6421ca4a66) due to same kind of issue, happened in [7.1.210](https://github.com/runtimeverification/k/commit/54b7a9c2753edb02697b4afe8058212fdebe26f1).

Comparing the respective nix derivation outputs showed the following differences:
<details><summary>Previously</summary>
<pre><code class="lang-xml">
&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot;?&gt;
&lt;metadata modelVersion=&quot;1.1.0&quot;&gt;
  &lt;groupId&gt;com.google.code.gson&lt;/groupId&gt;
  &lt;artifactId&gt;gson&lt;/artifactId&gt;
  &lt;versioning&gt;
    &lt;latest&gt;2.12.1&lt;/latest&gt;
    &lt;release&gt;2.12.1&lt;/release&gt;
    &lt;versions&gt;
      &lt;version&gt;1.1&lt;/version&gt;
      &lt;version&gt;1.4&lt;/version&gt;
      &lt;version&gt;1.5&lt;/version&gt;
      &lt;version&gt;1.6&lt;/version&gt;
      &lt;version&gt;1.7&lt;/version&gt;
      &lt;version&gt;1.7.1&lt;/version&gt;
      &lt;version&gt;1.7.2&lt;/version&gt;
      &lt;version&gt;2.0&lt;/version&gt;
      &lt;version&gt;2.1&lt;/version&gt;
      &lt;version&gt;2.2&lt;/version&gt;
      &lt;version&gt;2.2.1&lt;/version&gt;
      &lt;version&gt;2.2.2&lt;/version&gt;
[...]

PR #4820, opened 2025-05-18
Add Maven informational lock file and make Maven build more replicable

juliankuners:
In issue #4797 a problem was observed with nix packaging of Maven projects. In particular, upstream changes in Maven repositories caused the `mvnHash` in our derivation to be invalidated even though nothing changed in our derivation. This also breaks replicability of older derivations. This happened twice in recent times due to [GSON](https://mvnrepository.com/artifact/com.google.code.gson/gson) releasing two new version of their library, as described in the respective issue #4797 from before. Maven does not support locking dependencies and transitive dependencies.

Therefore, the following potential solutions/workarounds were proposed:
- locking dependencies in a best-effort manner, potentially using the Maven plugin [`dependency-lock-maven-plugin`](https://github.com/vandmo/dependency-lock-maven-plugin)
  - `dependency-lock-maven-plugin` does not lock dependencies for builds, but rather fails the build if the used dependencies mismatch the ones specified in a lock file, which is impractical for `nix` packaging
- creating a Maven mirror that does not change and contains all dependencies required for building K
- migrating to a tool such as Gradle that supports proper lock files

Migrating to Gradle causes more development effort compared to the other solutions. Locking dependencies in a best-effort way could work good enough when coupled with the lock file generating plugin `dependency-lock-maven-plugin` providing additional verbose information for when things break. The mirror could cause additional maintenance overhead in the future. Considering that the Java code in K is currently not being worked on, a least-effort solution seems more appropriate.

Therefore, this pull request introduces the script `update-maven-locks.sh` that uses `dependency-lock-maven-plugin`  to generate lock files that are committed to the repository that can be used to determine transitive dependencies that changed version in Maven over time. The plugin is run when building the k nix derivation `k-lock` and lock files can be extracted from the output of the derivation. This is done automatically when running the new script `update-maven-locks.sh`.

Additionally, when I was locking the GSON library version, I identified that the K Maven project actually did not update the GSON version being used and that it stayed fixed due to version requirements. Instead, `mvnHash` changing was caused by a `maven-metadata-central.xml` file being included in the offline Maven repository that was generated in nix. This file includes up-to-date information of the respective Maven dependency. Excluding this file actually causes the build to break. In particular, it fails to recognize the GSON library in the Maven repository. So apparently it must be included, though I am not exactly sure why and couldn't find any documentation indicating that this file is necessary for building. I added functionality to provide these files to the Maven derivation in a reproducible manner, i.e. providing the same revision of that file instead of fetching up-to-date version from the Maven repository. This way GSON will not cause the `mvnHash` to break again. I only identified GSON using and depending on this file in the K repository.


tothtamas28 (commented):
LGTM as far as I can tell. Asking for a second opinion from @jberthold.

mvn-lock/llvm-backend/src/main/native/llvm-backend/matching/dependencies-lock.json
jberthold, line 1: A bit weird that we place this lock file inside the submodule... but I guess unavoidable.

nix/resources/maven-metadata-central.xml
jberthold, line 1: Maybe we should rename this file to indicate that it is specifically patching `gson` metadata

jberthold (approved):

mvn-lock/llvm-backend/src/main/native/llvm-backend/matching/dependencies-lock.json
juliankuners: Yes the Maven submodule llvm-backend/matching is quite nested. But the lock files are placed there solely for manual diffing, because we don't want or need the plugin to break Maven builds in case dependencies change. Nix already breaks the build due to a changed Maven dependencies hash. Therefore the way we store the lock files in `mvn-lock` doesn't matter and we could instead rename the lock files to incorporate the module name. I just went with the directory approach because this is another explicit way of naming the respective Maven module.

Do you think incorporating the module name in the lock file names is cleaner than mirroring the Maven module structure?

nix/resources/maven-metadata-central.xml
juliankuners: Good point! The is xml now placed in a new sub-directory with the following path: `nix/resources/maven-metadata-central-replacements/GSON.xml`.

mvn-lock/llvm-backend/src/main/native/llvm-backend/matching/dependencies-lock.json
jberthold: If this structure works, it is definitely fine, no renaming required..
I actually missed the top-level `mvn-lock` here and thought we are writing into the git submodule structure.

PR #4829, opened 2025-06-03
Remove check on argument `input_file` of `_kore_print`

tothtamas28:
The check is too restrictive so in certain cases it produces spurious failures.


juliankuners (approved):
LGTM. I validated that this change fixes the issue observed in runtimeverification/imp-semantics#45, see [GitHub action](https://github.com/runtimeverification/imp-semantics/actions/runs/15424684171/job/43408604036?pr=45).

PR #4831, opened 2025-06-06
Migrate from `poetry` to `uv`

tothtamas28:


juliankuners:
The configuration for automatically updating `uv2nix` and `pyproject-build-tools` was introduced here: runtimeverification/devops#242.

juliankuners:
@tothtamas28 Are we ready to review and merge this PR?

tothtamas28:
> Are we ready to review and merge this PR?

I think so. Please resolve the conflict on `flake.lock`, then we can ask for reviews.

.github/actions/with-k-docker/Dockerfile
```diff
@@ -38,8 +38,8 @@ USER user
 WORKDIR /home/user
 
 ENV PATH=/home/user/.local/bin:${PATH}
-RUN    curl -sSL https://install.python-poetry.org | python3 - --version 2.1.1 \
-    && poetry --version
+RUN    curl -LsSf https://astral.sh/uv/install.sh | sh \
```
jberthold: This should probably use the same version as the non-docker setup-pyk-env above? See https://docs.astral.sh/uv/getting-started/installation/#standalone-installer
Currently it [seems to install 0.7.13](https://github.com/runtimeverification/k/actions/runs/15738586222/job/44358225794#step:5:1664) in CI.
```suggestion
ARG UV_VERSION
RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh \
```
The version would be read from `deps/uv_release` [by the docker action script](https://github.com/runtimeverification/k/blob/7a410b3ab30fe5056a598a5c4e2b6d19995f188b/.github/actions/with-k-docker/action.yml#L40-L45).

jberthold (approved):
LGTM, all makes sense, there is just one change to make so we run the dockerised pyk tests with the correct `uv` version. Pre-approving

.github/actions/with-k-docker/Dockerfile
tothtamas28: Good catch!

juliankuners: Thanks for catching this! I missed this Dockerfile. Fixed.

tothtamas28: `UV_VERSION` should also be passed to `docker build` in the `action.yml` though, right?

jberthold: Yes, that's what I meant by "docker action script", the script within that `action.yml` file should read it and pass it as a build argument.

tothtamas28: @juliankuners, can you open a PR with the fix?

jberthold: So, after the last change, we are _still_ installing uv 0.7.13 (a double slash in the URL is taken as a single slash I guess). We have to add reading the version and passing it as a variable in `actions/with-k-docker/action.yml`

juliankuners: Addressed in pull request runtimeverification/k#4835.
This PR also fixes release workflow issues with the Dockerfile. Namely, the PATH variable wasn't populated with the user bin directory where `uv` was installed to.

PR #4835, opened 2025-06-19
Fix Dockerfiles for release workflows

juliankuners:
The `poetry` to `uv` migration pull request #4831 causes the Dockerfile builds in release workflows [on CI to fail](https://github.com/runtimeverification/k/actions/runs/15757039533/job/44414666887). This is due to `uv` not being found at runtime as the underlying `bin` directory is missing in the PATH variable. This pull request fixes this by introducing a non-root user whose local bin directory `/home/${USER}/.local/bin` is added to the PATH variable.

I tested that the Dockerfiles successfully build on my local machine.

In addition, this pull request sets the `UV_VERSION` argument for the Dockerfile in the `with-k-docker` GitHub action that was missing before, which caused the resulting uv version that was installed to be of the latest version instead of a fixed specified version.

### DIFF
--- a/.github/actions/with-k-docker/Dockerfile
+++ b/.github/actions/with-k-docker/Dockerfile
@@ -32,8 +32,8 @@ RUN rm /kframework.deb
 
 ARG USER_ID=9876
 ARG GROUP_ID=9876
-RUN    groupadd -g ${GROUP_ID} user \
-&& useradd -m -u ${USER_ID} -s /bin/sh -g user user
+RUN    groupadd -g ${GROUP_ID} user                     \
+    && useradd -m -u ${USER_ID} -s /bin/sh -g user user
 
 USER user
 WORKDIR /home/user
--- a/.github/actions/with-k-docker/Dockerfile
+++ b/.github/actions/with-k-docker/Dockerfile
@@ -29,18 +29,20 @@ RUN apt-get -y clean
 RUN rm /llvm-backend.deb
 RUN rm /kframework.deb
 
-ENV PATH=/usr/bin:${PATH}
-ARG UV_VERSION
-RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | env UV_INSTALL_DIR="/usr/bin" sh \
-    && uv --version
 
 ARG USER_ID=9876
 ARG GROUP_ID=9876
 RUN    groupadd -g ${GROUP_ID} user \
-    && useradd -m -u ${USER_ID} -s /bin/sh -g user user
+&& useradd -m -u ${USER_ID} -s /bin/sh -g user user
 
 USER user
 WORKDIR /home/user
+
+ENV PATH=/home/user/.local/bin:${PATH}
+
+ARG UV_VERSION
+RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh \
+    && uv --version
 
 ENV PATH=/home/user/.elan/bin:${PATH}
 RUN    curl -O https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh \
--- a/.github/actions/with-k-docker/Dockerfile
+++ b/.github/actions/with-k-docker/Dockerfile
@@ -29,6 +29,11 @@ RUN apt-get -y clean
 RUN rm /llvm-backend.deb
 RUN rm /kframework.deb
 
+ENV PATH=/usr/bin:${PATH}
+ARG UV_VERSION
+RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | env UV_INSTALL_DIR="/usr/bin" sh \
+    && uv --version
+
 ARG USER_ID=9876
 ARG GROUP_ID=9876
 RUN    groupadd -g ${GROUP_ID} user \
@@ -36,11 +41,6 @@ RUN    groupadd -g ${GROUP_ID} user \
 
 USER user
 WORKDIR /home/user
-
-ENV PATH=/home/user/.local/bin:${PATH}
-ARG UV_VERSION
-RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh \
-    && uv --version
 
 ENV PATH=/home/user/.elan/bin:${PATH}
 RUN    curl -O https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,6 +90,7 @@ jobs:
   pyk-build-wheel:
     name: 'Pyk: Build Python wheel'
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
       - name: 'Get uv release'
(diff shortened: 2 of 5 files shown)


package/docker/Dockerfile.ubuntu-jammy
```diff
@@ -18,16 +18,29 @@ RUN    apt-get update
...
+ARG USER=user
+ARG GROUP=user
+ARG USER_ID=9876
+ARG GROUP_ID=9876
+RUN    groupadd -g ${GROUP_ID} ${GROUP} \
+    && useradd -m -u ${USER_ID} -s /bin/sh -g ${GROUP} ${USER}
+
+RUN chown -R ${USER}:${GROUP} /home/${USER}
+USER ${USER}:${GROUP}
+WORKDIR /home/${USER}
```
tothtamas28: I wonder if this is a breaking change for downstream repos that derive their own Dockerfiles from these images.

Is there a way to install a specific version of `uv` using `pipx`, the same way we did for `poetry` before the migration?

Alternarively (and this is the ideal solution), do you see an easy way to copy the wheel into the image from the environment so that installing `uv` becomes unnecessary in the first place?

juliankuners: Sure, I can create the wheel in a separate docker container and then copy the wheel over to the current one. This way we don't have to install `uv` in the image that might be inherited in other dependent Dockerfiles.

An alternative would be installing `uv` to a custom directory as root. But since the Dockerfile is inherited in other Dockerfiles, I'd say having a Dockerfile without `uv` pre-installed is cleaner, since other Dockerfiles might want to install other versions of `uv`.

tothtamas28: You might not even need a dedicated container for that, a job on a public runner can produce the wheel for other jobs.

juliankuners: Changed it now, so that `with-k-docker/Dockerfile` installs `uv` into `/usr/bin` as root.

For the noble/jammy Dockerfiles, a wheel is built by a GitHub job and uploaded as an artifact. This is then downloaded and copied into the docker images. This way, the noble/jammy docker images do not require `uv` to be installed. This seems important as dependent Dockerfiles install their own version of `uv`.

jberthold (approved):
Changes LGTM, good idea to skip `uv` installation for downstream K images.

.github/actions/with-k-docker/Dockerfile
```diff
@@ -29,6 +29,11 @@ RUN apt-get -y clean
 RUN rm /llvm-backend.deb
 RUN rm /kframework.deb
 
+ENV PATH=/usr/bin:${PATH}
+ARG UV_VERSION
+RUN    curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | env UV_INSTALL_DIR="/usr/bin" sh \
+    && uv --version
+
```
tothtamas28: This image is only ever run as `user`, what's the advantage of installing `uv` globally? I'd say installing `uv` as `user` is closer to how a developer would run these tests.

.github/workflows/release.yml
```diff
@@ -86,10 +86,36 @@ jobs:
             export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
             kup publish --verbose k-framework-binary .#k --keep-days 180
             kup publish --verbose k-framework-binary .#k.openssl.procps.secp256k1 --keep-days 180
-      
+
+  test-package-kframework-wheel:
+    name: 'K: Python wheel'
+    runs-on: [self-hosted, linux, normal]
```
tothtamas28:
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/release.yml
```diff
@@ -86,10 +86,36 @@ jobs:
             export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
             kup publish --verbose k-framework-binary .#k --keep-days 180
             kup publish --verbose k-framework-binary .#k.openssl.procps.secp256k1 --keep-days 180
-      
+
+  test-package-kframework-wheel:
```
tothtamas28: For consistency in naming with other `pyk`-related jobs (`pyk-build-docs`, `pyk-publish`), consider

```suggestion
  pyk-build-wheel:
```

tothtamas28: Also let's add something like `timeout-minutes: 5`, just in case.

package/docker/Dockerfile.ubuntu-jammy
```diff
@@ -18,16 +18,9 @@ RUN    apt-get update
...
-    && pip install dist/*.whl \
-    && rm -rf /pyk
-
-RUN rm -rf /kframework_amd64_ubuntu_jammy.deb
+COPY kframework-*.whl ./
+RUN    pipx ensurepath                              \
+    && . /root/.profile                          \
+    && pip install /kframework-*.whl             \
+    && rm -rf /kframework-*                      \
+    && rm -rf /kframework_amd64_ubuntu_jammy.deb
```
tothtamas28: Wildcards can probably be eliminated here. Also, `-rf` is no longer needed here I think.

```suggestion
COPY kframework.whl /kframework.whl
RUN    pipx ensurepath             \
    && . /root/.profile            \
    && pip install /kframework.whl \
    && rm /kframework.whl          \
    && rm /kframework_amd64_ubuntu_jammy.deb
```

(It'd be even better to remove the `.deb` as the last step of the previous `RUN`.)

package/docker/Dockerfile.ubuntu-noble
```diff
@@ -18,16 +18,9 @@ RUN    apt-get update
...
-    && pip install dist/*.whl --break-system-packages \
-    && rm -rf /pyk
-
-RUN rm -rf /kframework_amd64_ubuntu_noble.deb
+COPY kframework-*.whl ./
+RUN    pipx ensurepath                                          \
+    && . /root/.profile                                      \
+    && pip install /kframework-*.whl --break-system-packages \
+    && rm -rf /kframework-*                                  \
+    && rm -rf /kframework_amd64_ubuntu_noble.deb
```
tothtamas28: Same here.

tothtamas28 (commented):
Looks great overall, left a few minor comments.

.github/actions/with-k-docker/Dockerfile
juliankuners: You are right. This was an unnecessary leftover.
Addressed in new commits.

.github/workflows/release.yml
juliankuners: Addressed in new commits.

.github/workflows/release.yml
juliankuners: Good idea! I added both suggestions.

package/docker/Dockerfile.ubuntu-jammy
juliankuners: I removed the `-rf` and moved the `.deb` to prior installing the wheel.

I tried out the installation of a `kframework.whl` wheel before, but pip somehow requires the filename to include the version of the wheel. Thus, I left the filename as is and use wildcards accordingly.

package/docker/Dockerfile.ubuntu-noble
juliankuners: Resolved, see above.

.github/actions/with-k-docker/Dockerfile
tothtamas28, line 1: There might be accidental whitespace changes in this file.

tothtamas28 (approved):
LGTM!

.github/actions/with-k-docker/Dockerfile
juliankuners: Fixed!

juliankuners:
Release workflow ran successfully this time: https://github.com/runtimeverification/k/actions/runs/15777902775

PR #4843, opened 2025-07-16
Add fixture `load_kllvm`

tothtamas28:
Direct imports of `pyk.kllvm.load` are removed by relying on the fixture where needed. This eliminates the 5-10s startup delay in the integration test harness, as the `kllvm` Python module is now generated during test execution rather than at test collection.

### DIFF
--- a/pyk/src/pyk/testing/plugin.py
+++ b/pyk/src/pyk/testing/plugin.py
@@ -60,3 +60,13 @@ def profile(tmp_path: Path) -> Profiler:
 @pytest.fixture(scope='session')
 def kompile(tmp_path_factory: TempPathFactory) -> Kompiler:
     return Kompiler(tmp_path_factory)
+
+
+@pytest.fixture(scope='session')
+def load_kllvm(tmp_path_factory: TempPathFactory) -> None:
+    from ..kllvm.compiler import compile_kllvm
+    from ..kllvm.importer import import_kllvm
+
+    tmp_dir = tmp_path_factory.mktemp('kllvm')
+    module_file = compile_kllvm(tmp_dir)
+    import_kllvm(module_file)
--- a/pyk/src/pyk/testing/plugin.py
+++ b/pyk/src/pyk/testing/plugin.py
@@ -64,6 +64,21 @@ def kompile(tmp_path_factory: TempPathFactory) -> Kompiler:
 
 @pytest.fixture(scope='session')
 def load_kllvm(tmp_path_factory: TempPathFactory) -> None:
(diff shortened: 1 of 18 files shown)


pyk/src/tests/integration/kllvm/test_sorts.py
```diff
@@ -1,56 +1,67 @@
...
+    assert not actual
+
+
+def test_name(load_kllvm: None) -> None:
+    from pyk.kllvm.ast import CompositeSort
 
+    # Given
+    name = 'SortInt'
 
-@pytest.mark.parametrize('name', ('A', 'SortInt', ''))
```
juliankuners, line 29: In the updated code, this test case is executed using only one of the three previously specified values for parameter `name`. This was not described in the PR, therefore I ask whether this is reasonable?

pyk/src/tests/integration/kllvm/test_load.py
```diff
@@ -2,15 +2,14 @@
...
 
-import pyk.kllvm.load  # noqa: F401
-import pyk.kllvm.load_static  # noqa: F401
-from pyk.kllvm.compiler import compile_kllvm
-
 if TYPE_CHECKING:
     from typing import Final
 
 
 def test_kllvm_module() -> None:
```
juliankuners, line 9: Maybe add `load_kllvm` here as well?

pyk/src/pyk/testing/plugin.py
```diff
@@ -60,3 +60,13 @@ def profile(tmp_path: Path) -> Profiler:
 @pytest.fixture(scope='session')
 def kompile(tmp_path_factory: TempPathFactory) -> Kompiler:
     return Kompiler(tmp_path_factory)
+
+
+@pytest.fixture(scope='session')
```
juliankuners, line 65: Please document the fixture with a short comment that explains when and why to use this fixture in tests.

juliankuners (approved):
See comments, otherwise LGTM

pyk/src/tests/integration/kllvm/test_load.py
tothtamas28: This test still imports `pyk.kllvm.load`, just in the function body.

pyk/src/tests/integration/kllvm/test_sorts.py
tothtamas28: I think it is, it just reads back the set value.

pyk/src/pyk/testing/plugin.py
tothtamas28: Good point, done.

image

PR #4852, opened 2025-09-03
Fix macos release followup

juliankuners:
The previous pull request #4852 succesfully fixed an error that made the homebrew build of k not build in CI.

It turns out that there is another unrelated error during testing of that homebrew build, see [logs](https://github.com/runtimeverification/k/actions/runs/17402621638/job/49401283081#step:6:46). brew cannot find the local bottle (brew package) when trying to install.

### DIFF
--- a/.github/workflows/release-debug.yml
+++ b/.github/workflows/release-debug.yml
@@ -178,6 +178,7 @@ jobs:
           # explicitly instead.
           JAVA_HOME: ${{ env.JAVA_HOME_17_X64 }}
           HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
+          HOMEBREW_INTERNAL_ALLOW_PACKAGES_FROM_PATHS: 1
         run: |
           # The macOS public runners are prone to flakiness when running this
           # test suite, so the PL-tutorial is disabled for now.
--- a/.github/workflows/release-debug.yml
+++ b/.github/workflows/release-debug.yml
@@ -178,7 +178,7 @@ jobs:
           # explicitly instead.
           JAVA_HOME: ${{ env.JAVA_HOME_17_X64 }}
           HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
-          HOMEBREW_INTERNAL_ALLOW_PACKAGES_FROM_PATHS: 1
+          HOMEBREW_FORBID_PACKAGES_FROM_PATHS: 0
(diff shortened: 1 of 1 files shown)


juliankuners:
In the macos image that is used in the workflows, the brew version has been updated from 4.6.3 to 4.6.6. Though [the workflow runs](https://github.com/runtimeverification/k/blob/170a601171f2776b77520817bb0268e203c05a9b/package/macos/brew-build-and-update-to-local-bottle#L5) `brew update` in CI, which makes it unclear which versions were specifically used.

It looks like recent versions of brew disallow installing packages from a local path without throwing a warning or error. There is the environment variable `HOMEBREW_FORBID_PACKAGES_FROM_PATHS` that is enabled by default, but setting it to any value, including `0` or `false`, triggers the flag, see [here](https://github.com/Homebrew/brew/blob/1c959c22cee9eb333130633b30d380766e2dd09b/Library/Homebrew/env_config.rb#L675C7-L675C13).

There is also the undocumented environment variable `HOMEBREW_INTERNAL_ALLOW_PACKAGES_FROM_PATHS` that could be set to enable local paths, but the environment variable is `unset` [by brew beforehand](https://github.com/Homebrew/brew/commit/c143b0328029453f62114f76b2598d0838693750#diff-1c05b3e9ab474d3173985859b483da062303eb9899315a5de83f02a974941da8R1106). I also tried out the environment variable in CI as well as locally and it didn't work.

Finally, there's the environment variable `HOMEBREW_DEVELOPER` that can be set, which also enabled local paths. With this, the installation progresses in CI. Though this has the disadvantage that warnings are now treated as errors.

There is now a new issue that I cannot replicate locally on my x86_64 linux machine, [see logs](https://github.com/runtimeverification/k/actions/runs/17446597510/job/49547270003#step:6:250). brew cannot find the file `/opt/homebrew/Cellar/llvm/17.0.6` after installation of `llvm@17`. Looking at previous successful CI runs and my local installation with `brew install llvm@17 -v`, brew instead reads the file `/opt/homebrew/Cellar/llvm@17/17.0.6`. Somehow `@17` disappeared for no good reason.

juliankuners:
Brew builds are currently not planned to be actively maintained by specifically RV for the foreseeable future.

MacOS builds can also alternatively be installed with kup: https://docs.runtimeverification.com/kontrol/cheatsheets/kup-cheatsheet

PR #4860, opened 2025-09-30
Skip loading circularity module for non-circularity proofs

ehildenb:
Some proofs we're loading are quite large in terms of the configuration and the unevaluated functions in it. They end up taking a long time to load the circularity module dynamically into the backend, even though it won't be used.

This simply skips loading the circularities module for proofs that are not marked as circularities. It should not affect the functionality of the prover at all, and if it does it's likely a bug where we're using the circularities module in a place we shouldn't.

This does mean, that for the pyk prover, circularities _must_ be marked as such (like the sum-spec.k which is marked as such here), otherwise they won't be included automatically as an axiom.


palinatolmach (approved):

jberthold (approved):
LGTM

PR #4875, opened 2025-10-30
Streaming support for JsonRpcServer

RaoulSchaffranek:
This PR extends the JsonRpcServer with support for streaming.
JsonRpcMethods can now return a stream of bytes (Iterator[bytes]) instead of a JSON-serializable value.
The method must ensure that the produced stream is already JSON-encoded, as no extra validation is performed.
This allows us to transfer large responses without buffering the entire response in memory.

This is needed by the kontrol-node, when transferring large execution traces.


RaoulSchaffranek:
Closing and reopening in https://github.com/runtimeverification/k/pull/4877
I messed up the git history in this PR

PR #4879, opened 2025-11-05
Streaming support for JsonRpcServer

RaoulSchaffranek:
This PR extends the JsonRpcServer with support for streaming.
JsonRpcMethods can now return a stream of bytes (Iterator[bytes]) instead of a JSON-serializable value.
The method must ensure that the produced stream is already JSON-encoded, as no extra validation is performed.
This allows us to transfer large responses without buffering the entire response in memory.

This is needed by the kontrol-node, when transferring large execution traces.

### DIFF
--- a/pyk/src/pyk/rpc/rpc.py
+++ b/pyk/src/pyk/rpc/rpc.py
@@ -118,13 +118,12 @@ class JsonRpcSuccess(JsonRpcResult):
     id: Any
 
     def encode(self) -> Iterator[bytes]:
-        yield f'{{"jsonrpc":"2.0", "id": {self.id}, "result": '.encode('utf-8')
+        yield f'{{"jsonrpc":"2.0", "id": {self.id}, "result": '.encode()
         if isinstance(self.payload, Iterator):
-            for chunk in self.payload:
-                yield chunk
+            yield from self.payload
         else:
             yield json.dumps(self.payload).encode('utf-8')
-        yield '}'.encode('utf-8')
+        yield b'}'
 
 
 @dataclass(frozen=True)
@@ -132,15 +131,15 @@ class JsonRpcBatchResult(JsonRpcResult):
     results: tuple[JsonRpcError | JsonRpcSuccess, ...]
 
     def encode(self) -> Iterator[bytes]:
-        yield '['.encode('utf-8')
+        yield b'['
         first = True
         for result in self.results:
             if not first:
-                yield ','.encode('utf-8')
+                yield b','
             else:
                 first = False
             yield from result.encode()
-        yield ']'.encode('utf-8')
+        yield b']'
 
 
 class JsonRpcRequestHandler(BaseHTTPRequestHandler):
--- a/pyk/src/pyk/rpc/rpc.py
+++ b/pyk/src/pyk/rpc/rpc.py
@@ -118,8 +118,8 @@ class JsonRpcSuccess(JsonRpcResult):
     id: Any
 
     def encode(self) -> Iterator[bytes]:
-        id_encoded = json.dumps(self.id).encode('utf-8')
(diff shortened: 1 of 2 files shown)


pyk/src/pyk/rpc/rpc.py
```diff
@@ -106,32 +108,40 @@ def to_json(self) -> dict[str, Any]:
             'id': self.id,
         }
 
-    def encode(self) -> bytes:
-        return json.dumps(self.to_json()).encode('ascii')
+    def encode(self) -> Iterator[bytes]:
+        yield json.dumps(self.wrap_response()).encode('utf-8')
```
tothtamas28, line 112: This yields the whole `bytes` value, if you want to yield each byte, you can do
```python3
data = json.dumps(self.wrap_response()).encode('utf-8')
yield from (bytes([b]) for b in data)
```

pyk/src/pyk/rpc/rpc.py
```diff
@@ -96,7 +98,7 @@ class JsonRpcError(JsonRpcResult):
     message: str
     id: str | int | None
 
-    def to_json(self) -> dict[str, Any]:
+    def wrap_response(self) -> dict[str, Any]:
```
tothtamas28, line 101: The PR description states

> This PR extends the JsonRpcServer with support for streaming.

but it introduces what look like breaking changes. Are there other clients to this library other than `kontrol-node`?

pyk/src/pyk/rpc/rpc.py
```diff
@@ -106,32 +108,40 @@ def to_json(self) -> dict[str, Any]:
...
-        }
-
-    def encode(self) -> bytes:
-        return json.dumps(self.to_json()).encode('ascii')
+    def encode(self) -> Iterator[bytes]:
+        id_encoded = json.dumps(self.id)
+        version_encoded = json.dumps(JsonRpcServer.JSONRPC_VERSION)
+        yield f'{{"jsonrpc": {version_encoded}, "id": {id_encoded}, "result": '.encode()
+        if isinstance(self.payload, Iterator):
+            yield from self.payload
```
tothtamas28, line 125: Is `self.payload` a `bytes` now? If so, the type hint can be strengthened.

pyk/src/pyk/rpc/rpc.py
```diff
@@ -106,32 +108,40 @@ def to_json(self) -> dict[str, Any]:
...
+        else:
+            yield json.dumps(self.payload).encode('utf-8')
```
tothtamas28, line 127: This also yields the whole `bytes`.

pyk/src/pyk/rpc/rpc.py
RaoulSchaffranek: Good catch. I cannot think of a use case where a client library actually uses any `JsonRpcResult` subclasses directly. They were introduced as an internal abstraction layer when I implemented batch requests. Also, I'm not aware of any other user than `kontrol-node`. Therefore, the risk is very low here and can be easily addressed in downstream repositories.

pyk/src/pyk/rpc/rpc.py
RaoulSchaffranek: Don't think we need this fine granularity. The purpose of splitting the output stream into chunks is to keep the memory usage low. For example, `kontrol-node` writes a large (>1GB) JSON response to disk. The current design requires reading these files in full into memory before sending them over the network. The new design allows us to read and transfer these files in chunks. Additionally, the new design also avoids one circle of JSON parsing and serializing the files.

pyk/src/pyk/rpc/rpc.py
RaoulSchaffranek: Not necessarily. `payload` can still be any JSON serializable Python value, or it can be a `Iterator[bytes]` object. This is to keep backwards compatibility.

tothtamas28 (approved):

PR #4895, opened 2025-12-20
Testing build K and Publish to Cachix

F-WRunTime:
### Problem
The K Framework release workflow was experiencing "Connection reset by peer" errors when uploading large files to Cachix on macOS. This occurred during multipart uploads of large dependencies (e.g., `hs-backend-booster-dev-tools` ~748MB) when using `kup publish` on the self-hosted macOS runner.

### Solution
Switch from the self-hosted macOS runner (`self-macos-latest`) to the public GitHub-hosted macOS runner (`macos-latest`). Testing confirmed that `kup publish` works reliably on the public runner, eliminating the need for more complex workarounds.

### Changes
- **`.github/workflows/release.yml`**: Changed macOS runner from `self-macos-latest` to `macos-latest` (public runner)
- The `kup publish` commands remain unchanged - no need to replace with direct `cachix push`
- All three package variants (`.k`, `.k.openssl.secp256k1`, `.k.openssl.procps.secp256k1`) continue to be published with 180-day retention

### Testing
Added a [test job](https://github.com/runtimeverification/k/actions/runs/20732226383/job/59522438372?pr=4895#step:5:14) in `test-pr.yml` that validated `kup publish` works successfully on the public macOS runner with large file uploads.

### Impact
- Minimal code changes (just runner configuration)
- Maintains existing `kup publish` workflow
- Fixes upload timeout issues on macOS
- No functional changes to the publishing process


.github/workflows/release.yml
```diff
@@ -51,8 +51,8 @@ jobs:
         include:
           - runner: ubuntu-24.04
             os: ubuntu-24.04
-          - runner: self-macos-latest
-            os: self-macos-latest
+          - runner: macos-latest
+            os: macos-15
```
ehildenb, line 59: Are both the changes to the runner and the changes to the script below necessary? Or if we change the script we don't need to change the runner? Or if we change the runner we don't need to change the script?

.github/workflows/release.yml
```diff
@@ -76,17 +76,69 @@ jobs:
...
+          TOTAL_PATHS_K_PROCPS=$(echo "${STORE_PATHS_K_OPENSSL_PROCPS_SECP256K1}" | wc -l | tr -d ' ')
+          echo "Total store paths to push for .#k.openssl.procps.secp256k1: ${TOTAL_PATHS_K_PROCPS}"
+          echo "Pushing all ${TOTAL_PATHS_K_PROCPS} store paths for .#k.openssl.procps.secp256k1 to k-framework-binary cache..."
+          echo "${STORE_PATHS_K_OPENSSL_PROCPS_SECP256K1}" | cachix push k-framework-binary
+          echo "✅ .#k.openssl.procps.secp256k1 push completed successfully"
+          
+          # Pin the main output path
+          echo "Pinning .#k.openssl.procps.secp256k1 output path with 180-day retention..."
+          cachix pin k-framework-binary "${PLATFORM}.k.openssl.procps.secp256k1" "${K_OPENSSL_PROCPS_SECP256K1_PATH}" --keep-days 180
+          echo "✅ .#k.openssl.procps.secp256k1 pinned successfully"
```
juliankuners: I feel like this is a lot of repetitive code that could be generalized in a bash function. Using a bash function instead would make maintenance easier and reduce the number of lines.

.github/workflows/release.yml
```diff
@@ -76,17 +76,69 @@ jobs:
...
+          # Pin the main output path (equivalent to kup publish --keep-days 180)
+          echo "Pinning .#k output path with 180-day retention..."
+          cachix pin k-framework-binary "${PLATFORM}.k" "${K_PATH}" --keep-days 180
+          echo "✅ .#k pinned successfully"
+          
+          # Build and push .#k.openssl.secp256k1
+          echo "Building full .#k.openssl.secp256k1 package with all dependencies..."
+          K_OPENSSL_SECP256K1_PATH=$(nix build ".#packages.${PLATFORM}.k.openssl.secp256k1" --no-link --json | jq -r '.[].outputs.out')
+          DRV_K_OPENSSL_SECP256K1=$(nix-store --query --deriver "${K_OPENSSL_SECP256K1_PATH}")
+          STORE_PATHS_K_OPENSSL_SECP256K1=$(nix-store --query --requisites --include-outputs "${DRV_K_OPENSSL_SECP256K1}")
```
juliankuners: This snippet is taken from workflows that push to our `k-framework` cachix that includes built binaries and build-time dependencies. But in k-framework-binary, we only store built binaries without their build-time dependencies.

Therefore, this could should be changed to only push the binaries. See [`nix-store --query` docs](https://nix.dev/manual/nix/2.26/command-ref/nix-store/query#queries), specifically regarding flags `--requisites` and `--include-outputs`.

From what I understand, you would want to leave out the step where you query the derivation with `nix-store --query --deriver` and instead just query directly on the output path obtained from `nix build ... | jq ..` without the use of both `--requisites` and `include-outputs`.

Aka, something like this:
```
K_OPENSSL_SECP256K1_PATH=$(nix build ".#packages.${PLATFORM}.k.openssl.secp256k1" --no-link --json | jq -r '.[].outputs.out')
STORE_PATHS_K_OPENSSL_SECP256K1=$(nix-store --query "${K_OPENSSL_SECP256K1_PATH}")
```

On another note, I really don't know why we are separating `k-framework` and `k-framework-binary` in the first place, other than reducing the minimum trust base that for users of our cache that only opt-in to use our built binaries and not our build-time dependencies as a trusted nix cache.

juliankuners (changes requested):

.github/workflows/release.yml
F-WRunTime: It was a chosen pattern to use. However the separation does create repeated need to upload artifacts between PR and release which I would say just using 1 would reduce complexity and srpead. I would argue we just pick one and use it? Should we open that to a broader K group discussion and then open a new PR to consolidate? 
I'm open to this

F-WRunTime:
Decided to go with least amount of change needed although dependency on  kup would be simpler, better to not poke the bear and make more changes that necessary

.github/workflows/release.yml
juliankuners: Yeah, I researched a bit more to make sure there's nothing I oversee, but it looks to me as if we could just merge them. However, iirc there was only one cachix cache in the past, so I wonder what motivated the team to go with two this time instead.

Yes, let's start a K group discussion.

juliankuners (approved):

ehildenb (approved):

.github/workflows/release.yml
```diff
@@ -76,17 +80,18 @@ jobs:
...
-            kup publish --verbose k-framework-binary .#k.openssl.procps.secp256k1 --keep-days 180
+        run: |
+          # Install kup
+          export PATH="$(nix build github:runtimeverification/kup --no-link --json | jq -r '.[].outputs | to_entries[].value')/bin:$PATH"
+          
+          # Publish all three package variants using kup publish
+          # Using public macOS runner has proven reliable for large file uploads
+          kup publish --verbose k-framework-binary .#k --keep-days 180
+          kup publish --verbose k-framework-binary .#k.openssl.secp256k1 --keep-days 180
+          kup publish --verbose k-framework-binary .#k.openssl.procps.secp256k1 --keep-days 180
```
ehildenb: It looks like shellcheck wants the `with: ... packages: jq .... script: | ...` put back in?

PR #4907, opened 2026-04-08
pyk: add `node_id` to `custom_step`

juliankuners:
This PR adds a new parameter that is passed to custom steps in pyk. Specifically, the `node_id` is passed so that it can be used in downstream custom steps.

This is required for a downstream PR in kontrol (https://github.com/runtimeverification/kontrol/pull/1131) that adds the node id to console log cheatcode logging, which is implemented by means of a custom step.

This PR will require adjustments in some downstream dependencies as well such as kevm. However, not all downstream dependencies make use of custom steps such as, e.g., wasm-semantics and komet.

### DIFF
--- a/pyk/src/pyk/kcfg/semantics.py
+++ b/pyk/src/pyk/kcfg/semantics.py
@@ -35,7 +35,7 @@ class KCFGSemantics(ABC):
     """Check whether or not the semantics can make a custom step from a given ``CTerm``."""
 
     @abstractmethod
-    def custom_step(self, c: CTerm, cs: CTermSymbolic) -> KCFGExtendResult | None: ...
+    def custom_step(self, c: CTerm, cs: CTermSymbolic, node_id: int) -> KCFGExtendResult | None: ...
 
     """Implement a custom semantic step."""
 
@@ -61,7 +61,7 @@ class DefaultSemantics(KCFGSemantics):
     def can_make_custom_step(self, c: CTerm) -> bool:
         return False
 
-    def custom_step(self, c: CTerm, cs: CTermSymbolic) -> KCFGExtendResult | None:
+    def custom_step(self, c: CTerm, cs: CTermSymbolic, node_id: int) -> KCFGExtendResult | None:
         return None
 
     def is_mergeable(self, c1: CTerm, c2: CTerm) -> bool:
--- a/pyk/src/pyk/kcfg/semantics.py
+++ b/pyk/src/pyk/kcfg/semantics.py
@@ -40,12 +40,12 @@ class KCFGSemantics(ABC):
     """Implement a custom semantic step.
     
     Args:
-        c: Current configuration term representing the EVM state.
-        cs: Symbolic configuration term.
+        c: Current constrained term representing the state.
+        cs: ``CTermSymbolic`` for computing the custom step result.
         node_id: Current node id.
     
     Returns:
-        The KCFGExtendResult produced by this custom step if this custom step can produce one, otherwise returns None, 
+        The ``KCFGExtendResult`` produced by this custom step if this custom step can produce one, ``None`` otherwise.
     """
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


pyk/src/pyk/kcfg/semantics.py
```diff
@@ -35,7 +35,7 @@ def can_make_custom_step(self, c: CTerm) -> bool: ...
     """Check whether or not the semantics can make a custom step from a given ``CTerm``."""
 
     @abstractmethod
-    def custom_step(self, c: CTerm, cs: CTermSymbolic) -> KCFGExtendResult | None: ...
+    def custom_step(self, c: CTerm, cs: CTermSymbolic, node_id: int) -> KCFGExtendResult | None: ...
```
tothtamas28, line 38: What's the intended semantics of the `node_id`? Should it be optional?

juliankuners: I  don't see why it should be optional since `node_id` is always passed to it in `extend_cterm`. It is however optional for use by the custom step. Most custom steps will not make use of this. Only the `_exec_console_log_custom_step` custom step from https://github.com/runtimeverification/kontrol/pull/1131 so far would make use of this.

anvacaru (approved):

pyk/src/pyk/kcfg/semantics.py
tothtamas28: Thank you, makes sense.

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -35,7 +35,7 @@ def can_make_custom_step(self, c: CTerm) -> bool: ...
...
 
     """Implement a custom semantic step."""
```
tothtamas28: Please extend this docstring with an explanation of the parameters.

Here's a template for the syntax/convention: https://github.com/runtimeverification/k/blob/c9205d683b9c1d299c482c1506ce7dde3f4dc9c0/pyk/src/pyk/ktool/krun.py#L454-L470

tothtamas28 (approved):

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -35,9 +35,18 @@ def can_make_custom_step(self, c: CTerm) -> bool: ...
...
     @abstractmethod
-    def custom_step(self, c: CTerm, cs: CTermSymbolic) -> KCFGExtendResult | None: ...
-
-    """Implement a custom semantic step."""
+    def custom_step(self, c: CTerm, cs: CTermSymbolic, node_id: int) -> KCFGExtendResult | None: ...
+
+    """Implement a custom semantic step.
+    
+    Args:
+        c: Current configuration term representing the EVM state.
```
tothtamas28:
```suggestion
        c: Current constrained term representing the state.
```

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -35,9 +35,18 @@ def can_make_custom_step(self, c: CTerm) -> bool: ...
...
+        cs: Symbolic configuration term.
```
tothtamas28:
```suggestion
        cs: ``CTermSymbolic`` for computing the custom step result.
```

pyk/src/pyk/kcfg/semantics.py
```diff
@@ -35,9 +35,18 @@ def can_make_custom_step(self, c: CTerm) -> bool: ...
...
+        node_id: Current node id.
+    
+    Returns:
+        The KCFGExtendResult produced by this custom step if this custom step can produce one, otherwise returns None,
```
tothtamas28:
```suggestion
        The ``KCFGExtendResult`` produced by this custom step if this custom step can produce one, ``None`` otherwise. 
```

pyk/src/pyk/kcfg/semantics.py
juliankuners: Thanks! Included in latest commit.

PR #4911, opened 2026-04-20
Improvements to show, and new show from disc

mariaKt:
### Motivation

Displaying large proofs with `kmir show` caused out-of-memory failures. A completed 1411-node proof required **23.8 GB** of memory just to print its output, after the proof itself had already finished successfully.

### Investigation

The initial hypothesis was that the memory cost came from joining the entire output into a single string before printing (`'\n'.join(lines)`). Streaming the output line-by-line (via a generator) eliminated this duplication, but peak memory remained at 23.8 GB. The real cost was **loading the entire proof into memory**, parsing all node CTerms and constraint data from JSON into Python K-term objects, with ~6× memory overhead from Python object representation.

### Changes

This PR introduces two improvements:

**1. Generator-based `show_iter`**

`KCFGShow.show_iter` and `APRProofShow.show_iter` yield lines one at a time instead of collecting them into a list. The existing `show` methods are reimplemented as `list(self.show_iter(...))` to avoid code duplication. Tests confirm same output.

**2. Lazy-loading `show_iter_from_disk`**

`APRProofShow.show_iter_from_disk` displays a proof without loading the full proof into memory:

- **`LazyNode`** — duck-types `KCFG.Node`. Stores node ID, attrs, and a file path. Loads the CTerm from `nodes/{id}.json` only when `.cterm` is accessed for printing.
- **`LazyCSubst`** — duck-types `CSubst`. Holds the raw JSON dict and defers parsing into K-term objects until `.constraints` or `.subst` is accessed.
- **`APRProofStub`** — duck-types `APRProof`. Holds proof metadata (init, target, terminal, bounded, refutations) from `proof.json` and answers proof-level queries (`is_init`, `is_target`, `is_terminal`, etc.) without loading the KCFG.

These stubs are passed into the real `KCFG` data structures (edges, splits, covers, ndbranches accept them via duck typing) and the existing `pretty_segments` traversal works unchanged, with no reimplementation or duplication of the tree-drawing logic.

Loading is integrated into the existing code path: `KCFG.from_dict` accepts a `lazy` parameter that controls whether nodes are created as `LazyNode` stubs and covers use `LazyCSubst` wrappers. `KCFGStore.read_cfg_data_lazy` reads `kcfg.json` without loading node JSON files. `KCFG.read_cfg_data_lazy` provides the entry point, combining the two.

### Results

Tested on a 1411-node p-token proof (`burn_multisig_n1`):

| | `show_iter` (full load) | `show_iter_from_disk` (lazy) |
|---|---:|---:|
| Peak memory | 23.8 GB | 6.3 GB |
| Time | 4263s (~71 min) | 315s (~5 min) |
| Output lines | 10,221 | 10,221 |

**74% memory reduction, 13.5× faster, identical output.**

### Testing

- `show_iter_from_disk` output is compared against `show` output in three existing `test_imp.py` test functions (all tests that write proofs to disk). Both default and full-printer modes are exercised.
- Manually verified output match on external proofs with `APRProofNodePrinter` (proof-level attrs: init, target, terminal).
- All 98 existing pyk show-related tests pass.


pyk/src/pyk/kcfg/lazy.py
```diff
@@ -0,0 +1,151 @@
...
+            from ..cterm import CTerm
+
+            node_dict = json.loads(self._node_path.read_text())
+            self._cterm = CTerm.from_dict(node_dict['cterm'])
+        return self._cterm
+
+    def evict(self) -> None:
+        """Release the loaded CTerm from memory."""
+        self._cterm = None
+
```
ehildenb: I don't actually see where this is used, how does this work? I assume somehow this gets called whenever we're done printing a given node?

pyk/src/pyk/kcfg/lazy.py
```diff
@@ -0,0 +1,151 @@
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, LazyNode):
+            return self.id == other.id
+        # Also compare with real KCFG.Node
+        return hasattr(other, 'id') and self.id == other.id
```
ehildenb: Would this be equivalent anyway?

```suggestion
    def __eq__(self, other: object) -> bool:
        return hasattr(other, 'id') and self.id == other.id
```

Or perhaps the tighter:


```suggestion
    def __eq__(self, other: object) -> bool:
        if isinstance(other, LazyNode):
            return self.id == other.id
        # Also compare with real KCFG.Node
        return isinstance(other, KCFG.Node) and self.id == other.id
```

ehildenb:
Does it make sense to just have lazy-loading be the default anyway, and build it into the main KCFG class?

ehildenb:
@claude can you review this PR? And check if `evict` is actually used, and whether it makes sense to just have lazy loading be the default for KCFGs?

pyk/src/pyk/kcfg/lazy.py
mariaKt: I like the second one. I think it captures the intention (only LazyNode or Node) better.

mariaKt:
> Does it make sense to just have lazy-loading be the default anyway, and build it into the main KCFG class?

@ehildenb  I think that there is a great reason to build it into the main KCFG class. One of the large proofs I was running failed while it was running (not after completion, at the printing stage). While writing this, I found that this is due to the nodes are never released from memory while a proof is running (even though the nodes' json files are created while the proof runs).  Having a LazyNode would help in that, by enabling evicting nodes' CTerms and reloading them if ever needed.

I did not do this, as it was a much larger change, affecting not only the printing. I am happy to adapt this PR to do this.

pyk/src/pyk/kcfg/lazy.py
mariaKt: I intended to use it in pretty_segments, but missed it. I will address.

ehildenb:
I think it's a good idea overall, but we should probably check in with @tothtamas28 about it. I would prefer not to have two different classes with one duck-typing the other, if we can get the benefit by just making lazy loading the default, that would be great.

Tamas should be back next week to provide his opinion.

mariaKt:
I should be able to just edit the Node class, no additional class.

ehildenb:
If we have the laziness built-in by default, do we need the `APRProofStub` classes for example? Or do we get the benefit even with the main classes?

ehildenb:
I'm also wondering now if this is a good idea. We have to store the `_node_path` directly on the `Node` object, which is not how the data-model was done before. Before, we had clear separation between the storage on disk and the data representation itself. I think it should be fine, but for example is the `KCFGStore` class correctly handling these `_node_path` now too?

Currently, a `KCFG` already has a `KCFGStore`, so it already has the paths needed for loading/unloading from disk, can we reuse that infrastructure somehow when doing the lazy node loading?

mariaKt:
> If we have the laziness built-in by default, do we need the `APRProofStub` classes for example? Or do we get the benefit even with the main classes?

I think no, we would not need the stubs then. I removed one so far, for some of the current changes.

mariaKt:
> I'm also wondering now if this is a good idea. We have to store the `_node_path` directly on the `Node` object, which is not how the data-model was done before. Before, we had clear separation between the storage on disk and the data representation itself. I think it should be fine, but for example is the `KCFGStore` class correctly handling these `_node_path` now too?
> 
> Currently, a `KCFG` already has a `KCFGStore`, so it already has the paths needed for loading/unloading from disk, can we reuse that infrastructure somehow when doing the lazy node loading?

Right, I did need to add `node_path` to `Node`. I edited the `KCFGStore` (read_cfg_data and write_cfg_data) to propagate these paths, but I don't think this is what you might mean by "correctly handling these `_node_path`". I can take a closer look at the KCFGStore and see how to reuse that information instead, thanks for pointing it out.

mariaKt:
@ehildenb The issue is that, with the lazy approach, in order to find a node's `CTerm` I need to    
potentially read the node's json file. If not an explicit path, I need to be able to refer to the store, or the KCFG,   
so I definitely need some kind of access in that direction. claude also suggested a callback, which I hadn't considered. Is there an option from these, or another that you prefer?

tothtamas28:
> I think it's a good idea overall, but we should probably check in with @tothtamas28 about it. I would prefer not to have two different classes with one duck-typing the other, if we can get the benefit by just making lazy loading the default, that would be great.

The main question here I think is how to roll out this change across the stack. One option is to turn `KCFG` and its components into abstract classes, with separate implementations for lazy and eager loading. This would avoid duck-typing and prevent mixing node types within the same tree.

Downstream projects could then adopt the lazy implementation at their own pace and surface any issues. Once that transition is complete, we can remove the eager version.

The abstract layer would still be valuable, for example, it would make it easier to experiment with alternative storage backends (e.g., a database), I think there is room for optimization there.

mariaKt:
Thank you @tothtamas28 . I was wondering if you had any input on the following issue as well:
While working on the implementation of the Node with lazy loading as the default (https://github.com/runtimeverification/k/pull/4911/commits/44827c3d746f6d464a12c18612c4accb38313881) (of course that will change now, just for context), I realized that in order to be able to load a node's CTerm on demand (as needed in order to print or resume a proof and not load the entire proof), we need to be able to access the store within the Node (https://github.com/runtimeverification/k/pull/4911#issuecomment-4300876795). Do you have any insight or preference on how to deal with this?

tothtamas28:
> I realized that in order to be able to load a node's CTerm on demand (as needed in order to print or resume a proof and not load the entire proof), we need to be able to access the store within the Node (...). Do you have any insight or preference on how to deal with this?

My first thought would be to just store a reference to `KCFGStore` within the lazy node. Would that make sense?

Here are two design alternatives based on this idea.

## `LazyKCFG` and `EagerKCFG` as separate classes

```mermaid
classDiagram
    class KCFG <<abstract>>{
        Node node(int)
    }
    class Node <<abstract>>

    class LazyKCFG

    class LazyNode{
        KCFGStore store
        Optional[CTerm] term
    }

    class EagerKCFG{
        KCFGStore store
    }

    class EagerNode{
        CTerm term
    }

    KCFG --> Node
    KCFG <|-- LazyKCFG
    KCFG <|-- EagerKCFG
    Node <|-- LazyNode
    Node <|-- EagerNode
    LazyKCFG ..> LazyNode
    EagerKCFG ..> EagerNode
```

## A single `KCFG` for both strategies

While drawing the first alternative, I realized the two parallel class hierarchies can be collapsed into one. This one is probably closer to your current implementation. (But even in this design, I think extracting the abstract base classes would worth it in order to be able to freely experiment with alternative implementations.)

```mermaid
classDiagram
    class KCFG{
        KCFGStore store
        Node node(int)
    }

    class Node{
        Optional[KCFGStore] store
        Optional[CTerm] term
    }

    KCFG --> Node
```

The class invariant on `Node` is that either `store` or `term` is not `None`.

## Notes

A few thoughts regarding `KCFGStore`:
* Is there a downstream project that does not use the `KCFGStore` as a storage backend? If not, we should just commit to it and simplify the code as step nr. 1.
* Later we can even extract an interface from `KCFGStore` and figure out a way to abstract away the storage backend, decoupling storage from loading strategy.

PR #4914, opened 2026-04-27
Update regression new

ehildenb:
To start from a clean baseline while reimplementing the kompile pipeline in pyk, this updates the test expected output and uncomments some tests that are passing with the updated output. The main changes in the pyk printer from the Java one are:

- K-sequence `~>` is displayed with newlines.
- Empty K-sequence is not printed at the end of K-sequence on the `<k>` cell.
- List/Map/Set/RangeMap append are not displayed with newlines.
- The generated cells (`<generatedCounter>` and `<generatedTop>`) are not omitted.

The documentation file is also updated (by Claude) to reflect the current triage status of the regression-new test-suite (at `pyk/docs/regression-triage.md`).


pyk/regression-new/or-llvm/5.test.out
```diff
@@ -1,8 +1,12 @@
 <generatedTop>
   <k>
-    2 ~> test ( bar ~> .K ) ~> .K
+    2
+    ~> test ( bar ~> .K )
```
ehildenb, line 4: This change is a bit weird, but I would let it be. Looks like in our printer we are special-casing k-sequences directly below the `<k>` cell.

ehildenb:
MacOS tests are failing because of runner issues, will get them sorted.

pyk/regression-new/rangemap-tests-llvm/concat-4.test.out
```diff
@@ -1,4 +1,8 @@
-<k>
-  [ 1 , 2 ) r|-> 45
-  [ 2 , 5 ) r|-> 40 ~> .K
-</k>
+<generatedTop>
+  <k>
+    [ 1 , 2 ) r|-> 45 [ 2 , 5 ) r|-> 40 ~> .K
```
tothtamas28, line 3: > Empty K-sequence is not printed at the end of K-sequence on the `<k>` cell.

This doesn't seem to be enforced consistently.

Can we change the `pyk` printer to just print the `.K` in all cases? That would make the output more predictable.

ehildenb: Done.

tothtamas28 (approved):

PR #4915, opened 2026-04-27
pyk: context-aware sort-parameter inference in add_sort_params (port of Java AddSortInjections)

ehildenb:
The K frontend often emits parametric `KLabel`s with their sort parameters left off. This reworks pyk's `add_sort_params` into a context-aware algorithm mirroring the Java frontend's `AddSortInjections`: a single recursive sweep threads an expected sort downwards (so a parametric *return* sort is taken from its context) and resolves the remaining parameters from the argument sorts coming back up.

**Changes:**
- `add_sort_params` now takes an optional expected `sort` and resolves each parameter either bottom-up (argument-bound) or top-down (return-only); a parameter left genuinely free becomes a unique fresh sort variable (`#SortParam{Q0}`, `{Q1}`, …), with one such variable shared down a constraint spine. It asserts that a parametric return sort only stays free on the spine.
- `KDefinition.least_upper_bound`: a set-based, order-independent least upper bound over the subsort lattice (resolves siblings that share a common supersort).
- `KSort.contains`: a structural "does this sort mention that sort" query (mirrors Java `Sort.contains`), replacing a module-level helper.
- `sort` and `resolve_sorts` handle `KAs` and unfilled `KApply`, with recursive substitution into nested sort params.
- The `#SortParam` sentinel is not yet emittable to Kore: `_ksort_to_kore` rejects it with a clear error rather than producing invalid Kore, and `kast_to_kore` is left calling `add_sort_params` without an expected sort for now (opt-in), so pipeline emission is unchanged. Full design in `pyk/docs/2026-06-01-sortparam-kore-emission.md`.
- Date-prefix the existing `pyk/docs` markdown filenames for chronological sorting.

**Validation:**
- Unit tests for `add_sort_params` (bottom-up, top-down expected-sort resolution, spine sharing, the spine-violation assertion, conflict and unsortable-argument handling), plus `least_upper_bound`, `KSort.contains`, and `_ksort_to_kore` sentinel rejection. Full pyk unit suite green; `make -C pyk check` clean. 
- Not yet run locally: the konvert integration and `regression-new` suites (the machine's K install is pinned for a benchmark) — relying on CI.


pyk/src/pyk/kast/outer.py
```diff
@@ -1483,28 +1495,86 @@ def transform(
     # Best-effort addition of sort parameters to klabels, context insensitive
     def add_sort_params(self, kast: KInner) -> KInner:
         """Return a given term with the sort parameters on the `KLabel` filled in (which may be missing because of how the frontend works), best effort."""
+        # ML predicate labels whose result sort (Sort2) is context-dependent and not inferable
+        # from the arguments alone.  When Sort1 can be determined but Sort2 cannot, we fill Sort2
+        # with the sentinel KSort('#SortParam') so that downstream Kore emission can introduce a
+        # universally-quantified sort variable (Q0) in the axiom.
+        _ML_PRED_RESULT_SORT_PARAM = KSort('#SortParam')  # noqa: N806
+        _ML_PRED_LABELS = frozenset({'#Equals', '#Ceil', '#Floor', '#In'})  # noqa: N806
+
```
ehildenb: Unclear how this algorithm ends up handling cases with multiple sort parameters, since it uses just one sentinel value.

pyk/src/pyk/kast/outer.py
```diff
@@ -1000,6 +1001,59 @@ def let(self, *, require: str | None = None) -> KRequire:
...
+            for k, vs in _match_sort_params(p_sub, a_sub, params, subsorts_fn).items():
+                result.setdefault(k, []).extend(vs)
+        return result
+    if parametric.params and subsorts_fn is not None:
+        result = {}
+        for s in subsorts_fn(actual):
+            if s.name == parametric.name and len(s.params) == len(parametric.params):
+                for p_sub, a_sub in zip(parametric.params, s.params, strict=True):
+                    for k, vs in _match_sort_params(p_sub, a_sub, params).items():
+                        result.setdefault(k, []).extend(vs)
```
ehildenb, line 1033: Should we `break` if the `if s.name == ...` condition is true? We are looking for the first subsort that matches the given sort name right?

Also, can we factor out the `result = {} ... for p_sub, a_sub ... for k, vs ... result.setdefault(...)` logic into a helper inner function defined here? Something that computes piecewise the sort parameter bindings on arguments? Then we can reuse it here and the clause above this one.

pyk/src/pyk/kast/outer.py
```diff
@@ -1000,6 +1001,59 @@ def let(self, *, require: str | None = None) -> KRequire:
...
+    A param in ``params`` is considered unbound if it does not appear in ``bindings``.
+    """
+    if sort in params:
+        return bindings.get(sort)
+    if sort.params:
+        resolved = [_resolve_sort_partial(p, bindings, params) for p in sort.params]
+        if any(r is None for r in resolved):
+            return None
+        return KSort(sort.name, tuple(resolved))  # type: ignore[arg-type]
+    return sort
```
ehildenb: This is only used in one place. Is it reasonable to inline it to its use site?

PR #4494, opened 2024-07-02
Use portable sed invocation

Baltoli:
The `\|` operator is not portable to BSD / macOS sed, which makes some tests appear flaky when run on those platforms. This PR switches the offending command over to a wordier, but correct invocation.

PR #4917, opened 2026-05-07
Update regression-new test harness for pyk, enable checks directory again

ehildenb:
This PR brings in refactors from the last 2 years ago for the Java regression-new test-suite and updates the output in the `pyk/regression-new` test-suite for the `checks` subdirectory. All of the changes are expected (also made to the Java test-suite) or just changes to flag order for the `kompile` command that is invoked.

Around two years ago, there were several changes/refactors to the Java regression-new test-suite harness, which are copied over (and adapted for pyk) here:

- Removing the strategy module and tests: https://github.com/runtimeverification/k/pull/4115
- Factor out `ktest-common.mak` and better filtering of absolute paths in test output: https://github.com/runtimeverification/k/pull/4334
- Update the path filtering sed expression: https://github.com/runtimeverification/k/pull/4494

In addition, this PR brings in several updates that have been applied to the Java regression-new test-suite but not ported over to the pyk one, including changes to tests and entirely new tests added.

The `pyk/regression-new` suite is shifted onto our self-hosted runners, as the public ones are too slow and this is on the longest path for CI. The timeout is lowered to 20 minutes (despite running many more tests in the `checks` directory).


pyk/regression-new/checks/overloadKLabel.k.out
```diff
@@ -1,8 +1,8 @@
-[Error] Compiler: The attributes `klabel(_)` and `overload(_)` may not occur together.
+[Error] Compiler: The attributes `klabel` and `overload` may not occur together. Either remove `klabel(_._)`, or replace it by `symbol(_._)`
```
ehildenb: Updated output from Java: https://github.com/runtimeverification/k/blob/master/k-distribution/tests/regression-new/checks/overloadKLabel.k.out

gtrepta (approved):
Seems fine to me.

IMO the ideal solution will be to decouple the original regression tests from the Java code and make them flexible enough so that both the Java code and pyk can refer to them instead of maintaining their own mirrors.

ehildenb:
I agree that decoupling the tests would be best, I just figured that getting them to match each other and getting more of them passing with pyk was a good step in that direction. If we can get them all passing with pyk, we can probably even drop the Java ones actually.

PR #4918, opened 2026-05-11
pyk: fix add_cell_map_items over-wrapping, add --parser to pyk run

ehildenb:
`add_cell_map_items` was wrapping element-sort constructors in their cell-map item wrapper unconditionally. This caused incorrect KAST for productions like `getEntry(AccountCell)` where the argument expects the element sort `AccountCell`, not the collection sort `AccountCellMap`. The fix tracks the expected sort for each argument position and only wraps when the parent expects the map sort.

While investigating which regression tests this unblocks, pyk run was also missing `--parser`, which is used by several test cases to supply a custom input parser. This is now wired through from the CLI down to `_krun`.

Changes:
- `KDefinition.cell_map_item_info`: added cached property to `KDefinition` for extracting not only the element constructor for a given cell collection production, but also the sort of that constructor, and uses this factored out method in both of `KDefinition.{add_cell_map_items,remove_cell_map_items}`.
- `KDefinition.add_cell_map_items`: guard wrapping behind a check that the parent argument sort matches the cell map sort
- pyk run / `KRun.krun`: add `--parser` flag support
- Remove star-multiplicity, issue-582, krun-deserialize from regression-new/skipped; update their baselines to current compiler output

Validation:
- New unit tests in test_definition.py cover both the wrapping case (`_AccountCellMap_`) and the no-wrap case (`getEntry`)
- All three re-enabled regression tests pass end-to-end against the LLVM backend

### DIFF
--- a/pyk/pyproject.toml
+++ b/pyk/pyproject.toml
@@ -79,7 +79,7 @@ include = ["src/pyk"]
 [tool.pydocstyle]
 convention = "google"
 add-select = "D204,D401,D404"
-add-ignore = "D1"
+add-ignore = "D1,D202"
 
 [tool.isort]
 profile = "black"
--- a/pyk/pyproject.toml
+++ b/pyk/pyproject.toml
@@ -79,7 +79,7 @@ include = ["src/pyk"]
 [tool.pydocstyle]
 convention = "google"
 add-select = "D204,D401,D404"
-add-ignore = "D1,D202"
+add-ignore = "D1"
 
 [tool.isort]
 profile = "black"
--- a/pyk/src/pyk/kast/outer.py
+++ b/pyk/src/pyk/kast/outer.py
@@ -1139,12 +1139,12 @@ class KDefinition(KOuter, WithKAtt, Iterable[KFlatModule]):
     def cell_map_item_info(self) -> dict[str, tuple[str, KSort]]:
         """Map from cell label to (element-constructor label, cell-map sort).
 
-        Derived from cell-collection productions that carry both ELEMENT and WRAP_ELEMENT
-        attributes and whose element constructor is a known symbol.
+        Derived from cell-collection productions that carry both ``ELEMENT`` and ``WRAP_ELEMENT`` attributes and whose element constructor is a known symbol.
+
+        Example:
+            syntax AccountCellMap [cellCollection, hook(MAP.Map)]
+            syntax AccountCellMap ::= AccountCellMap AccountCellMap [assoc, avoid, cellCollection, comm, element(AccountCellMapItem), function, hook(MAP.concat), unit(.AccountCellMap), wrapElement(<account>)]
         """
-        # example:
-        # syntax AccountCellMap [cellCollection, hook(MAP.Map)]
-        # syntax AccountCellMap ::= AccountCellMap AccountCellMap [assoc, avoid, cellCollection, comm, element(AccountCellMapItem), function, hook(MAP.concat), unit(.AccountCellMap), wrapElement(<account>)]
         result: dict[str, tuple[str, KSort]] = {}
         for ccp in self.cell_collection_productions:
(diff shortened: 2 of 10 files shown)


pyk/src/pyk/kast/outer.py
```diff
@@ -1135,6 +1135,25 @@ def cell_collection_productions(self) -> tuple[KProduction, ...]:
...
+    @cached_property
+    def cell_map_item_info(self) -> dict[str, tuple[str, KSort]]:
+        """Map from cell label to (element-constructor label, cell-map sort).
+
+        Derived from cell-collection productions that carry both ELEMENT and WRAP_ELEMENT
+        attributes and whose element constructor is a known symbol.
+        """
+        # example:
+        # syntax AccountCellMap [cellCollection, hook(MAP.Map)]
+        # syntax AccountCellMap ::= AccountCellMap AccountCellMap [assoc, avoid, cellCollection, comm, element(AccountCellMapItem), function, hook(MAP.concat), unit(.AccountCellMap), wrapElement(<account>)]
```
tothtamas28:
```suggestion
        """Map from cell label to (element-constructor label, cell-map sort).

        Derived from cell-collection productions that carry both ``ELEMENT`` and ``WRAP_ELEMENT``
        attributes and whose element constructor is a known symbol.

        Example:

            syntax AccountCellMap [cellCollection, hook(MAP.Map)]
            syntax AccountCellMap ::= AccountCellMap AccountCellMap
                [assoc, avoid, cellCollection, comm, element(AccountCellMapItem),
                 function, hook(MAP.concat), unit(.AccountCellMap), wrapElement(<account>)]
        """
```

pyk/src/pyk/kast/outer.py
```diff
@@ -1511,36 +1530,37 @@ def _add_sort_params(_k: KInner) -> KInner:
...
-
-        cell_wrappers = {}
-        for ccp in self.cell_collection_productions:
-            if Atts.ELEMENT in ccp.att and Atts.WRAP_ELEMENT in ccp.att:
-                cell_wrappers[ccp.att[Atts.WRAP_ELEMENT]] = ccp.att[Atts.ELEMENT]
 
+        # Wrapping is correct only when the parent production expects the cell MAP sort (e.g.
+        # EntryCellMap), not when it expects the individual cell element sort (e.g. EntryCell).
+        # For example, EntryCellMapKey(<entry>(...)) takes EntryCell — the <entry> must NOT be
+        # wrapped, whereas _EntryCellMap_(<entry>(...), ...) expects EntryCellMap — wrapping is needed.
```
tothtamas28:
```suggestion
        """Wrap cell-map items in the syntactical wrapper that the frontend generates for them.

        See :func:`KDefinition.remove_cell_map_items`.

        Note:
            Wrapping is correct only when the parent production expects the cell MAP sort (e.g. ``EntryCellMap``),
            not when it expects the individual cell element sort (e.g. ``EntryCell``).

        Example:
            ``EntryCellMapKey(<entry>(...))``: takes ``EntryCell``, ``<entry>`` must NOT be wrapped
            ``_EntryCellMap_(<entry>(...), ...)``: expects ``EntryCellMap``, wrapping is needed
"""
```

pyk/pyproject.toml
```diff
@@ -79,7 +79,7 @@ include = ["src/pyk"]
 [tool.pydocstyle]
 convention = "google"
 add-select = "D204,D401,D404"
-add-ignore = "D1"
+add-ignore = "D1,D202"
```
tothtamas28:
```suggestion
add-ignore = "D1"
```

tothtamas28 (approved):

pyk/src/pyk/kast/outer.py
ehildenb: Done.

pyk/pyproject.toml
ehildenb: Done.

PR #4922, opened 2026-05-29
pyk: small fixes and additions across kast, ktool, and kore-rpc

ehildenb:
Bundles a set of small, independent improvements to pyk: a few bug fixes in the kast pipeline, a parser-side parity step for sort-decl arguments, opt-in safety flags on rule/claim construction, and a `ContextVar` that lets every outgoing kore-RPC request self-identify its originating claim.

**Changes:**
- **Bug fixes**
  - `kast/manip`: `on_attributes` recurses into `KDefinition` module sentences.
  - `kast/markdown`: `select_code_blocks` preserves source positions.
  - `kast/outer_parser`: bubble source + location attached only when the source path is known.
  - `ktool/utils`: `_which_kompile` no longer raises when `kompile` is absent.
- **Java-parity prep**
  - `kast/outer_parser`: accept `NAT` in sort-decl arguments.
- **Opt-in flags on rule / claim building** (defaults preserve current behaviour)
  - `kast/manip`: `keep_unsafe` and `filter_useless_constraints` on `build_rule` / `build_claim` (and upstream `ml_pred_to_bool` / `normalize_ml_pred`).
  - `cterm/cterm.py`: `cterm_build_rule` / `cterm_build_claim` forward the new flags.
- **Per-claim RPC attribution**
  - `kore/rpc`: module-level `client_label: ContextVar[str | None]` prefixes every outgoing JSON-RPC request id with the originating claim.
  - `proof/{reachability,implies}`: `Prover.init_proof` writes `client_label.set(proof.id)`.
  - Threading caveat: `ContextVar` is not propagated by `ThreadPoolExecutor`; multi-threaded consumers must `client_label.set(...)` inside their worker closure.

**Validation:**
- `make -C pyk check` clean (black, isort, autoflake, flake8, mypy, pydocstyle).
 - pyk unit + integration suites pass; new unit tests cover the kast/manip flags, the NAT sort-decl parse path, and the kore-rpc `ContextVar` semantics (default, set, persistence, token reset).


pyk/src/pyk/kast/manip.py
```diff
@@ -165,6 +165,8 @@ def _ml_constraint_to_bool(_kast: KInner) -> KInner:
                     exists_var = abstract_term_safely(_kast, base_name='Exists')
                     _LOGGER.warning(f'Converting #Exists condition to variable {exists_var.name}: {_kast}')
                     return exists_var
+            if keep_unsafe:
+                return _kast
```
gtrepta, line 172: The names of the parameters here are confusing. If you want `keep_unsafe` behavior, then you need `!unsafe`.

It looks like `unsafe` is only used [here](https://github.com/runtimeverification/k/blob/ef0d4c3f3b5a5873c582fb8c808b908ce99850c2/pyk/src/pyk/kcfg/show.py#L140), for printing...? Maybe it can be renamed.

gtrepta (approved):

pyk/src/pyk/kast/manip.py
ehildenb: I renamed `unsafe` to `abstract_unsafe`, so it's more descriptive of what it actually does, and I made a check that not both are set at the same time at the beginning of th efunction.

PR #4927, opened 2026-06-03
Optimize Nix release workflow

ehildenb:
The k-framework-binary (private, via kup) and k-framework (public, via cachix) publishes built the same three derivations on separate runners, recompiling K four times (2 jobs x 2 OS) for work that needs two builds. Fold them into one matrix job: the first publish populates the local Nix store so the second's nix build is an instant store hit. The two publishes stay independent (continue-on-error + if: always()) so a flaky upload to one cache neither blocks nor masks the other, and a final guard re-fails the job if either push failed, preserving the release gate the two separate jobs provided. Drop the removed cachix-release-dependencies from the release job's needs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ jobs:
           gh release upload --repo runtimeverification/k --clobber "v${version}" "${tarball}"
 
   cachix-release:
-    name: 'Cachix release (binary cache + dependency closure)'
+    name: 'Cachix release (binary cache `k-framework-binary` + dependency cache `k-framework`)'
     strategy:
       matrix:
         include:
@@ -79,17 +79,11 @@ jobs:
         with:
           name: k-framework-binary
 
-      # The binary publish and the dependency-closure publish build the same three
-      # derivations.  Running them as sequential steps on one runner means the build
-      # happens once: the first step populates the local Nix store, so the second
-      # step's `nix build` is an instant store hit rather than a full recompile.
-      #
-      # The two publishes target different caches with different tokens (private
-      # `k-framework-binary` via kup vs. public `k-framework` via cachix), so they are
-      # kept independent: `continue-on-error` plus `if: always()` ensures a flaky
-      # upload to one cache neither blocks nor masks the other, and the final guard
-      # re-fails the job if either push failed — preserving the release gate that the
-      # two separate jobs used to provide.
+      # Both publishes build the same three derivations, so they run as sequential
+      # steps on one runner: the first populates the local Nix store and the second is
+      # a store hit. They run independently (continue-on-error) so a flaky upload to
+      # one cache neither blocks nor masks the other, and a final guard fails the job
+      # if either push failed.
       - name: 'Publish K to k-framework-binary cache (private)'
         id: binary
         continue-on-error: true
@@ -116,7 +110,6 @@ jobs:
 
       - name: 'Build and cache K and K dependencies (public)'
         id: dependencies
-        if: always()
         continue-on-error: true
         uses: workflow/nix-shell-action@v3
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -101,38 +114,10 @@ jobs:
...
-      - name: 'Install Cachix'
-        if: ${{ !startsWith(matrix.os, 'self') }}
-        uses: cachix/cachix-action@v14
-        with:
-          name: k-framework-binary
-
-      - name: 'Build and cache K and K dependencies'
+      - name: 'Build and cache K and K dependencies (public)'
+        id: dependencies
+        if: always()
```
juliankuners: `if: always()` is unnecessary, since the previous step already declared `continue-on-error`

.github/workflows/release.yml
```diff
@@ -152,6 +137,15 @@ jobs:
             DRV_K_OPENSSL_PROCPS_SECP256K1=$(nix-store --query --deriver ${K_OPENSSL_PROCPS_SECP256K1})
             nix-store --query --requisites --include-outputs ${DRV_K_OPENSSL_PROCPS_SECP256K1} | cachix push k-framework
 
+      - name: 'Fail if either cachix publish failed'
+        if: always()
```
juliankuners: Same unnecessary `if: always()`, see above comment.

.github/workflows/release.yml
```diff
@@ -79,7 +79,20 @@ jobs:
...
+      #
+      # The two publishes target different caches with different tokens (private
+      # `k-framework-binary` via kup vs. public `k-framework` via cachix), so they are
+      # kept independent: `continue-on-error` plus `if: always()` ensures a flaky
+      # upload to one cache neither blocks nor masks the other, and the final guard
+      # re-fails the job if either push failed — preserving the release gate that the
+      # two separate jobs used to provide.
+      - name: 'Publish K to k-framework-binary cache (private)'
+        id: binary
+        continue-on-error: true
```
juliankuners: Bloated AI comment that reasons why the old pattern of two workflows was merged into one. Once we've adopted this pattern everywhere, this comment will be confusing since it's not clear that the workflows was split in two in the first place.

If you wish to add a clarifying comment, keep it short and concise, simply just describing what the workflow does.

Though I agree that the differentian between the caches `k-framework` and `k-framework-binary` was never properly documented. Though I doubt that the proper place is a long comment in a workflow in one of our dependency projects. Rather, there should be proper documentation in possibly https://github.com/runtimeverification/rv-nix-tools/tree/master/docs.

.github/workflows/release.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           gh release upload --repo runtimeverification/k --clobber "v${version}" "${tarball}"
 
   cachix-release:
-    name: 'k-framework-binary cachix release'
+    name: 'Cachix release (binary cache + dependency closure)'
```
juliankuners:
```suggestion
    name: 'Cachix release (binary cache `k-framework-binary` + dependency cache `k-framework`)'
```

juliankuners (commented):
See comments and suggestions.

juliankuners (approved):

.github/workflows/release.yml
ehildenb: Done.

.github/workflows/release.yml
ehildenb: Done.

.github/workflows/release.yml
ehildenb: Done.

.github/workflows/release.yml
ehildenb: Done.

palinatolmach (approved):

PR #4928, opened 2026-06-03
pyk: per-request haskell-backend RPC logging

ehildenb:
~Blocked on: https://github.com/runtimeverification/haskell-backend/pull/4151~
~Blocked on: https://github.com/runtimeverification/k/pull/4929~

Lets pyk request the haskell-backend's per-request log bundle and write it to disk, so downstream can diagnose fallback behaviour per RPC. The request carries the list of log entries to capture and the response returns them in-band; pyk owns the default set and writes each request's bundle to its own file. This pairs with the haskell-backend per-request-rpc branch, which adds the haskell-logging request field ([String]) and the haskell-log-entries response field to the five JSON-RPC endpoints and captures the named entries independently of the server's -l/--log-file configuration.

Changes:
- BoosterServer now emits the booster-form haskell-log CLI (--log-file FILE, repeated -l ENTRY) via a new overridable _haskell_log_cli_args hook; it previously sent the kore-rpc form (--log, --log-entries A,B), which made kore-rpc-booster exit on a usage error before opening its RPC port.
- KoreClient.{execute,implies,simplify,get_model,add_module} take haskell_logging: Iterable[str] | None — the entries to capture for that request — and send it as the haskell-logging wire field (omitted when None); the optional haskell-log-entries bundle is parsed onto every result type, and simplify returns a SimplifyResult (state, logs, haskell_log_entries).
- KoreClient.last_request_id exposes the JSON-RPC id of the most recent request issued by the client, used to name each per-request log file.
- CTermSymbolic owns the canonical default entry set (HASKELL_LOGGING_ENTRIES), overridable per instance (and through the cterm_symbolic factory) so downstream can tailor it; the on/off flag is threaded through execute/simplify/kast_simplify/implies and KCFGExplore.extend_cterm.
- CTermSymbolic takes a haskell_log_dir: when set, every RPC requests the bundle and the entries returned on the response are written to <haskell_log_dir>/<request_id>.jsonl (one JSON value per line, named by request id so concurrent proof workers never collide).
- pyk prove --haskell-logging points that sink at <save-directory>/haskell-logs/ (requires --save-directory).

Validation:
- make -C pyk check clean.
- Confirmed against the per-request-rpc backend implementation: it accepts haskell-logging: [String], routes the flat list to both engines (skipping names each does not recognise), tag-matches id-carrying contexts like Rewrite, and returns the captured entries in-band — pyk's default set is fully recognised, no skips.
- Integration and regression-new not run locally (the machine's K install is pinned for a benchmark) — relying on CI.


juliankuners (dismissed):

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -40,6 +41,25 @@
 _LOGGER: Final = logging.getLogger(__name__)
 
 
+#: Canonical default set of haskell-backend log entries to capture per request for fallback
+#: diagnosis.  Spans both engines: kore entry types (resolved by the backend against its log
+#: registry) and booster context tags.  Sent verbatim on the ``haskell-logging`` request field;
+#: the backend skips any name it does not recognise, so this can evolve without a lockstep backend
+#: release.  Override per `CTermSymbolic` (e.g. downstream semantics needing a different set).
+HASKELL_LOGGING_ENTRIES: Final[tuple[str, ...]] = (
```
tothtamas28:
```suggestion
HASKELL_LOGGING_ENTRIES: Final = (
```

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -40,6 +41,25 @@
 _LOGGER: Final = logging.getLogger(__name__)
 
 
+#: Canonical default set of haskell-backend log entries to capture per request for fallback
+#: diagnosis.  Spans both engines: kore entry types (resolved by the backend against its log
+#: registry) and booster context tags.  Sent verbatim on the ``haskell-logging`` request field;
+#: the backend skips any name it does not recognise, so this can evolve without a lockstep backend
+#: release.  Override per `CTermSymbolic` (e.g. downstream semantics needing a different set).
```
tothtamas28:
```suggestion
#: release.  Override per ``CTermSymbolic`` (e.g. downstream semantics needing a different set).
```

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -83,19 +105,53 @@ def __init__(
         log_succ_rewrites: bool = True,
         log_fail_rewrites: bool = False,
         booster_only_simplify: bool = False,
+        haskell_log_entries: Iterable[str] = HASKELL_LOGGING_ENTRIES,
```
tothtamas28, line 108: So logging is on by default?

pyk/src/pyk/cterm/symbolic.py
```diff
@@ -151,29 +210,42 @@ def execute(
         )
 
     def simplify(
-        self, cterm: CTerm, module_name: str | None = None, booster_only_simplify: bool | None = None
+        self,
+        cterm: CTerm,
+        module_name: str | None = None,
+        booster_only_simplify: bool | None = None,
+        haskell_logging: bool | None = None,
```
tothtamas28, line 218: So the class owns the list of log entry types to ask for, and requests just turn this on or off, right?

pyk/src/pyk/kore/rpc.py
```diff
@@ -48,6 +48,13 @@
 # `ContextVar` values are not propagated by `ThreadPoolExecutor`.
 client_label: ContextVar[str | None] = ContextVar('kore_rpc_client_label', default=None)
 
+# Records the JSON-RPC request id of the most recent wire request issued on the
+# calling thread (see `JsonRpcClient.request`).  A diagnostic caller snapshots
+# `KoreClient.last_request_id` immediately after each RPC to use the deterministic
+# `f'{client_label}-{NNN}'` id as a join key onto the captured per-request logs.
+# Thread-local so parallel proof workers never clobber each other's ids.
+_last_request_id: local = local()
+
```
tothtamas28: What's the rationale for tracking this per-thread and not per-client?

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # When the per-call flag turns logging on
+    cts.execute(dummy, haskell_logging=True)
+
+    # Then the configured entry set (the canonical default here) is what reaches KoreClient.execute
+    _args, kwargs = kore_client.execute.call_args
+    assert kwargs['haskell_logging'] == HASKELL_LOGGING_ENTRIES
+
+
+def test_execute_honours_custom_entry_set() -> None:
```
tothtamas28:
```suggestion
def test_execute_honors_custom_entry_set() -> None:
```

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,108 @@
...
+    from pathlib import Path
+
+    from pyk.cterm import CTerm
+
+
+def _mock_client_cts(response: object) -> tuple[Mock, CTermSymbolic]:
+    """Build a `CTermSymbolic` whose backend returns ``response`` and whose Kast<->Kore
+    conversions are stubbed, returning the backend `Mock` so a test can inspect call args.
+    """
+    kore_client = Mock()
```
tothtamas28: There's already a mock `KoreClient` in the unit test suite exposed with a fixture, maybe that can be reused here?

https://github.com/runtimeverification/k/blob/5d786ca3578881344001874d04bd4957756a32b0/pyk/src/tests/unit/kore/test_client.py#L54-L67

pyk/src/tests/unit/kore/test_client.py
```diff
@@ -228,13 +228,86 @@ def test_simplify(
...
     assert actual == expected
 
 
+_HASKELL_LOG_ENTRIES: Final = [
+    {'context': ['proxy', 'detail'], 'message': 'kore-simplify'},
+    {'label': 'EVM.foo', 'rule_id': 'abc', 'pre_hash': 'deadbeef', 'post_hash': 'f00d'},
+]
+
+
+def test_execute_parses_haskell_log_entries(kore_client: KoreClient, rpc_client: MockClient) -> None:
```
tothtamas28: It looks like all these tests follow the same template, consider extracting a parameterized test, or integrating them into the existing test functions even.

pyk/src/tests/unit/kore/test_rpc_client.py
```diff
@@ -78,3 +80,39 @@ def test_exceptions(
...
+    token = client_label.set('claim-x')
+    try:
+        transport.assume_response(json.dumps({'jsonrpc': '2.0', 'id': 'claim-x-001', 'result': {'k': 'v'}}))
+
+        # When (drive the underlying JsonRpcClient directly, bypassing Kore result parsing)
+        out = kore_client._client._default_client.request('simplify', state={'x': 1})
+
+        # Then
+        assert out == {'k': 'v'}
+        assert kore_client.last_request_id == 'claim-x-001'
```
tothtamas28: This test looks more like a `KoreClient` test, consider moving it there. Dependency on `MockTransport` can probably be eliminated.

pyk/src/tests/unit/kore/test_rpc_client.py
```diff
@@ -78,3 +80,39 @@ def test_exceptions(
...
+
+def test_last_request_id_is_thread_local(kore_client: KoreClient) -> None:
+    # Each thread must read back its own last-request id, never another thread's.
+    seen: dict[str, str | None] = {}
+    barrier = Barrier(2)
+
+    def worker(name: str) -> None:
+        _last_request_id.value = name
+        barrier.wait()  # both threads have set their value before either reads
+        seen[name] = kore_client.last_request_id
```
tothtamas28: This is also looks like a `KoreClient` test.

pyk/src/tests/unit/kore/test_server_cli_args.py
```diff
@@ -0,0 +1,90 @@
...
+    haskell_log_entries: list[str] | None = None,
+) -> KoreServer:
+    server = object.__new__(KoreServer)
+    server._log_axioms_file = log_axioms_file  # type: ignore[attr-defined]
+    server._haskell_log_format = haskell_log_format  # type: ignore[attr-defined]
+    server._haskell_log_entries = list(haskell_log_entries or [])  # type: ignore[attr-defined]
+    return server
+
+
+def _make_booster_server(
```
tothtamas28: These can be turned into fixtures.

pyk/src/tests/unit/kore/test_server_cli_args.py
```diff
@@ -0,0 +1,90 @@
...
+    log_axioms_file: Path | None,
+    haskell_log_format: KoreExecLogFormat = KoreExecLogFormat.JSON,
+    haskell_log_entries: list[str] | None = None,
+) -> BoosterServer:
+    server = object.__new__(BoosterServer)
```
tothtamas28: Consider using mocks of the actual classes instead that simply patch `start` into a no-op. This would let us test all other private methods as well.

tothtamas28: Alternatively, you can refactor the classes to call `start` only on `__enter__`. Then the real classes can be tested directly.

pyk/src/tests/unit/kore/test_server_cli_args.py
```diff
@@ -0,0 +1,90 @@
...
+    server._log_axioms_file = log_axioms_file  # type: ignore[attr-defined]
+    server._haskell_log_format = haskell_log_format  # type: ignore[attr-defined]
+    server._haskell_log_entries = list(haskell_log_entries or [])  # type: ignore[attr-defined]
+    return server
+
+
+class TestKoreServerHaskellLogCliArgs:
```
tothtamas28: These tests can be made top-level functions.

tothtamas28: Also consider turning them into parameterized tests.

pyk/src/tests/unit/kore/test_rpc_client.py
tothtamas28: This also looks like a `KoreClient` test.

pyk/src/pyk/cterm/symbolic.py
ehildenb: Done. abc9754

pyk/src/pyk/cterm/symbolic.py
ehildenb: Done. (abc9754)

pyk/src/pyk/cterm/symbolic.py
ehildenb: No — it's off by default. The populated default is which entries to request if logging is enabled, not a switch: logging only happens once haskell_log_dir is set (or a per-call flag turns it on). Reworded the comment to make that explicit. (abc9754)

pyk/src/pyk/cterm/symbolic.py
ehildenb: Yes, exactly. And following up on that ownership: the set is now also overridable through the cterm_symbolic factory (it previously only reached the legacy server -l path), so downstream legacy_explore/create_kcfg_explore callers can tailor it, not just directly-constructed CTermSymbolics. (f4aa11e)

pyk/src/pyk/kore/rpc.py
ehildenb: I checked _ProverPool and KoreServerPool — both key instances by current_thread().name, so each worker thread creates and owns its own KoreClient; no client is ever shared across threads. The thread-local was guarding a case that doesn't occur, so I switched last_request_id to a plain per-client attribute (recorded on JsonRpcClient, surfaced through the facade and KoreClient). (9a98324)

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Done. (abc9754)

pyk/src/tests/unit/kore/test_client.py
ehildenb: Done — the five near-identical execute/implies/simplify cases are now two endpoint-parameterized tests (requested vs. default). (96bccdc)

pyk/src/tests/unit/kore/test_rpc_client.py
ehildenb: Addressed, slightly differently than suggested: since last_request_id is now a JsonRpcClient property, the id-generation test lives directly on JsonRpcClient (no reach-through into kore_client._client._default_client, and it no longer needs the MockTransport helper), and a small KoreClient-level surfacing test moved into test_client.py. (9a98324)

pyk/src/tests/unit/kore/test_rpc_client.py
ehildenb: Removed — this was the thread-local test, and the behavior it covered no longer exists after the per-client redesign. (9a98324)

pyk/src/tests/unit/kore/test_server_cli_args.py
ehildenb: Partly: the two builders collapsed into one _server(server_cls, …) helper. I didn't make it a bare fixture since it's parameterized per case (server class, log file, format, entries). (b21dd80)

pyk/src/tests/unit/kore/test_server_cli_args.py
ehildenb: Seems like scope creep for this PR.

pyk/src/tests/unit/kore/test_server_cli_args.py
ehildenb: Done. (b21dd80)

pyk/src/tests/unit/kore/test_rpc_client.py
ehildenb: Done — single parameterized test_haskell_log_cli_args over a (server_cls, log-file, format, entries, expected) table covering both server forms and the no-file/no-entries cases. (b21dd80)

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: That fixture builds a real KoreClient over a mocked transport, whereas these tests mock the whole KoreClient to exercise CTermSymbolic in isolation. Reusing it would couple the CTermSymbolic unit tests to KoreClient/transport internals. I kept the isolated mock. Happy to extract a shared mock-KoreClient helper into a conftest if you'd prefer that over the local one.

tothtamas28: > That fixture builds a real KoreClient over a mocked transport, whereas these tests mock the whole KoreClient to exercise CTermSymbolic in isolation

Good point.

> Happy to extract a shared mock-KoreClient helper into a conftest if you'd prefer that over the local one.

No need to extract it, but turning it into an actual local fixture would be nice.

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,130 @@
...
+    # Given
+    kore_client, cts = _mock_client_cts(StuckResult(state=State(term=int_dv(2)), depth=1, logs=()))
+    dummy: CTerm = Mock()
+
+    # When the per-call flag turns logging on
+    cts.execute(dummy, haskell_logging=True)
+
+    # Then the configured entry set (the canonical default here) is what reaches KoreClient.execute
+    _args, kwargs = kore_client.execute.call_args
+    assert kwargs['haskell_logging'] == HASKELL_LOGGING_ENTRIES
```
tothtamas28: And that fixture could provide an interface similar to this, makes tests much easier to read:

```suggestion
def test_execute_on_requests_configured_entries() -> None:
    # Given a mock CTermSymbolic initialized with empty ``logs``
    cts = _mock_client_cts(logs=())
   
    # When the per-call flag turns logging on
    cts.execute(Mock(), haskell_logging=True)

    # Then the configured entry set (the canonical default here) is what reaches KoreClient.execute
    assert_client_execute_called_with_haskell_logging(HASKELL_LOGGING_ENTRIES)
```

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_execute_honors_custom_entry_set() -> None:
+    # Given a CTermSymbolic configured with a custom entry set
+    kore_client = Mock()
+    kore_client.execute.return_value = StuckResult(state=State(term=int_dv(2)), depth=1, logs=())
+    cts = CTermSymbolic(kore_client, Mock(), haskell_log_entries=['Rewrite', 'DebugApplyEquation'])
```
tothtamas28: `_mock_client_cts` can take whatever arguments `CTermSymbolic` takes (except of the `KoreClient` of course), so that these test functions collapse into a single parameterized test.

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,130 @@
...
+    assert kwargs['haskell_logging'] is None
+
+
+def _log_dir_cts(response: object, haskell_log_dir: Path) -> tuple[Mock, CTermSymbolic]:
+    kore_client = Mock()
+    kore_client.execute.return_value = response
+    cts = CTermSymbolic(kore_client, Mock(), haskell_log_dir=haskell_log_dir)
+    cts.kast_to_kore = Mock(return_value=Mock())  # type: ignore[method-assign]
+    cts.kore_to_kast = Mock(return_value=mlTop())  # type: ignore[method-assign]
+    return kore_client, cts
```
tothtamas28: Then this helper is gone as well.

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,130 @@
...
+    kore_client.last_request_id = 'proof-007'
+
+    # When
+    cts.execute(Mock())
+
+    # Then logging is requested and the bundle lands in <dir>/<request_id>.jsonl, one JSON value per line
+    _args, kwargs = kore_client.execute.call_args
+    assert kwargs['haskell_logging'] == HASKELL_LOGGING_ENTRIES
+    log_file = tmp_path / 'proof-007.jsonl'
+    assert json.loads(log_file.read_text().strip()) == entry
```
tothtamas28: Using the more powerful `_mock_client_cts` to deal with this test case:

```suggestion
def test_execute_writes_haskell_log_bundle(tmp_path: Path) -> None:
    # Given
    entry = {'context': ['Proxy'], 'message': 'simplifying'}
    log_file = tmp_path / 'proof-007.jsonl'
    
    cts = _mock_client_cts(haskell_log_dir=tmp_path)
    assume_client_execute_response_with_log_entries((entry,))
    assume_client_last_request_id('proof-007')
    
    # When
    cts.execute(Mock())

    # Then logging is requested and the bundle lands in <dir>/<request_id>.jsonl, one JSON value per line
    assert_client_execute_called_with_haskell_logging(HASKELL_LOGGING_ENTRIES)
    assert json.loads(log_file.read_text().strip()) == entry
```

pyk/src/tests/unit/cterm/test_symbolic.py
```diff
@@ -0,0 +1,130 @@
...
+            # Then the built CTermSymbolic requests exactly that set
+            assert cts._haskell_log_entries == ('Rewrite', 'DebugApplyEquation')
+
+
+def test_cterm_symbolic_defaults_to_canonical_entry_set() -> None:
+    # Given no override
+    with patch('pyk.cterm.symbolic.KoreClient'):
+        with cterm_symbolic(definition=Mock(), definition_dir=Mock(), start_server=False, port=1) as cts:
+            # Then the canonical default is used
+            assert cts._haskell_log_entries == HASKELL_LOGGING_ENTRIES
```
tothtamas28: These tests can be dropped I think? What value the private attribute takes after initialization is not a concern, only what entries there are in the `execute` request, which the previous tests cover thoroughly.

pyk/src/tests/unit/kore/test_client.py
```diff
@@ -228,13 +228,78 @@ def test_simplify(
...
     # Then
     rpc_client.assert_request('simplify', **params)
     assert actual == expected
 
 
+def test_last_request_id_surfaces_underlying_client_id(
+    kore_client: KoreClient, rpc_client: MockClient, mock: Mock
+) -> None:
+    # Given the underlying client reports an id and a normal response
+    mock.last_request_id = 'claim-x-001'
```
tothtamas28: What is `mock`?

pyk/src/tests/unit/kore/test_client.py
```diff
@@ -228,13 +228,78 @@ def test_simplify(
...
     # When
-    actual, _logs = kore_client.simplify(pattern)
+    actual, _logs, _entries = kore_client.simplify(pattern)
 
     # Then
     rpc_client.assert_request('simplify', **params)
     assert actual == expected
 
 
+def test_last_request_id_surfaces_underlying_client_id(
```
tothtamas28: I feel like all these test can be integrated into the existing harnesses, i.e. turn them into entries in the corresponding test data tables.

pyk/src/tests/unit/kore/test_client.py
```diff
@@ -228,13 +228,78 @@ def test_simplify(
...
+def _invoke_with_logging(client: KoreClient, method: str, haskell_logging: list[str] | None) -> Any:
+    match method:
+        case 'execute':
+            return client.execute(int_dv(0), haskell_logging=haskell_logging)
+        case 'implies':
+            return client.implies(int_bottom, int_top, haskell_logging=haskell_logging)
+        case 'simplify':
+            return client.simplify(And(INT, (int_top, int_top)), haskell_logging=haskell_logging)
+        case _:
+            raise AssertionError(method)
```
tothtamas28: This sort of parametrization doesn't make much sense imho.

pyk/src/tests/unit/kore/test_server_cli_args.py
tothtamas28: I can see how this can be the case for the second comment.

Regarding the first comment though: if you patch `start`, it'll eliminate the need to work around the type checker (with `__new__` and `# type: ignore` comments).

ehildenb: I just removed the tests, they didn't really test much, and this feature will be changing in the near future is my guess.

ehildenb:
@tothtamas28 I just removed all the test files. They were generated by Claude because CLaude wanted to test the code it had written. But I don't think they're needed. This is a logging feature, not something semantic, and it will likely change in the near future anyway.

pyk/src/tests/unit/kore/test_client.py
ehildenb: Removed.

pyk/src/tests/unit/kore/test_client.py
ehildenb: Removed.

pyk/src/tests/unit/kore/test_client.py
ehildenb: Removed.

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Removed.

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Removed.

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Removed.

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Removed.

pyk/src/tests/unit/cterm/test_symbolic.py
ehildenb: Removed.

tothtamas28 (approved):
Code looks good now.

Before merging, please make sure the PR title and description are cleaned up, since this is going to end up in the commit log.

PR #4930, opened 2026-06-08
Introduce `--step-timeout` option for `kprove` to ensure timely termination

### LINKED ISSUE #4924: Support progressive depth halving as a generic policy in `Prover.advance_proof`
[kontrol#1141](https://github.com/runtimeverification/kontrol/pull/1141) adds `--per-depth-timeout` to `kontrol prove`: each attempt has a stall window of `depth * per_depth_timeout` seconds in which the proof must commit at least one new node, otherwise the attempt is killed and `max_depth` is halved (down to 1). In [this review comment](https://github.com/runtimeverification/kontrol/pull/1141#issuecomment-4575606818), @ehildenb suggested factoring this into `Prover.advance_proof` so it's a generic policy rather than per-tool code.

Stevengre:
`advance_proof` now runs each step under the prover's `step_timeout` (a per-step wall-clock budget). A step that exceeds its budget is interrupted, the prover is asked to `shrink_step` (do less work per step), and the step is retried (the timed-out node was never committed, so it reappears from `get_steps`). If the prover cannot shrink further, advancement stops. Provers without a `step_timeout` run synchronously as before, so `advance_proof` stays generic and gains no new parameter.

Close https://github.com/runtimeverification/k/issues/4924.
PoC PR: https://github.com/runtimeverification/kontrol/pull/1141


ehildenb:
This looks pretty good to me. I guess I'm not an expert on whether this is the best way to handle this issue, but one question I have: when we interrupt a step, it will disconnect from that instance of the haskell backend, but does that also stop the haskell backend from working on that step at the same time? Or does the haskell backend keep working on it, we are just now trying some smaller step size ourselves?

pyk/src/tests/unit/test_advance_proof.py
```diff
@@ -0,0 +1,160 @@
...
+
+class _StepInterrupted(Exception):
+    """Raised inside `step_proof` when the prover is interrupted, mimicking a backend abort."""
+
+
+class CountingProof(Proof[int, int]):
+    """Minimal proof that needs `target` committed steps to pass."""
+
+    target: int
+    committed: int
```
ehildenb: I think this file is likely not needed, it's not really testing anything useful, and may make testing too brittle for this feature in the future. I'd recommend removing it. The integration test added should be enough.

Stevengre: Done! 11cb8ad59fc027214dc90cab0405bfd3a90efcf7

pyk/src/pyk/proof/proof.py
```diff
@@ -522,25 +552,59 @@ def advance_proof(
...
+                    for result in results:
+                        proof.commit(result)
+                    if iterations % maintenance_rate == 0:
+                        proof.write_proof_data()
+                        callback(proof)
+                if shrank_step:
+                    continue
+        finally:
+            if executor is not None:
+                executor.shutdown(wait=False)
```
ehildenb: Claude generated review with my guidance, I think this should be refactorde a bit:

● The timed path here is worth tidying. The `try/finally` only exists to `shutdown` the executor, but the substance is the inline `if timed:` block, which gives the loop body two parallel execution paths (timed vs. `else: results = self.step_proof(step)`) plus the `shrank_step` / `break` / `continue` dance threaded through the same loop. I'd extract the timed execution into one private helper so the loop body reads uniformly:

```python
results = self._execute_step(step, executor)   # None => timed out + shrank, re-fetch and retry
if results is None:
    shrank_step = True
    break
```

with `_execute_step` owning the `submit` / `result(timeout)` / `interrupt` + `shrink_step` logic. One caveat: the sentinel has to distinguish "shrank, retry" (today a `break` + `continue`) from "couldn't shrink, stop" (today a `return`), so use two distinct signals (e.g. a small `StepTimedOut`/`StepExhausted` exception pair, or a tagged result) rather than a bare `None`.

One thing to keep as-is, though: the timeout/shrink/retry policy belongs in the base `advance_proof`, not in `step_proof`. It's generic and driven entirely through the `shrink_step`/`interrupt` hooks — `ImpliesProver` overrides none of them and gets correct behavior (timeout → stop) for free. Folding it into `step_proof` would force every override to re-implement it, and would violate `step_proof`'s documented "must not modify `self`" contract (the shrink path mutates `self.execute_depth`), which is what makes it safe to run on the worker thread in the first place.

Stevengre: Done! 0c9f75ac5d8d2dddcdd970b76a64939b74036e69

ehildenb (approved):

ehildenb:
If you need to make big changes to get tests passing, then make sure to re-request reviews.

